PR #114, opened 2022-06-18
Config framework improvements

P0keDev:
- Merged `StorageHolder` and `ConfigHolder`, since having an abstract parent class doesn't seem necessary anymore
- Moved `@Config` parsing into `ConfigManager` instead of doing it in `FeatureRegistry`
- Removed `@Configurable` and `ConfigurableHolder`
- Config options can now only be within `Feature` classes, instead of any arbitrary object. The config category info is now defined in `@FeatureInfo`
- Added a config update listener to the `Feature` framework, including `userEnabled` handling for `UserFeature`
- A feature's config option holders now get stored within the feature itself, for easy feature-wise access
- Cleaned up the config command, mostly using the above addition. Also moved the value parser into `ConfigHolder` and greatly simplified it
- Added a `/wynntils config reset` command, which accepts both specific options and entire features
- Cleaned up `CustomColor` parsing, implemented `equals` so config defaults are handled correctly


magicus (approved):
LGTM

PR #115, opened 2022-06-18
Improve config setting saving

P0keDev:
This PR changes config functionality from only saving non-default settings to saving any settings that have been edited by the user, regardless of their current value. As already discussed, this prevents intentionally chosen settings from being altered by updates that modify default values - outside of this edge case, all other behavior will be the same.

`ConfigHolder` now has a `userEdited` flag, which defaults to false and is set to true any time `setValue` is called. Since the config file loading code uses `setValue`, any options stored to file (and thus, were at one point modified) will automatically have the flag set properly. If the `reset` function is called, meaning that the option has specifically been reset to the default, the flag is unset.

### DIFF
--- a/common/src/main/java/com/wynntils/core/config/ConfigHolder.java
+++ b/common/src/main/java/com/wynntils/core/config/ConfigHolder.java
@@ -19,6 +19,7 @@ public class ConfigHolder {
     private final Config metadata;
 
     private final Object defaultValue;
+    private boolean userEdited = false;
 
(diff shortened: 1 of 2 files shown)


kristofbolyai:
Does saveConfig remove the config key from the json file after a config has been reset?

P0keDev:
it will remove it, yes

kristofbolyai (approved):

magicus:
🎉

PR #117, opened 2022-06-19
Add a basic Overlay class and OverlayManager

kristofbolyai:
This PR is currently WIP, I created this so the team can see and discuss if they like this system. 

TODOs (not fully in the scope of this PR, but I may include them): 
-

- `Overlay`: This class is very bare bones at the moment, it should be improved. (This PR is focusing more on the OverlayManager and render system, the Overlay class should be detailed more in a separate PR)
- `RenderEvent`: Adding more `ElementType`s and canceling will allow us to "override" GUI parts like the health bar, necessary for some overlays.

Here is the render of the example overlay I added, called `DummyOverlayFeature`:

image


kristofbolyai:
I would like your opinion on the direction of this system, whether this is going in the right direction, or you are imagining something completely different. If you think this is a nice base, I will continue working on the TODOs listed above.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -0,0 +1,33 @@
...
+import com.mojang.blaze3d.platform.Window;
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.core.features.overlays.Overlay;
+import com.wynntils.core.features.overlays.annotations.OverlayHolder;
+import com.wynntils.core.features.overlays.annotations.RegisteredOverlay;
+import com.wynntils.mc.render.RenderUtils;
+import com.wynntils.utils.objects.CustomColor;
+
+@OverlayHolder
```
DalwynWasTaken: Same as by Config (Old Configurable), this isn't needed and you can simply get all the declared classes and then get the superclass.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -0,0 +1,33 @@
...
+public class DummyOverlayFeature extends UserFeature {
+    @RegisteredOverlay
```
DalwynWasTaken: I'll assume you want to do more with this later, otherwise this is also useless if you use the sorting method I mentioned above

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
kristofbolyai: This is used for explicit registration of overlays. Later, a feature will be able to hold Overlay, but not neccessary instantiate it.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
kristofbolyai: I mean, it is already used for stuff like render hooking and render "priority", although I should make it not limited to Pre and Post, but rather make it have a priority int.

DalwynWasTaken:
I think this is heading in the direction of having them both separated from features but allows features to own overlays. Which is good, I do believe the TODO's should be implemented at the basic level before any clear judgement can be given. 

And while you work on it, having overlay have some sort of overlayconfig field which would be the thing stored cross instance (location and if it's enabled. Everything as seen by the settings) or it could be OverlayPosition and you'd add config annotation if you want it saved. But this is more of a "keep this in mind" thing

P0keDev:
I like the setup of inner classes for overlays, but I don't think doing everything via annotations/reflection is the cleanest way to accomplish this, especially with the idea in mind of registering multiple instances of the same overlay. My immediate thought is that overlays should be explicitly registered in a feature's `init` method, which will also attach positional data (for configs) and maybe render data, although the latter could be kept as an annotation on the inner class. These overlays will be registered in a list within the feature, much like keybinds, so that when the feature is enabled or disabled, the feature can register/unregister all of its overlays with the `OverlayManager`. Here's a rough example:

```java
public class SomeFeature extends UserFeature {
    @Config(...)
    OverlayPosition position1 = new OverlayPosition(...);

    @Config(...)
    OverlayPosition position2 = new OverlayPosition(...);

    public void init() {
        registerOverlay(new SomeOverlay(), position1);
        registerOverlay(new SomeOverlay(), position2);
    }


    @Overlay(renderType = HUD) // etc
    public static class SomeOverlay extends Overlay {
        public void render(...) { ... }
    }
}
```

And then in `Feature`:

```java
    Map<Overlay, OverlayPosition> overlays = new HashMap<>();
    // ...
    protected void registerOverlay(Overlay overlay, OverlayPosition position) {
        overlays.put(overlay, position);
    }
    // ...
    public final void enable() {
        // ...
        OverlayManager.registerAll(overlays);
        // ...
    }
```

Incompleteusern:
Looking good, but I have a few complaints. I feel that `OverlayManager` should retain information about the feature that added that overlay. On that note, I think `registeredOverlays` in `OverlayManager` is unnecessary, and if a check that the registered overlay is a child class of the feature, it can be checked on initialization. If for some reason a feature needs to instantiate multiple overlays of the same kind, maybe use ids. Lastly, I feel that overlays should be created only when the parent feature is enabled/disabled.

kristofbolyai:
@Incompleteusern @P0keDev I implemented your suggestions, but the two of you had a little conflict of interest, so let me know if you don't like something.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -0,0 +1,89 @@
...
+
+        Pair<OverlayBase, OverlayPosition> overlayInfo = new Pair<>(overlay, position);
+        overlaysFeatureMap.get(registrar.getClass()).add(overlayInfo);
+
+        if (overlay.getClass().getAnnotation(Overlay.class).enabled()) {
+            instantiateOverlay(overlayInfo);
+        }
+    }
+
+    private static void instantiateOverlay(Pair<OverlayBase, OverlayPosition> overlayInfo) {
```
P0keDev: What's the point in this function when you're already passing an instance of OverlayBase? You don't need to create a new instance

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -74,6 +64,11 @@ private static void initializeFeature(Feature feature) {
             return;
         }
 
+        // OverlayManager event listener setup
+        if (featureClass.isAnnotationPresent(OverlayFeature.class)) {
```
P0keDev: If we're directly registering overlays in a feature's init, there's no need to have this annotation and additional handling, right?

P0keDev (commented):
Personally I would prefer it if overlay registration paralleled keybinds in its design (features keep track of their own overlays, register/unregister them to the central handler as needed). `OverlayManager` could keep track of overlays' "parent" features in addition to this, if that's also important. All up for debate, however

I would also advocate for renaming the overlay stuff to parallel the names of our feature framework, i.e. `Overlay` -> `OverlayInfo` and `OverlayBase` -> `Overlay`

magicus:
Overall, I think this is definitely heading in the right direction. I'll add some specific remarks in the code shortly, but let me first discuss one overall thing. 

I think the code is getting unnecessary complex due to the idea that enabling overlays mean instantiating them. Instead, I think we should treat overlays like features. Always instantiate them in the beginning. To enable/disable an overlay really just means to move it on or off the list of enabled overlays, which is traversed when calling the `render` methods on overlays. There is no cost associated with having an instance of an overlay that is created but not being called, so we should not be afraid of this. So enabling/disabling really just means shuffling instances from `enabledOverlays` to `disabledOverlays` or back again.

Second, while I originally championed the idea that we should allow a feature to have multiple instances of an overlay, I also must stress that this is a special case scenario (only info boxes seem to be relevant atm), and it's better if that part gets a bit messier, if it can make the normal case of one overlay instance per Overlay class simpler. I'm thinking something like this. Instead of registering OverlayPositions as `@Config`, lets do something like this:
```
public class SingleOverlayFeature extends UserFeature {
  @Overlay(name = "Fancy overlay", renderType = RenderEvent.ElementType.GUI)
  private Overlay fancyOverlay = new SingleOverlayFeature.FancyOverlay();

  public static class FancyOverlay extends Overlay {
      // ...
  }
}

public class MultiOverlayFeature extends UserFeature {
  @Overlay(name = "Tribbles overlay", renderType = RenderEvent.ElementType.GUI)
  private List<Overlay> tribbleOverlays = new ArrayList<SingleOverlayFeature.TribblesOverlay>();

  // ...
}
```

That is, the `@Overlay` annotation supports either an Overlay class directly, or a List of Overlays. In the latter case, we probably need to notify the OverlayManager whenever the list is updated. But, once again, since this is a special case, I think we can live with that. It's just two simple calls, whenever we add or remove overlays on the list.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
magicus: I think it's better to have purely annotations in Feature. We kind of determined that direction with how we handle Events, KeyPresses and Configs. Any deviation from that now will just make the code look and feel incoherent.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -189,94 +179,5 @@ public Object generate() {
...
-
-            // McIf.mc().getTextureManager().bindTexture(ICONS);
-        }
-
-        @SubscribeEvent
-        public static void onRenderPost(RenderEvent.Post e) {
-            if (!WynnUtils.onServer()) // || !McUtils.mc().playerController.isSpectator())
-            return;
-
-            McUtils.mc().getProfiler().push("postRenOverlay");
```
magicus, line 263: Do you think you can keep the profiler wrappers in the new overlay render code? I think it is nice to have it in place from the beginning, since overlay rendering is one of the ways that Wynntils can have unfortunately large impacts on game performance (rendering is expensive and is done multiple times per second), and it is easy to create bad performance overlays without realizing it.

magicus:
> I would also advocate for renaming the overlay stuff to parallel the names of our feature framework, i.e. `Overlay` -> `OverlayInfo` and `OverlayBase` -> `Overlay`

I second that. It really fits in better with what we have, and feels like more natural descriptions.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class OverlayManager {
+    private static final Map<OverlayBase, OverlayPosition> overlayInstanceMap = new HashMap<>();
+
+    private static final Map<Class<? extends Feature>, List<Pair<OverlayBase, OverlayPosition>>> overlaysFeatureMap =
```
magicus: If you heed my advice and make the OverlayManager only handle instances of Overlays, maybe this point will be moot, but there is some kind of redundancy here. 

If you have a separate mapping between an OverlayBase and a position, you don't also need to include the position in the mapping from feature to list of overlays. So the overlaysInstanceMap is really a "overlayPositionMap", mapping from overlay instance to position (which is the "configy" stuff about the placement), and "overlaysFeatureMap" (maybe skip the plural s?) is a map from Feature (class, as chosen here, or possibly instance instead?) to  a List of OverlayBase instances that the Feature holds.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -0,0 +1,89 @@
...
+            // Prevent duplication from default enabled Overlays
+            if (overlayInstanceMap.keySet().stream()
+                    .anyMatch(overlayBase -> overlayPositionPair.a.getClass().isInstance(overlayBase))) continue;
+            instantiateOverlay(overlayPositionPair);
+        }
+    }
+
+    @SubscribeEvent
+    public static void onRenderPre(RenderEvent.Pre event) {
+        renderOverlays(event, Overlay.RenderState.Pre);
```
magicus: Maybe I'm not knowledgeable enough about rendering, but what's the purpose of having both a pre and a post render? Would not a post render be enough? 

The current implementation is buggy, anyway. It just calls renderOverlays twice, and while it passes the Pre or Post renderState enum, this is just ignored, and the only effect is that the `render` method of the overlays are called twice.

My suggestion is that we drop the pre render part for now, and only if we show that there are overlays that really need that (for what? disabling vanilla elements?) we can add a pre-render method for them later on.

kristofbolyai: Oh, it didn't use to be ignored, I  removed the check by accident.

magicus: OK, that will solve the buggy part. :-) But can you give an example on why/where we need the pre-render event?

kristofbolyai: Pre render can be used to override Minecraft HUD like health bar, by making the event cancellable (which is easy to do, but this PR just doesn't do it), and using RenderEvent.ElementType. Post event can be used to draw on existing HUD elements.

kristofbolyai:
Third iteration! Tried to do everything as requested, although @magicus, you did not mention how you would store OverlayPosition, so I bundled it into the Overlay class. Overall the code is now simpler now. As requested by @P0keDev, the Feature class now holds information about its own overlays.

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -30,20 +36,39 @@
...
 
-        if (this.conditions.isEmpty()) return;
-        this.conditions.forEach(Condition::init);
+        if (!this.conditions.isEmpty()) this.conditions.forEach(Condition::init);
+    }
+
+    private void initOverlays() {
+        Field[] overlayFields = FieldUtils.getFieldsWithAnnotation(this.getClass(), OverlayInfo.class);
+        for (Field overlayField : overlayFields) {
+            if (overlayField.getType() != Overlay.class) continue;
```
magicus: Maybe log an error to catch mistakes in the code? Or even throw a runtime exception.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -4,41 +4,30 @@
...
-        this.visible = visible;
-        this.hookElements = hookElements;
-        this.growthX = growthX;
-        this.growthY = growthY;
+import com.mojang.blaze3d.platform.Window;
+import com.mojang.blaze3d.vertex.PoseStack;
+
+public abstract class Overlay {
+
+    public Overlay(OverlayPosition position, float width, float height) {
```
magicus, line 16: I think all fields should come before constructors and other methods in a class. In fact, I would have wished for Spotless to enforce that, but apparently it does not.

kristofbolyai: I will change the order, and also check out if there is a spotless "hack" for this.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public static void registerOverlay(Overlay overlay, OverlayInfo overlayInfo) {
+        overlayInfoMap.put(overlay, overlayInfo);
+
+        if (overlayInfo.enabled()) {
+            enabledOverlays.add(overlay);
+        }
+    }
+
+    public static void disableAllOverlaysForFeature(List<Overlay> overlays) {
```
magicus: I think the method should be renamed to just `disableOverlays`. (And similar for enable)

An alternative would be to actually send in a Feature, but then the logic of extracting the set of overlays in the feature need to be either moved here, or exposed in the Feature class, and neither seems really good.

And it is more general to just take a list of overlays and disable them, as you've done.

kristofbolyai: Yeah, I think it was my purpose to rename it but after I changed up the logic, I forgot.😅

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -0,0 +1,66 @@
...
+        McUtils.mc().getProfiler().pop();
+    }
+
+    private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState renderState) {
+        for (Overlay overlay : enabledOverlays) {
+            OverlayInfo annotation = overlayInfoMap.get(overlay);
+
+            if (renderState != annotation.renderAt() || event.getType() != annotation.renderType()) continue;
+
+            overlay.render(overlay.getPosition(), event.getPoseStack(), event.getPartialTicks(), event.getWindow());
```
magicus, line 59: When I see this code, I'm starting to think that a perhaps even simpler approach would be to store an `boolean enabled` in the Overlay class, and turn it on and off when enabling/disabling, and just skip the overlay in this for loop if it's not enabled. Otoh, it does not matter that much, and the current implementation with `enabledOverlays` is fine. Change it if you think it would be a nicer pattern, to store the knowledge about denabled/disabled in the Overlay itself, otherwise keep it if you think this is a responsibility of the Manager to know about.

kristofbolyai: Easier to get all enabled Overlays with this approach. Also with this style, we follow the "overlay doesn't have a understanding of state, and is only an abstraction of rendering on screen with specific settings", which is how I  think about overlays.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
magicus: I think I understand what P0ke means now. Or maybe that's not what he means, but that's what I suddenly see as a problem. :-)

We don't need the `@OverlayFeature` annotation, as far as I can see. If a feature has Overlay fields with the `@OverlayInfo` annotation, that is enough. We don't have to tag the class as well. That's just confusing, and adds no value, we only risk that we miss that we miss it and the Overlays do not work. If a class has Overlays with OverlayInfo, then it is an "overlay feature", otherwise not.

(The thing that got me confused is that the P0ke was talking about registering in the init function, but actually that registering is based on annotations as well, just that it is field annotations instead of class annotations. So I misunderstood this comment as to mean we should go back to some kind of call "initOverlays()" in the Feature, which I then argued against :-D)

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -0,0 +1,68 @@
...
+import com.wynntils.core.features.overlays.annotations.OverlayFeature;
+import com.wynntils.core.features.overlays.annotations.OverlayInfo;
+import com.wynntils.mc.event.RenderEvent;
+import com.wynntils.mc.render.RenderUtils;
+import com.wynntils.utils.objects.CustomColor;
+
+@OverlayFeature
+public class DummyOverlayFeature extends UserFeature {
+
+    @SuppressWarnings("UnusedDeclaration")
```
magicus: Why do we need the `@SuppressWarnings`? That seems a bit unfortunate. :-( 

Ah, I see, we create a field that is not accessed. Maybe we need to change the access modifier. Make it default (empty) or public.

kristofbolyai: I don't think changing access modifiers would help, I think we can live without the SuppressWarning tough, I only added it because IDEA complained.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -0,0 +1,68 @@
...
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay DummyRedBoxOverlay = new DummyRedBoxOverlay();
+
+    @SuppressWarnings("UnusedDeclaration")
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI, renderAt = OverlayInfo.RenderState.Pre)
+    private final Overlay DummyBlueBoxOverlay = new DummyBlueBoxOverlay();
```
magicus: For trivial overlays like this, that do not need to have any internal state etc, we should perhaps provide a simple generic `BasicOverlay` (or whatever), that takes a method handle to a render method. Then we could just create it like:
```
    @OverlayInfo(renderType = RenderEvent.ElementType.GUI, renderAt = OverlayInfo.RenderState.Pre)
    final Overlay dummyBlueBoxOverlay = new BasicOverlay(new OverlayPosition(10, 10), 50, 50, DummyOverlayFeature::renderBlueBox);

    public static void renderBlueBox(OverlayPosition overlayPosition, PoseStack poseStack, float partialTicks, Window window) {
        // ...
    }
```

I still we should keep the ability to override Overlay for more complex cases, if a simple, single render method is not enough, e.g. if we need to keep some local state for the overlay. Then an inner class is ideal.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
magicus: Ah, it's just IDEA? I thought it was the normal javac. 

IDEA can be told to consider annotations as accesses. I know we did that for `@SubscribeEvent` methods in legacy, otherwise they seemed to be never called. Don't remember how to do it though...

magicus:
I think this is starting to look pretty sweet! It feels like we're almost there, at the Perfect Overlay Design. :-)

@kristofbolyai I think bundling the position in the Overlay class was exactly the right decision. I just have three issues surrounding the overlay position thing.

1) We need to get back Dalwyns idea of storing the position as a configuration item into this framework. I'm not sure if we can mis-use the `@Config` annotation on the Position field in the Overlay, or (more likely) if we need to do some custom adaptation on configurations to get this to work smoothly. (We need some nice name for the configs, like `<feature>.<overlay>.[x|y]`, and I don't think we can generate that currently).

2) How should we do the initial, default position? It's one thing that if the user has dragged around the overlay box that we persist their choice. But it seems sub-optimal to hard-code the initial position as a pair of integers in the constructor of each Overlay. 

So even though I think an Overlay should hold it's position, we could perhaps allow it to be null, and not set from the constructor. And then maybe the OverlayManager can have some way of assigning default positions for the overlays? After all, it's a bit of a "holistic" system wide thing, to get a proper default where different overlays have good default that work with each other.

3) Right now the assumtion is that an Overlay has a fixed width and height. I think we will need to challenge this assumption rather soon, to allow resizable Overlays. We might consider preparing for this by storing the size not as a width and height field, but as an OverlaySize, similar to the OverlayPosition. An OverlaySize class can be subclassed with a FixedOverlaySize, or an RestrictedRangeOverlaySize, etc, to cater for different needs from different overlays. And if the OverlaySize has user changable values, they too need to be persisted in the config file. 

But we can also accept this PR as it is now, and continue working on these aspects as follow up issues.

kristofbolyai:
I would separate Overlay config saving into a PR, and then advanced Overlay characteristics into another one.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
magicus: I accept this. But with just a clarification: I think overlays can have some understanding of state, but that relates to "model state" (like the set of messages to show in the message pane). It does now know if it is shown or not, it just renders itself if asked to.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
magicus: @kristofbolyai Okay. I'm just thinking such a thing needs to both do a pre-render to cancel the Vanilla part, and then a post-render to setup our own health bar. The current solution only allows an Overlay to have one of each type. But, that is petty details. We can sort that out when we actually start to implement such Overlays. (My gut feeling is that we will end up with having a preRender and postRender method on Overlays instead of tagging them in OverlayInfo. But it's inconsequential at this point.)

kristofbolyai: How come we can't cancel the Pre event and also render our own thing?

common/src/main/java/com/wynntils/core/features/Feature.java
magicus: Actually, I realized you also need to look for `List.class` here, to be able to support the use case of multiple overlays, and iterate over the elements of that list, adding the overlays. We can do that later on as well, so long as we're not forgetting it.

kristofbolyai: List support is not currently added, I will work on it separately.

magicus:
@kristofbolyai I'm perfectly fine with that. Sometimes it's easier to have a checkpoint with a PR committed, and then you continue to refine details separately. Feel free to postpone any of my comments to a later PR, as long as they are adressed in some way or another later on. :-)

magicus (dismissed):

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
magicus: I tested around a bit. You can get rid of the warning by clicking on "Show quick fixes" and selecting "Suppress unused warning if annotated by ...". I don't know if we can set this as some kind of project default that we can distribute with the source. Possibly, but I don't know exactly how.

magicus: It's stored in .idea/misc.xml as a `<component name="EntryPointsManager">`. I think we can create a checked-in version of this data in some way. At least my local misc.xml contains stuff like net.minecraftforge.fml.common.Mod.EventHandler that I don't remember having added.

magicus: Sadly, the only choice seem to be to include the entire .idea/misc.xml in the git repo. :( I think this is a bad idea. But you can export an example config that can be imported into IntelliJ. I think that's the best thing we can do, or describe in a README how to get rid of those warnings yourself.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -162,12 +147,11 @@ public static void init() {
         registerFeature(new WynncraftButtonFeature());
         registerFeature(new TooltipScaleFeature());
         registerFeature(new DurabilityArcFeature());
+        // registerFeature(new DummyOverlayFeature()); TODO: We should have a way to register features with disabled state
```
magicus: I thought DebugFeatures did just that..?

Incompleteusern: DebugFeatures should only enabled for dev environment, this should be never be

kristofbolyai: We could move this into an examples folder and not register it?

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -0,0 +1,60 @@
...
+public class DummyOverlayFeature extends UserFeature {
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay ComplexRedOverlay = new DummyRedComplexOverlay();
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI, renderAt = OverlayInfo.RenderState.Pre)
+    private final Overlay BasicBlueOverlay =
+            new BasicOverlay(new OverlayPosition(50, 50), 75, 75, DummyOverlayFeature::renderBasicBlueBox);
+
+    @Override
```
magicus: Not that it really matters, but this seems a bit of a left-over :)

magicus (dismissed):
This looks soooo epic! 🥳

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
magicus: Whatever, it does not matter so much. I imagine we can remove this thing as soon as we start getting real examples of the Overlay system. I'm fine with a commented out line like this too.

magicus: I just discovered that we have an `@StartDisabled` annotation for Features. I assume this does what you wanted? :-D

magicus (approved):

PR #118, opened 2022-06-20
Chat item implementation

P0keDev:
This PR ports legacy's chat item feature to Artemis. The encoding format used is identical, so there is full cross-compatibility with legacy. 

image

To fully accomplish the feature, I created/changed the chat sent and received events. Their respective mixins use redirects now - I'm aware that this is a heavy-handed approach and should be avoided when possible, but the only other way of implementing the ability to modify received messages from multiple different features (which is not something that happens now, but most likely will in the future) would be to have some sort of central chat manager that can simply cancel the chat message and display its own. This is something that probably _should_ happen down the road, but I didn't want to undertake that within this PR.

### DIFF
--- a/common/src/main/java/com/wynntils/features/internal/ChatItemFeature.java
+++ b/common/src/main/java/com/wynntils/features/internal/ChatItemFeature.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.internal;
+
+import com.wynntils.core.features.InternalFeature;
+import com.wynntils.core.features.properties.EventListener;
+import com.wynntils.mc.event.ChatReceivedEvent;
+import com.wynntils.mc.event.KeyInputEvent;
+import com.wynntils.mc.mixin.accessors.ChatScreenAccessor;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wc.custom.item.GearItemStack;
+import com.wynntils.wc.utils.ChatItemUtils;
+import com.wynntils.wc.utils.WynnUtils;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.gui.components.EditBox;
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@EventListener
+public class ChatItemFeature extends InternalFeature {
+
+    private final Map<String, String> chatItems = new HashMap<>();
+
+    @SubscribeEvent
+    public void onKeyTyped(KeyInputEvent e) {
+        if (!WynnUtils.onWorld()) return;
+        if (!(McUtils.mc().screen instanceof ChatScreen chatScreen)) return;
+
+        EditBox chatInput = ((ChatScreenAccessor) chatScreen).getChatInput();
+
+        if (!chatItems.isEmpty() && (e.getKey() == GLFW.GLFW_KEY_ENTER || e.getKey() == GLFW.GLFW_KEY_KP_ENTER)) {
+            // replace the placeholder strings with the actual encoded strings
+            for (Map.Entry<String, String> item : chatItems.entrySet()) {
+                chatInput.setValue(chatInput.getValue().replace("<" + item.getKey() + ">", item.getValue()));
+            }
+            chatItems.clear();
+            return;
+        }
+
+        // replace encoded strings with placeholders for less confusion
+        Matcher m = ChatItemUtils.ENCODED_PATTERN.matcher(chatInput.getValue());
+        while (m.find()) {
+            String encodedItem = m.group();
+            String name = m.group("Name");
+            while (chatItems.containsKey(name)) { // avoid overwriting entries
+                name += "_";
+            }
+
+            chatInput.setValue(chatInput.getValue().replace(encodedItem, "<" + name + ">"));
+            chatItems.put(name, encodedItem);
+        }
+    }
+
+    @SubscribeEvent
+    public void onChatReceived(ChatReceivedEvent e) {
+        if (!WynnUtils.onWorld()) return;
+
+        Component message = e.getMessage();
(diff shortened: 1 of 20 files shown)


Incompleteusern (commented):
What is the difference between `customName` vs `itemName`? Also ChatItemFeature should probably have a webloadedcondition.

P0keDev:
> What is the difference between `customName` vs `itemName`? Also ChatItemFeature should probably have a webloadedcondition.

`itemName` is the original name of the item - generally, the standard wynncraft name w/ formatting but no % marker, but when we're creating item stacks from nothing, it's just the regular vanilla name like Diamond Helmet or whatever

Incompleteusern (dismissed):

magicus:
I like the idea of a central ChatManager. We're doing so much with the chat, it makes sense to have a central place to coordinate all of our meddling.

common/src/main/java/com/wynntils/features/internal/ChatItemFeature.java
```diff
@@ -0,0 +1,124 @@
...
+            // replace the placeholder strings with the actual encoded strings
+            for (Map.Entry<String, String> item : chatItems.entrySet()) {
+                chatInput.setValue(chatInput.getValue().replace("<" + item.getKey() + ">", item.getValue()));
+            }
+            chatItems.clear();
+            return;
+        }
+
+        // replace encoded strings with placeholders for less confusion
+        Matcher m = ChatItemUtils.ENCODED_PATTERN.matcher(chatInput.getValue());
```
magicus: We should not depend on a feature like this from the outside. But if this is temporary, and we will later turn this into a chat manager, where the ChatItemUtils could register a "chat transformer" or "chat processor" or whatever, then it is okay for now. But, to emphasize this "incorrect" dependence, could you at least extract this replacement into a separate method and put it in ChatItemUtils?

common/src/main/java/com/wynntils/features/internal/ChatItemFeature.java
```diff
@@ -0,0 +1,124 @@
...
+    public void onChatReceived(ChatReceivedEvent e) {
+        if (!WynnUtils.onWorld()) return;
+
+        Component message = e.getMessage();
+        List<MutableComponent> components =
+                message.getSiblings().stream().map(Component::copy).collect(Collectors.toList());
+        components.add(0, message.plainCopy().withStyle(message.getStyle()));
+
+        // chat item tooltips
+        if (ChatItemUtils.ENCODED_PATTERN.matcher(message.getString()).find()) {
```
magicus: Same here, please extract this part to a method in ChatItemUtils.

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
```diff
@@ -121,5 +125,19 @@ private static void takeScreenshot(Slot hoveredSlot) {
         McUtils.sendMessageToClient(
                 new TranslatableComponent("feature.wynntils.itemScreenshot.message", stack.getHoverName())
                         .withStyle(ChatFormatting.GREEN));
+
+        // chat item
```
magicus: I'm not sure I get this. If you screenshot an item, you also send it to the chat..?

I think some comments describing the purpose of this block would be great.

common/src/main/java/com/wynntils/wc/utils/ChatItemUtils.java
```diff
@@ -0,0 +1,247 @@
...
+     * Identifications are encoded either as the raw value minus the minimum value of that ID, or if the range is larger than 100,
+     * the percent value 0 to 100 of the given roll.
+     * Regardless of either case, this number is multiplied by 4, and the number of stars present on that ID is added.
+     * This ensures that the value and star count can be encoded into a single character and be decoded later.
+     *
+     * Powders are encoded as numerical values 1-5. Up to 4 powders are encoded into a single character - for each new powder,
+     * the running total is multiplied by 6 before the new powder value is added. Thus, each individual powder can be decoded.
+     *
+     * Rerolls are simply encoded as a raw number.
+     *
```
magicus, line 70: Maybe add something that this format is defined by Wynntils Legacy and can't be changed for backwards compatibility?

common/src/main/java/com/wynntils/mc/mixin/accessors/ChatScreenAccessor.java
```diff
@@ -0,0 +1,16 @@
...
+package com.wynntils.mc.mixin.accessors;
+
+import net.minecraft.client.gui.components.EditBox;
+import net.minecraft.client.gui.screens.ChatScreen;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+@Mixin(ChatScreen.class)
+public interface ChatScreenAccessor {
+    @Accessor("input")
```
magicus, line 14: Where are these two accessors used? I can't find any place where they are referenced.

magicus (commented):
I think this overall looks good. I've left a bunch of random remarks, questions or request. 

But I got really turned on by the tought of a ChatManager. :-) It would be nice if it could have things like `registerMessageToIgnore(String regex)`, `registerChatToMessageOverlay(String regex)` etc to make it easy to hijack certain wynn messages. And have a central point for separating the chat stream into separate kinds of messages -- global shouts, local player chat, NPCs talking, Wynn feedback messages, etc. We had some kind of tabbed chat interface in Legacy, but I never figured out how to get it to work to fit my needs.

common/src/main/java/com/wynntils/mc/mixin/accessors/ChatScreenAccessor.java
P0keDev: ChatItemFeature line 52 & ChatItemUtils line 212

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
P0keDev: The screenshot button essentially serves the double duty of creating a literal image screenshot, as well as providing the option to share via chat instead, if applicable to the item in question. This is how it functions in legacy

P0keDev:
If we do end up creating a custom chat interface (a worthwhile feature, imo), some kind of chat manager system could be built into it pretty easily. Definitely a big undertaking though

magicus:
I'm tempted at starting building an initial ChatManager that will just allow for hiding of some messages (like the Wynncraft public announcements), and classifying message types, so that it is e.g. possible to turn on/off local chat. That does not sound like a too big project. A custom chat interface otoh, that's a whole other story...

common/src/main/java/com/wynntils/wc/utils/ChatItemUtils.java
Incompleteusern: I mean, we could change it in legacy and artemis both if we wanted to.

kristofbolyai (dismissed):

kristofbolyai (approved):
Seems fine

PR #119, opened 2022-06-21
WynntilsCommand: Add /wynntils feature enable/disable

kristofbolyai:
Only works with UserFeature, as expected.

### DIFF
--- a/common/src/main/java/com/wynntils/commands/wynntils/WynntilsFeatureCommand.java
+++ b/common/src/main/java/com/wynntils/commands/wynntils/WynntilsFeatureCommand.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.commands.wynntils;
+
+import com.google.common.base.CaseFormat;
+import com.mojang.brigadier.arguments.StringArgumentType;
+import com.mojang.brigadier.context.CommandContext;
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/commands/wynntils/WynntilsFeatureCommand.java
```diff
@@ -0,0 +1,148 @@
...
+
+        if (feature.isEnabled()) {
+            context.getSource()
+                    .sendFailure(new TextComponent("Feature " + feature.getShortName() + " is already enabled!")
+                            .withStyle(ChatFormatting.RED));
+            return 1;
+        }
+
+        feature.setUserEnabled(true);
+        feature.tryUserToggle();
```
P0keDev, line 88: there should be a call to `saveConfig` here

common/src/main/java/com/wynntils/commands/wynntils/WynntilsFeatureCommand.java
```diff
@@ -0,0 +1,148 @@
...
+
+        if (!feature.isEnabled()) {
+            context.getSource()
+                    .sendFailure(new TextComponent("Feature " + feature.getShortName() + " is already disabled!")
+                            .withStyle(ChatFormatting.RED));
+            return 1;
+        }
+
+        feature.setUserEnabled(false);
+        feature.tryUserToggle();
```
P0keDev, line 134: here also

P0keDev (commented):
Looks good, just needs config saving

P0keDev (approved):

PR #122, opened 2022-06-23
 Make config options translatable

magicus:
As the title say. This PR also includes automatic generation of the translation key, so the code looks very sleek.

Note that this PR is based on https://github.com/Wynntils/Artemis/pull/121 and contains that as well, so the true changed brought by just this PR can only be seen once #121 has been integrated.

I'm not 100% happy with the solution for UserFeature#enabled. :( The way of settting a key in the Config is really just a hack. Ideally the system should be able to look up the correct name even if the config is inherited. But I got stuck trying to get that to work; I could find the declaring class, but I needed an instance to call `getTranslation()`.

Or. Maybe this is more evidence in favor of my position that the "enabled" switch is very special, and should not really part of the feature itself, but a part of the feature manager. In a way, it gets a bit mind staggering trying to think of what inherited config options really mean.


magicus:
Also, we should probably add some verification code that checks if we have translation strings for all feature names and all configs name/descriptions...

P0keDev:
For `userEnabled`, would it be possible to use a `%s` string reference in the name? Then `ConfigHolder#getDisplayName` could pass the name of `parent` into the translation - assuming the argument would get ignored for any translation without references

magicus:
@P0keDev It's certainly *possible*, but I also must say it feels a bit contrived. For now, I think the current change is OK. I imagine we will find out what to really do about `enabled` when we start doing the config GUI.

P0keDev:
That's true - we'll probably have specific handling for `userEnabled` that might not even use the display name. I was thinking that each feature would have its own "box" with the feature's name and a toggle switch - but those details are for another PR

P0keDev (dismissed):

Incompleteusern (dismissed):

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -57,6 +58,20 @@ public Config getMetadata() {
         return metadata;
     }
 
+    public String getDisplayName() {
+        if (!getMetadata().key().isEmpty()) {
+            return Language.getInstance().getOrDefault(getMetadata().key() + ".name");
```
kristofbolyai: Why not use `I18n#get`? It's simpler, and also has formatting support (not that it is needed this case).

magicus: I can't find this `I18n`. Which package/lib is it in?

kristofbolyai: `net.minecraft.client.resources.language` -> `I18n`

kristofbolyai: (its basically what you are doing but shorter)

kristofbolyai (approved):

PR #123, opened 2022-06-23
WIP: Create ChatManager

magicus:
I'm playing around a bit with what a ChatManager can, and should, do. Nothing is final, but if you have feedback about the current implementation, please let me know.


common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,59 @@
...
+
+        switch (e.getType()) {
+            case GAME_INFO -> parseGameInfo(e.getMessage());
+            case SYSTEM -> parseSystemMessage(e.getMessage());
+            case CHAT -> parseChatMessage(e.getMessage());
+        }
+    }
+
+    private static void parseChatMessage(Component message) {
+        // this is a real chat message, or wynntils stuff like
```
kristofbolyai: Wynntils stuff or Wynncraft stuff? I assume the latter?

common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public static void onChatReceived(ChatReceivedEvent e) {
+        if (!WynnUtils.onWorld()) return;
+
+        switch (e.getType()) {
+            case GAME_INFO -> parseGameInfo(e.getMessage());
+            case SYSTEM -> parseSystemMessage(e.getMessage());
+            case CHAT -> parseChatMessage(e.getMessage());
```
kristofbolyai: Perhaps a less universal name? (although I cant suggest a better one)

common/src/main/java/com/wynntils/core/chat/ChatManager.java
P0keDev: same with the method directly below this one

P0keDev:
Not necessarily within the scope of this PR, but what is the point/advantage of separating our character & world models into an interface and an implementation? This doesn't seem like an appropriate design when I can't envision a reason to ever have more than one class implementing each interface

(This is something that I've wondered for a while and was reminded of it after looking through this PR)

magicus:
@P0keDev You are right that this seems to mostly bring in uneccessary boilerplate. I will argue that there is a point in separating interface and implementation, even if there is only a single implementation, since it makes it clear what operations are "external" and supposed to be accessed from the outside. For the models, we also have public methods that are annotated callbacks, and I wanted to "hide" them.

But it was probably not worth the extra layer of abstraction. I'm removing this in https://github.com/Wynntils/Artemis/pull/145.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
magicus: Which is too universal you mean? `getType()`, or any of the enum values, or `parseChatMessage()`?

kristofbolyai: CHAT.

PR #127, opened 2022-07-01
ChatItemUtils: Bugfix + nitpicks

kristofbolyai:


P0keDev (approved):

common/src/main/java/com/wynntils/wc/utils/ChatItemUtils.java
```diff
@@ -271,29 +272,29 @@ private static Component createItemComponent(GearItemStack item) {
...
     private static String decodeString(String text) {
-        String decoded = "";
+        StringBuilder decoded = new StringBuilder();
         for (int i = 0; i < text.length(); i += 2) {
             int value = text.codePointAt(i) - OFFSET + 32;
-            decoded += (char) value;
+            decoded.append((char) value);
         }
-        return decoded;
+        return decoded.toString();
```
magicus, line 293: Actually modern Hotspot is very good at rewriting string operations to use StringBuilder under the hood, so in most cases you can keep using the easy-to-read syntax of string manipulation, and get StringBuilder performance for free.

PR #130, opened 2022-07-02
Overlays: Advanced positioning and sizing, support for overriding Minecraft GUI elements

kristofbolyai:
This PR adds: 

- [x] Advanced overlay positioning: This is achieved by having the screen divided into ninths and anchoring overlays into one ninth. Overlays also have vertical/horizontal alignment and offset values, making them fully customizable. 
- [x] Advanced overlay sizing: `GuiScaleRespectingOverlaySize` `RestrictedRangeOverlaySize` and `FixedOverlaySize`.
- [x] Support for overriding Minecraft GUI elements: This is easily achieved by already existing code, not much change was needed.  An example overriding the default crosshair was added.

Example of correct positioning (ignore the colors, those are my debug ninth renders):

Fullscreen:
image

Smaller window:
image

Different GUI scale (blue overlay respects GUI scale, red is fixed): 
image


common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -8,47 +8,106 @@
...
     @OverlayInfo(renderType = RenderEvent.ElementType.GUI, renderAt = OverlayInfo.RenderState.Pre)
-    private final Overlay BasicBlueOverlay =
-            new BasicOverlay(new OverlayPosition(50, 50), 75, 75, DummyOverlayFeature::renderBasicBlueBox);
+    private final Overlay BasicBlueOverlay = new BasicOverlay(
+            new OverlayPosition(
+                    15,
+                    15,
+                    OverlayPosition.VerticalAlignment.Top,
+                    OverlayPosition.HorizontalAlignment.Left,
+                    OverlayPosition.AnchorNinth.TopLeft),
```
P0keDev: minor detail for now, but this should be defined in its own field in the feature class so it can be collected as a config option

P0keDev: actually maybe it's a better idea to just register them from `OverlayManager`, since we always know exactly what we're dealing with and what needs to be registered - no need to deal with annotation handling

this would require a special method in `ConfigManager` but that's fine

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -4,24 +4,80 @@
...
 
-// TODO: This class is very basic atm
 public class OverlayPosition {
-    // Top-Left point of Overlay
-    private int x;
-    private int y;
 
-    private OverlayPosition() {}
+    private final int verticalOffset;
+    private final int horizontalOffset;
```
P0keDev: i'm not sure if all the fields here should be final - having to create a new position object and reassign it in the right places seems unnecessary. plus, if we intend on a gui placement screen that allows users to drag overlays around, being able to update these values in real-time and use them for rendering in that screen as well

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/GuiScaleRespectingOverlaySize.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.features.overlays.overlaySizes;
+
+public class GuiScaleRespectingOverlaySize extends OverlaySize {
```
P0keDev: not a fan of how wordy this name is; maybe `GuiScaledOverlaySize`?

P0keDev (commented):
some small comments after a quick look over - implementation looks good so far

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
kristofbolyai: What do you mean? Do you want to have OverlayPosition fields declared in the feauture, with @Config annotation and having those passed to the Overlay constructors?

kristofbolyai: Also, why can't we just look for Overlays and get the position field for them and load those?

P0keDev: > What do you mean? Do you want to have OverlayPosition fields declared in the feauture, with @config annotation and having those passed to the Overlay constructors?

that was my original thought yeah, but then i realized there's better ways of accomplishing what we need


> Also, why can't we just look for Overlays and get the position field for them and load those?

not sure what you mean by "look for", since we already handle/register each overlay in `OverlayManager` - we'd just need a bit of extra code to extract the position field (which is easy since we know exactly what it is, no annotation searching required) and then register it via `ConfigManager`

kristofbolyai: Yeah, I agree. So, if we understand each other correctly, no change needed atm? (and we will get back to this in the overlay config saving pr)

P0keDev: yeah, no need to implement it with this PR

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -6,35 +6,54 @@
 
 import com.mojang.blaze3d.platform.Window;
 import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.features.overlays.overlaySizes.FixedOverlaySize;
+import com.wynntils.core.features.overlays.overlaySizes.OverlaySize;
+import com.wynntils.utils.objects.Pair;
 
 public abstract class Overlay {
-    protected OverlayPosition position;
+    protected OverlayPosition overlayPosition;
```
magicus: I dunno, since we're in an `Overlay` already, is not `overlayPosition` just longer and redundant? Even if the type is `OverlayPosition` I think the variable can be just `position`. (And the same with overlaySize -> size).

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -4,24 +4,80 @@
...
+
+    private AnchorNinth anchorNinth;
+
+    public OverlayPosition(
+            int verticalOffset,
+            int horizontalOffset,
+            VerticalAlignment verticalAlignment,
+            HorizontalAlignment horizontalAlignment,
+            AnchorNinth anchorNinth) {
+        this.verticalAlignment = verticalAlignment;
```
magicus, line 23: This alignment is *relative* to the "ninth" sector? Otherwise it seems redundant.

kristofbolyai: What do you mean? Alignments are used to position the overlay in it's ninth with the specified offsets too.

common/src/main/java/com/wynntils/core/features/overlays/annotations/OverlayInfo.java
```diff
@@ -20,6 +20,8 @@
 
     RenderState renderAt() default RenderState.Post;
 
+    boolean cancelRender() default false;
```
magicus: You set this to true to override the vanilla rendering? It's not super clear. 

What about having the render state being a tri-state instead, Pre, Post and Replace? The latter would be called at Pre but also cancel the event. And you could skip the check in the init that we don't try a cancelRender for Post.

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/FixedOverlaySize.java
```diff
@@ -0,0 +1,23 @@
+/*
```
magicus: I'm really allergic to upper case letters in package names. 🤧  I don't think I've ever seen that before. In this case I think it's easy to resolve by just renaming the package `sizes`, since it is below `overlays` anyway.

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/FixedOverlaySize.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Override
+    public float getWidth() {
+        return (float) (this.width / McUtils.window().getGuiScale());
+    }
+
+    @Override
+    public float getHeight() {
+        return (float) (this.height / McUtils.window().getGuiScale());
```
magicus: I assume the gui scale thing works strangely, because on the surface of things it sounds like this would be the "gui scaled" overlay size, and the class below the "fixed" one...

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/RestrictedRangeOverlaySize.java
```diff
@@ -0,0 +1,34 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.features.overlays.overlaySizes;
+
+public class RestrictedRangeOverlaySize extends OverlaySize {
+    private final float maxWidth;
+    private final float maxHeight;
+
+    public RestrictedRangeOverlaySize(float width, float height, float maxWidth, float maxHeight) {
+        super(width, height);
```
magicus: Maybe check the max restriction in the constructor as well? :-D

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/FixedOverlaySize.java
kristofbolyai: Yeah, it seems strange, but this works as intended. I got it mixed up first try too.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -4,24 +4,80 @@
...
+
+    public VerticalAlignment getVerticalAlignment() {
+        return verticalAlignment;
+    }
+
+    public int getVerticalOffset() {
+        return verticalOffset;
+    }
+
+    public enum AnchorNinth {
```
magicus: What do you think about calling this `Sector` instead? Or perhaps `Section`?

(If you accept this, other references to "ninth" should be updated as well)

common/src/main/java/com/wynntils/mc/event/DisplayResizeEvent.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import net.minecraftforge.eventbus.api.Event;
+
+public abstract class DisplayResizeEvent extends Event {
```
magicus: I'm not really sure we need a Pre/Post version of this. I think in most cases we can get along with just a single event for all kinds of things that happens in vanilla. I think we should not create Pre/Post versions of everything "just in case", but only put in the separation if it turns out that we cannot do with a single event. For resizing this sounds unlikely to ever happen.

magicus:
Overall it looks great! 👍 Just had a few minor comments.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
magicus: So what I mean is that the ninths/sectors themselves are aligned relative to the entire screen. So the "first" is TopLeft, which mean it's aligned towards top and left on the screen. Then the "Alignment" here is alignment is alignment within the sector, so you can have something relative to the "top left" of the TopLeft ninth. One one hand, it's clear as day, and you could not really do it differently, but on the other hand, the word "alignment" becomes a bit fuzzy, since you need to be super clear about *what* you are aligning against.

I was mostly thinking out loud if there could be a way to make this completely unambiguous. But maybe the current code is the best we can do.

common/src/main/java/com/wynntils/core/features/overlays/overlaySizes/FixedOverlaySize.java
magicus: Maybe add a comment acknowledging this f*ckedupedness of vanilla and asserting that this does indeed works as the name says..?

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
kristofbolyai: I think we need to have a way to align to overlays in their ninths any way, and we even will use negative alignment offsets. Setting these values will become the job of the overlay management GUI, which should save overlay position in the most optimal way.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -54,13 +61,55 @@ private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState re
         for (Overlay overlay : enabledOverlays) {
             OverlayInfo annotation = overlayInfoMap.get(overlay);
 
-            if (renderState != annotation.renderAt() || event.getType() != annotation.renderType()) continue;
+            if ((annotation.renderAt() == OverlayInfo.RenderState.Replace && renderState != OverlayInfo.RenderState.Pre)
```
magicus: I suggest splitting this up in separate if statements, like:
if not the right render time - continue

if renderAt=replace:
  if not pre - continue
  setCanceled
else
 if not renderAt = state continue
fi

Or something like that.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -54,13 +61,55 @@ private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState re
...
+
+    @SubscribeEvent
+    public static void onResizeEvent(DisplayResizeEvent event) {
+        calculateNinths();
+    }
+
+    // Calculate the ninths when loading is finished (this acts as a "game loaded" event)
+    @SubscribeEvent
+    public static void gameInitEvent(TitleScreenInitEvent event) {
+        calculateNinths();
```
magicus: I'm thinking this should be calculateSections as well.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -20,6 +25,8 @@ public class OverlayManager {
 
     private static final Set<Overlay> enabledOverlays = new HashSet<>();
 
+    private static final List<Pair<Coordinate, Coordinate>> ninths = new ArrayList<>(9);
```
magicus: This should be `sections`.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -54,13 +61,55 @@ private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState re
...
+
+        ninths.clear();
+        for (int h = 0; h < 3; h++) {
+            for (int w = 0; w < 3; w++) {
+                ninths.add(new Pair<>(new Coordinate(w * wT, h * hT), new Coordinate((w + 1) * wT, (h + 1) * hT)));
+            }
+        }
+    }
+
+    public static Pair<Coordinate, Coordinate> getNinth(OverlayPosition.AnchorSection ninth) {
```
magicus: `getSection` (and `getSections` below)

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -54,13 +61,57 @@ private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState re
         for (Overlay overlay : enabledOverlays) {
             OverlayInfo annotation = overlayInfoMap.get(overlay);
 
-            if (renderState != annotation.renderAt() || event.getType() != annotation.renderType()) continue;
+            if (annotation.renderAt() == OverlayInfo.RenderState.Replace) {
```
magicus, line 68: You lost the `event.getType() != annotation.renderType()` in the process :(

kristofbolyai: Ooops. Fixed.

magicus (approved):
Ship it!

common/src/main/java/com/wynntils/core/features/overlays/BasicOverlay.java
```diff
@@ -6,22 +6,31 @@
 
 import com.mojang.blaze3d.platform.Window;
 import com.mojang.blaze3d.vertex.PoseStack;
-import com.wynntils.utils.objects.PentaConsumer;
+import com.wynntils.core.features.overlays.sizes.OverlaySize;
+import com.wynntils.utils.objects.QuadConsumer;
 
 public class BasicOverlay extends Overlay {
-    private final PentaConsumer<Overlay, OverlayPosition, PoseStack, Float, Window> renderConsumer;
+    private final QuadConsumer<Overlay, PoseStack, Float, Window> renderConsumer;
```
Incompleteusern, line 13: It might be good to just have this a functional interface, we probably won't use QuadConsumers in many places.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -6,35 +6,54 @@
...
-    public float getWidth() {
-        return width;
+    // Return the X where the overlay should be rendered
+    public int getRenderX() {
+        final Pair<Coordinate, Coordinate> section = OverlayManager.getSection(this.position.getAnchorSection());
+        return switch (this.position.getHorizontalAlignment()) {
+            case Left -> section.a.x() + this.position.getHorizontalOffset();
+            case Center -> (int) (section.a.x() + section.b.x() - this.getWidth()) / 2
+                    + this.position.getHorizontalOffset();
+            case Right -> (int) (section.b.x() - this.position.getHorizontalOffset() - this.getWidth());
```
Incompleteusern, line 45: The horizontal offset is added in the Left and Center cases, but subtracted in the Right case. It might be better to just have it always added. Same comment applies for vertical offset.

common/src/main/java/com/wynntils/features/user/DummyOverlayFeature.java
```diff
@@ -6,49 +6,104 @@
...
+    @SubscribeEvent
+    public void renderNinths(RenderEvent.Post event) {
+        Window window = McUtils.mc().getWindow();
+        int width = window.getGuiScaledWidth();
+        int height = window.getGuiScaledHeight();
+
+        int wT = width / 3;
+        int hT = height / 3;
+
+        for (Pair<Coordinate, Coordinate> ninth : OverlayManager.getSections()) {
```
Incompleteusern, line 104: It might be better to merge the ninth vs section terminology.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -20,6 +25,8 @@ public class OverlayManager {
 
     private static final Set<Overlay> enabledOverlays = new HashSet<>();
 
+    private static final List<Pair<Coordinate, Coordinate>> sections = new ArrayList<>(9);
```
Incompleteusern, line 28: `Pair<Coordinate, Coordinate>` could be made more descriptive.

Maybe something like

```java
public record BoxCoords(int x1, int x2, int y1, int y2) {}
```

Incompleteusern (commented):
I've finished this review after it was merged, so it might be better to deal with this in a separate pr, though only the second comment really matters.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: Why? If, let's say, my vertical alignment is aligned to bottom, any positive offset should move the overlay further from the bottom. Same with horizontal right.

Incompleteusern: It's inconsistent with Left and Center, and would make it harder to use imo.

kristofbolyai: Yeah, sure, I don't mind either way. I will open a PR in a bit.

PR #131, opened 2022-07-02
WynntilsMod/Reference: Add a way to check if mod is in development environment

kristofbolyai:


common/src/main/java/com/wynntils/core/Reference.java
```diff
@@ -12,4 +12,6 @@ public class Reference {
     public static final Logger LOGGER = LoggerFactory.getLogger(WynntilsMod.MOD_ID);
     public static String VERSION = "";
     public static int BUILD_NUMBER = -1;
+
+    public static boolean developmentEnvironment;
```
magicus, line 16: Hm. I mean, the fix seems correct, and I understand why you'd want to put this in Reference, but public static fields is a bit so-so. :-/ 

A public getter would be more kosher methinks. Maybe make the variable default ("package private") access, and set it from WynntilsMod anyway, but use getters when reading it?

kristofbolyai: For every variable in reference?

kristofbolyai: Also, what do you think of moving `MOD_STORAGE_ROOT` from WynntilsMod to Reference?

magicus: Hm. Do you need to ask those difficult questions? ;-) 

To clafify: I have no issue with public constants. So public static final is fine. 

I see now that VERSION and BUILD_NUMBER is not actually final, despite their names. Maybe we can continue that charade and pretend that developmentEnvironment is effectively final, and just rename it to DEVELOPMENT_ENVIRONMENT? (Or perhaps a shorter `DEV_ENV`?)

Hmmmm... Or maybe not. I don't really have a good solution here. Maybe the distinction between Reference and WynntilsMod is a bit synthetic, and we should just make sure we have a singleton WynntilsMod that contains all this kind of things? It's probably more elegant, design-wise, but more clumsy when accessing...

magicus: Yeah, maybe that'd be best. Skip the Refence class, move that stuff into WynntilsMod, make it a singleton, and make convenience static methods like `getLogger()` which does like `return getInstance().logger`...

If you are okay with that, I can fix it. Let's integrate this fix for now first, and then I can do some refactoring.

magicus (approved):

kristofbolyai:
I will merge this then, and you go ahead and fix this whenever you have time.

PR #135, opened 2022-07-03
Scoreboard Manager: QuestManager and ObjectiveManager

kristofbolyai:
This PR should handle:

- [x] Objective management and parsing
- [x] Quest parsing from scoreboard

The goal for both the quest and objective manager classes is to have a framework in place to add them as overlays.


common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/ObjectiveManager.java
```diff
@@ -0,0 +1,76 @@
...
+import com.wynntils.core.WynntilsMod;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ObjectiveManager {
+    // §b is guild objective, §a is normal objective and §c is daily objective
+    public static final Pattern OBJECTIVE_PATTERN = Pattern.compile("^§([b|ac])[- ]\\s§7(.*): *§f(\\d+)§7/(\\d+)$");
```
magicus: You don't need `|` as or in a `[]` group. (I think it is actually interpreted as the literal `|` character)

common/src/main/java/com/wynntils/mc/event/ScoreboardSetScoreEvent.java
```diff
@@ -0,0 +1,43 @@
...
+package com.wynntils.mc.event;
+
+import net.minecraft.server.ServerScoreboard;
+import net.minecraftforge.eventbus.api.Event;
+import org.jetbrains.annotations.Nullable;
+
+public class ScoreboardSetScoreEvent extends Event {
+    private final String owner;
+
+    @Nullable
```
magicus: I know IntelliJ likes to sprinkle those `@Nullable` around, and I just hate them. If we want to have some kind of static null safety checks, we need to use these rigorously. And the resulting cluttering of code is hardly worth it imho. So please kill them, kill them with 🔥  ;-)

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/Objective.java
```diff
@@ -0,0 +1,77 @@
...
+        // "  lvl 10 players: 0/24"
+        if (matcher.find()) {
+            goal = matcher.group(1);
+            try {
+                score = Integer.parseInt(matcher.group(2));
+                maxScore = Integer.parseInt(matcher.group(3));
+            } catch (NumberFormatException e) {
+                // Ignored, goal is already null
+            }
+        }
```
magicus: I realize you are mostly copying code that I wrote myself but I don't understand how this can work. 😆 How can we ever parse a multi-line objective with this code?

kristofbolyai: What is a multi line objective? There are objectives that are split into multiple lines?

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
```diff
@@ -0,0 +1,92 @@
...
+public class QuestManager {
+
+    private static final String TRACKED_QUEST_STRING = "§6§lTracked Quest:";
+
+    private static QuestInfo currentQuest = null;
+
+    public static void tryParseQuest() {
+        Scoreboard scoreboard = McUtils.player().getScoreboard();
+
+        List<Objective> objectives = scoreboard.getObjectives().stream().toList();
```
magicus: I could not wrap my head around this, until I realized that there is an `Objective` class in vanilla as well, so this is not "our" `Objective`. Maybe we should rename our class to like `WynnObjective` or `WcObjective` or such, to avoid this clash. It's one thing if completely unrelated code happens to have name clashes, but now we're basically dealing with the same stuff.

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
```diff
@@ -0,0 +1,92 @@
...
+
+        // Invalidate wrong parsing due to scoreboard refresh
+        if (!questLines.get(0).startsWith("§e")) {
+            return;
+        }
+
+        // An objective will appear among the quest lines if scoreboard is refreshing
+        for (int i = 1; i < questLines.size(); i++) {
+            if (ObjectiveManager.OBJECTIVE_PATTERN.matcher(questLines.get(i)).matches()) {
+                questLines = questLines.subList(0, i);
```
magicus: If that happens, is not the entire quest line segment broken? 
Also, modifying an array while iterating over it is scary, at best. Maybe use an iterator that allows for delete(), if you think that selectively deleting a objective lines is fine.

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
```diff
@@ -0,0 +1,92 @@
...
+                break;
+            }
+        }
+
+        questLines = questLines.stream()
+                .map(ChatFormatting::stripFormatting)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
+        currentQuest = new QuestInfo(
```
magicus: We are doing this entire method on each `ScoreboardSetScoreEvent`. It seems quite a lot of processing. Do you think there is a way to cache/shortcut this to figure out if the tracked quest has actually changed?

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
kristofbolyai: No, from my testing it always happens to be at the end of the quest description. Also, we break after and we only use the array size so it is safe, but if you want it changed anyways I can.

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/Objective.java
magicus: Yes, that was what I meant. From the comment:
```
        // Could also be multi line like:
        // "- Trade 24E with"
        // "  lvl 10 players: 0/24"
```

magicus: But honestly, I think I wrote that comment, and I can't really see how my code in legacy could deal with this properly. :-& Or maybe it did originally, but it has been rewritten since so that this handling has broken.

kristofbolyai: I think maybe Wynn changed it so they are single line only

magicus: Maybe we should do something like what you are doing in the quest manager, and build up the entire scoreboard, as seen on the screen by the user, before starting to parse things out of it (like a List<String>, basically). And just listen to the events as signals that we need to update this in-memory parsed scoreboard, and if it changed, reparse the parts that changed. (party health, tracked quest, etc)

kristofbolyai: I would rather confirm that multi line objectives still exist before making it unnecessary complicated, as for party tracking, we don't need to get the entire scoreboard

magicus:
I'm thinking like this. Look at this scoreboard screenshot:
scoreboard

We could create a representation of it in the ScoreboardManager, which basically contains everything between line 15 to line 2, with the formatting intact, in correct order, but ignoring the score value (which is just mis-used by Wynncraft to get the lines in order). We could have the different "scoreboard consumers/parser" grab different parts, so TrackedQuest would grab 15-10, Objectives 8-2. When the scoreboard changes in any of these "grabbed" parts, we only need to notify the owner. 

The big gain here will be to handle the party health update. As you say, it will update quite frequently, and we'd really like to avoid re-parsing the objectives and tracked quest for that. 

Also, that is a model that seem to fit in with how Wynncraft themselves use the scoreboard. So if they ever add a new section ("Campain of the week: - Buy two crates, pay for three!") it will likely fit right in and we can sort it out to it's own "consumer".

magicus:
And, of course, that will also help with multi-line objectives.

magicus:
Also, mostly for the record, I took a screenshot of how the scoreboard looks when you have party information as well:
scoreboard

It was formatted more or less as I remembered it, so it too matches the idea of scoreboard "parts".

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/Objective.java
kristofbolyai: Also rebuilding doesn't help with parsing or performance either

magicus: I'm not sure how we can confirm that. Maybe @HeyZeer0 knows, or can find out. I only occasionally got those, the only one I can remember was the one urging you to go throw emeralds on noobs.

kristofbolyai: Can you try explaining the benefits compared to now?

kristofbolyai:
> I'm thinking like this. Look at this scoreboard screenshot:
> 
> scoreboard
> 
> 
> 
> We could create a representation of it in the ScoreboardManager, which basically contains everything between line 15 to line 2, with the formatting intact, in correct order, but ignoring the score value (which is just mis-used by Wynncraft to get the lines in order). We could have the different "scoreboard consumers/parser" grab different parts, so TrackedQuest would grab 15-10, Objectives 8-2. When the scoreboard changes in any of these "grabbed" parts, we only need to notify the owner. 
> 
> 
> 
> The big gain here will be to handle the party health update. As you say, it will update quite frequently, and we'd really like to avoid re-parsing the objectives and tracked quest for that. 
> 
> 
> 
> Also, that is a model that seem to fit in with how Wynncraft themselves use the scoreboard. So if they ever add a new section ("Campain of the week: - Buy two crates, pay for three!") it will likely fit right in and we can sort it out to it's own "consumer".
> 
> 

Okay, just to clear this up. I don't mind doing this, but I am not 100% sold on the benefits. We save a little time on having a constructed scoreboard list, but otherwise every update and matching logic will stay the same mostly. Quest tracking is the only manager to need a fully constructed scoreboard to work, as every other can work with only regex/other types of matching on the event packet.

kristofbolyai:
I figured out a solution that should make this cleaner and way more performant. I will start working on it now.

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/Objective.java
P0keDev: the answer is that it never really did handle them properly - it only matched the second line, so the objective overlay would just display "lvl 10 players" as the objective. can't comment on whether the objective is still in the game or not

kristofbolyai: It was changed to trade with newbies so it is 1 line. If it is not, I will fix it later.

kristofbolyai:
@magicus This version should be much better than the previous solution. It is more performant, cleaner, and is much easier to adapt.

magicus:
To be fully clear: I am not 100% convinced of the gains either. :-) However, I have this distinct feeling that it will be easier to "do stuff" with the scoreboard if we start by getting it in a state where it can be manipulated in a way that corresponds to how Wynncraft (mis-)uses it, and which makes more sense. 

As I remember, the scoreboard API from mojang was quite targeted at doing, well, scoreboards, where players names were not really supposed to change, but their score could. Wynncraft is misusing this to keep scores constant, but changing (fake) players names, to be able to create an "always visible" box of text. 

By reconstructing this "box of text" as a concept in our code, we give ourselves a framework that matches how Wynncraft is presented to the user. I am convinced that this will, if done right, lay a good foundation to make current and future needs of analyzing/parsing the scoreboard simpler and more intuitive.

kristofbolyai:
> To be fully clear: I am not 100% convinced of the gains either. :-) However, I have this distinct feeling that it will be easier to "do stuff" with the scoreboard if we start by getting it in a state where it can be manipulated in a way that corresponds to how Wynncraft (mis-)uses it, and which makes more sense.
> 
> As I remember, the scoreboard API from mojang was quite targeted at doing, well, scoreboards, where players names were not really supposed to change, but their score could. Wynncraft is misusing this to keep scores constant, but changing (fake) players names, to be able to create an "always visible" box of text.
> 
> By reconstructing this "box of text" as a concept in our code, we give ourselves a framework that matches how Wynncraft is presented to the user. I am convinced that this will, if done right, lay a good foundation to make current and future needs of analyzing/parsing the scoreboard simpler and more intuitive.

After rewriting the code, I absolutely agree with you. At first I did not see the full scope of gains, but now, it is clear. 😅

magicus:
Oh, nice, a new version as I was writing my lithania. :-) 

I scanned briefly through the new version, and it seemed to be more in the direction I prefer. But my brain has run out of fuel for today; I've been doing too many code reviews both at work and on hobby projects. I'll have to come back tomorrow. I hope you can wait. :)

kristofbolyai:
> Oh, nice, a new version as I was writing my lithania. :-)
> 
> I scanned briefly through the new version, and it seemed to be more in the direction I prefer. But my brain has run out of fuel for today; I've been doing too many code reviews both at work and on hobby projects. I'll have to come back tomorrow. I hope you can wait. :)

I will fix the bugs in the meantime 😅

common/src/main/java/com/wynntils/mc/event/WynntilsScoreboardUpdateEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import java.util.Map;
+import java.util.Set;
+import net.minecraft.server.ServerScoreboard;
+import net.minecraftforge.eventbus.api.Event;
+
+public class WynntilsScoreboardUpdateEvent extends Event {
```
magicus: This is not really a Minecraft event and does not belong under `com.wynntils.mc.event`. I think we have a `com.wynntils.wc.event`, no?

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final Pattern PARTY_PATTERN = Pattern.compile("§e§lParty:§6\\s\\[Lv. (\\d+)]");
+    private static final Pattern PARTY_PLAYER_HEALTH_PATTERN =
+            Pattern.compile("-\\s\\[\\|\\|(\\d+)\\|\\|]\\s(.+)\\s\\[(\\d+)]");
+
+    // TimeUnit.Seconds
+    private static final int CHANGE_PROCESS_RATE = 1;
+
+    private static final List<ScoreboardLine> reconstructedScoreboard = new ArrayList<>();
+
+    private static final LinkedList<ScoreboardLineChange> queuedChanges = new LinkedList<>();
```
magicus, line 49:
```suggestion
    private static final List<ScoreboardLineChange> queuedChanges = new LinkedList<>();
```

It's nicer to hide implementation type in the declaration. :)

magicus: Oh, I see, you use it as a queue, so you need the type.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -0,0 +1,178 @@
...
+            return Optional.of(WynntilsScoreboardUpdateEvent.ChangeType.Objective);
+        }
+
+        if (withoutFormat.matches("À+")
+                || OBJECTIVE_HEADER_PATTERN.matcher(withoutFormat).matches()
+                || GUILD_OBJECTIVE_HEADER_PATTERN.matcher(withoutFormat).matches()) {
+            return Optional.empty();
+        }
+
+        // For unidentified text, assume it is a quest description
```
magicus: Let's see if we spot the multi-line objective again, then. Otherwise, this will break anyway, right? (All parsing that is based on a per-line approach will fail in the face of multi-line objectives afaict)

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -0,0 +1,178 @@
...
+                Comparator.comparingInt(ScoreboardLine::index).reversed());
+
+        WynntilsMod.getEventBus().post(new WynntilsScoreboardUpdateEvent(changeTypes));
+    };
+
+    private static Optional<WynntilsScoreboardUpdateEvent.ChangeType> getChangeType(ScoreboardLineChange change) {
+        String line = change.lineText();
+        String withoutFormat = ChatFormatting.stripFormatting(line);
+
+        if (withoutFormat == null) {
```
magicus: What about empty lines?

kristofbolyai: Empty lines have À.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -0,0 +1,178 @@
...
+    private static void getTrackedObjectivePlayer() {
+        Scoreboard scoreboard = McUtils.player().getScoreboard();
+
+        List<Objective> objectives = scoreboard.getObjectives().stream().toList();
+
+        if (objectives.isEmpty()) {
+            return;
+        }
+
+        trackedObjectivePlayer = objectives.get(0);
```
magicus, line 242: Are you feeling confident that this is OK, just grabbing the first? I remember I wrote some code in legacy trying to find a player named "sb_<playername>" to track the scoreboard changes. Iirc, there also was a "fb_<playername>" but it did not get all scoreboard updates. Or maybe not. It's all a bit unclear this much afterwards.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
magicus: I wrote a comment about that a few lines below when I found the À match, but Github seems to have flushed it down the drain. :-(

Anyway, in short, I was saying that in Legacy we had a function that "cleaned" strings from those À characters (and also like stripped leading/trailing spaces, I think), and we basically always used this. I think that was good, since otherwise these characters tended to creep up in all sorts of places, throwing regexps off.

So unless we *need* the À characters as markers, I think it is better if we get rid of them as soon as possible, and instead match for the empty string to find the empty lines.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
kristofbolyai: IMO, let's assume it is correct and we will fix it if it is not. I don't want to over engineer it if it is not needed.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
kristofbolyai: They are useful here, the count of them shows what section we are in.

kristofbolyai: We don't take advantage of that right now (we don't need to) but they can be useful this case

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
```diff
@@ -0,0 +1,73 @@
...
+                event.getChangeMap().get(WynntilsScoreboardUpdateEvent.ChangeType.Quest);
+
+        for (WynntilsScoreboardUpdateEvent.Change change : changes) {
+            if (change.method() == ServerScoreboard.Method.REMOVE) {
+                if (TRACKED_QUEST_STRING.equals(change.line())) {
+                    questTrackingStopped();
+                }
+            } else { // Method is CHANGE
+                // Use copy constructor to prevent threading issues
+                List<ScoreboardLine> reconstructedScoreboard =
```
magicus: I was about to add a comment on this in ScoreboardManager, but started by searching for use cases, and found this. :-) 

The problem is that you call `reconstructedScoreboard.sort()`, which will temporarily leave the scoreboard in a random state. If you sort it to a new List, and then just reassign the field, you will not have threading issues, and would not need to do a defensive copy here instead. (You'd still need to use `getReconstructedScoreboard()` and store the value in a local variable, so it does not change under your feet.)

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/ObjectiveManager.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private static WynnObjective guildWynnObjective = null;
+
+    private static final List<WynnObjective> WYNN_OBJECTIVES = new ArrayList<>();
+
+    @SubscribeEvent
+    public static void onScoreboardUpdate(WynntilsScoreboardUpdateEvent event) {
+        if (!event.getChangeMap().containsKey(WynntilsScoreboardUpdateEvent.ChangeType.Objective)) return;
+
+        Set<WynntilsScoreboardUpdateEvent.Change> changes =
```
magicus: Just thinking out loud here.. Do you think it would be better to have specialized Events instead, so you'd send like a `WynntilsScoreboardUpdate.ObjectiveEvent` instead, and don't have to filter out, and don't have to just get the relevant part of the map the first thing you do? It feels a bit like this is multiple different events baked into one.

kristofbolyai: I guess that works too if you want it to be like that.

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
kristofbolyai: What if get method sorted the list and assigned it? Sort is not expensive if it is already sorted

kristofbolyai: Oh actually your method is better, I get it now

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/ObjectiveManager.java
```diff
@@ -0,0 +1,95 @@
...
+
+        if (WYNN_OBJECTIVES.size() > 3) {
+            WynntilsMod.error("ObjectiveManager: Stored more than 3 objectives. Reset objective list.");
+            WYNN_OBJECTIVES.clear();
+            WYNN_OBJECTIVES.add(parsed);
+        }
+    }
+
+    private static void updateGuildObjective(WynnObjective parsed) {
+        guildWynnObjective = parsed;
```
magicus, line 67: It's a bit odd that "Cut down trees 1/3" and "Cut down trees 2/3" are *different* objectives. It feels like they should be the *same* objective, but have the progress updated. Maybe it's okay, since the only real practical purpose of the objectives is the upcoming overlay, which will likely just query the current objective at each render, and don't care. But from a modeling perspective, it has an uneasy ring to it.

But it's also easy to fix afterwards, if we want to, so I'm not saying you need to address this now.

magicus: My thinking is like this: if we say it's the same objective, we have the opportunity to put in a hook whenever the objective progress is changed. This could make us e.g. play a sound ("bling!") whenever an objective progress is increased, something which is perhaps not a show-stopper, but is definitely something I'd like to see. So it's more about creating an architecture that is beneficial and welcoming for future features like that.

kristofbolyai: Sure, it is 1 line change, I have the framework for it already.

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+                if (questLines.isEmpty()) {
+                    WynntilsMod.error("QuestManager: questLines was empty.");
+                }
+
+                currentQuest = new QuestInfo(
+                        questLines.get(0),
+                        StringUtils.joinWith(" ", questLines.stream().skip(1).collect(Collectors.toList())));
+
+                // Only update currentQuest once per batch.
```
magicus: Can you get `changes` that is like (`CHANGE`: ...., `REMOVE`: ...)? If so, could you not end up in a situation where you setup a new currentQuest here, and then replace it with `null` right after?

Or, hm, you are just looking if the header is removed. And reasonably the header could not be removed and the actual quest lines are changed. I guess...

TL;DR: This seems like it could be working, but it's more of the kind "no obvious problems" instead of "obviously no problems".

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
magicus: Okay, if you say. Personally I think I'd rather used matching headers to keep track of sections, but there's more than one way to do it, as Perl says.

common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/ObjectiveManager.java
magicus: Yes, I think it is better. As a general rule, events work best when they are "small" and "targeted".

magicus:
This version looks much better! There are still something nagging me, that I can't really put my finger on. But it's more about specific implementation details, or minor design choices, that can always be fixed afterwards.

Edit: I think I know what it is now. Unfortunately, the margin is to small... Er, I mean. I have to leave now, and don't have time to write it down. I'll get back to you later today.

common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
kristofbolyai: Order should be correct as it is the order of the packets. And the header is only removed if the quest tracking stops. I know it feels a bit strange, but this should work fine.

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -85,6 +86,8 @@ public static void init(String modVersion, boolean isDevelopmentEnvironment) {
         ClientCommandManager.init();
         KeyManager.init();
         CompassManager.init();
+        ScoreboardManager.init();
+
```
magicus: Even though the list of init:s is long, this break seems pretty arbitrary. If you find some reasonable way to group the manager, go ahead, but otherwise this is just one in a long list of bootstrapping modules.

kristofbolyai: Sure, I can just remove the newline. If it bothers us too much, we can find an organization for them later

magicus:
Ok, I'm back. So, maybe my opinion here is just that I would have implemented this differently myself, and there is no objective reason to chose my version over your. I cannot really say. But I'll try to explain my thinking, and then you can reject it or accept it, whatever you see fit.

In my mind, the role of the ScoreboardManager is to do both more and less work. It should do less "Wynn-centric" work, but more "preparatory" work. A central theme in how Wynncraft uses the scoreboard is that it is divided into segments (or sections, or groups, or whatever; I'll go with segments for now just to have a unique word), where each segment has a title, and some contents. We can with reasonable certainty say that a new segment starts with an empty line(or an `A), the header line which ends with a colon, and then the content is all non-empty lines. (The very first line is the URL which can be ignored, after which follows 0+ segments).

At this point, we're basically not making any assumption on the content. (If we *really* want to be content-agnostic, we can even just split by empty lines and ignore separating header and content).

I think the ScoreboardManager should deal with this "segments" as it basic datatype. A segment is basically a List<String> of formatted strings, or if we want to be fancy, a `record Segment(String header, List<String> content)`. Any reference to changes should be relative to the segment. So if the first line in the content changes, it should not be seen as "line 13 in the scoreboard", but "line 1 in this segment".

(Continuing...)

magicus:
Furthermore, ScorboardManager should have "handlers" for different segments. A handler basically registers a regexp for the header, and gets callbacks whenever the segment is created, disappears, or changes content. 

So ScoreboardManager divides to scoreboard into segments, and for each segment checks if there is a registered handlers that can take care of it, and notifies that handler whenever that segment changes (or is created/destroyed). To do this, it has to do about the same thing that it currently does, going over the incoming changes and matching them with the "cached" value of the scoreboard. It just needs to separate these changes into segments.

This would push the ScoreboardManager more towards the `core` part of the code, rather than `wc`, making it more a foundational API. The respective scoreboard segment handlers, otoh, are clearly "wynncraft" code.

magicus:
(A slight complication is that the guild objectives has "content" (a counter) in the heading, but I don't think it changes anything per se, we just need to remember that the header can changes as well and that counts as a content change.)

I also believe that with this solution, we could relax the strictness of the change notifications as well. I think it will be enough to tell a single segment to refresh it's data, if any line of the contents has changed, rather than to send it a specific set of changed lines. E.g. if we have a party:
```
|1234| member1
|4321| member2
|2222| member3
```

and member1 and 3 gets health changed:
```
|999| member1
|4321| member2
|1111| member3
```
in the current model, we'd get a change report saying that line 1 and 3 in the segment is different, so we can skip re-parsing line 2. But this comes at a cost, which is the increased complexity in how we communicate this change information as a set of deltas for specific lines, with content and line number. I believe that in the grand scheme of things, the cost of reparsing the entire segment if any part of it has changed (esp. if we do this max once per second!) is completely trivial. What we *don't* want to do is to have to reparse the Tracked Quest if the party health changes. But by separating this into segments, I think we are fine with asking each segment to reparse if any line of it changed.

Accepting this means that we just need to call like `onSegmentChange(Segment newValue)` for each handler, and just send them the new segment (header + content). I think this will make for a more clear interface, that is easier to extend should new segments arrive, and is robust and performant enough.

magicus:
And, finally, this model will easily handle multi-line objectives. :-) (And as an odd bonus, if Wynncraft were to suddenly swap the order of segments in-game, without changing the contents, we wouldn't see this as a change in content, since it isn't.)

Now I'm done. *phew*

kristofbolyai:
I will consider this and think through implementation details. If it is not a huge pain, I may do it. 

The only reason I would not do this is because I don't think we need ScoreboardManager to be that big of a handler and make it part of the core code, as it only handles 3-4 overlays at most. I can see the better code structure as a benefit, but I am not sure if we should have a 1000 line scoreboard manager just so we can re-render that with our own overlays.

magicus:
I don't think it needs to be much larger. The only "real" changes are in the central loop of ScoreboardManager, where you need to match the incoming changes to the respective segment. But you still need to kind of do that anyway. It's just that it's done with more specific knowledge about the format of each individual line, where unknown lines are supposed to be quest lines.

magicus:
And I'm not saying ScoreboardManager needs to move to `core` -- it is just "approaching" core. Even if we strip the knowledge of what kind of segments there are, it is still a wynncrafty construction with segments in the scoreboard anyway, so I'm fine with keeping it in `wc`. But it will raise the abstraction of ScoreboardManager, which I think is a good thing, regardless of which package it belongs to.

kristofbolyai:
Okay, I will try to make this system change, however it might be a compromise of the two. I will get back to you when it is done.

magicus:
I'm feeling a bit old-school (or maybe just old) when I have trouble reading all these chained streaming operations... But I think this looks good. You were right in that it required more added complexity than I thought. I think it might be possible to shortcut some operations by using some cleverness, but it's probably not worth it, and it might end up making the code harder to read rather than the other way around. I am happy to see that the tracked quest and objective handlers got simplified. That means that future party etc handlers will likely be simpler to write, too.

magicus (approved):
You've been at this far to long. :-) Thanks for indulging me. Time to ship it!

kristofbolyai:
Now comes the pain. Rendering.

magicus:
🎉 🎆

magicus:
Yeah. Rendering.... *sigh* I'll... er... I'll go back to finishing the Chat Manager, if you don't mind? 😄

PR #136, opened 2022-07-04
Item Highlight Improvements & ItemProperty Rework

### LINKED ISSUE #128: Unidentified Revealed Mythics Glow Missing
image

Revealed mythics don't glow on the market or in booths. The issue is probably more general - these are just the places where I found it.

P0keDev:
This PR changes the `ItemProperty` system to allow for `PropertyType` interfaces that can be implemented by different item properties. This allows for "classes" of properties: `ItemTierProperty` and `CosmeticTierProperty` both represent something that corresponds to an inventory highlight color, so they implement `HighlightProperty`. This design lets us avoid making entire custom itemstacks for items that only require very basic handling that isn't unique to them, instead delegating this to properties that can be handled universally in feature code. Thus, custom itemstacks are reserved for items that require a lot of unique processing/modification, such as gear items.

An example of a potential property type might be a text overlay property, which would be applicable to tiers of certain items, consumable charges, shorthand identifiers (tp scrolls), etc. An item might have multiple such properties, which is why I also added support for collecting multiple properties of a given type.

The changes to the highlight feature mostly involve rewriting it to use the new `HighlightProperty` type. Thanks to the more universal `ItemTierProperty`, it will now work on every applicable item, which fixes #128.


common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -123,7 +122,7 @@ public static void init() {
         // debug
         registerFeature(new ConnectionProgressFeature());
         registerFeature(new PacketDebuggerFeature());
-        registerFeature(new DummyOverlayFeature());
+        // registerFeature(new DummyOverlayFeature());
```
magicus: Is this due to https://github.com/Wynntils/Artemis/issues/138?

common/src/main/java/com/wynntils/features/user/ItemHighlightFeature.java
```diff
@@ -87,10 +88,16 @@ public void onRenderSlot(SlotRenderEvent.Pre e) {
         if (!inventoryHighlightEnabled) return;
 
         ItemStack item = e.getSlot().getItem();
-        if (!(item instanceof HighlightedItem highlightedItem)) return;
```
magicus, line 90: Given the similarities between this and the hotbar version, maybe you can extract this into a `getHighlightColor` method, which just returns `CustomColor.NONE` if it should not be highlighted (for whatever reason)?

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
```diff
@@ -35,19 +36,37 @@ public List<Component> getOriginalTooltip() {
...
 
     public void addProperty(ItemProperty property) {
-        if (getProperty(property.getClass()) != null) return; // don't allow duplicate properties of same type
+        if (getProperty(property.getClass()) != null) return; // don't allow duplicate properties
         this.properties.add(property);
     }
 
-    public <T extends ItemProperty> T getProperty(Class<T> propertyType) {
+    /**
+     * Returns the specified property or the first property of the specified type, if it exists.
```
magicus, line 44: It's not so much an "or" here, it's really just "return the first property of the specified type", is it not?

(Same goes for hasProperty)

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
```diff
@@ -35,19 +36,37 @@ public List<Component> getOriginalTooltip() {
     }
 
     public void addProperty(ItemProperty property) {
-        if (getProperty(property.getClass()) != null) return; // don't allow duplicate properties of same type
+        if (getProperty(property.getClass()) != null) return; // don't allow duplicate properties
```
magicus: I'm not sure I get this. How can we ever set multiple properties of the same class? Or what does `getProperties` do? Is the idea that you can insert multiple subtypes with addProperty and then send a request with the supertype to getProperties?

(And, you now have a `hasProperty`, why not use it? ;-))

magicus:
At times I felt a bit lost in the abstractions, but I think this looks good. It has a cleaner feel to it than the previous iteration.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
P0keDev: well no, since i was getting annoyed by it in dev environment - i'd rather not have my whole screen colored into different sections, even if i'm just testing stuff

kristofbolyai: We need a way to disable debug features / explicitly enable them

common/src/main/java/com/wynntils/features/user/ItemHighlightFeature.java
P0keDev: a helper method within the feature, you mean? i can do that, although it'd need a way to differentiate between hotbar and inventory too

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
P0keDev: no, because it can take a specific `ItemProperty` or a `PropertyType`, and the latter will find the first property present that implements that type

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
P0keDev: the point is to prevent adding multiple instances of the same property, yes - probably not necessary, but might as well be safe. you're right, this should use `hasProperty`

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
P0keDev: maybe DebugFeature should extend UserFeature?

magicus: Or maybe "Debug" is more of an aspect of either User or System Features.

In any case, we should be able to disable features. I think it's coming up soon. :) Or, can't we do it already using commands?

common/src/main/java/com/wynntils/features/user/ItemHighlightFeature.java
magicus: Yes, that what was I meant. 

I just thought it looked like the initial checks where identical between the two methods, up until the point of the actual rendering, so I thought that part could be extracted to a method which returns either null/Color.NONE if we should not render any highlight, and otherwise the color to use.

But it's not super important and it does not have to be done as part of this PR.

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
magicus: But you are scoping T to extend ItemProperty? How can you then send in a PropertyType?

Maybe this is all obvious to y'all, but I'm getting lost in the type hierarchy and casting here. 

What is the purpose of the PropertyType?

common/src/main/java/com/wynntils/features/user/ItemHighlightFeature.java
magicus: I see now that you already did this. Don't know how I missed that.

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
Incompleteusern: I think its suppose to represent properties with multiple implementations.

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
```diff
@@ -35,19 +36,37 @@ public List<Component> getOriginalTooltip() {
...
+     */
+    public boolean hasProperty(Class<?> propertyType) {
         // getProperty returns null if no property of the given type is present
         return (getProperty(propertyType) != null);
     }
+
+    /**
+     * Returns all the present properties of the specified type.
+     */
+    public <T extends PropertyType> List<T> getProperties(Class<T> propertyType) {
```
Incompleteusern, line 65: How would getProperties be used? Would it be used to deal with potential "conflicts" in `PropertyType`s with the same functionality, or am I just confused

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
P0keDev: we can disable user features atm, but debug features don't have any framework for being toggled

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
P0keDev: The `ItemProperty` scope is removed in this PR, so the method can accept both

`PropertyType` is for properties with different implementations that accomplish the same thing. In this PR specifically, regular gear items and cosmetic items differ far too much to handle with the same `ItemProperty` class, but the end goal is the same: draw a highlight behind the item. Thus, the `HighlightProperty` type exists so that the highlight rendering code doesn't have to define separate handling for every applicable `ItemProperty` - if some implementation of `HighlightProperty` is present, it immediately knows how to draw the highlight

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
P0keDev: the intended use is for types that don't necessarily conflict with themselves, such as text overlays. if an item has multiple text overlays on it, there needs to be a way to collect and render all of them

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
magicus: Right, you remove the scope. I misread the diff.

It's still a bit messy in my head, but I think we can blame that more on me than your code. 😉

magicus (approved):

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
P0keDev: it's certainly a bit convoluted, but due to the nature of what we're doing here i think it's either this, or lots and lots of `if` statements (legacy wynntils' approach). i'll try to add some documentation in the repo wiki later so everyone can get a clear picture of it

PR #137, opened 2022-07-04
Remove a few warnings

iwei20:
I'm not a very experienced Java dev/modder, but I thought this would be a fun learning exercise. I tried to find best practices for each of these, though I'm not sure on all of them.

- Most of the warnings were being generated from `AutoCloseable` resource leaks from the `Minecraft` class. These were corrected by holding a single static variable and wrapping any required fields in variables instead of returning `Minecraft` itself. `mc()` still remains. 
  - I couldn't find if `Minecraft.getInstance()` was guaranteed to be nonnull, so that is perhaps one consideration here.
- Deprecated `JsonParser` instantiation was replaced with static method
- Replace array in RequestHandler with `List` to circumvent unchecked conversion
- Suppress unchecked conversion warning for `Class<T>` conversion in `PacketEvent` 
   - I am very fishy about this, because if `GenericEvent` requires the class in its constructor, shouldn't we also?
- Suppress unchecked conversion warning from `Class<HashMap>` to `Class<HashMap<String, String>>` for use with `fromJson`
- Commented out/removed some unused or todo code
- Didn't suppress other required deprecated methods because I thought those might be important to keep in mind and work around (`NativeImage` `makePixelArray` and `sha1` for getting server icon)


kristofbolyai:
I mean this sounds reasonable, but @magicus would know the java side of this better than me.

magicus:
@iwei20 Warnings in Java is not standardized, but up to each java compiler to implement. (This was a really bad idea, btw.) Are you using IntelliJ, or some other compiler to find these warnings?

magicus:
I'm assuming IntelliJ, since I see the same warnings. 

So, first rule when hunting warnings, is to question if the warning is legit. Many times, compilers warn too broadly, which in some situations give too many false positives. If that is the case, the warning should be disabled, rather than having the code changed.

This is the case for the autocloseable Minecraft instance. It's just bad luck that the Minecraft class inherits this interface, and IntelliJ is just being stupid. We can't go create methods for all fields we want to access. Instead, turn off that warning. (I don't think there is a good way to check in a configuration to turn off warnings for everyone, unfortunately :-()

Reverting all these changes can be a tough challenge, so it might be better for you to create new PRs, addressing one warning at a time. That way it is easier to see if the warning was sound, if there is a pattern to the fixes that can result in the need to address the problem at a higher level, etc.

I'll look through the rest of the code and come with some more comments.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -192,9 +192,12 @@ public static boolean tryLoadItemList() {
         handler.addRequest(new RequestBuilder(apiUrls.get("Athena") + "/cache/get/itemList", "item_list")
                 .cacheTo(new File(API_CACHE_ROOT, "item_list.json"))
                 .handleJsonObject(json -> {
-                    translatedReferences = gson.fromJson(json.getAsJsonObject("translatedReferences"), HashMap.class);
-                    internalIdentifications =
-                            gson.fromJson(json.getAsJsonObject("internalIdentifications"), HashMap.class);
+                    @SuppressWarnings("unchecked")
```
magicus: If you think this is worth handling (and I'm tempted to agree, unchecked warnings are bad), using TypeTokens is probably better. See e.g. https://github.com/google/gson/blob/master/UserGuide.md#collections-examples

common/src/main/java/com/wynntils/core/webapi/account/WynntilsAccount.java
```diff
@@ -18,23 +18,22 @@
...
 import java.util.HashMap;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 import javax.crypto.SecretKey;
 import net.minecraft.util.Crypt;
 import org.apache.commons.codec.binary.Hex;
 
 public class WynntilsAccount {
 
-    private static final ScheduledExecutorService service = Executors.newSingleThreadScheduledExecutor(
```
magicus: What was the issue here?

common/src/main/java/com/wynntils/core/webapi/request/RequestHandler.java
```diff
@@ -77,23 +79,21 @@ public Thread dispatchAsync() {
     }
 
     public Thread dispatch(boolean async) {
-        List<Request>[] groupedRequests;
```
magicus, line 80: This is good! Combining `List` with `[]` is utter madness. :-)

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -405,14 +405,16 @@ public Object getTransferData(DataFlavor flavor) throws UnsupportedFlavorExcepti
...
     public static BufferedImage createScreenshot(RenderTarget fb) {
-        NativeImage image = new NativeImage(fb.width, fb.height, false);
-        RenderSystem.bindTexture(fb.getColorTextureId());
-        image.downloadTexture(0, false);
-        image.flipY();
-
-        int[] pixelValues = image.makePixelArray();
         BufferedImage bufferedimage = new BufferedImage(fb.width, fb.height, BufferedImage.TYPE_INT_ARGB);
-        bufferedimage.setRGB(0, 0, fb.width, fb.height, pixelValues, 0, fb.width);
+        try (NativeImage image = new NativeImage(fb.width, fb.height, false)) {
```
magicus, line 409: 👍

common/src/main/java/com/wynntils/utils/objects/Location.java
```diff
@@ -93,7 +93,6 @@ public final BlockPos toBlockPos() {
     }
 
     public Location clone() throws CloneNotSupportedException {
-        Location clone = (Location) super.clone();
```
magicus, line 96: I think you are forgetting that super.clone() can have side effects. You cannot just remove it. I don't remember the details about clone (it was a bad design decision anyway), but I'm pretty sure you are required to call super.clone.

common/src/main/java/com/wynntils/utils/objects/ThrowingBiPredicate.java
```diff
@@ -5,7 +5,6 @@
 package com.wynntils.utils.objects;
 
 @FunctionalInterface
-public interface ThrowingBiPredicate<T, U, Exception extends Throwable> {
-
-    boolean test(T t, U u) throws Exception;
+public interface ThrowingBiPredicate<T, U, EXC extends Throwable> {
```
magicus: Actually just use `E`. The convention is to use single-letter for type templates in Java.

common/src/main/java/com/wynntils/wc/impl/CharacterImpl.java
```diff
@@ -102,10 +100,11 @@ public UUID getId() {
         }
 
         public static CharacterInfo parseCharacter(ItemStack itemStack) {
+            /*
```
magicus: This is half-finished code, yes, but what are you trying to achieve?

common/src/main/java/com/wynntils/mc/event/PacketEvent.java
```diff
@@ -17,6 +17,7 @@
 public class PacketEvent<T extends Packet<?>> extends GenericEvent<T> {
     private final T packet;
 
+    @SuppressWarnings("unchecked") // Technically an unsafe typecast, but nobody wants to pass up the packet class
```
magicus, line 20: This is okay. There cracks where generics were glued into Java are showing up when dealing with `Class`es of generic types.

magicus:
I've looked through all of it now. Everything except the Minecraft instance thing looks okay. I don't know if you can easily revert just that change, since it's so pervasive. If you can, fine, do it, and the rest is OK.

iwei20:
Thanks for the review! Should more of the `Minecraft` field wrapping methods be removed for consistency?

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -192,9 +192,10 @@ public static boolean tryLoadItemList() {
         handler.addRequest(new RequestBuilder(apiUrls.get("Athena") + "/cache/get/itemList", "item_list")
                 .cacheTo(new File(API_CACHE_ROOT, "item_list.json"))
                 .handleJsonObject(json -> {
-                    translatedReferences = gson.fromJson(json.getAsJsonObject("translatedReferences"), HashMap.class);
+                    Type hashmap_type = new TypeToken<HashMap<String, String>>() {}.getType();
```
magicus: Please use normal camelCase. See also `majorIdsType` below.

common/src/main/java/com/wynntils/utils/objects/Location.java
magicus: I'm not sure that the comments adds anything of value? All non-expression statements are in place for their side effects, that's not anything you need to state.

magicus:
Looks better now! I'm not sure how to think about the McUtils wrapping functions. If they are used in just a few places, it might be better to remove them. But that sounds like a separate issue; it's not about fixing warnings, so do it in a separete PR if so.

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
```diff
@@ -46,7 +46,7 @@ public void addProperty(ItemProperty property) {
      */
     public <T> T getProperty(Class<T> propertyType) {
         for (ItemProperty property : properties) {
-            if (property.getClass().equals(propertyType)) return propertyType.cast(property);
```
magicus: This looks like a merge error..?

iwei20: err, I don't think so? I saw the condition changed from main so I took that and combined it with the class cast()

magicus: Agreed. Github presented it strangely to me. Sorry for the noise. :-(

magicus (approved):
This looks good now. Thanks for hanging in there!

PR #139, opened 2022-07-04
Feature: OverlayInfo enabled setting is now honored by Feature class

### LINKED ISSUE #138: Overlays are enabled even when feature is disabled by conditions
If a feature fails to load duo to condition failure, overlays (and possibly keybinds/events) still get loaded.

kristofbolyai:
Fixes #138


common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -61,7 +61,7 @@ private void initOverlays() {
             try {
                 Overlay overlay = (Overlay) FieldUtils.readField(overlayField, this, true);
                 OverlayInfo annotation = overlayField.getAnnotation(OverlayInfo.class);
-                OverlayManager.registerOverlay(overlay, annotation);
+                OverlayManager.registerOverlay(overlay, annotation, shouldEnable);
```
magicus: Am I being anal if I think you can call `this.canEnable()` here instead of passing it as an argument?

magicus:
The bug mentions keybinds as well. Did you verify that these did not have a problem?

kristofbolyai:
> The bug mentions keybinds as well. Did you verify that these did not have a problem?

I will check keybinds, but events work fine and keybinds are being handled at the same time so

common/src/main/java/com/wynntils/core/features/Feature.java
kristofbolyai: Eh, fine by me either way

magicus: Then I think we should not pass instance variables as arguments to private helper methods. :-)

kristofbolyai:
Keybinds are handled properly.

magicus (dismissed):

Incompleteusern (dismissed):

kristofbolyai:
This commit aims to fix the duplicate registration of overlays while not respecting the enabled state. Now, it should be cleaned up, and work as intended (intention is to have OverlayInfo#enabled by changeable by the user).

magicus:
Have you repurposed this PR?

Please refrain from force-pushing to PRs. It is better to merge. Force pushing makes the PR history confusing, old comments do not apply and it's hard to make heads and tails of what's going on.

kristofbolyai:
> Have you repurposed this PR?
> 
> Please refrain from force-pushing to PRs. It is better to merge. Force pushing makes the PR history confusing, old comments do not apply and it's hard to make heads and tails of what's going on.

I only force pushed to amend the latest commit. Also, yes I found a second regarding overlay states. Fixed that too.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -28,18 +28,20 @@ public class OverlayManager {
...
-        }
     }
 
     public static void disableOverlays(List<Overlay> overlays) {
         enabledOverlays.removeIf(overlays::contains);
     }
 
-    public static void enableOverlays(List<Overlay> overlays) {
-        enabledOverlays.addAll(overlays);
+    public static void enableOverlays(List<Overlay> overlays, boolean ignoreState) {
```
magicus, line 37: Is this called with true? I can't find it.

kristofbolyai: It is not right now, but we will when I add Overlay config options so the user can enable/disable them (by commands at least atm).

common/src/main/java/com/wynntils/core/features/overlays/annotations/OverlayInfo.java
```diff
@@ -13,7 +13,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 public @interface OverlayInfo {
-    /* Whether the annotated overlay instantiated at registration */
+    /* Whether the annotated overlay enabled */
```
magicus, line 16:
```suggestion
    /* Whether the annotated overlay is enabled by default */
```

kristofbolyai: Is that actually what OverlayInfo#enabled is? We will use that field to save Overlay enabled config status too (as OverlayManager is in-memory only).  So we will have OverlayManager and OverlayInfo both have the state of the Overlay? I guess we should deal with this in the overlay configs PR.

kristofbolyai: I feel like removing enabled/disabled state from the manager would be the easiest, we don't even need reflection as OverlayInfo is cached to a map already.

kristofbolyai: Anyway, problem of the future, but I wouldn't change that comment.

magicus: Well, now I'm not sure what this means either. :-) What have you done. :) Ok, fine, leave the comment be. We'll make this better when fixing overlay config.

magicus (approved):

Incompleteusern (approved):

PR #141, opened 2022-07-06
Add classifiers to jar files 

DevScyu:
Adds `-fabric` `-forge` and `-quilt` to respective jars


kristofbolyai (approved):
Yes finally. I was too stupid for this.

PR #142, opened 2022-07-06
HealthPotionBlockerFeature: Fix translation key

kristofbolyai:


magicus (approved):
My mistake, I renamed the translation keys to systematically match feature class names, but missed this. :( Thanks for noticing.

PR #143, opened 2022-07-06
QuestInfoOverlayFeature: Add QuestInfoOverlayFeature and FontRenderer

kristofbolyai:
The first actual working overlay in Artemis! 🎉

As expected, this PR is a bit bigger in size, since I had to adapt the Overlay framework to fix issues and make it more intuitive and usable. 

Most important changes: 
- `Overlay` class now has vertical and horizontal alignment overrides. This is done so the alignment we use to save the overlay position does not dictate the rendering alignment. These settings are meant to be config options, exposed to the user. If no override is present, the overlay position alignment is used.
- `FontRenderer` class: This class is quite useful even in it's first iteration. It has support for loads of features; we can draw multi-line texts, aligned to our needs, line broken if a max width is present, and GUI scaled. The actual FontRenderer "API" is quite nice too, it is quite easy to do rendering with the wrapper methods. We can improve this class as we are using it, I did not see the need to create a bunch of hypothetically useful methods. 
- `CommonColors` class: Nice to have, easy colors
- Last but not least, `QuestInfoOverlayFeature`'s `QuestInfoOverlay`. As you can see from the code, it is really clean and easy to add overlays.

There are a few smaller changes, mostly based around the changes listed above. 

Few screenshots of the overlay (ignore the red box, it was used as a debug tool):

Alignments: 
  image
  image
  image

GUI scaling:
image
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.user;
+
+import com.mojang.blaze3d.platform.Window;
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.core.features.overlays.Overlay;
+import com.wynntils.core.features.overlays.OverlayPosition;
+import com.wynntils.core.features.overlays.annotations.OverlayInfo;
+import com.wynntils.core.features.overlays.sizes.GuiScaledOverlaySize;
+import com.wynntils.mc.event.RenderEvent;
+import com.wynntils.mc.render.FontRenderer;
+import com.wynntils.mc.render.HorizontalAlignment;
+import com.wynntils.mc.render.LineRenderTask;
+import com.wynntils.mc.render.VerticalAlignment;
+import com.wynntils.utils.objects.CommonColors;
+import com.wynntils.wc.utils.scoreboard.quests.QuestInfo;
+import com.wynntils.wc.utils.scoreboard.quests.QuestManager;
+import java.util.ArrayList;
+import java.util.List;
+
+public class QuestInfoOverlayFeature extends UserFeature {
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay QuestInfoOverlay = new QuestInfoOverlay();
+
+    public static class QuestInfoOverlay extends Overlay {
+        public QuestInfoOverlay() {
+            super(
+                    new OverlayPosition(
+                            5,
+                            -5,
+                            VerticalAlignment.Top,
+                            HorizontalAlignment.Right,
+                            OverlayPosition.AnchorSection.TopRight),
+                    new GuiScaledOverlaySize(300, 50),
+                    HorizontalAlignment.Left,
+                    VerticalAlignment.Middle);
+        }
+
+        @Override
+        public void render(PoseStack poseStack, float partialTicks, Window window) {
+            QuestInfo currentQuest = QuestManager.getCurrentQuest();
+
+            if (currentQuest == null) {
+                return;
+            }
+
+            List<LineRenderTask> toRender = new ArrayList<>();
+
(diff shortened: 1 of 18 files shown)


Incompleteusern:
Unironically the red box is stylish.

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public FontRenderer() {
+        this.font = ((MinecraftAccessor) McUtils.mc()).getFont();
+    }
+
+    public static FontRenderer getInstance() {
+        return INSTANCE;
+    }
+
+    public int drawLine(
```
magicus: Maybe "renderText"? "drawLine" sounds like it will draw a straight line on the screen.

magicus:
Looks fantastic! Try to get some better names for the "drawLine" family of function, and you're all set.

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
kristofbolyai: Fixed.

magicus (approved):
🚢 it!

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
```diff
@@ -0,0 +1,75 @@
...
+
+            toRender.add(TextRenderTask.getWithHorizontalAlignment(
+                    "Tracked Quest Info:", this.getWidth(), CommonColors.GREEN, this.getRenderHorizontalAlignment()));
+            toRender.add(TextRenderTask.getWithHorizontalAlignment(
+                    currentQuest.quest(), this.getWidth(), CommonColors.ORANGE, this.getRenderHorizontalAlignment()));
+            toRender.add(TextRenderTask.getWithHorizontalAlignment(
+                    currentQuest.description(),
+                    this.getWidth(),
+                    CommonColors.WHITE,
+                    this.getRenderHorizontalAlignment()));
```
HeyZeer0, line 61: Be careful with this pattern, this is going to generate a bunch of garbage and significantly reduce the time between each Young GC causing lag spikes. A good rule for rendering is to always try splitting logic from the actual rendering code,since whatever you place inside the render block will run each frame!

kristofbolyai: I can extract toRender and update the text, so there are no new instances if it. I will do this later.

kristofbolyai: Actually, I can't extract it as it can change. @HeyZeer0 do you have something in mind to optimize this as this is already pretty optimized imo?

magicus: I agree with HeyZeer0 that rendering code is one of the few places were we need to be performance conscious. 

I think you can create the toRender list ahead of time, and just call the render method in the render callback.

You might need to create some communication when task changes. Perhaps the TrackedQuestManager needs to send an event when the tracked quest changes? Then the feature can just listen to that event, and modify the toRender list. And if the tracked quest disappear, just set toRender to null, and skip it in the render method.

kristofbolyai: The problem with making them static is not the strings, but the overlay alignment changes.

magicus: If the user changes configuration you mean?

magicus: I thought we said that we should have a way for features to get a callback when a config change is happening. That will probably need to be extended to cover overlays, too.

kristofbolyai: Yeah

kristofbolyai: > I thought we said that we should have a way for features to get a callback when a config change is happening. That will probably need to be extended to cover overlays, too.

Yeah, but we don't have that right now

magicus: Ok, so when we fix that, we can update the render function to be more efficient.

PR #147, opened 2022-07-07
Overlays: Support configurations

### LINKED ISSUE #148: `RenderEvent` does not work get fired on Forge
This is caused by the overriding method in forge's GUI wrapper. Not sure if we can work around this.

kristofbolyai:
This pull request is quite big, reviewing might be easier per commit. 

Changes:
- Fixed several crashes related to configs
- `ConfigHolder`/`ConfigManager` now supports overlays
- `WynntilsConfigCommand` also has full overlay support. This command also got a bit of a refactor.
- Separated `TextRenderTask` into line and `TextRenderSetting`.
- Removed `DummyOverlayFeature`.
- `QuestInfoOverlay` now uses the optimized render method.

EDIT: 
Fortunately I opened the forge version while setting up my new development space for Artemis in Linux (super fast btw!), to come to the unfortunate realization that forge breaks at least 1 of our mixins/events (#148). The only solution to this is, having forge have "forge only" mixins, and posting the events there. See https://github.com/Wynntils/Artemis/pull/147/commits/1ae272e57ca7be0726eff090a3c778413d95f6dc for implementation. This also means that this PR fixes #148. Moreover, I guess we now have support for forge only mixins. :cold_sweat:


forge/src/main/java/com/wynntils/forge/mixins/ForgeIngameGuiMixin.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraftforge.client.gui.ForgeIngameGui;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ForgeIngameGui.class)
+public abstract class ForgeIngameGuiMixin {
+
+    // TODO: This is kinda hacky, but we can't shadow minecraft here
```
magicus: Unless you actually intend to fix it, don't mark it as a TODO. (I think it's an okay solution, for all practical purposes there ever only is a single Minecraft instance. I think this passing around of it mostly is a (mis-guided?) attempt to optimize).

forge/src/main/java/com/wynntils/forge/mixins/ForgeIngameGuiMixin.java
```diff
@@ -0,0 +1,33 @@
...
+    //       so we have to use the instance.
+
+    @Inject(method = "render", at = @At(value = "HEAD"))
+    public void onRenderGuiPre(PoseStack poseStack, float partialTick, CallbackInfo ci) {
+        EventFactory.onRenderGuiPre(
```
magicus, line 24: I think you should add a comment to the "common" injection that it does not apply to Forge, and reference this class.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -47,6 +49,12 @@ public String getFieldName() {
     }
 
     public String getJsonName() {
+        if (parent instanceof Overlay) {
+            // "FeatureName.OverlayName.settingName"
+            return parent.getClass().getDeclaringClass().getSimpleName() + "."
+                    + parent.getClass().getSimpleName() + "." + field.getName();
+        }
+        // "FeatureName.settingName"
         return parent.getClass().getSimpleName() + "." + field.getName();
```
magicus: Do we want to normalize the names to initial-lowercase? Like how we do for the translation keys. so like "fakeFeature.dummyOverlay.horizontalSize" etc.

In my eyes, that'd look better.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -62,14 +70,14 @@ public String getDisplayName() {
         if (!getMetadata().key().isEmpty()) {
             return I18n.get(getMetadata().key() + ".name");
         }
-        return parent.getTranslation(field.getName() + ".name");
```
magicus, line 65: Are we sure parent is Translatable? If so, maybe we should even declare it as such in the constructor.

Otherwise, nowadays there is a nice pattern matching instanceof operation:
`if (!parent instanceof Translatable translatabe) return;` and then you can use `translatable` as a correctly casted alias for parent.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -85,7 +93,7 @@ public Object getValue() {
     public boolean setValue(Object value) {
         try {
             FieldUtils.writeField(field, parent, value, true);
-            parent.updateConfigOption(this);
+            ((Configurable) parent).updateConfigOption(this);
```
magicus: Actually, I think it *would* make sense to declare parent as `Configurable`.

magicus: ... and maybe Configurable should implement/extend Translatable... I realize it's a bit of mixing of concerns, but there is no good way in Java either to require that the parent sent to the constructor is both Configurable and Translatable.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -26,15 +27,14 @@
 import java.util.ArrayList;
 import java.util.List;
 import org.apache.commons.lang3.reflect.FieldUtils;
+import org.jetbrains.annotations.NotNull;
```
magicus: Kill it with 🔥 ! :-)

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -121,14 +121,33 @@ private static List<ConfigHolder> collectConfigOptions(Feature feature) {
         if (featureInfo == null || featureInfo.category().isBlank()) return null;
 
         String category = featureInfo.category();
-        List<ConfigHolder> options = new ArrayList<>();
 
-        // collect options
-        for (Field f : FieldUtils.getFieldsWithAnnotation(feature.getClass(), Config.class)) {
-            Config metadata = f.getAnnotation(Config.class);
-            options.add(new ConfigHolder(feature, f, category, metadata));
+        loadFeatureOverlayConfigOptions(feature, category);
```
magicus: "feature, category" or "category, feature"? :-D Chose one and stick to it...

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -121,14 +121,33 @@ private static List<ConfigHolder> collectConfigOptions(Feature feature) {
...
+
+        return getConfigOptions(category, feature);
+    }
+
+    private static void loadFeatureOverlayConfigOptions(Feature feature, String category) {
+        // collect feature's overlays' config options
+        for (Overlay overlay : feature.getOverlays()) {
+            List<ConfigHolder> options = getConfigOptions(category, overlay);
+
+            CONFIG_HOLDERS.addAll(options);
```
magicus: These actions are shared with `registerFeature`. All must be done to correctly register options. It should be extracted to a common method and called from both here and `registerFeature`.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -4,24 +4,42 @@
...
+public abstract class Overlay implements Translatable, Configurable {
+    private List<ConfigHolder> configOptions = new ArrayList<>();
+
+    @Config
     protected OverlayPosition position;
 
+    @Config
     protected OverlaySize size;
 
+    @Config(key = "overlay.wynntils.overlay.userEnabled")
```
magicus, line 33: What does the `key` here mean? Does that not lead to all overlays being turned on/off at the same time? Also, in terms of getting a nice looking user fronting name, would not just "enabled" be better than "userEnabled"?

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -47,6 +65,49 @@ public Overlay(
 
     public abstract void render(PoseStack poseStack, float partialTicks, Window window);
 
+    @Override
+    public final void updateConfigOption(ConfigHolder configHolder) {
+        // if user toggle was changed, enable/disable feature accordingly
+        if (configHolder.getFieldName().equals("userEnabled")) {
```
magicus, line 73: Once again we're running into "enabled" as a special case. I'm getting more and more firm in my belief that the "enabled" status is not an aspect of the thing (feature, overlay) *itself* but rather on the owner (feature manager, or the feature owning the overlay (or possibly the overlay manager).

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -47,6 +65,49 @@ public Overlay(
...
+                OverlayManager.disableOverlays(List.of(this));
+            }
+            return;
+        }
+
+        // otherwise, trigger regular config update
+        onConfigUpdate(configHolder);
+    }
+
+    protected void onConfigUpdate(ConfigHolder configHolder) {}
```
magicus: I was like "did I not just look at that method?!?" and it took me some time to figure out that they had different names. I remember that there were some kind of reason for this split, but I am questioning it. This is sooo confusing. Can we redesign this to get rid of the split? What was the reason? Is it the "enabled" thingy again? If so, I swear I'm betting more and more upset about it. :-/

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -47,6 +65,49 @@ public Overlay(
...
+
+    /** Gets the name of a feature */
+    @Override
+    public String getTranslatedName() {
+        return getTranslation("name");
+    }
+
+    @Override
+    public String getTranslation(String keySuffix) {
+        return I18n.get("overlay.wynntils." + getNameCamelCase() + "." + keySuffix);
```
magicus: This makes it sound like overlays do not belong to a feature.

Maybe that is a better solution. Maybe we should offer "features" as well as "overlays", and make both first-class citizens. I think that would be a completely valid choice. I would support a redesign for this end.

But the current design says overlays belong to features. If we want to stick with that, then this should be reflected in the key.

common/src/main/java/com/wynntils/commands/wynntils/WynntilsConfigCommand.java
```diff
@@ -332,4 +489,127 @@ private static int resetAllConfigOptions(CommandContext<CommandSourceStack> cont
...
+        response.append(new TextComponent("Subcategory: ")
+                        .withStyle(ChatFormatting.WHITE)
+                        .append(new TextComponent(config.getMetadata().subcategory())))
+                .append("\n");
+        response.append(new TextComponent("Description: ")
+                        .withStyle(ChatFormatting.WHITE)
+                        .append(new TextComponent(config.getDescription())))
+                .append("\n");
+        return response;
+    }
```
magicus, line 614: I think this looks good (barring the discussion of how overlay config names should be built, if they should belong to a feature or not).

But.

The entire class is very repetitive. At some point, we might consider seeing if we can raise the abstraction a bit, or if that is just hopeless given the Brigadier API...

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -30,6 +35,24 @@ public OverlayPosition(
         this.horizontalOffset = horizontalOffset;
     }
 
+    public OverlayPosition(String string) {
```
magicus, line 38: Maybe better make this a static factory  method `from` than a constructor?

Where is this used? In json config file parsing? I'm a bit worried about the ordering requirement for the fields, it will look like you can give them in any order but this will not be accepted.

kristofbolyai: It is used in config parsing reflection, for the WynntilsConfigCommand. I didn't make it accept any order, as we will only use this for the command, and the actual config gui will not need this.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -122,5 +122,17 @@
   "feature.wynntils.tooltipScale.universalScale.name": "Universal Scale",
   "feature.wynntils.userFeature.userEnabled.description": "Should this feature be enabled?",
   "feature.wynntils.userFeature.userEnabled.name": "Enabled",
-  "feature.wynntils.wynncraftButton.name": "Wynncraft Button Feature"
+  "feature.wynntils.wynncraftButton.name": "Wynncraft Button Feature",
+  "overlay.wynntils.overlay.horizontalAlignmentOverride.description": "Set this if you want to override the position horizontal alignment",
```
magicus, line 129: Right, now I understand (I think). For translation keys, we want the key to be unique per "kind" of configuration for that overlay. But when storing config for these values, we need a config json key that is unique per instance of overlay.

I got confused since for features we have had a 1-to-1 relationship between "kinds" of features and instances of them.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
kristofbolyai: (See the ConfigHolder for usage.)

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: Yeah for config values in abstract classes, we only need to define per superclass, not per every implementation.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
magicus: I see. Then I guess it is okay. We can always extend it later, if needed. But you can still make it a factory method instead of a constructor?

common/src/main/java/com/wynntils/commands/wynntils/WynntilsConfigCommand.java
kristofbolyai: Yeah, I tried my best, but wasn't really my intention to recode this fully.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: We can't really do anything about this :(.

magicus:
I apologize if the review came in a strange order and sounded more rambling than usual. :) I tried reviewing per commit, and it had some merits, but it also made it harder to see the whole picture, or find the right place to add comments.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: Feel free to rework it :). Aside from the joke, yeah, perhaps that is better, once again I don't want to rewrite the whole Feature/Overlay impl this pr.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
kristofbolyai: Make parent Configurable and throw RuntimeException in ctor if it is not Translatable? I don't want to mix the two into a singular thing?

@magicus, do you accept this?

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
kristofbolyai: That would require changing the ConfigHolder value parsing to include even more reflection stuff? Isn't it fine to leave it as a string constructor?

magicus: Fine, leave it as a constructor.

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
```diff
@@ -48,24 +82,15 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
...
-            toRender.add(TextRenderTask.getWithHorizontalAlignment(
-                    "Tracked Quest Info:", this.getWidth(), CommonColors.GREEN, this.getRenderHorizontalAlignment()));
-            toRender.add(TextRenderTask.getWithHorizontalAlignment(
-                    currentQuest.quest(), this.getWidth(), CommonColors.ORANGE, this.getRenderHorizontalAlignment()));
-            toRender.add(TextRenderTask.getWithHorizontalAlignment(
-                    currentQuest.description(),
-                    this.getWidth(),
-                    CommonColors.WHITE,
-                    this.getRenderHorizontalAlignment()));
+            toRender[1].setText(currentQuest.quest());
```
magicus, line 85: As a future possible enhancement, we can have the quest manager send out an even when the quest changes, so we need to set the text only once and not once per render. But we can skip it for now.

kristofbolyai: I thought about that, and I don't think the cost of registering overlays into event bus, then listening for an event is much more optimized than setting a string reference.

magicus (approved):

PR #149, opened 2022-07-08
QuestInfoOverlay: Fix quest parsing and rendering

kristofbolyai:


common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -81,7 +81,7 @@ public void renderText(
         int line = 0;
         int partBegin = 0;
 
-        for (int i = parts.size() - 1; i >= 0 && partBegin < parts.size() - 1; i--) {
```
magicus, line 84: Maybe this is a simple as it can get, and/or maybe it doesn't matter, but this is really complicated logic. As you might have noticed, I have this "don't make me think!" attitude about code. :-) Everything that is not immediately clear to me is "sus". :-D

I wonder if it's possible to make the logic clearer. For one, are you rendering this "backwards"? Why?

(I did not really react on this when you created the code, but now when you change it I realize I don't understand what your fix is doing...)

magicus: If you say "whatever, not important, leave it be" I will accept that, too. I know we have more important issues to address.

kristofbolyai: The loop is backwards so we can get the biggest part that we can fit in the maxWidth.

kristofbolyai: I think this loop always has to be a bigger one, I will add a clarifying comment for the best of the two worlds.

kristofbolyai: I think we could use recursion here, or this loop, where I much rather prefer loops.

kristofbolyai:
I could refactor this code to be a bit smaller (and remove the duplicate for loop), but that is not meant to be the scope of this PR, and I (or someone else if they will) can just do it separately.

magicus (approved):

magicus:
If you ever feel motivated to clean up the text rendering, feel free to do it. :-) Otherwise, it's not that important.

PR #150, opened 2022-07-08
WynncraftPauseScreenFeature: Add WynncraftPauseScreenFeature

kristofbolyai:
Looks:

image


common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
```diff
@@ -0,0 +1,72 @@
...
+        renderables.set(3, classSelection);
+
+        Button hub = replaceButtonFunction(
+                (Button) renderables.get(4),
+                new TranslatableComponent("feature.wynntils.wynncraftPauseScreen.hubButton.name"),
+                (button) -> {
+                    McUtils.mc().setScreen(null);
+                    McUtils.mc().mouseHandler.grabMouse();
+                    McUtils.player().chat("/hub");
+                });
```
Incompleteusern, line 57: Instead of returning the button and setting afterwards, `replaceButtonFunction` could be fed only the number and automatically do the replacing given its name

Incompleteusern (commented):
LGTM. We could also make the `Wynncraft.com` button redirect to the wynncraft website.

Incompleteusern (dismissed):

common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
kristofbolyai: I didn't want to pass around the list. Old habbit from C++.

Incompleteusern: Makes sense

common/src/main/java/com/wynntils/mc/EventFactory.java
```diff
@@ -161,7 +161,7 @@ public static void onScreenCreated(Screen screen, Consumer<AbstractWidget> addBu
         if (screen instanceof TitleScreen titleScreen) {
             post(new TitleScreenInitEvent(titleScreen, addButton));
         } else if (screen instanceof PauseScreen gameMenuScreen) {
```
magicus:
```suggestion
        } else if (screen instanceof PauseScreen pauseMenuScreen) {
```
(and below)

common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
```diff
@@ -0,0 +1,72 @@
...
+
+        renderables.set(4, hub);
+
+        event.getPauseScreen().clearWidgets();
+
+        for (Widget renderable : renderables) {
```
magicus, line 63: I thought you should only add the new buttons? Now it looks like you re-add all the old buttons as well.

kristofbolyai: Yeah, we clear all buttons and re-add all. There is no nice way to replace buttons at specific indexes.

magicus: Ah, now I see, you do clearWidgets() first. Hm. Since you want to replace buttons. Hm. Yeah, maybe this is the best way.

(For the record: I'm not sure this is the ideal set of buttons in the pause menu. But we can revisit this when we have more actual functionality.)

magicus:
The screenshot is from some different code than this? There is no Territory Map or Wynncraft.com in this code, what I can see?

Incompleteusern:
That's due to the wynncraft texture pack iirc

magicus (approved):

kristofbolyai:
> The screenshot is from some different code than this? There is no Territory Map or Wynncraft.com in this code, what I can see?

I think that is changed by Wynncraft itself. Let me make sure on a vanilla version.

magicus:
Oh, can you do such things in Vanilla? I had no idea. How does that work?

kristofbolyai:
Vanilla: 
image

magicus:
Ah, I get it. They have changed the text for "Open for LAN" and "Achievements" in the English locale. I have a non-English locale selected, so I have never seen this.

PR #152, opened 2022-07-09
Features: Add ItemLockFeature

kristofbolyai:
Works as expected. Single limitation is commented in https://github.com/Wynntils/Artemis/commit/78ccc63c2ea3b3f4634e06648eec766d59ac3808.

I am mostly happy with the feature itself, and the Character implementation is turning out to be consistent, I have not had a single problem with it during testing. 

The only thing which stands out is the config changes. Again. I managed to shoehorn map parsing into this PR, but I am not that happy with it. We need some special kind of `ConfigHolder` for special cases, like this, where GSON needs a little help. Feel free to discuss the implementation ideas here. However, I would rather work on that in a separate PR (even right after this one is merged). (Or maybe you like this implementation for the special configs? I doubt that.) If you want me to remove config changes, I suppose I can remove them, and have item lock slots not be persistent until we figure this out. I think I already got an idea while writing this up, but I want to hear your opinions :). 

EDIT: https://github.com/Wynntils/Artemis/pull/152/commits/49951cdb907ab4ce5bacd368b486d5fb6f0213e8 should handle type overwrites pretty good. There is no easy way to do this. The PR also includes a few serialization problem fixes I found while testing it. 

Demo:
image


common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -89,7 +89,14 @@ private static void loadConfigOptions(List<ConfigHolder> holders) {
...
             JsonElement holderJson = configObject.get(holder.getJsonName());
-            Object value = gson.fromJson(holderJson, holder.getType());
+            Object value;
+
+            if (holder.getTypeOverride() == null) {
+                value = gson.fromJson(holderJson, holder.getType());
+            } else {
+                value = gson.fromJson(holderJson, holder.getTypeOverride());
+            }
+
```
P0keDev: are you certain all of this (and the related type handling elsewhere) is necessary? i'm pretty sure gson is capable of automatically determining types in a map

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -0,0 +1,160 @@
...
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.inventory.ClickType;
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@FeatureInfo(category = "IWillRemoveThisWhenIFix#151")
+public class ItemLockFeature extends UserFeature {
```
P0keDev, line 41: the order of methods and variables in this class is very jumbled imo - variables and feature/event/helper methods should all be grouped together

common/src/main/java/com/wynntils/core/config/Config.java
```diff
@@ -20,4 +20,7 @@
 
     /** Whether this option should be visible to users */
     boolean visible() default true;
+
+    /** Internal configs ignore ConfigHolder's userEdited field and always save */
+    boolean internal() default false;
```
P0keDev: maybe instead of requiring this extra flag, `ConfigHolder` should check whether `userEdited` is true _or_ if the field is in a non-default state. previously these two statements were redundant because we only edited fields through the holders, but it would work as a catch-all that doesn't need us to supply this extra info

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -126,6 +132,9 @@ public boolean setValue(Object value) {
     }
 
     public boolean isUserEdited() {
+        if (this.metadata.internal()) {
+            return true;
+        }
         return userEdited;
     }
```
P0keDev: whether or not this gets changed according to my other comment, this method should probably get renamed to something more universal, like `shouldSave`

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -0,0 +1,160 @@
...
+                8,
+                8,
+                Texture.ITEM_LOCK.width() / 2,
+                Texture.ITEM_LOCK.height() / 2);
+    }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onInventoryClickEvent(ContainerClickEvent event) {
+        if (!(McUtils.mc().screen instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
+        if (event.getClickType() != ClickType.THROW) return;
```
P0keDev: i would like to see the full functionality from wynntils legacy ported, i.e. an option to block _all_ interaction with a slot as opposed to just dropping it

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: i did not find a way to make gson handle it properly and we do the same in WebManager.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -89,7 +92,14 @@ private static void loadConfigOptions(List<ConfigHolder> holders) {
 
             // read value and update option
             JsonElement holderJson = configObject.get(holder.getJsonName());
-            Object value = gson.fromJson(holderJson, holder.getType());
+            Object value;
+
+            if (holder.getTypeOverride() == null) {
```
magicus: Why is this logic not part of holder.getType? Seems to me that the "override" should be internal to the ConfigHolder.

kristofbolyai: holder.getType returns a `Class<?>` and typeOverride is an actual Type. Can I just cast the Type to Class<? extends Type>? Not sure.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -6,10 +6,12 @@
 
 import com.wynntils.mc.render.HorizontalAlignment;
 import com.wynntils.mc.render.VerticalAlignment;
+import java.io.Serializable;
+import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public class OverlayPosition {
+public class OverlayPosition implements Serializable {
```
magicus: Is this necessary? Serialization is soo flawed and very tricky to get right. Is it json stuff that requires this?

kristofbolyai: Yeah it just needed the implementation, no other changes.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -83,6 +85,23 @@ public String toString() {
                 + anchorSection + '}';
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        OverlayPosition that = (OverlayPosition) o;
```
magicus: I think there is a helper class in one of the google libraries we are including that create a general equals by comparing the fields of the class. Don't remember its name, but I think it's better to use it, otherwise it's easy to like add a field and miss updating equals etc. (If you can't find it, look in Legacy where I think it is used here and there, or let me know and I'll go look it up.)

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -83,6 +85,23 @@ public String toString() {
...
+        return verticalOffset == that.verticalOffset
+                && horizontalOffset == that.horizontalOffset
+                && verticalAlignment == that.verticalAlignment
+                && horizontalAlignment == that.horizontalAlignment
+                && anchorSection == that.anchorSection;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(verticalOffset, horizontalOffset, verticalAlignment, horizontalAlignment, anchorSection);
```
magicus: I think it has a hash support function, too.

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -0,0 +1,163 @@
...
+
+@FeatureInfo(category = "IWillRemoveThisWhenIFix#151")
+public class ItemLockFeature extends UserFeature {
+
+    @RegisterKeyBind
+    private final KeyHolder lockSlotKeybind = new KeyHolder(
+            "Lock Slot", GLFW.GLFW_KEY_H, "Wynntils", true, ItemLockFeature::tryChangeLockStateOnHoveredSlot);
+
+    @Config(visible = false)
+    private static Map<Integer, Set<Integer>> classSlotLockMap = new HashMap<>();
```
magicus, line 48: Should this not be `Map<UUID, Set<Integer>>`?

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @SubscribeEvent
+    public void onContainerRender(ContainerRenderEvent event) {
+        if (!(event.getScreen() instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
+
+        Character character = WynnUtils.getCharacter();
+
+        if (character == null || character.getCharacterInfo() == null) return;
+
+        for (Integer slotId :
```
magicus, line 71: I'm vaguely worried that we have a loop in render code. In all practical situations, this is unlikely to matter much, since there are only so many characters you can have. But as a general principle I really want to see rendering code being as minimal as possible. You could extract the inner set of locked slots into a cache, and update that cache whenever the player switches character. I'm pretty sure the Character model sends out an event when that happens, that you could listen to.

Edit: I misread the code. The loop is over slots, not characters. Nevermind.

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
kristofbolyai: Character impl was changed for the time being.

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -0,0 +1,163 @@
...
+    private static Map<Integer, Set<Integer>> classSlotLockMap = new HashMap<>();
+
+    @TypeOverride
+    private static final Type classSlotLockMapType = new TypeToken<HashMap<Integer, Set<Integer>>>() {}.getType();
+
+    @Config
+    public static boolean blockAllActionsOnLockedItems = false;
+
+    @SubscribeEvent
+    public void onKeyPress(InventoryKeyPressEvent e) {
```
magicus, line 57: Why is this not handled by registering our KeyHolder? Is there a special extra step for handling key presses in the inventory? Should we perhaps pass an additional boolean includeContainerKeypress to the KeyHolder constructor, to have it do this?

common/src/main/java/com/wynntils/utils/ObjectUtils.java
```diff
@@ -0,0 +1,111 @@
...
+public class ObjectUtils {
+    /**
+     * Returns a copy of the object, or null if the object cannot
+     * be serialized.
+     */
+    public static Object copy(Object orig) {
+        Object obj = null;
+        try {
+            // Make a connected pair of piped streams
+            PipedInputStream in = new PipedInputStream();
```
magicus: Yikes! Can't we use clone instead?

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
kristofbolyai: MoreObjects

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
kristofbolyai: I was not sure, itemscreenshot feature does the same "hack". Perhaps make an issue and deal with it later?

common/src/main/java/com/wynntils/utils/ObjectUtils.java
kristofbolyai: Object doesn't implement cloneable by default? If you can figure out a better alternative, please say so, I don't like this that much either.

common/src/main/java/com/wynntils/wc/model/Character.java
```diff
@@ -90,17 +104,32 @@ public int getLevel() {
...
+                }
+
+                matcher = CLASS_PATTERN.matcher(line);
+
+                if (matcher.matches()) {
+                    classType = ClassType.fromName(matcher.group(1));
+                }
             }
 
-            return new CharacterInfo(null, false, 0, UUID.randomUUID());
```
magicus, line 103: Oh, I never finished this! 😊 

I could have sworn I did a try-to-find-best-matching character, and if none were found, return its UUID. But apparently I only had the design in my head but no implementation. :(

I should get back to this and try to get a implementation that is as robust as we can get, and which supports autojoin etc. I believe we can get it pretty reliable.

common/src/main/java/com/wynntils/utils/ObjectUtils.java
magicus: To begin with, your implementation can be replaced with the one in `SerializationUtils#clone`. It does the same (possibly with almost identical code) but we don't have to maintain it. The main restriction here is that the classes need to be serializable.

Another suggestion I got from the interwebz was to use e.g.: 
```
User deepCopy = gson.fromJson(gson.toJson(orig), User.class);
```

I'd prefer that over serialization, if I have to chose. (And if it works okay.)

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: So the two calls to `fromJson` are actually overloaded with different argument types. I see. It might be that there is nothing to be done then.

I'm not sure you can cast like that, but I'm guessing you'd be able to create a Type given a `Class<?>`, so it's possible you can have getType return a Type in all situations, the overrideType if it's specified, and a Type created from the Class otherwise.

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
magicus: I saw that further down. I'll try to get some time to fix it properly. In the meantime, this'll do.

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
magicus: Make an issue and deal with it later. It's a winning concept, pushing work to future you (or future us). :-) 

(Have you tested it it does not work without it?)

kristofbolyai: It does not work.

kristofbolyai: Tracked by #153.

kristofbolyai:
This should be good now :).

common/src/main/java/com/wynntils/mc/mixin/LocalPlayerMixin.java
```diff
@@ -29,4 +31,11 @@ private void redirectChat(ClientPacketListener connection, Packet<?> packet, Str
 
         connection.send(new ServerboundChatPacket(result.getMessage()));
     }
+
+    @Inject(method = "drop", at = @At("HEAD"), cancellable = true)
+    private void onDropPre(boolean fullStack, CallbackInfoReturnable<Boolean> cir) {
```
magicus, line 36: Could this not be just a normal CallbackInfo?

kristofbolyai: MC development plugin suggests that (as it should match method return).

kristofbolyai: And drop returns boolean.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -35,7 +38,7 @@ public class ConfigManager {
     private static final List<ConfigHolder> CONFIG_HOLDERS = new ArrayList<>();
     private static File userConfig;
     private static JsonObject configObject;
-    private static Gson gson;
+    static Gson gson;
```
magicus: Not 100% happy with exposing fields like this, but I'm willing to turn a blind eye to for this time. :)

magicus (dismissed):

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: We need this gson instance so it has the same adapters both in ConfigHolder and ConfigManager.

kristofbolyai: If you can suggest a better alternative, let me know.

magicus:
I think I found a way to get unique names for characters, btw. (Not affecting this PR, just sharing the good news :-))

The Wynncraft API allows you to query users. Here is my user: [ca3ee631-a36c-48f4-9007-3c3c52c79840](https://api.wynncraft.com/v2/player/ca3ee631-a36c-48f4-9007-3c3c52c79840/stats)

The `classes[].name` field seem to combine class and a unique number. I've got both `archer` and `archer1`, etc. I'm trying to figure out if you have like `mage`, `mage1` and `mage2`, and delete the character corresponding to `mage1`, that the other ones still keep their name. Hopefully they do, so the names are stable.

Then we only need to figure out which of these character we actually selected. Hopefully the combination of class and level is unique. If not, we'll have to do more thorough analysis.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: The standard solution is to make a public accessor :-). I know it adds boilerplate but I think it's good to try to stick with the practice of keeping fields private. (As Mojang most definitely not do :))

magicus (approved):

PR #154, opened 2022-07-09
ConfigManager: Remove FeatureInfo annotation requirement

### LINKED ISSUE #151: UserFeatures config problems
I was going to create a PR fixing this, but this one is a controversial topic, so I thought an issue might be better to start with. 

`ConfigManager` has a requirement that a `Feature` must be annotated with `FeatureInfo` to get the `Feature`'s category or else config options won't load. This is obviously not a good thing. 

Here is the things we could do: 
- Make `FeatureInfo` annotation required by throwing Exception on feature registration 
- Make `ConfigManager` assume blank ("") category for features without annotation (the annotation already has `""` as default category).

kristofbolyai:
Fixes #151.

### DIFF
--- a/common/src/main/java/com/wynntils/core/config/ConfigManager.java
+++ b/common/src/main/java/com/wynntils/core/config/ConfigManager.java
@@ -131,7 +131,7 @@ public class ConfigManager {
         FeatureInfo featureInfo = feature.getClass().getAnnotation(FeatureInfo.class);
 
         String category = "";
-        // feature has no category defined, can't create options
+        // feature has no category defined, default to empty category
         if (featureInfo != null) {
             category = featureInfo.category();
         }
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -129,10 +129,12 @@ public static void saveConfig() {
 
     private static List<ConfigHolder> collectConfigOptions(Feature feature) {
         FeatureInfo featureInfo = feature.getClass().getAnnotation(FeatureInfo.class);
-        // feature has no category defined, can't create options
-        if (featureInfo == null || featureInfo.category().isBlank()) return null;
 
-        String category = featureInfo.category();
+        String category = "";
+        // feature has no category defined, can't create options
```
P0keDev: i think you forgot to change this comment? since it's no longer true that we ignore options

P0keDev (approved):

PR #155, opened 2022-07-10
ConfigManager: Create default config file

kristofbolyai:
A file named 'default.conf.json` is created on every start, including every config holder. This can be used to modify settings without guessing the JSON format.


magicus:
Can you post an example of this file?

kristofbolyai:
> Can you post an example of this file?

https://pastebin.com/pu83KDVL

magicus:
The line 
```
  "itemHighlightFeature.inventoryOpacity": 1.0,{
```
looks very broken. Just a copy/paste error, or some exporting issues?

Also lines 93-94:
```
  "itemLockFeature.userEnabled": true
}
```

magicus (approved):
The code looks good as far as I can tell, but this approval assumes the errors in the json file were a copy/paste error. If the file is in fact malformed, this needs to be fixed before integrating.

kristofbolyai:
Rechecked, and it was a paste error.

PR #160, opened 2022-07-12
1.12-like tooltip wrapping

### LINKED ISSUE #129: Item Tooltip Sizing and Positioning
When I halve my Minecraft window to one side of the screen (yes, I do that sometimes - don't beat me - for item management stuff where I need several things next to each other), the tooltip on some items isn't inside the game window bounds, but it pushed out to the side. It becomes unreadable when the left half of a tooltip is missing.

TomaSajt:
This PR fixes tooltips going past the screen when hovering (with `ItemGuessFeature` turned on)
As stated in #129 vanilla text-wrapping doesn't seem to be implemented. However Forge has a built-in way of handling this, so I took the code from `ForgeHooksClient` and put them in the pre-existing mixin. This fix seems to work fine on all 3 modloaders. (I did not put in a special case for Forge, but it doesn't seem to show any problems, so I'll leave it like that).


kristofbolyai:
Mostly I would like confirmation from @P0keDev as they wrote this redirect.

kristofbolyai (dismissed):
Seems fine. Thank you for fixing this.

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
```diff
@@ -80,12 +87,35 @@ private void renderTooltipPre(
         ItemTooltipRenderEvent e = EventFactory.onItemTooltipRenderPre(poseStack, itemStack, mouseX, mouseY);
         if (e.isCanceled()) return;
 
-        instance.renderTooltip(
-                poseStack,
-                instance.getTooltipFromItem(e.getItemStack()),
-                e.getItemStack().getTooltipImage(),
-                e.getMouseX(),
-                e.getMouseY());
+        var tooltipsFromItem = instance.getTooltipFromItem(e.getItemStack());
```
P0keDev: I think it's best to explicitly declare the type of a variable

P0keDev (commented):
have you tested how this interacts with tooltips that get scaled by the scaling feature? i suspect this implementation would mess up the height calculation, so it might be best to include this logic within the feature, if possible. in general, it's best to keep logic outside of the mixin itself as much as possible, and that rule isn't really being followed here

TomaSajt:
I have a feeling it will be hard to make a robust system that wraps text AND scales down text. My idea would be to first scale by universal then wrap and then shrink. This will make it so it will always be completely on-screen, however the tooltip will not reach the side of the screen, like it is supposed to. If anyone has some ideas please share them.

kristofbolyai:
> I have a feeling it will be hard to make a robust system that wraps text AND scales down text. My idea would be to first scale by universal then wrap and then shrink. This will make it so it will always be completely on-screen, however the tooltip will not reach the side of the screen, like it is supposed to. If anyone has some ideas please share them.

This seems like the most sensible solution. But, if we wrap the tooltip width if needed, then decrease height if needed, how come we won't reach the side of the screen? And if we don't is that really a problem (if the tooltip is readable otherwise)?

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -177,4 +178,29 @@ public static List<Component> stripDuplicateBlank(List<Component> lore) {
 
         return newLore;
     }
+
+    public static int getOptimalTooltipWidth(Screen screen, List<FormattedText> tooltips, int mouseX) {
+        int width = tooltips.stream()
+                .mapToInt(McUtils.mc().font::width)
+                .max()
+                .orElse(0);
```
kristofbolyai: Why do we need this?

kristofbolyai: I am mostly referring to the `orElse` part. Github doesn't highlight it specifically.

magicus:
> I have a feeling it will be hard to make a robust system that wraps text AND scales down text.

Just use trusty old Knuth. https://github.com/jaroslov/knuth-plass-thoughts/blob/master/plass.md Don't you kids learn *anything* in school these days? 😜 

But more seriously, of course it's a tradeoff if we allow both wrapping and scaling, whereby the end result (all text fits within the screen) can be achieved in different ways. So we need to have some kind of priority ("scale first, and only wrap if we scaled 50% and it did not help", or "wrap first until we reach a text box with the same aspect ratio as the screen, and then scale down", or something else).

In general, I think scaling is a bad solution. On my old computer, some tooltips where hard to read since they got scaled down too much. In legacy we actually had an option to scroll the tooltips with the mouse wheel. That was really my favorite, but unfortunately the implementation was buggy and did not always work, so I had to leave it. We could perhaps revive the idea in Artemis, and implement it properly.

Also, as wrapping goes, most of the time tooltips where scaled down for me was not due to too long lines, but too many lines. Especially a lengthy lore could easily extend a tooltip past my screens limited height. So we might perhaps also consider "unwrapping" lines, turning a 8-line lore into 2, but longer, lines, to avoid scaling.

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
```diff
@@ -77,15 +84,14 @@ private void renderTooltipPre(
...
+                poseStack, itemStack, Collections.unmodifiableList(tooltips), mouseX, mouseY);
         if (e.isCanceled()) return;
-
-        instance.renderTooltip(
-                poseStack,
-                instance.getTooltipFromItem(e.getItemStack()),
-                e.getItemStack().getTooltipImage(),
-                e.getMouseX(),
-                e.getMouseY());
+        List<ClientTooltipComponent> clientTooltips = e.getTooltips().stream()
```
magicus: I'm not happy with this. I never liked the `@Redirect` in the first place (P0ke might remember my opposition ;-)). And now we're replacing vanilla code with something that is not even pretending to do the same as the original code. 

I don't have any alternative design suggestions, and I can't say that this will definitely break. But the more we mess around with vanilla code, the more we risk:

1) incompatibilities with future minecraft versions
2) incompatibilities with other mods
3) crashing the game if our code misbehaves (or has unintended interactions with other mods, or pieces of vanilla we did not consider), which
4) can't be turned off by disabling a feature.

Maybe there's no other way. But we need to be very, very sure about that, and also be sure the benefit outweighs the risks.

kristofbolyai (commented):
LGTM. Obviously, we still have the redirect, not sure if we can do anything about it. 

No approval yet since I didn't have time to test it.

TomaSajt:
I ended up doing the scale->wrap->fit approach, and it now works as expected. It's not perfect, but if the scale stays below 5 or so, the extra space on the side won't be very noticeable.
Also I removed most of the bloat from the mixin, so compatibility won't really be an issue.

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
kristofbolyai: Actually if we turn off the feature we kinda call the same method with the same parameters, so it *should* be fine. Obviously, we don't want many redirects like this at all, I agree with that.

kristofbolyai:
@TomaSajt Can you run  `git commit -m "Chore: Empty commit to rerun checks" --allow-empty` and `git push`? The bot is broken for first time contributors and build status gets stuck.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
TomaSajt: `intSteam::max` returns `Optional<int>` which can be empty when there are no elements, 0 is the default width if empty

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
```diff
@@ -52,8 +55,19 @@ public void setMouseY(int mouseY) {
 
     @Cancelable
     public static class Pre extends ItemTooltipRenderEvent {
-        public Pre(PoseStack poseStack, ItemStack itemStack, int mouseX, int mouseY) {
+        private List<Component> tooltips;
```
P0keDev, line 58: it seems a bit strange to define this specific to the Pre event, but ultimately i suppose it doesn't really matter. if you really wanted to be pedantic, you could make all the variables/getters universal and define the setters in Pre

P0keDev (dismissed):
overall, looks good to me!

kristofbolyai:
> @TomaSajt Can you run `git commit -m "Chore: Empty commit to rerun checks" --allow-empty` and `git push`? The bot is broken for first time contributors and build status gets stuck.

@TomaSajt can you do this again?

kristofbolyai (dismissed):

kristofbolyai:
Thank you for adding this!

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -15,10 +15,11 @@
 import net.minecraft.network.chat.MutableComponent;
 import net.minecraft.network.chat.Style;
 import net.minecraft.network.chat.TextColor;
+import net.minecraft.network.chat.TextComponent;
 import org.jetbrains.annotations.Nullable;
 
 public class ComponentUtils {
-    public static String getFormatted(Component component) {
+    public static String getFormatted(FormattedText component) {
```
magicus: What's this change? I can't see this fit on, also what is a FormattedText and how does it relate to Component?

TomaSajt: the Component interface extends the FormattedText interface.

This was a change that I was not really fond of as well, as I did not write this original functionality. I'm not really sure how it works but as far I know the method is supposed to return the §-color-coded string from a Component. I needed a way to create a Component from a FormattedText, and found that this method doesn't use Component-only behavior, so I changed the parameter type to FormattedText instead.
I used this in the formattedTextToComponent method which goes ahead and converts every FormattedText supplied from splitText back to Component by calling this getFormatted function.
This was the change that allowed me to feed back a List\<Component\> in the ItemTooltipRenderEvent.Pre instead of the vanilla internal-wizardry I was doing.
I maybe should've also changed the name of the parameter. (should I?)

(also, If you know of a better way to go from FormattedText to Component, I'll gladly use that)

kristofbolyai: If you change the type, please change the argument name too. Also, I think it is a reasonable change.

kristofbolyai: But I will wait for magicus' comment before merging.

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -19,10 +19,10 @@
 import org.jetbrains.annotations.Nullable;
 
 public class ComponentUtils {
-    public static String getFormatted(FormattedText component) {
+    public static String getFormatted(FormattedText formattedText) {
```
TomaSajt: These method names are becoming a bit confusing with all my `FormattedText` usage. I might go with `getStringFormatted` and `getStringUnformatted`.

common/src/main/java/com/wynntils/features/user/TooltipScaleFeature.java
```diff
@@ -10,10 +10,13 @@
...
+import com.wynntils.mc.utils.ComponentUtils;
 import com.wynntils.mc.utils.McUtils;
 import com.wynntils.wc.utils.WynnUtils;
+import java.util.Collections;
+import java.util.List;
 import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.network.chat.Component;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @FeatureInfo(stability = Stability.STABLE, category = "Item Tooltips")
```
magicus, line 22: I just realized the name of the class, `TooltipScaleFeature`, is not as spot-on anymore, now that we do both scaling and wrapping. I don't have any good suggestion (`TooltipFittingFeature`???), but if you figure out a good name that captures both scaling and wrapping, please feel free to rename the class. :) (If you do, remember to update the i18n resources.)

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
magicus: I'm not even sure why we have a `Pre` event. We're only intercepting tooltip rendering once, yes? I only think we should need to bother with .Pre and .Post events if we *actually* are sending two different events. And I'm always willing to refactor, if it turns out that what was one event need to turn into two.

magicus: Also, `onItemTooltipRenderPre` should be just `onItemTooltipRender`.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
magicus: I was thinking, we really need a clear name -- and possibly a separate type -- for "string with format codes", like "§eHello". Calling it "formatted" is clearly bad, since it gets confused with Vanilla's type system. Perhaps "coded string"? Or "fmtcode string"? 

This is really a different issue from this PR, but it aggravates the problem and makes it more urgent to solve. But I've been thinking a bit about this for some time before. I think we should consider introducing a separate type for CodedString (of whatever we end calling it), which basically just has a String as a field. But the point is that we can use Java's type system to make sure we handle strings properly. Then we can convert from a Component to a CodedString, or we can strip the control formatting from the CodedString and just return a bare String, etc. But at all points it will be clear if we're dealing with a CodedString or a Component. (And we should never really deal with a bare String, unless we're consuming it.)

TomaSajt: I noticed that in the class `Character` `getUnformatted` is used, but still checks for §-codes.
*snippet from Character:*
```java
ComponentUtils.getUnformatted(e.getTitle()).equals("§8§lSelect a Class"))
```
This is probably because Wynncraft uses §-coded strings for these container names. So this is not actually "unformatted" or "raw" or whatever we're gonna call it. I think we should do some actual string replacement to get rid of the §-codes in `getUnformatted` instead of just hoping that FormattedString::getString (or Component::getString) doesn't give us §-coded text.
*(Added a TODO to getUnformatted)*

common/src/main/java/com/wynntils/features/user/TooltipScaleFeature.java
TomaSajt: I ended up using TooltipFittingFeature, I think it's the best term to summarize what it's doing.

kristofbolyai (dismissed):
Looks fine IMO. I can change that event separately to match our format if the team wants me to, but it's fine for now.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
magicus: Whatever we need to do likely merits a separate PR, and perhaps some discussions before that. So let's do this change for now, and then we can come back and revisit it when we've decided how to best handle this.

TomaSajt:
I realized the way I'm using getCodedString is leading to data-loss, as it isn't able to handle non-§-coded colors properly. This means all the custom coloring will disappear.
image
*(the lerp'd colors don't have an actual §-color)*

---
This probably will mean I have to find another way to go from FormattedText to Component without converting to String.
I'll convert this PR to a draft until I find a way to do that.

TomaSajt:
Well it seems like I found a pretty straightforward way of doing it, I thought it was going to take more time.

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
magicus: I see now that we had the `Pre` subclass before this PR, so I'll accept that it stays for now. But I think it should go in the future. To quote the wise William of Occam: "Do not multiply Event classes beyond necessity". Let's keep the set of events as tight and specialized as needed just for the momdent, but instead have no qualms about introducing new events as soon as as the need arise.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -40,24 +41,25 @@ public Optional<Object> accept(Style style, String string) {
...
         if (component == null) return null;
 
-        return ComponentUtils.getFormatted(component);
+        return getCodedString(component);
     }
 
     @Nullable
-    public static String getUnformatted(String loreString) {
-        MutableComponent component = Component.Serializer.fromJson(loreString);
+    public static String getUnformattedString(String jsonString) {
```
magicus: I'm seriously lost at this. It all seems just a mess to me. It definitely breaks the "don't make me think" rule. I can accept it for now, if it is needed. But we definitely need to find a better way to handle strings to make it unambiguous.

magicus: ("this" meaning all these changes to this file)

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -177,4 +179,43 @@ public static List<Component> stripDuplicateBlank(List<Component> lore) {
 
         return newLore;
     }
+
+    public static Component formattedTextToComponent(FormattedText formattedText) {
```
magicus, line 182: If I get this straight, with this new FormattedText -> Component, all the other formatted string changes above is not really needed? If so, can you revert them?

TomaSajt: OK, I'll revert

TomaSajt:
I reverted most of the changes unrelated to this PR except a few parameter names in `ComponentUtils` which I think are more suitable.

magicus (approved):
Looks good now! Thanks for your contribution, and your patience. :)

kristofbolyai:
@TomaSajt Hey, I just logged in to test the latest PRs, and I found an issue with this PR. 
Item lore is missing newlines:

Without this PR:
image
(as it should be)

With this PR:
image


I assume these were caused by the `ComponentUtil` changes. Do you mind fixing this or should I open an issue for it?

PR #161, opened 2022-07-14
Screens: Add Overlay Management GUI

kristofbolyai:
Added overlay management GUI! This PR actually contains two GUIs, one for selecting an Overlay to be edited (`OverlaySelectionScreen`), and the second one for actual editing (`OverlayManagementScreen`). Open the GUI with pressing the "Wynntils Menu" in the game pause menu.

Quite a big PR, sorry reviewers. I encourage you to also try this in-game, just to see how it all behaves.

Rough feature list:
- [x] Resizing overlay
  - [x] With corners
  - [x] With edges 
- [x] Moving overlay
- [x] Disable overlay
- [x] Help tooltips
- [x] Draw overlay name
- [x] Hide disabled overlays option
- [x] A way to set alignment overrides (arrows)
- [x] Test mode -> 
PreviewRender method in overlays (this gets drawn when management screen is open instead of actual render, making it possible to render a static preview for overlays depending on Wynncraft itself)
- [x] Show ninths/sections in non test mode.
- [x] Show render alignment with overlay name
- [x] Individual screen to disable overlays and select overlay for editing
- [x] Alignment lines (snapping)
  - [x] At specific screen parts
  - [x] Depending on other overlays

Here is a few screenshots of the GUIs:
image
image
image
image
image
image
image
image
image
image
image


common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -83,6 +85,55 @@ public String toString() {
...
+
+        // 3. Calculate render positions for new alignment
+        OverlayPosition newOverlayPositionTemp =
+                new OverlayPosition(0, 0, verticalAlignment, horizontalAlignment, section);
+
+        float renderX = Overlay.getRenderX(newOverlayPositionTemp, overlay);
+        float renderY = Overlay.getRenderY(newOverlayPositionTemp, overlay);
+
+        // 4. Calculate the offsets to match the current render position, but factor in drag
+
```
kristofbolyai: todo update this comment

common/src/main/java/com/wynntils/screens/lists/entries/OverlayEntry.java
```diff
@@ -0,0 +1,133 @@
...
+        FontRenderer.getInstance()
+                .renderText(
+                        poseStack,
+                        translatedName,
+                        3,
+                        (OverlayList.getItemHeight()
+                                        - FontRenderer.getInstance()
+                                                        .calculateRenderHeight(List.of(translatedName), width)
+                                                / 2f)
+                                / 2f,
```
kristofbolyai: spotless epic fail

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -83,7 +82,11 @@ private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState re
                 }
             }
 
-            overlay.render(event.getPoseStack(), event.getPartialTicks(), event.getWindow());
+            if (testMode) {
+                overlay.renderPreview(event.getPoseStack(), event.getPartialTicks(), event.getWindow());
+            } else {
```
kristofbolyai, line 89: todo fix if

Incompleteusern (dismissed):
Finally got around to looking at this, LGTM

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -67,6 +70,8 @@ public Overlay(
 
     public abstract void render(PoseStack poseStack, float partialTicks, Window window);
 
+    public abstract void renderPreview(PoseStack poseStack, float partialTicks, Window window);
```
magicus: I think it's a bit sad that we need to have this for all Overlays. :( Can we not just make a default implementation that just shows the overlay title? Otherwise it increases the burden to implement an Overlay. Or, can we not use the *real* render function for previews?

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
```diff
@@ -74,7 +73,7 @@ private static void takeScreenshot(Slot hoveredSlot) {
         List<Component> tooltip = stack.getTooltipLines(null, TooltipFlag.Default.NORMAL);
         WynnItemUtils.removeLoreTooltipLines(tooltip);
 
-        Font font = McUtils.mc().font;
+        Font font = FontRenderer.getInstance().getFont();
```
magicus, line 76: What's the difference?

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
```diff
@@ -106,13 +105,7 @@ private static void takeScreenshot(Slot hoveredSlot) {
...
         poseStack.pushPose();
         poseStack.scale(scalew, scaleh, 1);
-        RenderUtils.drawTooltip(
-                tooltip.stream()
-                        .map(Component::getVisualOrderText)
-                        .map(ClientTooltipComponent::create)
-                        .collect(Collectors.toList()),
-                poseStack,
-                font);
+        RenderUtils.drawTooltip(poseStack, tooltip, font, true);
```
magicus, line 108: Was this an unrelated change?

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
```diff
@@ -96,5 +122,21 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
...
+        @Override
+        public void renderPreview(PoseStack poseStack, float partialTicks, Window window) {
+            recalculatePreviewRenderTasks(); // we have to force update every time
+
+            FontRenderer.getInstance()
+                    .renderTextsWithAlignment(
+                            poseStack,
+                            this.getRenderX(),
+                            this.getRenderY(),
+                            Arrays.stream(toRenderPreview).toList(),
```
magicus: Why is toRenderPreview not a List to begin with..?  (The same question goes for toRender, I assume)

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -365,17 +367,12 @@ public static void drawArc(PoseStack poseStack, CustomColor color, int x, int y,
                 Texture.ARC.height());
     }
 
-    public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack poseStack, Font font) {
-        int tooltipWidth = 0;
-        int tooltipHeight = lines.size() == 1 ? -2 : 0;
+    public static void drawTooltip(
```
magicus, line 370: What's the deal with the tooltip changes? Is that a separate fix that crept in, or do you need it for the overlay manager?

kristofbolyai: It is used in the overlay manager to render help tooltips like this. 
image

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
kristofbolyai: Change of method signature.

kristofbolyai: Behaves the same way, but the method is more generalized.

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
kristofbolyai: Nothing at the moment, but I guess it is nice to keep this universal in case we change something up in the future.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+                    minY = overlay.getRenderY();
+                    maxY = overlay.getRenderY() + overlay.getHeight();
+                }
+                case Bottom -> {
+                    minX = overlay.getRenderX();
+                    maxX = overlay.getRenderX() + overlay.getWidth();
+                    minY = overlay.getRenderY() + overlay.getHeight() - MAX_CLICK_DISTANCE / 2f;
+                    maxY = overlay.getRenderY() + overlay.getHeight() + MAX_CLICK_DISTANCE / 2f;
+                }
+                default -> {
```
magicus, line 276: Is Edge an enum? I thought the switch expression barfed on `default` if all enum values were covered..?

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: If you agree with me making renderPreview just call render, and still have it to be overridden? Let's say the quest info overlay or the game update overlay needs a separate preview, and something like the health bar overlay doesn't.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: This prevents variable might not have been initialized.

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
kristofbolyai: I guess I can change that up.

common/src/main/java/com/wynntils/screens/OverlaySelectionScreen.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    protected void init() {
+        overlayList = new OverlayList(this);
+    }
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        poseStack.pushPose();
+
+        this.fillGradient(poseStack, 0, 0, this.width, this.height, -1072689136, -804253680);
```
magicus: What are these magic numbers? If it's colors, can you make them constants and perhaps express them as RGB codes instead?

common/src/main/java/com/wynntils/screens/OverlaySelectionScreen.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public boolean mouseReleased(double mouseX, double mouseY, int button) {
+        return overlayList.mouseReleased(mouseX, mouseY, button);
+    }
+
+    @Override
+    public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
+        if (keyCode == 256 && this.shouldCloseOnEsc()) {
```
magicus: Is there some pre-defined constant you can use instead of 256? Sounds like it's ESC...

common/src/main/java/com/wynntils/screens/OverlaySelectionScreen.java
kristofbolyai: I took this from the MC source code, I am not sure about those colors. It is used to render the gray background when a GUI is open. I will check this out.

common/src/main/java/com/wynntils/screens/lists/entries/OverlayEntry.java
magicus: I agree that spotless handles this anti-gracefullly. But, I think you can also see it as a sign that the function call is too complex. Try extracing the height calculation into a temp variable first. By giving it a name, it will also be clearer what this is supposed to do. A win-win! :)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -153,7 +153,9 @@
   "feature.wynntils.playerGhostTransparency.name": "Transparent Player Ghost Feature",
   "feature.wynntils.playerGhostTransparency.playerGhostTranslucenceLevel.description": "The level of translucence that should be applied to player ghosts",
   "feature.wynntils.playerGhostTransparency.playerGhostTranslucenceLevel.name": "Player Ghost Translucence",
-  "feature.wynntils.questInfoOverlayFeature.name": "Quest Status Overlay Feature",
+  "feature.wynntils.questInfoOverlay.name": "Quest Status Overlay Feature",
+  "feature.wynntils.questInfoOverlay.overlay.questInfo.description": "Display the currently tracked quest's info",
+  "feature.wynntils.questInfoOverlay.overlay.questInfo.name": "Quest Info Tracker Overlay",
```
magicus: Why did not `overlay.wynntils.questInfo.name` work? I must admit I'm a bit lost here, but I must say that the new name is quite a mouthful. And also, is this an unrelated fix or something that is needed by the overlay manager?

kristofbolyai: I am pretty sure you wanted this i18n key format for the overlay keys. `  "feature.wynntils.questInfoOverlayFeature.name": "Quest Status Overlay Feature",` is incorrect because it contained the "Feature" part of the feature, which should be left out.

magicus:
I'm very impressed that you managed to get this together! The screenshots look very promising. But it's a lot of code to go through. I've made one pass but I think I need to do another when I'm not about to fall asleep. =)

Also, I think your suggestion about actually trying it out is good. But that also means I need even more time to be able to test this PR properly.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
magicus: Yes, that sounds better. I can see how you are worried that some overlays may not be visible, and so you want to feed them dummy data. But it also feels a bit odd to make it a separate render function, when it really is like a "get your data from somewhere else". I can't see how we can implement that in a nice way, though.

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
magicus: Maybe we should have a McUtils.getFont() instead?

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
magicus: And the old implementation was too limited for that?

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
kristofbolyai: We could but I don't see any benefit atm. I imagine `McUtils.getFont()` could be the default Font for Minecraft, while our FontRenderer could have a Font with specific settings. We will see how this turns out in the future.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
magicus: As a warning? That sounds really odd. How on earth could we get there with `value` uninitialized? Is that the IntelliJ compiler?

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
kristofbolyai: With this change we can pass regular components for a tooltip, not the really specific tooltip ones, and this render call will just convert it if needed. The boolean was added so there is a difference between normal tooltips and ones with a "header" (where there should be more space after the first line).

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: So I got only myself to blame for this? :-D

This might have gotten worse since QuestInfoOverlayFeature had "Overlay" in it's name. Maybe we can rename it just "QuestInfoFeature" later on.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: image

kristofbolyai:
This should be everything, Waiting for your next review round.

common/src/main/java/com/wynntils/screens/OverlaySelectionScreen.java
```diff
@@ -32,7 +32,8 @@ protected void init() {
     public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
         poseStack.pushPose();
 
-        this.fillGradient(poseStack, 0, 0, this.width, this.height, -1072689136, -804253680);
+        int backgroundColor = CommonColors.DARK_GRAY.withAlpha(200).asInt();
+        this.fillGradient(poseStack, 0, 0, this.width, this.height, backgroundColor, backgroundColor);
```
magicus, line 51: Is this correct? Now the gradient goes from backgroundColor to backgroundColor? Previously it was two different colors...

kristofbolyai: Those two colors were basically the same, just some encoding stuff (probably happened while obfuscation). This does the same thing.

kristofbolyai:
Just noticed an issue with this PR! TODO for feature me: vertical and horizontal offsets in OverlayPosition should be a scale value (where the scale value is calculated depending on the anchor section size) and not pixels, because that won't work with resizing. More systematic changes to Overlays, although I think this will be pretty straight forward :\).

kristofbolyai:
(This only affects `OverlayPosition` and `OverlayPosition#getBestPositionFor`, feel free to review anything else.)

magicus:
I have now tested the PR locally. Overall it is very impressive and intuitive! 

I ran into half a dozen of of snags. I'm listing them here, none of them are blockers for integrating this PR, and can be fixed in a follow-up if you want.

First problem (we can fix this separately). I got a crash, or rather, an infinite loop, in `calculateRenderHeight`, when I tried resizing the tracked quest window to be very tall and narrow.

Second issue: the overlay list screen needs a close button. Now I had to reach for the keyboard to press esc to finish manipulating the overlays, which are all done by mouse.

Third issue: The "Test" button functionality is sound, but the UI is not really clear. I think it would be better to replace the Test button with a "Show dummy content", and when you press it, it gets replaced with a new button "Show outlines", so you toggle between the "dummy content" and "outline" modes.

Fourth issue: I assume that pressing "Close" instead of "Apply" will lose changes (I realize as I write this that I never tested this). Maybe we should also have a "Reset" or "Restore" button that restore the setting to how it was before we started changing the overlay, but does not leave the screen. Maybe also a "Default" which gets a Overlay back to the default coded in the Feature. 

Fifth issue: Adding all these buttons I'm requesting might make the UI take up quite a bit of screen real estate, and can possibly interfere with placement of overlays. I don't know if this is a real problem, nor really what to do about it. Maybe in the future we can have the graphics team make us a nice texture with drawn icons for "save" "apply" "reset" and "default", and a tooltip hover to explain the functionality if it is not apparent. That way we can make it much more compact.

Sixth issue: The snapping functionality was really nice. However, when I tried snapping the Tracked Quest info to the top right corner (and right-align it), it did not look good. The text went all the way out to the border of the screen, so the ":" were almost invisible. I think we should have a pixel margin to the edge of the screen, so that snapping to the edge of the screen is not really really the edge. Or, perhaps better, have yet another "snap line" just outside the screen border, so you can chose if you *really* want to snap to the edge (might fit for some overlays) or if you want the margin.

magicus:
More comments:

* When selecting vertically centered text alignment, the name of the overlay is not centered, but is put below the center.
* The green color for "inactive" overlays is attracting the eye too much. Perhaps make them gray instead to show that we can't move them. (And the active overlay could in fact be that green rather than cyan)
* We don't need to have "Overlay" in the name of overlays on the management screen. Not only is it redundant, on my resolution at least, it also turns `Health Bar` into
```
Health Bar
Overlay
```

magicus:
* It would be nice if pressing shift made you ignore snapping behavior. (I tried this instinctively, so I believe this is quite common).
* The overlay selection list has the name of the overlays not really at the center of the "button".
* The red color of disabled overlays in the selection list is shining quite aggressively. Perhaps here too use a muted gray instead?

common/src/main/java/com/wynntils/core/features/overlays/Edge.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.wynntils.utils.objects.Pair;
+import java.util.Map;
+import net.minecraft.world.phys.Vec2;
+
+public enum Edge {
+    Top(Corner.TopLeft, Corner.TopRight),
+    Left(Corner.TopLeft, Corner.BottomLeft),
+    Right(Corner.TopRight, Corner.BottomRight),
+    Bottom(Corner.BottomLeft, Corner.BottomRight);
```
magicus, line 14: I'm not sure I should even write this, since it doesn't matter too much, but the relationship of Edge and Corner feels kind of backwards. At least if you compare with how things are named in English. A "Top Left" corner sounds like it combines a "Top" and "Left" edge. That is, edge feels more atomic, and corners are defined given which edges meet, instead of having corners atomic and edges defined by which corners they combine.

Feel free to ignore this if you think it creates too much work to redesign this.

common/src/main/java/com/wynntils/core/features/overlays/BasicOverlay.java
```diff
@@ -12,29 +12,39 @@
 
 public class BasicOverlay extends Overlay {
     private final QuadConsumer<Overlay, PoseStack, Float, Window> renderConsumer;
+    private final QuadConsumer<Overlay, PoseStack, Float, Window> renderPreviewConsumer;
 
     public BasicOverlay(
```
magicus, line 17: Do we need both the float/float and the OverlaySize constructors? If anything, I'd rather see a simplified constructor without the renderPreviewConsumer argument (which uses the renderConsumer for both). But if we create that variant, and we want it in both constructors, we're up to four constructors which seems a bit too much...

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -109,16 +116,17 @@ public String getTranslatedName() {
 
     @Override
     public String getTranslation(String keySuffix) {
-        return I18n.get("feature.wynntils" + getDeclaringFeatureNameCamelCase() + ".overlay." + getNameCamelCase() + "."
-                + keySuffix);
+        return I18n.get("feature.wynntils." + getDeclaringFeatureNameCamelCase() + ".overlay." + getNameCamelCase()
```
magicus, line 118: How come we did not notice this bug? 🤔 All those bug fixes we see scare me a bit. We have no automatic testing, and apparently even our ad-hoc testing is not good enough...

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -146,25 +162,51 @@ public float getRenderedHeight() {
...
-        return switch (this.position.getVerticalAlignment()) {
-            case Top -> section.y1() + this.position.getVerticalOffset();
-            case Middle -> (int) (section.y1() + section.y2() - this.getHeight()) / 2
-                    + this.position.getVerticalOffset();
-            case Bottom -> (int) (section.y2() + this.position.getVerticalOffset() - this.getHeight());
+    public float getRenderY() {
+        return getRenderY(this.position, this);
+    }
+
+    public static float getRenderX(OverlayPosition position, Overlay overlay) {
```
magicus: I'm not sure why you made this static and take in an overlay. I think it can (and should) be an instance method. The one place you call it as a static method can just as well be called as an instance method on the given overlay.

(It is also a bit weird to have the same name to a static and instance method)

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -175,4 +217,15 @@ public HorizontalAlignment getRenderHorizontalAlignment() {
     public VerticalAlignment getRenderVerticalAlignment() {
         return verticalAlignmentOverride == null ? position.getVerticalAlignment() : verticalAlignmentOverride;
     }
+
+    public Map<Corner, Vec2> getCornersMap() {
```
magicus: If I get this right you put the four corners in a map, and then you extract two of them in Edge. This is the only use of this method, yes?

What if you make it into a switch expression `getCornerPos` that takes in a Corner and return the position for that corner?

magicus: Ah, I see now, you also iterate over it in the manager. But you could probably still have this a getter for a method, and iterate over all possible values of the Corner enum there.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -65,6 +58,14 @@ public static void onRenderPost(RenderEvent.Post event) {
     }
 
     private static void renderOverlays(RenderEvent event, OverlayInfo.RenderState renderState) {
+        boolean testMode = false;
+
+        if (McUtils.mc().screen instanceof OverlayManagementScreen screen) {
+            testMode = screen.isTestMode();
+        }
+
+        boolean shouldRender = !(McUtils.mc().screen instanceof OverlayManagementScreen);
```
magicus: The reason you're not just bailing out here is that we might need to do setCanceled for all overlays?

magicus: But you've already tested the instanceof above, so you can set shouldRender to true always and add a false assignment in the if body. That would make it more analogous to testMode.

common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
magicus: Ok.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -501,6 +498,52 @@ public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack pos
...
+        poseStack.translate(renderX, renderY, renderZ);
+        drawTooltip(poseStack, componentLines, font, firstLineHasPlusHeight);
+
+        poseStack.popPose();
+    }
+
+    public static int getToolTipWidth(List<ClientTooltipComponent> lines, Font font) {
+        int tooltipWidth = 0;
+
+        for (ClientTooltipComponent clientTooltipComponent : lines) {
```
magicus: This looks very old-skool. :) Is there no max() methods on streams?

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -501,6 +498,52 @@ public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack pos
...
+            }
+        }
+
+        return tooltipWidth;
+    }
+
+    public static int getToolTipHeight(List<ClientTooltipComponent> lines) {
+        int tooltipHeight = lines.size() == 1 ? -2 : 0;
+
+        for (ClientTooltipComponent clientTooltipComponent : lines) {
```
magicus: Same here, there ought to be a sum() method. (If there isn't, Java's stream library is too crippled :().

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        if (testMode) {
+            FontRenderer.getInstance()
+                    .renderText(
+                            poseStack,
+                            this.width / 2f,
+                            this.height - 160,
+                            new TextRenderTask(
```
magicus: Breaking out the render task to a local variable might help spotless not f*ck this up completely...

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+                                overlay.getRenderedHeight() + yOffset,
+                                overlay.getRenderHorizontalAlignment(),
+                                overlay.getRenderVerticalAlignment());
+            }
+
+            if (isMouseHoveringOverlay(selectedOverlay, mouseX, mouseY) && selectionMode == SelectionMode.None) {
+                RenderUtils.drawTooltipAt(
+                        poseStack,
+                        mouseX
+                                - RenderUtils.getToolTipWidth(
```
magicus: Same here, extract tooltipWidth to a local variable to help formatting (and readability).

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+                        false);
+            }
+        }
+
+        super.render(poseStack, mouseX, mouseY, partialTick); // This renders widgets
+    }
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (testMode) return super.mouseClicked(mouseX, mouseY, button);
```
magicus, line 221: Does this mean that in test mode you could be clicking around in the background UI? How well does that work? Sounds a bit scary.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
magicus: I had to go dig this down. It is by design, in http://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#13.4.26. The problem is that in theory, we can compile this class first, then add a new enum and just recompile the Edge class, and then this code would have an unaccounted-for path. So this is correct.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+
+            return false;
+        }
+
+        return super.mouseClicked(mouseX, mouseY, button);
+    }
+
+    @Override
+    public boolean mouseReleased(double mouseX, double mouseY, int button) {
+        if (testMode) return false;
```
magicus, line 301: So for pressed in testMode we pass it on to super, but not for release? Is this intentional? Can't we get "stuck" on some background UI element?

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+            if (keyCode == GLFW.GLFW_KEY_RIGHT) {
+                index += 1;
+            } else {
+                index -= 1;
+            }
+
+            HorizontalAlignment[] values = HorizontalAlignment.values();
+            index = (values.length + index) % values.length;
+            selectedOverlay.setHorizontalAlignmentOverride(values[index]);
+            saveConfigForOverlay();
```
magicus: Does this gets saved even if we do not press Apply?

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+    @Override
+    public boolean mouseDragged(double mouseX, double mouseY, int button, double dragX, double dragY) {
+        if (testMode) return false;
+
+        // Order:
+        //  - Corners
+        //  - Edges
+        //  - OverlayArea
+
+        handleOverlayCornerDrag(button, dragX, dragY);
```
magicus: Maybe a switch on `selectionMode` instead of just calling all three blindly? More for a readability thing, I realize the code works as well when written like this.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -0,0 +1,651 @@
...
+
+        for (int i = 2; i <= ALIGNMENT_LINES_MAX_SECTIONS_PER_AXIS; i++) {
+            for (int j = 1; j < i; j++) {
+                verticalAlignmentLinePositions.add((float) j / i);
+                horizontalAlignmentLinePositions.add((float) j / i);
+            }
+        }
+
+        for (Overlay overlay :
+                OverlayManager.getOverlays().stream().filter(Overlay::isEnabled).collect(Collectors.toSet())) {
```
magicus: Not that it matters greatly, but a simple `toList()` would be shorter, and in practice it's not relevant that a set would have sufficed. (In practice, Lists are often used in Java even when order is ignored.)

common/src/main/java/com/wynntils/screens/OverlaySelectionScreen.java
```diff
@@ -0,0 +1,108 @@
...
+                0,
+                Texture.OVERLAY_SELECTION_GUI.width(),
+                Texture.OVERLAY_SELECTION_GUI.height(),
+                Texture.OVERLAY_SELECTION_GUI.width(),
+                Texture.OVERLAY_SELECTION_GUI.height());
+
+        FontRenderer.getInstance()
+                .renderText(
+                        poseStack,
+                        "Overlays",
```
magicus: The string literal should be i18n'd.

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
```diff
@@ -0,0 +1,99 @@
...
+import com.wynntils.screens.OverlaySelectionScreen;
+import com.wynntils.screens.lists.entries.OverlayEntry;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import net.minecraft.client.gui.components.ContainerObjectSelectionList;
+
+public class OverlayList extends ContainerObjectSelectionList<OverlayEntry> {
+    private static final int ITEM_HEIGHT = 25;
+    private final Set<Overlay> overlays;
```
magicus: I don't really like how the list presented to the user is based on an (unordered) set. We should sort the set according to some principle (overlay name alphabetically, if nothing else).

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    protected int getRowTop(int index) {
+        return this.y0 - (int) this.getScrollAmount() + index * this.itemHeight + this.headerHeight + 1;
+    }
+
+    @Override
+    public int getRowWidth() {
+        return 161;
```
magicus: Did I say how much I loved the typescript warning that said "No magic numbers in code!"? :-) Please make this a static final instead.

magicus:
I've finished my second pass now. I think I got most issues covered, but I think I'll need to look through everything closely again a final time when you're done handling my current batch of change requests...

magicus:
But I must emphasize that this is looking extremely professional and cool! I can't imagine how much time you must have spent trying to get this to work so smoothly, and how to handle all the twisting logic of inverting alignments and flipping coordinates for different corners. (That always make my head hurt; I've just glossed through that in the review, hoping that if it was incorrect it would show up quickly in the testing...)

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
Incompleteusern: I think it does save position if you move things around.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: To be fair this was a written when translations where not used anywhere for overlays so we couldn't see wrong behavior even with testing.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: The static method has it is use case and cannot be replaced by the instance method, but I can rename it for clarity.

kristofbolyai: Actually, I cannot come up with a better name, but the use case is in OverlayPosition line 124-125.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: Not the background GUI, but the Screen code, which handles button clicks. If we don't have this, we can't interact with the buttons.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: Explained why mousePressed is important above, this has no use case when in test mode.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
magicus: I need to verify this. If you just press close, and not apply, it should neither save a new placement or size of the overlay, nor the text alignment inside the overlay.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
magicus: Why not:

```
        float renderX = overlay.getRenderX(newOverlayPositionTemp);
        float renderY = overlay.getRenderY(newOverlayPositionTemp);
```

or am I missing something?

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -101,18 +101,14 @@ public void onClose() {
...
-                            this.height - 160,
-                            new TextRenderTask(
-                                    "Test mode on.",
-                                    new TextRenderSetting(
-                                            0,
-                                            CommonColors.WHITE,
-                                            FontRenderer.TextAlignment.CENTER_ALIGNED,
-                                            FontRenderer.TextShadow.NORMAL)));
+            TextRenderTask renderTask = new TextRenderTask(
+                    "Test mode on.",
```
magicus: This should be i18n'd. (Also, drop the trailing period).

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: I just fixed this. Thanks for pointing out.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: Oh my bad. Fixed.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -293,7 +294,10 @@ public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
             VerticalAlignment[] values = VerticalAlignment.values();
             index = (values.length + index) % values.length;
             selectedOverlay.setVerticalAlignmentOverride(values[index]);
-            saveConfigForOverlay();
+            selectedOverlay.getVisibleConfigOptions().stream()
+                    .filter(configHolder -> Objects.equals(configHolder.getFieldName(), "verticalAlignmentOverride"))
```
Incompleteusern: Can't `Overlay#getConfigOptionFromString` be used?

kristofbolyai:
> Sixth issue: The snapping functionality was really nice. However, when I tried snapping the Tracked Quest info to the top right corner (and right-align it), it did not look good. The text went all the way out to the border of the screen, so the ":" were almost invisible. I think we should have a pixel margin to the edge of the screen, so that snapping to the edge of the screen is not really really the edge. Or, perhaps better, have yet another "snap line" just outside the screen border, so you can chose if you _really_ want to snap to the edge (might fit for some overlays) or if you want the margin.


image
I think this feels alright, I have a bit of "this is what you have asked for" vibe. Fix for this is more specialized snapping line positions.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
kristofbolyai: Fixed.

kristofbolyai:
> First problem (we can fix this separately). I got a crash, or rather, an infinite loop, in `calculateRenderHeight`, when I tried resizing the tracked quest window to be very tall and narrow.

Already fixed in #172

kristofbolyai:
As for the button changes, I will leave that up to someone with better GUI design skills :\).

kristofbolyai:
Fixed every issue, and added a few improvements too. The blocking issue, wrong position calculation should be sorted out now. 

Waiting for the reviews. @magicus if you are reading this, your attention was also requested in the staff discord :^).

magicus:
New comments after playing around with the latest version:

* I still find it a bit hard to locate the overlay that you have selected to edit. Maybe we can make it pulsate a bit? Like each time you render it, you modify the color slightly, to make it lighter and then darker again. I think this can be implemented fairly easy, and it will probably make it much easier to spot.
* I once again hanged calculateRenderHeight. This time, I accidentally tried to resize the overlay to a "negative" width, i.e. I dragged the left edge too far so it passed the right edge. We should probably have some minimum size for overlays that you can't resize below.
* The title "Overlays" on the overlay list is very crowded. Is it possible to give it some more space, and maybe center it horizontally on the "page"?
* Resizing in the horizontal direction is slow and the mouse movement feels scaled down. Edit: this might be some macOS retina 4x thing; it seems to work different in full screen and windowed mode. Probably needs more testing.
* It should be possible to exit the overlay placement screen with esc (== close) (and possibly also with enter (== apply))
* There are Z order rendering issues with tooltips and the overlay list. See screenshot.

TL;DR repetition of some of my older comments. You don't have to fix them but please acknowledge that you've seen my remarks. :-)

* Center-aligned text is not placed properly, see screenshot.
* Overlay list needs close button, not only esc.
* Please remove "Overlay" from the end of the name of overlays when showing them in the placement screen
* Name over the overlay in the green/gray rectangle in the overlay list is not centered vertically, but slightly further down than the center. (See screenshot)
* It would be nice if pressing shift made you ignore snapping behavior.

As always, I'm open to fixing some or all of these issues as follow-ups.

kristofbolyai:
Oops! I forgot to do the Shift to disable snapping! As for the centering for rendering, I have tried to debug it, with no luck (see the fix and reverse commit😅).

kristofbolyai:
Just to keep track, the center alignment issues are duplicate likely, fixing one properly will fix the other one. 

As for remove "overlay" how would I do that without breaking proper localization?

kristofbolyai:
> The title "Overlays" on the overlay list is very crowded. Is it possible to give it some more space, and maybe center it horizontally on the "page"?

Placement is intentional. If you check the texture very closely, there is space left out for a search box, which will be a future addition. Not sure how I could fix this otherwise.

kristofbolyai:
> I once again hanged calculateRenderHeight. This time, I accidentally tried to resize the overlay to a "negative" width, i.e. I dragged the left edge too far so it passed the right edge. We should probably have some minimum size for overlays that you can't resize below.

As mentioned earlier, the method should be fixed in the other PR of mine. As for the minimum size, I would rather work on that when the game update overlay is merged, so I can also test the patched calcRenderHeight. Feel free to make an issue about this, if not, I will in a few days.

kristofbolyai:
> There are Z order rendering issues with tooltips and the overlay list. See screenshot.

Saw this, it is harder to fix this than just Z value. I will try to fix it, if not, issue it is.

kristofbolyai:
> Resizing in the horizontal direction is slow and the mouse movement feels scaled down

Never experienced this, I will try to reproduce.

kristofbolyai:
> It should be possible to exit the overlay placement screen with esc (== close) (and possiblt also with enter (== apply))

👍

kristofbolyai:
This should be everything. Let me know if I missed something.

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -213,6 +230,15 @@ private float calculateRenderHeight(List<TextRenderTask> toRender) {
         return (float) (height - NEWLINE_OFFSET / 2 * McUtils.window().getGuiScale());
```
magicus: This looks sus. I think this is the reason the vertically centered text gets incorrectly offset. Why would you subtract half the newline offset multiplied with the gui scale? I think some parentheses are missing.

magicus: Also, the for loop (github won't let me comment on those lines) need some failsafe, so it does not just stack up `lines` and reset i to parts.size() if it fails to get a sublist. (I think that is what is causing the infinite looping.)

kristofbolyai: https://github.com/Wynntils/Artemis/pull/172/files#diff-4b2bbac5a5d922ca9738a301beb33915427e526a4cf0e22f97ec98cebbc46c5d for the second comment.

magicus:
> As for remove "overlay" how would I do that without breaking proper localization?

Yeah, that... Hm. Maybe we should drop the "Overlay" from the English i18n strings then? Seeing it in context makes the "Overlay" part quite redundant.

magicus:
> Placement is intentional. If you check the texture very closely, there is space left out for a search box, which will be a future addition. Not sure how I could fix this otherwise.

Ok, I accept this for now. I was mostly worried about vertical space though, but I will not force you to update the texture. Maybe the CT wants to have a go at it later on anyway.

magicus:
> > Resizing in the horizontal direction is slow and the mouse movement feels scaled down
> 
> Never experienced this, I will try to reproduce.

I think it might have to do with "incorrect" snapping. That is, if the box is snapped horizontally, and you want to change the vertical size, it acts like it's always snapped and makes it "sticky" and change-resistant. Or maybe it's if you change a corner and not an edge, and move in the direction that is not snapped. Don't remember really, I did not see any really clear pattern.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -313,6 +313,12 @@ public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
             selectedOverlay
                     .getConfigOptionFromString("horizontalAlignmentOverride")
                     .ifPresent(configHolder -> selectedOverlay.updateConfigOption(configHolder));
+        } else if (keyCode == GLFW.GLFW_KEY_ENTER) {
+            ConfigManager.saveConfig();
```
magicus, line 345: I think you need an ` McUtils.mc().setScreen(new OverlaySelectionScreen());` here as well. At least nothing visibly happens when you press enter.

kristofbolyai: Right, oops.

common/src/main/java/com/wynntils/screens/OverlayManagementScreen.java
```diff
@@ -144,20 +144,6 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
                         overlay.getWidth(),
                         overlay.getHeight());
 
-                int xOffset =
```
magicus: Actually I think this part was good. (If I understand the code correctly)

While the alignment now works correctly, it will align the text with the *outside* of the box, making the text overwrite the box border. It would look better if it were aligned with the *inside* of the box (and possibly with a one pixel margin). I think this code tried to achieve that. (Even though it's odd that the additional offset was different in different directions (3 vs 5)).

kristofbolyai: It caused weirdness with small overlays. Maybe we can just have an universal offset. I will try that in a second.

magicus:
The animation looked good, but it is a bit short. I think it could repeat over and over, at least until the user has done any kind of interaction with the overlay (move or resize, or change alignment).

kristofbolyai:
I think we can make it longer, but I wouldn't wait for user input. Do you have a good amount of seconds in mind?

magicus:
This is starting to look real good. See if you can fix the animation cycling and the text overwriting borders, and I think you're good to go.

I'd like to look through the entire code one more time as well, but it is too late for me tonight. I'll come back to this tomorrow.

magicus:
Why not wait for input? I think it would be a good signal that the user found the overlay. But sure, if you want a fixed time, try with something like 5 seconds. (But even then I think it should stop if the user starts to interact earlier than 5 seconds).

kristofbolyai:
Fixed the text issue, and I ended up adding the animation until user input thing. Looks okay I guess.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -175,4 +220,22 @@ public HorizontalAlignment getRenderHorizontalAlignment() {
     public VerticalAlignment getRenderVerticalAlignment() {
         return verticalAlignmentOverride == null ? position.getVerticalAlignment() : verticalAlignmentOverride;
     }
+
+    public Vec2 getCornerPoints(Corner corner) {
+        return switch (corner) {
```
magicus, line 225: Not anything that needs changing, just a note: why does not the compiler complain here? If we update the Corner enum, it might as well getting to a situation where it don't know what to return. That's just odd...

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
```diff
@@ -43,35 +43,68 @@ public QuestInfoOverlay() {
...
+                                this.getWidth(), CommonColors.ORANGE, this.getRenderHorizontalAlignment())),
+                new TextRenderTask(
+                        "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer tempus purus in lacus pulvinar dictum. Quisque suscipit erat pellentesque egestas volutpat.",
+                        TextRenderSetting.getWithHorizontalAlignment(
+                                this.getWidth(), CommonColors.WHITE, this.getRenderHorizontalAlignment())));
 
         private void recalculateRenderTasks() {
-            toRender[0].setSetting(TextRenderSetting.getWithHorizontalAlignment(
-                    this.getWidth(), CommonColors.GREEN, this.getRenderHorizontalAlignment()));
+            toRender.get(0)
```
magicus, line 75: I don't really want to drag this out further (just had a final look through before I was about to approve the whole thing), but I could not help realize that these recalculate functions might be made simpler and more similar by moving the GREEN, ORANGE and WHITE into a final static array, and looping over the toRender and color array. (In fact, they could be made so similar so as to make them a single method, which takes in the toRender or toRenderPreview arrays as argument. (That would almost get us to the point I were requesting before, with just changing the "input source".)

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -206,11 +223,20 @@ private float calculateRenderHeight(List<TextRenderTask> toRender) {
                     }
                 }
 
-                height += lines * (font.lineHeight + NEWLINE_OFFSET);
+                height += lines * (font.lineHeight) + NEWLINE_OFFSET * (lines - 1);
```
magicus:
```suggestion
                height += lines * font.lineHeight + NEWLINE_OFFSET * (lines - 1);
```

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -206,11 +223,20 @@ private float calculateRenderHeight(List<TextRenderTask> toRender) {
...
             }
         }
 
-        return (float) (height - NEWLINE_OFFSET / 2 * McUtils.window().getGuiScale());
+        return (float) (height / 2 * McUtils.window().getGuiScale());
```
magicus, line 230: Not something that needs changing, just me being curious: Why do we divide the height by 2? It seems to me that we do a proper calculation of height. I assumed it was the NEWLINE_OFFSET that were supposed to be divided by 2. But it seems to work, I mean, the alignment look nice and all. I just don't understand why. :-)

common/src/main/java/com/wynntils/utils/MathUtils.java
```diff
@@ -59,4 +59,15 @@ public static int integerFromRoman(String numeral) {
...
+
+    public static double map(double sourceNumber, double fromA, double fromB, double toA, double toB) {
+        double deltaA = fromB - fromA;
+        double deltaB = toB - toA;
+        double scale = deltaB / deltaA;
+        double negA = -1 * fromA;
+        double offset = (negA * scale) + toA;
+        double finalNumber = (sourceNumber * scale) + offset;
+        int calcScale = (int) Math.pow(10, 5);
+        return (double) Math.round(finalNumber * calcScale) / calcScale;
```
magicus: This calcScale thing. 🤔 My first thought was that you wanted to limit the result to a certain number of significant (decimal) digits. But is that really what you are doing? It seems to me that it does nothing at all for large numbers, and risk making very small numbers all 0.

magicus (dismissed):
I say this is okay now. I added a few more comments on improvements. If you want to address them, I'd be happy, but I will not require it.

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
kristofbolyai: I am kinda lost in the GUI scaling math stuff, originally it was supposed to be (height - NEWLINE_HEIGHT) / 2 but now newline height is added in a different place.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: This type of switch statements, (assignment and return ones) force us to implement all enums. So if there is a change in enum values, it doesn't compile, while the regular switch does.

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
magicus: You and me both, brother :-)

magicus (approved):
Ship it!

magicus:
🎉

PR #163, opened 2022-07-15
Item Text Overlay Framework

P0keDev:
This PR adds the `TextOverlayProperty` property type and corresponding `ItemTextOverlayFeature` to render text overlays. For an actual implementation of text overlays, I created `PowderTierProperty`, which includes both the powder type highlight and tier text overlay. I also made some small tweaks to `HighlightProperty` and the properties that implement it, to better handle config options.

Text overlay example:
image


common/src/main/java/com/wynntils/core/webapi/profiles/item/ItemTier.java
```diff
@@ -87,13 +87,22 @@ public ChatFormatting getChatFormatting() {
         return chatFormatting;
     }
 
+    public boolean isHighlightEnabled() {
+        try {
+            return highlightEnabled.call();
+        } catch (Exception e) {
+            e.printStackTrace();
```
magicus: 1) Use logger instead.
2) make return in the catch block, then the comment is unnecessary.

common/src/main/java/com/wynntils/core/webapi/profiles/item/ItemTier.java
```diff
@@ -87,13 +87,22 @@ public ChatFormatting getChatFormatting() {
...
+        }
+        return false; // some error occurred
+    }
+
     public CustomColor getHighlightColor() {
         try {
-            if (highlightEnabled.call()) return highlightColor.call();
+            return highlightColor.call();
         } catch (Exception e) {
             e.printStackTrace();
```
magicus: Here as well...

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
```diff
@@ -0,0 +1,78 @@
...
+        if (!wynnItem.hasProperty(ItemProperty.TEXT_OVERLAY)) return;
+
+        for (TextOverlayProperty overlayProperty : wynnItem.getProperties(ItemProperty.TEXT_OVERLAY)) {
+            boolean contextEnabled = hotbar ? overlayProperty.isHotbarText() : overlayProperty.isInventoryText();
+            if (!overlayProperty.isTextOverlayEnabled() || !contextEnabled) continue; // not enabled or wrong context
+
+            TextOverlayProperty.TextOverlay textOverlay = overlayProperty.getTextOverlay();
+
+            PoseStack poseStack = new PoseStack();
+            poseStack.translate(0, 0, 300); // items are drawn at z300, so text has to be as well
```
magicus, line 63: We should not draw it on a z on top of the item? Like 301 (or 299, if z goes the other way). It might not make any practical difference, but it seems more robust.

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
```diff
@@ -0,0 +1,78 @@
...
+        if (!hotbarTextOverlayEnabled) return;
+
+        drawTextOverlay(e.getStack(), e.getX(), e.getY(), true);
+    }
+
+    public void drawTextOverlay(ItemStack item, int slotX, int slotY, boolean hotbar) {
+        if (!(item instanceof WynnItemStack wynnItem)) return;
+        if (!wynnItem.hasProperty(ItemProperty.TEXT_OVERLAY)) return;
+
+        for (TextOverlayProperty overlayProperty : wynnItem.getProperties(ItemProperty.TEXT_OVERLAY)) {
```
magicus, line 56: This for loop... Can we have more than one text overlay per item? If we *really* want to support that, we'd likely have to offset the texts, so as not to have them completely overlap. 

Or, if we in practice only support one text overlay, maybe make that an explicit consistency check, and remove the loop.

common/src/main/java/com/wynntils/utils/MathUtils.java
```diff
@@ -20,4 +22,41 @@ public static float inverseLerp(float a, float b, float value) {
     public static double inverseLerp(double a, double b, double value) {
         return (value - a) / (b - a);
     }
+
+    public static int integerFromRoman(String numeral) {
```
magicus, line 26: I was like "there must be a apache commons method to do this"... and then 5 minutes later "wat! really not?" *sigh*

P0keDev: i experienced the same thing lol - ended up copying this method from legacy, so i guess that's the reason it existed

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
kristofbolyai: We do Z 400 for the item highlights iirc

magicus: So maybe Z 200 for text then?

kristofbolyai: Why not 401?

magicus: I thought highlights were placed behind the items, while text should be placed in front of. So I assumed that lower Z were closer to the viewer, and higher further away. But maybe my mental model is incorrect?

P0keDev: z 200 draws the text behind the item (larger z = higher layer)

i didn't test values above 300, but my worry would be that text would get drawn over other inventory elements, like the slot highlight square

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
P0keDev: the placement of text is defined by the properties, so in the (probably rare) case of potentially overlapping texts, we'll just need to take care to give them different offsets

P0keDev: also, text overlays that exist in different contexts (inventory vs hotbar) could theoretically be applied to the same item

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
kristofbolyai: Oh yeah so 300 should be fine then. If it's a problem we can fix that later.

P0keDev:
after testing more Z values for the text rendering, 350 works fine but 400 causes this to happen: 
image

personally i think it's best to avoid drawing at any z higher than what is strictly necessary, and z 300 should work just fine

common/src/main/java/com/wynntils/utils/MathUtils.java
kristofbolyai: I am pretty sure we have a better version of this in legacy where it can convert anything basically.

P0keDev: if we do, i didn't find it - we definitely have this one

kristofbolyai: Ah it is for the other way around.

common/src/main/java/com/wynntils/wc/custom/item/properties/PowderTierProperty.java
```diff
@@ -0,0 +1,67 @@
...
+        if (chatColor == null) chatColor = ChatFormatting.WHITE;
+
+        highlightColor = CustomColor.fromChatFormatting(chatColor);
+
+        // parse tier
+        Matcher powderMatcher = WynnItemMatchers.powderNameMatcher(item.getHoverName());
+        String powderNumeral = "I";
+        if (powderMatcher.matches()) {
+            powderNumeral = powderMatcher.group(2);
+        }
```
kristofbolyai, line 38: What if it doesn't match? Why do we assume TIER 1?

kristofbolyai (approved):
Looks good, I only have that question about the powder numeral matching.

common/src/main/java/com/wynntils/wc/custom/item/properties/PowderTierProperty.java
P0keDev: the same pattern is used to detect powder items in the first place, so it should be completely impossible for any of it to fail anyway - i didn't choose tier 1 for any particular reason

kristofbolyai: I see, it's fine then I guess.

common/src/main/java/com/wynntils/features/user/ItemTextOverlayFeature.java
magicus: Ok!

common/src/main/java/com/wynntils/wc/utils/WynnItemMatchers.java
```diff
@@ -25,6 +25,8 @@ public class WynnItemMatchers {
     private static final Pattern ITEM_RARITY_PATTERN =
             Pattern.compile("(Normal|Set|Unique|Rare|Legendary|Fabled|Mythic)( Raid)? (Item|Reward).*");
     private static final Pattern DURABILITY_PATTERN = Pattern.compile("\\[(\\d+)/(\\d+) Durability\\]");
+    public static Pattern POWDER_PATTERN =
+            Pattern.compile("§[2ebcf8].? ?(Earth|Thunder|Water|Fire|Air|Blank) Powder ([IV]{1,3})");
```
magicus, line 29: (As a side note, I spent longer than I care to admit trying to think about if it was possible to express roman numerals properly in RE. :-))

magicus (approved):
Ship it!

PR #164, opened 2022-07-15
Add properties file to support hotspot-agent

magicus:
If anyone knows the proper gradle-fu needed to make sure this file only gets included in dev builds, please go ahead and suggest it. :) (Otherwise, it doesn't harm anything by being in the release builds as well.)


kristofbolyai (dismissed):

magicus:
Actually, I need to do gradle magic anyway. Must also pass proper command line options to the JVM, so this fix is only half-way there...

kristofbolyai:
> Actually, I need to do gradle magic anyway. Must also pass proper command line options to the JVM, so this fix is only half-way there...

JVM args should be set up by the user, since if the hotswap agent is not present, and we inject the args, it does not run?

magicus:
@kristofbolyai Sorry, it was a bit too much Yoda grammar for me to parse. I think you are asking if a non-hotswap JVM fails if we pass the hotswap -XX argument. If so, the answer is: yes. So we need to verify that the JVM can accept the hotswap arguments before adding them.

I pushed a half-fix; I setup the args conditionally on a bool, but I have yet not figured out how to detect the proper value of this bool. I guess I need to do something like `Runtime.exec()` and try to tease out the path to the JVM from gradle.

Also, we should probably add the sponge mixin jar to the repo and add something like:
```
          vmArgs "-javaagent:$PROJECT_ROOT/utils/sponge-mixin-0.11.4+mixin.0.8.5.jar"
```
to get hotswap working in mixins, according to Loom docs. If this plays well with hotswap agent, that is. I have not been able to verify this; my environment got all messed up and I can't do any hotswapping at all, right now. :-(

kristofbolyai:
Yes I meant that. Also I am not sure if we need mixin hotswaps, it might increase swap time and there is little benefit

magicus:
If I recall correctly, the mixin hotswapping also allowed you to set breakpoints in mixins, which is definitely beneficial. I got mixin hotswapping working on one of my other Fabric mods; I don't remember that it was any slower at all, but it was messy when you upgraded to a new version of Loom.

kristofbolyai:
> If I recall correctly, the mixin hotswapping also allowed you to set breakpoints in mixins, which is definitely beneficial. I got mixin hotswapping working on one of my other Fabric mods; I don't remember that it was any slower at all, but it was messy when you upgraded to a new version of Loom.

Minecraft development plugin allows breakpoints

magicus:
Ah, I see.

build.gradle
```diff
@@ -8,6 +8,8 @@ architectury {
     minecraft = rootProject.minecraft_version
 }
 
+def usingHotswapAgent = true
```
kristofbolyai: What do we think of defaulting this to false, and having this in as an easy "opt-in"?

magicus: Yeah, that is perhaps a good solution for now. Or even setting your own default value in `~/.gradle/gradle.properties`.

magicus: I just wanted to run `java -XX:HotswapAgent=fatjar --version` and check the exit code, but I failed to figure out how to do that. Sometimes I really hate Gradle.

kristofbolyai (approved):

Issue #23, opened 2022-02-27
`WorldStateImpl` occasionally has wrong world state


kristofbolyai:
During development I have found that `WorldStateImpl#onServer` returns `false` on multiple occasions, while connected to the server. This is likely related to the mixins that call the world state change events.

P0keDev:
~~Potentially related: having Litematica for Fabric installed stops Artemis from detecting world joins at all, even though the lobby and class selection triggers still work~~

EDIT: this is not actually the case, my bad

kristofbolyai:
I have finally made progress on this bug! Sometimes, the events that we use to check world state (this case `PlayerDisplayNameChangeEvent`), does not have the correct UUID, so `e.getId().equals(WORLD_UUID)` fails. 

According to git blame, @magicus wrote the WORLD_UUID code. Can you elaborate where the UUID checking came from and is it neccessary?

kristofbolyai:
Oh, I see now. Legacy has the same issue. Weird. I guess we should not depend on using UUIDs then. I will fix this bug, but I will wait for magicus before starting on it.

kristofbolyai:
The strangest thing is that only a full restart of the game fixes this issue, relogging does not. Weird.

magicus:
I don't really remember. I think I just traced packets that were sent and tried to look for suitable patterns to lock onto. What is the UUID in the failing cases? Is it a second, alternative UUID (which might carry some significance, perhaps?), or is it just random, or is it left out entirely?

kristofbolyai:
It was `3d524cd9-c75b-3410-b1da-de66593bd4f6` in my case, however I don't know if it is random or not. Until I restarted, it seemed consistent.

magicus:
Looking at the code, it seems the UUID is necessary to extract the world name.

kristofbolyai:
> Looking at the code, it seems the UUID is necessary to extract the world name.

How so? 

```java
    @SubscribeEvent
    public void update(PlayerDisplayNameChangeEvent e) {
        if (!onServer()) return;

        if (e.getId().equals(WORLD_UUID)) {
            Component displayName = e.getDisplayName();
            String name = ComponentUtils.getUnformatted(displayName);
            Matcher m = WORLD_NAME.matcher(name);
            if (m.find()) {
                String worldName = m.group(1);
                setState(State.WORLD, worldName);
            } else {
                WynntilsMod.error("World name did not match pattern: " + name);
            }
        }
    }
 ```

[...]

PR #165, opened 2022-07-15
HealthPotionBlockerFeature: Fix health potions being consumed at max health when looking at block

TomaSajt:
I noticed when testing my previous PR, that sometimes the `HealthPotionBlockerFeature` wouldn't work, so I looked into it, and it turns out Minecraft sends not only `UseItem` packets, but `UseItemOn` packets, when looking at a block. This made the game display the normal "You're at max health" text, but it also consumed the potion.

This PR fixes the feature by cancelling `UseItemOn` packets too.


kristofbolyai:
This looks good, though I just noticed that this feature is outdated compared to legacy. 

In legacy, we have a max health threshold when potions should be blocked (like block potions above 97%). Do you wanna update this feature (it is pretty simple), or should we just accept this as is?

TomaSajt:
> Do you wanna update this feature (it is pretty simple), or should we just accept this as is?

Sure, I'll try to port it over.

kristofbolyai:
> > Do you wanna update this feature (it is pretty simple), or should we just accept this as is?
> 
> Sure, I'll try to port it over.

Cool, thanks

TomaSajt:
I didn't add effective-percentage based blocking, like in legacy, as there isn't yet an action-bar parsing system in place to get the real health of the player.

Incompleteusern (approved):

common/src/main/java/com/wynntils/features/user/HealthPotionBlockerFeature.java
```diff
@@ -12,25 +13,45 @@
...
+        Component response = getBlockResponse(e);
+        if (response != null) {
+            e.setCanceled(true);
+            McUtils.sendMessageToClient(response);
+        }
+    }
+
+    @SubscribeEvent
+    public void onPotionUseOn(PacketSentEvent<ServerboundUseItemOnPacket> e) {
+        if (getBlockResponse(e) != null) e.setCanceled(true);
```
kristofbolyai, line 40: So we don't need to send a chat message here because the other event already sent one? Did you test this by looking at the sky for example?

TomaSajt: Yes, I checked looking at the sky too.

TomaSajt:
I think this feature is ready for now, however there were I think 2 times while testing when the feature just didn't work at all. I think this is not an issue introduced in this PR. My hypothesis is that the world state is still sometimes incorrect, so `Wynnutils.onWorld()` returns false. *I don't have evidence of this, just a guess*

P0keDev:
> I think this feature is ready for now, however there were I think 2 times while testing when the feature just didn't work at all. I think this is not an issue introduced in this PR. My hypothesis is that the world state is still sometimes incorrect, so `Wynnutils.onWorld()` returns false. _I don't have evidence of this, just a guess_

your hypothesis is correct: #23

magicus:
1) Action bar parsing is implemented in my ChatManager branch. Maybe I should commit that part, even if the entire ChatManager is not done?

2) The workd state thing is irritating indeed. Maybe we should have a helper function for the typical "if not on work return" check, that can print log output if it fails, so we at least can diagnose that this is what is happening.

TomaSajt:
> 1) Action bar parsing is implemented in my ChatManager branch.

Oh... #168 is doing action bar handling too... It's my bad for not asking about the scope of your PR. I don't mind if I have to discard my changes, but I think the Manager way of doing stats might be more favourable (for storing health and mana). I'll leave up the PR until you're done with your chat manager.

PR #166, opened 2022-07-16
Fix unnecessarily removing empty lines from lore

TomaSajt:
In #160 I accidentally introduced a bug, where sometimes the empty lines from the lore would disappear. This PR fixes that.


common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -213,8 +214,10 @@ public static List<Component> wrapTooltips(List<Component> tooltips, int maxWidt
     }
 
     public static List<Component> splitComponent(Component component, int maxWidth) {
-        return McUtils.mc().font.getSplitter().splitLines(component, maxWidth, Style.EMPTY).stream()
```
P0keDev, line 216: i assume the issue was just that this function returns an empty list if given a blank/newline component?

TomaSajt: Yes, sorry for not specifying.

P0keDev (approved):

PR #168, opened 2022-07-16
Add ActionBarManager

TomaSajt:
This PR adds `ActionBarParserFeature` which gets things like total current health from the actionbar. However this won't really be usable until there is a way to load CharacterInfo without entering the class selection screen as it is a dependency of this.
(it could be usable if we put the health data outside of CharacterInfo, however I don't think that's the way to go)


TomaSajt:
I also added a TODO to get the partial spell progress. (I'll try to implement it myself in the near future)

common/src/main/java/com/wynntils/wc/model/Character.java
```diff
@@ -108,6 +111,30 @@ public int getId() {
...
+        public int getMaxHealth() {
+            return maxHealth;
+        }
+
+        public int getHealth() {
+            return health;
+        }
+
+        public int getMaxMana() {
+            return 20;
```
kristofbolyai: Not 2.0 ready

kristofbolyai (commented):
I think this should be more of a manager (like OverlayManager or ScoreboardManager), and not an internal feature. That will also mean that we can just have something like ActionBarManager#getCurrentHealth and we don't have to worry about CharacterInfo, which is more of a class data holder. 

This is a relatively small change, but this fits Artemis' code structure better.

Thank you for adding this, I am planning to work on health and mana bar overlays after my draft PR, this will be definitely useful.

common/src/main/java/com/wynntils/wc/model/Character.java
TomaSajt: New commit now has parsed mana too

common/src/main/java/com/wynntils/wc/utils/ActionBarManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private static final Pattern ACTIONBAR_PATTERN =
+            Pattern.compile("§c❤ ([0-9]+)/([0-9]+)§0\\s+(.+)\\s+§b✺ ([0-9]+)/([0-9]+)");
+
+    private static String previousActionBar = null;
+
+    private static int currentHealth = 100;
+    private static int maxHealth = 100;
+    private static int currentMana = 20;
+    private static int maxMana = 20;
```
kristofbolyai: We assume these default values? Can we just make it null, or what is the use?

TomaSajt:
```java
Pattern.compile("(?:§❤ *([0-9]+)/([0-9]+))?.*? {2,}(?:§([LR])§-(?:§([LR])§-§([LR])?)?)?.*".replace("§", "(?:§[0-9a-fklmnor])*"));
```
this is the legacy regex pattern
I see that it checks for every possible §-code at multiple places. In my version I just used the color that I logged in the console.
I don't know of any event the game, where the hp is not red or the mana is not aqua. If you know of somewhere where it's a different color, I'll use this more generic version instead.

kristofbolyai:
> ```java
> Pattern.compile("(?:§❤ *([0-9]+)/([0-9]+))?.*? {2,}(?:§([LR])§-(?:§([LR])§-§([LR])?)?)?.*".replace("§", "(?:§[0-9a-fklmnor])*"));
> ```
> 
> this is the legacy regex pattern I see that it checks for every possible §-code at multiple places. In my version I just used the color that I logged in the console. I don't know of any event the game, where the hp is not red or the mana is not aqua. If you know of somewhere where it's a different color, I'll use this more generic version instead.

I think it is fine to use the legacy pattern even if it is a bit larger, regex is very fast already and this won't affect much, but it will be less prone to bugs/Wynn changes.

common/src/main/java/com/wynntils/wc/utils/ActionBarManager.java
TomaSajt: It was arbitrary, I could do -1 like in legacy, but I guess null wouldn't be that bad either. My only concert about null is querying the health stats before the first actionbar event, like maybe the HP-bar overlay tries to render before the event. I could also use the Integer boxed type and then you could handle null from the overlay. For now i'll go with -1, if that's ok.

kristofbolyai: Sure, -1 should be okay, though I hope it won't cause weirdness. It just has to be handled correctly, just like null.

kristofbolyai (approved):
Looks good! Thank you for contributing.

TomaSajt:
I guess this PR is a semi-duplicate of #123. ~~I'll delete this once that's completed~~

P0keDev:
personally i'm in favor of separate chat and action bar managers, because even if action bar updates are technically chat messages, it's not a natural connection to make within the context of the game itself. thus, i think it's fine to merge this and deal with conflicts or expand it later on

P0keDev (approved):

kristofbolyai:
Agree, I would like ChatManager to handle chat tabs, message cancelling and overwriting, etc, this is mostly unrelated.

TomaSajt:
I just realized that instead of `PacketReceivedEvent<ClientboundChatPacket>` I could have used `ChatReceivedEvent`. It's essentially the same, but not so low-level. Might change it when I'm done with the skill parsing thing, (but probably not for a while, because I'm traveling abroad tomorrow.)

PR #170, opened 2022-07-19
WIP: Bounding Volume Hierarchy with Level Of Detail

Kepler-17c:


magicus:
Lots of complex data structures. What do you intend to use these for?

Kepler-17c:
> Lots of complex data structures. What do you intend to use these for?

My first intention was only using them for the path-finder. But the BVH can be extended to allow tree cuts and LOD, which would make it perfect for handling the in-game map: All areas could just be added with their in-game coords (maybe even on a block-level) and the presentation would request a specific area + zoom level.

Kepler-17c:
And in case you wonder, this is only part of what I actually have for the path-finder. I just split it, so this part can be used for the map first without the need of waiting for my full path-finder to be completed.

Kepler-17c:
If you have any questions about the architecture/ideas/concepts/..., just lmk

magicus:
I'm not fond of the idea of pushing a lot of abstract code that is not being used. We can keep this as a Draft PR for an extended period of time, if you want to show what code could be available to other parts of the code.

But pushing in dead code, before we have functionality that actually uses it, is a no-no to me.

Kepler-17c:
Yea, of course - it's drafted, because it's not ready yet - I'll also add WIP prefix to make the status more present

Kepler-17c:
Making some good progress: Fixed some bugs, but found an annoying Heisenbug (I know what to look for, though)

magicus (changes requested):

magicus:
@Kepler-17c This PR has been in the draft state for a long time. I will close it now. If you dedice to continue working on this functionality, please open a new PR from your branch!

PR #172, opened 2022-07-23
Features+Overlays: Add game notifications, GameNotificationOverlay, NotificationFeature and PouchRedirectFeature

kristofbolyai:
This PR adds the base game update overlay, with it's actual features and parsing left out (except `PouchRedirectFeature`, which is the example use case). As this overlay is mostly called by chat events, I am holding off on implementing that before #123 is discussed/merged. The PR also fixes a few issues in `FeatureRegistry`, and changes `FontRenderer` to support multiple lines, and coloring with color codes, and bug fixes. 

There is a structural change in this PR as well. As you can see, I created sub-packages in the `features.user` package to put already categorized features, as I believe we should have the categorization in the code structure, not just the actual mod. If any of you don't like those changes, I can revert it, but I think it is a simple and reasonable change.

Example: 
image


Incompleteusern (dismissed):

common/src/main/java/com/wynntils/core/objects/MessageContainer.java
```diff
@@ -0,0 +1,53 @@
...
+import com.wynntils.mc.render.TextRenderTask;
+
+public class MessageContainer {
+    TextRenderTask message;
+    long endTime;
+
+    public MessageContainer(String message) {
+        this.message = new TextRenderTask(message, TextRenderSetting.DEFAULT);
+        this.endTime = System.currentTimeMillis()
+                + (long) (GameUpdateOverlayFeature.getInstance().gameUpdateOverlay.messageTimeLimit * 1000);
```
magicus: Not liking that functionality in `core` depends on a Feature... :-/ 

I'm not sure of the best way to solve this, but I think it seems reasonable that the default time limit should be a part of the MessageContainer class instead.

I assume you have it this way so that it can be a `@Config` in the Feature... It's a tricky one.

magicus: ... or maybe this class should be an inner class/helper class to GameUpdateOverlayFeature?

kristofbolyai: Yes, this is a helper class, but I including it in the feature seemed to be a little overwhelming, and I also do not want to see `GameUpdateOverlayFeature.MessageContainer` every time we need to use it.

magicus:
And, also, please can we figure out a better name for this functionality? "Update" sounds more like we're pushing a new version of the mod.

"Notifications"? "Alerts"? "Notices"? "Announcements"? "NoticeBoard"?

kristofbolyai:
`GameStateUpdate`?

kristofbolyai:
`GameNotification` or `GameStateNotification` is also okay IMO.

common/src/main/java/com/wynntils/features/user/redirects/PouchRedirectFeature.java
```diff
@@ -0,0 +1,71 @@
...
+            Matcher matcher = EMERALD_POUCH_PICKUP_PATTERN.matcher(stringWithFormattingCodes);
+            if (matcher.matches()) {
+                event.setCanceled(true);
+
+                // If the last emerald pickup event was less than 3 seconds ago, assume Wynn has relayed us an "updated"
+                // emerald title
+                // Edit the first message it gave us with the new amount
+                // editMessage doesn't return the new MessageContainer, so we can just keep re-using the first one
+                if (lastEmeraldPouchPickup > System.currentTimeMillis() - 3000 && emeraldPouchMessage != null) {
+                    GameUpdateOverlayFeature.editMessage(emeraldPouchMessage, stringWithFormattingCodes);
```
magicus: Will this reset the timer? If the message change, we need the full time to be able to read it. Otherwise, it might be about to disappear in a few milliseconds, and the user will in practice miss the updated text.

kristofbolyai: `editMessage` calls reset timer method.

magicus:
I have some more overarching thoughts that I'm not sure how to express (I'm already half-asleep, perhaps not the best time to review code ;-)).

First, I think we need to express some connection between the GameUpdate/Notification overlay, and the PouchRedirectFeature. That is, if you enable pouch redirect, but do not have the game update/notification overlay visible, you will just lose information, and if you are not accustomed to Wynntils and have just fiddled around a bit, it can be hard to figure out what went wrong.

I'm not sure how. Maybe make the Redirect feature "dependent" on the Overlay to be shown for it to be active? Or just print a warning if the user is about to mess up. But something needs to be there to prevent information loss.

Second, I think maybe the game notification system (I like that better than "game update" anyway, so let's go with it for the time being) is so pervasive (at least in Legacy), that it deserves to be more of a `core` functionality, rather than a UserFeature. I'm thinking something like this:

PouchRedirectFeature should call like `com.wynntils.core.Notification.post("Added N emeralds to pouch")`, so there is a single "channel" for sending out these kinds of notifications. Then the NotificationOverlayFeature can show them on the screen. Or, we can send them as chat messages, if the Overlay is not enabled. We'd just need to design where and how to make the cut between the "core" part and the "overlay" part of the functionality, and make the API between them proper and kosher.

magicus:
... and maybe my second thought resolves the issues that I raised in the first, if we can fallback to printing to the chat instead. :-) 

I'd better get some sleep now. 💤

kristofbolyai:
I like these ideas! I will get around implementing something similar and we will see how it will turn out.

kristofbolyai:
These commits should be mostly what magicus planned out earlier.

common/src/main/java/com/wynntils/features/user/overlays/GameNotificationOverlayFeature.java
```diff
@@ -0,0 +1,175 @@
...
+        @Override
+        public void render(PoseStack poseStack, float partialTicks, Window window) {
+            List<MessageContainer> toRender = new ArrayList<>();
+
+            ListIterator<MessageContainer> messages = messageQueue.listIterator(messageQueue.size());
+            while (messages.hasPrevious()) {
+                MessageContainer message = messages.previous();
+
+                if (message.getRemainingTime() <= 0.0f) {
+                    messages.remove(); // remove the message if the time has come
```
magicus, line 109: Something does still not feel right about the design. It should not be the responsibility of the Overlay to remove messages, if the timer is in the messages themselves. 

There is something schizophrenic about the notifications, how they are partially a "core" functionality, but still tightly coupled to how they are displayed. I think either the core notification system should only accept new messages and pass them on, without knowing anything about timers, or the queue of not-yet-expired notifications should reside in the notification manager. (I think I like the latter approach better).

common/src/main/java/com/wynntils/features/user/redirects/PouchRedirectFeature.java
```diff
@@ -0,0 +1,71 @@
...
+import com.wynntils.core.notifications.MessageContainer;
+import com.wynntils.core.notifications.NotificationManager;
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import com.wynntils.mc.utils.ComponentUtils;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@FeatureInfo(category = "Game Update Overlay Redirects")
```
magicus:
```suggestion
@FeatureInfo(category = "Game Notification Redirects")
```

common/src/main/java/com/wynntils/features/user/overlays/GameNotificationOverlayFeature.java
kristofbolyai: I think removing message queue in the NotificationManager and only sending out events is fine. But I think features consuming these messages can have their own system of storing them.

magicus: I'm still not convinced but I'm prepared to turn a blind eye for this time only. ;-)

magicus: Just to clarify: I could be okay with the Feature keeping a queue of notifications, and the manager just responsible for sending them out as they arrive.

But, then the time delay should be kept out of the manager, since that is part of the overlay queue handling.

kristofbolyai: That is the case?

common/src/main/java/com/wynntils/features/user/overlays/QuestInfoOverlayFeature.java
```diff
@@ -24,7 +24,7 @@
 import com.wynntils.wc.utils.scoreboard.quests.QuestManager;
 import java.util.List;
 
-@FeatureInfo(category = "overlays")
+@FeatureInfo(category = "Overlays")
```
magicus, line 27: Btw, did we ever discuss if the categories should be i18n'ed, or made into like an enum so we keep a proper collection of what categories there are?

common/src/main/java/com/wynntils/features/user/overlays/GameNotificationOverlayFeature.java
kristofbolyai: Oh well there is a time limit there too. I can remove that I guess.

common/src/main/java/com/wynntils/features/user/overlays/QuestInfoOverlayFeature.java
kristofbolyai: I18nd enum was settled on iirc, but we never did it.

magicus (approved):

PR #173, opened 2022-07-24
Overlays: Add health bar and mana bar overlays

kristofbolyai:
This PR adds health and mana bar overlays. These are the first replacement overlays, they seem to work fine. Tested it with forge and fabric, and I applied the forge specific fixes. 

Fill animations are missing compared to legacy, they can be added later, this is just a base for improvement (basically changes are instant and not animated).

Example: 
image


common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
```diff
@@ -56,4 +60,46 @@ public void onRenderGuiPre(PoseStack poseStack, CallbackInfo ci) {
...
+            int k,
+            int l,
+            boolean bl,
+            CallbackInfo ci) {
+        if (EventFactory.onRenderHearthsPre(poseStack, this.minecraft.getWindow())
+                .isCanceled()) {
+            ci.cancel();
+        }
+
+        RenderSystem.setShaderTexture(0, GuiComponent.GUI_ICONS_LOCATION); // we have to reset shader texture
```
magicus, line 83: Why is that? Is this needed even if no-one listens to the event? Or is the event handler messing things up? If so, should not they clear this up?

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsFeature.java
```diff
@@ -0,0 +1,231 @@
...
+import com.wynntils.mc.render.Texture;
+import com.wynntils.mc.render.VerticalAlignment;
+import com.wynntils.utils.objects.CommonColors;
+import com.wynntils.utils.objects.CustomColor;
+import com.wynntils.wc.utils.ActionBarManager;
+
+@FeatureInfo(category = "overlays")
+public class CustomBarsFeature extends UserFeature {
+    @OverlayInfo(renderType = RenderEvent.ElementType.HealthBar, renderAt = OverlayInfo.RenderState.Replace)
+    private final Overlay HealthBarOverlay = new HealthBarOverlay();
```
magicus: Field names traditionally begin with lower case letter.

It gets even messier since now we have a class and an instance with the very same name. Legal java, but not good practice.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -501,6 +501,126 @@ public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack pos
...
+            float y1,
+            float x2,
+            float y2,
+            int textureX1,
+            int textureY1,
+            int textureX2,
+            int textureY2,
+            float progress,
+            boolean background) {
+        if (background) {
```
magicus: Looks like this is basically two different functions combined, `drawProgressBarBackground` and `drawProgressBarForeground`. Maybe you should split it up instead of sending in a `backgorund` boolean?

Or see if you can unify the two code blocks a bit. It does not seem impossible, from a quick glance.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -501,6 +501,126 @@ public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack pos
         poseStack.popPose();
     }
 
+    /** drawProgressBar
+     * Draws a progress bar(textureY1 and textureY2 now specify both textures with background being on top of the bar)
```
magicus: In the comments, `(` is missing space in front of it in several places.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -501,6 +501,126 @@ public static void drawTooltip(List<ClientTooltipComponent> lines, PoseStack pos
...
+            int textureX2,
+            int textureY2,
+            float progress,
+            boolean background) {
+        if (background) {
+            Matrix4f matrix = poseStack.last().pose();
+
+            RenderSystem.enableTexture();
+            RenderSystem.setShader(GameRenderer::getPositionTexShader);
+            RenderSystem.setShaderTexture(0, texture.resource());
```
magicus: Is this what is causing the 
```
 RenderSystem.setShaderTexture(0, GuiComponent.GUI_ICONS_LOCATION)
```
in the mixin?

Can we read out the current shader texture, save it, and restore it, in this function?

kristofbolyai: We cant't. Basically we only have 1 method for rendering healths, food and air. We overwrite the shader in between those render calls, and we need to reset them, since the MC source only sets it before the method call.

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
kristofbolyai: I think it is to set the state to the original one in the mixin, instead of the things that use this event.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
magicus: But we could call `RenderSystem.setShaderTexture(0, GuiComponent.GUI_ICONS_LOCATION)` here instead, and not have it in the mixin, right?

kristofbolyai: We could, but if we use this for anything else (we will) we don't need it.

magicus: (I'm not sure it's better, though...)

kristofbolyai: I would much rather call it in the mixin than here

magicus: Ok.

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsFeature.java
magicus: If you fix the field names I'm ok with this PR.

magicus (approved):

PR #174, opened 2022-07-25
Add QuickCastFeature

TomaSajt:
This PR will try to add a quick-casting system to the game from legacy.
(PR is still work in progress.)


kristofbolyai:
Can you share a bit of detail about your plan? Are you planning on porting legacys safeguarded system with the bugs fixed? Are you checking if we really cast a spell and retry if not?

TomaSajt:
I still have a lot of things I'm not aware of about legacy, for example those bug fixes. I'm not really sure of what you mean exactly by the safeguard.

I am planning on making it so it completes the spell, like in legacy, but haven't gotten to it yet.

I saw that there were some shenanigans with packet sending and retrying, I don't really know how that works either. (I just put in a 1 tick delay between sending the packets)

kristofbolyai:
Okay so, legacy checks for correct weapon type, skill point for the weapon, and other requirements. I would like to see these here as well. Secondly, legacy sends a packet, waits for the action bar update "R - -", then sends the next one (R L -), etc. We discussed this and I think we can just send all 3 packets essentially at the same time, then check if the spell was cast, then resend if not (accounting for partial progress like R R -).

kristofbolyai:
I also remember @P0keDev being knowledgeable about this topic, so I would like to hear from him as well.

TomaSajt:
The feature is semi-stable now, it has proper spell completion.
If the partial spell is not compatible, it won't queue it up, you'll need to re-quick-cast after the spell timed out.
There still isn't a system in place to prevent overlapping quick-casts (I'll work on that tomorrow).
The method I used for spell forgetting is just waiting 40 ticks. (That seems to be the time needed on the server too, but lag can sometimes mess this up). I think this way is not that bad, so if no one has an issue with it, I'll keep it.
The weapon, class and level requirements have also been implemented.
The Archer/Hunter class has their spells reversed properly.
The packet queue is polled every 3 ticks when not empty (to avoid lag messing up spell-casts by accidentally overlapping two inputs).
I still have the i18n to do.

common/src/main/java/com/wynntils/wc/utils/WynnItemMatchers.java
```diff
@@ -84,6 +84,14 @@ public static boolean isEmeraldPouch(ItemStack itemStack) {
         return itemStack.getHoverName().getString().startsWith("§aEmerald Pouch§2 [Tier");
     }
 
+    /**
+     * Returns true if the passed item has an attack speed
+     */
+    public static boolean isWeapon(ItemStack heldItem) {
```
magicus: It's not necessarily a "held item", is it?

common/src/main/java/com/wynntils/wc/utils/SpellManager.java
```diff
@@ -0,0 +1,72 @@
...
+import com.wynntils.mc.event.TitleSetTextEvent;
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class SpellManager {
+    private static final Pattern SPELL_PATTERN = Pattern.compile(
+            "§([LR]|Right|Left)§-§([LR?]|Right|Left)§-§([LR?]|Right|Left)§".replace("§", "(?:§[0-9a-fklmnor])*"));
+    public static final boolean[] NO_SPELL = new boolean[0];
```
magicus: I realize this is how it was done in Legacy, but I'm not really fond of it. The spells are not properly "boolean". Yes, I understand it's a two-state variable, but they should be enums, since they carry no meaningful truth value.

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
```diff
@@ -0,0 +1,145 @@
...
+            new KeyHolder("Cast 4th Spell", GLFW.GLFW_KEY_V, "Wynntils", true, QuickCastFeature::castFourthSpell);
+
+    private static final Pattern INCORRECT_CLASS_PATTERN =
+            Pattern.compile("§✖§ Class Req: (.+)".replace("§", "(?:§[0-9a-fklmnor])*"));
+    private static final Pattern LVL_MIN_NOT_REACHED_PATTERN =
+            Pattern.compile("§✖§ (.+) Min: ([0-9]+)".replace("§", "(?:§[0-9a-fklmnor])*"));
+    private static final Packet<?> RIGHT_CLICK_PACKET = new ServerboundUseItemPacket(InteractionHand.MAIN_HAND);
+    private static final Packet<?> LEFT_CLICK_PACKET = new ServerboundSwingPacket(InteractionHand.MAIN_HAND);
+    private static final Queue<Packet<?>> SPELL_PACKET_QUEUE = new LinkedList<>();
+    private static final boolean RIGHT = true;
```
magicus: This also ties in with my comment about an enum in SpellManager.

(It need not necessary be LEFT and RIGHT; to account for archer it could also be like PRIMARY, SECONDARY).

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
```diff
@@ -0,0 +1,145 @@
...
+        List<Boolean> remainder = new ArrayList<>(spell.subList(partialSpell.length, spell.size()));
+        remainder.stream().map(x -> x ? RIGHT_CLICK_PACKET : LEFT_CLICK_PACKET).forEach(SPELL_PACKET_QUEUE::add);
+    }
+
+    @SubscribeEvent
+    public void onTick(ClientTickEvent e) {
+        if (!WynnUtils.onWorld() || e.getTickPhase() != ClientTickEvent.Phase.END) return;
+        if (SPELL_PACKET_QUEUE.isEmpty()) return;
+        if (--packetCountdown > 0) return;
+        if (McUtils.inventory().selected != lastSelectedSlot)
```
magicus: Please don't use multi-line if statements without braces.

TomaSajt: Yeah, sorry, the spotless formatting git-hook sometimes wraps when I don't want it to.

common/src/main/java/com/wynntils/wc/utils/SpellManager.java
TomaSajt: I went with the name SpellUnit.

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
TomaSajt: I ended up leaving them as boolean as I didn't want to make another enum for this.

TomaSajt:
What I did not manage to do is prevent regular attacking while the queue is processing. When I tried to cancel the regular packets, it would cancel the queue packets too. I don't know of a way of distinguishing between those two types packets, as they are sent together. If anyone knows a way, I'll do that.

kristofbolyai:
> What I did not manage to do is prevent regular attacking while the queue is processing. When I tried to cancel the regular packets, it would cancel the queue packets too. I don't know of a way of distinguishing between those two types packets, as they are sent together. If anyone knows a way, I'll do that.

Can you confirm it is actual attacking and not visual?

TomaSajt:
> Can you confirm it is actual attacking and not visual?

I'm not sure I know what you mean by that.

The problem is that I want to cancel the sending of the Swing and Use packets, but the feature also sends them and I don't want to cancel those.
So either I don't cancel anything (so the player can sometimes mess up the spell-cast if they interfere with a regular Swing/Use) or I cancel the packets, but then I can't send the quick-cast either.

common/src/main/java/com/wynntils/wc/utils/SpellManager.java
```diff
@@ -0,0 +1,83 @@
...
+
+        SpellUnit[] spell = new SpellUnit[size];
+        for (int i = 0; i < size; ++i) {
+            spell[i] = spellMatcher.group(i + 1).charAt(0) == 'R' ? SpellUnit.RIGHT : SpellUnit.LEFT;
+        }
+
+        return spell;
+    }
+
+    public static void tryUpdateSpell(String text) {
```
Incompleteusern: `tryUpdateSpell` is called from both `SubtitleSetTextEvent` and `ActionBarManager`, is this intentional?

TomaSajt: This was intentional, however I am not really happy with this being outside of `SpellManager`. This was to avoid putting another actionbar parsing thing outside of `ActionBarManager`.

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
magicus: Can't you use the same enum for both? I'm thinking the only enum you need is PRIMARY/SECONDARY, all the way through, except at the very end, where you decide if primary is SwingPacket or UseItemPacket, depending on if you are an Archer or not.

magicus:
The more I think about it, the less certain I am that the split between SpellManager and QuickCastFeature is correct.

I think the functionality of both classes belong just in the QuickCastFeature. The point is, if we disable the QuickCastFeature, there is no need ti have SpellManager running and doing stuff each tick. That kind of defeats the purpose of having features that can be turned off.

If, in the future, we find that spell parsing is needed by more than one Feature, we can find a way to extract common functionality then.

magicus:
However, I understand if QuickCastFeature can become quite long. I think we should figure out a systematic way to handle features that needs (or benefits from) to be split in multiple classes.
Perhaps a package/class like `com.wynntils.features.quickcast.SpellManager`? (That is, create a package name based on the Feature main class.)

magicus:
At the very least, it should be clear that to implement the QuickCast feature we need two different, but interconnected parts: One part that sends the actual packets, and one part that determines the current spell progress. (`SpellSend` and `SpellDetect` or whatever.)

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
TomaSajt: The main problem is that the ingame spell gui doesn't distinguish between an L as an archer or an L as a mage, the parser is gonna think see the same, even though one is primary and one is secondary. I could do the class detection while parsing the spell, but there it is not guaranteed that the player is still holding the weapon. Since auto-class-join class detection still isn't implemented Character can't be used to fetch the class reliably.

magicus: Hm. Okay, let's keep it at this for the time being. I hope we can return to this later, if we have proper class detection implemented.

Or maybe we should not consider like "LRL" and "RLR" to be the "same" spell, and actually try to keep L and R properly separated. 

I dunno. I'm not 100% happy with the current solution but I don't have anything better to offer.

common/src/main/java/com/wynntils/utils/StringUtils.java
```diff
@@ -65,4 +65,13 @@ public static String[] wrapTextBySize(String s, int maxPixels) {
 
         return result.toString().split("\n");
     }
+
+    /**
+     * Creates a new pattern, but replaces all occurrences of '§' in the regex with an expression for detecting any color code
+     * @param regex - the expression to be modified and compiled
+     * @return a Pattern with the modified regex
+     */
+    public static Pattern compileCCRegex(String regex) {
```
TomaSajt, line 74: I don't really know where to put this method. Also, not sure about the name either.

magicus: That was a clever idea to get simple but proper matching for color codes. Did you come up with it or was it in legacy? Also, it's a good thing to move it to StringUtils, I do think it belongs here.

Maybe you should strip it down to the bare str.replace() part, and leave the Pattern.compile to the caller. Then you can name it like `replaceColorCode`. I think that would do fine.

common/src/main/java/com/wynntils/wc/utils/ActionBarManager.java
```diff
@@ -41,7 +43,7 @@ public static void onActionBarUpdate(ChatReceivedEvent e) {
         maxMana = Integer.parseInt(matcher.group(5));
 
         String centerText = matcher.group(3);
```
magicus: To do this really properly, I think we should have the ActionBarManager send out a wc event, like `ActionBarMessageUpdateEvent` with the group(3) content, and have the feature listen for it. That way we can make sure no "manager" depends on a "feature".

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
```diff
@@ -0,0 +1,224 @@
...
+    private final KeyHolder castFourthSpell =
+            new KeyHolder("Cast 4th Spell", GLFW.GLFW_KEY_V, "Wynntils", true, QuickCastFeature::castFourthSpell);
+
+    private static final Pattern SPELL_PATTERN =
+            StringUtils.compileCCRegex("§([LR]|Right|Left)§-§([LR?]|Right|Left)§-§([LR?]|Right|Left)§");
+    private static final Pattern INCORRECT_CLASS_PATTERN = StringUtils.compileCCRegex("§✖§ Class Req: (.+)");
+    private static final Pattern LVL_MIN_NOT_REACHED_PATTERN = StringUtils.compileCCRegex("§✖§ (.+) Min: ([0-9]+)");
+
+    public static final SpellDirection[] NO_SPELL = new SpellDirection[0];
+    private static SpellDirection[] partialSpell = NO_SPELL;
```
magicus: Could you either add a comment like "What we know about the current spell in progress", or perhaps rename it "spellInProgress"? Otherwise the purpose of this was a bit vague to me.

common/src/main/java/com/wynntils/utils/StringUtils.java
TomaSajt: The idea was from legacy code.
The main reason was for ccregex creation to be shorter, so spotless formatting won't make it a 2-liner. I'm not a big fan of the replace-only idea, as this method doesn't really make sense outside of regex creation. I'll think about it.

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
TomaSajt: That's a much bettern name, I'll change it when I get to a computer.

common/src/main/java/com/wynntils/wc/utils/ActionBarManager.java
TomaSajt: I was also thinking about something like this but didn't implement it. It is done now.

magicus (approved):
Looks good now. I guess we'll need to take this for a spin, and fix the wrinkles as we go. The legacy variant has had tons of incremental fixes and improvements, to make it more robust in real-world scenarios. It is unlikely we get this right the first time.

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
```diff
@@ -0,0 +1,230 @@
...
+
+        for (String lore : loreLines) {
+            Matcher matcher = LVL_MIN_NOT_REACHED_PATTERN.matcher(lore);
+            if (!matcher.matches()) continue;
+            sendCancelReason(new TranslatableComponent(
+                    "feature.wynntils.quickCast.levelRequirementNotReached", matcher.group(1), matcher.group(2)));
+            return;
+        }
+
+        boolean isSpellInverted = isArcher;
```
kristofbolyai, line 164: This boolean is kinda duplicated, perhaps just rename the first one?

kristofbolyai (approved):
Looks good! Thanks for adding this!

common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
TomaSajt: It's because lamdas can only use final or effectively final variables.

PR #181, opened 2022-07-28
Make for loop out of duplicated code in QuestInfoOverlayFeature

magicus:


magicus:
I am doing something wrong with the I18n business, but I don't understand what or why. :-( The texts do not get translated, they only show up as the key. @kristofbolyai Any idea?

magicus:
Is maybe translations not available at time of Overlay instantiation? But that sounds strange as well, that should not happen until quite late, or..?

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
```diff
@@ -20,16 +20,30 @@
...
     @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
     private final Overlay questInfoOverlay = new QuestInfoOverlay();
 
     public static class QuestInfoOverlay extends Overlay {
+        private static final List<CustomColor> TEXT_COLORS =
+                List.of(CommonColors.GREEN, CommonColors.ORANGE, CommonColors.WHITE);
+
+        private final List<TextRenderTask> toRender = createRenderTaskList("", "");
+        private final List<TextRenderTask> toRenderPreview = createRenderTaskList(
+                I18n.get("feature.wynntils.questInfoOverlay.overlay.testQuestName") + ":",
```
kristofbolyai: I am not sure if I18n is initialized when this is created

kristofbolyai: Is this the only one not working, or the other one as well?

magicus: Do you know at what point I18n is initialized? I'm guessing this is a Minecraft vanilla thing. Can we delay creation of Overlays until after that point? I'm thinking this will not be the only Overlay that will want to have string literals, and it would be nice if we didn't have to jump through hoops to get it.

kristofbolyai: I have no idea, but this is strange overall. I can test it in the evening.

magicus: Neither of the two I18n literals are working. 

It might also be some environment issue here on my side. I know I have had problems with resources getting placed correctly by IntelliJ. But otoh new textures seem to appear fine, so I don't really think so.

kristofbolyai: Try building the mod and see the localization file in there?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -175,6 +175,8 @@
   "feature.wynntils.questInfoOverlay.name": "Quest Status Overlay Feature",
   "feature.wynntils.questInfoOverlay.overlay.questInfo.description": "Display the currently tracked quest's info",
   "feature.wynntils.questInfoOverlay.overlay.questInfo.name": "Quest Info Tracker",
+  "feature.wynntils.questInfoOverlay.overlay.testQuestName": "Test quest",
```
kristofbolyai: Btw this should be `feature.wynntils.questInfoOverlay.overlay.questInfo.testQuestName` to match our format.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -175,6 +175,8 @@
...
+  "feature.wynntils.questInfoOverlay.overlay.title": "Tracked Quest Info",
```
kristofbolyai: Same as above

common/src/main/java/com/wynntils/features/user/QuestInfoOverlayFeature.java
magicus: I have verified that the strings are in the jar file, so it is not a resource building problem.

I searched the web to try to dig up some "later" init hook that could be called after Fabric's I18n is in place, but failed to find any useful info. :( I've just dropped a question at the Fabric modders discord.

magicus: Recommendations from the Fabric discord was to either:
1) do translation lookup at render time (I'm not to fond of this), or
2) use a custom resource reload listener, to get a hook back to after i18n has loaded (https://fabricmc.net/wiki/tutorial:custom_resources). 

I'm inclined towards going down route #2, and make a way to do our main init stuff in an "after-resources-are-loaded-init". But I'm not sure what the implications of this is, really.

For the time being, I'm thinking of reverting the I18n bits of this PR to get the rest pushed, and then implement the init-with-resources-loaded, and then getting back to fixing I18n here.

magicus: (The reload listener needs some additional checks to make sure we only call it once and not on manual resource reloads)

kristofbolyai:
As coehlrich suggested, maybe try using TranslatabeComponent for the preview and make it a string in render?

magicus:
> As coehlrich suggested, maybe try using TranslatabeComponent for the preview and make it a string in render?

Yes, that is basically option 1 i noted above.

The reason I'm not really fond of this was that it surprised me that I18n was not available, and I think we might ran into similar issues in the future. As always, one of the goals of Artemis is to make writing new Features as painless as possible. And I think having I18n available at run-time but not init-time is a WTF-factor that makes it unnecessarily hard.

magicus:
I give up on the resource reload listener. :-( I can't get it to work, and I don't understand why. We'll just have to keep from using I18n in Feature initialization...

kristofbolyai (approved):

PR #183, opened 2022-07-29
TeleportScrollProperty: Fix issues with teleport scroll text property

### LINKED ISSUE #169: Test Result of Item Tier / Scroll Letters
Success:
- scrolls have the town letter
- powders have their tier

Fail:
- Town pairs like Nesaak and Nemract get same letter
- boat passes don't have letters
- broken keys where the Forgery note is missing don't have letters
- old dungeon keys don't have letters
- emerald pouches don't have their tier

kristofbolyai:
Fixes https://github.com/Wynntils/Artemis/issues/169#issuecomment-1189971952.

This also adds 2 char text overlay, instead of 1 char.

New look: 
image


magicus:
As @DonkeyBlaster says in #169:
"Well first 2 would not solve the problem in the first place - NEmract and NEsaak"

Maybe we should have a list of hard-coded exceptions, like "Nemract" => "Nm", "Nesaak" => "Ns"? There can't be that many collisions.

magicus:
Or maybe a curated list of abbreviations for all destinations? If there is no other place than Detlas starting with "D", then "D" should be enough (and "De" overkill).

magicus (approved):

kristofbolyai:
If someone wants to go ahead and add those alias overwrites, go ahead, but I prefer to merge this as it was only intended as a bug fix.

PR #186, opened 2022-07-29
Test PR to fix GitHub actions

kristofbolyai:


kristofbolyai:
Well, seems like we cannot do anything without admin privs to fix this.

PR #188, opened 2022-07-30
ChatManager: Split out NPC dialogs, determine message recipient group and filter on info messages

magicus:
I'll restart without the GAME_INFO changes, since these collided with stuff already done by TomaSajt.


magicus:
I think this is about ready to merge now.

There are much more that can be done, but I'm happy for now.

The main functionality is that it tries to "counteract" the new Wynntils NPC dialog system, so it splits out the NPC dialog and tries to restore the normal chat. My idea here is that we should have an overlay presenting the chat dialog; but for now the NPC dialog is just posted (once) to the chat.

As a DebugFeature, I also tag all chat messages with the "RecipientType". The idea here is that we should be able to separate between party, guild, global, etc chats.

Finally, there is a skeleton implementation for filtering info messages (everything that is not a chat written by a human in the other end). This too needs to be extended, and many of the filtered messages should probably be sent as Notifications instead.

(I'm thinking that the `@Config` variables in that Feature should really be a tri-state, not a boolean. Like `enum HandleInfoMsg { KEEP, HIDE, REDIRECT_TO_NOTIFICATION }` or something like that.

magicus:
Some more comments about this. The problem here is the way Wynntils has implemented the new NPC dialog system. They essentially rewrite the entire chat history with a single chat message containing lots and lots of lines, ending with the NPC dialog. The actual chat messages are greyed out. Any new chat messages appearing during this time will appear in the greyed out (what I called MessageType.BACKGROUND) section. I look for these changes, and send them out as individual chat lines.

The benefit of doing this is not only that we can move the NPC dialog elsewhere, but also that message filtering can be done with messages recieved during an NPC dialog. Otherwise that is an impossible project.

What we lose, however, is the proper color formatting of chat messages. What Wynncraft does is that once the NPC dialog is over, they redraw once more the entire chat history, but this time without the NPC messages, *and* the actual chat messages having restored colors.

But this idea of modifying the colors of chat messages, long after they actually were received, could not possibly be implemented in Wynntils in a sane way. But this also lead to a kind of regression; when not using Wynntils, chat messages received during a NPC dialog will, eventually, get proper colors without Wynntils, but not with Wynntils. I'm prepared to live with that, but you ought to know about that limitation.

magicus:
This is also the reason for the double set of pattern matches (normal and background).

I'd also like to add a feature that is *gain* by this Wynntils treatment: All onHover/onClick functionality on the Components are lost when Wynntils redraws the entire chat history. So if you have a message like "Click here to join the party", and before clicking there, goes into an NPC dialog, when you leave that dialog, the clickability of the chat message is gone. This will not be the case with Wynntils multiline splitting.

Nevertheless, I tought the multiline splitting sounded a bit scary, so it is off by default, and only enabled if someone turns it on in ChatManager. (For a devel build, this will happen due to the debug feature). Reasons for turning this on can be both to get NPC dialogs into an overlay, and to be able to get proper filtering of chat messages even while in an NPC chat (otherwise messages received during a dialog can not be properly processed). We'll have to think a bit on how to present that to the user.

magicus:
@kristofbolyai @P0keDev Any chance you can review this? I have a couple of dependent changes planned/ongoing, like a NPC dialog overlay and porting the translation service from legacy, which depends on this PR.

kristofbolyai:
> @kristofbolyai @P0keDev Any chance you can review this? I have a couple of dependent changes planned/ongoing, like a NPC dialog overlay and porting the translation service from legacy, which depends on this PR.

When I get home, eta 1-2 hours max.

magicus:
Thanks! I suspect it might take a few iterations to get this into ship-shape, so I'd like to get that process started.

One thing that could be controversial (and that I can remove/disable), is the default enabled on the testing feature, this tags all chat messages by the type it detected (normal/background and recipient group (party, guild, etc)). I have also enabled that NPC messages are sent out just once in chat (like the old-style Wynncraft behavior). I'd prefer to keep this on by default for a while, to see how well the detection works, but if this is too controversial, we can turn it off by default and I'll turn it on myself.

kristofbolyai:
First observarion: 
image

We kinda lost the "shift to progress" in dialogue message. Can we add it back somehow?

common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,221 @@
...
+                        dialog.push(line);
+                    }
+                } else {
+                    newChatLines.push(line);
+                }
+            }
+        } else {
+            // After a NPC dialog screen, Wynncraft sends a "clear screen" with line of ÀÀÀ...
+            // We just ignore that part
+            if (trailingLine.matches("À+")) {
```
kristofbolyai, line 170: There was a bug in Wynntils, where users could send a long string of these characters (and fool the mod into an incorrect state)? Is this present here?

common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private static RecipientType getRecipientType(Component message, MessageType messageType) {
+        String msg = ComponentUtils.getCoded(message);
+        if (messageType == MessageType.SYSTEM) {
+            // System type messages can only be shouts or "info" messages
+            // We call this MessageType.NORMAL anyway...
+            if (RecipientType.SHOUT.matchPattern(msg, MessageType.NORMAL)) {
+                return RecipientType.SHOUT;
+            }
+            return RecipientType.INFO;
```
kristofbolyai: Qodana suggestion: extract this return (if there was not a specific reason not to do so.)

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private static final Pattern BACKGROUND_SOUL_POINT_1 =
+            Pattern.compile("^(§r§8)?As the sun rises, you feel a little bit safer...$");
+    private static final Pattern BACKGROUND_SOUL_POINT_2 = Pattern.compile("^§r§7\\[(\\+\\d+ Soul Points?)\\]$");
+
+    private static final Pattern BACKGROUND_LEVEL_UP_1 = Pattern.compile("^(§r§8)?.* is now level .* in §.*$");
+    private static final Pattern BACKGROUND_LEVEL_UP_2 =
+            Pattern.compile("^(§r§8)?\\[!\\] Congratulations to §r.* for reaching (combat )?§r§7level .*!$");
+
+    private static final Pattern BACKGROUND_VIP_LOGIN =
```
kristofbolyai: I assume this works for every rank? Perhaps rename this?

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -222,4 +222,8 @@ public static List<Component> splitComponent(Component component, int maxWidth)
         if (split.isEmpty()) split.add(new TextComponent(""));
         return split;
     }
+
+    public static String stripFormattingCodes(String msg) {
```
kristofbolyai: Why is this better than `ChatFormatting#stripFormattingCodes`?

kristofbolyai (commented):
This is my first run through. but I am tired so I will take a look at tomorrow too.

magicus:
> We kinda lost the "shift to progress" in dialogue message. Can we add it back somehow?

Yes, and no. You can get back the entire Wynncraft "NPC screen" thingie (where the entire chat history is sent like once a second), by `ChatManager.disableNpcDialogExtraction()`. This is in fact the default, and the only reason you are seeing this removed is since `ChatManagerTestFeature` calls `ChatManager.enableNpcDialogExtraction()` in its `onEnable`. Disable that feature *) and you will see the old behavior.

But. 

That also means that we will not be able to process any chat messages that arrive while being in the NPC dialog. That part only works because I effectively disable this "screen-wise" dialog system, cancelling all the multiline chat "screens".

So a better way forward is to take the extracted dialog and present it differently to the user, in terms of an overlay. I indent to do that as a follow-up. But in the mean time, I just send out the extracted NPC messages to the chat. At that point, the ChatManagerTestFeature serves no purpose, and instead the enable/disable of npc dialog extraction in ChatManager will depend on wether you activate the overlay or not.

*) I just realize that I should turn off this in a `onDisable` as well, so you won't need to restart the mod for the disable to take effect.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
magicus: I've tried my darndest to get this foolproof. The `RecipientType` regexps are all matching the whole line (surrounded by `^` and `$`), and should catch properly what kind of message this is (a shout, system announcement etc). 

In this particular case, you're looking at the code which interprets the multiline chat "screens" sent during NPC dialog. The `newLines` list contains all the last few lines in the chat screen, which were not in place when the last time a "normal" one line chat message appeared. I can't see how a user can manipulate this.

My entire assumption here is that a user cannot send a `\n` (newline) in a single chat message. It can't be done in normal vanilla chat, and if someone tries that with some kind of mod, Wynncraft really *should* filter it out. If they don't, well, then someone can write a message that looks like this:
```
[12/33/Ar] MyRealName: hello \n
[105/333/Sh] Salted: This is a fake message really sent by MyRealName
```
with proper formatting codes etc. If they do, it would be indistinguishable from a real message from Salted. There is no way either a normal vanilla user, nor Wynntils, would be able to separate that from a real message.

I do belive my code is robust enough (and stateless enough) to not be able to be tricked into some kind of incorrect state, though.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
magicus: I'm not sure what Qodana is talking about here. There is no common part? Yes, SHOUT is one member of RecipientType.values(), but seriously, how could you extract that meaningfully?

I claim Qodana is just being stupid right now. If you see a way to simplify this code, please explain to me in better terms than Qodana is doing...

kristofbolyai: It's about the return RecipientType.INFO. That is common. Anyways this is just a nit.

magicus: Ah, I see, it wants to extract the return INFO.

Well, you could do that. I just think this captures the semantics better. Initially, I had SYSTEM always return INFO, until I discovered that SHOUTS where (sometimes? or possibly always?) coded as SYSTEM, so I had to put in a special case for that.

So the semantics is like:
* for system, the default case is "info" unless it is a shout
* for normal, the default case is that it's a chat message. if it's not, the fallback is that it's "info"

But if we can't silence Qodana any other way, I can rearrange this.

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
magicus: Yes, agree.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
kristofbolyai: Ah, it is alright then. Qodana should only report problems with added code (although this might not be the case atm, perhaps it is buggy)

P0keDev:
> But this idea of modifying the colors of chat messages, long after they actually were received, could not possibly be implemented in Wynntils in a sane way. 

i know that in legacy, we modified chat messages after they were added to chat using an id system, but maybe that was a forge functionality?

kristofbolyai:
> > But this idea of modifying the colors of chat messages, long after they actually were received, could not possibly be implemented in Wynntils in a sane way. 
> 
> i know that in legacy, we modified chat messages after they were added to chat using an id system, but maybe that was a forge functionality? 

There is a similar thing, check out NotificationsFeature.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
magicus: Well, it's not, it's actually slightly worse. 😄 I just didn't know about it. I looked through our Utils class but did not think Vanilla would have anything like that...

magicus:
> > But this idea of modifying the colors of chat messages, long after they actually were received, could not possibly be implemented in Wynntils in a sane way.
> 
> i know that in legacy, we modified chat messages after they were added to chat using an id system, but maybe that was a forge functionality?

Ok, so, yes, we could do something like this:
 * When in an NPC screen, send out new chat message like now, in background mode, but tag each message with a unique id
 * When the NPC dialog is finished, and the last multiline is sent (which includes color information), re-parse the new lines that had been sent out as background mode, and associate each line with the corresponding unique id
 * For each of these, send out a ChatModifiedAfterBeingSentEvent with the unique id and the new color-coded string
 * Have chat listeners do something relevant with this. For instance, a new Feature which implements chat tabs like in legacy, would need to find and update the corresponding chat lines in their per-tab history.

It is *possible*. I just did not consider it a "sane way".

magicus:
Can we, at the very least, save that for the future as an upgrade to the functionality? ("Enhancement: restore color to messages sent during NPC dialogs" or something like that)

magicus:
Let me also be clear that this "color restoration screen" is only partially restoring what would have been, had the user not been in a NPC dialog. A normal chat line is sent as a quite complex Component, with hover boxes for a lot of things (like user and world info). These are not restored; since the NPC "screens" are not sent as Components, but instead as a single string with formatting codes.

So we can restore the chat messages to look like they would have been, if we had not been in an NPC dialog (like how Wynntils does themselves), but we can never recover how they *really* should have looked if we had not been in a dialog. That's a limitation set by Wynncraft.

magicus:
Let me also add that to do be able to use this "updated" chat messages, we *need* to implement something like the tabbed chat pages, even if we support only one page, since we need to have a history of the latest sent chat messages, and be able to clear the screen from chat messages and redraw them with the updated colors. (Just like Wynn does).

Hm. Hmmm... We could possibly let the last "screen" through, with the updated colors. The drawback of this is that messages that we have filtered away (like "You feel much safer.... +1 soul point") will suddently (re-)appear. I think that'd be a worse user experience.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,222 @@
...
+            if (updatedMessage == null) {
+                e.setCanceled(true);
+            } else if (!updatedMessage.equals(message)) {
+                e.setMessage(updatedMessage);
+            }
+            return;
+        }
+
+        if (extractDialog) {
+            handleMultilineMessage(codedMessage);
```
kristofbolyai, line 104: This method only handles NPC chat stuff? Is there any more messages with new lines that is not an npc dialogue?

common/src/main/java/com/wynntils/core/chat/ChatManager.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /**
+     * Return a "massaged" version of the message, or null if we should cancel the
+     * message entirely.
+     */
+    private static Component handleChatLine(Component message, String codedMessage, MessageType messageType) {
+        RecipientType recipientType = getRecipientType(message, messageType);
+
+        System.out.println("Handling chat: " + ComponentUtils.getCoded(message) + ", type:" + messageType
```
kristofbolyai: println in non debug feature (or anywhere at all)?

kristofbolyai (commented):
Seems okay to me otherwise

common/src/main/java/com/wynntils/core/chat/ChatManager.java
magicus: This handles the NPC dialog scenario. But when that happens, there are a "pre screen" and "post screen" that does the same thing, but does not contain any of the NPC lines. Kind of a "cls" thing.

I have not seen this system being used for anything else than NPC dialogs. I could not even imagine what that would be. The one thing I know exists but haven't tested it with, is the multi option selections things. I only saw them in Winter Detlas; I don't know if there are any such present in the world today. I imagine the code will work mostly fine anyway, but the dialog itself may be multi-line, which I currently chose not to handle until I see evidence that such a thing exists.

As I said before, I hope Wynncraft prohibits users from posting `\n` to the chat, otherwise all bets are off.

common/src/main/java/com/wynntils/core/chat/ChatManager.java
magicus: Oops. That should not be there.

magicus:
@kristofbolyai Can we tweak Qodana to stop complaining about `Redundant character escape `\\]` in RegExp`? That's the most annoying warning ever. The technical rules of regexp says that in some cases the `\\` can be removed, but that is not trivial to know when, and it breaks symmetry and actually makes the regexp harder to read, and more prone to errors if you modify it.

kristofbolyai:
> @kristofbolyai Can we tweak Qodana to stop complaining about `Redundant character escape `\\]` in RegExp`? That's the most annoying warning ever. The technical rules of regexp says that in some cases the `\\` can be removed, but that is not trivial to know when, and it breaks symmetry and actually makes the regexp harder to read, and more prone to errors if you modify it.

I will open a PR this evening, please remind me if I don't.

magicus:
@kristofbolyai I opened https://github.com/Wynntils/Artemis/pull/197 to handle Qodana. I also chose to silence the UnnecessaryReturn. I can see how it might have benefits, but I can also see how we will have many list of test-with-return statements like the one in the filter feature, where it will just be annoying.

magicus:
Edit: This came out of order. It's fixed in the Silence Qodana PR.

I'm getting more and more annoyed at Qodana. I mean, how could anybody ever think this would be an appropraite warning: `2 consecutive spaces in RegExp` ??? Jeez. Of course there are two spaces, since we want to match with two spaces. That is just plain stupid.

kristofbolyai:
> Edit: This came out of order. It's fixed in the Silence Qodana PR.
> 
> 
> 
> I'm getting more and more annoyed at Qodana. I mean, how could anybody ever think this would be an appropraite warning: `2 consecutive spaces in RegExp` ??? Jeez. Of course there are two spaces, since we want to match with two spaces. That is just plain stupid.

It is complaining about not using %s%s so it is more visible or smth. IDEA does this too (well Qodana is IDEA backend but yeah)

magicus:
@kristofbolyai Was this OK?

kristofbolyai:
> @kristofbolyai Was this OK?

I mean it is okay, I don't mind either way, I am just saying there is logic to the warning (although it works either way and is just visual).

magicus:
Yeah, sorry, I meant this PR.

kristofbolyai (approved):
Seems fine, though if you could add start disabled to the debug stuff it would be nice. Fine either way.

magicus:
I'd like to keep it on for a while, to verify that everything works smoothly. We can turn it off in a few days time or so.

PR #189, opened 2022-07-30
Add feature to hide Vanilla info bar (to avoid duplicated mana/health display)

magicus:
Add feature to hide Vanilla info bar (to avoid duplicated mana/health display)

We could consider making this part of the Mana/Health overlay, but it also sort of orthogonal do that.


kristofbolyai:
Wouldn't this fit into the specific overlay's feature class?

I am planning to do this with quest info (hide from scoreboard), and I thought I would just add it into the feature itself. What do you think?

magicus:
We can fold it into the mana/health Overlay feature, but to do this properly we should probably then only filter the relevant parts of the info bar. That is, if you have the Mana overlay but not the health overlay, we should remove the health part of the info bar (presumably replacing it with spaces to keep alignment), and vice versa. 

This PR took a more heavy-handed approach, by just disabling the entire info bar.

We don't have to integrate this. It was a spin-off from my attempt to split up the work in my former ChatManager branch into different branches depending on what it was doing, since the info bar changes overlapped with TomaSajt's work.

magicus:
But yes, it might be better to have a way for the ActionBarManager to blank out the separate parts (health, message and mana) individually, and cancel the entire thing if it's just blanks all the way.

kristofbolyai:
So, just to confirm, we are going with the more refined option? If so, do you plan on doing that in this PR, or what is your plan?

magicus:
Yes, we're going with the more refined option. I can try and to it in this PR unless someone beats me to it. (I'm rapidly running out of time available to do Wynntils, and my focus now is getting the ChatManager finished)

kristofbolyai:
We should not need this anymore.

PR #190, opened 2022-07-30
CI: Add Qodana static code analysis

### LINKED ISSUE #187: Qodana code report
Qodana code report should be a much better replacement for legacy's Codefactor. 

What do you think? @magicus @P0keDev @Incompleteusern 

Example PR/Report: https://github.com/kristofbolyai/Artemis/pull/1#issuecomment-1200123934

I have tried to embed the report into the comment, but I only could manage to send the link. It should be fine.

kristofbolyai:
This PR is adding #187.


kristofbolyai:
Also, if anyone is interested, we need to commit `.idea/misc.xml` to specify JDK version for Qodana. It is a bit eh, but we can live with it I guess.

magicus:
> Also, if anyone is interested, we need to commit `.idea/misc.xml` to specify JDK version for Qodana. It is a bit eh, but we can live with it I guess.

That's close to a no-no. :-( I have personally set up a special JDK 17 with hotswapping for Artemis (as opposed to my normal JDK 17), and this is determined by the special name I've given it, which is recorded in .idea/misc.xml.

magicus:
Do you mean Qodona as run on Github needs a .idea/misc.xml file? That sounds really odd. There must reasonably be some other way to specify JDK version to Qodona.

kristofbolyai:
> Do you mean Qodona as run on Github needs a .idea/misc.xml file? That sounds really odd. There must reasonably be some other way to specify JDK version to Qodona.

It needs the file to specify JDK version. Their GH discussions page suggested this solution.

magicus:
I see. But they also recommend that it can be created in a prior step in the github action file. I think that is preferable. They post a suggested content of the file. I'd be fine with us pushing like `.idea/misc-template.xml` and copy that file during the github run. It could also serve as a template for new users, with the annotation classes as entry points (to stop IntelliJ from marking properly annotated methods as unused).

kristofbolyai:
@magicus Fixed the `misc.xml` problem. We now have `misc-example.xml` and rename it.

magicus (approved):

PR #191, opened 2022-07-30
ScoreboardManager: Add a way to disable scoreboard segments within features

kristofbolyai:
Example implementation of this capability is `QuestInfoOverlayFeature`. 

This PR `ScoreboardManager` bug fixes. Now, the manager is so precise, it can be used to reconstruct the whole scoreboard, without the user even noticing.

### DIFF
--- a/common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
+++ b/common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
@@ -8,8 +8,10 @@ import com.wynntils.core.WynntilsMod;
 import com.wynntils.mc.event.ScoreboardSetScoreEvent;
 import com.wynntils.mc.utils.McUtils;
 import com.wynntils.utils.objects.Pair;
+import com.wynntils.wc.event.ScoreboardSegmentAdditionEvent;
 import com.wynntils.wc.event.WorldStateEvent;
 import com.wynntils.wc.model.WorldState;
+import com.wynntils.wc.utils.WynnUtils;
 import com.wynntils.wc.utils.scoreboard.objectives.ObjectiveManager;
 import com.wynntils.wc.utils.scoreboard.quests.QuestManager;
 import java.util.ArrayList;
@@ -17,6 +19,7 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -140,8 +155,53 @@ public class ScoreboardManager {
                 }
             }
         }
+
+        handleScoreboardReconstruction();
     };
 
+    private static void handleScoreboardReconstruction() {
```
magicus, line 162: How would this work if the user logs on to another server? I think one of the (less important) design goals of Artemis is that the mod should be "transparent" on other servers, so you should in principle be able to have it always enabled, even if playing on other servers.

Is there some logic that skips this process if not on Wynncraft?

magicus: Found it now in onSetScore

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -193,6 +193,8 @@
   "feature.wynntils.pouchRedirectFeature.redirectEmeraldPouch.name": "Redirect Emerald Pouch Messages",
   "feature.wynntils.pouchRedirectFeature.redirectIngredientPouch.description": "Should messages about ingredients being added to your pouch be redirected to the game update ticker?",
   "feature.wynntils.pouchRedirectFeature.redirectIngredientPouch.name": "Redirect Ingredient Pouch Messages",
+  "feature.wynntils.questInfoOverlay.disableQuestTrackingOnScoreboard.description": "Should quest tracking be enabled on the scoreboard? This option only works if QuestInfoOverlay is turned on.",
```
```suggestion
  "feature.wynntils.questInfoOverlay.disableQuestTrackingOnScoreboard.description": "Should quest tracking be removed from the scoreboard? This option only works if QuestInfoOverlay is turned on.",
```

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -140,8 +155,53 @@ public class ScoreboardManager {
...
+        Scoreboard scoreboard = McUtils.player().getScoreboard();
+
+        List<String> skipped = new ArrayList<>();
+
+        for (Segment parsedSegment : segments) {
+            boolean cancelled = WynntilsMod.getEventBus().post(new ScoreboardSegmentAdditionEvent(parsedSegment));
+
+            if (cancelled) {
+                skipped.addAll(parsedSegment.getScoreboardLines());
```
magicus, line 171: I'm thinking that if all segments are removed, maybe the entire scoreboard should be hidden..?

kristofbolyai: It is hidden that case.

kristofbolyai: The reason you don't see code for that is, no matter if we have a display scoreboard, if there are no scores, Minecraft does not display it.

magicus: What about the top heading "play.wynncraft.com"?

kristofbolyai: That is the scoreboard name, if no "players" have score, it doesn't display.

magicus: Ok, fine!

magicus (approved):

PR #192, opened 2022-07-31
Fix remaining annoying Vanilla log output

magicus:
Fix remaining annoying Vanilla log output.

First step:
* Fix "Received passengers for unknown entity"

### DIFF
--- a/common/src/main/java/com/wynntils/features/internal/FixPacketBugsFeature.java
+++ b/common/src/main/java/com/wynntils/features/internal/FixPacketBugsFeature.java
@@ -5,6 +5,7 @@
 package com.wynntils.features.internal;
 
 import com.wynntils.core.features.InternalFeature;
+import com.wynntils.mc.event.AddEntityLookupEvent;
 import com.wynntils.mc.event.BossHealthUpdateEvent;
 import com.wynntils.mc.event.RemovePlayerFromTeamEvent;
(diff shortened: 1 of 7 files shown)


magicus:
Remaining to be fixed:
* "Duplicate entity UUID"
* Fix NPE:
```
 java.lang.NullPointerException: Cannot invoke "net.minecraft.client.gui.components.LerpingBossEvent.setName(net.minecraft.network.chat.Component)" because the return value of "java.util.Map.get(Object)" is null
	at net.minecraft.client.gui.components.BossHealthOverlay$1.updateName(BossHealthOverlay.java:91) ~[minecraft-project-@fabric-merged-named.jar:?]
```
* "Received packet for unknown team X: team action: REMOVE, player action: null"

magicus:
I have not yet been able to fix the "Duplicate entity UUID" issue. This seem to happen when tracking horses, at world login a horse is somehow being glitchily updated.

I've tracked the most likely cause of the error to be in PersistentEntitySectionManager, where updateStatus calls startTracking which adds the entity when it is already added.

The check before seems to be that the entity should be moving from a non-visible to a visible "entity section" (like a chunk but for entities, I guess). I'm guessing something is a bit glitchy there.

magicus:
Now all buggy Wynncraft warning spam seems to be gone.

kristofbolyai (approved):

PR #196, opened 2022-08-01
Port Translation from legacy

magicus:


magicus:
NPC dialogue and chat (info messages, and/or player chat) can now be translated. Quest tracking is not implemented yet; that will happen in another PR.

If we get the "update post factum" thing on chat messages that we discussed wrt background encoding, we can use that as well to provide a translation. It would be nice. (The problem is that translation is asynchronous and potentially really slow in comparison, so we can't wait on it). But it will also be a huge new feature that is not present in legacy, so maybe it can wait until we are closer to parity.

magicus:
I know there are some stuff in the translation "engine" part that is not really used right now. I'll look over it at a second run, including getting proper analysis of "prefix" of chat messages, so we don't try to translate user names (we do right now with this PR, sometimes it looks hilarious :-)).

kristofbolyai (commented):
Seems okay, I will give it another pass tomorrow.

common/src/main/java/com/wynntils/features/user/TranslationFeature.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.network.chat.TextComponent;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@StartDisabled
+public class TranslationFeature extends UserFeature {
+    public static TranslationFeature INSTANCE;
+
+    @Config
+    public String languageName = "";
```
kristofbolyai, line 26: Is this fine as empty? Perhaps make it null?

magicus: It is fine as empty. In general, my principle is to prefer the empty string before null for strings, and to prefer the empty set/list/map before null for set/list/map. 

But maybe I should put in an explicit check so I don't rely on the translation API to return an error. (Or maybe I already do, don't remember; that code is stolen directly from Legacy)

common/src/main/java/com/wynntils/core/services/CachingTranslationService.java
```diff
@@ -0,0 +1,91 @@
...
+
+        translateNew(message, toLanguage, handleTranslation);
+    }
+
+    public static synchronized void saveTranslationCache() {
+        try {
+            if (translationCaches == null) return;
+
+            File f = new File(TRANSLATION_CACHE_ROOT, "translations.json");
+            Gson gson = new Gson();
```
kristofbolyai: Can we cache this Gson instance? (and use it in load)

kristofbolyai:
I think I remember debugging a log in legacy, which led to this code. Not sure if I fixed it though...

magicus:
> I think I remember debugging a log in legacy, which led to this code. Not sure if I fixed it though...

Like in someone posted a stack dump, and it included some translation stuff in general? Or was there some specific code you thought was bad?

kristofbolyai:
Yeah, it was a ticket containing a log. I don't really remember now so :shrug:

kristofbolyai:
Also, what happens if translation is not available? Do we get the English text?

kristofbolyai:
Also does this work with the overlay?

magicus:
The default behavior is to keep both original and add a translation. If the translation fails, you only get the original. If you have selected not to keep original, and translation fails, you get nothing. (I think; I have not yet seen translation fail).

Ideally, we should build in more support to have translations. Like if we get a replacement for the vanilla chat (with tabs etc). Then I'd like to have things like maybe putting original and translated chat/info side by side, or showing original when hovering etc.

The problem is that translation is a bit so-so, and sometimes you need to see the original as well to understand. For instance, if "Cinfras County Middle Upper" is translated to "Cinfas län medelöverst" you might need to see the original to understand what region is referred to.

For NPC dialogs, quest tracking prompts and system info, we are dealing with static strings, and can, at least in theory, plug in some kind of database with manually translated phrases. If we ever get there, we can perhaps get the kind of quality that an English parallel is not needed. But for user chat this will never happen. (Otoh, translating user chat is perhaps not that important)

So this PR is setting the framework for getting translations delivered, but I think we can work out a couple of iterations on how to present them in the best way.

kristofbolyai (approved):
I see, this should be fine then.

magicus:
NPC dialogue works perfectly with the overlay. Without it, I cannot translate since all I get is this huge "screen" every second.

The dialogue first appears in English, and then it is replaced as soon as the translation is done. Looks okay, for now (but see the discussion above on having better ways to handle this).

PR #197, opened 2022-08-02
Silence a few warnings from Qodana

magicus:
It was a bit tricky to determine the names of the inspections. I ended up downloading the report, looking through the files until I found the `inspectionName` key, and then grepped that. Don't know if there is a better way.

### DIFF
--- a/qodana.yaml
+++ b/qodana.yaml
@@ -7,3 +7,6 @@ include:
   - name: RedundantLabeledSwitchRuleCodeBlock
 exclude:
   - name: UnstableApiUsage
+  - name: RegExpRepeatedSpace
+  - name: RegExpRedundantEscape
+  - name: UnnecessaryReturn


kristofbolyai (approved):

magicus:
Argh, `maven.parchmentmc.org` is down, so the build does not succeed, and hence I cannot merge this. :-( Stupid systems. I think we need to allow main developers (me, you and P0ke) to override this restriction, not only @DevScyu.

kristofbolyai:
I think we can leave the double spaces and use the escaped version (%s). As for finding these tags, you can download Qodana report, use python http server host and open it, and you get a nice GUI where you can disable inspections and download the new qodana.yml.

magicus:
I've never even heard about having `%s` as an escape for space in regexp. It does not work on regex101.com, which is my standard go-to place for testing out regexps. I can't see any gain in changing real, working rexeps to some non-standard variant, just to make a stupid warning in Qodana keep quiet.

kristofbolyai:
Oh sorry, it is \s. https://www.w3schools.com/jsref/jsref_regexp_whitespace.asp

magicus:
Well, that is not the same. ` ` is just a single space character. `\s` matches all kinds of whitespace. I still claim this Qodona rule is dumb as f*ck.

PR #198, opened 2022-08-02
Fix spurious disconnect events, causing "onServer" checks to fail

magicus:
I believe this will solve #23.

The problem here was that the vanilla disconnect method were called even if the connection was not open (?). I can't say I fully understand why, most likely this is used for different kinds of connections, not only the "main" connection to the server. But this makes us at least avoid the disconnect event that typically happened shortly after logging in. (And that's the reason I've seen to explain the world state info bug.)


kristofbolyai (approved):
Seems like a valid fix, we will see.

PR #199, opened 2022-08-02
Let the buttons in the overlay management screen have priority over overlays

magicus:
Otherwise you can't press the buttons when an overlay is "under" them.


kristofbolyai (approved):
Thanks! Saw this issue but never got around fixing it.

magicus:
My default placement of the NPC dialogue is smack in the middle of the screen, so it became very apparent (and annoying). :-)

PR #200, opened 2022-08-02
Make NPC dialogue an overlay

magicus:
Also fixes bug in ChatManager and FontRenderer.


common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -133,12 +133,40 @@ public void renderText(
...
+                continue;
+            }
+            String thisCode = lastPart.substring(index, index + 2);
+            if (thisCode.charAt(1) == 'r') {
+                // it's a reset code, we can stop looking
+                break;
+            }
+            // prepend to codes since we're going backwards
+            lastPartCodes = thisCode + lastPartCodes;
+
```
magicus, line 163: You know what, Qodana? I'm starting to get really pissed off with you. There is no `StringBuild.prepend()`. This rule of yours should have checked the order of the arguments to the `+`, not just naively assume that we don't know how to use a StringBuilder.

kristofbolyai (commented):
This looks good. One thing I would like to confirm before I approve this though. At the moment, if we disable this feature (and the debug one), do we get the old, unmodified, Wynncraft style NPC messages (with no modification, including the "shift to continue", and the grayed out stuff)? 

If yes, this looks good!. If not, maybe we should add it in this PR. 

Also, I am not 100% sure about the overlay positioning, but I don't know where else to put it...

P0keDev:
could you include a screenshot of the overlay in action?

Incompleteusern:
image
from discord

magicus:
> At the moment, if we disable this feature (and the debug one), do we get the old, unmodified, Wynncraft style NPC messages (with no modification, including the "shift to continue", and the grayed out stuff)?

That is the intention, yes. I have not verified it, but it it does not work like you describe, I'd consider it a bug and get it fixed. In fact, come to think of it, I'm not sure the logic is sound what happens if you enable the overlay but not the debug feature. I probably need to have like a "refcount" on how many features want to have NPC dialogue extractions enabled.

magicus:
I'd prefer to fix that when I disable the debug feature by deafult, if that's okay.

magicus:
As for positioning, I'd be fantastically cool if we could render the dialogues as speech bubbles floating in the world next to the characters. :-) At least, I think it would be cool. Maybe it'd just be impractical.

But I'm open for suggestion for other placements. To the left, just above the chat lines is another obvious choice. But I think that if you interact with an NPC, the dialog is kind of the most central thing to your play experience at that time, and the placement should mimic that.

magicus:
I've now adapted the fix to the recently repaired calculateRenderHeight. I think it is about as clear for commit as it'll ever be.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -0,0 +1,155 @@
...
+            // TODO: Replace this with a more stylish background?
+            RenderUtils.drawRectBorders(
+                    poseStack,
+                    CommonColors.DARK_GRAY,
+                    this.getRenderX(),
+                    this.getRenderY(),
+                    this.getRenderX() + this.getWidth(),
+                    this.getRenderY() + textHeight + 10,
+                    1,
+                    1.8f);
```
kristofbolyai: I think we can just not render a border for the overlay, it might look better. If you agree to do this, I will approve after this is changed, if you think we should keep this, LMK.

magicus: Let me check how that looks. I think it might be a bit bland. 

I'll be occupied elsewhere for a few days so it might be quiet for me for a while.

kristofbolyai:
Also I would like to mention that this does not seem to work in production.

magicus:
What is not working? 

And "in production", you mean as in contrast to "beta"? I have never run on beta, so I've tested it only on production. To be fair, I've only tested it with my level 11 test character, so just Detlas/Ragni NPCs, but they all worked fine.

kristofbolyai:
> What is not working? 
> 
> 
> 
> And "in production", you mean as in contrast to "beta"? I have never run on beta, so I've tested it only on production. To be fair, I've only tested it with my level 11 test character, so just Detlas/Ragni NPCs, but they all worked fine.

Built jar in multimc, fabric loader + api. I assume it is because we don't have the debug feature.

magicus:
Oh. Bummer. I guess I need to do that proper ref-count activation anyway, unless I can come up with some better solution.

magicus:
Ok, i ditched the debug feature completely, and made sure the NPC dialogue extraction is enabled iff the NPC dialogue overlay is enabled.

I think this should be good to go in now.

kristofbolyai:
Have you tested this works in production (some launcher)?

kristofbolyai (approved):
Seems fine

magicus:
There should be no problem with production since I don't use debug features any longer. Do you want me to test it anyway?

kristofbolyai:
Yes please just to make sure.

magicus:
Working fine.

PR #201, opened 2022-08-02
Mod: Fix manager init order and WebManager bug

kristofbolyai:
#188 ordered the init of managers in alphabetical order. This won't work, we need a solid order for the inits. This PR "reverts" it to original state. I also managed to found that `WebManager` does not post the event it is supposed to, and that `WebLoadedCondition` satisfaction is not always calculated correctly either. Fixed both of those issues.


magicus:
Bootstrapping is always tricky. But I'm a bit disappointed that we already managed to build in intra-manager dependencies. :-( Can we try to make the dependencies more explicit, either by code or in comments (that is, what is depending on what else, and why)

Also, what broke? I did not notice anything in my branch. (That's not saying nothing broke, I believe you)

kristofbolyai:
Item stats, as web manager was the last to load, not features.

kristofbolyai:
As for the comments, sure, but I can't do it until tomorrow evening. If you care to merge this today, you could push a commit with comments added. If not, I will do it tomorrow.

(iirc, the only important thing is that FeatureManager depends on WebManager, other managers might have connections, but overall their init order should not matter. Don't quote me on this though, there might be more dependencies)

kristofbolyai:
Hoping to merge this ASAP, this is an quite annoying bug.

P0keDev (approved):

magicus (approved):

PR #205, opened 2022-08-03
Fix the disconnect problems, for real this time

magicus:
I gave up on using the Connection.disconnect method, it was too imprecise, and kept triggering randomly. (I assume this was used for other kinds of connections, like some microsoft "phone home" functionality or so)

Instead I now inject the disconnect event either when the user presses the "Leave server" button, or when the main packet connection gets an IO exception.


kristofbolyai (approved):
Seems okay! Let's hope this works.

kristofbolyai:
Any case, it is "better" not to have a disconnect event by accident, by having them when they are not supposed to happen.

PR #207, opened 2022-08-03
FontRenderer: Fix calculateRenderHeight

kristofbolyai:
This PR fixes issues with calculateRenderHeight. It should work as intended in all cases now. @magicus, I think it would be beneficial to merge this before #200.


P0keDev (approved):

common/src/main/java/com/wynntils/mc/render/FontRenderer.java
```diff
@@ -213,11 +219,14 @@ private float calculateRenderHeight(List<TextRenderTask> toRender) {
                             .size();
                 }
 
-                height += lines * font.lineHeight + NEWLINE_OFFSET * (lines - 1);
+                height += lines * font.lineHeight;
+                totalLineCount++;
```
magicus, line 223: @kristofbolyai Should that not be like `totalLineCount += lines`?

kristofbolyai: At one point I thought so, I will check.

kristofbolyai: Tested this, seems to work fine.

magicus: I'm just confused as to how this can be correct. Can you explain, apart from just "it seems to work"? If not, maybe the logic needs some more straightening out so that it becomes apparent.

kristofbolyai: I will take a look.

PR #208, opened 2022-08-03
Sort Item Guesses by level

Incompleteusern:
image


common/src/main/java/com/wynntils/wc/custom/item/UnidentifiedItemStack.java
```diff
@@ -68,30 +70,49 @@ public UnidentifiedItemStack(ItemStack stack) {
...
-        if (items == null) return;
+        if (items == null || items.isEmpty()) return;
+
+        tooltip.add(new TranslatableComponent("feature.wynntils.itemGuess.possibilities"));
+
+        Map<Integer, List<MutableComponent>> levelToItems = new TreeMap<>();
 
-        MutableComponent guesses = new TextComponent("");
         for (String item : items) {
+            ItemProfile profile = WebManager.getItemsMap().get(item);
```
kristofbolyai, line 80: Based on WebManagers design, I assume ItemGuessFeature loads the items? (Just to confirm)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -46,8 +46,9 @@
   "feature.wynntils.healthPotionBlocker.threshold.name": "Threshold (%%)",
   "feature.wynntils.healthPotionBlocker.thresholdReached": "You are already at %s%% of your max health!",
   "feature.wynntils.ingredientPouchHotkey.name": "Ingredient Pouch Keybind Feature",
+  "feature.wynntils.itemGuess.levelline": "  §a- §7Lv. %s ",
```
kristofbolyai: 'levelLine' perhaps?

kristofbolyai (commented):
Looks good overall! 

PS: Are you back developing? 😮

common/src/main/resources/assets/wynntils/lang/en_us.json
Incompleteusern: That would be better

common/src/main/java/com/wynntils/wc/custom/item/UnidentifiedItemStack.java
Incompleteusern: It doesn't as of now, so I'll try to remedy that.

common/src/main/java/com/wynntils/wc/custom/item/UnidentifiedItemStack.java
```diff
@@ -68,30 +70,49 @@ public UnidentifiedItemStack(ItemStack stack) {
...
+        }
+
+        for (Map.Entry<Integer, List<MutableComponent>> entry : levelToItems.entrySet()) {
+            int level = entry.getKey();
+            List<MutableComponent> itemsForLevel = entry.getValue();
+
+            MutableComponent guesses = new TextComponent("  ");
+
+            guesses.append(
+                    new TranslatableComponent("feature.wynntils.itemGuess.levelline", level != -1 ? level : "?"));
```
kristofbolyai: `levelLine` begins with spaces, but we also init `guesses` with spaces? Can we move all to levelLine, or to the component? (Perhaps the second option is better)

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: Also, by having the space as the last char, we seem to have double space in the output? Is that intentional?

image

Incompleteusern: No

common/src/main/java/com/wynntils/wc/custom/item/UnidentifiedItemStack.java
```diff
@@ -68,30 +69,49 @@ public UnidentifiedItemStack(ItemStack stack) {
...
-            if (guesses.getSiblings().size() > 0)
-                guesses.append(new TextComponent(", ").withStyle(ChatFormatting.GRAY));
+            guesses.append("§7: ");
 
-            guesses.append(itemDesc);
-        }
+            guesses.append(itemsForLevel.stream()
+                    .reduce((i, j) -> i.append(new TextComponent(", ").withStyle(ChatFormatting.GRAY))
+                            .append(j))
+                    .get());
```
kristofbolyai: If `itemsForLevel` is empty, reduce produces an empty optional, meaning the .get will fail. Perhaps add an `orElse` or a `isPresent` check.

(@magicus I believe this is the first time Qodana has pointed out something useful, although yes, it can and is stupid a lot of times)

Incompleteusern: I don't think `itemsForLevel` can be empty as any creation of an entry in the TreeMap also adds an element to the list.

kristofbolyai: Then what happens if there is only 1 item?

Incompleteusern: I think it still works, but I will add a check nonetheless.

kristofbolyai: Thanks!

kristofbolyai (approved):
Looks good IMO.

PR #184, opened 2022-07-29
ServerIcon: Fix race condition and add additional safety check

kristofbolyai:
Fixes: 
```
---- Minecraft Crash Report ----
// Don't do that.

Time: 5/23/22, 5:25 PM
Description: Unexpected error

java.lang.NullPointerException: Cannot invoke "com.mojang.blaze3d.platform.NativeImage.m_84982_()" because "this.f_117977_" is null
	at net.minecraft.client.renderer.texture.DynamicTexture.m_117992_(DynamicTexture.java:21) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:classloading}
	at com.mojang.blaze3d.systems.RenderSystem.m_69884_(RenderSystem.java:211) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:classloading}
	at com.mojang.blaze3d.systems.RenderSystem.m_69495_(RenderSystem.java:198) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:classloading}
	at com.mojang.blaze3d.platform.Window.m_85435_(Window.java:333) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:classloading,pl:runtimedistcleaner:A}
	at net.minecraft.client.Minecraft.m_91383_(Minecraft.java:1066) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:mixin,pl:accesstransformer:B,pl:runtimedistcleaner:A,re:classloading,pl:accesstransformer:B,pl:mixin:APP:wynntils.mixins.json:MinecraftMixin,pl:mixin:A,pl:runtimedistcleaner:A}
	at net.minecraft.client.Minecraft.m_91374_(Minecraft.java:663) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:mixin,pl:accesstransformer:B,pl:runtimedistcleaner:A,re:classloading,pl:accesstransformer:B,pl:mixin:APP:wynntils.mixins.json:MinecraftMixin,pl:mixin:A,pl:runtimedistcleaner:A}
	at net.minecraft.client.main.Main.main(Main.java:205) ~[client-1.18.2-20220404.173914-srg.jar%2353!/:?] {re:classloading,pl:runtimedistcleaner:A}
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?] {}
[...]

PR #212, opened 2022-08-03
WynncraftButtonFeature: Fix weirdness

kristofbolyai:
Superseded https://github.com/Wynntils/Artemis/pull/184.

Crash log: https://hst.sh/ukiyejodek

Honestly, I am still not 100% sure what is causing this. 

What I could find:

DynamicTexture constructor is not called on a render thread, thus is being put on render queue. For some reason, this render queue is called again when rendering window. Mostly what I did here should be safe already, however I still added the null checks, they can't hurt. 

What I believe goes on is when the constructor call is replayed again on the render thread, nativeImage is somehow getting closed. Or I don't know. All of this is weird. My idea is to put the assignment of `serverIconLocation` to the renderThread too, making it not set before the render call is not finished. That might be the key to solving this.

If this PR does not solve the issues, we should just not do this. I do not believe updating server icon for the button is worth random crashes when starting the game.


magicus (approved):
It's hard to say if this is correct. It does not look wrong, but neither did the old code. Let's try this fix.

Incompleteusern (approved):

PR #213, opened 2022-08-04
WynncraftButtonFeature: Disable dynamic icon loading

kristofbolyai:
I am a bit sad we have to do this, I think loading the icon dynamically is a cool concept. What I don't think is cool is crashing the user on launch, repeatedly. For now, let's disable dynamic loading, and use an actual texture file. If anyone fixes the dynamic stuff up, we can just enable this again.


P0keDev (approved):

magicus:
Actually, a problem with the old solution was that the first time you started Wynntils, you got the wrong icon. That gave a really bad first impression. So I think this is kind of an improvement. At any case, if we want to reintroduce dynamic icons (have Wynncraft ever changed their icon?), we must make sure we start by using this fallback texture at the first run at the very least.

PR #215, opened 2022-08-04
Character: Fix CharacterInfo recognition

kristofbolyai:
(This also fixes `ItemLockFeature`, which depends on this.)


kristofbolyai:
@magicus @P0keDev @Incompleteusern 

I feel like it is time to bring this up. I think we can all agree all of our time is limited. However, when some of us gets a little bit of free time, and produces 4-5 PRs in an afternoon (me usually), and all of you are busy, I can get a bit stuck, because I am playing/working on a version of the mod that is missing 3-4 bugfixes. Do you think it would be fine for us, who (most of time knows what they are doing) to merge (mostly bugfix / few liner PRs) without reviews? The PR itself would be still there, with reviews requested, so you can use GH notifications to check the PRs out, but I believe this would speed up / ease development much. 

I chose this PR to comment on, as this is the textbook case of what I am talking about.

P0keDev (approved):

magicus:
I would be okay with that. As long as we agree that "well well, but it's merged now" is an adequate answer to a post-commit request for changes, i.e. the burden is still on the PR submitter to listen to (well-founded) criticism and request for changes in the PR, even if it happens post-commit.

But I'd also like to say that you can merge in all the bug fixes in a local "bolyai-best-of-breeds" branch, and then create a new branch from that with your new feature. You can then (relatively) easily copy these changes back to a main-based branch. So you can get to work on a bug-fixed baseline and still do new development.

Not saying we should do that instead, just pointing it out as a possibility. I used to do that a lot when I started doing work on Legacy.

kristofbolyai:
> I would be okay with that. As long as we agree that "well well, but it's merged now" is an adequate answer to a post-commit request for changes, i.e. the burden is still on the PR submitter to listen to (well-founded) criticism and request for changes in the PR, even if it happens post-commit.
> 
> 
> 
> But I'd also like to say that you can merge in all the bug fixes in a local "bolyai-best-of-breeds" branch, and then create a new branch from that with your new feature. You can then (relatively) easily copy these changes back to a main-based branch. So you can get to work on a bug-fixed baseline and still do new development.
> 
> 
> 
> Not saying we should do that instead, just pointing it out as a possibility. I used to do that a lot when I started doing work on Legacy.

Okay, agreed. 

Unrelated: please check staff discord internal notices \:)

PR #221, opened 2022-08-04
CustomBarsFeature: Don't render if not on world

kristofbolyai:


P0keDev (approved):

magicus:
Thanks! :) This has annoyed me a bit every time I've started Wynntils, but I were always about to debug something else and got sidetracked away from fixing it.

PR #223, opened 2022-08-04
Qodana: Bump version

kristofbolyai:


kristofbolyai:
@magicus This version should have pull request mode, meaning only new issues will be reported. This does not make it smarter (except the update to the newer linters), but we can hope :).

Issue #224, opened 2022-08-04
ChatManager wrong chat type


kristofbolyai:
@magicus 
![image](https://user-images.githubusercontent.com/49001742/182939598-31382771-d4dd-47b7-9c5a-97f23b797301.png)
`[21:45:06] [Render thread/INFO] (Minecraft) [CHAT] BACKGROUND-PARTY: §r§8[§r§7VIP+§r§8] §r§7FancyFedora§r§8 has just logged in!`

This breaks `InfoMessageFilterFeature` as expected.

kristofbolyai:
![image](https://user-images.githubusercontent.com/49001742/182940626-c9692fef-8764-4b32-ad78-beb2c65f5bef.png)

Shift clicking soul points, assigned different types, not sure if that is what you want.

HighCrit:
![image](https://user-images.githubusercontent.com/35771251/183042511-6b791fed-f6ff-430e-ac0b-eb596a740c55.png)
What is the distinction between the messagetypes normal and system? As I expected bomb bells to be classified as system, but they're not.

magicus:
Thanks for the reports, I'll look into them!

magicus:
@HighCrit THe difference is if Wynncraft tags the chat packet as `CHAT` or `SYSTEM`. I don't know the reason for their choice, or if it has any practical difference. (They look the same to the user.) I assume this is a vanilla distinction that does not really apply to Wynncraft. But the difference is yet another clue for us to correctly differentiate between certain types of messages.

magicus:
@kristofbolyai With different types, you mean SYSTEM/NORMAL? As I said to HighCrit above, that is basically just reporting what Wynncraft choose.

kristofbolyai:
> @kristofbolyai With different types, you mean SYSTEM/NORMAL? As I said to HighCrit above, that is basically just reporting what Wynncraft choose.

Yeah, it is fine them, except the issue description.

kristofbolyai:
![image](https://user-images.githubusercontent.com/49001742/183655175-2a63c78d-62ce-4bb7-9545-3aa04d0fcf69.png)
 I think this issue is related. Shown on joining game.

kristofbolyai:
(Only some text is filtered)

magicus:
What do you mean? The general advertising done by Wynncraft is not filtered. This is due to two reasons:
1) It's a lot of different messages, and hence a lot of additional filters, and
[...]

PR #225, opened 2022-08-04
InfoMessageFilterFeature: Add local join messages and color to soul point message

kristofbolyai:
image

When in NPC dialogue, local join messages are broken. See #224.

This PR also adds all the missing I18n that was left out (I assume accidentally).


common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -77,6 +78,8 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
         String msg = e.getCodedMessage();
         MessageType messageType = e.getMessageType();
 
+        System.out.println("msg = " + msg);
```
Incompleteusern: leftover debug I presume.

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -176,4 +193,21 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
...
+                            + ChatFormatting.DARK_GREEN + "] " + ChatFormatting.GREEN + playerName);
+            case "VIP+" -> NotificationManager.queueMessage(
+                    ChatFormatting.GREEN + "→ " + ChatFormatting.DARK_AQUA + "[" + ChatFormatting.AQUA + "VIP+"
+                            + ChatFormatting.DARK_AQUA + "] " + ChatFormatting.AQUA + playerName);
+            case "HERO" -> NotificationManager.queueMessage(
+                    ChatFormatting.GREEN + "→ " + ChatFormatting.DARK_PURPLE + "[" + ChatFormatting.LIGHT_PURPLE
+                            + "HERO" + ChatFormatting.DARK_PURPLE + "] " + ChatFormatting.LIGHT_PURPLE + playerName);
+            case "CHAMPION" -> NotificationManager.queueMessage(
+                    ChatFormatting.GREEN + "→ " + ChatFormatting.YELLOW + "[" + ChatFormatting.GOLD + "CHAMPION"
+                            + ChatFormatting.YELLOW + "] " + ChatFormatting.GOLD + playerName);
```
Incompleteusern: This lacks a default branch, and has a decent bit of duplicated code

kristofbolyai: No need for default branch, I will fix duplication.

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
kristofbolyai: Yep

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
HighCrit: We thought so too when HERO was the highest rank. But this shouldn't crash if that changes so 🤷

kristofbolyai: Champ branch does not even work, wynn sends hero format for them. Also, we can't just put an universal format.

HighCrit:
Do we want to remove the abilities to hide login announcements with this PR?

kristofbolyai:
> Do we want to remove the abilities to hide login announcements with this PR?

Is there an option to do so in legacy?

HighCrit:
> > Do we want to remove the abilities to hide login announcements with this PR?
> 
> Is there an option to do so in legacy?

Not to my knowledge, but I'd personally prefer to hide instead of redirect it and since it was already in place before this PR we could just leave it in. 

Don't know whether that would be allowed though. As it is one of the rank's perks.

magicus:
Thanks for fixing missing i18n. We should probably have some way to check that we actually have proper i18n keys for all our automated translation stuff, it's so easy to miss when you don't have a literal i18n key in the code.

magicus:
The thought didn't even cross my mind that it'd be considered disallowed to hide login messages. Can we get that verified with Wynncraft staff somehow? @HeyZeer0?

As I think I said before, my original plan was to have an enum `{ KEEP, HIDE, REDIRECT }` for all of these groups, but the PR was large enough as it was. So, unless it is verboten, we should perhaps start by applying this principle to login messages. That way, we won't lose the ability to disable them. (As I'd like to do, personally)

kristofbolyai:
FilterType should be added now.

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -153,14 +167,23 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
...
+                if (matcher.find()) {
                     e.setCanceled(true);
+                    if (loginAnnouncements == FilterType.HIDE) {
+                        return;
+                    }
+
+                    String playerName = matcher.group(3);
+                    String rank = matcher.group(2);
+
+                    sendLoginMessage(playerName, rank);
```
magicus, line 181: The duplication here is a bit annoying. :( Don't see a very good way to address it, though. Maybe we should make sure the normal and BACKGROUND_ patterns have the same groups (you can always add a dummy `()` to pad it), and they we collect the normal and background patterns into two FilterPattern instances, and select the correct one based on the chat type? 

That sounds like a separate PR, though...

magicus (approved):

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
kristofbolyai: Agreed, I think something like was discussed in DC.

PR #227, opened 2022-08-05
Fix outline rendering

P0keDev:
The original implementation rendered the outline with offsets in 8 different directions (4 cardinals + 4 diagonals), but this didn't look very good. Now it's only offset in the cardinal directions, which matches legacy's outline implementation.


kristofbolyai:
Can you show a screenshot?

kristofbolyai (approved):

PR #228, opened 2022-08-06
Load Main-map.png

Incompleteusern:
Looking for feedback.


common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -231,12 +237,89 @@ public static boolean tryLoadItemList() {
...
+    public static CompletableFuture<Boolean> tryLoadMap() {
+        if (apiUrls == null || !apiUrls.hasKey("MainMap")) return CompletableFuture.completedFuture(false);
+
+        final File mapDirectory = new File(API_CACHE_ROOT, "map");
+
+        mapDirectory.mkdirs();
+
+        final CompletableFuture<Boolean> result = new CompletableFuture<>();
+
+        handler.addAndDispatchAsync(new RequestBuilder(apiUrls.get("MainMap"), "main_map.info")
```
magicus: I am very, very skeptical towards bringing this old bad `CenterX` thing into Artemis.

I want to have a new, json-based metadata file with map information. I propose this:
`maps.json`:
```
[
{ "file": "main-map.png",
  "x1": -3443,
  "z1": 1002,
  "x2": 2443,
  "z2": -3922
} 
]
```

This is interpreted that one tile of the map (in this case, the entire world map) resides in the file `main-map.png`. The top left pixel on the map png corresponds to the block at the position X=-3443,Z=1002, and the bottom right corresponds to X=2443,Z=-3922. It is a failure if the size of the png does not match the size spanned by the coordinates 2D BB, as given in the metadata.

HighCrit: @magicus would we want to limit ourselves to 1 pixel = 1 block? We could have an option for higher resolution maps in the future. Also, should we add hashes to the metadata? That way we only have to download new/updated areas. Although that might prevent users from using a custom map.

Incompleteusern: I agree, but this is not the pr to change things as this would likely require changes to Artemis. `CenterX` and `CenterZ` actually refer to the bottom right X and Z coordinates so they have been renamed here.

magicus: You mean that it would require changes to Athena? 

I still think this *is* the place to introduce such changes. It's better that you make a proper json file, even if you have to check it in as part of the resources of Artemis, than to write code that parses to old and bad stuff.

I also think that we make the Athena/server part faaaar to complicated. For static data caching, we just need to fetch it off a github repo. Wham, bam, done. The Athena logic is really only needed for stuff like user tracking, etc.

So better yet, make a json file and put it at [Data-Storage](https://github.com/Wynntils/Data-Storage) in e.g. a new `Artemis` directory, and get it using http from a raw link like `https://raw.githubusercontent.com/Wynntils/Data-Storage/master/item-changes.json`.

magicus: And yes, @HighCrit's idea of a hash is an excellent idea. If we download the metadata, and the hash matches the png file we have on disk in our cache, we're good, otherwise we need to refetch it.

magicus: @HighCrit Acutally, my first attempt at the json format included a `"scale": 1`, but then I removed it. I think we better start with that assumption. If we ever want to change it, it is easy to add the scale factor back. But in the meantime, we will be creating a lot of extra work for ourselves, trying to pass around a scale factor that is not used. And in the end, if it is not tested, it will likely be broken anyway the day we introduce a non-1-to-1 map. So my recommendation is: "Yes, it might be a good idea, but not just yet".

Incompleteusern: The current implementation of map fetching on the serverside resides in [WynntilsWebsite-API](https://github.com/Wynntils/WynntilsWebsite-API/tree/master/maps).

Incompleteusern: @magicus This should be added now, with the additional pr of https://github.com/Wynntils/WynntilsWebsite-API/pull/7.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -231,12 +234,69 @@ public static boolean tryLoadItemList() {
...
+    public static CompletableFuture<Boolean> tryLoadMap() {
+        if (apiUrls == null || !apiUrls.hasKey("AMainMap")) return CompletableFuture.completedFuture(false);
+
+        final File mapDirectory = new File(API_CACHE_ROOT, "map");
+
+        mapDirectory.mkdirs();
+
+        final CompletableFuture<Boolean> result = new CompletableFuture<>();
+
+        handler.addAndDispatch(new RequestBuilder(apiUrls.get("AMainMap"), "main_map.info")
```
magicus:
```suggestion
        handler.addAndDispatch(new RequestBuilder(apiUrls.get("AMainMap"), "map")
```

This is, afaict, just a marker that we use to distinguish different request. Better make it a bit nicer.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -231,12 +234,69 @@ public static boolean tryLoadItemList() {
...
+                .cacheTo(new File(mapDirectory, "main-map.txt"))
```
magicus:
```suggestion
                .cacheTo(new File(mapDirectory, "maps.json"))
```

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -231,12 +234,69 @@ public static boolean tryLoadItemList() {
...
+                .handleJson(json -> {
+                    JsonObject mapData = json.getAsJsonArray().get(0).getAsJsonObject();
```
magicus: Since we only have a single map image, this works fine for now. But maybe add a comment describing this limitation?

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
```diff
@@ -0,0 +1,87 @@
...
+        this.texture = texture;
+        this.x1 = x1;
+        this.z1 = z1;
+        this.x2 = x2;
+        this.z2 = z2;
+        this.textureWidth = textureWidth;
+        this.textureHeight = textureHeight;
+
+        // Remove this if we ever have non 1 to 1 maps
+        assert x2 - x1 == textureWidth;
```
magicus, line 39: We can calculate width and height from the input coordinates. 

However, we need to validate that this information matches the actual texture. So you should check the calculated width/height (from x12/z12) with `texture.getPixels().getWidth()` etc. And if it fails, throw an exception (a runtime exception is fine, you don't have to declare it).

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
```diff
@@ -0,0 +1,87 @@
...
+    public int getWorldXPosition(double textureX) {
+        return (int) Math.round(textureX + x1);
+    }
+
+    public int getWorldZPosition(double textureY) {
+        return (int) Math.round(textureY + z1);
+    }
+
+    public ResourceLocation resource() {
+        if (mapResource == null) {
```
magicus: Why is this not done in the constructor?

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
```diff
@@ -42,7 +42,7 @@ public ServerIcon(ServerData server, boolean allowStale, Consumer<ServerIcon> on
 
         // Try default
         ResourceLocation destination =
-                new ResourceLocation("wynntils/servers/" + Hashing.sha1().hashUnencodedChars(server.ip) + "/icon");
```
magicus, line 45: This seem unrelated..?

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -67,6 +72,8 @@ public class WebManager {
 
     private static @Nullable WynntilsAccount account = null;
 
+    private static @Nullable MapProfile map = null;
```
magicus: Ditch the `@Nullable`. They are just making the code cluttered.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
Incompleteusern: How does this differ from what occurs right now?

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
Incompleteusern: It's shoehorned in, as a small fix.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
HighCrit: We should also have variable names for the files. I commented on https://github.com/Wynntils/WynntilsWebsite-API/pull/7 with an idea on how to do that. Which is trivial, really.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
Incompleteusern: I said more about this in the discord, but `TextureManager` is null when the mod initalizer for fabric is called, thus creating it in the constructor leads to a crash or an unregistered resource location.

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
magicus: What is the fix about? The texture does look fine to me?

common/src/main/java/com/wynntils/core/webapi/WebManager.java
magicus: The "obvious" solution is to name the files after the hash, I guess. But then we should really have some method of cleaning out old files as well. So if the files are named individually per segment, we need not bother about that part.

HighCrit: We could also remove all files that did not match any hash in that directory

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
magicus: 1) You need to pass in the height to the constructur, even though it can be calculated. Instead you have an assert to verify that the calculation is correct.

2) You need to check the actual texture dimensions.

magicus: Okay, I see now that you send in the texture size. But that's putting the responsibility on the caller. This should be done by this class instead.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
magicus: Does the crash happen with the ResourceLocation constructor, or the `register` call? If it is the latter, the ResourceLocation should still be created in the constructor.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -231,12 +234,69 @@ public static boolean tryLoadItemList() {
                 .useCacheAsBackup()
                 .build());
 
-        handler.dispatch(false);
-
         // Check for success
         return isItemListLoaded();
     }
 
+    public static CompletableFuture<Boolean> tryLoadMap() {
```
magicus: Actually, I think we should try to make clear that we are handling more than one map, so the method should be like `tryLoadMaps`, the cache should be like:
```
final File mapsDirectory = new File(API_CACHE_ROOT, "maps");
```

etc.

And the more I think about it, the stronger I feel that we should also correctly loop over all maps in the json response, and not just hard-code one map.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
magicus: Yes, that is a solution. Or we name the files after the coordinates: "-1331,33,-12,553.png" or so.

HighCrit: Well, if we split the larger map file into smaller maps. Or wynncraft has another expansion. We are still stuck with leftover files

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
Incompleteusern: The `register` call. If the ResourceLocation is created in the constructor, then how can registration of texture manager be done later?

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
Incompleteusern: Minecraft's server list stores icons at this resource location, which means that this location should be checked for an already existing icon.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
Incompleteusern: Leftover files are really only a problem when files are renamed around, which should be rather infrequent. I feel it might be better to deal with that in a separate pr.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
Incompleteusern: Will do.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
Incompleteusern: Will do.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
magicus: There are many ways that could be done. Flag it with a boolean, for instance. The thing that bothers me with this code is actually how you hard-code string literals in the constructor. These should have been gotten from the constructor.

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
magicus: But we switched to using our own static icon instead..?

kristofbolyai: We use our static icon if we don't have a cache of it.

Incompleteusern: This is about the original implementation.

common/src/main/java/com/wynntils/core/webapi/profiles/MapProfile.java
Incompleteusern: This should be resolved.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
Incompleteusern: Should be resolved.

magicus (dismissed):

magicus (approved):

PR #232, opened 2022-08-06
ConfigHolder: Fix valueChanged when reflection equals does not work

kristofbolyai:
Comments should explain why I have done this. This PR fixes item lock feature not saving locked slots (and maybe other configs not saving as well).


magicus (approved):
If you say this is fine, I say this is fine.

However, this seems to be a field troubled with issues. I wonder if we should treat this differently. Maybe we can't reliably detect if changes has happened. Maybe we need some way to detect changes as they happen instead, rather than trying to look for changes after the fact.

kristofbolyai:
> If you say this is fine, I say this is fine.
> 
> 
> 
> However, this seems to be a field troubled with issues. I wonder if we should treat this differently. Maybe we can't reliably detect if changes has happened. Maybe we need some way to detect changes as they happen instead, rather than trying to look for changes after the fact.

Proposed changes on discord.

PR #234, opened 2022-08-06
Features: Add TradeMarketAutoOpenChatFeature

kristofbolyai:
This feature opens trade market automatically if an input from user is required (filter, set amount, etc).

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/TradeMarketAutoOpenChatFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/TradeMarketAutoOpenChatFeature.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.user;
+
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.mc.event.ScreenClosedEvent;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wc.event.ChatMessageReceivedEvent;
+import com.wynntils.wc.utils.WynnUtils;
+import java.util.regex.Pattern;
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class TradeMarketAutoOpenChatFeature extends UserFeature {
+
+    // Type the price in emeralds or type 'cancel' to cancel:
+    // Type the amount you wish to buy or type 'cancel' to cancel:
+    // Type the item name or type 'cancel' to cancel:
+    private static final Pattern TYPE_TO_CHAT_PATTERN = Pattern.compile("§6Type the .* or type 'cancel' to cancel:");
+
+    private boolean openChatWhenContainerClosed = false;
+
+    @SubscribeEvent
+    public void onChatMessageReceive(ChatMessageReceivedEvent event) {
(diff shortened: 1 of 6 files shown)


P0keDev (approved):

common/src/main/java/com/wynntils/features/user/TradeMarketAutoOpenChatFeature.java
```diff
@@ -0,0 +1,41 @@
...
+    public void onChatMessageReceive(ChatMessageReceivedEvent event) {
+        if (!WynnUtils.onWorld()) return;
+
+        if (TYPE_TO_CHAT_PATTERN.matcher(event.getCodedMessage()).matches()) {
+            openChatWhenContainerClosed = true;
+        }
+    }
+
+    @SubscribeEvent
+    public void onContainerClose(ScreenClosedEvent event) {
```
P0keDev, line 35: do we use this same logic in legacy? just wondering if it's at all possible for the container to close before the message is sent

kristofbolyai: I didn't check until I wrote this, but yes, same logic.

common/src/main/java/com/wynntils/features/user/TradeMarketAutoOpenChatFeature.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.regex.Pattern;
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class TradeMarketAutoOpenChatFeature extends UserFeature {
+
+    // Type the price in emeralds or type 'cancel' to cancel:
+    // Type the amount you wish to buy or type 'cancel' to cancel:
+    // Type the item name or type 'cancel' to cancel:
+    private static final Pattern TYPE_TO_CHAT_PATTERN = Pattern.compile("§6Type the .* or type 'cancel' to cancel:");
```
magicus, line 21: In general, I like regexp patterns to include `^` and `$`, that makes sure we match the whole line, which makes the risk lower that we match on a partial message, potentially sent by a user.

common/src/main/java/com/wynntils/features/user/TradeMarketAutoOpenChatFeature.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private boolean openChatWhenContainerClosed = false;
+
+    @SubscribeEvent
+    public void onChatMessageReceive(ChatMessageReceivedEvent event) {
+        if (!WynnUtils.onWorld()) return;
+
+        if (TYPE_TO_CHAT_PATTERN.matcher(event.getCodedMessage()).matches()) {
+            openChatWhenContainerClosed = true;
```
magicus, line 30: Should we have some kind of additional safeguards here? Like that the chat message should arrive with a container open? Or that it is reset whenever a container is opened? I'm just (slightly) worried that this bool is going to get in an incorrect state, somehow.

kristofbolyai: Legacy impl is quite similar, I guess I could add a reset.

magicus: It's up to you.

PR #237, opened 2022-08-06
Custom item stack counts

HighCrit:
image


common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
```diff
@@ -0,0 +1,39 @@
...
+import com.wynntils.wc.custom.item.properties.type.TextOverlayProperty;
+import net.minecraft.ChatFormatting;
+
+public class SkillPointProperty extends ItemProperty implements TextOverlayProperty {
+    private static final CustomColor TEXT_COLOR = CustomColor.fromChatFormatting(ChatFormatting.WHITE);
+    private final TextOverlay textOverlay;
+
+    public SkillPointProperty(WynnItemStack item) {
+        super(item);
+        item.setCount(1);
```
P0keDev: it might be better to define this sort of "override" as an attribute of `TextOverlayProperty` and do the actual handling in the feature code

HighCrit: Do you mean like `isInventoryText`? Or should I make another property type like `ItemCountProperty`?

P0keDev: similar to `isInventoryText`, yes

kristofbolyai:
Could we get the numbers colored according to the skill?

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
HighCrit: I went a different route. As I only found a way to implement it by iterating over the entire inventory to set the item counts to 1 on specific items or do it ever render event. But I do lack some knowledge of Artemis' codebase so please let me know if there is another way.

HighCrit:
> Could we get the numbers colored according to the skill?

image
:P

common/src/main/java/com/wynntils/wc/custom/item/properties/DailyRewardMultiplierProperty.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.wynntils.mc.utils.ItemUtils;
+import com.wynntils.utils.objects.CommonColors;
+import com.wynntils.wc.custom.item.WynnItemStack;
+
+public class DailyRewardMultiplierProperty extends CustomStackCountProperty {
+    public DailyRewardMultiplierProperty(WynnItemStack item) {
+        super(item);
+        // Multiplier line is always on index 3
+        String value = String.valueOf(ItemUtils.getLore(item).get(3).charAt(25));
```
kristofbolyai: Safeguard so we don't crash if lore is different?

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
```diff
@@ -0,0 +1,24 @@
...
+import net.minecraft.ChatFormatting;
+
+public class SkillPointProperty extends CustomStackCountProperty {
+    public SkillPointProperty(WynnItemStack item) {
+        super(item);
+
+        char colorCode = ComponentUtils.getCoded(item.getHoverName()).charAt(16);
+        // Current skill point amount is always on line 4 i.e. index 3.
+        String pointsLine = ItemUtils.getLore(item).get(3);
+        String points = pointsLine.substring(6, pointsLine.indexOf('p') - 1);
```
kristofbolyai: Am I paranoid if I would rather put this whole block in a try catch?

kristofbolyai (commented):
Looks fine otherwise, I would like @P0keDev to have the final say though.

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
HighCrit: An exception there won't crash the client and will just make the item invisible. 

But if I add safeguards I'd have to check whether the name has 17 characters or more, line 4 exists in the lore and that the points line is longer than 6 chars and that the first p exists and comes after index 6. 

If we expect the lore to change I would just iterate over all the lore lines with a regex. A performance hit I was trying to avoid.

common/src/main/java/com/wynntils/wc/custom/item/properties/DailyRewardMultiplierProperty.java
HighCrit: Analogous to https://github.com/Wynntils/Artemis/pull/237#discussion_r939674722

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
kristofbolyai: Let's just skip it then, it will be fine.

HighCrit: I would like a second opinion on this choice though. @P0keDev or @magicus any thoughts?

magicus: I'm not really fond of "crashing instead of checking" as a programming style, even if we think that this will not affect anything. At the very least, the stack trace will clutter up the log, maybe hiding other, real issues. 

I think you can surround it with a `try { ... } catch (NPE | IOOB) { // ignore }` and just be happy about it?

magicus:
Would it be too much of feature creep to ask if you can also add the skill symbol overlays, just like in legacy? (If you think it is, we'll save it for later)

HighCrit:
> Would it be too much of feature creep to ask if you can also add the skill symbol overlays, just like in legacy? (If you think it is, we'll save it for later)

I don't know what feature you're referring to

magicus:
Screenshot 2022-08-07 at 19 09 43

HighCrit:
@magicus like this 
image

magicus (dismissed):

magicus (approved):

kristofbolyai (approved):

PR #239, opened 2022-08-07
Overlay crash

Incompleteusern:
Add crash info for overlays during crash.


kristofbolyai (approved):

magicus (approved):
This is probably not a bad idea, but I'm also a bit unsure how this would ever help us in case of a crash. 

If we really wanted good debug logging, I'd rather recommend we figure out if the actual crash includes frames from a Wynntils Feature, and if so, dump the `@Config` options and/or all instance variables.

magicus:
And in a slightly related way, we used to have some kind of performance measurement around the overlay rendering code, to measure the performance impact of our overlays. I think that'd be good to have back. In general, I have no qualms about our performance, but the one exception is rendering.

PR #241, opened 2022-08-07
Remove all @Nullable annotations

magicus:
This is a kind of "all-in or nothing" thing. If we annotate for nullness checking it should be done consistently (I don't think it's worth it, though). Right now, it's just randomly cluttering the code.


magicus:
If anyone has an idea how we can add a check to keep this out, let me know. Maybe we can add it to Spotless?

magicus:
Yes, we could. Done.

kristofbolyai:
Do we not care about @NonNull?

magicus:
You mean the random guy you just inadvertently pinged? 😆 

You are correct, we do care about that as well. I just did not remember what the annotation was called, and there were no current instances of it in the code base.

An alternative is to check for `import com.jetbrains.annotations`; that will catch any other random IntelliJ annotations as well. I was briefly considering this but thought it was more indirect. But perhaps that is better?

magicus:
@kristofbolyai I was wrong, there were NonNull annotations, but there was a PEBKAC issue with searching. ;-)

Also, the import check seems to work fine. And I convinced myself it was actually better. After all, we should not have imports from random jetbrain packages.

kristofbolyai (approved):

PR #244, opened 2022-08-07
Cleanups, the IntelliJ way

magicus:
I ran all the IntelliJ analyzers, and fixed the places where I agreed that the code was sub-optimal (about 10%, so this is not very efficient). Also found and fixed a couple of real bugs.

### DIFF
--- a/common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
+++ b/common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
@@ -101,24 +101,25 @@ public class ServerIcon {
             return;
         }
 
-        NativeImage nativeImage;
         try {
-            nativeImage = NativeImage.fromBase64(iconString);
+            try (NativeImage nativeImage = NativeImage.fromBase64(iconString)) {
+                Validate.validState(nativeImage.getWidth() == 64, "Must be 64 pixels wide");
+                Validate.validState(nativeImage.getHeight() == 64, "Must be 64 pixels high");
+
+                synchronized (this) {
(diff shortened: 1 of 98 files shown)


common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
```diff
@@ -101,24 +101,25 @@ private synchronized void loadServerIcon(ResourceLocation destination) {
             return;
         }
 
-        NativeImage nativeImage;
         try {
-            nativeImage = NativeImage.fromBase64(iconString);
+            try (NativeImage nativeImage = NativeImage.fromBase64(iconString)) {
```
kristofbolyai, line 105: Not sure about this but it should be fine because it is disabled.

kristofbolyai (commented):
Are you sure the final changes don't conflict with reflection/gson? If yes this is fine.

magicus:
I'm not sure how I verify if the gson stuff breaks anything. I accepted a lot of these changes at first, but I actually thought I reverted them in the end, since I was not sure. Did I miss any?

common/src/main/java/com/wynntils/mc/utils/objects/ServerIcon.java
magicus: I looked at the NativeImage code and it really looked like it should have it's close() method called. So I agreed with IntelliJ and said that this was an autocloseable miss.

magicus:
I removed the finals from IdentificationProfile. Could not find any more for gson classes.

PR #248, opened 2022-08-07
More IntelliJ fixes

magicus:
I now found out how to enable All Of The Checks! :-) I'm wading through false positives, but am actually finding quite a few gems.


magicus:
This PR is probably easiest to review per commit. The "major" commits are https://github.com/Wynntils/Artemis/pull/248/commits/5370f4fa56e98fb61d15e648cce45fe1f199c383 (which basically makes stuff private if it is possibly) and https://github.com/Wynntils/Artemis/pull/248/commits/b1c8fa7bf4270a300e9d361c75ee456c965968e5 (which makes utility classes final and adds a private constructor); these affect a lot of files with the same kind of changes. The rest are more of a mixed bag.

common/src/main/java/com/wynntils/commands/wynntils/WynntilsConfigCommand.java
```diff
@@ -30,12 +30,14 @@
 import net.minecraft.network.chat.TextComponent;
 import org.apache.commons.lang3.StringUtils;
 
-public abstract class WynntilsConfigCommand {
-    public static final SuggestionProvider<CommandSourceStack> featureSuggestionProvider =
+public final class WynntilsConfigCommand {
+    private WynntilsConfigCommand() {}
```
kristofbolyai: Why?

kristofbolyai: (talking about every empty constructor)

Incompleteusern: I think the goal is to avoid unsubclassed abstract classes, which is done with the final + private constructor.

kristofbolyai: Yeah but this is kinda annoying and messy?

magicus: As @Incompleteusern says, it is to disallow instantiation. If you think it is too messy, I can remove it.

magicus: (I agree that it is a bit messy, but ut also makes it clear it is a utility class. I'm on the fence for that one)

kristofbolyai: This one kinda bothers me? What about @P0keDev @Incompleteusern @HighCrit ?

magicus: If it bothers you, I'll remove it.

HighCrit: In the past a decision was made to not have empty constructors on utility classes. When I proposed an empty constructor or one that throws an exception or both.

HighCrit: I'm in favour of preventing instantiation on utility classes. But I won't lose a night's sleep over it. Wynntils isn't really used as a dependency anywhere and we're all capable enough to know or notice it in PRs.

magicus: I just removed them, so, hey... :) I think they have some value, but as HighCrit says it also unlikely that we will make any mistakes regarding this, and they do add a bit of clutter. So let's avoid the clutter, then.

kristofbolyai (approved):
Seems fine. Hope these changes won't cause big issues later on (although they look fine).

kristofbolyai:
@magicus are we sure we need the inspection profile though? Isn't it just an "enable every setting" file?

magicus:
@kristofbolyai The inspection profile isn't strictly needed. It's just a shared profile that any of you can use when doing code analysis from IntelliJ. It contains the combination of enabled checks that I found useful. (I'm not saying all checks are perfect, or free from false positives, but they tend to catch real issues.) Stupid and/or irrelevant checks have been turned off.

I'd really like to see a way to go between the Qodona config and this inspection profile.

PR #250, opened 2022-08-07
Added ability to render durability arc on hotbar

HighCrit:


Incompleteusern (approved):

kristofbolyai:
Would it be much to ask to fix the arc rendering? We did not have a working line renderer and it is just using fixed textures, unlike legacy, which is dynamic. No need to do this though before merging this if you don't want to.

PR #260, opened 2022-08-09
OverlayManager: Send config update on enable/disable

kristofbolyai:
Fixes NPC dialog overlay.


magicus:
What is broken?

kristofbolyai:
Dialogue extraction does not get enabled. It is because we do not send config updates on first init of overlay.

magicus:
Ah, I did not notice that, just toggled it on and off all the time. Thanks for fixing!

magicus (approved):

PR #262, opened 2022-08-09
FontRenderer: Fix OUTLINED italic/bold text

kristofbolyai:
Fixes notifications with bold/italic style.


magicus:
So the idea is to remove colors but not bold/italic? What is the use case?

kristofbolyai:
If we remove bold formatting from shadow but not the actual text, the shadow rendering is incorrect.

magicus:
Ah, right, for the outline. Now the penny drops!

magicus (approved):

PR #263, opened 2022-08-09
QuestManager: Add coloring to quest description

kristofbolyai:
image

### DIFF
--- a/common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
+++ b/common/src/main/java/com/wynntils/wc/utils/scoreboard/quests/QuestManager.java
@@ -10,8 +10,12 @@ import com.wynntils.wc.utils.scoreboard.ScoreboardHandler;
 import com.wynntils.wc.utils.scoreboard.ScoreboardManager;
 import com.wynntils.wc.utils.scoreboard.Segment;
 import java.util.List;
+import java.util.regex.Pattern;
+import net.minecraft.ChatFormatting;
 
 public class QuestManager implements ScoreboardHandler {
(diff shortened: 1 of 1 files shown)


magicus:
There is some kind of similar formatting in the original text. It is not simpler to try and keep it?

(Otherwise, please adjust the magenta coordinate color. It sticks out too much.)

kristofbolyai:
What color do you recommend?

magicus:
I dunno. Cyan? Yellow? What is the color used by Wynncraft?

magicus:
Or leave it be for now. I think we need to do a "color overhaul" when we've come a bit further. Sometimes it looks a bit too much like a child's coloring book. :)

Or. We could have color profiles. That'd actually be cool. So we define a few "kinds" of highlights, like "general highlight", "coordinates", "names" etc, and we can have themes with nicely tuned colors for these.

kristofbolyai:
image

HighCrit:
In Wynntils we highlight coordinates in the same color as the items in the example. But we could make the color an option

kristofbolyai:
I think they should have separate colors. Making it an option is a bit messy before we do the InternalFeature refactoring.

magicus:
I agree that they should have different colors. The new screenshot looks nice. Is this using wynn colors but with coordinates changed to aqua (from white) and general text changed from gray to white?

magicus (approved):

kristofbolyai:
> I agree that they should have different colors. The new screenshot looks nice. Is this using wynn colors but with coordinates changed to aqua (from white) and general text changed from gray to white?

Yes

PR #264, opened 2022-08-09
Fix UA being unavailable during user account setup

HighCrit:
Also forgot one User Agent instance.

### DIFF
--- a/common/src/main/java/com/wynntils/core/webapi/WebManager.java
+++ b/common/src/main/java/com/wynntils/core/webapi/WebManager.java
@@ -76,17 +76,19 @@ public final class WebManager {
 
     private static List<MapProfile> maps = null;
 
-    private static String userAgent;
+    private static final String USER_AGENT;
 
-    public static void init() {
-        tryReloadApiUrls(false);
-        setupUserAccount();
-
-        userAgent = String.format(
+    static {
+        USER_AGENT = String.format(
                 "Wynntils Artemis\\%s-%d (%s)",
                 WynntilsMod.getVersion(),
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -76,19 +76,21 @@ public final class WebManager {
 
     private static List<MapProfile> maps = null;
 
-    private static String userAgent;
+    private static final String USER_AGENT;
```
magicus: Can't you do like: 
```
    private static final String USER_AGENT = String.format(...);
```
?

magicus: static clinit blocks are the worst...

HighCrit: During testing (on Artemis) yesterday VERSION was still an empty string and the static block fixed it. But am unable to reproduce it now 🤔.

magicus: We're kind of doing the version setup backwards, but the way I read the code the version string should be set at the second line in WynntilsMod.init(), and the WebManager not inited until later in that method. I don't know if we have some racy thing calling methods on the WebManager before that, causing it to be loaded too early.

magicus: In any case, the static block would not fix anything, so it was probably just pure luck, if there is a race.

magicus (approved):

Issue #153, opened 2022-07-09
KeyHolders do not work in inventories


kristofbolyai:
Fix this by adding KeyHolder "scopes" (Inventory, Game, All, etc..), so the keyholder is active in the specific scope.

PR #268, opened 2022-08-09
Features: Add ItemCompareFeature

kristofbolyai:
This feature has two "modes":

1, Compare to equipped armor (space to toggle)
image
image

2, Compare to selected item:
image
image


kristofbolyai:
This seems to work with smaller game windows + tooltip scale feature, although I can see issues arising. In the end, I think if it is not a breaking bug, and only happens in specific screen sizes, it should not block the merging of this feature, and we can look into that later (as it very well might be that tooltip scaling would have to be rewritten again, it is quite intrusive).

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,165 @@
...
+            RenderUtils.drawArc(CommonColors.LIGHT_BLUE, slot.x, slot.y, 200, 1, 6, 8);
+        }
+    }
+
+    @SubscribeEvent
+    public void onItemTooltipRenderEvent(ItemTooltipRenderEvent.Post event) {
+        if (!(McUtils.mc().screen instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
+        if (!COMPARE_MODE_ON) return;
+
+        if (abstractContainerScreen.hoveredSlot != null) {
```
HighCrit: Could  you invert this condition and return, to reduce nesting.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,165 @@
...
+    public void onRenderSlot(SlotRenderEvent.Pre event) {
+        Slot slot = event.getSlot();
+        if (slot.getItem() == COMPARED_ITEM) {
+            RenderUtils.drawArc(CommonColors.LIGHT_BLUE, slot.x, slot.y, 200, 1, 6, 8);
+        }
+    }
+
+    @SubscribeEvent
+    public void onItemTooltipRenderEvent(ItemTooltipRenderEvent.Post event) {
+        if (!(McUtils.mc().screen instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
```
HighCrit: Do we not want to compare with the items send in chat? Or does the current implementation not support that?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,165 @@
...
+
+                Optional<GearItemStack> matchingArmorItemStack = armorSlots.stream()
+                        .filter(itemStack -> itemStack instanceof GearItemStack gItemStack
+                                && gItemStack.getItemProfile().getItemInfo().getType()
+                                        == hoveredGearItemStack
+                                                .getItemProfile()
+                                                .getItemInfo()
+                                                .getType())
+                        .map(itemStack -> (GearItemStack) itemStack)
+                        .findFirst();
```
HighCrit: Would there be a better approach for rings? Arrow keys to switch between them 🤔

HighCrit (changes requested):

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: I think that would not be that easy. We can do that as an improvement.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: I think coding that might not be the most clean code for little benefit.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: Done

magicus:
Cool thing. Is this a port of legacy functionality, or a new idea?

How do you select an item for comparison?

common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java
```diff
@@ -75,6 +76,12 @@ protected void onInit(ImmutableList.Builder<Condition> conditions) {
         registerProperty(WynnItemMatchers::isIngredient, IngredientProperty::new);
     }
 
+    @Override
+    protected boolean onEnable() {
+        return (WebManager.isItemGuessesLoaded() || WebManager.tryLoadItemGuesses())
+                && (WebManager.isItemListLoaded() || WebManager.tryLoadItemList());
+    }
```
Incompleteusern, line 83: This is an internal feature, so this effectively permanently loads item guesses/list.

kristofbolyai: Yeah, but we need this for GearItemStack.

kristofbolyai:
> Cool thing. Is this a port of legacy functionality, or a new idea?
> 
> How do you select an item for comparison?

New idea, Space to compare to equipped armor, C to select item for comparison and hover over second item to compare.

magicus:
It would have been nice to have https://github.com/Wynntils/Artemis/issues/153 for this.

common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java
magicus: Is there any reason to not permanently load the item data?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private static GearItemStack COMPARED_ITEM = null;
+    private static boolean COMPARE_MODE_ON = false;
+
+    @SubscribeEvent
+    public void onRenderSlot(SlotRenderEvent.Pre event) {
+        Slot slot = event.getSlot();
+        if (slot.getItem() == COMPARED_ITEM) {
+            RenderUtils.drawArc(CommonColors.LIGHT_BLUE, slot.x, slot.y, 200, 1, 6, 8);
+        }
```
Incompleteusern, line 46: Is this rendered in the example pictures?

kristofbolyai: image

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,167 @@
...
+public class ItemCompareFeature extends UserFeature {
+    @RegisterKeyBind
+    private final KeyHolder toggleCompareModeKeybind =
+            new KeyHolder("Compare mode", GLFW.GLFW_KEY_SPACE, "Wynntils", true, () -> {});
+
+    @RegisterKeyBind
+    private final KeyHolder compareSelectKeybind =
+            new KeyHolder("Select for comparing", GLFW.GLFW_KEY_C, "Wynntils", true, () -> {});
+
+    private static GearItemStack COMPARED_ITEM = null;
```
magicus: Since these are not constants, I suggest normal camel case naming. (I'm surprised Spotless does not complain..?)

magicus: Also no reason to have them static, is there? I see Qodana complains about it, too.

magicus: And also, could you not just have like `private boolean isCompareMode() { return comparedItem != null; }`, and skip the bool field?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,167 @@
...
+                    .map(itemStack -> (GearItemStack) itemStack)
+                    .findFirst();
+
+            matchingArmorItemStack.ifPresent(gearItemStack -> handleCompareTo(
+                    new PoseStack(),
+                    event.getMouseX(),
+                    event.getMouseY(),
+                    abstractContainerScreen,
+                    hoveredGearItemStack,
+                    gearItemStack));
```
magicus: I'm suggesting you extract the `gearItemStack` out of this block (and let it be null, if not found). Then you can have a structure like

if comparedItem != null
   compareToItem = comparedItem
else
   compareToItem = extacted gearItemStack
endif

if compareToItem != null handleCompareTo(....compareToItem)

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: Clarified this.

common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java
Incompleteusern: My original idea was that if a certain web data such as item guesses isn't used by any features, it isn't loaded. We can ditch that if necessary.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
magicus: In fact, maybe rather inline handleCompareTo into onItemTooltipRenderEvent, and extract the gearItemStack extraction code to a separate method.

If not, please consider renaming it to something like "renderCompareToItem".

common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java
kristofbolyai: I just think it is a hassle to load it everywhere we need to, although per feature reloading of required web data is nice. (Not that reloading everything in WebManager is so much stuff though...)

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private static void handleCompareTo(
+            PoseStack poseStack,
+            int mouseX,
+            int mouseY,
+            AbstractContainerScreen<?> inventoryScreen,
+            GearItemStack hoveredGearItemStack,
+            GearItemStack matchingArmorItemStack) {
+        if (hoveredGearItemStack == matchingArmorItemStack) return;
```
magicus: This can't happen, can it? I don't think you need to protect for that. And if it *can* happen, there's something wrong with the logic in onItemTooltipRenderEvent, and we should abort much earlier.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: Got it.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,167 @@
...
+
+        Screen screen = McUtils.mc().screen;
+
+        if (screen == null) return;
```
magicus: You already did:
```
 if (!(McUtils.mc().screen instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
 ```
in onItemTooltipRenderEvent. (I think this speaks for combining the two methods)

File: common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java (hunk 4 above)
Comment (Incompleteusern) [reply common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java:83]: It might be nice to just permaload commonly-used like item map, while more niche things can be on demand.

Comment (Incompleteusern) [reply common/src/main/java/com/wynntils/features/internal/ItemStackTransformerFeature.java:83]: Chances are that you won't disable every single item map using feature.

File: common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java (hunk 6 above)
Comment (magicus) [reply common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java]: I think that would also help with what seems like a hole right now. It seems that COMPARED_ITEM is not always reset to null when COMPARE_MODE_ON is disabled. I wonder if the onRenderSlot cannot be tricked into rendering "ghost" items, where COMPARE_ITEM is set but compare mode is off.

**Comment by kristofbolyai** [comment]
This is likely better now.

File: common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java [hunk 10]
```diff
@@ -0,0 +1,154 @@
...
+    public void onRenderSlot(SlotRenderEvent.Pre event) {
+        Slot slot = event.getSlot();
+        if (slot.getItem() == comparedItem) {
+            RenderUtils.drawArc(CommonColors.LIGHT_BLUE, slot.x, slot.y, 200, 1, 6, 8);
+        }
+    }
+
+    @SubscribeEvent
+    public void onItemTooltipRenderEvent(ItemTooltipRenderEvent.Post event) {
+        if (!(McUtils.mc().screen instanceof AbstractContainerScreen<?> abstractContainerScreen)) return;
```
magicus: Thinking a bit more of this, I think that `null instanceof AnyClass` is always true, so maybe we should combine this with a null check as well. Otoh, you can't really get a tooltip event unless in some kind of screen, can you?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,154 @@
...
+        } else if (comparedItem != null) {
+            itemToCompare = comparedItem;
+        }
+
+        if (itemToCompare == null) {
+            return;
+        }
+
+        if (itemToCompare == hoveredGearItemStack) {
+            itemToCompare = null;
```
magicus: Maybe just a return here, and skip the duplicated null check on itemToCompare?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
HighCrit: Isn't the tooltip drawn with chat items?

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
```diff
@@ -0,0 +1,154 @@
...
+        } else {
+            abstractContainerScreen.renderTooltip(poseStack, itemToCompare, mouseX + hoveredWidth + 10, mouseY);
+        }
+
+        poseStack.popPose();
+    }
+
+    @SubscribeEvent
+    public void onScreenClose(ScreenClosedEvent event) {
+        compareToEquipped = false;
```
magicus, line 124: Qodana is still b*tching about that the variables are static. I don't think there is any reason to, our features are created as singletons.

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
kristofbolyai: That uses a renderutils method iirc.

Incompleteusern:
> It would have been nice to have #153 for this.

This could be done in a separate pr by `KeyManager`

common/src/main/java/com/wynntils/features/user/tooltips/ItemCompareFeature.java
P0keDev: chat items use the vanilla tooltip rendering method, so it would trigger the event. but at the same time, the screen wouldn't be null, it would be the chat gui

magicus (approved):

PR #271, opened 2022-08-09
Feature: Unidentified Item Icons

HighCrit:
Renders small icons of the itemtype on top of unidentified items for easier recognition. 
image


common/src/main/java/com/wynntils/features/user/inventory/UnidentifiedItemIconFeature.java
```diff
@@ -0,0 +1,48 @@
...
+                12,
+                12,
+                unidentifiedItem.getItemType().getIconTextureX(),
+                unidentifiedItem.getItemType().getIconTextureY(),
+                16,
+                16,
+                Texture.GEAR_ICONS.width(),
+                Texture.GEAR_ICONS.height());
+    }
+}
```
kristofbolyai, line 48: Maybe this feature should only contain configs, and the actual rendering etc should be done by ItemProperties?

HighCrit: This is an analogous setup to the durability arc, item highlights and text overlay. 🤔
If we want a more generic approach for rendering texture on top of items we can,  but what else will it be used for?

kristofbolyai: I am not sure, perhaps the others will have stronger opinions. I am fine with this being a feature or a item property impl.

P0keDev: an ItemProperty still requires a feature to do rendering code - i think this implementation is fine; imo it's safe to assume that we won't need this functionality for any other type of item, and on the off chance that we do, we can make the necessary modifications

common/src/main/java/com/wynntils/core/webapi/profiles/item/ItemType.java
```diff
@@ -5,28 +5,59 @@
...
+        return iconTextureX;
+    }
+
+    public int getIconTextureY() {
+        return iconTextureY;
+    }
+
+    public static ItemType fromString(String typeStr) {
+        return switch (typeStr.toLowerCase(Locale.ROOT)) {
+            case "spear", "hammer", "scythe", "warrior", "knight", "wa" -> SPEAR;
```
magicus: Are we ever going to get any other than the first? Like "Unidentified Scythe" -- is that a thing?

I see you have copied this from legacy. I'm just surprised, this was not how I left it when I created the item type overlays in legacy...

magicus: I see. Seems I just used a pre-existing function, which did what I needed (and more). Not sure we need to carry that baggage over to Artemis.

HighCrit: I copied it from legacy. Although it might be feature envy to add the other cases without a feature that uses them.

magicus: I'd say remove all string literals except the ones we need.

common/src/main/java/com/wynntils/core/webapi/profiles/item/ItemType.java
```diff
@@ -5,28 +5,59 @@
...
+        return iconTextureX;
+    }
+
+    public int getIconTextureY() {
+        return iconTextureY;
+    }
+
+    public static ItemType fromString(String typeStr) {
+        return switch (typeStr.toLowerCase(Locale.ROOT)) {
+            case "spear" -> SPEAR;
```
magicus: Actually, I think this can be simlified to something along the lines of `ItemType.valueOf(typeStr.toUpperCase)`... Didn't realize until now that all our enum names match the strings.

HighCrit: Yes, but should we be concerned about the IllegalArgumentException it may throw 🤔.

magicus: Catch it and return null, to get the same behaviour as the current code.

common/src/main/java/com/wynntils/mc/render/Texture.java
```diff
@@ -14,6 +14,7 @@ public enum Texture {
     ITEM_LOCK("item_lock.png", 16, 16),
     LOOTRUN_LINE("path_arrow.png", 16, 16),
     MANA_BAR("bars_mana.png", 256, 256),
+    GEAR_ICONS("gear_icons.png", 64, 64),
```
kristofbolyai: Keep the alphabetical order please.

magicus (dismissed):
Fix the texture name ordering and this looks good to me.

common/src/main/java/com/wynntils/mc/render/Texture.java
HighCrit: I tried to order it on file name, then got confused what ordering was used 😅

magicus:
(I still think the weapon overlay icons are really, really difficult to see, but we did a couple of iterations with the CT in legacy, and this was the best we came up with. But maybe at some point we can do another attempt, to make the difference between wand and spear clearer, and to make the bow not so thin that it almost disappears, and the relic looking like a squashed insect :-)).

PR #272, opened 2022-08-09
ScoreboardManager+ObjectiveManager: Fix parsing

### LINKED ISSUE #235: ObjectiveOverlay fails to realize completion of objective
image

Likely a parsing error.

### LINKED ISSUE #259: Somebody ated the space on the scoreboard??
image
Who ated the space?? 🤔 🤯  it is a mystery...

kristofbolyai:
I admit this is a bit more heavy handed approach, but this seems to work fine every use case. 

Should fix #259, #235. (In #235, some objectives are just resent as completed by Wynncraft, I have seen this in legacy. /class or rejoin will fix this on Wynn's side. ObjectiveManager will now handle this at parsing.)

### DIFF
--- a/common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
+++ b/common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/wc/utils/scoreboard/objectives/ObjectiveManager.java
```diff
@@ -83,7 +83,9 @@ public static List<WynnObjective> getObjectives() {
 
     @Override
     public void onSegmentChange(Segment newValue, ScoreboardManager.SegmentType segmentType) {
-        List<WynnObjective> objectives = reparseObjectives(newValue);
+        List<WynnObjective> objectives = reparseObjectives(newValue).stream()
+                .filter(wynnObjective -> wynnObjective.getScore() != wynnObjective.getMaxScore())
```
magicus:
```suggestion
                .filter(wynnObjective -> wynnObjective.getScore() < wynnObjective.getMaxScore())
```
would be clearer I think

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -67,30 +68,37 @@ public final class ScoreboardManager {
...
+            ScoreboardLineChange processed = queueCopy.pop();
 
             if (processed.method() == ServerScoreboard.Method.REMOVE) {
+                for (ScoreboardLine lineToRemove : scoreboardLineMap.values().stream()
+                        .filter(scoreboardLine -> Objects.equals(scoreboardLine.line(), processed.lineText()))
+                        .toList()) {
+                    scoreboardLineMap.remove(lineToRemove.index());
+                }
                 continue;
             }
```
magicus, line 93: how about removing continue and making this into a proper else part? I think it would make it clearer -- if we're removing, do one thing, otherwise add/update the content.

common/src/main/java/com/wynntils/wc/utils/scoreboard/ScoreboardManager.java
```diff
@@ -67,30 +68,37 @@ public final class ScoreboardManager {
...
         }
 
         Set<String> changedLines = new HashSet<>();
-        while (!queuedChanges.isEmpty()) {
-            ScoreboardLineChange processed = queuedChanges.pop();
+        while (!queueCopy.isEmpty()) {
+            ScoreboardLineChange processed = queueCopy.pop();
 
             if (processed.method() == ServerScoreboard.Method.REMOVE) {
+                for (ScoreboardLine lineToRemove : scoreboardLineMap.values().stream()
```
magicus, line 84: Also, this could be lambdas all way, right? Skip the for and toList and to the remove in a forEach() instead.

kristofbolyai: forEach does not work there for some reason. It is 2 am so I am either insane or I am missing something (It also didn't work now and before commiting this). For should be fine \:).

magicus: That's odd. It looks like it should work just fine. But ok, send it in as-is, and I can play around with it later.

magicus (approved):

kristofbolyai:
Build seem to be broken. Github :crab:

kristofbolyai:
Oh it is not.

kristofbolyai:
Ah, it is.

PR #274, opened 2022-08-10
ItemProperties: Add MaterialProperty

kristofbolyai:
image


common/src/main/java/com/wynntils/wc/custom/item/properties/TieredCraftingItemProperty.java
```diff
@@ -0,0 +1,58 @@
...
+    public boolean isInventoryHighlight() {
+        return true;
+    }
+
+    @Override
+    public boolean isHotbarHighlight() {
+        return false;
+    }
+
+    protected abstract ChatFormatting getPrimaryColor(IngredientTier tier);
```
magicus: Could you rename these so it is more clear that they are used in parsing? I don't have any really good names, maybe `getPrimaryColorForParsing`?

magicus (approved):

PR #277, opened 2022-08-10
ItemCompareFeature: Reset properly

kristofbolyai:


HighCrit:
Do we want to reset that? Does this not allow us to compare bank items to marketplace items? (I can't check rn because I'm away)

kristofbolyai:
Ah I never though about this

PR #278, opened 2022-08-10
Improved icons for unidentified item overlay

magicus:
HighCrit's fix for the un-id'ed item overlays reminded me exactly how annoyed I was at the poor visibility of the weapon overlays.

Here comes an updated icon set. It is extra challenging since we only have 16x16 to play with; that is the reason why the scaled-down version of the "standard" wynn weapons looked like crap. What I have done:

* Completely redesigned the weapon icons. 

One leading principle was that they should be easy to distinguish from another. This was achieved by using distinct forms, but also by using different slopes for all weapons. This means that once you've learned which is which, you can easily distinguish a wand from a spear just by looking at the slope.

The second leading principle was that the weapons should follow the look and style of the armor/accessories icons. That is, they should have a clear outline and a simple gradient inside this outline.

* All icons are brighter. On top of the colored un-ID boxes, the dark and grayish colors just did not stand out clearly enough. Most notably is that I've made the necklace ribbon brighter, it almost looked black previously.

* I've also made some fixes to the ring (darker and more clearly delimiting outline), and the bracelet (the black "inside" is now different from the outline).

I think these are superior to the old icons. I've been playing for a while with them now, and it is a delight to be able to distinguish weapons just as easy as you can tell a helmet from a chestplate.


magicus:
Oh, and some screen shots!

kristofbolyai:
Can we make this a config variable? Also an in-game screenshot would be nice.

kristofbolyai:
I like the accessory changes, not sure about the weapons. They don't feel like Wynn (that is why I would make this a config)

magicus:
Well, we *can* but I'd say let's not do it unless the users come complaining they dislike the icons and want their old mashed-insect-relic back. :)

HighCrit:
I honestly prefer the current weapons over these a lot more.

magicus:
Sounds like I'm the only one who wants clarity. ;-) Alright, I'll make it an option.

HighCrit:
> Sounds like I'm the only one who wants clarity. ;-) Alright, I'll make it an option.

Well if it's clarity you're after. You could try outlining it. Or you can check if you'd prefer the items draw more. It's what I implemented along the way using the Item renderer. But was unable to get the accessories working.

magicus:
@HighCrit I'm not sure I'm following you. I tried painting an outline to the original "Wynn" weapons, but it just got messy and ugly.

What did you not get working with the accessories?

HighCrit:
> @HighCrit I'm not sure I'm following you. I tried painting an outline to the original "Wynn" weapons, but it just got messy and ugly.
> 
> What did you not get working with the accessories?

The accessories are different damaged version of flint and steal. But all rendered as crosses.

P0keDev:
> > @HighCrit I'm not sure I'm following you. I tried painting an outline to the original "Wynn" weapons, but it just got messy and ugly.
> > 
> > What did you not get working with the accessories?
> 
> The accessories are different damaged version of flint and steal. But all rendered as crosses.

you need to set the `Unbreakable` nbt tag on the item stack

HighCrit:
Also I don't know why but I just don't like the dagger, relik and spear. And I feel like the wand would look better mirrored.

HighCrit:
@P0keDev I did or at least I think I did.

magicus:
As for the mirroring, as I said in the description, on of the goals were to have different slopes for the different weapons, so not all look like a dark gray "slash". 

But I'll add this as an option, and the default is the old legacy Wynn style. I hope everyone is happy now. :)

magicus:
@kristofbolyai @HighCrit Is this ok, now that it's just an option?

kristofbolyai (approved):

HighCrit (approved):

PR #279, opened 2022-08-11
Add info boxes

ryanzhoudev:


common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,164 @@
...
+    private final Overlay infoBox3Overlay = new InfoBoxOverlay3();
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay infoBox4Overlay = new InfoBoxOverlay4();
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay infoBox5Overlay = new InfoBoxOverlay5();
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay infoBox6Overlay = new InfoBoxOverlay6();
```
kristofbolyai, line 51: I imagined this as a special overlay that can have any amount of instances. I just did not add framework for it yet. This is fine until I do that.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,164 @@
...
+                            toRender,
+                            this.getWidth(),
+                            this.getHeight(),
+                            this.getRenderHorizontalAlignment(),
+                            this.getRenderVerticalAlignment());
+        }
+
+        @Override
+        public void renderPreview(PoseStack poseStack, float partialTicks, Window window) {
+            FontRenderer.getInstance()
```
kristofbolyai, line 166: If you are not doing anything with this method, you can just not overwrite it and it will call render.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,164 @@
...
+                            this.getRenderY(),
+                            toRender,
+                            this.getWidth(),
+                            this.getHeight(),
+                            this.getRenderHorizontalAlignment(),
+                            this.getRenderVerticalAlignment());
+        }
+    }
+
+    public static class InfoBoxOverlay1 extends InfoBoxOverlay {
```
kristofbolyai, line 179: Why do we need these? We support multiple instances of the same overlay.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
kristofbolyai: If you are doing this for the different translations, you can just overwrite the getTranslatedName to add the id of the overlay.

kristofbolyai:
Also, if you have questions about the framework, let me know.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
ryanzhoudev: I will add preview text for later, with different colors and formatting codes and such. This is just for now

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
ryanzhoudev: Originally did this to separate the `content` config setting for each box without having 6 extra fields in the parent class. I will try out a different way of doing this in a bit.

kristofbolyai:
I have a proposal for info variables. I will write it up later today.

ryanzhoudev:
"Theoretically working" meaning not working for now. Can't get MinecraftFunctions.XFunction to init properly. Same goes for Y and Z, didn't check others. Works when you override the returns on all of them and just return `0`.

kristofbolyai:
Superseeded by #468

PR #280, opened 2022-08-11
Add feature to turn off potion glint

magicus:
I think 1.18 has redone the potion glint thing, or the resource pack in 1.12 handled this better, or whatever. Now it's turning me crazy, anyway.


HighCrit:
You forgot to add the translation line

HighCrit:
Also, do we want to move away from Minecraft's terminology of foil?

HighCrit (changes requested):
You forgot to add the translation line

magicus:
@HighCrit  Well, the functionality has always ever been called "glint" by everyone I've ever heard. In fact, I had a hard time finding the right method in Minecraft since it was not named glint. So I'd like to stick to that terminology as much as we can. The only exception is the mixin, where the name of the mixin should match the name of the method we're injecting into.

HighCrit (approved):

PR #284, opened 2022-08-11
InfoMessageFilterFeature: Add color to background soul point notifications

kristofbolyai:


HighCrit:
I recall this already being purple. What scenario does this fix?

kristofbolyai:
It is purple if we get it in a non npc dialogue state. This fixes it when we are in dialogue.

Incompleteusern (approved):

PR #287, opened 2022-08-11
Mod: Rework how managers work (along with other changes)

kristofbolyai:
What does this PR do?

- Managers now have an abstract class and a registry. Managers can be either a feature dependency or a persistent manager, which is always loaded. When a manager is a feature dependency, they are only enabled when any of the features dependant on it are enabled.
- OverlaySelectionScreen was changed to respect parent feature enabled state
- WebManager: Added a method to load commonly used objects
- ItemStackTransformerFeature was moved to being a manager. Reasoning is that I can see players wanting a minimal "install" of Artemis, and this feature was quite "heavy". Let's give the user the option to disable this if they don't need it.
- Renamed InternalFeature to StatelessFeature as it fits the use case better. They not does not sound like something the mod needs internally.

What this PR is not:
- I originally wanted to convert every InternalFeature to a manager or feature, but I do not think that is the right thing to do right now. I know @magicus expressed interest in removing InternalFeatures all together. I think this PR is a nice start.

I know the Manager abstract class is kinda niche, but there is no nice way to do inheritance with static methods.

PS: I am going on holiday on Sunday, meaning Saturday will be my last day when I can code. I hope to finish all the change requests by then, as I see this PR as quite a blocking one in terms of development.


magicus:
This will take a while to review.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Set;
+import org.apache.commons.lang3.reflect.MethodUtils;
+
+public class ManagerRegistry {
+    private static final List<Class<? extends Manager>> PERSISTENT_MANAGERS = new ArrayList<>();
+    private static final Map<Class<? extends Manager>, List<Feature>> MANAGER_DEPENDENCIES = new HashMap<>();
+    private static final Set<Class<? extends Manager>> ENABLED_MANAGERS = new HashSet<>();
+
+    public static void init() {
+        registerPersistentDependency(CompassManager.class);
```
magicus: This looks like alphabetic order. Did the bootstrap order issue resolve itself with this patch?

kristofbolyai: The bootstrap issue is not present here because persistent managers are loaded before features and their dependency manager. Order does not matter with persistent managers.

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
```diff
@@ -27,6 +28,10 @@ public class OverlayList extends ContainerObjectSelectionList<OverlayEntry> {
             new TextComponent("Left click on the overlay to edit it."),
             new TextComponent("Right click on the overlay to disable/enable it."));
 
+    private static final List<Component> DISABLED_PARENT_TOOLTIP_LINES = List.of(
```
HighCrit, line 32: Shouldn't these be translatable?

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
```diff
@@ -54,14 +59,25 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
...
-                    false);
+            if (!hovered.getOverlay().isParentEnabled()) {
+                RenderUtils.drawTooltipAt(
+                        poseStack,
+                        mouseX,
+                        mouseY,
+                        100,
+                        DISABLED_PARENT_TOOLTIP_LINES,
+                        FontRenderer.getInstance().getFont(),
+                        false);
```
HighCrit, line 76: A list of features to enable would be helpful for the user I think.

HighCrit (dismissed):

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
kristofbolyai: These were never translatable. I don't plan on fixing this rn because of the resource manager init thing we have not sorted out yet.

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
kristofbolyai: I will add the feature to the tooltip.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
magicus: So basically the ordering problem was "first persistent managers (in any order), then feature managers (in any order)"?

kristofbolyai: There might be cases where it will not be true (Web configs will need WebManager for configs). At the moment, this works fine.

magicus: Is CompassManager really essential? I thought it was just used by features like beacon tracking.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+        registerPersistentDependency(ConfigManager.class);
+        registerPersistentDependency(ClientCommandManager.class);
+        registerPersistentDependency(KeyManager.class);
```
magicus: Do you think you can rename KeyManager into KeyBindManager while you're at it? I've lost count at how many times I need to check what kind of keys it's managing, thinking it's either dungeon keys, or SSH keys, but never realizing it's keybindings. :)

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -138,16 +137,15 @@ public static void init() {
         registerFeature(new ConnectionProgressFeature());
         registerFeature(new PacketDebuggerFeature());
 
-        // internal
+        // stateless
```
magicus: "stateless" in this terminology is supposed to mean "non-enableable"? 

I'm not saying the latter is a better word, but "stateless" is very general, and makes it more sound like there is no fields, which is patently false. So we might need to work on the  name a bit.

"always-on"? "permanent"?

common/src/main/java/com/wynntils/core/managers/Manager.java
```diff
@@ -0,0 +1,16 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.managers;
+
+/**
+ * Managers can have two static methods:
+ * <p>
+ * init: Called when manager is enabled
+ * <p>
+ * disable: Called when manager is disabled
```
magicus, line 12: Do we do this?

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+        } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException e) {
+            WynntilsMod.error(e.getMessage());
+        }
+    }
+
+    private static void tryDisableManager(Class<? extends Manager> manager) {
+        try {
+            MethodUtils.invokeExactStaticMethod(manager, "disable");
+        } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException e) {
+            WynntilsMod.error(e.getMessage());
```
magicus: Maybe we don't need to treat the absence of a disable method as an error? I think it should be optional to implement.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+    }
+
+    private static void updateManagerState(Class<? extends Manager> manager) {
+        List<Feature> dependencies = MANAGER_DEPENDENCIES.get(manager);
+
+        if (ENABLED_MANAGERS.contains(manager)) {
+            if (dependencies == null || dependencies.isEmpty()) {
+                disableManager(manager);
```
magicus: Are you sure you need to split this into two functions? (same for enable) The difference is not really clear to me.

common/src/main/java/com/wynntils/features/stateless/ChatItemFeature.java
```diff
@@ -22,19 +22,15 @@
...
+            ImmutableList.Builder<Condition> conditions, ImmutableList.Builder<Class<? extends Manager>> dependencies) {
         conditions.add(new WebLoadedCondition());
     }
 
-    @Override
-    protected boolean onEnable() {
-        return WebManager.isItemListLoaded() || WebManager.tryLoadItemList();
-    }
-
     @SubscribeEvent
```
magicus: It's not at all clear to me why this is not a normal user enabled feature? We're doing a lot of slightly scary stuff here. Why should the user not be able to turn this off?

I'm thinking a reason for this might be that there is some kind of coupling/dependency between this feature and the screenshot feature. But then we should be able to express "recommends" or "depends" relationships between features instead.

common/src/main/java/com/wynntils/features/stateless/FixPacketBugsFeature.java
```diff
@@ -22,7 +22,7 @@
 import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
-public class FixPacketBugsFeature extends InternalFeature {
```
magicus: This should probably rather be a debug feature. Or a feature that the user can turn off, if it starts to cause crashes.

common/src/main/java/com/wynntils/features/stateless/FixSpellOverwriteFeature.java
```diff
@@ -16,7 +16,7 @@
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
-public class FixSpellOverwriteFeature extends InternalFeature {
+public class FixSpellOverwriteFeature extends StatelessFeature {
```
magicus: Why can't this be a normal user enabled feature? If I understand it correctly, it blocks Wynncraft from updating the held weapon when casting a spell..? Is that because it interferes with other parsing we do, or because of it looking nicer? If it interferes, what is it interfering with?

common/src/main/java/com/wynntils/features/stateless/LootrunFeature.java
```diff
@@ -23,7 +24,7 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @StartDisabled
-public class LootrunFeature extends InternalFeature {
```
magicus: The LootrunFeature is a chapter of it's own... I'd like to redesign this quite a lot. A start is to split between "recoding" and "playback", respectively. We should also have a way to enable events for recording and playback independently.

The split between LootrunUtils and LootrunFeature also seems rather arbitrary. I think we should consider making selected parts of LootrunUtils and LootrunFeature into a LootrunManager. Maybe all of it. 

That change is worth a PR of it's own, so I won't say much more right now on this class.

common/src/main/java/com/wynntils/features/stateless/NotificationsFeature.java
```diff
@@ -2,16 +2,16 @@
...
-import com.wynntils.core.features.InternalFeature;
+import com.wynntils.core.features.StatelessFeature;
 import com.wynntils.features.user.overlays.GameNotificationOverlayFeature;
 import com.wynntils.mc.utils.McUtils;
 import com.wynntils.wc.event.NotificationEvent;
 import net.minecraft.network.chat.TextComponent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
-public class NotificationsFeature extends InternalFeature {
+public class NotificationsFeature extends StatelessFeature {
```
magicus: Maybe this can be solved similar to how I do NPC chats. Don't even send out the event if the Overlay is not active, but print it to the chat instead, in the NotificationManager.

magicus:
I think there is good reason to reconsider the entire category of "stateless" features. I've tried to argue for each and every one of them. The exception is Lootrun, but as I said I think that requires a redesign of it own. In the meantime, it can be a user feature, so we can get rid of the "internal/stateless" category, and we'll solve the proper details later.

common/src/main/java/com/wynntils/wc/custom/item/ItemStackTransformManager.java
```diff
@@ -78,14 +71,8 @@ protected void onInit(ImmutableList.Builder<Condition> conditions) {
         registerProperty(WynnItemMatchers::isMaterial, MaterialProperty::new);
     }
 
-    @Override
-    protected boolean onEnable() {
-        return (WebManager.isItemGuessesLoaded() || WebManager.tryLoadItemGuesses())
```
magicus: Is this since these are permanently loaded now?

magicus:
Is this a correct list of "dependent" managers? They are a bit more scattered and harder to find now...

ItemStackTransformManager
TranslationManager
ActionBarManager
ChatManager
ScoreboardManager

magicus:
I suggest the follow names instead:
"core managers" instead of "persistent managers"
"feature managers" instead of "dependent managers/dependencies"

For clarify maybe we should even split the Manager interface into two marker interfaces, `CoreManager` and `FeatureManager`.

magicus:
Otherwise I think this look good. You have found some a crucial red line between the core and the feature managers. (The feature managers is mostly what I called "model" in my initial architecture. I still think we can tweak them a bit more, but this is a very good start and definitely worth doing, even if we should continue some more tweaking later on.)

magicus:
Actually, thinking about it, maybe we should follow my previous line of reasoning, and use the term "model" instead of "feature manager". So we'd have a ChatModel, ScoreboardModel, etc. (And then possibly just call "core managers" for "managers". Or keep calling them "core managers" even if it's slightly redundant, just to be clear and specific)

I think I'd like that.

kristofbolyai:
> Actually, thinking about it, maybe we should follow my previous line of reasoning, and use the term "model" instead of "feature manager". So we'd have a ChatModel, ScoreboardModel, etc. (And then possibly just call "core managers" for "managers". Or keep calling them "core managers" even if it's slightly redundant, just to be clear and specific)
> 
> 
> 
> I think I'd like that.

But then WorldModel and CharacterModel would have to be changed, and also what would happen with model loader?

common/src/main/java/com/wynntils/features/stateless/FixSpellOverwriteFeature.java
kristofbolyai: I have no idea, I think p0ke just didn't think about it being disabled. Although it does temporarily break item transformation, and it won't become a gear item stack.

common/src/main/java/com/wynntils/features/stateless/ChatItemFeature.java
kristofbolyai: Iirc it should be fine ti disable this, only chat item parsing breaks.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
kristofbolyai: Which two functions are you talking about?

common/src/main/java/com/wynntils/core/managers/Manager.java
kristofbolyai: ScoreboardManager

magicus:
The ModelLoader was kind of the embryo to the ManagerRegistry, even though I did not know it at the time. :) WorldModel and CharacterModel should be reshaped into "Model Managers". I don't think it's hard to do, but if it is, let me know and I can help with that.

kristofbolyai:
> The ModelLoader was kind of the embryo to the ManagerRegistry, even though I did not know it at the time. :) WorldModel and CharacterModel should be reshaped into "Model Managers". I don't think it's hard to do, but if it is, let me know and I can help with that.

I am not exactly sure what you want me to do. Remove ModelLoader and merge it into ManagerRegistry? 
(and at the same time also load core managers there)?

common/src/main/java/com/wynntils/wc/custom/item/ItemStackTransformManager.java
kristofbolyai: Yes

common/src/main/java/com/wynntils/features/stateless/NotificationsFeature.java
kristofbolyai: done

common/src/main/java/com/wynntils/features/stateless/LootrunFeature.java
kristofbolyai: Got it, new PR later.

common/src/main/java/com/wynntils/features/stateless/FixSpellOverwriteFeature.java
kristofbolyai: If we were to "break" item gear parsing and for example rely spell cast feature on it, it would not be nice.

common/src/main/java/com/wynntils/features/stateless/FixPacketBugsFeature.java
kristofbolyai: Done

common/src/main/java/com/wynntils/features/stateless/ChatItemFeature.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
kristofbolyai: It is used by territory command. I guess I can make it not persistent, but then we will have no use case it for now.

kristofbolyai: Done

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
kristofbolyai: Done

kristofbolyai:
Big changes. Model / CoreManager concept should be separated now.

magicus:
Fair enough. Let's push this now, and I can take over and try to get my primordial models to fit into what we actually evolved into. I'll just to a final pass through the code.

common/src/main/java/com/wynntils/mc/mixin/PotionItemMixin.java
```diff
@@ -14,7 +14,7 @@
 
 @Mixin(PotionItem.class)
 public abstract class PotionItemMixin {
-    @Inject(method = "isFoil()B", at = @At("HEAD"), cancellable = true)
```
magicus, line 17: Why remove the method signature?

kristofbolyai: It is not needed, and actually, some intelij plugin even complained. 

As far as I know, by standard, method signatures and param types are left out when there is not any other method with the same name.

common/src/main/java/com/wynntils/screens/lists/OverlayList.java
magicus: @HighCrit We can fix this after https://github.com/Wynntils/Artemis/pull/294

common/src/main/java/com/wynntils/mc/mixin/PotionItemMixin.java
magicus: I see now I messed it up completely. FIxing in separate PR.

magicus:
I've finished my final pass through the code. This looks great now. Ship it!

magicus (approved):

PR #288, opened 2022-08-11
Add mana upgrade tables

magicus:
Like this.

Also fixes bug with isReskinned.

### DIFF
--- a/common/src/main/java/com/wynntils/wc/custom/item/IntelligenceSkillPointsItemStack.java
+++ b/common/src/main/java/com/wynntils/wc/custom/item/IntelligenceSkillPointsItemStack.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wc.custom.item;
+
+import com.wynntils.wc.custom.item.properties.ItemProperty;
+import com.wynntils.wc.custom.item.properties.SkillPointProperty;
+import com.wynntils.wc.model.Character;
+import com.wynntils.wc.objects.SpellType;
+import com.wynntils.wc.utils.WynnUtils;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.TooltipFlag;
+
+public class IntelligenceSkillPointsItemStack extends WynnItemStack {
+    private List<Component> tooltip;
+
+    public IntelligenceSkillPointsItemStack(ItemStack stack) {
+        super(stack);
+    }
+
+    @Override
+    public void init() {
+        List<Component> newTooltip = new ArrayList<>(getOriginalTooltip());
+        newTooltip.addAll(getTooltipExtension());
+        tooltip = newTooltip;
+    }
+
+    @Override
+    public List<Component> getTooltipLines(Player player, TooltipFlag isAdvanced) {
+        if (tooltip == null) {
+            return getOriginalTooltip();
+        }
+
+        return tooltip;
+    }
+
+    private List<Component> getTooltipExtension() {
+        List<Component> tooltipExtension = new ArrayList<>();
+        tooltipExtension.add(new TextComponent(""));
+        tooltipExtension.addAll(getManaTables());
+        return tooltipExtension;
+    }
+
+    private List<Component> getManaTables() {
+        Character.CharacterInfo characterInfo = WynnUtils.getCharacter().getCharacterInfo();
+        SkillPointProperty property = getProperty(ItemProperty.SKILL_POINT);
+        int intelligencePoints = property.getSkillPoints();
+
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
```diff
@@ -15,7 +15,9 @@
 import net.minecraft.ChatFormatting;
 
 public class SkillPointProperty extends CustomStackCountProperty {
-    private static final Pattern POINT_PATTERN = Pattern.compile("^.+?(\\d+) points .+$");
+    private static final Pattern POINT_PATTERN = Pattern.compile("^§7    (\\d+) points      §r§6\\d+ points$");
```
HighCrit: This pattern won't match on HB, the whitespace has changed iirc

common/src/main/java/com/wynntils/wc/custom/item/WynnItemStack.java
```diff
@@ -68,4 +68,9 @@ public <T extends PropertyType> List<T> getProperties(Class<T> propertyType) {
         }
         return collected;
     }
+
+    /**
+     * Called when all properties are setup on this stack
+     */
+    public void init() {}
```
HighCrit, line 75: Is this function not what we've been doing with the constructor so far?

HighCrit (changes requested):

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
ryanzhoudev: In HB, you can also buy cost decreases with ability points. The mana system is also completely different. I think the best thing to do here is to ignore HB for now, then fix (or ignore entirely) once HB is released. If we were to take the AP cost decreases into account, we would need to access all pages of the ability tree.

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
```diff
@@ -15,16 +15,20 @@
...
     private final int skillPoints;
 
     public SkillPointProperty(WynnItemStack item) {
         super(item);
 
+
         char colorCode = ComponentUtils.getCoded(item.getHoverName()).charAt(16);
         String points = "";
         for (String lore : ItemUtils.getLore(item)) {
+            if (lore.contains("point")) {
```
HighCrit: Debug line?

HighCrit (dismissed):

common/src/main/java/com/wynntils/wc/model/Character.java
```diff
@@ -111,7 +111,7 @@ public static CharacterInfo parseCharacter(ItemStack itemStack, int slotNum) {
             List<String> lore = ItemUtils.getLore(itemStack);
 
             int level = 0;
-            ClassType classType = null;
+            String className = "";
```
kristofbolyai, line 114: Why?

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
HighCrit: @DonkeyBlaster but the skill point pattern is still used for other features

ryanzhoudev: Ah okay

common/src/main/java/com/wynntils/wc/model/Character.java
magicus: Sorry, I should have explained this further. This is a bug fix for Character, which did not properly detect if a player were reskinned. The old code did `ClassType.isReskinned(classType.getName())` but that was just begging the question since `classType.getName()` always returned the non-reskinned name. So we needed to save the *actual* name as given in the tooltip. (As usual, I prefer to use "" instead of null for empty strings, unless I'm absolutely sure no NPEs are possible, and I want to signal some kind of error condition)

common/src/main/java/com/wynntils/wc/custom/item/properties/SkillPointProperty.java
magicus: Ooops.. 😊

common/src/main/java/com/wynntils/wc/custom/item/IntelligenceSkillPointsItemStack.java
```diff
@@ -0,0 +1,95 @@
...
+
+        int closestUpgradeLevel = Integer.MAX_VALUE;
+        int level = characterInfo.getLevel();
+
+        LinkedList<Component> newLore = new LinkedList<>();
+
+        for (int j = 0; j < 4; j++) {
+            SpellType spell = SpellType.forClass(characterInfo.getClassType(), j + 1);
+
+            if (spell.getUnlockLevel(1) <= level) {
```
kristofbolyai, line 65: is that meant to be 1?

magicus: Yes, that means that the spell has not even been unlocked yet.

kristofbolyai: Tbh I have no idea what that 1 means there but if you say so.

kristofbolyai (dismissed):

kristofbolyai (approved):

PR #291, opened 2022-08-12
FunctionManager: implement framework for handling "info" functions

magicus:


common/src/main/java/com/wynntils/core/functions/Function.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.functions;
+
+public interface Function {
+    String getValue(String argument);
```
kristofbolyai: String... ?

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -100,6 +101,7 @@ public static void init(String modVersion, boolean isDevelopmentEnvironment) {
 
         // Forced Post Init, this should be last init in almost any case.
         FeatureRegistry.init(); // Needs WebManager, ConfigManager (KeyManager, OverlayManager, ClientCommandManager,
+        FunctionRegistry.init();
```
kristofbolyai: Init before features and after managers?

common/src/main/java/com/wynntils/core/functions/FunctionRegistry.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static Optional<Function> forName(String functionName) {
+        return FunctionRegistry.getFunctions().stream()
+                .filter(function -> function.getName().equals(functionName))
+                .findFirst();
+    }
+
+    public static void init() {
+        // debug
+        registerFunction(new TestFunction());
```
kristofbolyai: Is this where we would register functions?

common/src/main/java/com/wynntils/core/functions/Function.java
```diff
@@ -0,0 +1,13 @@
...
+
+    String getName();
+
+    default void init() {}
```
kristofbolyai: I assume we are waiting on the manager pr for dependencies

common/src/main/java/com/wynntils/core/functions/Function.java
magicus: Now T

common/src/main/java/com/wynntils/core/functions/FunctionRegistry.java
magicus: Yes.

magicus: I'm leaving in the two test functions as placeholders. When we start to get a couple of real functions, we can remove them.

common/src/main/java/com/wynntils/core/functions/Function.java
magicus: Yeah, maybe. We'll have to see how we sort that out. At the moment, I assume functions has no special dependencies as such. There is an `EnablableFunction` which can request stuff to initialize and/or check for the presense of prerequisites. Let's build on top of that when the manager PR has been sorted out.

magicus:
The template parser is still missing. It's a bit more complex to get right, so I'd like to return to it in a separate PR. Otherwise I'm pretty happy with this design. The two test functions should be removed when we start to get some real functions in their place.

common/src/main/java/com/wynntils/commands/wynntils/WynntilsFunctionCommand.java
```diff
@@ -0,0 +1,208 @@
...
+                    .sendFailure(
+                            new TextComponent("Function does not need to be enabled").withStyle(ChatFormatting.RED));
+            return 0;
+        }
+
+        boolean success = FunctionManager.enableFunction(enableableFunction);
+
+        if (!success) {
+            context.getSource()
+                    .sendFailure(new TextComponent("Feature could not be enabled").withStyle(ChatFormatting.RED));
```
Incompleteusern: Should this say `Function` could not be enabled?

common/src/main/java/com/wynntils/core/functions/Function.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.functions;
+
+import com.google.common.base.CaseFormat;
+import net.minecraft.client.resources.language.I18n;
+
+public abstract class Function<T> {
```
kristofbolyai: Translatable interface?

kristofbolyai:
This can now use new managers

magicus:
@kristofbolyai I'm not sure how new managers would fit in. I think I'll keep the function init as a separate step, with I18n, close to the feature init (after the latest i18n init fix). Am I missing something?

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.TextComponent;
+
+/** Manage all built-in {@link Function}s */
+public final class FunctionManager {
```
kristofbolyai: is this not a manager? should we call this registry?

magicus: Ah, that's what you meant. Hm. I started by calling it FunctionRegistry, but then it got more logic like `getSimpleValueString`, or the rather complex `enableFunction` so I renamed it "manager" to reflect that it was just not a simple and stupid registry.

However, it does not listen to subscription events. And the functions themselves have I18n support. ... but that's not needed at class creation time... right. Yeah, ok, fine, let's make this a CoreManager then.

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -0,0 +1,111 @@
...
+            enableableFunction.init();
+        }
+    }
+
+    public static List<Function> getFunctions() {
+        return FUNCTIONS;
+    }
+
+    public static boolean enableFunction(Function function) {
+        if (!(function instanceof EnableableFunction<?> enableableFunction)) return true;
```
kristofbolyai: Why not take EnableableFeature as arg? Also I can't see where we call this

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -0,0 +1,111 @@
...
+                .findFirst();
+    }
+
+    public static Component getSimpleValueString(
+            Function function, String argument, ChatFormatting color, boolean includeName) {
+        MutableComponent header = includeName
+                ? new TextComponent(function.getTranslatedName() + ": ").withStyle(ChatFormatting.WHITE)
+                : new TextComponent("");
+
+        if (function instanceof EnableableFunction<?> enableableFunction && !isEnabled(enableableFunction)) {
```
kristofbolyai: Is this how we want to handle this systematically? Let's say chest count, an enableable function, but it can still produce it's latest active measurement when disabled. Can we someone add this type of functions into the abstraction?

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.functions;
+
+public abstract class EnableableFunction<T> extends Function<T> {
```
kristofbolyai: ActiveFunction?

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
```diff
@@ -0,0 +1,21 @@
...
+    public void init() {}
```
kristofbolyai: Can we get Manager dependency registration here like features do?

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
kristofbolyai: What I think we need is separating the "active" part. Is calculation the active part, like current speed or is the "recording"/data collecting the active part? If the later, can't we always spit out the last recorded value?

common/src/main/java/com/wynntils/functions/TestFunction.java
```diff
@@ -0,0 +1,14 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.functions;
+
+import com.wynntils.core.functions.Function;
+
+public class TestFunction extends Function<Integer> {
+    @Override
+    public Integer getValue(String argument) {
+        return 42;
```
HighCrit: There is no need for this function to exist. We have a real world example.

HighCrit (changes requested):

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
magicus: Good point about the argument. I'll fix.

This is currently called when doing `/wynntils function enable <function>`. My thinking is that it should also be called when an EnableableFunction is added to an info box overlay.

magicus: Actually, about the argument, if I change that, that will force the caller to know about what kind of function it is. If I keep it, the info box can always call "enable" and know they are doing the right thing. So I think I'll prefer to keep it this way, if that is OK with you?

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
magicus: Maybe I'm building an abstraction for a use case that does not exist, but my idea here is that there are potential functions that need to collect data by having a subscription to an event, and cache the last value (or calculate something based on these events, like a moving average, or a sum, etc). Such functions are not "one-shot", like e.g. a function to get your current X and Z position, for which we can instantly query Minecraft.

For this type of functions to be usable, they need to have been "running". I tried to think of a good name, but was not really satisfied with what I could come up with. "ListeningFunction", "BackgroundFunction", "UpdatingFunction", "NonStaticFunction", etc. In the end I crept back to an old OOP practice of just saying what kind of extra method these functions had, namely "enable". I'm still not happy with the name, but it was kind of usable. I do appreciate any suggestions for better name, though! 

An alternative design is to kind of turn this on it's head, assume that all functions are like this, and make the simpler ones a special subclass "StaticFunction". Feels a bit like an anti-OOP-pattern though.

Ok, so given that we assume that this kind of functions need to exist, what should we do if we query them when they are not enabled? Whatever you get back, it's not going to be a valid value. I chose the hard way, of saying "NO!" and clearly marking the result as invalid. But sure, you could certainly return "something" and get a stale value. I'm just worried that we return a misleading value. 

I think basically this boils down to "is it better to say nothing at all unless it is correct, or to try our hardest to give a value even if it's wrong"? I dunno. I prefer the former but maybe I'm a hardliner perfectionist.

Otoh, we are still going to have a situation where a function, once it is enabled, might not have gotten the kind of events it needs, before the first call to `getValue`. So maybe that in itself indicates we should have a more relaxed attitude.

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
magicus: It's definitely less of a mouthful. 

Yeah, it might work. I'll try renaming it and we'll see if it looks ok.

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
magicus: Probably. :) Do you want me to code it in right now, or do it later if/when the need actually arise?

common/src/main/java/com/wynntils/functions/TestFunction.java
magicus: I'll remove it as soon as we have real functions. But I'm not sure what real world example you are thinking of..? I do need some kind of test function in this PR, so I'm planning on removing this in a follow-up when we start to port over legacy functions.

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
magicus: I'll try to rewrite it using a more relaxed approach, where the functions are returning cached values if they have.

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
kristofbolyai: okay either way

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
kristofbolyai: Return cached value but also add a way to see cache age?

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
kristofbolyai: Ah this is a gonna need a bit more changes than I first thought. This is fine now, and I will probably add an interface for feature and functions to handle this.

common/src/main/java/com/wynntils/functions/TestFunction.java
HighCrit: @magicus what is this then? common/src/main/java/com/wynntils/functions/WorldNameFunction.java

HighCrit (changes requested):

common/src/main/java/com/wynntils/functions/TestFunction.java
HighCrit: And with a real world example I am referring to an actual implementation as we might expect in production

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
magicus: I was thinking something like that. But it also makes the API more complicated, and I don't know how to present the information to the user. I think some of these details will also have to get straightened out as we start to implement real functions. Right now, it's a lot of guess work.

common/src/main/java/com/wynntils/functions/TestFunction.java
magicus: Ok, I'll port over a handful of "info variables" from legacy, so we can see this API in action. It's a fair point, which I'm usually trying to make. ;-)

common/src/main/java/com/wynntils/core/functions/EnableableFunction.java
magicus: Like a `ModelDependant`? That's fine for me, as long as you give `Function` a default empty implementation, so trivial functions are not burdened with boilerplate.

common/src/main/java/com/wynntils/functions/TestFunction.java
magicus: We lack the source data for most of our info variables. I made a list at https://github.com/Wynntils/Artemis/issues/301 for keeping track of what is still left to do. I implemented two categories of functions: those that derive their value from Minecraft, and those that derive their value from the general environment (basically, the JVM).

magicus: Ok, I think I got this covered for now. If anything, we should perhaps try to get some ActiveFunctions as well. I'll see if there is any I can implement without far too much work.

HighCrit: For some reason my comment disappeared regarding the already present worldnamefunction :?

HighCrit (approved):

PR #294, opened 2022-08-12
Load Features after I18n is initialized

magicus:
This ResourceLoadStateTracker hook is called async by the Minecraft() constructor, when all resources has finished loading. That seemed to be the proper place to put a hook for having all resources available.

However, this also causes a bit of a jump on the loading progress bar. So I chose to put only the features there, since Managers take far to long to load (!). (Possibly due to Architectury transformations?) This otoh means that it will take extra time before you even start to see the progress bar, so it's not really more helptful, but it is not as ugly. :-)

I believe a reasonable separation between "managers" and "features" is that managers should never be user-facing, so there should be no need for i18n anyway.


magicus:
And yeah, there's no way we're going to send events for this mixin...

kristofbolyai:
What about static final stuff in screens?

magicus:
I did not really think of screens, but that should not be a problem, should it? Static stuff executes when the class is loaded. It is loaded when it is referenced. Screens should only be referenced from features. So I18n should be available.

I can't see a reason to reference Screens from managers. Once again, managers should not contain "user fronting" code.

kristofbolyai (dismissed):
This seems fine. How do you know it is the transformer what is slow specifically with managers? Could it be WebManager only?

magicus:
Wtf happened now? :( I somehow messed up a a merge which turned into a rebase.

kristofbolyai (approved):

PR #295, opened 2022-08-12
Fix isFoil mixin signature

magicus:


magicus:
Yes, `B` is not boolean. And there was an argument. Sloppy me. :(

kristofbolyai (approved):

kristofbolyai:
Are you using auto complete for this kind of work?

kristofbolyai (approved):

magicus:
No..? The Minecraft IntelliJ plugin validates signatures (sort-of) for me, but I have not gotten it to auto complete. Am I missing something? What I usually do is to check that the signature matches by clicking on the little redirect icon in the gutter. If the signature is correct, it takes me to the corresponding method in Mojangs code.

magicus:
But I'd looove for the Minecraft plugin to have like, you right-click on a method in Minecraft, and select "Create a mixin for this method"...

kristofbolyai:
image

kristofbolyai:
This comes up after CTRL + Space for me.

kristofbolyai:
image

magicus:
Hm, okay, but if you select one of those, do you get the signature as well? I do get the same, but when I select one I only get the method name, and that is kind of trivial... It's the signature that's the PITA.

kristofbolyai:
> Hm, okay, but if you select one of those, do you get the signature as well? I do get the same, but when I select one I only get the method name, and that is kind of trivial... It's the signature that's the PITA.

You get the signature when it is needed. When there is only 1 method with that name, the name is enough in the mixin. When there is a need for full signature, it auto completes that.

magicus:
Hm. I prefer to have the full signature. If that ever changes, I want to have an early failure.

PR #303, opened 2022-08-13
Fixes a crash on unknown itemtypes (e.g. tomes)

HighCrit:
I think we should add the texture and itemtype entry when we have/use them. Which is not the case currently.


common/src/main/java/com/wynntils/wc/custom/item/UnidentifiedItemStack.java
```diff
@@ -30,7 +29,7 @@
 
 public class UnidentifiedItemStack extends WynnItemStack {
     private final List<Component> tooltip;
-    private ItemType itemType;
+    private Optional<ItemType> itemType;
```
kristofbolyai: I think optionals are not meant to be fields. Make this an ItemType, but keep `getItemType` an Optional (and return empty if null)

magicus (approved):
LGTM

Issue #304, opened 2022-08-13
Combine auto progress with NPC dialogue overlay


kristofbolyai:

magicus:
You mean if the overlay is disabled while it is showing a dialogue? What did you expect to happen? 😮  The next dialogue screen sent should work as normal. Does it not?

kristofbolyai:
Nothing appeared in chat.

kristofbolyai:
It might be because of my PR though, but I doubt. The thing should not depend on dialog extraction being disabled, but rather if it is being "consumed". See #305.

magicus:
I'm still not sure what you mean, though. Did all NPC chat disappear, like, forever and always, if you do not have the overlay enabled? Or did you disable the overlay while in the middle of a dialogue, and that messed up that current NPC interaction?

kristofbolyai:
Iirc, if you do not have the overlay enabled, but the feature still enabled dialog extraction, then messages wont be sent in chat (and since we don't have an overlay, we cannot see it). We would need to look at the users of the NPC dialogue, and send it to chat if no-one is using it.

magicus:
Agree, I thought this was the case. If it's not, then it is definitely a bug. I'll look more closely into it.

magicus:
I fail to reproduce by turning on and off the overlay and turning on and off the actual feature. I'll look into the code to see if there is something else that can cause us to lose dialogues.

magicus:
Yes, looks like this can happen if you turn on auto progress but not NPC overlay.

magicus:
I think I would like to solve this by folding the auto progress functionality into the NpcOverlay feature itself. I have a hard time seeing how the autoprogress could ever work fine if we do not pick out the dialogue and present it in an overlay. And I frankly doubt that anyone would ever want to do that.

magicus:
@kristofbolyai There is also a bug in auto progress, where it got into a bad state of being disabled, but still showed "0 seconds left" all the time. I triggered it by enabling and disabling Npc overlay, auto progress, etc, in some random order.

![image](https://user-images.githubusercontent.com/6209108/187198567-b3ff4f6a-671a-4038-895c-9116309093c0.png)

kristofbolyai:
> @kristofbolyai There is also a bug in auto progress, where it got into a bad state of being disabled, but still showed "0 seconds left" all the time. I triggered it by enabling and disabling Npc overlay, auto progress, etc, in some random order.
> 
[...]

PR #305, opened 2022-08-13
Features: Add dialog auto-progress feature

kristofbolyai:
image

"Time until auto progress" is only displayed for overlay, as it is waiting on #304. If this is merged before, it should be fixed when that issue is fixed.


common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -73,7 +76,7 @@ public final class ChatModel extends Model {
             Pattern.compile(" +§[47]Press §r§[cf](SNEAK|SHIFT) §r§[47]to continue§r$");
     private static final Pattern EMPTY_LINE_PATTERN = Pattern.compile("^\\s*(§r|À+)?\\s*$");
 
-    private static boolean extractDialog = false;
+    private static final Set<Feature> dialogExtractionDependents = new HashSet<>();
```
Incompleteusern, line 79: Isn't there already a dependency system by `ManagerRegistry?` Why do we need a new one here?

kristofbolyai: This one is an inner dependency of a manager (model)

Incompleteusern: What do you mean by "inner dependency"?

kristofbolyai: So ChatModel handles a few stuff related to chat, and npc dialog extraction is a optional thing it can do. This is one way to implement it but @magicus designed this, and I am not sure how it is meant to be used.

magicus: Do you think we can extract this so we can use the ManagerRegistry dependency system? That'd be far nicer. Maybe we can split it into a ChatModel and NpcDialogueModel, and have the ChatManager pass "chat screens" to the latter, if it is enabled?

magicus: (I'm not really fond of this `extractDialog` solution, there just was nothing better available when I wrote it)

kristofbolyai: I tried it but the code was too mixed together. Perhaps you could do this better.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -126,12 +130,27 @@ private void renderDialogue(PoseStack poseStack, String currentDialogue) {
...
+            Optional<Long> millisecondsUntilProgress =
+                    NpcDialogAutoProgressFeature.INSTANCE.millisecondsUntilProgress();
+
+            List<TextRenderTask> renderTaskList;
+
+            renderTaskList = millisecondsUntilProgress
+                    .map(timeUntilProgress -> List.of(
+                            new TextRenderTask("§cPress SNEAK to continue", renderSetting),
+                            new TextRenderTask(
+                                    ChatFormatting.GREEN + "Auto-progress: "
```
magicus, line 143: What about having a bar that's filling up instead? If you start with 2 seconds and just display integer numbers of remaining seconds, it gives very little feedback.

kristofbolyai: Sure

common/src/main/java/com/wynntils/features/user/NpcDialogAutoProgressFeature.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private final Runnable sendShiftRunnable = () -> {
+        McUtils.sendPacket(new ServerboundPlayerCommandPacket(
+                McUtils.player(), ServerboundPlayerCommandPacket.Action.PRESS_SHIFT_KEY));
+    };
+
+    private ScheduledFuture<?> lastScheduledFuture = null;
+
+    @Config
+    public static int dialogAutoProgressTime = 2000; // Milliseconds
```
magicus: I think the time should scale with the length of the message. Possibly with a constant offset. So instead two variables:
minimumTimeBeforeProgress = 1500;
extraTimePerWord = 50; 
and count the number of words (or spaces) in the dialog.
(I have no idea if 50 is a good number, you'll have to check that and tweak it a bit, I guess)

magicus:
If we have the NpcDialogue overlay turned off, do you think we should signal the auto progress to the user in some other way? Perhaps rewriting the chat "screen"? Or is it just a lot of work for nothing, since all reasonable users will want to keep that enabled anyway? It will look a bit weird, though, since the chat screen will say "Press SNEAK to continue".

kristofbolyai:
> If we have the NpcDialogue overlay turned off, do you think we should signal the auto progress to the user in some other way? Perhaps rewriting the chat "screen"? Or is it just a lot of work for nothing, since all reasonable users will want to keep that enabled anyway? It will look a bit weird, though, since the chat screen will say "Press SNEAK to continue".

With npc dialog extraction, we send whatever message we want? So we can just append this to "Press Sneak..."

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: Ok, I'll have a look at it later. Let's do it this way for now.

magicus:
Yes, but if you disable Npc Dialogue overlay, you won't get .... oh. I see, you just added so you get Dialogue extraction anyway. But if you do that, you need to find some other way to publish the dialog to the user. You can re-send it as chat message, but then it will be like the old-style chat, slowly disappearing up the chat history.

The "flash a chat screen, complete with history" thing that Wynncraft does is not something we can both keep and modify. So if you disabled extraction, you'll get the dialogue kept at the bottom of the screen, as with vanilla. But then you can't do anything with the dialog.

kristofbolyai:
> Yes, but if you disable Npc Dialogue overlay, you won't get .... oh. I see, you just added so you get Dialogue extraction anyway. But if you do that, you need to find some other way to publish the dialog to the user. You can re-send it as chat message, but then it will be like the old-style chat, slowly disappearing up the chat history.
> 
> 
> 
> The "flash a chat screen, complete with history" thing that Wynncraft does is not something we can both keep and modify. So if you disabled extraction, you'll get the dialogue kept at the bottom of the screen, as with vanilla. But then you can't do anything with the dialog.

You can actually do smth like Wynn does. Check out what notification manager does when overlay is not enabled (for editing).

magicus:
But then I'll need to keep a chat history, if you mean I should resend everything. I'd rather implement a better chat handler.

kristofbolyai:
> But then I'll need to keep a chat history, if you mean I should resend everything. I'd rather implement a better chat handler.
> 
> 

You can send a message, then update it, which makes it latest message again. It should be fine for quests.

magicus:
I don't think I understand what you mean. But if you have a plan on how to fix this, please do so.

magicus:
Just to be clear: Getting the delay correlated to the length of the text is the one blocking issue as I see it. A progress bar is "nice to have" but can be put in later. And the extraction weirdness if you're not using the overlay is not that relevant. If you are not using overlays for NPC dialogs, you are going to suffer in sooo many ways anyway.

kristofbolyai:
> Just to be clear: Getting the delay correlated to the length of the text is the one blocking issue as I see it. A progress bar is "nice to have" but can be put in later. And the extraction weirdness if you're not using the overlay is not that relevant. If you are not using overlays for NPC dialogs, you are going to suffer in sooo many ways anyway.

Another blocking factor is me being on vacation 😉

common/src/main/java/com/wynntils/features/user/NpcDialogAutoProgressFeature.java
```diff
@@ -0,0 +1,83 @@
...
+                McUtils.player(), ServerboundPlayerCommandPacket.Action.RELEASE_SHIFT_KEY));
+
+        if (event.getCodedDialog() == null) return;
+
+        lastScheduledFuture = executor.schedule(sendShiftRunnable, dialogAutoProgressTime, TimeUnit.MILLISECONDS);
+    }
+
+    @SubscribeEvent
+    public void onKeyInput(KeyInputEvent event) {
+        if (event.getKey() != GLFW.GLFW_KEY_SPACE) return;
```
HighCrit: Should this not be a user definable key?

HighCrit (changes requested):

common/src/main/java/com/wynntils/features/user/NpcDialogAutoProgressFeature.java
```diff
@@ -0,0 +1,92 @@
...
+    public static NpcDialogAutoProgressFeature INSTANCE;
+
+    @RegisterKeyBind
+    public final KeyBind cancelAutoProgressKeybind =
+            new KeyBind("Cancel Dialog Auto Progress", GLFW.GLFW_KEY_Y, false, this::cancelAutoProgress);
+
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
+
+    private final Runnable sendShiftRunnable = () -> {
+        McUtils.sendPacket(new ServerboundPlayerCommandPacket(
```
magicus, line 37: I thought it was the action of sneaking that was causing the dialog to progress. Or is this Vanilla's weird way of handling sneak, even if it is remapped?

magicus (approved):

HighCrit (approved):

common/src/main/java/com/wynntils/features/user/NpcDialogAutoProgressFeature.java
kristofbolyai: Yes, this is very likely a source code thing, or perhaps the mapping is "wrong".

PR #306, opened 2022-08-13
Add a lot of "character" functions from legacy, and also fix some bugs

magicus:
I have added a bunch more functions.


magicus:
@kristofbolyai I sorta, kinda did the `ModelDependant` as we talked about. It's not very well done, though. The problem is that I don't like the current model of Feature to have a init-method that will populate a local list. I'd rather see that the Features (and Functions) provide methods declaring their need, like the current `ModelDependant` interface I created here. (And similar for Dependencies.)

But I did not want to start changing all Features before I had buy-in from you. Hence this hacky half-solution.

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -30,6 +33,13 @@ private static void registerFunction(Function<?> function) {
         if (function instanceof ActiveFunction<?> activeFunction) {
             activeFunction.init();
         }
+        // FIXME: This is sort of hacky. We should have these as ActiveFunctions instead,
```
kristofbolyai, line 36: We already have enabled functions stored though?

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -72,6 +75,19 @@ public static void addDependency(Feature dependant, Class<? extends Model> depen
         updateManagerState(dependency);
     }
 
+    public static void addFunctionDependency(
```
kristofbolyai, line 78: I assume this is only here because features don't implement the interface yet?

magicus: Yes, if/when we unify the dependency interface between features and functions, this will go away (or maybe rather the current Feature variant goes away).

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
magicus: So, I cheated a bit here, partially because I was not sure how to continue with the dependency change I proposed.

There is really no point in having a separate DependentFunction the way I did it, I realized half-way through. We need to enable them, to register the manager dependeny. Without enabling, the manager is not necessarily loaded. 

Problem is, I don't know what we really say about non-loaded managers. What happen if I call ActionBarModel.getMana(), even if we have not registered ActionBarModel as a dependency? Will it fail? Will we just get random bad data back? The idea we had for ActiveFunctions with "well, try return what you got and if it's bad, it's bad", is not really working if we try calling a manager/model, and it is allowed to crash since it's not properly enabled.

magicus: So for the time being, I register the manager dependencies when loading the functions. That is no good if we really want to be able to not load all managers. But I think we need more clarity on what to expect if you're interacting with a manager that is not loaded. If that is okay, guaranteed to not crash, but can give dummy data, then this is fine.

But if they might crash, then we still will need to check if we're enabled, and if not, return null or "N/A" or whatever.

kristofbolyai: So manager changes again?

magicus: Yeah... question is what to change. Maybe it's just a declaration, stating that managers must work without crashing, and at most deliver random data, when not enabled.

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
```diff
@@ -25,9 +27,23 @@ public class CharacterManager extends CoreManager {
...
 
+    private static final int[] LEVEL_UP_XP_REQUIREMENTS = {
+        110, 190, 275, 385, 505, 645, 790, 940, 1100, 1370, 1570, 1800, 2090, 2400, 2720, 3100, 3600, 4150, 4800, 5300,
+        5900, 6750, 7750, 8900, 10200, 11650, 13300, 15200, 17150, 19600, 22100, 24900, 28000, 31500, 35500, 39900,
+        44700, 50000, 55800, 62000, 68800, 76400, 84700, 93800, 103800, 114800, 126800, 140000, 154500, 170300, 187600,
+        206500, 227000, 249500, 274000, 300500, 329500, 361000, 395000, 432200, 472300, 515800, 562800, 613700, 668600,
+        728000, 792000, 860000, 935000, 1040400, 1154400, 1282600, 1414800, 1567500, 1730400, 1837000, 1954800, 2077600,
+        2194400, 2325600, 2455000, 2645000, 2845000, 3141100, 3404710, 3782160, 4151400, 4604100, 5057300, 5533840,
+        6087120, 6685120, 7352800, 8080800, 8725600, 9578400, 10545600, 11585600, 12740000, 14418250, 16280000,
+        21196500, 23315500, 25649000, 249232940
```
ryanzhoudev, line 42: Maybe we should calculate the 106 xp requirement instead of hardcoding in case values change in the future?

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
```diff
@@ -136,5 +158,67 @@ public String toString() {
...
+         *
+         * Also check that {@code {@link #getMaxSoulPoints()} >= {@link #getSoulPoints()}},
+         * in which case soul points are already full
+         */
+        public int getTicksToNextSoulPoint() {
+            if (McUtils.mc().level == null) return -1;
+            return 24000 - (int) (McUtils.mc().level.getDayTime() % 24000);
+        }
+
+        public int getTicksToNextManaRegen() {
```
ryanzhoudev: I think we should leave this out since we don't have the MR timer bar in anyway, and mana regen ticks will be removed in 2.0

ryanzhoudev: Or mark as deprecated

ryanzhoudev (dismissed):

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
magicus: Oh, oh... You don't know what hornet's nest you're stirring! :) I've spent a lot of time with spread sheets, reverse engineering formulas, and manual testing. Overall, Wynncrafts functions can be quite erratic. 

I understand your thinking, but I think we do best to keep it hard-coded. 

I can see if I can dig out some links to how these values where extracted, though.

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
magicus: Ok, I'll remove it. It is not used in this PR, and I just copied it over alongside the rest of the functions.

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
ryanzhoudev: I believe 1-105 is all hardcoded without much formula, 106 should just be the sum of everything 1-105

ryanzhoudev (dismissed):

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
magicus: I see. It does not seem worth it to calculate it. If values are changed, we still need to update the array. I added a comment about this observation, though.

ryanzhoudev (dismissed):

common/src/main/java/com/wynntils/wc/model/CharacterManager.java
```diff
@@ -136,5 +162,58 @@ public String toString() {
...
+        public float getXpProgress() {
+            return McUtils.player().experienceProgress;
+        }
+
+        public int getXpLevel() {
+            return McUtils.player().experienceLevel;
+        }
+
+        public int getXpPointsNeededToLevelUp() {
+            return LEVEL_UP_XP_REQUIREMENTS[getXpLevel() - 1];
```
kristofbolyai: Crash if our level is too big? Add safety?

ryanzhoudev (approved):

PR #307, opened 2022-08-13
Create a FilterAdminCommandsFeature

magicus:
This will hide unusable admin commands such as `/wynnproxy` etc from tab expand.

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
@@ -15,7 +15,8 @@ import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class FilterAdminCommandsFeature extends UserFeature {
-    private static final List<String> FILTERED_COMMANDS = Arrays.asList("bungee",
+    private static final List<String> FILTERED_COMMANDS = Arrays.asList(
+            "bungee",
             "connect",
             "galert",
             "gcountdown",
--- a/common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.user;
+
+import com.mojang.brigadier.tree.CommandNode;
+import com.mojang.brigadier.tree.RootCommandNode;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.mc.event.CommandsPacketEvent;
+import java.util.Arrays;
+import java.util.List;
+import net.minecraft.commands.SharedSuggestionProvider;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class FilterAdminCommandsFeature extends UserFeature {
+    private static final List<String> FILTERED_COMMANDS = Arrays.asList(new String[] {
+        "bungee",
+        "connect",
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
```diff
@@ -0,0 +1,51 @@
...
+
+        RootCommandNode<SharedSuggestionProvider> newRoot = new RootCommandNode<>();
+        for (CommandNode<SharedSuggestionProvider> child : root.getChildren()) {
+            if (!FILTERED_COMMANDS.contains(child.getName())) {
+                newRoot.addChild(child);
+            }
+        }
+
+        // We also need to add the arguments for all commands
+        newRoot.addChild(root.getChild("args"));
```
Incompleteusern: What does this do?

magicus: All command literals have a hardcoded dependency on an "args" string argument node. I assume this is something setup by whatever mod they have to inject commands like this (bungee, perhaps). 

In theory, I should recurse over all child nodes and see if they have more argument types to be added. At least, that's what I assume I'm supposed to do. But I could not be bothered. Should Wynncraft drastically change how their admin commands are provided, then the user need to disable this feature until we can fix it.

kristofbolyai: (keep in mind this is only command suggestions, so even if we filter out something we should not, the user doesn't lose much.)

common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
```diff
@@ -0,0 +1,51 @@
...
+            "wcl",
+            "wynnproxy");
+
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onCommandPacket(CommandsPacketEvent event) {
+        RootCommandNode<SharedSuggestionProvider> root = event.getRoot();
+
+        RootCommandNode<SharedSuggestionProvider> newRoot = new RootCommandNode<>();
+        for (CommandNode<SharedSuggestionProvider> child : root.getChildren()) {
+            if (!FILTERED_COMMANDS.contains(child.getName())) {
```
HighCrit, line 42: Wouldn't a set be more performant?

HighCrit (changes requested):

common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
magicus: Maybe. I doubt you can even measure it.

The Rules of Optimizations:
1) Don't Optimize
2) (Only for experts) Don't Optimize Yet.

:-)

But I can agree that it makes things a bit clearer (and that's worth more than premature optimization) if you call things a Set when you just need Set-like behavior.

magicus: ... only there is no `asSet`. :( At least not that I can find.

common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
magicus: It seemed like I could remove it without any bad effects. I don't fully understand the Brigadier structure. But if it works without it, I'll skip it.

common/src/main/java/com/wynntils/features/user/FilterAdminCommandsFeature.java
HighCrit: > ... only there is no `asSet`. :( At least not that I can find.

`new HashSet<String>(Arrays.asList(...))` is the best I could find 👀

HighCrit (approved):

PR #309, opened 2022-08-13
Add command expansion for commands provided by Wynncraft

magicus:
The idea seems sound. But I'm getting stuck in generics-valley-of-death and can't figure out how to solve this without the raw access warning. :(

And it will be a bit of an effort to encode all Wynncraft's commands.


magicus:
I'm satisfied with this patch, at the time. Ideally, all "complex" commands should be encoded like "friend" is done here. Furthermore, even "friend" can be enhanced with suggestion providers.

But. I can't be bothered to do all that right now. This is pure 80/20 Rule -- this solves the most pressing problem of getting the command name tab expanded. And getting the complex arguments supported is a suitable beginners task, that I gladly leave to someone else. :)

magicus:
I fixed `/toggle` as well, since it is a proper pain in the behind...

magicus:
I, eh, ended up doing all the commands properly anyway. :-)

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/user/AddCommandExpansionFeature.java
```diff
@@ -0,0 +1,267 @@
...
+
+        root.addChild(literal("p")
+                .then(argument("msg", StringArgumentType.greedyString()))
+                .build());
+
+        root.addChild(literal("r")
+                .then(argument("msg", StringArgumentType.greedyString()))
+                .build());
+
+        CommandNode<CommandSourceStack> node4 = literal("msg")
```
HighCrit, line 168: Why does this one have a `4` behind the variable name?

common/src/main/java/com/wynntils/features/user/AddCommandExpansionFeature.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    private void addMiscCommandNodes(RootCommandNode root) {
+        root.addChild(literal("report")
+                .then(argument("player", StringArgumentType.string())
+                        .then(argument("reason", StringArgumentType.greedyString())))
+                .build());
+
+        root.addChild(literal("switch")
+                .then(argument("world_num", IntegerArgumentType.integer()))
```
HighCrit, line 183: `WC1` is also a valid input. We should also verify with Zeer0 whether the YT server is accessible with this command and how.

HighCrit (changes requested):

common/src/main/java/com/wynntils/features/user/AddCommandExpansionFeature.java
magicus: Debug code left behind...

common/src/main/java/com/wynntils/features/user/AddCommandExpansionFeature.java
magicus: I'll make it a string instead.

PR #311, opened 2022-08-14
Make DevAuth allow alt account

magicus:


kristofbolyai (approved):

HighCrit:
Is this necessary? I personally removed the line in the jvm args that points to .devauth in the Artemis directory. So it now uses the one in my user home with 3 MC accounts

magicus:
Well, it doesn't hurt. I think our setup by default creates IntelliJ profiles that point to this file.

PR #314, opened 2022-08-14
Make /feature, /function and /config commands in their own right

magicus:
There is no good reason to shoehorn these into `/wynntils`. For one, they are long enough as it is, and without the `/wynntils` prefix, they are a bit shorter. Secondly, now `/wynntils` is more concentrated about Wynntilsy things, like discord links etc.

I was initially worried about clashes with Wyncraft commands, but my two previous PRs (hide admin commands, and make real commands tab expansion aware) have alleviated any such fears.


common/src/main/java/com/wynntils/commands/ConfigCommand.java
```diff
@@ -2,14 +2,18 @@
  * Copyright © Wynntils 2022.
  * This file is released under AGPLv3. See LICENSE for full license details.
  */
-package com.wynntils.commands.wynntils;
+package com.wynntils.commands;
+
+import static net.minecraft.commands.Commands.literal;
```
kristofbolyai: Ew. Do we like static imports?

magicus: In general, no, but I thought it made the complex Brigadier node building much more readable. But if you really dislike it, I can remove them.

kristofbolyai: Isn't it nice to remove it to keep to our standards?

magicus: OK, I'll fix.

kristofbolyai (approved):

PR #315, opened 2022-08-14
Refactor the classes in utils, mc/utils and wc/utils to better locations

magicus:
So, this one is going to look quite massive, and I understand if it can be hard to review. I recommend checking out the branch yourself and having a look in your IDE.

I've been increasingly annoyed at the mess in the "utils" folders. We have been mixing together things that should properly be called Managers or Models (which model a certain aspect of Minecraft or Wynncraft), with pure utility classes (which provide a mixed bag of static functions). Stuff have ended up in `mc` when it should have been in `wc`, or vice versa. The generic `utils` did not contain all generic utility classes, etc.

I've tried to straighten out this. It is perhaps not perfect, but it is sure as h*ll better than what we have right now. Since refactorings like this is always tricky, if they are kept too long, I propose that if this is mostly acceptable to you, we integrate it more or less as it is, and then we fix stuff that did not get right in a follow-up PR. I promise I won't use "but it's already integrated" as an argument, even if the consensus ends up being reverting the majority (or all!) of this PR.


magicus:
To at least give some kind of visual overview, this is how the "utils" directories look in IntelliJ with this patch.

Screenshot 2022-08-14 at 23 54 03

kristofbolyai (dismissed):
Seems okay to me

kristofbolyai (approved):

magicus:
@kristofbolyai Sorry, I did the merge in IntelliJ (due to conflicts) and then Github thinks I need a re-review. (That is at least the pattern I think happens; if you merge from main in the github web GUI, a re-review is not needed... Somewhat irritating...)

Issue #301, opened 2022-08-13
Re-implement old info variables as functions


magicus:
Mostly for the record, here are all info variables that was supported by legacy that are still not implemented. Note that some of the names are aliases.

```
"adps"
"areadps"
"beacdist"
"beacon_distance"
"blocks"
"bpm"
"bps"
"dry_boxes"
"dry_streak_boxes"
"dry_streak"
"e"
"eb"
"emeraldblocks"
"emeralds"
"hp_pot_charges"
"hp_pot_combined"
"hp_pot_max"
"inv_free"
"inv_slots"
"kmph"
"kpm_raw"
"kpm"
"le"
"liquidemeralds"
"loc"
"location"
"manaregen_timer"
"money_desc"
"money"
"mp_pot"
"mr_timer"
"party_count"
"party_owner"
"ping"
"potions_health_charges"
"potions_health_combined"
"potions_health_max"
"potions_mana"
"pouch_free"
"pouch_slots"
"pouch"
"terguild_pref"
"terguild"
"territory_owner_prefix"
"territory_owner"
"uptime_h"
"uptime_m"
"xpm_raw"
"xpm"
"xppm"
```

kristofbolyai:
Pretty much done.

PR #316, opened 2022-08-15
Add horse info functions

ryanzhoudev:
- Horse xp, level, max level, tier


magicus:
Several of these are already implemented in https://github.com/Wynntils/Artemis/pull/306, which is just awaiting a review to be merged.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+import com.wynntils.core.functions.ActiveFunction;
+import com.wynntils.mc.utils.InventoryUtils;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wc.custom.item.WynnItemStack;
+import com.wynntils.wc.custom.item.properties.ItemProperty;
+import net.minecraft.world.item.ItemStack;
+
+public class HorseFunctions {
+
+    public static class horseLevelFunction extends ActiveFunction<String> {
```
magicus: Please use conventional naming, i.e. UpperCamelCase, for classes, even nested classes.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+        @Override
+        public String getValue(String argument) {
+            if (InventoryUtils.findHorseSlotNum() == -1) return "??";
```
magicus: If you want, you can return `null` and it will be interpreted by the info manager as "N/A".

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    public static class horseLevelMaxFunction extends ActiveFunction<String> {
+        @Override
+        public String getValue(String argument) {
+            if (InventoryUtils.findHorseSlotNum() == -1) return "??";
+
+            ItemStack is = McUtils.player().getInventory().getItem(InventoryUtils.findHorseSlotNum());
+            if (!(is instanceof WynnItemStack horse)) return "??";
```
magicus: Seems you could extract this into a `private static WynnItemStack getHorse()` in `HorseFunctions`, to get rid of some duplication.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
ryanzhoudev: I think in this specific area, we should do ??. If you were to use something like (%horse_level%/%horse_level_max%), it would return (N/A/N/A) which I think looks rather funny

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,63 @@
...
+
+            return String.valueOf(horse.getProperty(ItemProperty.HORSE).getLevel());
+        }
+    }
```
magicus, line 44: You also need to declare your dependency on `ItemStackTransformModel`:
```
        @Override
        public List<Class<? extends Model>> getModelDependencies() {
            return List.of(ItemStackTransformModel.class);
        }
```

common/src/main/java/com/wynntils/wc/custom/item/properties/HorseProperty.java
```diff
@@ -0,0 +1,45 @@
...
+    public HorseProperty(WynnItemStack item) {
+        super(item);
+
+        // Parse xp, level, max level, and tier
+        List<String> lore = ItemUtils.getLore(item);
+
+        tier = Integer.parseInt(lore.get(0).substring(7));
+        level = Integer.parseInt(lore.get(1).substring(9, lore.get(1).indexOf('/')));
+        maxLevel = Integer.parseInt(lore.get(1).substring(lore.get(1).indexOf('/') + 1));
+        xp = Integer.parseInt(lore.get(4).substring(6, lore.get(4).indexOf('/')));
```
magicus: Overall I prefer regexp for this kind of matching. It is clearer what is going on, and if the format changes it will fail correctly, instead of extracting random subparts of the strings.

But if you don't feel like redoing it as a regex I'm (reluctantly) okay with it.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,61 @@
...
+    private static WynnItemStack getHorse() {
+        ItemStack is = McUtils.player().getInventory().getItem(InventoryUtils.findHorseSlotNum());
+        if (!(is instanceof WynnItemStack horse)) return null;
+        return horse;
+    }
+
+    public static class HorseLevelFunction extends ActiveFunction<String> {
+        @Override
+        public String getValue(String argument) {
+            if (InventoryUtils.findHorseSlotNum() == -1) return "??";
```
magicus: You can move that part too into getHorse()...

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,54 @@
...
+        if (InventoryUtils.findHorseSlotNum() == -1) return null;
+        ItemStack is = McUtils.player().getInventory().getItem(InventoryUtils.findHorseSlotNum());
+        if (!(is instanceof WynnItemStack horse)) return null;
+        return horse;
+    }
+
+    public static class HorseLevelFunction extends ActiveFunction<String> {
+        @Override
+        public String getValue(String argument) {
+            if (getHorse() == null) return "??";
```
magicus: Please store the result of getHorse() in a local variable, and use that instead. There is an optimization win (we don't have to go through all the inventory slots twice to locate the horse), but the big win is in stability. If something where to change in the background, we could fail even if you did check previously.

In general, this is something you always should do in programming. It's called [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) -- "time of check to time of use" and is a classical source of race condition.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,54 @@
...
+import com.wynntils.mc.utils.InventoryUtils;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wc.custom.item.WynnItemStack;
+import com.wynntils.wc.custom.item.properties.ItemProperty;
+import net.minecraft.world.item.ItemStack;
+
+public class HorseFunctions {
+
+    private static WynnItemStack getHorse() {
+        if (InventoryUtils.findHorseSlotNum() == -1) return null;
```
magicus: Here is also another source of TOCTOU. Store the slot num in a local variable instead.

magicus (dismissed):

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -160,5 +161,11 @@ public static void init() {
         registerFunction(new EnvironmentFunctions.MemMaxFunction());
         registerFunction(new EnvironmentFunctions.MemUsedFunction());
         registerFunction(new EnvironmentFunctions.MemPctFunction());
+
+        registerFunction(new HorseFunctions.HorseLevelFunction());
```
kristofbolyai, line 201: Can we get all registries in alphabetical order?

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static WynnItemStack getHorse() {
+        int horseSlot = InventoryUtils.findHorseSlotNum();
+        if (horseSlot == -1) return null;
+        ItemStack is = McUtils.player().getInventory().getItem(horseSlot);
+        if (!(is instanceof WynnItemStack horse)) return null;
+        return horse;
+    }
+
+    public static class HorseLevelFunction extends ActiveFunction<String> {
```
kristofbolyai: Is this really what we want to call "active"? This rather fits the DependantFunction concept (or whatever it is called now)

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,96 @@
...
+        @Override
+        public String getValue(String argument) {
+            WynnItemStack horse = getHorse();
+            if (horse == null) return "??";
```
kristofbolyai: Can we add a generic UNKNOWN string const (for every function)

common/src/main/java/com/wynntils/functions/HorseFunctions.java
kristofbolyai: (This applies to every horse function.

kristofbolyai: What I mean is I don't see any "caching", just "active" fetching.

common/src/main/java/com/wynntils/wc/custom/item/properties/ItemProperty.java
```diff
@@ -20,6 +20,7 @@ public abstract class ItemProperty {
     public static final Class<TeleportScrollProperty> TELEPORT_SCROLL = TeleportScrollProperty.class;
     public static final Class<SkillPointProperty> SKILL_POINT = SkillPointProperty.class;
     public static final Class<SkillIconProperty> SKILL_ICON = SkillIconProperty.class;
+    public static final Class<HorseProperty> HORSE = HorseProperty.class;
```
kristofbolyai: Alphabetical order

common/src/main/java/com/wynntils/wc/custom/item/properties/HorseProperty.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class HorseProperty extends ItemProperty {
+
+    private final Pattern HORSE_PATTERN = Pattern.compile(
+            "§7Tier (\\d)§6Speed: (\\d+)/(\\d+)§6Jump: \\d+/\\d+§5Armour: None§bXp: (\\d+)/100(?:§cUntradable Item)?(:?§7Name: (.+))?");
+    private int xp = -1;
+    private int level = -1;
+    private int maxLevel = -1;
+    private int tier = -1;
+    private String name = null;
```
kristofbolyai: empty string as default value?

common/src/main/java/com/wynntils/functions/HorseFunctions.java
magicus: So, yes, it is a DependantFunction, not strictly an "active" one. But as we talked about in the previous PR, this does not really make sense. We need to load the manager, before it can start giving real values. So there is really only two kinds of Functions, those that can return a value at any time, with no prereq's, (like `clock`), and those which has some kind of dependency. 

If they need to call into a Manager/Model, then this needs to be loaded.

magicus: The alternative is to always load the ItemStackTransformerModel. But if we want to be able to *not* have it always loaded, we somehow need to say that it needs loading for this function. And you can't really load and unload managers willy-nilly each time you want to read a value from a function...

kristofbolyai: But then, the is cached field (idk name) is not updated here? And it does not make sense to so... Basically it would mean every tick it is "updated"

common/src/main/java/com/wynntils/functions/HorseFunctions.java
magicus: That is already present, and it is `null`. This is interpreted as a unknown/invalid (currently represented by `N/A` in the UI).

common/src/main/java/com/wynntils/wc/custom/item/properties/HorseProperty.java
magicus: Yes, agree. Strings should be "", not null, if at all possible.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
kristofbolyai: We return "??" though?

common/src/main/java/com/wynntils/functions/HorseFunctions.java
magicus: I'm not sure I follow you completely here, but are you worried about performance issues? In the old code, info boxes were updated every 4th tick. I doubt it will make much of an impact to query the inventory for horse data 5 times a second. If we should start caching this, we'll run into the old:

There's only two hard things in computer science:
* Naming things, and
* Cache invalidation

kristofbolyai: Am I not up to date on ActiveFunction structure? Does it not have a "is cached" field of some sort? We don't seem to update that

magicus: So in a way I think what you're saying is that there is a conceptual difference between functions that has a `@SubscribeEvent` and gets updated whenever that event happens, and in between that returns the cached value, and a function like this, that goes and fetches the value "fresh" each time it is called, but does so from a Model that needs to be enabled first. I agree with that. But it's just an "implementation detail". 

From the PoV of the code calling the function, both of them has the property that they need to be "enabled" to give valid data back, even if the reason for this differs. (That was why I called them EnableableFunctions originally :))

kristofbolyai: Okay. I think I get it now, but I don't want to reason on something I'm not yet fully familiar with. I will get back to this.

magicus:
I checked https://github.com/Wynntils/Artemis/issues/301. There are a bunch of aliases for these functions as well:
```
"h_lvl"
"h_mlvl"
"h_tier"
"h_xp"
```

You add aliases by something like this:
```
        @Override
        public List<String> getAliases() {
            return List.of("h_lvl");
        }
```

common/src/main/java/com/wynntils/functions/HorseFunctions.java
magicus: Yes, currently in the PR. I suggested above to @DonkeyBlaster that he should return `null` instead. I'm guessing he hasn't been around to it yet.

ryanzhoudev: I mentioned [here](https://github.com/Wynntils/Artemis/pull/316#discussion_r945554495) that having N/A in those two specific functions may look weird, I can change it though if you don't think it will be an issue.

magicus: It is probably better then to replace the "N/A" value in `FunctionManager.getSimpleValueString()` to `??`, and return null here. I agree that it is probably more readable to have the "??" than the "N/A" string.

common/src/main/java/com/wynntils/functions/HorseFunctions.java
magicus: I'm open for improvements on the Function design as well. It is not fully clear to me how to solve this in the best way. I made an attempt that I think is okay, but it does not shine like the only obvious way...

common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -0,0 +1,122 @@
...
+            return List.of(ItemStackTransformModel.class);
+        }
+
+        @Override
+        public List<String> getAliases() {
+            return List.of("h_lvl");
+        }
+    }
+
+    public static class HorseLevelMaxFunction extends ActiveFunction<String> {
```
magicus:
```suggestion
    public static class HorseLevelMaxFunction extends ActiveFunction<Integer> {
```
and then you can skip the String.valueOf.

Same goes for all the other horse functions as well, except `HorseNameFunction`.

magicus (dismissed):
Yeah, ship it!

magicus:
@DonkeyBlaster ... but you must turn it into a real PR instead of a Draft, and resolve merge conflicts. I'll re-approve after that.

magicus (approved):

PR #317, opened 2022-08-15
Our commands should only be available if we're on a Wynncraft server

magicus:
Thanks to @coehlrich for noticing this.


kristofbolyai (dismissed):

HighCrit (changes requested):
No translation lines?

common/src/main/java/com/wynntils/core/commands/ClientCommandManager.java
```diff
@@ -59,16 +58,15 @@ private static void registerCommand(CommandBase command) {
     }
 
     @SubscribeEvent
```
coehlrich: I think you forgot to remove this line

common/src/main/java/com/wynntils/features/user/CommandsFeature.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.wynntils.features.user;
+
+import com.wynntils.core.commands.ClientCommandManager;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.mc.event.ChatSentEvent;
+import com.wynntils.wc.utils.WynnUtils;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class CommandsFeature extends UserFeature {
```
HighCrit, line 13: Also do we want this feature to be enabled/disabled by the user?

common/src/main/java/com/wynntils/core/commands/ClientCommandManager.java
magicus: Yep. Nice spotted!

common/src/main/java/com/wynntils/features/user/CommandsFeature.java
magicus: Yes, I think it makes sense. Once again, if we do something bad. Or if, let's say Wynncraft introduces a `/feature` command, the user can disable this. Most of the features that the user can disable makes no real sense (in the end, if you disable all of them you might as well not use Wynntils :)) but I think it makes sense that each and every bit of modification we do can be individually turned off.

magicus:
> No translation lines?

I keep forgetting those. :( '/feature list' shows I'm not the only one. ;-) We should have a check for this in FeatureRegistry, really.

kristofbolyai:
> > No translation lines?
> 
> 
> 
> I keep forgetting those. :( '/feature list' shows I'm not the only one. ;-) We should have a check for this in FeatureRegistry, really.

I would rather do this in spotless with a custom module

magicus:
> I would rather do this in spotless with a custom module

Sounds like the proper place, yeah. But also much more work. What did you call it? "Sounds nice and really complicated."? :-D If you want to do it (after your vacation, no hurry) you're more than welcome. Also, ideally we'd also check functions, config variables and overlays.

HighCrit (approved):

PR #319, opened 2022-08-15
Cache gradle data

magicus:
While the gradle action caches downloaded dependencies, it does not cache Architectury/loom data, which needs to be re-generated each and every build.

Also, if we run the format and build check consecutively we can utilize the gradle cache, source code and jdk that we have already downloaded.

With these changes, formatting and building time goes down to 2 minutes or less.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,6 @@ jobs:
           java-version: 17
 
       - name: Cache Gradle data
-          id: cache-gradle
           uses: actions/cache@v3
           with:
             path: .gradle
@@ -60,7 +59,6 @@ jobs:
           java-version: 17
 
       - name: Cache Gradle data
-          id: cache-gradle
           uses: actions/cache@v3
           with:
             path: .gradle
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,13 @@ jobs:
           distribution: "temurin"
           java-version: 17
 
+      - name: Cache Gradle data
+          id: cache-gradle
+          uses: actions/cache@v3
+          with:
+            path: .gradle
+            key: ${{ runner.os }}-gradle--${{ hashFiles('**/build.gradle', '**/settings.gradle', '**/gradle.properties') }}
+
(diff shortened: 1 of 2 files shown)


magicus:
(I still need to verify that it works as it should in a PR)

magicus:
@DevScyu It seems that the checks done on this PR is not using the new .github/workflow file, but the old one..? That surprised me a bit. Do you know if this is a setting on the repo? Or if there is a way to trigger tests with the code that is actually in the PR?

kristofbolyai:
Merge on your fork, then make a PR there

kristofbolyai:
It is a github universal thing related to pull request and pull request target triggers.

magicus:
Thanks. One would think I should know about such things, but the GHA setup we have at OpenJDK (which is part of my day-time work) has a quite different setup where we tweak it heavily to fit our needs, so the concepts I've learnt do not apply here.

magicus:
@kristofbolyai Not sure I understood you correctly. I made a PR from my branch on my fork, to main on my fork. https://github.com/magicus/Artemis/pull/1
But it still does not run the new GHA workflow. So maybe that was not what you meant?

magicus:
... and not actually merging the code into my `main` and opening a PR from that to here either... *sigh*

kristofbolyai:
> @kristofbolyai Not sure I understood you correctly. I made a PR from my branch on my fork, to main on my fork. https://github.com/magicus/Artemis/pull/1
> 
> But it still does not run the new GHA workflow. So maybe that was not what you meant?

Merge this to Magicus fork Main.  Then open a Pr to magicus main from a magicus branch.

magicus:
Ah, now I think I understand. I need to "implement" this on my `main`, and then test it by making a new dummy PR.

magicus:
Ok, now I have verified that it works, including the Spotless apply commit:
https://github.com/magicus/Artemis/actions/runs/2861294049

kristofbolyai (approved):
Seems okay. I think there might be a better way, but it did not work for me, and this is good enough.

PR #322, opened 2022-08-15
Rename package wc to wynn

magicus:


kristofbolyai (approved):

magicus:
@DevScyu Did my GHA workflow changes mess with settings on the repo? It seems it is waiting for a job named "build", which is now named "format-and-build" or "Spotless and Build" (depending on if it is the job name or the workflow name).

magicus:
There also seems to be no way for me to override his, like I can if there's a missing reviewer. :(

PR #325, opened 2022-08-16
Add ability to log items to console

HighCrit:
I basically have to write this everytime I do something with items. So I figured why not make a feature of it. Press INSERT (default) while hovering an item in a container or when holding an item in your main hand.

image


magicus (approved):
👍

PR #326, opened 2022-08-16
Add callback for key presses in inventory to KeyBinds, and cleanup

### LINKED ISSUE #153: KeyHolders do not work in inventories

magicus:
Including refactoring/cleanup of KeyBinds, like more consistent naming, and simplify KeyBind constructor.

This fixes https://github.com/Wynntils/Artemis/issues/153.


kristofbolyai:
This kinda needs to fix inventory keybinds blocking non inventory keybinds (space bind in inv does not work for jump)

magicus:
I'm not sure what you mean. If you select a key that has a different function in gameplay, you can't check for it in inventory? Is the InventoryKeyPressEvent not even sent for that key?

kristofbolyai:
> I'm not sure what you mean. If you select a key that has a different function in gameplay, you can't check for it in inventory? Is the InventoryKeyPressEvent not even sent for that key?

Other way around. They do not work in gameplay when you bind them to do smth in inventory, even when inventory is not open. I am afraid forge had to write a whole system for keybind conflicts. I wonder how we should deal with that

common/src/main/java/com/wynntils/core/keybinds/KeyBind.java
```diff
@@ -0,0 +1,66 @@
...
+        return keyMapping;
+    }
+
+    public void onPress() {
+        if (onPress == null) return;
+
+        onPress.run();
+    }
+
+    public void onInventoryPress(Slot hoveredSlot) {
```
HighCrit, line 53: Do we always want the hovered slot?

Incompleteusern: I think we do, since some Keybinds such as item compare use hovered item information

magicus: Well, implementations are free to ignore it. :) But it makes sense to provide it for those that care.

magicus:
So I've solved this the same way as we did for all Features. This is not really new code, I just moved it to the KeyBindManager. 

We do not register a new key bind with Minecraft. We listen to the onKeyPressImInventory event and check if the key is being pressed down at that time.

magicus:
It is possible that this can be solved more elegantly, but now at least all Features only has to implement a lambda, and if we ever want to change the method how we detect the key press, we only need to do so in one place.

common/src/main/java/com/wynntils/core/keybinds/KeyBind.java
HighCrit: I meant it more like. Don't we want to have the event accessible for implementations that need it.

magicus: There's not much else in the event. KeyCode and scanCode we've already matched for. That leaves `modifiers`. If we ever feel like we need them, we can pass them on to the lambda as well.

HighCrit (approved):

PR #327, opened 2022-08-16
(hopefully) fixes the intelligence skillpoint item disappearing

byBackfish:
(hopefully) fixes the intelligence skillpoint item disappearing under certain circumstances.

image
Context: https://canary.discord.com/channels/394189072635133952/946471329857675307/1009132445330845726

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/item/IntelligenceSkillPointsItemStack.java
+++ b/common/src/main/java/com/wynntils/wynn/item/IntelligenceSkillPointsItemStack.java
@@ -62,6 +62,9 @@ public class IntelligenceSkillPointsItemStack extends WynnItemStack {
         for (int j = 0; j < 4; j++) {
             SpellType spell = SpellType.forClass(characterInfo.getClassType(), j + 1);
 
+            if (spell == null) {
+                continue;
+            }
             if (spell.getUnlockLevel(1) <= level) {
                 // The spell has been unlocked
                 int nextUpgrade = spell.getNextManaReduction(level, intelligencePoints);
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/wynn/item/IntelligenceSkillPointsItemStack.java
```diff
@@ -62,6 +62,9 @@ private List<Component> getManaTables() {
         for (int j = 0; j < 4; j++) {
             SpellType spell = SpellType.forClass(characterInfo.getClassType(), j + 1);
 
+            if (spell == null) {
```
HighCrit, line 65: This does fix the NPE but doesn't fix `spell` being `null`. Which shouldn't be as we have fail over spells with class `None`.

HighCrit (changes requested):

common/src/main/java/com/wynntils/wynn/objects/SpellType.java
```diff
@@ -156,6 +157,7 @@ public static SpellType forClass(ClassType classRequired, int spellNumber) {
                 return spellType;
             }
         }
+        WynntilsMod.error("SpellType not found for " + classRequired + " | Spell: " + spellNumber);
         return null;
```
HighCrit, line 161: I.e. we should prevent this being reachable and throw an illegal argument exception if we do somehow reach it.

byBackfish:
Fixed. Turns out the user was using `autojoin` (`/toggle autojoin`) and it therefore skipped the class screen, resulting in artemis not knowing what class he was

HighCrit:
> Fixed. Turns out the user was using `autojoin` (`/toggle autojoin`) and it therefore skipped the class screen, resulting in artemis not knowing what class he was

Still a problem that needs solving 👀

kristofbolyai:
> > Fixed. Turns out the user was using `autojoin` (`/toggle autojoin`) and it therefore skipped the class screen, resulting in artemis not knowing what class he was
> 
> 
> 
> Still a problem that needs solving 👀

It will be in a new PR shortly.

PR #329, opened 2022-08-16
Fix currentCharacter being null when using autojoin

byBackfish:
Fixes various bugs containing the `currentCharacter`, like the intelligence skill point item disappearing


common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -53,6 +55,12 @@ public static boolean hasCharacter() {
     }
 
     public static CharacterInfo getCharacterInfo() {
+        if (currentCharacter == null) {
+            McUtils.sendMessageToClient(new TextComponent(
```
HighCrit: This could spam the user imo this should be send only once at world join.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -53,6 +55,12 @@ public static boolean hasCharacter() {
...
+                            "Could not find your class type. Disable your autojoin (/toggle autojoin) and try again.")
+                    .withStyle(ChatFormatting.RED));
+            currentCharacter = new CharacterInfo(ClassType.None, false, McUtils.player().experienceLevel, 0);
```
HighCrit, line 66: If we discover it's `null` we should consider looking in the compass menu, using something similar to FakeInventory in legacy, for the actual current class. @magicus @kristofbolyai any thoughts?

HighCrit (dismissed):

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
kristofbolyai: IMO, FakeInventory is prone to being buggy, if it is not a must to use it, let's don't.

byBackfish: yea we should. I mentioned this in the artemis discussions channel, and got told there is a fix for that planned, and that i should just fix it as it is right now

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
byBackfish: Alright yea. Wynncraft itself sends a "you are using autojoin" message, so ill just look for that and then send the wynntils one i guess

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
byBackfish: > IMO, FakeInventory is prone to being buggy, if it is not a must to use it, let's don't.

how else would you want to fix that? We could tell the user to open that menu, but thats also not the best idea

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
byBackfish: > This could spam the user imo this should be send only once at world join.

but in what way would it spam the user? Wouldnt it only send it to the user the first time he uses autojoin? The other times he would have to use the /class menu, and that would create the normal CharacterInfo

byBackfish: Nevermind. The message is also sent the second the player does /class

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
HighCrit: > IMO, FakeInventory is prone to being buggy, if it is not a must to use it, let's don't.

We could also execute `/class` when the user joined using autojoin and then `/toggle autojoin` to turn it off. After which  we make it a Wynntils feature that auto-selects the last class known to us. I believe legacy does this by slot id.

byBackfish: Wynntils could also forcefully do `/toggle autojoin`, but i dont really know if that is a good solution

byBackfish: Yea, i think doing `/class` automatically is good enough, not sure tho

magicus: I think Legacy requires you to run /class at least once when having autojoin, then we save this info an assumes it is the same. This will break if you ever log in without Wynntils, changes class with /class, and then goes back to Wynntils. So I'm not sure that either is a good idea.

magicus: I think we could be justified for doing fakeinventory for determining class when we connected using autojoin. And then we can offer a WynntilsAutojoin feature so if people prefer to use our solution we can do it better.

magicus: We will need to do fake inventory anyway to get quest info etc. So we can't avoid it.

HighCrit: > I think Legacy requires you to run /class at least once when having autojoin, then we save this info an assumes it is the same. This will break if you ever log in without Wynntils, changes class with /class, and then goes back to Wynntils. So I'm not sure that either is a good idea.

Well then it just selects the wrong class very infrequently. 🤷

byBackfish: > We will need to do fake inventory anyway to get quest info etc. So we can't avoid it.

Yea, and Wynntils would only display that fake inventory when the player joins with autojoin. And even then, the fakeinventory would only be there for the first few seconds of joining. I think thats the best idea

byBackfish: > > I think Legacy requires you to run /class at least once when having autojoin, then we save this info an assumes it is the same. This will break if you ever log in without Wynntils, changes class with /class, and then goes back to Wynntils. So I'm not sure that either is a good idea.
> 
> Well then it just selects the wrong class very infrequently. 🤷

but whats really against using a fakeinventory? Yea it can be buggy sometimes, but its still one of the best ways to the current class

P0keDev: > I think Legacy requires you to run /class at least once when having autojoin, then we save this info an assumes it is the same. This will break if you ever log in without Wynntils, changes class with /class, and then goes back to Wynntils. So I'm not sure that either is a good idea.

we could include some logic to check if the current class "matches" the one stored, and if it seems like it doesn't, we could give the user a warning message telling them they should /class to ensure wynntils will function correctly

P0keDev: imo the less invasive we are the better, so forcing /class every time or creating our own auto join functionality isn't desirable

magicus: I'm not sure why you are so afraid of fake inventory. I have a prototype working in Artemis right now, and it's looking very good. We send packets to click with the compass (just like how we send packets to click with the horse slot for the horse mount keybind feature). I intercept the window open packet and the slot contents packet, and then I send a "close container" packet back. Wham, bam, done. Nothing is even shown to the user since I cancel the packets. 

I can possibly imagine this going awry if we do it while the user is making some weird stuff. But if we do this automatically as soon as we join the world, I have a hard time seeing what could go wrong.

HighCrit: > but whats really against using a fakeinventory? Yea it can be buggy sometimes, but its still one of the best ways to the current class

An alternative would be to scan the current inventory for items and look which class is listed behind the green checkmark on weapon reqs.

magicus: But that will fail if you hold no weapons. 

Once more: reading the compass menu is 100% accurate. Why is this such a big deal to not do that, so that you all are thinking up a lot of probabilistic methods instead?

byBackfish: How does wynntils handle this atm? Does it also just use a fakeinventory with the compass?

HighCrit: > How does wynntils handle this atm? Does it also just use a fakeinventory with the compass?

https://github.com/Wynntils/Artemis/pull/329#discussion_r947196160

Incompleteusern: > I'm not sure why you are so afraid of fake inventory. I have a prototype working in Artemis right now, and it's looking very good. We send packets to click with the compass (just like how we send packets to click with the horse slot for the horse mount keybind feature). I intercept the window open packet and the slot contents packet, and then I send a "close container" packet back. Wham, bam, done. Nothing is even shown to the user since I cancel the packets.
> 
> I can possibly imagine this going awry if we do it while the user is making some weird stuff. But if we do this automatically as soon as we join the world, I have a hard time seeing what could go wrong.

I feel that fake inventory needs to be more stable and should fail safer; I think someone has lost a mythic to fake inventory in legacy

kristofbolyai: I fixed that bug. Took a few years  though..

kristofbolyai (approved):
I think this is fine as is. We will fix the autojoin issues later.

HighCrit:
> I think this is fine as is. We will fix the autojoin issues later.

This issue isn't too pressing to have fixed immediately. Therefore, I think we should solve it properly, supporting autojoin. Am also not 100% okay with a getter having behaviour besides getting/returning data. But if either @magicus or @P0keDev also approves I'll dismiss my review.

kristofbolyai:
> > I think this is fine as is. We will fix the autojoin issues later.
> 
> 
> 
> This issue isn't too pressing to have fixed immediately. Therefore, I think we should solve it properly, supporting autojoin. Am also not 100% okay with a getter having behaviour besides getting/returning data. But if either @magicus or @P0keDev also approves I'll dismiss my review.

I think this fix is alright though. We need to make sure manager getters don't cause NPEs.

magicus (approved):
Agree. Let's go with this for now, and fix it better in the future.

Issue #157, opened 2022-07-10
Assign stable UUIDs to characters


magicus:
I have a plan, but I'm not going to be able to address it for a while. I'll sketch my idea here, and leave it open for anyone else to implement if they want. (Otherwise I'll do it when I get around to it.)

From the Wynncraft API we can get a json description of a player. Here is one describing me: https://api.wynncraft.com/v2/player/ca3ee631-a36c-48f4-9007-3c3c52c79840/stats
This contains, among other things, an array with all classes. The order in this array is similar to the order in the class selection screen, but not the same. API order is "total level", but class selection is "combat level". So there is not a simple 1-to-1 match that way. Also, API updates rather slowly. Actual changes in the characters take time (hours?) to propagate.

But. Thew good news is that the API has a stable naming of characters. In the object describing each class, there is a field "name", which will be unique per character. It is comprised of the class type (e.g. "archer") plus a monotonically increasing number (starting at 0, which is omitted). So the first archer you create is named "archer", the second "archer1", the next "archer2" etc. Even if you create and remove classes, numbers are not reused (confirmed).

What remains is to match characters found in the class selection screen with these characters from the API. We'll have to do a bit of fuzzy matching, but I think it'll be okay for all practical purposes. We can match on:
 * class
 * reskinned or not (the API name is "darkwizard" for a reskinned mage)
 * combat level (bearing in mind that the class selection level can be slightly higher, but never lower, since the API lags a bit)
 * combat XP percentage (with the same caveat; also if level has increased then XP can be lower as well)
 * number of finished quests

It would be really bad luck if a player has two characters that match on all these properties. If that happens I think we might be excused if we make a mismatch. (We can detect if this happens and either print a warning, or refuse to do matching until they have changed level/number of quests) on one of them.)

Finally, I think we should assign these characters an unique UUID. We can do this by:
```
String uniqueName = player.getUUID() + "-" + character.getName();
characterUUID = UUID.nameUUIDFromBytes(uniqueName.getBytes());
```
or if we want to be really fancy and do it the Right Way for v3 UUIDs:
```
public static final UUID WYNNCRAFT_CHARACTER_NAMESPACE = new UUID("ac64c234-48c3-4e3f-ad29-6694affcfbfa"); // randomly generated UUID that just needs to be kept the same

String uniqueName = WYNNCRAFT_CHARACTER_NAMESPACE + "-" + player.getUUID() + "-" + character.getName();
[...]

PR #330, opened 2022-08-17
ContainerQueryManager: Parse Quest Book, and fix autojoin character info

magicus:
I'm trying to create a better and more robust "fake inventory", but I'm calling it the Container Query Manager, since that better reflects the purpose of the class.

The main use cases are:
 * Quest Book analysis
 * Profile page (Compass page) analysis, either in the case of autojoin, or if we want to get additional info about the character (like notifying the player if there is a new Daily chest etc)

I could possibly also think of other use cases, like querying the Trade Market on your trades, etc.


magicus:
The current PoC just opens the Quest Book upon login, and looks through the first three pages.

I'm planning to create some kind of "scriptable" interface, so the user can like of setup rules like "if the title is such-and-such, and item in slot X is such-and-such, press slot X, and then expect title T2". I'm not really sure how to make this nice yet...

common/src/main/java/com/wynntils/core/managers/ContainerQueryManager.java
```diff
@@ -0,0 +1,168 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class ContainerQueryManager extends CoreManager {
+    private static int containerId = -2;
+    private static int transactionId = 0;
+
+    private static String lookForTitle;
+    private static final String QUESTS_TITLE = "§0[Pg. 1] §8mag_icus'§0 Quests";
+    private static final String QUESTS_TITLE_2 = "§0[Pg. 2] §8mag_icus'§0 Quests";
+    private static final String QUESTS_TITLE_3 = "§0[Pg. 3] §8mag_icus'§0 Quests";
```
Incompleteusern: Is this intentional? EDIT: Oh draft

magicus: Yeah, very much draft. :-) I just wanted to show a bit what I'm doing and kind of where I'm heading.

The concrete matching stuff needs to be moved to a QuestBookExtractionModel or something like that.

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
```diff
@@ -0,0 +1,134 @@
...
+
+                    return QUESTS_TITLE_3;
+                };
+                return QUESTS_TITLE_2;
+            };
+            return QUESTS_TITLE;
+        });
+    }
+
+    private void onScanQuestBook2() {
```
magicus: @kristofbolyai Does this seem like a good API to "script" an interaction with a container?

kristofbolyai: Seems fine, although it is harder to debug for sure.

magicus: When you say "harder to debug", you're not comparing with [QuestManager.readQuestBook()](https://github.com/Wynntils/Wynntils/blob/fdde2d1fece9f419615a72fd36b0b75415684eec/src/main/java/com/wynntils/modules/questbook/managers/QuestManager.java#L83) in Legacy, are you? :)

magicus: But I am open for suggestions on how to improve debuggability. My goal is to have a way to easily describe an interaction with a "hidden" inventory, where you can click on stuff, check the results and build up some kind of model of what you can extract from the information.

kristofbolyai: First off, what do you think of having a "QueryBuilder" class, then have a build method, to get the actual query you execute. That means good code separation, and you can step through code easier, as the containers used to build the query are not mixed up with the executing query's container.

magicus: That was kind of what I was aiming at, but it might not have been completely separated. Let me update my sketch.

magicus: @kristofbolyai Something like this? Is that more of what you where thinking of?

kristofbolyai: Seems better, but we will need to see the actual impl. Structural changes wouldn't require logic rewrite.

magicus:
This ended up being quite a massive PR. :) But I wanted to make sure the ContainerQuery mechanism worked well for the two most important scenarios: getting the Character info page (in case of autojoin), and for parsing the Quest Book. The latter was really worth a PR of it's own, but, well, now it is done.

For the moment the QuestBookFeature just dumps a list of all quests to the console. It's not particularly useful :-D but the idea is that this will form the basis upon which we can build a better Quest Book GUI. I've already done more than enough for quest book handling in this PR.

magicus:
I have tested this quite a bit while developing. There are certainly holes and bugs, but I feel quite confident that it is likely to not crash the game or do anything stupid if things go wrong. I've really, really tried to program the container query as defensive as I can.

magicus:
It is worth noting that the container query is extremely speedy. I can't even tell a background query is going in when logging in. Compared to Legacy, where the "update quest book" took like a second or two, during which everything lagged.

For a complete quest book analysis, we also need to look ad discoveries and miniquests. I've chosen to leave that be for the moment. Miniquests can be accessed using the scripted query API, but discoveries can have an arbitrarily large (well, not really perhaps, but it feels like it :)) number of pages, and will need to use the QueryStep interface more directly.

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @SubscribeEvent
+    public void onQuestBookReloadedEvent(QuestBookReloadedEvent e) {
+        System.out.println("All quests after reload:");
+        QuestBookModel.getQuests().forEach(System.out::println);
+    }
+
+    private void rescanQuestBook() {
+        System.out.println("Requesting rescan");
```
HighCrit: You forgot some prints

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -0,0 +1,153 @@
...
+                    .filter(query -> query.getName().equals(firstStepName))
+                    .findAny()
+                    .isEmpty()) {
+                queuedQueries.add(firstStep);
+            }
+            return;
+        }
+
+        Screen screen = McUtils.mc().screen;
+        if (screen != null && screen instanceof AbstractContainerScreen) {
```
HighCrit: Hold on... Does qodana have a valid point?

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -0,0 +1,153 @@
...
+            // Another inventory screen is already open, cannot do this
+            firstStep.onError("Another container screen is already open");
+            return;
+        }
+
+        if (McUtils.player().containerMenu.containerId != 0) {
```
HighCrit, line 51: Is containermenu never null?

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        int id = e.getContainerId();
+        if (id != containerId) {
+            raiseError("Another container opened");
+            return;
+        }
+
+        if (containerId == lastHandledContentId) {
+            // Wynncraft sometimes sends contents twice; just drop this silently
```
HighCrit, line 108: I'd imagine they do this when you stack unstackable items

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
```diff
@@ -0,0 +1,83 @@
...
+                .matchTitle(getQuestBookTitle(1))
+                .processContainer(c -> processQuestBookPage(c, 1))
+                .clickOnSlotMatching(NEXT_PAGE_SLOT, Items.GOLDEN_SHOVEL, getNextPageButtonName(2))
+                .matchTitle(getQuestBookTitle(2))
+                .processContainer(c -> processQuestBookPage(c, 2))
+                .clickOnSlotMatching(NEXT_PAGE_SLOT, Items.GOLDEN_SHOVEL, getNextPageButtonName(3))
+                .matchTitle(getQuestBookTitle(3))
+                .processContainer(c -> processQuestBookPage(c, 3))
+                .clickOnSlotMatching(NEXT_PAGE_SLOT, Items.GOLDEN_SHOVEL, getNextPageButtonName(4))
+                .matchTitle(getQuestBookTitle(4))
```
HighCrit: Is there no way to beat this hard coded values?

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -78,12 +82,25 @@ public static void onWorldStateChanged(WorldStateEvent e) {
...
-                    .withStyle(ChatFormatting.DARK_RED));
+            WynntilsMod.info("Scheduling character info query");
+            scanCharacterInfoPage();
         }
     }
 
+    private static void scanCharacterInfoPage() {
+        ScriptedContainerQuery query = ScriptedContainerQuery.builder("Character Info Query")
+                .useItemInHotbar(InventoryUtils.COMPASS_SLOT_NUM)
+                .matchTitle("^§.\\d+§. skill points remaining$")
```
HighCrit, line 97: Odd, wouldn't expect this to be the title of the compass 🤔

HighCrit (changes requested):

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
magicus: Apparently. I was so sure null was instanceof anything, but apparently it is not so.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
magicus: No, it's sometimes when you click on the "next page" button in the Quest Book. I have no idea why.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
magicus: What do you mean? I have explicitly tried to make the matching as exact as possible, to minimize the risk that we're randomly clicking around in some incorrect container. There was some war stories about folks who've lost mythics due to the Legacy fake inventory being buggy, so I want this to be "better safe than sorry". That means I try to match exactly what I expect, and if it does not match, I will fail.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: Yeah, it is odd, isn't it? But that is the actual title of the character info menu.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
magicus: No, it is either set to the open container, or the inventory (in which case containerId == 0).

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
magicus: I intended this to be a log output; changing. (If there ever is a strange crash or behavior I think we'd like to see if a container query was started nearby in time.)

HighCrit: Shouldn't we use theWynntilsMod logger

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
HighCrit: I mean the fact that we've specified each quest book page's name

magicus: That is very much by purpose. I could have skipped the title matching completely, or made it just a more fuzzy regexp match, but I really wanted to make sure we go get to exactly the page we want.

This is for instance how I discovered that some quest pages where sent twice -- I sent the "click on 'go to page 2'" packet, got an update, but it was still for page 1.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
HighCrit: Nvm, I was confused, it's the container's title not the item's

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
```diff
@@ -0,0 +1,83 @@
...
+                .processContainer(c -> processQuestBookPage(c, 4))
+                .onError(msg -> WynntilsMod.warn("Error querying Quest Book:" + msg))
+                .build();
+
+        query.executeQuery();
```
HighCrit:
```suggestion
        QueryBuilder query = ScriptedContainerQuery.builder("Quest Book Query")
                .useItemInHotbar(InventoryUtils.QUEST_BOOK_SLOT_NUM)
                .onError(msg -> WynntilsMod.warn("Error querying Quest Book:" + msg));
        
        for (int page = 1; page <= QUEST_BOOK_SIZE; page++) {
                query.matchTitle(getQuestBookTitle(page))
                .processContainer(c -> processQuestBookPage(c, page))
                .clickOnSlotMatching(NEXT_PAGE_SLOT, Items.GOLDEN_SHOVEL, getNextPageButtonName(page + 1));
        }
                
        query.build()
                .executeQuery();
```
I was more aiming at reducing the magic numbers by doing something like the above. If we wanted to future proof it more we can read the book in the questbook's sidebar to get the number of pages in total (e.g. Math.ceil(137 / 41))

HighCrit: (I coded the suggestion on my phone, please don't blindly commit)

magicus: Ah, I see, you wanted to reduce duplication of code by introducing a loop. I guess that is fine, even though it is a bit unusual way of using a builder. If it is not, I'll fix the API so it works.  I'll take your suggestion and make sure it works, thanks!

magicus: Fixed.

Fun fact: the new code is as many lines as the old. This one is hovering on the border where a loop is better than "unrolling" it. My hard rule is "never loop for 2 items", and my general rule is "only loop for 3 if it improves readability". Normally I'd agree that 4 items are enough to always be better with a loop, but this is really a border case, since the first container is different.

HighCrit (dismissed):

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class QuestBookFeature extends UserFeature {
+    @RegisterKeyBind
+    private final KeyBind questBookKeyBind =
+            new KeyBind("Rescan Quest Book", GLFW.GLFW_KEY_K, true, this::rescanQuestBook);
+
+    @Override
+    protected void onInit(
+            ImmutableList.Builder<Condition> conditions, ImmutableList.Builder<Class<? extends Model>> dependencies) {
+        super.onInit(conditions, dependencies);
```
Incompleteusern: Unnecessary as of now, and not done anywhere else as of now.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
Incompleteusern, line 97: matchTitle should support `Pattern`s too.

magicus:
@kristofbolyai You okay with this, too?

kristofbolyai:
I will review this in a bit, I haven't gotten around doing Artemis work today.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -41,6 +42,7 @@ public static void init() {
         registerPersistentDependency(OverlayManager.class);
         registerPersistentDependency(WebManager.class);
         registerPersistentDependency(WorldStateManager.class);
+        registerPersistentDependency(ContainerQueryManager.class);
```
kristofbolyai: Alphabetical order.

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
```diff
@@ -0,0 +1,50 @@
...
+        if (e.getNewState() == WorldStateManager.State.WORLD) {
+            WynntilsMod.info("Scheduling quest book query");
+            rescanQuestBook();
+        }
+    }
+
+    @SubscribeEvent
+    public void onQuestBookReloadedEvent(QuestBookReloadedEvent e) {
+        System.out.println("All quests after reload:");
+        QuestBookModel.getQuests().forEach(System.out::println);
```
kristofbolyai: Can we move this to logger debug (in case someone uses this in production, it does not spam log and does not produce big log files)?

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryStep.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.model.container;
+
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.inventory.MenuType;
+
+public interface ContainerQueryStep {
```
Incompleteusern, line 10: Nice comments!

common/src/main/java/com/wynntils/wynn/model/container/ScriptedContainerQuery.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        public ScriptedContainerQuery build() {
+            if (startAction != null || handleContent != null || verification != null) {
+                throw new IllegalStateException("Partial contents only for last step");
+            }
+            return query;
+        }
+
+        private void checkForCompletion() {
```
Incompleteusern, line 203: The fact that creating multiple steps in the builder is done implicitly should be commented

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class QuestBookModel extends Model {
+    private static final int NEXT_PAGE_SLOT = 8;
+
+    private static List<QuestInfo> quests = List.of();
+    private static List<QuestInfo> newQuests;
+
+    public static List<QuestInfo> getQuests() {
+        return quests;
+    }
```
kristofbolyai: Can we move this to the other getters?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
```diff
@@ -0,0 +1,196 @@
...
+    private static final Pattern LEVEL_MATCHER = Pattern.compile("^§..§r§7 Combat Lv. Min: §r§f(\\d+)$");
+    private static final Pattern REQ_MATCHER = Pattern.compile("^§..§r§7 (.*) Lv. Min: §r§f(\\d+)$");
+
+    private final String name;
+    private final QuestStatus status;
+    private final QuestLength length;
+    private final int level;
+    private final String nextTask;
+    /** Additional requirements as pairs of <"profession name", minLevel> */
+    private final List<Pair<String, Integer>> additionalRequirements;
```
kristofbolyai, line 30: Wouldn't this be nicer as it's own record/class?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
```diff
@@ -0,0 +1,196 @@
...
+        return "QuestInfo[" + "name=\""
+                + name + "\", " + "status="
+                + status + ", " + "length="
+                + length + ", " + "minLevel="
+                + level + ", " + "nextTask=\""
+                + nextTask + "\", " + "additionalRequirements="
+                + additionalRequirements + "]";
+    }
+
+    public static QuestInfo parseItem(ItemStack item) {
```
kristofbolyai, line 82: The defensive programming here :\)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private static String getDescription(LinkedList<String> lore) {
+        // The last two lines is an empty line and "RIGHT-CLICK TO TRACK"; skip those
+        List<String> descriptionLines = lore.subList(0, lore.size() - 2);
+        // Every line begins with a format code of length 2 ("§7"), skip that
+        // and join everything together, trying to avoid excess whitespace
+        String description = String.join(
+                        " ",
+                        descriptionLines.stream().map(line -> line.substring(2)).toList())
+                .replaceAll("  ", " ")
```
kristofbolyai:
```suggestion
                .replaceAll("\\s+", " ")
```
?

kristofbolyai:
This should close #157?

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -78,12 +82,25 @@ public static void onWorldStateChanged(WorldStateEvent e) {
 
         if (e.getNewState() == WorldStateManager.State.WORLD
                 && e.getOldState() != WorldStateManager.State.CHARACTER_SELECTION) {
-            McUtils.sendMessageToClient(new TextComponent(
-                            "Could not find your class type. Disable auto join (/toggle autojoin) and try again.")
-                    .withStyle(ChatFormatting.DARK_RED));
+            WynntilsMod.info("Scheduling character info query");
```
Incompleteusern, line 89: Mention autojoin here

common/src/main/java/com/wynntils/features/user/QuestBookFeature.java
magicus: I can remove it fully, as well. It was just some kind of demo code to show that I could parse the Quest Book fully.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: I'm thinking we can expand the API whenever the need arises. Are you ok with that? 

The "backend" API is very general, we have a "verificator" that gets the container title and menutype (basically what "kind" of container it is), and returns go/no-go if we got the right one. There is currently one implementation that checks for exact string match, and one for regexp (and neither checks menutype), but we can certainly expand this, to one that takes a Pattern for regexp, or whatever.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryStep.java
magicus: Thanks!

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
Incompleteusern: Im fine with this. We can expand if necessary.

common/src/main/java/com/wynntils/wynn/model/container/ScriptedContainerQuery.java
magicus: I added javadoc to QueryBuilder. I hope you find it acceptable.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookModel.java
magicus: I reordered it, but I'm not sure it got better. This one is a public getter, one of the two public methods on the manager. The other to methods named "get*" are more like "calculateThisString" than reading a field like a normal getter.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
magicus: Yeah, probably. If is valid for like 3 quests out of 100 though. I was mad enough at how much effort it took to parse and handle those requirements as it were... 

If anything, we should probably have some higher level representation of requirements ("Fishing lv 100", "Intelligence lv 55", "Combat lv 12") and use it everywhere we're talking about level requirements of some kind of aspect of your character. Until we get that, I'd prefer if this stays like this. Ok?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
magicus: I've tried my best not to lose any more mythics. :)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
kristofbolyai: Fine by me

Incompleteusern (approved):
🚢 !

magicus:
> This should close #157?

No, that is unrelated. We used to have a bug report (I believe?) on how autojoin messes up CharacterManager, but I can't find it now. What this solves is that we will always have a valid CharacterInfo, even if autojoin is used.

#157 on the other hands is about assigning a stable ID to characters, so we can save e.g. settings per character (like slot locks) and be sure we can always match them up again.

PR #336, opened 2022-08-17
Update Pull Request section of README

Incompleteusern:


kristofbolyai (approved):
This is is alright as is, but if you could include a few words on run configs, it would be nice.

If you do so, please include that deleting .idea/runConfigurations and reloading the gradle project fixes most issues that appear with the run configs. (also space in project path is really buggy)

HighCrit (approved):

Incompleteusern:
> This is is alright as is, but if you could include a few words on run configs, it would be nice.
> 
> If you do so, please include that deleting .idea/runConfigurations and reloading the gradle project fixes most issues that appear with the run configs. (also space in project path is really buggy)

I don't think I would be the best for writing about run configs.

PR #338, opened 2022-08-19
ObjectivesOverlayFeature: Fix render height when multiline

### LINKED ISSUE #298: Incorrect offsets for Daily objective rendering when multi line
image

kristofbolyai:
Fixes #298

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/overlays/ObjectivesOverlayFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/overlays/ObjectivesOverlayFeature.java
@@ -101,7 +101,9 @@ public class ObjectivesOverlayFeature extends UserFeature {
             final int barHeight = this.enableProgressBar ? 5 : 0;
             final int barWidth = 182;
             final float actualBarHeight = barHeight * (this.getWidth() / barWidth);
-            final float renderedHeight = 9 + actualBarHeight;
+            final float renderedHeight = FontRenderer.getInstance()
+                            .calculateRenderHeight(List.of(guildObjective.asObjectiveString()), this.getWidth())
+                    + actualBarHeight;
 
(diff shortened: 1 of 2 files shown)


magicus:
@kristofbolyai This also includes https://github.com/Wynntils/Artemis/pull/305? Seems you did not base the fix on main.

kristofbolyai:
Just fixed it.

common/src/main/java/com/wynntils/features/user/overlays/ObjectivesOverlayFeature.java
```diff
@@ -181,8 +183,13 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
             final int barHeight = this.enableProgressBar ? 5 : 0;
             final int barWidth = 182;
             final float actualBarHeight = barHeight * (this.getWidth() / barWidth);
-            final float renderedHeight = SPACE_BETWEEN + 9 + actualBarHeight;
-            final float fullHeight = renderedHeight * objectives.size() - SPACE_BETWEEN;
+            final float renderedHeightWithoutTextHeight = SPACE_BETWEEN + actualBarHeight;
+            final float fullHeight = (float) (renderedHeightWithoutTextHeight * objectives.size()
```
magicus, line 187: If we ever start doing something like this in some other place, I think we should move it to a utility class, but for now it can reside here.

magicus: It's kind of icky. I assume it's doing the right thing but I'm a bit too tired to really wrap my head around it, and convince myself that it is. This is no complaint on your code, I believe the problem is actually more complex than one would think at first glance, and complex problems require complex solutions.

magicus (approved):

PR #339, opened 2022-08-19
Do not go to CHARACTER_SELECTION state if we have autojoin

magicus:
I also cleaned up some ugly stuff with state detection.


common/src/main/java/com/wynntils/wynn/model/WorldStateManager.java
```diff
@@ -142,21 +144,22 @@ public static void onTabListFooter(PlayerInfoFooterChangedEvent e) {
...
-                setState(State.HUB, "", null);
+                setState(State.HUB, "");
             }
         }
     }
 
     @SubscribeEvent
     public static void update(PlayerDisplayNameChangeEvent e) {
         if (!onServer()) return;
+        if (!e.getId().equals(WORLD_NAME_UUID)) return;
```
kristofbolyai, line 155: UUIDs again? What is this used for?

magicus: Slightly quicker return in case it's not about world changes, and slightly higher accuracy. I used to pick this up dynamically, but it never changes, and iirc @HeyZeer0 confirmed that it is always the same.

kristofbolyai: I hope this won't bite us again. But let's go with it.

PR #341, opened 2022-08-19
Fix Event thread safety

magicus:
This PR fixes:
 * ClientPacketEvents are no longer sent twice, once on the IO thread and once on the main thread, but only on the main thread
 * Events are now annotated if they are sent on anything but the main thread. This is checked while running in development mode

The code will signal one warning, which will be fixed by https://github.com/Wynntils/Artemis/pull/339

### DIFF
--- a/common/src/main/java/com/wynntils/core/events/EventThread.java
+++ b/common/src/main/java/com/wynntils/core/events/EventThread.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.events;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface EventThread {
+    Type value() default Type.RENDER;
+
+    enum Type {
+        RENDER,
+        IO,
+        WORKER,
+        ANY
+    }
+}
--- a/common/src/main/java/com/wynntils/core/events/EventThread.java
+++ b/common/src/main/java/com/wynntils/core/events/EventThread.java
@@ -20,8 +20,8 @@ public @interface EventThread {
 
     enum Type {
         RENDER, // The main thread a.k.a the Render thread
-        IO,     // Any Netty Client IO thread
+        IO, // Any Netty Client IO thread
         WORKER, // A worker thread, from a Minecraft or Wynntils thread pool
-        ANY     // Any thread at all
+        ANY // Any thread at all
     }
 }
--- a/common/src/main/java/com/wynntils/core/events/EventThread.java
+++ b/common/src/main/java/com/wynntils/core/events/EventThread.java
@@ -9,15 +9,19 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * Inform what type of thread the event is allowed to be sent on. Events without an
+ * explicit annotation is considered to be allowed on Type == RENDER only.
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 public @interface EventThread {
     Type value() default Type.RENDER;
 
     enum Type {
-        RENDER,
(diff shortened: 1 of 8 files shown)


common/src/main/java/com/wynntils/core/events/EventThread.java
```diff
@@ -0,0 +1,23 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface EventThread {
+    Type value() default Type.RENDER;
+
+    enum Type {
+        RENDER,
+        IO,
+        WORKER,
+        ANY
```
Incompleteusern: Please comment what these are. I think Render and IO are minecraft internal, and Worker is artemis internal, but once again it's hard to know.

common/src/main/java/com/wynntils/mc/event/PacketEvent.java
```diff
@@ -28,13 +29,15 @@ public T getPacket() {
...
     @Cancelable
+    @EventThread(EventThread.Type.ANY)
     public static class PacketSentEvent<T extends Packet<?>> extends PacketEvent<T> {
         public PacketSentEvent(T packet) {
             super(packet);
         }
     }
 
     @Cancelable
+    @EventThread(EventThread.Type.ANY)
```
Incompleteusern, line 40: Shouldn't packet sending/receiving only occur on IO thread?

magicus: You'd think. :) I have not dug into the details, but basically this mixin is injected deep in networking code, which is apparently used in stuff like scanning servers for the join server list, and whatsnot. We only use it for debugging purposes, so we're fine with that.

common/src/main/java/com/wynntils/core/events/EventThread.java
magicus: Ok, fixed.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -34,10 +35,15 @@
...
 public abstract class ClientPacketListenerMixin {
+    private static boolean isRenderThread() {
+        return (McUtils.mc().isSameThread());
+    }
+
     @Inject(
             method = "handleCommands(Lnet/minecraft/network/protocol/game/ClientboundCommandsPacket;)V",
             at = @At("HEAD"))
     private void handleCommandsPre(ClientboundCommandsPacket packet, CallbackInfo ci) {
+        if (!isRenderThread()) return;
```
HighCrit, line 46: I would've expected this more on the worker thread than the render but 🤷

common/src/main/java/com/wynntils/core/events/EventThread.java
```diff
@@ -0,0 +1,27 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Inform what type of thread the event is allowed to be sent on. Events without an
+ * explicit annotation is considered to be allowed on Type == RENDER only.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface EventThread {
+    Type value() default Type.RENDER;
```
HighCrit, line 19: I might've even removed the default value, such that people get used to defining it.

HighCrit (approved):

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
magicus: What do you mean by "worker" thread? The "Render" thread is the main game thread. I'd had called it "Main" but Minecraft calls it "Render Thread" so I stick with that.

common/src/main/java/com/wynntils/core/events/EventThread.java
magicus: When I started the annotation process, it turned out that almost all events are thrown (correctly) on the main/render thread. So this annotation is really about the few events that are not. If anything, I could remove RENDER, and force you to declare using the Render thread by not having the annotation at all, but I thought I would keep it to make it possible to tell this explicitly.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
HighCrit: Ah, now I get it

PR #343, opened 2022-08-19
Let EventFactory only send out events if we are on Wynncraft

magicus:
This increases mod "safety" and decreases complexity. Basically, none of our events are sent out unless we're on a wynncraft server, so each listener do not need to implement checks (which anyway was missing in like half or more of the cases). A few exceptions are present, using `postAlways`, and that is basically needed to detect logon to Wynncraft, and to set up the main menu button.


magicus:
I have painstakingly followed up uses of each and every event and verified that this change is okay. That was not fun, but it makes it possible for me to say with certainty that I do not think this will break anything.

There are some instances of `WynnUtils.onWorld()` scattered around the code. I've kept most of it. I don't really think this makes a lot of sense most of the time; it was probably just about as arbitrarily selected for a check as `onServer`, but in other cases it actually *do* have meaning, like prohibiting minimap from rendering until you are on an actual world. So I've left most of these calls intact.

common/src/main/java/com/wynntils/mc/EventFactory.java
```diff
@@ -107,15 +107,19 @@
 /** Creates events from mixins and platform dependent hooks */
 public final class EventFactory {
     private static <T extends Event> T post(T event) {
```
kristofbolyai, line 109: Maybe a postOnCondition with a predicate would be better have, and we could have a static onServer predicate cached. 

Or it can stay like this until we need to change it.

magicus: Haha. :-) Actually, my first attempt was to add a `boolean condition` to `post`. But then it turned out to have the same condition everywhere, except a few cases which had `true`. This made them not stand out, since your eyes started kind of filtering out the last argument. 

I was also initially thinking "maybe some events should be sent only `onServer`, and others `onWorld`", but that was not a good approach either. What we really want to do is filter `onServer`; the few moments we spend in the hub or character selection screen is irrelevant -- the point is to not send events on single-player and other servers. A handful of features must explicitly be disabled on hub or char sel screen (like minimap), but they can do so themselves.

It was also very confusing, not knowing if the events where delivered or not. I think this is better. As soon as we have confirmed that we are on Wynncraft, we start sending events. Additional filtering can be done by listeners.

kristofbolyai (approved):

PR #345, opened 2022-08-20
Screens: Add settings screen

kristofbolyai:


kristofbolyai:
Status update 2022.08.22: 
Feature list is now semi-working with basic search functionality. 

image
image

Incompleteusern:
I feel that rounded edges of the button kinda clashes as the scroll bar is already round, on that note I think making it look similar to overlay config (red if enabled, green if disabled, square boxes) might work. In general, it might be good to make feature config and overlay config look more similar (though I definitely prefer the feature config color choices).

In the future, maybe we could have customizable color palettes for the settings screen, but that's a bit meta and might have problems, along with being infeasible right now.

magicus:
I think we should not sweat to much over graphic design at this point. Hopefully we can get some graphic designers to step up; my impression is that there is usually a ratio of 3:1 of people wanting to contribute who can do graphics relative to those who can do coding.

kristofbolyai:
Also I would like to mention the pain that working with MC screens/guis cause. It is, for sure, something we need to tackle when we get the new textures. If there is no good framework we can shadow, we might as well create one...

magicus:
There are GUI frameworks, of course. For configuration screens, for instance, there are several popular, ClothConfig for instance. I'm not sure how well they will fit the "look and feel" of Wynncraft, though. Or if we think that matters.

kristofbolyai:
Status update 2022.08.24: 
Today was focused on small bug fixes in the feature list code. Now scrolling is less buggy, and mostly works (and looks) fine. I also tested different resolutions, like 1080p, and the screen is looking okay (there are some alignment issues with multi-line feature names, we can fix that by taking out "feature" of the end of every feature name).

I also went ahead and improved on our search widget, as I believe it will be something we will re-use in many different places. Now it has a working cursor!
image

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
```diff
@@ -113,8 +104,6 @@ protected void renderBackground(PoseStack poseStack) {
                 Texture.FEATURE_LIST_BACKGROUND.height());
     }
 
-    // FIXME: This is incorrect, entries have different heights
-    //        Consider not using this in renderList
```
Incompleteusern: How is this fixed now?

magicus:
I agree that we can take out "Feature". It looked good in the beginning, but in the context of all features it just looks weird.

magicus:
Regarding the search (or rather "filter") box, one feature I tend to like for such is the ability to match not only substrings, but just consecutive characters. Like "heapob" would match "HEAlth POtion Blocker". (It could be matched by "loo" as well, "heaLth pOtiOn blocker", so if we do that, we should probably reorder results so those where the match *is* consecutive("LOOtrun") comes on top.)

Not saying you have to do this, but if you get inspired and think it sounds like a cool thing to implement, I won't protest. :)

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
kristofbolyai: It is not, but it does not make that big of a difference. I will fix it either way though (although not here, but in the render code where I actually use this).

kristofbolyai:
Status update 2022.08.25:
More improvements to feature list code and search bar.

Localization now does not have " Feature" for every feature name.

image

Search bar now supports "fuzzy"(ish) matching (basically what @magicus described https://github.com/Wynntils/Artemis/pull/345#issuecomment-1226926441).

image
image

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
kristofbolyai: Fixed now.

kristofbolyai:
Status update 2022.08.25 part 2: 

Scroll button!

image


PS: We also hit 1k lines of code... I haven't even added anything about selecting a feature, and displaying it's specific config options :sweat_smile:

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
```diff
@@ -0,0 +1,262 @@
...
+    public void reAddEntriesWithSearchFilter(String searchText) {
+        this.clearEntries();
+        this.setScrollAmount(0);
+
+        float renderHeight = 0;
+
+        String lastCategory = "";
+
+        for (Feature feature : FeatureRegistry.getFeatures().stream()
+                .filter(feature -> !(feature instanceof DebugFeature))
```
HighCrit, line 335:
```suggestion
                .filter(feature -> !(feature instanceof DebugFeature) || WynntilsMod.isDevelopmentEnvironment())
```
I think it'd be nice for our devs to easily access debugfeatures in menu.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -390,6 +390,98 @@ public static void drawArc(
...
+                    .endVertex();
+            bufferBuilder
+                    .vertex(matrix, midX + sinAngle * innerRadius, midY - cosAngle * innerRadius, z)
+                    .color(color.r, color.g, color.b, color.a)
+                    .endVertex();
+        }
+
+        bufferBuilder.end();
+        BufferUploader.end(bufferBuilder);
+        RenderSystem.disableBlend();
```
HighCrit: Why don't you just call drawArc and calculate the values of the arguments in drawRoundedCorner

kristofbolyai: Done. Had to modify drawArc a bit though... I first thought PoseStack translations would work, but they did not.

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
kristofbolyai: Maybe, but I am not sure if we should do this. Devs can and should be able to use the config command and/or edit the config files, then reload the configs with already existing command.

magicus: I agree. Or we can hand-edit the files to enable/disable features and recompile.

kristofbolyai: > I agree. Or we can hand-edit the files to enable/disable features and recompile.

Can you clarify which comment you agree with 😅?

Incompleteusern: I agree that it would still be more convenient to use the menu

kristofbolyai:
Status update 2022.08.27: 

Vertically aligned categories.

image

kristofbolyai:
Status update 2022.08.27 part 2:

The settings screen buttons were updated. Now there is the expected close without saving, and save buttons. 

Features can be selected now. Selection can be moved up/down with arrows. Selecting a feature will play the UI click sound. 

Selected feature name is now displayed, and a "scrollbar" is drawn. This is in preparation for config list rendering.

image

Incompleteusern:
It might be nice to color the boxes instead later on, but currently this is not supported and has no need to go out of the way to support.

kristofbolyai:
@Incompleteusern https://github.com/Wynntils/Artemis/pull/345/commits/eb9522b158b8cae24d6b0ac4191402ab6a6a3ced solves code duplication, but it has a little cost...
I can't return early + I have to use atomic references.

Incompleteusern:
> @Incompleteusern [eb9522b](https://github.com/Wynntils/Artemis/commit/eb9522b158b8cae24d6b0ac4191402ab6a6a3ced) solves code duplication, but it has a little cost... I can't return early + I have to use atomic references.

Feel free to revert that if the cost is too much in your opinion, both are fine for me.

kristofbolyai:
> > @Incompleteusern [eb9522b](https://github.com/Wynntils/Artemis/commit/eb9522b158b8cae24d6b0ac4191402ab6a6a3ced) solves code duplication, but it has a little cost... I can't return early + I have to use atomic references.
> 
> Feel free to revert that if the cost is too much in your opinion, both are fine for me.

I think it is fine until I only update hovered on click. If we have to do it in render, to render a tooltip or smth, I will definitely revert.

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
magicus: I agree with the agreeableness of all agreements. Agreed? :-)

Seriously, though, I thought that debug features do not need GUI configs. They are a special case tool. If we start adding them in devel builds, it only means the config screen will be not-as-in-production for devel builds, which is bad, for very little gain.

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
```diff
@@ -0,0 +1,189 @@
...
+        float renderY = settingsScreen.getBarHeight() + padding;
+
+        final float xOffset = this.width / 35f;
+
+        for (int i = scrollIndexOffset; i < configWidgets.size(); i += 2) {
+            ConfigOptionElement configWidgetLeft = configWidgets.get(i);
+            ConfigOptionElement configWidgetRight = configWidgets.size() > i + 1 ? configWidgets.get(i + 1) : null;
+            float renderHeight = this.height / 4f;
+
+            if (renderY + renderHeight > this.y + this.height) break;
```
Incompleteusern, line 161: This can be simplified to
```java
            if (renderY > this.y + 3/4f * this.height) break;
```

kristofbolyai: I am not sure this refactor is much simpler either.

kristofbolyai:
Status update 2022.08.28:

Features now display their config options, with scroll bar!
Feature userEnabled state can be toggled with a switch.
Boolean config options can be toggled.
Setting screen ESC now reverts settings to their previous state.
Fixed FontRenderer alignment bug.

image

magicus:
So what about my suggestion to have the enable switch next to the actual feature, in the list? I'm thinking that would make it very easy to see if a feature is actually enabled, and to turn it on and off.

kristofbolyai:
> So what about my suggestion to have the enable switch next to the actual feature, in the list? I'm thinking that would make it very easy to see if a feature is actually enabled, and to turn it on and off.

I think I missed it, if you mentioned this before. I can do that. Fine by me either way.

magicus:
I tried to describe it in the discord channel, but without a mockup or sketch it's hard to get across ideas of GUI design. 

I think it would make for a very smooth UX, so if you can do it, please do.

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
Incompleteusern: It gets rid of the `renderHeight` variable

common/src/main/java/com/wynntils/screens/settings/elements/BooleanConfigOptionElement.java
```diff
@@ -0,0 +1,56 @@
...
+import com.wynntils.screens.settings.widgets.FeatureSettingWidget;
+
+public class BooleanConfigOptionElement extends ConfigOptionElement {
+    public BooleanConfigOptionElement(ConfigHolder configHolder, FeatureSettingWidget featureSettingWidget) {
+        super(configHolder, featureSettingWidget);
+    }
+
+    @Override
+    protected void renderConfigAppropriateButton(PoseStack poseStack, float width, float height) {
+        if (configHolder.getType().equals(Boolean.class)) {
```
Incompleteusern: Isn't this unnecessary?

kristofbolyai: It is now. There was a need in another iterations.

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
```diff
@@ -0,0 +1,308 @@
...
+    private static final CustomColor BORDER_COLOR = new CustomColor(86, 75, 61, 255);
+    private static final CustomColor FOREGROUND_COLOR = new CustomColor(177, 152, 120, 255);
+    private static final CustomColor SCROLLBAR_COLOR = new CustomColor(137, 117, 92, 255);
+
+    private final WynntilsSettingsScreen settingsScreen;
+
+    private Feature cachedFeature = null;
+    private List<ConfigOptionElement> configWidgets = new ArrayList<>();
+    private ConfigOptionElement hoveredConfigWidget = null;
+    private boolean enabledStateChangeable = false;
```
Incompleteusern: This variable is unused as of now

kristofbolyai: It is not? I check it when rendering the on off switch?

kristofbolyai: I will get rid of it anyways if I move the switches.

Incompleteusern: You opted to use
```java
         final Texture switchTexture = selectedFeature.isEnabled() ? Texture.SWITCH_ON : Texture.SWITCH_OFF;
```
there instead

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
kristofbolyai: Well I need the renderHeight variable now anyways

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
```diff
@@ -0,0 +1,308 @@
...
+                        this.width,
+                        this.height / 50f,
+                        this.width,
+                        CommonColors.WHITE,
+                        FontRenderer.TextAlignment.CENTER_ALIGNED,
+                        FontRenderer.TextShadow.OUTLINE);
+    }
+
+    private void renderEnabledSwitch(PoseStack poseStack, Feature selectedFeature) {
+        if (!enabledStateChangeable) return;
```
kristofbolyai: @Incompleteusern

Incompleteusern: ohhhhh

kristofbolyai:
Status update 2022.08.29:

Text box inputs! I generalized the search widget, and now we have a text input widget as well. Now we properly handle and care about what text box is selected/focused, compared to when we only had the search widget.

Text config types: They have a parse indicator. I am planning for these to work with most config holders, and we can add special support for numbers, etc. later.

I also changed the size of the switches. They look better like this on smaller displays.

This screen now reached a state where it is completely usable to edit any feature configs. I only have to add overlay config editing, and it will be "ready" to ship. However, I do plan to add a special button for enums, so it is easier to switch between them.

image
image

image

kristofbolyai:
Also, almost hit 2000 lines of code.

common/src/main/java/com/wynntils/screens/settings/elements/BooleanConfigOptionElement.java
```diff
@@ -16,41 +16,39 @@ public BooleanConfigOptionElement(ConfigHolder configHolder, FeatureSettingWidge
...
     @Override
     public void mouseClicked(double mouseX, double mouseY, int button) {
         Boolean value = (Boolean) configHolder.getValue();
 
-        System.out.println("value = " + value);
-
         if (value == null) {
             configHolder.setValue(false);
         } else {
             configHolder.setValue(!value);
```
Incompleteusern, line 50: This sets the value, so even if you quit without saving, the config is still applied when `Configmanager#save` is called

kristofbolyai: Yeah but we reload saved configs from the file. So it is either saved or discarded.

kristofbolyai:
image

kristofbolyai:
Status update 2022.08.30:

This is finally ready for reviews!

Changes since last time:
Moved the feature switch
Added custom config switches for enums and custom colors
Fixed stuff

Tested on 1080p and 3440*1440. This is likely not perfect on every resolution, but we can deal with strange cases later.

Screenshots of the current state:
image
image
image
image

kristofbolyai:
Let's start the review process. This is gonna take a few iterations :\).

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,98 @@
...
+                                : (textBoxInput.substring(0, cursorPosition)
+                                        + cursorChar
+                                        + textBoxInput.substring(cursorPosition)),
+                        this.x + 5,
+                        this.x + this.width - 5,
+                        this.y + 6.5f,
+                        this.width,
+                        CommonColors.WHITE,
+                        FontRenderer.TextAlignment.LEFT_ALIGNED,
+                        FontRenderer.TextShadow.NORMAL);
```
Incompleteusern, line 59: Cursor char should render when focused, even if empty

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,98 @@
...
+                this.y,
+                0,
+                this.width,
+                this.height,
+                0,
+                0,
+                Texture.SEARCH_BAR.width(),
+                Texture.SEARCH_BAR.height(),
+                Texture.SEARCH_BAR.width(),
+                Texture.SEARCH_BAR.height());
```
Incompleteusern, line 73: Don't we have a method already to render an entire bar? If not, we should have one.

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
```diff
@@ -0,0 +1,215 @@
...
+
+            while (font.width(builder.toString()) < maxTextWidth - cursorWidth
+                    && stringPosition < this.textBoxInput.length()) {
+                builder.append(textBoxInput.charAt(stringPosition));
+
+                stringPosition++;
+            }
+
+            renderedText = builder.toString();
+        }
```
Incompleteusern, line 104: What does this code do?

magicus:
I plan to start by checking out the PR and testing it locally. I understand that at some point, behavioral fixes might have to wait for follow up PRs, but I want to see that it works roughly okay for me.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: The method we already have is not scaling up the texture, but rather, rendering it "multiple" times as a pattern.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: Why? When we are not focused, we render "Search...", when we are focused, we render the current search with the cursor char.

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
kristofbolyai: This is basically to only render text that fits into the text widget (truncate if string is too long). I can add code comments to document this.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: Fixed.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
Incompleteusern: Last time I checked, when we are focused and empty, it still renders "Search..."

kristofbolyai: image
image

kristofbolyai: image

kristofbolyai: I am not sure when you checked. It is working now, I might have fixed it yesterday or today though.

Incompleteusern:
image
The text boxes are small for my screen size, but that is best for a separate pr.

On another note, the addition and removal of the cursor is annoying, as it shifts the text back and forth, but is not that big of a concern.

A bug occurs when the cursor is placed at the end of a text input book that has enough characters to fill it entirely, since then the cursor position can desync when pressing the left key once when its rendered.

This is caused by
```java
            int stringPosition = Math.min(textBoxInput.length() - 1, cursorPosition);
```
where cursorPosition equals `textBoxInput.length()`; subtracting one causes no change in rendering but changes the cursor position;

A fix is likely to replace
```java
        cursorPosition = Math.min(textBoxInput.length(), cursorPosition + 1);
```
with
```java
        cursorPosition = Math.min(textBoxInput.length() - 1, cursorPosition + 1);
```
which on a note also makes the earlier `Math.min` for `stringPosition` moot, if that is correct.

kristofbolyai:
> image
> 
> The text boxes are small for my screen size, but that is best for a separate pr.
> On another note, the addition and removal of the cursor is annoying, as it shifts the text back and forth, but is not that big of a concern.
> 
> A bug occurs when the cursor is placed at the end of a text input book that has enough characters to fill it entirely, since then the cursor position can desync when pressing the left key once when its rendered.
> 
> This is caused by
> 
> ```java
>             int stringPosition = Math.min(textBoxInput.length() - 1, cursorPosition);
> ```
> 
> where cursorPosition equals `textBoxInput.length()`; subtracting one causes no change in rendering but changes the cursor position;
> 
> A fix is likely to replace
> 
> ```java
>         cursorPosition = Math.min(textBoxInput.length(), cursorPosition + 1);
> ```
> 
> with
> 
> ```java
>         cursorPosition = Math.min(textBoxInput.length() - 1, cursorPosition + 1);
> ```
> 
> which on a note also makes the earlier `Math.min` for `stringPosition` moot, if that is correct.

I will increase the size, but I've tried fixing the blinking and it did not work, if you want, you can work on that as a follow up.

kristofbolyai:
Added support for overlay config options, I left that out by accident. 

image

magicus:
So, I need to take the bull by the horns. Making a deep-dive review now.

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -210,6 +210,10 @@ public boolean canEnable() {
         return true;
     }
 
+    public boolean canUserEnable() {
+        return getConfigOptionFromString("userEnabled").isPresent();
```
magicus: Seem a bit backwards to determine this from config strings...

kristofbolyai: This could be an instaceof check too, yes.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -383,6 +406,68 @@ public static void drawArc(
...
+        // Edges
+        drawLine(poseStack, borderColor, x + offset, y, x2 - offset, y, z, lineWidth);
+        drawLine(poseStack, borderColor, x2, y + offset, x2, y2 - offset, z, lineWidth);
+        drawLine(poseStack, borderColor, x + offset, y2, x2 - offset, y2, z, lineWidth);
+        drawLine(poseStack, borderColor, x, y + offset, x, y2 - offset, z, lineWidth);
+
+        // Corners
+        poseStack.pushPose();
+        poseStack.translate(-1, -1, 0);
+        drawRoundedCorner(poseStack, borderColor, x, y, z, innerRadius, outerRadius, Mth.HALF_PI * 3);
```
magicus, line 451: `Mth`?

kristofbolyai: Mth is Minecraft's MathUtils class.

Incompleteusern: Name of minecraft math class.

common/src/main/java/com/wynntils/screens/overlays/OverlaySelectionScreen.java
```diff
@@ -24,9 +24,11 @@ public class OverlaySelectionScreen extends Screen {
     private static final int BUTTON_HEIGHT = 20;
 
     private OverlayList overlayList;
+    private final Screen lastScreen;
 
     public OverlaySelectionScreen() {
         super(new TranslatableComponent("screens.wynntils.overlaySelection.name"));
```
magicus, line 32: I think in general there is a pattern in Minecraft where you send in the "previous" screen as the first argument to the Screen constructor. There used to be in 1.12 at least. I think we can try and follow it.

kristofbolyai: I will see if there is such thing here. I don't think so though.

common/src/main/java/com/wynntils/screens/settings/WynntilsSettingsScreen.java
```diff
@@ -0,0 +1,239 @@
...
+    public void resize(Minecraft minecraft, int width, int height) {
+        this.init();
+        super.resize(minecraft, width, height);
+    }
+
+    @Override
+    public boolean mouseScrolled(double mouseX, double mouseY, double delta) {
+        if (featureList.isMouseOver(mouseX, mouseY)) {
+            return featureList.mouseScrolled(mouseX, mouseY, delta);
+        } else if (featureSettingWidget.isMouseOver(mouseX, mouseY)) {
```
magicus, line 155: Is this not handled by super, since `featureSettingWidget` is set by `addRenderableWidget`?

kristofbolyai: This is needed so the "order" of processing the click is how I want it, compared to super.

magicus: I might be wrong here, but if not, that would also make a case for also adding the `featureList` to the renderable widget list. I think that is in general more akin to how the guis of minecraft work. I understand this has been in the works for long, and I will not require any additional works just "because it looks better", but if you agree with me you might consider this :)

kristofbolyai: I don't think feature list can be added as a widget though.

kristofbolyai: (because of the superclass)

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
```diff
@@ -0,0 +1,362 @@
...
+    @Override
+    public boolean isMouseOver(double mouseX, double mouseY) {
+        return mouseX <= getRenderWidth()
+                && mouseY >= this.settingsScreen.getBarHeight()
+                && mouseY <= this.settingsScreen.getBarHeight() + this.height;
+    }
+
+    @Override
+    protected FeatureListEntryBase getEntryAtPosition(double mouseX, double mouseY) {
+        AtomicReference<FeatureEntry> returnValue = new AtomicReference<>(null);
```
magicus, line 110: Why AtomicReference? Is there a race here?

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,94 @@
...
+    private final Component DEFAULT_TEXT = new TranslatableComponent("screens.wynntils.searchWidget.defaultSearchText");
+
+    public SearchWidget(
+            int x,
+            int y,
+            int width,
+            int height,
+            Consumer<String> onUpdateConsumer,
+            WynntilsSettingsScreen settingsScreen) {
+        super(x, y, width, height, new TextComponent("Search..."), onUpdateConsumer, settingsScreen);
```
magicus: Why not use `DEFAULT_TEXT` here?

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        this.renderBg(poseStack, McUtils.mc(), mouseX, mouseY);
```
magicus, line 40: Maybe rename it `renderBackground` to match other places?

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,94 @@
...
+                        Objects.equals(textBoxInput, "") && !isFocused()
+                                ? DEFAULT_TEXT.getString()
+                                : (textBoxInput.substring(0, cursorPosition)
+                                        + cursorChar
+                                        + textBoxInput.substring(cursorPosition)),
+                        this.x + 5,
+                        this.x + this.width - 5,
+                        this.y + 6.5f,
+                        this.width,
+                        CommonColors.WHITE,
```
magicus: If we are showing the default "Search..." string, make it gray or light gray to show more clearly that it is not a real search term.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
```diff
@@ -0,0 +1,94 @@
...
+
+        return false;
+    }
+
+    @Override
+    public void updateNarration(NarrationElementOutput narrationElementOutput) {}
+
+    @Override
+    protected void removeFocus() {
+        this.setTextBoxInput("");
```
magicus, line 93: What does this do? Remove the search term if you e.g. tab away from it? Is that really correct?

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
```diff
@@ -0,0 +1,216 @@
...
+        this.settingsScreen = settingsScreen;
+    }
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        poseStack.pushPose();
+
+        poseStack.translate(this.x, this.y, 0);
+
+        this.renderBg(poseStack, McUtils.mc(), mouseX, mouseY);
```
magicus, line 67: `renderBackground`?

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: This is what the superclass method is called.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: Remove search if you esc away:

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
kristofbolyai: Same as before, superclass method name

common/src/main/java/com/wynntils/screens/settings/lists/FeatureList.java
kristofbolyai: Needed for the code below.

common/src/main/java/com/wynntils/screens/settings/widgets/FeatureSettingWidget.java
```diff
@@ -0,0 +1,306 @@
...
+                renderWidth = fullWidth / 2;
+                float renderX = xOffset + fullWidth / 2;
+
+                if (isMouseOverConfigWidget(actualRenderX, actualRenderY, renderWidth, renderHeight, mouseX, mouseY)) {
+                    this.hoveredConfigElement = configWidgetRight;
+                }
+
+                if (configWidgetRight.getConfigHolder().getParent() instanceof Overlay
+                        && configWidgetRight.getConfigHolder().getParent() != lastOverlayConfigTitle) {
+                    renderY += renderHeight + padding;
```
magicus, line 184: I'm not sure I understand the if condition here. At what circumstance do we want/need to render the right widget to the left?

kristofbolyai: When there is an overlay, we need to render the title. If the left config field is the first for the overlay, it is easy case. If it is the right, we shift the right config title to be the left one, also rendering the title next iteration.

common/src/main/java/com/wynntils/screens/settings/elements/CustomColorConfigOptionElement.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    protected void renderSuccessState(PoseStack poseStack) {
+        poseStack.pushPose();
+
+        poseStack.translate(-getTextInputHeight() * 1.2f, 0, 0);
+
+        if (!this.lastParseSuccessful) {
+            RenderUtils.drawLine(
```
magicus, line 31: Maybe add comment "make a red cross if parsing failed" or something like that. Took me a while to figure out.

common/src/main/java/com/wynntils/core/features/Feature.java
magicus: I'd prefer that.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
magicus: Ah. Weird name. It used to be MathUtils, iirc. Much better imho.

common/src/main/java/com/wynntils/screens/settings/WynntilsSettingsScreen.java
magicus: ok

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
magicus: And you can't use that? What's the difference by the super's backgorund render method, and your?

kristofbolyai: I mean superclass's method is called renderBg, and I would rather override that than make a new method.

magicus: Ah, ok.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
Incompleteusern: Could this be undesirable?

kristofbolyai: You can click away and search will stay. Esc for quick search reset

common/src/main/java/com/wynntils/screens/settings/elements/ConfigOptionElement.java
```diff
@@ -9,11 +9,13 @@
 import com.wynntils.mc.objects.CommonColors;
 import com.wynntils.mc.objects.CustomColor;
 import com.wynntils.mc.render.FontRenderer;
+import com.wynntils.mc.render.HorizontalAlignment;
 import com.wynntils.mc.render.RenderUtils;
 import com.wynntils.mc.render.VerticalAlignment;
 import com.wynntils.screens.settings.WynntilsSettingsScreen;
 import com.wynntils.screens.settings.widgets.FeatureSettingWidget;
 import net.minecraft.client.gui.narration.NarrationElementOutput;
+import oshi.hardware.CentralProcessor;
```
magicus: :-D I guess spotless will take care of that.

common/src/main/java/com/wynntils/screens/widgets/SearchWidget.java
kristofbolyai: This is actually the widget name, so it should not be "search.."

common/src/main/java/com/wynntils/screens/overlays/OverlaySelectionScreen.java
kristofbolyai: No such thing in 1.18.2

kristofbolyai:
@Incompleteusern text alignment still broken 
image

kristofbolyai:
Fixed.

magicus (dismissed):
Looks good enough for now.

magicus:
We will have to continue tweaking this, though. This is how it looks on my screen with lower resolution. :-D
image

magicus:
But it's good enough for now. Let's ship it and move on.

Incompleteusern (dismissed):

PR #350, opened 2022-08-28
Command changes

### LINKED ISSUE #347: /feature list add ordering + categories
Currently is is unordered. Also enabled color is yellow, and disabled is red. This is weird.

### LINKED ISSUE #348: /wynntils should have feature and config commands as sub-commands
I agree that these commands can be on their own, but it is weird not to have them as sub-commands as well.

kristofbolyai:
Fixes #347 and fixes #348.

https://github.com/Wynntils/Artemis/commit/291351a607d4570d97ccbac32be9c8752f3cc2fc ...
Every time I see this superclass, I get a headache... Anyways, this should be a much more fitting name, as always on is just not right (see LootrunFeature). I think we can keep the superclass, but use it in really specific edge cases, like the ones we use it for already. Hence, I updated the comments too, just to make sure people think twice (or more times hopefully) before using it.

All other changes should be pretty straightforward.

Here is a screenshot of the new `/feature list` (https://github.com/Wynntils/Artemis/commit/5d4e6c5072f282e72a4e0be4c2061762ba3731de): 
image


common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -23,7 +23,7 @@
 import net.minecraft.client.resources.language.I18n;
 import net.minecraft.world.phys.Vec2;
 
-public abstract class Overlay implements Translatable, Configurable {
+public abstract class Overlay implements Translatable, Configurable, Comparable<Overlay> {
```
magicus, line 26: I agree with Qodana. We should implement `equals`.

common/src/main/java/com/wynntils/commands/FeatureCommand.java
```diff
@@ -51,17 +55,54 @@ private LiteralCommandNode<CommandSourceStack> buildListNode() {
...
+                if (feature.isEnabled()) {
+                    translatedName += " {ENABLED DEBUG}";
+                } else {
+                    translatedName += " {DISABLED DEBUG}";
+                }
+            }
+
+            System.out.println("feature.getCategory() = " + feature.getCategory());
+
+            if (!Objects.equals(lastCategory, feature.getCategory())) {
```
magicus, line 91: I think this would break if we *only* had uncategorized features. :) This is more an academic question, of course, but the warning bell triggered in the back of my head. Maybe just add a comment `// this assumes we have at least one categorized feature` to show the intent.

magicus: Or, yeah, maybe not breaks. We would not get a category header at all. Which would probably be correct. I dunno. Just ignore my ranting, I guess.

kristofbolyai: Also, debug line. Oops. 

And no, it would not. Default is null, uncategorized is "". So, it would trigger.

magicus (approved):

common/src/main/java/com/wynntils/commands/FeatureCommand.java
magicus: Ah, I thought uncategorized were null.  Good.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
kristofbolyai: I don't know how we can improve on the default one? Do we want equals to not care about instances, but rather overlay types? I am not sure if that is better.

magicus: Default equals just implements identity. We should compare the very same fields as we look at in the `compareTo` method. We should also implement a hash function that returns a hash based on those. That is "best practice" java, to not get caught in pitfalls. You don't have to do it now, but we should make sure we never implements `equals` or `hashCode` without the other, and never `compareTo` without the other two.

PR #352, opened 2022-08-28
Make feature category an enum, and fix i18n for categories

magicus:
This should not go in before https://github.com/Wynntils/Artemis/pull/350.


kristofbolyai:
Nice. I kinda needed this for setting screen anyways.

kristofbolyai:
How can one get the category from the feature though?

kristofbolyai:
> How can one get the category from the feature though?

Ah, I see, this is not fully implemented in Feature class yet.

magicus:
@kristofbolyai I think this is okay now, I'll just verify that it actually works :)

magicus:
Yeah, works fine

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1,4 +1,10 @@
 {
+  "core.wynntils.category.inventory": "Inventory",
```
kristofbolyai, line 2: woudln't feature.wynntils.category be better? As these are feature categories

kristofbolyai: Fine either way by me, though.

HighCrit (approved):
Only suggestion is bolyai's note but it's fine either way

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: That was my first thought as well, but we have `feature.wynntils.<NAME>` as a pattern, where this is a reserved namespace for feature `<NAME>`. While we might never have a feature named CategoryFeature, I still felt it was mixing two different things.

PR #354, opened 2022-08-28
Use asserts to verify that our assumptions are held

magicus:


kristofbolyai:
Another good PR for settings screen 😀

magicus:
@kristofbolyai I'm doing what I can to assist you. The settings screen is necessary, and a lot of work. And it inspires me to raise the quality of other, settings-related stuff in Artemis.

common/src/main/java/com/wynntils/core/keybinds/KeyBindManager.java
```diff
@@ -42,7 +42,7 @@ public static void onKeyPress(InventoryKeyPressEvent e) {
 
     public static void registerKeybind(KeyBind toAdd) {
         if (hasName(toAdd.getName())) {
-            throw new IllegalStateException("Can not add " + toAdd + " since the name already exists");
+            throw new IllegalStateException("Can not add kebind " + toAdd.getName() + " since the name already exists");
```
HighCrit: “Kebind”

HighCrit (changes requested):

common/src/main/java/com/wynntils/core/keybinds/KeyBindManager.java
magicus: Kaboom.

HighCrit (approved):

kristofbolyai (approved):

kristofbolyai:
We should make a PR to enable assertions in our Github Action build.

magicus:
Should we? We're not running it, so it won't show anything. That will only show like assertions in Gradle...

kristofbolyai:
Ah right. No brain moment. I just wanted a way to test PRs.

magicus:
With that being said, it would be fantastic if we had a way to run tests.

But the only ways I can think of require tons of work, and would probably not be able to make legally on GHA. :(

PR #356, opened 2022-08-28
Change render order of minimap

### LINKED ISSUE #355: Minimap Overlay covers other vanilla features like tab list and f3

Incompleteusern:
Fixes #355


kristofbolyai (approved):

kristofbolyai:
Can't we change the OverlayInfo default to Pre? It would be helpful, I do not see Post useful for many cases.

Incompleteusern:
> Can't we change the OverlayInfo default to Pre? It would be helpful, I do not see Post useful for many cases.

That might break something and would require discussion, so best left for a separate pr.

PR #360, opened 2022-08-28
Make Loom generate only client runs in IntelliJ

magicus:


kristofbolyai:
What docs did you use for this? Because there is some other thing they call client only jars iirc.

magicus:
I asked for advice in the Architectury discord, and got this reply: https://discord.com/channels/792699517631594506/1013533572914753536/1013536825228415056
> add loom.clientOnlyMinecraftJar() to the root build.gradle in subprojects

magicus:
It is implied, but not documented, at https://fabricmc.net/wiki/documentation:fabric_loom, which has the example:
```
	// When set only server related features and jars will be setup.
	serverOnlyMinecraftJar()
```

kristofbolyai (approved):
Should be okay then. This should improve AW/AT processing time too.

magicus:
AW/AT?

kristofbolyai:
Access widener/Access transformer. 
When remapping jars, now they can skip the server ones.

magicus:
Ah. Yeah.

PR #361, opened 2022-08-28
Make container query more robust (fixes #358)

### LINKED ISSUE #358: Quest book query failure
```[19:56:10] [Render thread/INFO] (Minecraft) [CHAT] Your class has automatically been selected. Use /class to change your class, or /toggle autojoin to turn this feature off.
[19:56:10] [Render thread/INFO] (Minecraft) [CHAT] 
[19:56:10] [Render thread/INFO] (Minecraft) [CHAT]                  §bSee yourself live on the web map: 
[19:56:10] [Render thread/INFO] (Minecraft) [CHAT]                            §d§nmap.wynncraft.com
[19:56:10] [Render thread/INFO] (wynntils) Scheduling character info query
[19:56:10] [Render thread/INFO] (wynntils) Entering world WC28
[19:56:10] [Render thread/INFO] (wynntils) Scheduling quest book query
[19:56:10] [Render thread/INFO] (wynntils) Requesting rescan of Quest Book
[19:56:11] [Render thread/INFO] (wynntils) Deducing character CharacterInfo[classType=Warrior/Knight, reskinned=false, level=105, id=-1]
[19:56:11] [Render thread/WARN] (wynntils) Error querying Quest Book:Container contents without associated container open
```

This does not seem to happen every time.

magicus:
I also refactored the ClientTickEvent. Of all events, this one is sent by far the most often. And yet, all callers needed to check the status and filter away 50% of the calls... Which also robbed the forge event bus of the possibility to optimize away event calls.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
+++ b/common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
@@ -71,18 +71,11 @@ public class ContainerQueryManager extends CoreManager {
 
     @SubscribeEvent
     public static void onMenuOpened(MenuEvent.MenuOpenedEvent e) {
-        if (currentStep == null && lastStep == null) return;
-        if (currentStep == null && lastStep != null) {
-            // We're in a possibly bad state. We have failed a previous call, but
-            // we might still get the menu opened (perhaps after a lag spike).
-            boolean matches = lastStep.verifyContainer(e.getTitle(), e.getMenuType());
-            if (matches) {
-                WynntilsMod.warn("Closing container '" + e.getTitle().getString()
-                        + "' due to previously aborted container query");
-                lastStep = null;
-                e.setCanceled(true);
-            } else {
-                lastStep = null;
+        if (currentStep == null) {
+            if (lastStep != null) {
+                // We're in a possibly bad state. We have failed a previous call, but
+                // we might still get the menu opened (perhaps after a lag spike).
+                handleFailedOpen(e);
             }
             return;
         }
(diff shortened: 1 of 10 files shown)


magicus:
Hopefully, this should fix #358. I can at least not reproduce it any longer.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -58,24 +62,40 @@ public static void runQuery(ContainerQueryStep firstStep) {
...
         if (!firstStep.startStep(null)) {
             raiseError("Cannot execute first step");
         }
     }
 
     @SubscribeEvent
     public static void onMenuOpened(MenuEvent.MenuOpenedEvent e) {
-        if (currentStep == null) return;
+        if (currentStep == null && lastStep == null) return;
+        if (currentStep == null && lastStep != null) {
```
kristofbolyai: Can't we "simplify" this? Or did you want to avoid right shift?

magicus: Different opinions on what "simple" means. The first check is the normal case, just wanted it to be clearly visible. But I agree it looks not so nice, will fix.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -58,24 +62,40 @@ public static void runQuery(ContainerQueryStep firstStep) {
...
+            // We're in a possibly bad state. We have failed a previous call, but
+            // we might still get the menu opened (perhaps after a lag spike).
+            boolean matches = lastStep.verifyContainer(e.getTitle(), e.getMenuType());
+            if (matches) {
+                WynntilsMod.warn("Closing container '" + e.getTitle().getString()
```
Incompleteusern: The container isn't actually closed?

magicus: We do ` e.setCanceled(true)` on the open request. This means the client will not see this as a request to open a container. So technically it's more "prohibit that it does not open" than "close", but that did not make a catchy error message :)

kristofbolyai (approved):
Seems okay.

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
Incompleteusern: Yes, but this is fired in response to a packet from the server. If we cancel the event, that just means we don't handle the packet client side, while the server still believes that the container is open.

PR #367, opened 2022-08-30
Quest Manager refactoring

magicus:
In preparation of the upcoming /quest command, refactor QuestBookManager: 
 * Rename to QuestManager
 * Collect all quest related functionality under one roof
 * Split up the unwieldy QuestManager in several, simpler and more specialized classes


common/src/main/java/com/wynntils/commands/QuestCommand.java
```diff
@@ -0,0 +1,154 @@
...
+                error.append(new TextComponent("\n - ").withStyle(ChatFormatting.GRAY))
+                        .append(new TextComponent(quest.getName()).withStyle(ChatFormatting.WHITE));
+            }
+            context.getSource().sendFailure(error);
+            return 0;
+        }
+
+        QuestInfo quest = matchingQuests.get(0);
+
+        MutableComponent response =
```
kristofbolyai: It will be so much fun to I18n commands. (This also applies to almost every other command class though..)

magicus: Not sure we need to do that, though. If we consider commands to be "advanced user" thing, it can probably stay English. (Don't know if you ever run like a localized Linux, and just gotten so annoyed at error messages being incomprehensible in your native language instead of English :-))

kristofbolyai (dismissed):
LGTM

Incompleteusern:
Still reviewing, but likely is fine

common/src/main/java/com/wynntils/gui/widgets/QuestButton.java
```diff
@@ -125,27 +118,8 @@ private void trackQuest() {
...
-                            .getAsString();
-                    Utils.openUrl(baseUrl + WebUtils.encodeForWikiTitle(pageTitle));
-                    return true;
-                })
-                .build();
-
-        RequestHandler handler = new RequestHandler();
-
-        handler.addAndDispatch(req, true);
+        McUtils.soundManager().play(SimpleSoundInstance.forUI(SoundEvents.EXPERIENCE_ORB_PICKUP, 1.0F));
```
Incompleteusern, line 121: We should have a helper method that plays sounds, not really this pr however

common/src/main/java/com/wynntils/wynn/model/quests/DialogueHistoryQueries.java
```diff
@@ -0,0 +1,99 @@
...
+                            createActualDialogueHistoryQuery(pageCount);
+                            break;
+                        }
+                    }
+                });
+
+        queryBuilder.build().executeQuery();
+    }
+
+    private void createActualDialogueHistoryQuery(int pageCount) {
```
Incompleteusern: I would suggest removing the `actual` from this, as this is more of a continuation of the query

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
```diff
@@ -0,0 +1,177 @@
...
+        this.nextTask = nextTask;
+        this.additionalRequirements = additionalRequirements;
+        this.isMiniQuest = isMiniQuest;
+        this.pageNumber = pageNumber;
+        this.tracked = tracked;
+    }
+
+    public String getName() {
+        return name;
+    }
```
Incompleteusern, line 66: Could this reasonably be made a record?

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
```diff
@@ -0,0 +1,177 @@
...
+        return level;
+    }
+
+    public String getNextTask() {
+        return nextTask;
+    }
+
+    public void setNextTask(String nextTask) {
+        this.nextTask = nextTask;
+    }
```
Incompleteusern, line 100: I suggest renaming `task` to `step` as we use `task` already for render tasks

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public static List<QuestInfo> getMiniQuests() {
+        return miniQuests;
+    }
+
+    public static List<QuestInfo> getQuestsSorted(QuestSortOrder sortOrder) {
+        return sortQuestInfoList(sortOrder, quests);
+    }
+
+    public static List<QuestInfo> getMiniQuestsSorted(QuestSortOrder sortOrder) {
```
Incompleteusern: Rename to have `Sorted` before `MiniQuests`/`Quests`?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -0,0 +1,190 @@
...
+    public static void toggleTracking(QuestInfo questInfo) {
+        CONTAINER_QUERIES.toggleTracking(questInfo);
+    }
+
+    public static void openQuestOnWiki(QuestInfo questInfo) {
+        final String baseUrl = "https://wynncraft.fandom.com/wiki/";
+
+        // TODO handle mini quest
+        String name = questInfo.getName();
+        String wikiQuestPageNameQuery = WebManager.getApiUrl("WikiQuestQuery");
```
Incompleteusern: check if apiurls exists before using

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -0,0 +1,190 @@
...
+            return PLAYER_LOCATION.distanceToSqr(location.toVec3());
+        }
+
+        @Override
+        public int compare(QuestInfo quest1, QuestInfo quest2) {
+            Optional<Location> loc1 = quest1.getNextLocation();
+            Optional<Location> loc2 = quest2.getNextLocation();
+            return (int) (getDistance(loc1) - getDistance(loc2));
+        }
+    }
```
Incompleteusern, line 180: This can reasonably be extracted to another class

common/src/main/java/com/wynntils/wynn/model/quests/QuestScoreboardHandler.java
```diff
@@ -29,35 +23,38 @@ public void onSegmentChange(Segment newValue, ScoreboardModel.SegmentType segmen
...
                         .append(" ");
             }
         }
 
-        String descriptionTrimmed = description.toString().trim();
+        Optional<QuestInfo> questInfoOpt =
+                QuestManager.getQuestFromName(questName.toString().trim());
+        if (questInfoOpt.isEmpty()) {
+            WynntilsMod.warn("Cannot match quest from scoreboard to actual quest: " + questName);
+            return;
```
Incompleteusern, line 42: Perhaps do something like 
```java
questInfo.setNextTask("Unknown");
```
if possible so the player knows that parsing has failed instead of showing an outdated task

Incompleteusern (dismissed):
A few minor comments, LGTM

magicus:
@Incompleteusern More or less all your comments applied as much to the code before I moved it. (The one exception is that I added the "play sound" line for opening on the wiki). 

But I'll go through them one by one and see what I can do.

common/src/main/java/com/wynntils/gui/widgets/QuestButton.java
magicus: It's not that complicated, but I agree, a helper would be a good thing.

https://github.com/Wynntils/Artemis/issues/509

common/src/main/java/com/wynntils/wynn/model/quests/DialogueHistoryQueries.java
magicus: Well, it *is* creating the actual query, since the other part is just a helper to extract the number of pages we need to query.

I've renamed it slightly, and added some more clarification, so the flow is hopefully clearer.

Me and boylai need to come up with a better way to create dynamic queries, so we don't need to know the number of pages beforehand. But that is a future exercise.

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
magicus: No, since nextTask is not final (and actually `status` should not be either). If we really wanted to, maybe we should split this into like a constant `QuestInfo` with just title, level, etc, and a `QuestProgress` which updates as we progress through a quest. 

It will help the conceptual design, but it will also make us need more code to compensate for the abstraction. I'll have to think about it. So not this PR at least. (But it's likely not the last one in QuestManager)

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
magicus: And we also use "step" for the next step in queries, which is much tighter coupled with the QuestManager than rendering is. 

I've used the term `nextTask` consistently, so it's never called just `task`. 

I think we'll just have to accept that commonplace words like "step" and "task" can be used with different meaning in different contexts.

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Actually, this should be the main way to get quests; the "raw" methods are (currently) not even used. I removed those and skipped the "sorted" part. If we ever need raw access again, let's name it something with "raw" instead.

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Sigh, the apiurls overhead is just bad. :(

Fixed it. Also noticed that the url pointed to the old domain...

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: What do you mean? It is another class already. Or do you mean that it should be a top-level class in its own file? If so, I disagree. This is just a small helper class; it was slightly too large to be a lambda, but it's only used in sortQuestInfoList.

Incompleteusern: I feel that it might be used in other places, but we can move it if that occurs. Indeed, it is only used here for now.

common/src/main/java/com/wynntils/wynn/model/quests/QuestScoreboardHandler.java
magicus: But if that happens we don't have a questInfo... If you're talking about what to do in case `questInfoOpt.isEmpty()`.

I think this error is unlikely enough that we don't have to worry about it. And it won't cause a crash.

Incompleteusern: It's more that I'd rather we not show the user outdated quest info when parsing failed, though this is niche as you say.

Incompleteusern (approved):

PR #372, opened 2022-08-31
ItemCompareFeature: Reset compare mode when world state changes

### LINKED ISSUE #332: Item to compare persists through a relog but doesn't draw circle
image

kristofbolyai:
Alternatively we can do this by implementing comparable (equals, compareTo) on all of our custom item stacks. I don't think this small fix is worth all that effort, but when we get those implemented, we can revert this to some extend (to allow /class changes).

Fixes #332


magicus (approved):

magicus:
When you say this could be fixed by a compareTo on custom item stacks, do you mean that the world it was retrieved from would be included in the comparison?

kristofbolyai:
No. The problem is we are comparing item stacks in render, and the item and it's stats are the same, but it is a different instance now because /class or user reopened some GUI. So it breaks the feature.

PR #375, opened 2022-08-31
MiniMapOverlayFeature: Fix map being black underwater

### LINKED ISSUE #355: Minimap Overlay covers other vanilla features like tab list and f3

kristofbolyai:
Fixes #355


Incompleteusern (approved):
It might be nice to have a specialized overlay gui hook in the future, or automatically have `RenderSystem.setShaderColor(1f, 1f, 1f, 1f);` be called, but this works for now

kristofbolyai:
> It might be nice to have a specialized overlay gui hook in the future, or automatically have `RenderSystem.setShaderColor(1f, 1f, 1f, 1f);` be called, but this works for now

Yeah, I don't want to disclose how much time I spent looking at Gui#render, but I could not figure out where the water overlay is rendered. I have a bet it is right before the debug screen, so there is not really anything we can do.

PR #376, opened 2022-09-01
Fixup .gitattributes

obj-obj:
`text=auto` makes git only apply line endings to files it thinks is text, rather than affecting all files, so file types don't have to be specified as binary manually.

https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto


.gitattributes
```diff
@@ -1,2 +1 @@
-* text eol=lf
-*.png binary
\ No newline at end of file
+* text=auto eol=lf
```
kristofbolyai: Why put them in the same line though?

kristofbolyai:
Good catch though

.gitattributes
magicus: Missing linebreak at EOF

obj-obj: Oops

obj-obj:
You also might want to consider adding a `.editorconfig` (https://editorconfig.org/)

magicus (approved):

magicus:
> You also might want to consider adding a `.editorconfig` (https://editorconfig.org/)

We have automatic formatting from Spotless. If you care to craft a `.editorconfig` that matches the Spotless formatting so much it seems worth the point, go ahead. Otherwise, it would probably make more sense to add more IntelliJ options, since that is the IDE that all core Artemis developers are using. (Not saying any other IDE is not allowed, just that we have optimized our workflow for IntelliJ).

PR #379, opened 2022-09-03
Features: Add CustomCommandKeybindsFeature

kristofbolyai:
Fun fact: This allows users to use keybinds to set/reset configs. Full screen map done. :P


common/src/main/java/com/wynntils/features/user/CustomCommandKeybindsFeature.java
```diff
@@ -0,0 +1,84 @@
...
+    private String keybindCommand3 = "";
+
+    @Config
+    private String keybindCommand4 = "";
+
+    @Config
+    private String keybindCommand5 = "";
+
+    @Config
+    private String keybindCommand6 = "";
```
Incompleteusern, line 33: Code dup is annoying, but necessary as of now

Incompleteusern:
> Fun fact: This allows users to use keybinds to set/reset configs. Full screen map done. :P

Not quite, as only one command can be run per keybind

common/src/main/java/com/wynntils/features/user/CustomCommandKeybindsFeature.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @RegisterKeyBind
+    private final KeyBind executeKeybind6 = new KeyBind(
+            "Execute 6th Custom Command Keybind",
+            GLFW.GLFW_KEY_UNKNOWN,
+            true,
+            () -> this.executeKeybind(keybindCommand6));
+
+    private void executeKeybind(String keybindCommand) {
+        keybindCommand = keybindCommand.trim();
```
Incompleteusern: Modifcations should be done only once in `onConfigUpdate`

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -67,6 +68,72 @@
 public final class FeatureRegistry {
     private static final List<Feature> FEATURES = new ArrayList<>();
 
+    public static void init() {
```
magicus, line 71: You moved it to make it easier to find?

kristofbolyai: Yes, this is literally the only method we touch in the class.

magicus (commented):
Would be nice if we could get to a point where we could generalize this. I agree with Incomplete's idea on cleaning up the command at config time. Otherwise it looks good.

Incompleteusern (approved):

PR #380, opened 2022-09-03
Fix color not being added

Incompleteusern:
image

Continuation of #377, fixes a bug

PR #382, opened 2022-09-03
Fix TextRenderSetting using the wrong values

byBackfish:
Was looking at #380 and then saw some other cases.


kristofbolyai (approved):

Incompleteusern:
Thanks!

magicus:
@byBackfish Good thinking!

PR #387, opened 2022-09-06
Initial issue templates

Incompleteusern:
Here are initial templates for issues, feel free to give suggestions/changes.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,29 @@
+---
+name: Bug report
+about: Create a report to help us improve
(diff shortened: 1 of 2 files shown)


magicus:
I disagree with the idea of having templates, at least for this point in time. They will just be annoying for us. We don't have a high amount of external bug reports of low quality that we need to slow down.

If anything, we could consider having a simple link to the site whatever it was called, where people where contributing ideas, in the feature request template.

magicus:
... but I guess even having such a template means we will need extra clicks just to open a simple bug report, so on second thought, I dislike even that.

Incompleteusern:
Good point, unnecessary as of now.

PR #389, opened 2022-09-07
Enable all text overlays for items in hotbar by default.

ekgame:
<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>
	<td>
</table>

There is a setting for enabling/disabling text overlays in hotbars, however it is disregarded for most items. This pull request makes all items render in the hotbar as expected when the setting is enabled.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/item/properties/ConsumableChargeProperty.java
+++ b/common/src/main/java/com/wynntils/wynn/item/properties/ConsumableChargeProperty.java
@@ -28,9 +28,4 @@ public class ConsumableChargeProperty extends CustomStackCountProperty {
     public boolean isTextOverlayEnabled() {
         return ItemTextOverlayFeature.consumableChargeEnabled;
     }
-
-    @Override
-    public boolean isHotbarText() {
-        return true;
-    }
 }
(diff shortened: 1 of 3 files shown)


kristofbolyai:
> There is a setting for enabling/disabling text overlays in hotbars, however it is disregarded for most items.

What do you mean?

kristofbolyai:
Oh, I see. What I don't get is how this change does not ignore the "hotbar text overlays" setting? 

Does the fix make it so everything  gets rendered unless specified separately otherwise or the setting is turned off?

ekgame:
Whether or not text overlays should be rendered in the hotbar is handled here: https://github.com/Wynntils/Artemis/blob/f441312ae88d1d99ffc2764f4938f442bbce8fe9/common/src/main/java/com/wynntils/features/user/inventory/ItemTextOverlayFeature.java#L84 

It can be turned on/off in the settings: 
paveikslas


However, the `isHotbarText` method is an override - if it returns false, it disregards the setting and refuses to render the overlays as expected.

kristofbolyai (approved):

kristofbolyai:
Nice! Thanks.

PR #391, opened 2022-09-07
OverlaySelectionScreen: Fix back button

kristofbolyai:


common/src/main/java/com/wynntils/screens/overlays/OverlaySelectionScreen.java
```diff
@@ -42,7 +41,7 @@ protected void init() {
                 BUTTON_WIDTH,
                 BUTTON_HEIGHT,
                 new TranslatableComponent("screens.wynntils.overlaySelection.close"),
-                button -> McUtils.mc().setScreen(lastScreen)));
+                button -> McUtils.mc().setScreen(new PauseScreen(true))));
```
magicus, line 44: Should it not go to the Wynntils menu? I thought that was were we came from?

kristofbolyai: Aah right. Problem was it was stuck switching between the two overlay screens.

PR #393, opened 2022-09-08
Fix casting spells while dropping items.

ekgame:
|The issue|
|-|
| <video src="https://user-images.githubusercontent.com/1730245/189118157-5646699c-fc5a-4345-8f12-fe950494f6ca.mp4"> |

Apparently, sometime between 1.13-1.18, they added a hand swing when you drop an item. This feature introduces an issue in Wynncraft where if you are holding a bow and start dropping items from inventory - it starts triggering the spell sequence.

This patch adds a mixin to effectively bypass the hand swing call when dropping items (only when dropping directly from inventory screen). This fixes the issue of inadvertently casting spells that would be caused by swinging your hand.


magicus:
Good catch, but I'd like to see a slightly different resolution. First of all, is it in any way possible to detect if this is a non-spell-casting start from the spell detection code? E.g. by checking if a container screen is open. I think this would be the best solution.

If that is not feasible, then we need to fix this the "proper" way, and that is to create and send out an event which we then will cancel. Otherwise, this fix will affect players even if playing on singleplayer or other servers, and that is one of the stated goals of Artemis not to do that.

kristofbolyai:
> Good catch, but I'd like to see a slightly different resolution. First of all, is it in any way possible to detect if this is a non-spell-casting start from the spell detection code? E.g. by checking if a container screen is open. I think this would be the best solution.
> 
> If that is not feasible, then we need to fix this the "proper" way, and that is to create and send out an event which we then will cancel. Otherwise, this fix will affect players even if playing on singleplayer or other servers, and that is one of the stated goals of Artemis not to do that.

Basically, make this a feature, and use events so they are only sent out on Wynn. Otherwise, nice fix.

magicus:
@kristofbolyai Well, we already have a feature, but it's buggy... What we need is an event. 

But. As I said, my preferred solution is to skip the mixin altogether and instead check if we're inside an container UI when looking for the first click.

kristofbolyai:
@ekgame Are you planning on continuing work on this PR?

ekgame:
Yes, I'll revisit and update when I have time it. 

> @kristofbolyai Well, we already have a feature, but it's buggy... What we need is an event.
> 
> But. As I said, my preferred solution is to skip the mixin altogether and instead check if we're inside an container UI when looking for the first click.

I'm not sure I understand how the preferred solution would work. In what context is only the first click relevant? Is there some logic that tries to detect a spell being cast and modify the event somehow? I don't really see a less hacky way to patch this than simply using a mixin.

Note that this is not a bug with the quick cast feature or with the mod - this issue also happens in vanilla, I just feel like it's appropriate for the mod to try and patch it.

kristofbolyai:
> Yes, I'll revisit and update when I have time it.
> 
> > @kristofbolyai Well, we already have a feature, but it's buggy... What we need is an event.
> > But. As I said, my preferred solution is to skip the mixin altogether and instead check if we're inside an container UI when looking for the first click.
> 
> I'm not sure I understand how the preferred solution would work. In what context is only the first click relevant? Is there some logic that tries to detect a spell being cast and modify the event somehow? I don't really see a less hacky way to patch this than simply using a mixin.
> 
> Note that this is not a bug with the quick cast feature or with the mod - this issue also happens in vanilla, I just feel like it's appropriate for the mod to try and patch it.

I myself not sure what magicus meant, but I think I know what he wants. 


Post an event in the mixin, make it cancelable. If the event is cancelled, cancel the swing. (If it is not, don't modify vanilla behavior). Then, create a user feature, and make it cancel this event. This way, this fix is only applied on Wynncraft (due to the nature of EventFactory) and can be disabled by the user (due to the nature of UserFeature).

Is it clear now?

common/src/main/java/com/wynntils/features/alwayson/FixCastingSpellsFromInventoryFeature.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * This feature fixes an issue where a Archers holding bows may start
+ * casting spells just by dropping items from the inventory screen.
+ */
+public class FixCastingSpellsFromInventoryFeature extends StateManagedFeature {
+
+    @SubscribeEvent
+    public void onSetSlot(ArmSwingEvent event) {
+        if (event.getActionContext() == ArmSwingEvent.ArmSwingContext.DROP_ITEM_FROM_INVENTORY_SCREEN
```
kristofbolyai: This might have gotten lost in our conversation (or I failed to communicate this), but we don't really need all of the mixins to post events for all action contexts when we only use one? I think they should he removed, and only post when context is this one (and we can just extend the event when we need it). Keeping the enum is fine.

ekgame:
Now the fix is implemented using events and a feature, as described. Note that I elected to make the feature always on, because the other "fix" type feature is also always on and I don't see a good reason to turn it off.

common/src/main/java/com/wynntils/features/alwayson/FixCastingSpellsFromInventoryFeature.java
```diff
@@ -0,0 +1,25 @@
...
+import com.wynntils.core.features.StateManagedFeature;
+import com.wynntils.mc.event.ArmSwingEvent;
+import net.minecraft.world.InteractionHand;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+/**
+ * This feature fixes an issue where a Archers holding bows may start
+ * casting spells just by dropping items from the inventory screen.
+ */
+public class FixCastingSpellsFromInventoryFeature extends StateManagedFeature {
```
kristofbolyai: We can make this an UserFeature. I get there is no point in disabling this, but StateManagedFeatures are reserved for edge cases in Artemis (and likely will be removed). 

Overall, imo it is fine to give the user the option to disable this.

common/src/main/java/com/wynntils/mc/mixin/PlayerMixin.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Redirect(
+            method = "drop(Lnet/minecraft/world/item/ItemStack;ZZ)Lnet/minecraft/world/entity/item/ItemEntity;",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lnet/minecraft/world/entity/player/Player;swing(Lnet/minecraft/world/InteractionHand;)V"))
+    private void onSwingInInventoryScreen(Player player, InteractionHand hand) {
+        if (!EventFactory.onArmSwing(ArmSwingEvent.ArmSwingContext.DROP_ITEM_FROM_INVENTORY_SCREEN, hand)
```
kristofbolyai: Please flip this if condition. Return if event is cancelled, otherwise swing (to reduce code right shift)

kristofbolyai:
> Now the fix is implemented using events and a feature, as described. Note that I elected to make the feature always on, because the other "fix" type feature is also always on and I don't see a good reason to turn it off.

Disabling the other fix feature breaks item parsing in Artemis. That is why it is always turned on.

kristofbolyai (commented):
Overall, this is looking good. Remove the "unused" mixins fron MinecraftMixin, and it will be ready for shipping IMO.

common/src/main/java/com/wynntils/mc/mixin/MinecraftMixin.java
```diff
@@ -37,4 +41,91 @@ public void tickPost(CallbackInfo ci) {
     public void resizeDisplayPost(CallbackInfo ci) {
         EventFactory.onResizeDisplayPost();
     }
+
+    @Redirect(
```
magicus: Holy smoke, there's a lot of `@Redirect`s instead of `@Inject` here!

magicus (changes requested):

magicus:
I'm sorry I've been away from Wyntils development for a while. 

This still looks like a *waaay* more complicated fix than is needed for the problem, as described.

The issue, as far as I can understand, is that we start casting a spell even though the inventory is open. In that case, we should just check if there is an inventory screen open, and if it is, we should not consider it part of the spell casting.

magicus:
There is a thing like `mc().player.inventory` or `mc().player.container` that you can check for non-null.

magicus:
And if that does not work out, for some reason, we should at the very least not have Redirects instead of Injects. Redirects are very unfriendly, and will make other mods break.

kristofbolyai:
> There is a thing like `mc().player.inventory` or `mc().player.container` that you can check for non-null.

McUtils.mc().screen instanceof AbstractContainerScreen<?>. Inject this into the swing method, cancel if true.

magicus:
With that said, I kind of liked the 
```
public enum ArmSwingContext {
        DROP_ITEM_FROM_INVENTORY_SCREEN,
        DROP_ITEM_FROM_HOTBAR,
        ATTACK_OR_START_BREAKING_BLOCK,
        BREAKING_BLOCK,
        INTERACT_WITH_BLOCK,
        INTERACT_WITH_ENTITY,
        USE_ITEM,
    }
```
thing. It might turn out useful in the future. I'm not too keen in taking in code that is not being used (after all, code that is "for future use", or "was used before, but keep it along for now" is bound to be broken, and is introducing complexity into the code for no reason), but we can at least try to remember this PR if we want to revive it. (Given that we replace the Redirects with Injects).

kristofbolyai:
Hey, @ekgame. As I don't want this PR hanging, I will take over and finish this. Thank you for pointing the issue out, and bringing us an almost perfect code for fixing it.

common/src/main/java/com/wynntils/mc/event/ArmSwingEvent.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public enum ArmSwingContext {
+        DROP_ITEM_FROM_INVENTORY_SCREEN,
+        DROP_ITEM_FROM_HOTBAR,
+        ATTACK_OR_START_BREAKING_BLOCK,
+        BREAKING_BLOCK,
+        INTERACT_WITH_BLOCK,
+        INTERACT_WITH_ENTITY,
+        USE_ITEM,
```
Incompleteusern: This enum is only implemented for one variable, so likely comment out the other values for now.

magicus:
I realize now I misunderstood the bug report. :-( Apologies. I thought it was about our quick cast spell logic. I missed the post where @ekgame explicitly said it wasn't.

That explains why I was pushing for a completely bonkers solution instead. :)

magicus (approved):
This looks good. Thanks everyone, and once again sorry for misunderstanding and sounding so negative. :(

Incompleteusern (approved):

ekgame:
Hey, sorry about over-engineering a feature and seemingly abandoning it. I'm glad somebody picked it up in my absence.

kristofbolyai:
> Hey, sorry about over-engineering a feature and seemingly abandoning it. I'm glad somebody picked it up in my absence.

No problem, I hope we did not scare you away from contributing. :cold_sweat:

magicus:
I'd like to continue on the ArmSwingContext thing. Or, perhaps differently, we have a very odd collection of events right now for when the player left or right clicks, while holding stuff, or on blocks etc. These should be more aligned.

PR #394, opened 2022-09-09
chore(ci): Port Publishing from Wynntils

HighCrit:
A port of https://github.com/Wynntils/Wynntils/pull/572


magicus:
While I do like the thought of publishing automatically, this seems a bit premature for Artemis. We are nowhere near being able to publish anything. Also, perhaps even more importantly, with the mod in such flux, we should not be hampered by changelog-like commit descriptions.

magicus:
So, to be clear, sure, we can port this now, but we should not run the publish targets, and we should be very lax in the variance of commit descriptions we accept.

(Also, there seemed to be some javascript version stuff, since we don't even really have version numbers in Artemis, have you done any adaptations to that code to try and handle Artemis, or is it just copy/paste from legacy?)

(And there, too, we will need proper versioning, but we're not there yet.)

kristofbolyai:
I think it is fine to have alpha builds, Artemis is usable as of right now, and many people use it now.

HighCrit:
> While I do like the thought of publishing automatically, this seems a bit premature for Artemis. We are nowhere near being able to publish anything.  

This is precisely what the release types are for. If everyone shipped finished products there would be no need for beta or alpha release types. Also, this allows people to more easily see and manage available updates in their modpacks and is one of the reasons why I don't daily drive Artemis yet.


>  Also, perhaps even more importantly, with the mod in such flux, we should not be hampered by changelog-like commit descriptions.  

Looking at previous commits it doesn't look like too big of a change - just remove 'ure' from 'feature' and add a semicolon after 'fix'.

> So, to be clear, sure, we can port this now, but we should not run the publish targets  

If we don't publish to targets what is the point of porting?


> and we should be very lax in the variance of commit descriptions we accept.

I can remove the commit constraint but then we need to pay extra attention to the merge commit titles


> (Also, there seemed to be some javascript version stuff, since we don't even really have version numbers in Artemis, have you done any adaptations to that code to try and handle Artemis, or is it just copy/paste from legacy?)

The version of Artemis is setup exactly the same way as Wynntils in build.gradle so 🤷 


> (And there, too, we will need proper versioning, but we're not there yet.)

I would like to see use starting at 0.1.1 and when Artemis is ready release it as 1.0.0

@magicus

magicus:
To be clear, I think we definitely should publish the mod, and we should have proper versioning, and we might even consider the "changelog from commit messages" thing (even if I am decidedly more skeptical towards that -- the target audience is very different between commit messages and changelogs; the former is for developers keeping track of the source code, the latter is for end users. Any overlap is just seemingly, and will fail to consider the needs of either (or both) of these groups. A good changelog cannot be created automatically from good commit messages.)

I just think it is still too early. If we start publishing this too soon, we will just get dragged down by user support. I frankly don't think publishing alpha releases on an external site is a good thing. The fact that CurseForge allows you to tag a release as "alpha" does not mean it is a good idea to use, any less than the `<blink>` html tag ever was a good idea.

DevScyu (approved):

.github/workflows/build.yml
```diff
@@ -31,18 +29,115 @@ jobs:
...
+          create-summary: true
+          version-file: ./version.json
+          skip-git-pull: true
+
+  publish-gh:
+    name: Publish to Github Releases
+    runs-on: ubuntu-latest
+    needs: [ changelog, artifact ]
+    steps:
+      - uses: actions/download-artifact@v3
```
DevScyu, line 78: Probably will need a name for the artifact download

.github/workflows/publish.yml
```diff
@@ -0,0 +1,67 @@
...
+        type: string
+    secrets: inherit
+
+jobs:
+  release-to-cf:
+    name: Release to CurseForge
+    runs-on: ubuntu-latest
+    needs: [changelog, artifact]
+    steps:
+      - uses: actions/download-artifact@v3
```
DevScyu, line 33: will need artifact name to download

DevScyu (changes requested):

.github/workflows/build.yml
HighCrit: I used the following https://github.com/actions/download-artifact#download-all-artifacts

.github/workflows/build.yml
```diff
@@ -31,18 +29,115 @@ jobs:
...
+
+      - name: Create release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            - "forge/build/libs/wynntils-${{ steps.changelog.outputs.tag }}.MC1.18.2-forge.jar"
+            - "fabric/build/libs/wynntils-${{ steps.changelog.outputs.tag }}.MC1.18.2-fabric.jar"
```
HighCrit, line 85: Can we use both the multiline and array syntax simultaneously?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Upload to Modrinth
+        uses: RubixDev/modrinth-upload@v1.0.0
+        with:
+          token: ${{ secrets.MODRINTH_API_TOKEN }}
+          file_path: "${{ inputs.file_path }}"
```
HighCrit, line 59: We may be able to move Modrinth a as it supports multiple files per version, but i couldn’t find whether the action supported it.

kristofbolyai:
Well, I don't know what to do with this PR. I am not against publishing, but @magicus voiced concerns...

.github/workflows/build.yml
magicus: No, that looks odd. I'm no yaml expert, but I think this will make a string "- forge/... - fabric/..." etc, instead of an array [forge/..., fabric/...]

magicus:
Ok, I hear that you really want this. I am removing my objection. But I think the actual code might need a pass, or so. (I have not looked at it specifically until now.)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,67 @@
+on:
+  workflow_call:
```
magicus, line 2: So we will not auto-publish, but it will require a manual intervention? That sounds better to me, I thought you wanted to automatically push every build.

Edit: Oh, I see, this is called from the `on:push` anyway...

.github/workflows/publish.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+      - uses: actions/download-artifact@v3
+
+      - name: Upload to CurseForge
+        uses: itsmeow/curseforge-upload@v3
+        with:
+          file_path: "${{ inputs.file_path }}"
+          game_endpoint: minecraft
+          game_versions: "${{ inputs.cf_mc_version }}"
+          project_id: 303451
```
magicus, line 41: Should we not have our own project? This is the Wynntils project. I think it will be very confusing to users, publishing Artemis builds there. Or won't it?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,67 @@
...
+          token: ${{ secrets.MODRINTH_API_TOKEN }}
+          file_path: "${{ inputs.file_path }}"
+          name: "${{ inputs.release_name }}"
+          version: "${{ inputs.tag }}"
+          changelog: "${{ inputs.changelog }}"
+          game_versions: "${{ inputs.mr_mc_version }}"
+          release_type: alpha
+          loaders: "${{ inputs.mr_loader }}"
+          featured: false
+          project_id: dU5Gb9Ab
```
magicus, line 67: Same here; this is the legacy Wynntils project.

.github/workflows/build.yml
```diff
@@ -1,16 +1,14 @@
-name: Spotless and Build
+name: Publish Release
```
magicus, line 1: Both the name of the file and this name here is conflicting, and neither seem to be good. 

This file should be named `push.yml`. The name should probably be `On push`, or possibly `Build and publish`.

.github/workflows/build.yml
```diff
@@ -31,18 +29,115 @@ jobs:
...
+          release-count: 1
+          pre-commit: .pre-commit.js
+          create-summary: true
+          version-file: ./version.json
+          skip-git-pull: true
+
+  publish-gh:
+    name: Publish to Github Releases
+    runs-on: ubuntu-latest
+    needs: [ changelog, artifact ]
```
magicus, line 76: I think you mean that it needs `build`. There is no `artifact` dependency afaics.

.github/workflows/build.yml
```diff
@@ -31,18 +29,115 @@ jobs:
...
+      release_name: "Wynntils (Artemis) ${{ steps.changelog.outputs.tag }}-forge"
+      tag: ${{ steps.changelog.outputs.tag }}
+      changelog: ${{ steps.changelog.outputs.changelog }}
+      cf_mc_version: "9008,7498,8326,8634"
+      mr_mc_version: 1.18.2
+      mr_loader: forge
+      secrets: inherit
+
+  # Minecraft 1.18.2: 9008
+  # Forge: 7499
```
magicus, line 110: By this, do you mean `Fabric`? (Also same question about Quilt below)

.github/workflows/build.yml
magicus: I don't really like the lack of consistency here. Do we *know* that the output files are really named as the changelog action claims, `-${{ steps.changelog.outputs.tag }}`? If so, we should use that very name instead of the `*` when uploading the artifacts. Or do the changelog module rename the files?

.github/workflows/build.yml
```diff
@@ -31,18 +29,115 @@ jobs:
...
-        uses: gradle/gradle-build-action@v2
+  changelog:
+    name: Generate Changelog
+    needs: [build]
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.changelog.outputs.tag }}
+      skipped: ${{ steps.changelog.outputs.skipped }}
+      clean_changelog: ${{ steps.changelog.outputs.clean_changelog }}
+      changelog: ${{ steps.changelog.outputs.changelog }}
```
magicus, line 54: Is this a path to the changelog? But how can that be accessed in later steps, since this is run on a separate VM? 🤔

magicus: Or is it just the changelog as a string? How is it formatted? Can it really be passed on properly to all other steps?

.pre-commit.js
```diff
@@ -0,0 +1,17 @@
+const fs = require("fs");
```
magicus, line 1: So what is this? Just a helper for the `changelog` action`? If so, I think it should move to `.github/helpers`

.github/workflows/build.yml
HighCrit: Yes 😅

.github/workflows/build.yml
HighCrit: The changelog is passed to the desired workflow. Just like it is for Wynntils but for Artemis we control the intermediary workflow that pushes to Modrinth and Curseforge. However, Scyu shared another workflow that does all in one.

build.gradle
```diff
@@ -121,12 +121,17 @@ allprojects {
...
-
     //If someone can find a way to access the mcVersion without using it as a param please do
     String toString(Object minecraft_version) {
-        "${major}.${minor}.${patch}+${this.build}.MC${minecraft_version}"
+        "${major}.${minor}.${patch}.MC${minecraft_version}"
     }
 }
+
+task installLocalGitHook(type: Copy) {
+    from new File(rootProject.rootDir, '.github/hooks/commit-msg')
```
magicus, line 131: I think you mean utils/git-hooks...

This looks very hacky to me. Is it impossible to make git pickup a hook automatically? Using gradle here to potentially copy a file on every build seems the wrong way to go about this.

build.gradle
```diff
@@ -121,12 +121,17 @@ allprojects {
...
+    into { new File(rootProject.rootDir, '.git/hooks') }
+    fileMode 0775
+}
+
+prepareKotlinBuildScriptModel.dependsOn installLocalGitHook
```
magicus, line 136: Why setting up stuff for Kotlin? We have no Kotlin support.

.github/workflows/build.yml
HighCrit: You are correct

.github/workflows/build.yml
HighCrit: @DevScyu 

> Can we use both the multiline and array syntax simultaneously?

.github/workflows/publish.yml
HighCrit: We could change it to workflow call but then we should discus when we plan to release new features. It may also help to use the stability info of features to make it user configurable which 'stream' they are on.

version.json
```diff
@@ -0,0 +1,3 @@
+{
+  "version": "0.1.1"
```
magicus, line 2: That sounds like a very random version number to start with. Even if we all agree that bumping the major version from 0 to 1 should happen at release, what to you think is the criteria for bumping minor and patch?

I don't understand how all these github actions work or interact. Is there some way this code associates either an automatically incremented build number (like a traditional CI build system like Jenkins), or an arbitrarily/random (e.g. latest git hash) build number? Or is this the version string in it's entirety? Then it either needs to be updated for every push, or we will publish multiple builds with the same version string. Neither is ideal in any way.

.github/workflows/publish.yml
HighCrit: Personally I don't think so, we publish it as an alpha. And the sites clearly mark which MC version is targeted.

.github/workflows/build.yml
magicus: Ok, I don't know how this is done in Legacy, so you'll have to explain this PR to me like I'm four years old. ;-)

Is it a string? Formatted? How? Markdown?

magicus:
Also, if we should follow these new changelog-required prefixes, I doubt there is a `chore(ci):` prefix. In fact, this does not seem a chore at all, to me.

version.json
HighCrit: Version bumps are dictated by:
```
fix: fixed NPE in class (increases the REVISION number by 1)
feat: Added minimap (increases the MINOR number by 1)
chore: fixed typo (does not increase version number)
ci: added github actions (does not increase version number)
docs: updated readme (does not increase version number)
```

build.gradle
HighCrit: @DevScyu

build.gradle
HighCrit: @DevScyu

.pre-commit.js
HighCrit: @DevScyu

version.json
magicus: So how does that work? Do the changelog action push commits changing version.json?

Also, you realize this will mean we will end up with version numbers like `0.29.301`? (Traditional semver assumed that the last part were updated for a new release, not for every committed fix)

kristofbolyai:
We will have to re-do this whole PR since legacy publishing changed a lot. Closing this (until we port it or indefinitely as a new PR might be better.)

PR #396, opened 2022-09-10
WebManager: Map loading changes

kristofbolyai:
This PR adds more logic, replaces map list with a much faster concurrent object and tries to remove the "hacky" parts of map piece downloading.

### DIFF
--- a/common/src/main/java/com/wynntils/core/webapi/WebManager.java
+++ b/common/src/main/java/com/wynntils/core/webapi/WebManager.java
@@ -38,13 +38,14 @@ import java.net.URLConnection;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import net.minecraft.ChatFormatting;
 import net.minecraft.network.chat.ClickEvent;
 import net.minecraft.network.chat.MutableComponent;
@@ -76,7 +77,7 @@ public final class WebManager extends CoreManager {
 
     private static WynntilsAccount account = null;
 
-    private static List<MapProfile> maps = new ArrayList<>();
+    private static Queue<MapProfile> maps = new ConcurrentLinkedQueue<>();
 
     private static final String USER_AGENT = String.format(
             "Wynntils Artemis\\%s-%d (%s)",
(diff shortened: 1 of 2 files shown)


kristofbolyai:
@Incompleteusern can you confirm these changes are okay?

common/src/main/java/com/wynntils/features/user/overlays/map/MiniMapOverlayFeature.java
```diff
@@ -107,7 +107,7 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
 
             // TODO replace with generalized maps whenever that is done
             if (WebManager.isMapLoaded()) {
-                MapProfile map = WebManager.getMaps().get(0);
+                MapProfile map = WebManager.getMaps().peek();
```
HighCrit, line 110: Why no .poll() isnt or do we need to contents in the future elsewhere?

kristofbolyai: This is render loop. We need this every render.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -286,34 +295,35 @@ public static CompletableFuture<Boolean> tryLoadMaps() {
...
-                        }
-                    });
+                    if (downloadedMapProfiles.size() == mapArray.size()) {
+                        maps = downloadedMapProfiles;
+                        result.complete(true);
+                    } else {
+                        WynntilsMod.error("Expected to have" + mapArray.size() + " map parts, got "
+                                + downloadedMapProfiles.size() + " parts.");
+                        result.complete(false);
+                    }
```
Incompleteusern, line 326: The reason for the hack is that having added a bunch of map download requests done asynchronously, we need to do an operation when all are complete.

Now, we are doing the whole map download synchronously, which likely dramatically increases startup time

kristofbolyai: Yeah, but can't that make a race somewhere? As we do not group requests, and only dispatch 1 by 1

kristofbolyai:
@Incompleteusern are you fine with these changes, or do you think we should close this?

common/src/main/java/com/wynntils/core/webapi/WebManager.java
Incompleteusern: I feel the solution here is to add better grouping functionality/use the existing grouping functionality and code that runs when the entire group is done by default.

Incompleteusern:
I would rather we keep the async functionality and the hack than remove the hack.

Incompleteusern:
I may try my own hand at removing the hack sometime.

PR #398, opened 2022-09-10
CustomBarsOverlayFeature: Add Mana Bank and Health Bank overlay bars

kristofbolyai:
image
image

I already know @magicus won't like https://github.com/Wynntils/Artemis/commit/0445b0626910db1a4ad8f3f722f75012524f1795 . However, there is no other way of doing this, sanely. I think this is our best bet. I made it so if the events are not used, we do not "override" the default minecraft rendering.


magicus:
I have not played 2.0 nearly enough. Can there only be one "bank" at the time?

kristofbolyai:
> I have not played 2.0 nearly enough. Can there only be one "bank" at the time?

Yes, it is a mage ability upgrade. It is a boss bar, if any mana is present in the bank.

magicus:
You are absolutely correct that I don't like the changes in `BossHealthOverlayMixin`. :-) This will cause incompatibilities with any other mod that tries to modify how boss bars is rendered. I understand if it is tricky to get this right otherwise. If you can hold on a while to this and let me think about it for a while. (Or if you push it, and are prepared to work with me changing the implementation if we figure out a better way.)

kristofbolyai:
@magicus revert last commit and push?

kristofbolyai:
@magicus This should revert the commit you don't like, is it fine to merge otherwise?

magicus:
So this means you can get a new overlay, but you can't cancel the boss bar, so you'll get the same info twice?

If so, well, it's better code-wise, but it's worse UX-wise...

kristofbolyai:
> So this means you can get a new overlay, but you can't cancel the boss bar, so you'll get the same info twice?
> 
> 
> 
> If so, well, it's better code-wise, but it's worse UX-wise...

Yep..

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
```diff
@@ -120,25 +142,75 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
...
+                    new GuiScaledOverlaySize(81, 21));
         }
 
         @Override
-        protected void onConfigUpdate(ConfigHolder configHolder) {}
+        public void render(PoseStack poseStack, float partialTicks, Window window) {
+            if (!WynnUtils.onWorld()) return;
+
+            LerpingBossEvent poolEvent = null;
+            String bloodPoolPercent = "";
```
magicus, line 180: This is not a rendering thing. Can you extract this into some Util class (one existing, or invent a new if none seem good enough), so other code in the future that wants to read the blood pool value can do so easily? (Same for the Mana Bank)

kristofbolyai:
> So this means you can get a new overlay, but you can't cancel the boss bar, so you'll get the same info twice?
> 
> 
> 
> If so, well, it's better code-wise, but it's worse UX-wise...

Mc only can render 3 boss bars, so I think this limit can be reached (guild info, mob hp, blood pool, shaman mask). This make is so it is seen even then.

magicus:
But yes, if you want to get this in quickly, I can accept it with a double bar, and then we can attack the problem on how to hide a boss bar selectively, in the least intrusive way, as a separate PR.

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
magicus: @kristofbolyai Did you miss this, or intentionally skip doing it?

kristofbolyai: ooh, I forgot. I will do this later separately.

kristofbolyai: I will keep tabs on this and do it.

magicus: Thanks! <3

PR #399, opened 2022-09-10
Features: Add AbilityRefreshRedirectFeature

kristofbolyai:
image


magicus (approved):
Looks good. We might consider collecting all "chat -> notification" redirects in one place later on, though, and select which messages to redirect by config variables instead.

PR #405, opened 2022-09-13
Report modloader to server

Incompleteusern:
The first commit serves the purpose of reporting the mod loader to the server, while the latter two commits are minor changes


common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -73,24 +77,25 @@ public static void info(String msg) {
         LOGGER.info(msg);
     }
 
-    public static void onResourcesFinishedLoading() {
-        if (featuresInited) return;
```
kristofbolyai, line 77: Why remove this?

magicus: Yeah, you can't do that. This is a guard to make sure we only do the init once. Resources can reload any number of times, we only want the first.

Also, the idea with initFeatures() was to create an abstraction barrier, to create for ourselves a way of doing "init when resources are available". I am not fond of you removing it. But maybe it needs comments to explain why it is as it is.

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -99,14 +104,9 @@ public static void init(String modVersion, boolean isDevelopmentEnvironment) {
...
 
-    private static void initFeatures() {
-        // Init all features. Now resources (i.e I18n) are available.
-        FeatureRegistry.init();
-    }
-
     private static void parseVersion(String versionString) {
         if (isDevelopmentEnvironment()) {
-            LOGGER.info("Wynntils running on version " + versionString);
+            LOGGER.info("Wynntils running on version {}", versionString);
```
magicus: If you are updating this, why not add the modloader as well? It is useful if we ever see parts of a log. Also, the message is really unclear. "running on version ..." sounds like we are giving the java version. But I assume this is the Wynntils version..?

So maybe something like `"Wynntils version {} (using {} on Minecraft {}) is now loaded", modVersion, modLoader, mcVersion` or however it needs to be to be correct..?

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -136,4 +136,10 @@ public Object generate() {
             }
         });
     }
+
+    public enum ModLoader {
+        Forge,
```
magicus: Enum are still to be considered consts, i.e. FORGE, ... We have been very bad at enforcing this. I am surprised that Spotless does not catch this.

magicus: Unless there have been a shift in Java conventions lately? That might have happened. But to me, the UpperCamelCase is strictly reserved for classes.

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -118,6 +119,9 @@ public static void reset() {
         // tryLoadTerritories
         territories.clear();
 
+        // tryLoadMaps
```
magicus, line 122: Why this?

(Please don't mix different bug fixes in a separate PR, it makes for a messy history, messy reverts in case of failures, and all kind of confusion)

common/src/main/java/com/wynntils/mc/mixin/ResourceLoadStateTrackerMixin.java
```diff
@@ -17,6 +17,6 @@ public abstract class ResourceLoadStateTrackerMixin {
     private void onResourceManagerReloadPost(CallbackInfo info) {
         // This is the signal that Minecraft has finished loading the initial resources,
         // or a resource pack has been reloaded
-        WynntilsMod.onResourcesFinishedLoading();
```
magicus: It is possible that the old name was not enough descriptive, but the new name is really no better. This is called after all resources has been (re-)loaded. This can happen multiple time. But the fact that the resources are done loading (which is an async event) is important.

fabric/src/main/java/com/wynntils/fabric/WynntilsModFabric.java
```diff
@@ -14,12 +14,14 @@ public class WynntilsModFabric implements ClientModInitializer {
     @Override
     public void onInitializeClient() {
         Optional<ModContainer> wynntilsMod = FabricLoader.getInstance().getModContainer("wynntils");
-        if (wynntilsMod.isEmpty()) {
```
magicus, line 17: Why is this better? The old code follows our traditional style of:

if (thingsAreNotOk) return;

doRealStuff;

common/src/main/java/com/wynntils/core/WynntilsMod.java
Incompleteusern: Modloader is pretty obvious from the logs, but will do

common/src/main/java/com/wynntils/core/webapi/WebManager.java
Incompleteusern: Will do from now on

common/src/main/java/com/wynntils/core/WynntilsMod.java
Incompleteusern: This should be enforced better in general, but done

Incompleteusern:
This attempts to replace the resource hook with a hook that runs when minecraft finishes initializing

magicus:
> This attempts to replace the resource hook with a hook that runs when minecraft finishes initializing

I see that you are doing that. But why are you doing that? It is certainly not required for reporting the modloader to the server.

If you really think that needs to be done, it should be done in a separate PR. Such a change would need especially much testing and verifying in Vanilla code that it *really* guarantees that resources are loaded. The change you are removing was put in place explicitly to guarantee that resources are in place. Not only do you need to put that change in a separate PR, you'd also need to argue why that is a better way to verify that resources are loaded, than to get a signal when resources are loaded.

magicus (changes requested):
Please remove init sequence changes from this PR. And also, in the future, try to keep each PR to handling a separate concern.

Incompleteusern:
> Please remove init sequence changes from this PR. And also, in the future, try to keep each PR to handling a separate concern.

I will revert

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -99,15 +109,12 @@ public static void init(String modVersion, boolean isDevelopmentEnvironment) {
         ManagerRegistry.init();
     }
 
-    private static void initFeatures() {
```
magicus: Please restore this as well.

magicus (approved):
Looks good now. Please remember to keep separate PRs for separate issues in the future, this will make the process much smoother for everyone.

Incompleteusern:
> Looks good now. Please remember to keep separate PRs for separate issues in the future, this will make the process much smoother for everyone.

I will after this experience

PR #408, opened 2022-09-13
CustomBarsOverlayFeature: Move parsing to WynnBossBarUtils

kristofbolyai:
As requested by @magicus

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
@@ -30,15 +30,11 @@ import com.wynntils.wynn.event.ActionBarMessageUpdateEvent;
 import com.wynntils.wynn.model.ActionBarModel;
 import com.wynntils.wynn.utils.WynnBossBarUtils;
 import com.wynntils.wynn.utils.WynnUtils;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import net.minecraft.client.gui.components.LerpingBossEvent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @FeatureInfo(category = FeatureCategory.OVERLAYS)
 public class CustomBarsOverlayFeature extends UserFeature {
-    private static final Pattern BLOOD_POOL_PATTERN = Pattern.compile("§cBlood Pool §4\\[§c(\\d+)%§4\\]");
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
```diff
@@ -301,22 +294,14 @@ public ManaBankBarOverlay() {
...
-                    manaBankMaxPercent = matcher.group(2);
-                    break;
-                }
-            }
+            Pair<LerpingBossEvent, Matcher> matcherPair = WynnBossBarUtils.getLerpingBossEvent(MANA_BANK_PATTERN);
+            LerpingBossEvent bankEvent = matcherPair.a;
 
             if (bankEvent == null) return;
 
+            String manaBankPercent = matcherPair.b.group(1);
```
magicus: Can you like make a public method in WynnBossBarUtils like `getManaBank` and `getBloodPool(Percent)`? And move the patterns over there. For efficiency, let `getManaBank` return a Pair of strings. (From an API point of view, two methods returning the current and max value would have been nicest, but that'd require doubling the parsing.

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import com.wynntils.mc.utils.ComponentUtils;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.utils.Pair;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.client.gui.components.LerpingBossEvent;
+
+public class WynnBossBarUtils {
+
+    public static Pair<LerpingBossEvent, Matcher> getLerpingBossEvent(Pattern titlePattern) {
```
magicus: ... and then this can just be a private helper method.

byBackfish:
We could also just make a `registerPattern` method that accepts a pattern and a callback, and then listen to the BossHealthUpdateEvent (event that is fired once the bossbar updates) and then check for the patterns and fire the callback with the regex match

kristofbolyai:
> We could also just make a `registerPattern` method that accepts a pattern and a callback, and then listen to the BossHealthUpdateEvent (event that is fired once the bossbar updates) and then check for the patterns and fire the callback with the regex match 

I don't know where we would put it, but no need to overcomplicate it imo

byBackfish:
> I don't know where we would put it, but no need to overcomplicate it imo

Would be pretty cool to have something like that, but yea, it might be a bit "overkill"

magicus (approved):

PR #410, opened 2022-09-14
Proposal for Gradle Submodules

DevScyu:
# Proposal for Gradle Submodules

Allows better code separation so that package dependencies are stricter.

` com.wynntils.mc` can only use from `common`
`com.wynntils.wynn` can use from `common` and `mc`
`com.wynntils.features` can use from `common`, `wynn` and `mc`

`com.wynntils.wynn` should not depend on `com.wynntils.features`
`com.wynntils.mc` should not depend on `com.wynntils.wynn`

Some problem files of incorrect package dependency:
`wynntils-wynn/src/main/java/com/wynntils/wynn/utils/WynnItemUtils.java`
`wynntils-wynn/src/main/java/com/wynntils/wynn/model/LootrunModel.java`
(14 files total rely on `com.wynntils.features` in the `com.wynntils.wynn` package, **this is bad**!)
`wynntils-mc/src/main/java/com/wynntils/mc/EventFactory.java`
`wynntils-mc/src/main/java/com/wynntils/mc/mixin/GuiMixin.java`
`wynntils-mc/src/main/java/com/wynntils/mc/utils/ComponentUtils.java`
(3 files total rely on `com.wynntils.wynn` in the `com.wynntils.mc` package)


Incompleteusern:
This would be ideal, but we should first resolve all package conflicts

PR #413, opened 2022-09-17
Screens: Port legacy quest book 

kristofbolyai:
Missing: Dialogue history + Mini quests (and discovery stuff). These are due in a follow up PR. (Accessing Wynn quest book is possbile via shift+click).

I think this PR is getting big, let's start the reviews, and I can work on mini-quests and dialogue history later.

Currently ported:

Main menu: 
image
image
image

Quest Book:
image
image
image


magicus:
Heh. I started trying to do this but I could only just get the background to render, so I more or less gave up.

magicus:
@kristofbolyai FWIW, here is how far I came. You might find some inspiration there (or maybe not):
https://github.com/magicus/Artemis/tree/porting-legacy-guis

Edit: looking through this again, it just looks a mess. I followed the process of "keep copying classes from legacy until it compiles, and then start pruning", and that was obviously a bad choice.

magicus:
Actually, maybe the only thing worth using from that branch might be this snippet, where I replaced our old re-implementation of open an URL with a call into vanilla:
```
    public static void openUrl(URI uri) {
        Util.getPlatform().openUri(uri);
    }
```

kristofbolyai:
TODO: Fix class screen bug

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -80,14 +86,27 @@ public static void onWorldStateChanged(WorldStateEvent e) {
...
+                // get a chance to read the character info from the character selection menu.
+                // Instead, we send a container query to read it from the character (compass) menu.
+                WynntilsMod.info("Scheduling character info query");
+
+                // This time, we need to scan character info and profession info as well.
+                scanCharacterInfoPage();
+            } else {
+                // We did not auto-join, we have a correct ID already.
+                int oldId = currentCharacter.id;
+                scanCharacterInfoPage();
```
magicus: This means we will always do `scanCharacterInfoPage`, which is calling the container query API ("fake inventories"). I understood there were some hesitations about the stability of this, but maybe those fears have been alleviated by now..?

kristofbolyai: We need to do this so profession info is calculated. The query itself seemes to be pretty stable.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -199,13 +231,34 @@ public static CharacterInfo parseCharacterFromCharacterMenu(ItemStack itemStack,
...
         public String toString() {
-            return "CharacterInfo[classType=" + classType + ", reskinned=" + reskinned + ", level=" + level + ", id="
-                    + id + ']';
+            return "CharacterInfo{" + "classType="
+                    + classType + ", reskinned="
+                    + reskinned + ", level="
+                    + level + ", id="
+                    + id + ", professionInfo="
+                    + professionInfo + '}';
         }
```
magicus, line 258: Qodana has a point. It makes sense to have a proper toString for ProfessionInfo.

common/src/main/java/com/wynntils/features/user/WynntilsQuestBookFeature.java
```diff
@@ -0,0 +1,51 @@
...
+public class WynntilsQuestBookFeature extends UserFeature {
+    private static final String QUEST_BOOK_NAME = "§dQuest Book";
+
+    @RegisterKeyBind
+    private final KeyBind openQuestBook = new KeyBind(
+            "Open Quest Book", GLFW.GLFW_KEY_K, true, () -> McUtils.mc().setScreen(new WynntilsMenuScreen()));
+
+    @SubscribeEvent
+    public void onUseItem(UseItemEvent event) {
+        if (McUtils.player().isShiftKeyDown()) return;
```
magicus: So if you hold down shift you get to the original Quest Book? Was this the behavior in Legacy? (I did not know that...)

kristofbolyai: It wasn't. But this is not complete yet and we need some way to access the wynn book to be able to access some info.

common/src/main/java/com/wynntils/features/user/WynntilsQuestBookFeature.java
```diff
@@ -0,0 +1,51 @@
...
+
+        tryCancelQuestBookOpen(event);
+    }
+
+    private static void tryCancelQuestBookOpen(Event event) {
+        ItemStack itemInHand = McUtils.player().getItemInHand(InteractionHand.MAIN_HAND);
+
+        if (itemInHand != null
+                && ComponentUtils.getCoded(itemInHand.getHoverName()).equals(QUEST_BOOK_NAME)) {
+            event.setCanceled(true);
```
magicus, line 51: Iirc, in Legacy, we actually had two separate settings -- 1) open the Wynntils questbook by a hotkey, and 2) redirect the Wynncraft questbook to the Wynntils one. I think this is a good idea and that we can have a config for this "replacement" behavior.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+
+        this.searchWidget = new QuestBookSearchWidget(
+                (int) (Texture.QUEST_BOOK_BACKGROUND.width() / 2f + 15),
+                0,
+                Texture.QUEST_BOOK_SEARCH.width(),
+                Texture.QUEST_BOOK_SEARCH.height(),
+                this::updateQuestsFilter,
+                this);
+
+        WynntilsMod.getEventBus().register(this);
```
magicus, line 70: I looks a bit "unbalanced" to register the event bus in the constructor, but unregister in `onClose`. Maybe register in `init` instead?

magicus: In fact, why is `searchWidget` created in the constructor instead of in `init`?

kristofbolyai: Init is run on every resize. Searchwidget should not lose search into on resize. Also, we should not register this multiple times...

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+                        0,
+                        CommonColors.YELLOW,
+                        HorizontalAlignment.Center,
+                        FontRenderer.TextShadow.NORMAL);
+        poseStack.popPose();
+    }
+
+    private List<Component> getTooltipLinesForQuest(QuestInfo questInfo) {
+        List<Component> tooltipLines = new ArrayList<>() {
+            {
```
magicus: Nicely spotted, Qodana! :)

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+                        Texture.QUEST_BOOK_BACKGROUND.height() * 1.3f - 6f,
+                        0,
+                        CommonColors.YELLOW,
+                        HorizontalAlignment.Center,
+                        FontRenderer.TextShadow.NORMAL);
+        poseStack.popPose();
+    }
+
+    private List<Component> getTooltipLinesForQuest(QuestInfo questInfo) {
+        List<Component> tooltipLines = new ArrayList<>() {
```
magicus: I think this should be a public method in `QuestInfo` instead.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+                                questInfo.getLength().toString().toLowerCase(Locale.ROOT)))
+                        .withStyle(ChatFormatting.WHITE)));
+        tooltipLines.add(new TextComponent(""));
+
+        if (questInfo.getStatus() != QuestStatus.COMPLETED) {
+            tooltipLines.add(new TextComponent(questInfo.getNextTask()).withStyle(ChatFormatting.GRAY));
+            tooltipLines.add(new TextComponent(""));
+        }
+
+        if (questInfo.getStatus() != QuestStatus.CANNOT_START) {
```
magicus: ... at least up to this point. That would allow us to potentially reuse the code that generates the "actual" quest lore. The part below is more specifically tied to this Screen.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        renderBackgroundTexture(poseStack);
+
+        // Make 0, 0 the top left corner of the rendered quest book background
+        poseStack.pushPose();
+        final float translationX = getTranslationX();
+        final float translationY = getTranslationY();
+        poseStack.translate(translationX, translationY, 1f);
+
+        renderTitle(poseStack);
```
magicus: There seem to be a fair amount of overlap between this class and `WynntilsMenuScreen`. Do you think they could share a common abstract ancestor that takes care of the "questbooky" things of the screen?

common/src/main/java/com/wynntils/screens/settings/WynntilsSettingsScreen.java
```diff
@@ -51,7 +51,6 @@ public class WynntilsSettingsScreen extends Screen {
 
     public WynntilsSettingsScreen() {
         super(new TranslatableComponent("screens.wynntils.settingsScreen.name"));
-        lastScreen = McUtils.mc().screen;
```
magicus, line 54: I've said it before, but I liked the practice in Vanilla of always sending in a "parent" or "lastScreen" to a Screen constructor. If that is always used, it is easy to know where to return to, and the "chaining" of screens is very obvious.

That way, if we e.g. add a hotkey for going to settings, the Feature that implements it can call WynntilsSettingsScreen(null), and then we will exit right back with esc, which is the expected behavior in such a case, but if we got there from the wynntils main menu, we get back to it.

common/src/main/java/com/wynntils/screens/widgets/BackButton.java
```diff
@@ -0,0 +1,67 @@
...
+    public BackButton(int x, int y, int width, int height, Screen backTo) {
+        super(x, y, width, height, new TextComponent("Back Button"));
+        this.backTo = backTo;
+    }
+
+    @Override
+    public void renderButton(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        Texture backArrow = Texture.BACK_ARROW;
+        if (this.isHovered) {
+            RenderUtils.drawTexturedRect(
```
magicus, line 28: This is a very common pattern, of selecting the left or right half of a texture. I think you should make a helper function in RenderUtils for it. If you also update overlays_icon.png and settings_icon.png to have their half horizontally instead of vertically, you can use the very same function for all these icons. (The next best thing is to have two functions, like RenderUtils.drawTexturedRectHorizontalHalf() and drawTexturedRectVerticalHalf)

magicus: And it should basically only take the `Texture`, x, y, and a boolean leftHalf. 

(In fact, it might *also* be helpful to have a `drawTexturedRect` overload that takes in a Texture and extract the size and resource() automatically; this is another common pattern in this PR)

common/src/main/java/com/wynntils/screens/widgets/ReloadQuestsButton.java
```diff
@@ -0,0 +1,67 @@
...
+                    reloadButton.width() / 2f,
+                    reloadButton.height(),
+                    reloadButton.width(),
+                    reloadButton.height());
+        }
+    }
+
+    @Override
+    public void onPress() {
+        QuestBookManager.rescanQuestBook();
```
magicus, line 62: Do we also need to signal to the GUI to reload, or will that happen automatically? Hm, yes, maybe the Screen listens to the QuestBookRescanned event, or whatever it is called...

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
```diff
@@ -40,34 +44,62 @@ public TextInputBoxWidget(
...
+                        0,
+                        CommonColors.WHITE,
+                        HorizontalAlignment.Left,
+                        FontRenderer.TextShadow.NORMAL);
+
+        poseStack.popPose();
+    }
+
+    protected void checkForHeldBackspace() {
+        if (KeyboardUtils.isKeyDown(GLFW.GLFW_KEY_BACKSPACE)) {
```
magicus, line 88: this is to get repeating backspace? Why are not all characters repeating? How do Vanilla do this? Is there some functionality they have that we could reuse? (You don't necessarily have to do that in this PR)

common/src/main/java/com/wynntils/utils/Utils.java
```diff
@@ -4,16 +4,73 @@
...
+    /**
+     * Open the specified URL in the user's browser if possible, otherwise copy it to the clipboard
+     * and send it to chat.
+     * @param url The url to open
+     */
+    public static void openUrl(String url) {
+        Util.getPlatform().openUri(url);
+    }
+
+    public static void copyToClipboard(String s) {
```
magicus: The clipboard functions are not used anymore now with the new openUrl, but I assume you think they could be good to have for future uses..?

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -125,10 +147,14 @@ public static final class CharacterInfo {
         // This was implemented the same way by legacy.
         private final int id;
 
-        private CharacterInfo(ClassType classType, boolean reskinned, int level, int id) {
+        private final ProfessionInfo professionInfo;
+
+        private CharacterInfo(
+                ClassType classType, boolean reskinned, int level, int id, ProfessionInfo professionInfo) {
```
magicus, line 149: This makes me realize that we do not update our CharacterInfo, iirc. So any level up, either in combat level or professions, will not be recorded, and thus the quest book requirements will be wrong. This needs to be fixed, but perhaps not in this PR.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final int NEXT_PAGE_SLOT = 8;
+
+    private static List<QuestInfo> quests = List.of();
+    private static List<QuestInfo> newQuests;
+
+    public static void init() {}
+
+    @SubscribeEvent
+    public static void onQuestsReloaded(QuestBookReloadedEvent event) {
+        McUtils.player().sendMessage(new TextComponent("Quest Book scanned.").withStyle(ChatFormatting.GRAY), null);
```
magicus: Do we really need this? I know Legacy did, but that was since it was slow and clumsy, and not always done at the right moment. Now we rescan everytime we open the quest book.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @SubscribeEvent
```
magicus: You might want to lower priority, so the quest book rescanning is always scheduled after the player info scan.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: I am not sure, it can indicate that something is wrong.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: This also causes a bug. (Scripted container queries dont have a queue and overlap each other..) I think we do not need to do this anymore, tracking does not depend on quest book anymore.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    public static void trackQuest(QuestInfo questInfo) {
+        ScriptedContainerQuery.QueryBuilder queryBuilder = ScriptedContainerQuery.builder("Quest Book Quest Pin Query")
+                .onError(msg -> WynntilsMod.warn("Problem pinning quest in Quest Book: " + msg))
+                .useItemInHotbar(InventoryUtils.QUEST_BOOK_SLOT_NUM)
+                .matchTitle(getQuestBookTitle(1));
+
+        if (questInfo.getPageNumber() > 1) {
```
magicus, line 93: If you start saving pageNumber for a quest, maybe you could save the slotNumber for that quest as well, so you don't have to search for it when found the right page?

But, are we sure this is realiable enough? Quests are reordered in the Wynncraft quest book when they are finished. In the extreme case, a quest gets finished while you have the quest book screen opened. (Don't know how, but this can perhaps happen if you need to get some item as last requirement, and it is picked up, or if you enter a NPC dialogue, stops progressing at the last stage, enters the wyntils book, and after a while the NPC dialog times out and the quest gets finished).

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
```diff
@@ -122,6 +140,14 @@ private static String getQuestName(ItemStack item) {
         return m.group(1);
     }
 
+    private static boolean isQuestTracked(ItemStack item) {
+        String rawName = item.getHoverName().getString();
```
magicus, line 195: I'm thinking maybe QUEST_NAME_MATCHER from QuestBookManager could move to this class, and be used here? I don't really like when we have multiple ways of doing the same thing.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestStatus.java
```diff
@@ -5,12 +5,25 @@
...
-    CANNOT_START;
+    COMPLETED(new TextComponent("Completed!").withStyle(ChatFormatting.GREEN)),
+    STARTED(new TextComponent("Started...").withStyle(ChatFormatting.YELLOW)),
+    CAN_START(new TextComponent("Can start...").withStyle(ChatFormatting.YELLOW)),
+    CANNOT_START(new TextComponent("Cannot start...").withStyle(ChatFormatting.RED));
+
+    private final Component component;
+
+    QuestStatus(Component component) {
+        this.component = component;
```
magicus: This took me a while to understand. The `component` is how we describe the status in our own hover, not how Wynncraft sends data and we parse it from the string. Maybe add a comment or name the variable slightly different to signal this?

magicus (commented):
Overall I think this looks very impressive!

I spend almost 1,5h I did not really have to review this. :-/ But I realize you've spent countless more to implement this and I'm very grateful. We soo need this.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: I could have sworn I added a queue to the scripted container queries, just since they did have a chance to overlap and collide? If I did not do that, it is definitely needed.

common/src/main/java/com/wynntils/features/user/WynntilsQuestBookFeature.java
magicus: I do like it, though... Please keep it for the future as well (perhaps with a setting to turn it off)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: I don't think there is a queue (as there is also no way for a query to signal completion.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
magicus: Aha.

magicus: Maybe add what you just said as a comment! :)

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+                continue;
+            }
+
+            tracked = quest;
+            return;
+        }
+    }
+
+    // FIXME: We only need this hack to stop the screen from closing when tracking Quest.
+    //        Adding a proper way to add quests with scripted container queries would mean this can get removed.
```
magicus, line 209: After reading through all the code, I don't get this. You are tracking using scripted container queries..? Is this from an older iteration?

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -0,0 +1,513 @@
...
+    public void setFocusedTextInput(TextInputBoxWidget focusedTextInput) {}
+
+    public void setQuests(List<QuestInfo> quests) {
+        this.quests = quests;
+        this.maxPage = (quests.size() / QUESTS_PER_PAGE + (quests.size() % QUESTS_PER_PAGE != 0 ? 1 : 0)) - 1;
+        this.setCurrentPage(0);
+    }
+
+    public void setTracked(QuestInfo tracked) {
+        this.tracked = tracked;
```
magicus, line 390: I'm apparently missing something in the flow. I assume the QuestBookManager needs to be told to track the quest as well, but I fail to find where that is happening.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
kristofbolyai: I use a modified version of quest analysis query for tracking a quest. But clicking on a quest to track force-closes the container, "breaking" the query. In reality, it sends a warning, but works.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
magicus: I found it, in `mouseClicked`.

common/src/main/java/com/wynntils/screens/widgets/QuestButton.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    // Not called
+    @Override
+    public void onPress() {}
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (button == GLFW.GLFW_MOUSE_BUTTON_LEFT) {
+            if (this.questInfo.getStatus() != QuestStatus.CANNOT_START) {
```
magicus: Maybe break out this (and the "open wiki") part to separate methods, so `mouseClicked` is just a small dispatcher.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: Yes, I understand why. I was just pointing out that when I started doing this (for auto-join) I met with opposition for doing it always. (I think we should do it always, like you do now.)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: Then maybe send the error message in case of failure to the chat instead of the log?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus:
```
    public static void runQuery(ContainerQueryStep firstStep) {
        if (currentStep != null) {
            // Only add if it is not already enqueued
            if (queuedQueries.stream()
                    .filter(query -> query.getName().equals(firstStepName))
                    .findAny()
                    .isEmpty()) {
                queuedQueries.add(firstStep);
            }
            return;
        }
```

magicus: So yes, there is a queue, and yes, the container query manager knows when it is done (that is when the query do not return a "next step").

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
magicus: Hm, I might need to help you craft the query, since that should work. You just have to make sure you tell the query maanger that you expect the last click to close the container. I explicitly designed it to handle this very use case, even though I never quite got around to doing the `/quest track` that I was planning. (see https://github.com/magicus/Artemis/tree/quest-book-stuff)

magicus: Basically, my thinking was that I should get all the QuestBookManager stuff in place using commands, so when writing the GUI we'd only need to consider the UI stuff. I never got further than this when I ran out of time to spend on Wynntils, though. :-(

https://github.com/Wynntils/Artemis/compare/main...magicus:Artemis:quest-book-stuff?expand=1

kristofbolyai: Yeah, I would appreciate if you could help me with this part.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: This case, could you help out with constructing a query which does not need the page number saved? Since this is the second query, perhaps you could do it in a follow up PR (and merge this "buggy")? What do you think?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: Nice, though, I will still remove this since this is not needed anymore.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
kristofbolyai: https://github.com/Wynntils/Artemis/issues/416

common/src/main/java/com/wynntils/utils/Utils.java
kristofbolyai: I will just remove this for now. No dead code :\)

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
kristofbolyai: I think legacy did `Keyboard.enableRepeatEvents()`. I cannot find that option atm, this should be good enough.

common/src/main/java/com/wynntils/screens/widgets/ReloadQuestsButton.java
kristofbolyai: Screen listens to the event.

common/src/main/java/com/wynntils/screens/widgets/BackButton.java
kristofbolyai: I would leave this as a follow-up (RenderUtils refactor) PR. Are you fine with that?

common/src/main/java/com/wynntils/screens/settings/WynntilsSettingsScreen.java
kristofbolyai: Yeah, but this can also cause bugs (looping between 2 screens). I think it is safe to assume we don't open the settings screen from anywhere else for now.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
magicus: I’ll see what I can do. If I do not have time to do anything right now, push this as it is and I’ll rewrite it afterwards and remove the workarounds.

kristofbolyai:
@magicus note that I have not missed your reviews, I just did not have time to do a few of them. I requested review so you can review everything else in the meantime :).

kristofbolyai:
Should be everything.

common/src/main/java/com/wynntils/screens/widgets/BackButton.java
magicus: Yes.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: I'm not sure I understood why this is not needed anymore. :) I'll see if I can figure it out and come back and answer my own question. Otherwise feel free to enlighten me. :)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: We can certainly start with this query and I can try to improve it later.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: We do not need to analyse quests on world join. Legacy needed the quest book always tracked so quest overlay worked. We only need quest book tracking when the custom quest book GUI is open. So, we don't need this here, and we can query the quest book when we open the GUI.

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
magicus: I looked around the Vanilla code a bit now. There is a net.minecraft.client.gui.components.EditBox. It seems to be quite full-fledged. Why are we not using that?

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: Thanks for opening a bug. I was thinking it was needed but at the moment it was just too much work for me :)

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: I think I am following now, yes. :) Wynncraft quest tracking is reset each time you leave a world, right?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -6,53 +6,41 @@
...
-    }
-
     /**
      * Trigger a rescan of the quest book. When the rescan is done, a QuestBookReloadedEvent will
      * be sent. The available quests are then available using getQuests.
      */
     private static void queryQuestBook() {
         ScriptedContainerQuery.QueryBuilder queryBuilder = ScriptedContainerQuery.builder("Quest Book Query")
-                .onError(msg -> WynntilsMod.warn("Problem querying Quest Book: " + msg))
+                .onError(msg -> {
```
magicus, line 37: There is also an .onError in trackQuest that could use the same code of doing both log and chat. (Perhaps we should have a convenience method for that in QueryBuilder?)

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
kristofbolyai: Ah. I see. I can try to use that for every text box. Another PR though?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: What's the difference?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
```diff
@@ -95,6 +103,50 @@ public String toString() {
...
+                            .withStyle(ChatFormatting.WHITE))));
+        }
+
+        tooltipLines.add(new TextComponent("-")
+                .withStyle(ChatFormatting.GREEN)
+                .append(new TextComponent(" Length: ").withStyle(ChatFormatting.GRAY))
+                .append(new TextComponent(StringUtils.capitalizeFirst(
+                                questInfo.getLength().toString().toLowerCase(Locale.ROOT)))
+                        .withStyle(ChatFormatting.WHITE)));
+        tooltipLines.add(new TextComponent(""));
```
magicus: Maybe this is nitpicking, but I think it would be nicer if this part ended without the trailing empty line, and that this instead was added as the first thing by the questbook screen gui.

common/src/main/java/com/wynntils/screens/widgets/QuestButton.java
```diff
@@ -0,0 +1,147 @@
...
```
magicus: I was perhaps not entirely clear. I think also the left button stuff should move to a separate method.

kristofbolyai: I forgot this..

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: Sorry if I was not really clear. We have two instances of the queryBuilder in this class. I think both of them might benefit from having an error handler that prints error message both to the log and to the chat. (Basically copying this code to `trackQuest` as well)

Then I thought copying code is bad, and having an error handler sending the same message both to the log and the chat might be a common scenario for when a scripted container query fails, so maybe we should have a special support method for that, like `.onErrorReport(String msg)` that does this. But that was more of an idea for future enhancements on the container query api.

magicus:
I have looked through your additional commits, and it looks good. I'd still like to go through the code once more in its entirety, and am not sure I have time enough to do it tonight (still waiting for a late work zoom meeting with US to start...).

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: I see. Yes, this is a nice idea. (but if we print similar messages for different queries, we cannot easily tell what happened)

common/src/main/java/com/wynntils/screens/widgets/TextInputBoxWidget.java
magicus: Another PR.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -80,14 +86,27 @@ public static void onWorldStateChanged(WorldStateEvent e) {
...
+                // Instead, we send a container query to read it from the character (compass) menu.
+                WynntilsMod.info("Scheduling character info query");
+
+                // This time, we need to scan character info and profession info as well.
+                scanCharacterInfoPage();
+            } else {
+                // We did not auto-join, we have a correct ID already.
+                int oldId = currentCharacter.id;
+                scanCharacterInfoPage();
+                currentCharacter = new CharacterInfo(
```
magicus: Actually, is this correct? scanCharacterInfoPage() is only scheduling an async scan; you can not trust that any values has been updated when it returns. And in it we call:
`                    currentCharacter =
                            CharacterInfo.parseCharacterFromCharacterMenu(characterInfoItem, professionInfoItem, -1);`
so we will set the currentCharacter there anyway, and most likely overwriting what you are setting here.

kristofbolyai: Yeah, this is bad. How do we fix it though, in a sensible way?

magicus: I think what you want to do is to make `scanCharacterInfoPage` take in an `int id` which is either -1 or oldId, and use it instead.

And the entire if statement boils down to wether id should be set to -1 (we don't know), or currentCharacter.id (as in we do know).

magicus: And maybe in the future rewrite this even more so we can dynamically update the CharacterInfo. But that's another PR.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: So, my idea is that we take in a specific "prefix" message when calling it, and then append the error message. Like here `"Problem querying Quest Book: " + msg`, so we'd call this as `.onErrorReport("Problem querying Quest Book")`

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
magicus: You resolved this without doing anything? Did you disagree or just forgot? 

To clarify: my point is that I don't like that we both have a pattern and a `rawName.endsWith("§e[Tracked]")` that tries to do more or less the same thing, with the logic spread between two classes.

magicus:
Meeting was canceled. :) I looked through the code. Basically looks fine now, I just added a last few remarks.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
kristofbolyai: I removed the duplicate pattern, I thought this was that suggestion. I will look into this tomorrow.

magicus: I might start to confuse things and forget exactly what I requested as well. :) But in general I am wary of "almost-duplicated" matches (this was the cause of much grief in legacy), and this looked like one such instance.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: I think that now you can remove the `currentCharacter = new CharacterInfo(...)` from here, and just rely on the assignment in scanCharacterInfoPage.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestInfo.java
kristofbolyai: Yeah, but, how would I remove this duplication? Perhaps use the pattern here as well, and check if there is the tracked part. Since we use a lot of static methods for parsing, there is no simple way

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
magicus: @kristofbolyai Not sure if this comment got lost in the torrent of other comments, but if you remove this reassignment, I think you're all good to go.

kristofbolyai: I saw this, but I already turned off my PC yesterday. Doing this today.

magicus (approved):
:shipit:

magicus:
🍾

PR #418, opened 2022-09-21
Refactor Custombars

Incompleteusern:
This pr does the following
- Code deduplication through the usage (`BaseBarOverlay` is the interface)
- Converts rendered text to use icons, which is more in line with legacy
  - Awakened: ۞
  - Focus (not implemented): ➶ (not in legacy)
  - Arcanist Bank: ☄
  - Blood Pool: ⚕ 

TODO Images:
- Focus Image
- Me dying to Qira


magicus:
I'm not sure I understand the "Why?" here... That is, what is your goal and reason for doing this?

It looks like you are removing all support for the actionbar meters (health and mana). That is certainly not OK.

Incompleteusern:
> I'm not sure I understand the "Why?" here... That is, what is your goal and reason for doing this?
> 
> It looks like you are removing all support for the actionbar meters (health and mana). That is certainly not OK.

The reason for doing this is to make it easier to create new bars in the future, as well as removing the code duplication and standardizing the code.

Actionbar meters are still supported, and I will post an image when I have finished testing everything. 

I apologize that the .diff is mangled, and it might be best to read the file in its own context.

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
```diff
@@ -7,38 +7,45 @@
...
-        for (LerpingBossEvent event : McUtils.mc().gui.getBossOverlay().events.values()) {
+        Collection<LerpingBossEvent> events =
+                McUtils.mc().gui.getBossOverlay().events.values();
+
+        for (LerpingBossEvent event : events) {
             matcher = titlePattern.matcher(ComponentUtils.getCoded(event.getName()));
             if (matcher.matches()) {
                 poolEvent = event;
+                if (cancel) {
+                    events.remove(event);
```
kristofbolyai: Well, this is kinda weird. But this sure does its job and is not that intrusive as my solution with redirect.

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
```diff
@@ -7,38 +7,45 @@
...
+    public static final BarProgress NO_AWAKENED_PROGRESS = new BarProgress(0, 0, -1);
+    public static final BarProgress NO_FOCUS = new BarProgress(0, 0, -1);
 
     private static final Pattern BLOOD_POOL_PATTERN = Pattern.compile("§cBlood Pool §4\\[§c(\\d+)%§4\\]");
     private static final Pattern MANA_BANK_PATTERN = Pattern.compile("§bMana Bank §3\\[(\\d+)/(\\d+)§3\\]");
     private static final Pattern AWAKENED_PROGRESS_PATTERN = Pattern.compile("§fAwakening §7\\[§f(\\d+)/(\\d+)§7]");
     private static final Pattern FOCUS_PATTERN = Pattern.compile("§eFocus §6\\[§e(\\d+)/(\\d+)§6]");
 
-    private static Pair<LerpingBossEvent, Matcher> getLerpingBossEvent(Pattern titlePattern) {
+    private static Pair<LerpingBossEvent, Matcher> getLerpingBossEvent(Pattern titlePattern, boolean cancel) {
```
kristofbolyai: There might be a better name for `cancel`. Perhaps `disableVanillaRendering`? (or anything else that fits better)

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
```diff
@@ -294,178 +309,124 @@ public FocusBarOverlay() {
...
-        public CustomColor textColor = CommonColors.LIGHT_BLUE;
-
-        public ManaBarOverlay() {
+    public static class FocusBarOverlay extends BaseBarOverlay {
+        public FocusBarOverlay() {
             super(
                     new OverlayPosition(
                             -30,
-                            52,
+                            -150,
```
kristofbolyai, line 382: Why?

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
```diff
@@ -211,78 +222,82 @@ public AwakenedProgressBarOverlay() {
...
-            if (!WynnUtils.onWorld()) return;
+        public String icon() {
+            return "⚕";
+        }
 
-            WynnBossBarUtils.AwakenedBar awakenedBar = WynnBossBarUtils.getAwakenedBar();
-            if (awakenedBar.equals(WynnBossBarUtils.NO_AWAKENED_PROGRESS)) return;
+        @Override
+        public WynnBossBarUtils.BarProgress progress() {
+            return WynnBossBarUtils.getBloodPool(!INSTANCE.shouldDisplayOnBossBar);
```
kristofbolyai: `CustomBarsOverlayFeature.this` instead of `INSTANCE`?

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
kristofbolyai: I notice you changed positions for other overlays as well. What does this mean? Where are they now?

Incompleteusern: This is because `BaseBarOverlay` was added, which then shifts all the classes in the .diff downward. In reality, there shouldn't be any changes to the positions.

kristofbolyai: Oh, GH is so weird..

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
Incompleteusern: This is an unhappy side effect that the overlay has to reference the config statically. I am glad for an alternative.

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
Incompleteusern: That would likely be better

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
magicus: I'm not really following where and how this is happening in time. So the boss bar gets data gets updated in vanilla, but then this method is called, and it removes the boss bar from the set of boss bars? But whenever Vanilla gets a new boss bar update packet, it will recreate the data? What if Vanilla happens to render the bars before this method is called? Will the boss bar flicker in and out?

I agree that this is less intrusive, but I must feel convinced it is stable and reliable.

magicus:
Yes, the github diff view had particular problems presenting this change in a good way. I think I understand a bit better what you are doing now.

magicus:
Do we have some "prior art" for moving overlays to separate files? I think this Feature class is started to get quite crowded. It would be good if `BaseBarOverlay` could move to a separate file.

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
Incompleteusern: I have yet to fully test this yet. I hope it does not flicker serendipidiously :-), but if it does, I may have to resort to more intrusive methods

Incompleteusern: EDIT: This does not work sadly

Incompleteusern:
This should basically be ready for review, barring images which I have not had time to add yet.

kristofbolyai (approved):
Seems good to me

PR #419, opened 2022-09-21
Screens: Add WynntilsDialogueHistoryScreen

kristofbolyai:
@magicus Some changes were needed for dialogue parsing to work with container queries. I've sent some details in discord, but if it is not clear why the change was needed, let me know.

Screenshots:

image
image
image
image
image
image


common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,302 @@
...
+    protected void init() {
+        QuestBookManager.scanDialogueHistory();
+
+        this.addRenderableWidget(new BackButton(
+                (int) ((Texture.QUEST_BOOK_BACKGROUND.width() / 2f - 16) / 2f),
+                65,
+                Texture.BACK_ARROW.width() / 2,
+                Texture.BACK_ARROW.height(),
+                new WynntilsMenuScreen()));
+
```
Incompleteusern: Space here is a little weird

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        renderBackgroundTexture(poseStack);
+
+        // Make 0, 0 the top left corner of the rendered quest book background
+        poseStack.pushPose();
+        final float translationX = getTranslationX();
+        final float translationY = getTranslationY();
```
Incompleteusern, line 120: Can be inlined

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,302 @@
...
+                        VerticalAlignment.Middle);
+    }
+
+    @Override
+    public boolean mouseScrolled(double mouseX, double mouseY, double delta) {
+        if (delta < 0) {
+            setCurrentPage(getCurrentPage() + 1);
+        } else if (delta > 0) {
+            setCurrentPage(getCurrentPage() - 1);
+        }
```
Incompleteusern: Perhaps could be replaced with something like
```java
setCurrentPage(getCurrentPage() + (delta > 0 ? 1 : -1));
```

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,302 @@
...
+import net.minecraft.client.gui.components.AbstractButton;
+import net.minecraft.client.gui.components.Widget;
+import net.minecraft.client.gui.components.events.GuiEventListener;
+import net.minecraft.client.resources.language.I18n;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.network.chat.TranslatableComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class WynntilsDialogueHistoryScreen extends WynntilsMenuPagedScreenBase {
```
Incompleteusern, line 36: Nice to see something similar to `QuestBookListPage` here

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,302 @@
...
+                            .withStyle(ChatFormatting.GREEN));
+
+            RenderUtils.drawTooltipAt(
+                    poseStack,
+                    mouseX,
+                    mouseY,
+                    100,
+                    tooltipLines,
+                    FontRenderer.getInstance().getFont(),
+                    true);
```
Incompleteusern, line 217: Perhaps extract this from the conditionals and instead set tooltipLines in the if statements

common/src/main/java/com/wynntils/screens/WynntilsMenuScreen.java
```diff
@@ -207,6 +226,27 @@ private void renderButtons(PoseStack poseStack, int mouseX, int mouseY) {
 
             if (hovered) {
                 this.hovered = button;
+            }
+
+            if (!button.dynamicTexture()) {
```
Incompleteusern, line 231: What is this check for?

common/src/main/java/com/wynntils/screens/WynntilsMenuScreenBase.java
```diff
@@ -91,4 +91,12 @@ protected void renderDescription(PoseStack poseStack, String description) {
                         HorizontalAlignment.Left,
                         FontRenderer.TextShadow.NONE);
     }
+
+    protected float getTranslationY() {
+        return (this.height - Texture.QUEST_BOOK_BACKGROUND.height()) / 2f;
+    }
```
Incompleteusern, line 97: `Translation` being used clashes with the already present usage of translating text into languages. Perhaps `translated` might work?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -123,6 +139,84 @@ private static void findQuestForTracking(ContainerContent container, QuestInfo q
...
+            int page = i;
+            queryBuilder
+                    .clickOnSlot(0)
+                    .matchTitle(getQuestBookTitle(1))
+                    .setWaitForMenuReopen(false)
+                    .processContainer((c) -> {
+                        ItemStack dialogueHistoryItem = c.items().get(0);
+
+                        if (!ComponentUtils.getCoded(dialogueHistoryItem.getHoverName())
+                                .equals("§bDialogue History")) return;
```
Incompleteusern, line 201: This check is only necessary once in `scanDialogueHistory` technically, so check could be reasonably removed. It is fine to keep, however.

Incompleteusern (dismissed):
A few minor comments

common/src/main/java/com/wynntils/screens/WynntilsMenuScreen.java
kristofbolyai: Some textures have 2 states, and they "animate" when hovered. Some are stateless.

common/src/main/java/com/wynntils/screens/WynntilsMenuScreenBase.java
kristofbolyai: Perhaps, although I think it is pretty clear what this does.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: Yeah, it is just safer to do so.

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
kristofbolyai: I did not do this because there might be some special tooltip rendering where it is not always the same call. I will clean this up next commit / when I add mini-quest parsing.

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
kristofbolyai: Yeah, it is really basic though, but I like that about this. If we do item guides, it will be easy.

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
kristofbolyai: I did not want to, on purpose.

common/src/main/java/com/wynntils/wynn/model/CharacterManager.java
```diff
@@ -99,6 +99,7 @@ public static void onWorldStateChanged(WorldStateEvent e) {
             } else {
                 // We did not auto-join, we have a correct ID already.
                 int oldId = currentCharacter.id;
+                System.out.println("oldId = " + oldId);
```
magicus: Left over debug code?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -80,7 +96,7 @@ private static void processQuestBookPage(ContainerContent container, int page) {
         if (page == 4) {
             // Last page finished
             quests = newQuests;
-            WynntilsMod.postEvent(new QuestBookReloadedEvent());
+            WynntilsMod.getEventBus().post(new QuestBookReloadedEvent.QuestsReloaded());
```
magicus: This seem like a merge error..? I recently turned `WynntilsMod.getEventBus().post` into `WynntilsMod.postEvent` but this seem to revert that.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -123,6 +139,84 @@ private static void findQuestForTracking(ContainerContent container, QuestInfo q
...
+
+                    if (!ComponentUtils.getCoded(dialogueHistoryItem.getHoverName())
+                            .equals("§bDialogue History")) return;
+
+                    for (String line : ItemUtils.getLore(dialogueHistoryItem)) {
+                        Matcher matcher = DIALOGUE_HISTORY_PAGE_PATTERN.matcher(line);
+
+                        if (matcher.matches()) {
+                            int pageCount = Integer.parseInt(matcher.group(2));
+                            createActualDialogueHistoryQuery(pageCount);
```
magicus, line 159: The scripted queries are really clumsy for this use case. I was planning to make it easier to create dynamic queries, but it turned out to be a hard problem defining an API with good usability to do dynamic querying, so I kind of got stuck. But this code might perhaps serve as a good example, so if I can do something that makes it better (so that we don't have to first check the number of pages, and *then* redo everything to iterate over them), I think we'd have a much better solution. 

For the time being though, this is a creative (and good-enough) solution.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -6,28 +6,44 @@
...
     private static List<QuestInfo> quests = List.of();
     private static List<QuestInfo> newQuests;
+    private static List<List<String>> dialogueHistory = List.of();
+    private static List<List<String>> newDialogueHistory;
 
     public static void init() {}
 
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public static void onWorldStateChanged(WorldStateEvent e) {
+        quests = List.of();
```
magicus, line 43: This was a bug that we did not clear quests when changing worlds? :-o

kristofbolyai: Yeah, but we re-queried on screen init so it was okay.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: Yeah I will fix this

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
```diff
@@ -143,8 +143,10 @@ public static void onContainerSetContent(ContainerSetContentEvent e) {
             }
         } else {
             // We're done
+            ContainerQueryStep oldStep = currentStep;
```
magicus: Does `endQuery()` reset `currentStep`? Something in "oldStep" leads my thinking astray and got me confused. Maybe you can call it "lastStep" eller "finalStep"?

kristofbolyai:
@magicus as you can see, I added an onComplete handler for container queries, but did not end up using it. I think it is fine to leave this in as it has use cases, likely. What do you think?

common/src/main/java/com/wynntils/wynn/model/container/ContainerQueryManager.java
kristofbolyai: It does reset, yes.

magicus:
`onComplete` is fine, yes. It's a fine line between having dead code and designing for the future. :-) I think we both agree that in general there should be no dead code, but some exceptions like this is probably fine.

common/src/main/java/com/wynntils/screens/WynntilsDialogueHistoryScreen.java
```diff
@@ -0,0 +1,297 @@
...
+                    mouseX,
+                    mouseY,
+                    100,
+                    RELOAD_TOOLTIP,
+                    FontRenderer.getInstance().getFont(),
+                    true);
+            return;
+        }
+
+        if (this.hovered instanceof QuestsPageButton) {
```
magicus, line 196: It seems like Qodana is correct. `hovered` is private, initialized to null, and I can't see any setters.

kristofbolyai: Line 232?

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -183,17 +188,13 @@ public boolean charTyped(char codePoint, int modifiers) {
 
     @Override
     public boolean mouseScrolled(double mouseX, double mouseY, double delta) {
-        if (delta < 0) {
-            setCurrentPage(getCurrentPage() + 1);
-        } else if (delta > 0) {
-            setCurrentPage(getCurrentPage() - 1);
-        }
+        setCurrentPage(getCurrentPage() + (delta > 0 ? -1 : 1));
```
magicus, line 191: Just a reflection. I noticed aBackfish's PR on legacy, where there was a ScrollDirection.UP/DOWN selection. Maybe we should consider having a general function that converts a mouse scrolled delta to a direction, and that can be affected by settings. (Is there not already an inverted scroll setting in Vanilla, btw?)

Not saying it should be part of this PR, just ranting a bit. :)

kristofbolyai: Yeah, if there is not one, we could have one.

common/src/main/java/com/wynntils/screens/WynntilsQuestBookScreen.java
```diff
@@ -278,21 +279,31 @@ private void renderTooltip(PoseStack poseStack, int mouseX, int mouseY) {
...
-                        poseStack,
-                        (currentPage + 1) + " / " + (maxPage + 1),
-                        Texture.QUEST_BOOK_BACKGROUND.width() / 2f,
-                        Texture.QUEST_BOOK_BACKGROUND.width(),
-                        Texture.QUEST_BOOK_BACKGROUND.height() - 25,
-                        0,
-                        CommonColors.BLACK,
-                        HorizontalAlignment.Center,
-                        FontRenderer.TextShadow.NONE);
+        if (this.hovered instanceof DialogueHistoryButton) {
```
magicus, line 285: I'm not sure why Qodana is complaining about this here but not in `renderTooltip()`. But can you please check if there is some substance behind this claim?

kristofbolyai: I have checked this and tested in-game, and it seems to work fine. I have no idea what Qodana is doing.

magicus (approved):

PR #420, opened 2022-09-21
Features: Add BeaconBeamFeature

kristofbolyai:
This PR also adds auto tracking for quest description coordinates. `/compass` will be added in a separate PR.

image


Incompleteusern:
Custom beam color would be cool, but that would require a hybrid between manager and feature, which we still need to resolve on how that should work.

Incompleteusern (dismissed):

kristofbolyai:
> Custom beam color would be cool, but that would require a hybrid between manager and feature, which we still need to resolve on how that should work.

There is a config for custom beam color?

Incompleteusern:
> > Custom beam color would be cool, but that would require a hybrid between manager and feature, which we still need to resolve on how that should work.
> 
> There is a config for custom beam color?

Oh, sorry, I got lost.

magicus:
You're on a roll! 👀

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
```diff
@@ -0,0 +1,96 @@
...
+        QuestInfo questInfo = event.getQuestInfo();
+
+        Matcher matcher = COORDINATE_PATTERN.matcher(questInfo.description());
+        if (!matcher.matches()) return;
+
+        Location parsedLocation = new Location(
+                Integer.parseInt(matcher.group(1)),
+                Integer.parseInt(matcher.group(2)),
+                Integer.parseInt(matcher.group(3)));
+
```
magicus: This functionality should be moved to QuestInfo.

kristofbolyai: (While I am there, I will rename it to `ScoreboardQuestInfo` as the other QuestInfo is for quest book stuff.

common/src/main/java/com/wynntils/wynn/model/scoreboard/quests/QuestHandler.java
```diff
@@ -53,5 +55,6 @@ public void onSegmentRemove(Segment segment, ScoreboardModel.SegmentType segment
 
     public static void resetCurrentQuest() {
         currentQuest = null;
```
magicus: Maybe move extract both these changes to a common private `setQuest` helper? To help make sure we send the event properly if we ever will set it from new code.

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
```diff
@@ -0,0 +1,96 @@
...
+        poseStack.pushPose();
+        float partials = event.getPartialTick();
+        Vec3 vec = McUtils.mc().getCameraEntity().getEyePosition(partials);
+        poseStack.translate(-vec.x, -vec.y, -vec.z);
+
+        Location location = CompassModel.getCompassLocation().get();
+        poseStack.translate(location.x, location.y, location.z);
+
+        BeaconRenderer.renderBeaconBeam(
```
magicus, line 70: Aha, there is a vanilla function we can hijack. :) I was always wondering how much effort must have gone into the compass bean in Legacy :)

kristofbolyai: It actually took me 2 months to get this to render even with this 😅

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
```diff
@@ -0,0 +1,96 @@
...
+                Integer.parseInt(matcher.group(1)),
+                Integer.parseInt(matcher.group(2)),
+                Integer.parseInt(matcher.group(3)));
+
+        CompassModel.setCompassLocation(parsedLocation);
+    }
+
+    @SubscribeEvent
+    public void onRenderLevelLast(RenderLevelLastEvent event) {
+        if (McUtils.player() == null || CompassModel.getCompassLocation().isEmpty()) return;
```
magicus: We're not testing for null on McUtils.player() elsewhere (except in WebManager where I assume it is copied over from Legacy).

I can't imagine what could cause player to be null, as long as we are playing on a world. That just seems like everything should come crashing down burning. And we already have our event filtering so events are only sent out if we are really connected to Wynncraft.

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
magicus: Oh my! 😱

magicus (approved):

PR #422, opened 2022-09-21
Overlay crash protection

magicus:
If an overlay crashes during rendering, log it, give user feedback and disable the overlay.

Also fixes issue in NpcDialogueOverlayFeature where the dialogue might not be cleared if the overlay was disabled while showing a dialogue, as discovered by testing this new crash protection. :)

I indent to push similar crash protection to other parts of "components" that are executed by our core framework in follow-up PRs.


Incompleteusern (approved):
Nice usage of disabling the overlay

PR #425, opened 2022-09-22
ChatModel: Greatly improve parsing

### LINKED ISSUE #292: Encode Component into strings losslessly
Several parts of Wynntils (like the ChatManager, and the Translation API) would benefit from us being able to convert a Component into a text encoding, which can then losslessly be converted back to a Component. Basically, I'm thinking something like this:

`§6{Hello there, }§7{Stranger!} §0{Click on §L{http://www.wynntils.com;this link}!}`

In the best of all worlds, we could even use this encoding to do regexp matching. The current code that converts to the "classic" format codes is a one-way street, basically since we lose the "end of formatting" information that is present in the Component. While it might have it's uses (it's served us well, so far), I think it is suboptimal. But the regexp matching on strings is much easier than trying to match an actual Component. So a faithful but still readable representation of Components as a string would be a benefit for all of the mod.

Before deciding on a format, I need to look into what kind of formatting is allowed by Component. The format also needs to be regexp friendly. And trivial to parse and generate.

### LINKED ISSUE #334: NPC dialogue overlay should also catch "non-screen" dialogue
image
Multiple dialog lines during the Hero of Gavel quest are not shown in the overlay but in chat.

kristofbolyai:
Fixes #334. 

This PR implements lossless processing for chat messages, fixing the issue where clickables are not working (mainly fixing dialogues where the user needs to pick an option).

This PR also fixes the issue where non-progressable quest dialogue was not displayed on the NPC dialogue overlay. 

This does not implement #292, but fixes issues related to it in ChatModel.


magicus:
Ooooh! Exciting. :-) I'm looking forward to going through this in detail. I'm also very happy that you are helping out with the chat stuff, it's better to have shared knowledge of all parts of the code and I started feeling a bit lonely in the chat corner. :)

kristofbolyai:
> Ooooh! Exciting. :-) I'm looking forward to going through this in detail. I'm also very happy that you are helping out with the chat stuff, it's better to have shared knowledge of all parts of the code and I started feeling a bit lonely in the chat corner. :)

I edited the description. There is no lossless encoding of components to strings, but I did fix issues lossless encoding could have fixed in ChatModel. Still, I think this is a pretty good fix.

magicus:
> fixing dialogues where the user needs to pick an option

Did you find any such examples in the wild? I've not seen any since Winter Detlas. (I might have asked this before...)

kristofbolyai:
> > fixing dialogues where the user needs to pick an option
> 
> 
> 
> Did you find any such examples in the wild? I've not seen any since Winter Detlas. (I might have asked this before...)

https://wynncraft.fandom.com/wiki/The_Canary_Calls

Step 2 has both options and non-shiftable dialogue.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -110,19 +109,19 @@ public static void onChatReceived(ChatPacketReceivedEvent e) {
     }
 
     private static void handleMultilineMessage(Component message) {
-        String msg = ComponentUtils.getCoded(message);
-
-        List<String> lines = new LinkedList<>(Arrays.asList(msg.split("\\n")));
+        List<Component> lines = new LinkedList<>(message.getSiblings());
```
magicus, line 111: I'm seriously not really understanding how this can work. From what I remember, the "\n" was embedded in the middle of messages, and the "sibling" relationship segmented the message according to formatting instead. That was kind of the reason I started using formatted strings instead of Components. 

How do you ensure that `lines` contains just "lines"?

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -131,31 +130,35 @@ private static void handleMultilineMessage(Component message) {
...
+        LinkedList<Component> newChatLines = new LinkedList<>();
         LinkedList<String> dialog = new LinkedList<>();
 
-        if (NPC_FINAL_PATTERN.matcher(newLines.getFirst()).find()) {
+        String firstNewLineCoded = ComponentUtils.getCoded(newLines.getFirst());
+
+        boolean noShiftToContinueDialog = false;
+
+        if (NPC_DIALOGUE_PATTERN.matcher(firstNewLineCoded).matches()) {
+            noShiftToContinueDialog = true;
```
magicus: What do you think about `nonBlockingDialogue` as term instead?

common/src/main/java/com/wynntils/core/chat/ChatModel.java
kristofbolyai: From what I've seen, when the chat contents are resent, it is a huge, empty TextComponent, with each message being a sibling (these components might have additional siblings, but we do not care about them). As we do not split on "\n",  I needed to change a few places where we expected empty lines, etc. This seemed to work fine.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -131,31 +130,35 @@ private static void handleMultilineMessage(Component message) {
...
+            handleNpcDialog(List.of(firstNewLineCoded), false);
+        } else if (NPC_FINAL_PATTERN.matcher(firstNewLineCoded).find()) {
             // This is an NPC dialog screen.
             // First remove the "Press SHIFT to continue" trailer.
             newLines.removeFirst();
-            if (newLines.getFirst().isEmpty()) {
```
magicus, line 150: There used to be an empty line inbetween. Like:
```
xxxxx: ....

[1/3] Key Master: Are you the gatekeeper?

Press SHIFT to continue
```

We needed to remove that, both to trim down the message sent as an event, and to avoid falsely triggering the detection of the "end" (that is, the start, since we're doing this in reverse) of the NPC message.

I don't understand how that is handled if you remove this?

kristofbolyai: We remove all empty lines before this code is reached.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -243,15 +250,15 @@ private static Component handleChatLine(Component message, String codedMessage,
         return event.getMessage();
     }
 
-    private static void handleNpcDialog(List<String> dialog) {
+    private static void handleNpcDialog(List<String> dialog, boolean autoProgressing) {
```
magicus: Here, too, I think I prefer `nonBlocking`. (Or, if you want to, inverse the logic and use `blocking`)

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -100,7 +102,7 @@ public void onNpcDialogue(NpcDialogEvent e) {
             scheduledAutoProgressKeyPress = null;
         }
 
-        if (autoProgress) {
+        if (autoProgress && e.isAutoProgressing()) {
```
magicus: Shouldn't this test be inverted? If this means "NPC chat event is nonBlocking", then we should not schedule a sneak key press...

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: This might have been a change with 2.0 then. I think I'll need to verify this myself. It is indeed very welcome if this is the case. I am 99% certain it did not look like that when I created this functionality.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
kristofbolyai: Oops

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -110,19 +109,19 @@ public static void onChatReceived(ChatPacketReceivedEvent e) {
...
             // If we have no history, all lines are to be considered new
             lines.forEach(newLines::addLast);
         } else {
             // Figure out what's new since last chat message
-            for (String line : lines) {
-                String noCodes = ComponentUtils.stripFormatting(line);
+            for (Component line : lines) {
+                String noCodes = ComponentUtils.getUnformatted(line);
+                if (noCodes.isBlank()) continue;
+
```
magicus, line 124:
```suggestion
                // Remove all empty lines
                if (noCodes.isBlank()) continue;

```

kristofbolyai:
I did notice a bug with dialogue parsing. Nothing serious, I think siblings will still work. I will try to fix it soon.

kristofbolyai:
> I did notice a bug with dialogue parsing. Nothing serious, I think siblings will still work. I will try to fix it soon.

Fixed now. I also removed some empty line filtering logic, it is not needed anymore.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: Well, it is still the case that `\n` are embedded. But perhaps it does not matter:

```
TextComponent{text='
                       ', siblings=[TextComponent{text='Welcome to Wynncraft!
', siblings=[], style=Style{ color=dark_gray, bold=true, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='               ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='play.wynncraft.com', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=ClickEvent{action=OPEN_URL, value='http://play.wynncraft.com'}, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text=' ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='-/-', siblings=[], style=Style{ color=dark_gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text=' wynncraft.com

', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='                       ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='Wynncraft 2.0 is OUT!
', siblings=[], style=Style{ color=gray, bold=true, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='                     ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='Read about the new abilities
', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='                         ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='and archetypes ', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='here!', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=true, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}, TextComponent{text='

', siblings=[], style=Style{ color=gray, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}], style=Style{ color=null, bold=null, italic=null, underlined=null, strikethrough=null, obfuscated=null, clickEvent=null, hoverEvent=null, insertion=null, font=minecraft:default}}
```

Edit: I tried copy/paste from IntelliJ watch but it replaced the `\n` as shown in the watch window with real line breaks, so it just looks odd here. Every new line here is a "\n" in the text. And this is an entire sibling of the top-level TextComponent.

kristofbolyai: When the siblings have newlines, then it is some special Wynncraft message. I think it is fine to keep them as a "bunch" (for our purpose of quest dialogue parsing).

magicus: All "normal" chat history (user sent messages) seem to have one sibling per line. So these "multiline" advertising messages perhaps don't matter that they are several lines long.

But... I wonder what happens when they are starting to creep out of the history range. Can it lead us to fail to detect the "last message"?

I am seriously worried a bit that this entire change is making the chat manager less robust. If nothing else, we will now be *relying* on the assumption that there will be basically one sibling per line (or several-lines-long-messages). If this breaks for some reason, due to chat message types we have not yet spotted, the entire logic can just come crumbling down, and deliver very weird results to the user.

magicus: To feel 100% confident, I'd rather see that we transformed the TextComponent tree, so each top-level sibling contained exactly one line. For the normal chat messages, this is already the case, so we need not do anything there.

kristofbolyai: This change might make us more reliant on Wynn in a way yes. But, we used to rely on newlines being in specific places, in a specific format (always newline before "Press shift to continue", etc.). I think this state of ChatModel is almost as "breakable" as the old version.

kristofbolyai: > To feel 100% confident, I'd rather see that we transformed the TextComponent tree, so each top-level sibling contained exactly one line. For the normal chat messages, this is already the case, so we need not do anything there.

I am not sure doing this would be any better, as there is a really limited amount of messages which still have new lines, and even if they exist, they do not break our parsing system.

magicus: Some pseudo code on how I'm thinking:
```
TextComponent newComponent = new MutableComponent();
for sibling : TextComponent.getSiblings() {
  String text = sibling.getString();
  if (text.indexOf("\n") == text.length()) {
    // all is fine
   newComponent.add(sibling);
    continue
  }
  // else we need to either split this sibling up, or join it with the next, or possiblt both
 doTrickyStuff();
}
textComponent = newComponent;
```
The only really complicated thing is the `doTrickyStuff()`....

kristofbolyai: Also, technically, if we do this, this should be able to recurse on all siblings of all siblings. 

Still, I am not sure why we are trying to "optimize" the Component representation of a message (it feels like we are doing that). We could do this, but, either way, both the optimized and the normal component has the full information of a "message" (even if the said message is multi-line).

magicus: > I am not sure doing this would be any better, as there is a really limited amount of messages which still have new lines, and even if they exist, they do not break our parsing system.

Maybe. I'm not fully convinced. I'm basically worrying about two things:

1) That there might be "special" messages that break this behavior, but that we have not yet seen. This can only be answered by running for a long time and seeing that it does not break.

2) That there is some edge case with how the "last stored chat line" and the  system for how new lines are detected with these ads spanning multiple lines. I cannot say for certain that it will break. But I *can* say that when I wrote the code I made the assumption that we got lines split on line breaks, and now that is not true anymore. So the logic needs to be scrutinized. And I think we need to make some practical tests, where we try to fill the chat so let's say just haft the first welcome or advertising message is left in the chat history, and then we go talk to an NPC, and see what happens.

3) There might be some point 3 as well, but I can't put my finger on it yet...

magicus: Let me phrase things differently: I am not overly confident in the stability of this code. In contrast with e.g. the Container Query which I have a lot of faith in, this was tricky to get right, I made a lot of false starts and had to fix things when developing the code that I did not expect when I wrote it. So in my eyes it is very brittle. And I worry that this change will make it even more brittle. When I instead would like to make it more robust.

magicus: It is not about optimizing storage, it is about storing the history of the chat on a per-line basis, since minecraft buffers N lines (I think it is 32, or something like that), and that the check-backwards-for-actually-new-messages work on a per-line basis. The disconnect between actual lines as a basic type (which my original code had) and siblings as a basic type, which this PR has, makes me worried that things can go wrong in ways that is hard to predict and hard to test.

magicus: To put it bluntly:
```
        List<String> lines = new LinkedList<>(Arrays.asList(msg.split("\\n")));
```
was true, but

```
        List<Component> lines = new LinkedList<>(message.getSiblings());
```

is a lie. This is not a list of lines. It is a list of siblings. And that will have consequences. I like the idea to keep the messages as Components. But I'd *also* like to keep the idea that `lines` are, in fact, lines.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: You resolved this without fixing?

common/src/main/java/com/wynntils/features/user/TranslationFeature.java
```diff
@@ -110,8 +110,8 @@ private String wrapCoding(String origCoded) {
     }
 
     private static class TranslatedNpcDialogEvent extends NpcDialogEvent {
-        public TranslatedNpcDialogEvent(String codedDialog) {
-            super(codedDialog);
+        public TranslatedNpcDialogEvent(String codedDialog, boolean autoProgressing) {
```
magicus, line 113: Also, to be clear, I think the name "nonBlocking" (or perhaps even better if we switch it around and use "blocking" (which then will be true for the "traditional" NPC dialogues) should be propagated everywhere, into this event and where the event is read, etc.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
kristofbolyai: Ah, I renamed some other boolean with the wrong terminology.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
kristofbolyai: I agree that this might not be the most stable/bug-free code we have. However, parsing chat data will always be somewhat uncertain. However, I do think that this PR is not that big of a downgrade to our code in terms of safety/stability. 

Yes, I am a bit paranoid about merging this PR. But, I don't think we can do this much better. Basically, the logic is quite simple now. Is current like a NPC Dialog, according to the pattern, or is it not? The other code is just handling the problem of Wynn re-sending messages.

kristofbolyai: Overall, I am not even sure how the component should look like when we "squash" it.

image

What do we do with the spaces that make this centered? Do we count the first line as line 1, and the 2-3 as a line again? I think you can see what the problem is. We cannot really turn this message into 1 line, without making assumptions. And, one of our goal is to have less assumptions in the code. So we just fix a problem by creating essentially the same problem somewhere else.

magicus: I've started trying to write a method for doing this reshuffling of the TextComponent.
Basically what I need to be happy is that this line is replaced with a 
`List<Component> lines = getLinesAsSiblings(message);`

I can help trying to write `getLinesAsSiblings` if you want.

common/src/main/java/com/wynntils/features/user/TranslationFeature.java
kristofbolyai: Should be renamed everywhere.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
kristofbolyai: After writing the code, and having proper component - line relationship, do you want to put back the newline checks (is it even necessary if every quest message is its own sibling)?

kristofbolyai: Writing the code is half the problem (obviously, it is tricky). I am just not sure this is any better (as I assume it won't be worse. Although, in my head, the more modifications we do to the strings, the more bugs we will introduce).

magicus: I found another example of a message that do not follow the "one sibling per line rule":

```
[22:44:03] [Render thread/INFO] (Minecraft) [CHAT] §3[INFO§3]§b TheNelston rewarded §r§31024 Emeralds§r§b to TheNelston.
[22:44:03] [Render thread/INFO] (Minecraft) [CHAT] §r§3Rewards can be claimed in the Member Menu.
```

This was sent as a single sibling.

kristofbolyai: You mentioned your concern was that, when writing the ChatModel code, you had the assumption that the `lines` list always contains 1 chat line.

What if, we remove this assumption, and re-factor the ChatModel around this (although, I am not sure if there is anything to rewrite), instead of messing with components?

kristofbolyai: > I found another example of a message that do not follow the "one sibling per line rule":
> 
> ```
> [22:44:03] [Render thread/INFO] (Minecraft) [CHAT] §3[INFO§3]§b TheNelston rewarded §r§31024 Emeralds§r§b to TheNelston.
> [22:44:03] [Render thread/INFO] (Minecraft) [CHAT] §r§3Rewards can be claimed in the Member Menu.
> ```
> 
> This was sent as a single sibling.

Seems like relevant messages are "grouped" together. Maybe we could even use this whole sibling - parent relation to our benefit? Example: Let's say we want to filter out the ad. We check for the first line, and we can just remove it's siblings as well. Same with the guild rewards. No need for multiple regexes.

magicus: Yes, but what happens the first line scroll of the screen, but the second is still there? And you enter a NPC dialogue...

magicus: The chat system of Vanilla is still line-based. That's a hard fact that we need to relate to, and it determines how we must approach this, imho.

kristofbolyai: > Yes, but what happens the first line scroll of the screen, but the second is still there? And you enter a NPC dialogue...

I think Wynn either sends the whole component again (and mc just forgets it) or neither is sent. I don't think we confirmed this is not the case?

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -100,7 +102,7 @@ public void onNpcDialogue(NpcDialogEvent e) {
             scheduledAutoProgressKeyPress = null;
         }
 
-        if (autoProgress) {
+        if (autoProgress && !e.isBlocking()) {
```
magicus: Either I'm confused, or you are confused. Pressing "sneak" is only relevant for blocking NPC dialogues, right? The ones that are not blocking is just announced to the chat.

So the autoProgress scheme in NpcDialogueOverlayFeature is only releavant for `e.isBlocking()`. So I think this test is still backwards.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -195,8 +198,11 @@ private void renderDialogue(PoseStack poseStack, String currentDialogue) {
 
             // Render "To continue" message
             List<TextRenderTask> renderTaskList = new LinkedList<>();
-            TextRenderTask pressSneakMessage = new TextRenderTask("§cPress SNEAK to continue", renderSetting);
-            renderTaskList.add(pressSneakMessage);
+
+            if (!currentlyBlocking) {
```
magicus: And here as well. `currentlyBlocking` is set to true if we get a blocking NPC dialogue, i.e. one which waits for sneak. So if, and only if, the dialog we are showing is blocking, then we should print this text message.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
kristofbolyai: Blocking is a weird word in this scenario. I think of blocking as "this dialouge blocks the user from progressing by shift". You can think of it as "this dialogue blocks progress until shift"

kristofbolyai: perhaps this is not that good of a name after all?

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -122,6 +124,7 @@ private ScheduledFuture<?> scheduledSneakPress(String msg) {
     @SubscribeEvent
     public void onWorldStateChange(WorldStateEvent e) {
         currentDialogue = null;
+        currentlyBlocking = true;
```
magicus: You do not need to set this here. `currentlyBlocking` is only checked in `renderDialogue`, and that is only called if we have a `currentDialgue`, and that is set at the same time as you set `currentlyBlocking` in the event handler. So this is just confusing. (Especially as you set it to true)

magicus:
Another question: How do you know when to clear the non-blocking NPC dialogues from the overlay? I can't really see that this is done now, so I assume they will stay on screen until a new dialogue comes into chat? That might be annoying...

kristofbolyai:
> Another question: How do you know when to clear the non-blocking NPC dialogues from the overlay? I can't really see that this is done now, so I assume they will stay on screen until a new dialogue comes into chat? That might be annoying...

Yeah, but most of the time when this happens, there is some animation. I guess I could use the same system for calculating a display time as auto progress?

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
magicus: Ok, seems we interpret "blocking" completely different here then! Yes, to  me, it "obviously" meant that progress was blocked, waiting for shift. :)

`needsConfirmation`? Is that ok? So the "normal" NPC dialogues that require you to press shift has `needsConfirmation` true?

magicus:
Have we not had this discussion before, in discord or a bug report? I'm thinking of quests where someone is calling out lines, one by one, that explain a part of the mystery, while you traverse further and further into the maze leading to the boss battle. Like https://wynncraft.fandom.com/wiki/The_Hunger_of_Gerts_Part_2_(1.20)#Stage_12

magicus:
Trying to parkour or fight mobs with a NPC dialog plastered over your screen might be ... non-helpful....

kristofbolyai:
@magicus did the renames, and added auto progress. 

Are you willing to accept this PR as is? If not, what are the minimum changes you request? 

I think, this is fine for the time being. We can come back to this and figure out a nice solution (if there is one...).

magicus:
I am working on a `List<Component> splitComponentInLines(Component message)` method. It's almost done but there are some details left to flesh out.

But I think I can just inject that in your patch once it's pushed. I think that will alleviate my fears. 

Let me double check the patch for any other things, and then I think you're good to go.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
```diff
@@ -131,68 +130,59 @@ private static void handleMultilineMessage(Component message) {
...
-                    if (EMPTY_LINE_PATTERN.matcher(line).find()) {
-                        dialogDone = true;
-                        // Intentionally throw away this line
-                    } else {
-                        dialog.push(line);
-                    }
+            for (Component lineComponent : newLines) {
+                String line = ComponentUtils.getCoded(lineComponent);
+
+                if (NPC_DIALOGUE_PATTERN.matcher(line).find()) {
```
magicus, line 161: Ooh, I did not spot this before. You cannot do this. Not all "NPC" dialogue lines are formatted like that. (That was why I kept the blank lines before, to be able to tell when the "NPC dialogue" lines were done. 

For an example, go and talk to the Detlas citizen standing next to the entrance from Nivla/Ragni. The dialogue ends like "You walk away, wondering if the citizen stands there all day greeting visitors" or something like that.

common/src/main/java/com/wynntils/features/user/TranslationFeature.java
magicus: Should as in "I agree, I need to fix this", or "this should have been fixed"? Because if it's the latter, that does not seem to be the case, or there is something weird with what github shows.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -195,11 +211,14 @@ private void renderDialogue(PoseStack poseStack, String currentDialogue) {
...
+            if (currentlyNeedsConfirmation) {
+                TextRenderTask pressSneakMessage = new TextRenderTask("§cPress SNEAK to continue", renderSetting);
+                renderTaskList.add(pressSneakMessage);
+            }
+
+            if (scheduledAutoProgress != null && !scheduledAutoProgress.isCancelled()) {
+                long timeUntilProgress = scheduledAutoProgress.getDelay(TimeUnit.MILLISECONDS);
                 TextRenderTask autoProgressMessage = new TextRenderTask(
                         ChatFormatting.GREEN + "Auto-progress: "
                                 + Math.max(0, Math.round(timeUntilProgress / 1000f))
```
magicus, line 224: I feel like I'm being a real PITA tonight, but this still does not seem right. :-)

Overall I think this latest iteration makes most sense, there are two "kinds" of things we can schedule, an "press sneak for me automatically" and one "remove this overlay after a while". But. The text you add here applies only to the former, since "Press Y to cancel" does not make sense in the latter situation.

Or, at least, this requires some more thought. What would it mean to cancel? Perhaps you need extra time to read it. But if you cancel it, can you "restart" the timer that will clear the message? How? Or is that just making things complicated?

And if you missed the message, how can you re-read it? I don't think these "direct to chat" messages are stored in the quest book history, so if you're a slow reader, it's just lost.

magicus:
I created a splitComponentInLines() method in https://github.com/Wynntils/Artemis/pull/429. If you include it, I think you can properly just start replacing String with Component as you do.

However, as I tried to explain, I don't think you can drop all blank lines, since they are needed for proper detection of dialogue lines.

So I think this PR will in the end boil down to:
1) keeping the general structure on how the "chat screens" are segmented into parts
2) ... but use proper Components instead of coded strings
3) ... and introduce an additional check if a chat message is a "free-floating, no-confirmation-needed" dialogue, which 
4) ... is shown by the NPC Overlay in a suitable way.

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
magicus: I just had a revelation: Perhaps we can solve all the above problems by this behavior: For a scheduled auto progress for non-need-confirmation messages, the text says "Auto-Progress: .... Press Y to move to chat", and pressing Y will then cancel the NPC Dialogue overlay, but at the same time send out the dialogue to the chat. This way you can either give yourself more time to read it, and/or you can get it out of the way.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
kristofbolyai: Why? If we do not have a parsable quest, we set current dialogue to be empty. ...or is the problem that we don't parse the "you walk away.." part? It is not an actual dialogue though, is it?

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: No, we have not tested this. Or at least, I have not.

common/src/main/java/com/wynntils/core/chat/ChatModel.java
magicus: It is treated as part of the dialogue. You need to press shift to get past it. Go do Detlas and check with your code. I don't think it will work.

kristofbolyai:
@magicus Do you plan to implement the new line parser or should I do it?

magicus:
To be honest, I think this PR should be closed, and that we open two new ones. One in which we switch String to Component, and one in which we catch the no-confirmation-needed messages. It's just gotten so messy by now. I can do the former; after all, all the hard work was in creating the splitComponentInLines. And then you can pick out the "free-floating" NPC dialoges and make a separate PR out of it. Sounds OK?

kristofbolyai:
> To be honest, I think this PR should be closed, and that we open two new ones. One in which we switch String to Component, and one in which we catch the no-confirmation-needed messages. It's just gotten so messy by now. I can do the former; after all, all the hard work was in creating the splitComponentInLines. And then you can pick out the "free-floating" NPC dialoges and make a separate PR out of it. Sounds OK?

Yes, good idea.

PR #434, opened 2022-09-23
Use Component instead of String as base for the ChatModel

magicus:


magicus:
This introduces a slight regression since we have no `ComponentUtils.componentFromCoded()`, i.e. a way to create a Component from a coded string (like "§eHello"). So translated messages currently lose their formatting.

Other than that, I tested this a bit and it seems to work. Always slightly scary with changes to the ChatManager, but I think this is ok.

magicus:
This makes the need for #292 even less. (That being said, I still think it could be useful in the future to be able to do a full, loss-less roundtrip between coded strings and Components.)

kristofbolyai (approved):

PR #435, opened 2022-09-23
WIP: Handle non-confirmation NPC dialogues

magicus:
These are the ones where the NPC message just goes to the chat as normal, not requiring you to press sneak.

There are actually two kinds of those, "disappearing" and normal chat messages. The "disappearing" ones are just like normal confirmation NPC dialogues, but they do not require sneak. I have only seen those in King's Recruit. These are handled by Wynncraft by sending out "screens", and when the chat is done, the screen is "redrawn" without it.

The second one is just like normal chat messages.

There are a slew of problems with getting these to work as you'd expect. This is just a draft.

Needs fixing in the framework:
- [ ] Disappearing messages should have a NULL sent when they are cleared. 
- [ ] This means we need an enum to separate between "chat style", "need confirmation", "need no confirmation", not just a boolean
- [ ] There is (at least one) bug when different kinds of messages are mixed, so old "chat style" messages pop up in the overlay when new "need no confirmation" messages gets removed, or something like that.

Needs fixing in the Overlay:
- [ ] "chat style" messages needs a time out. Can probably reuse code from @kristofbolyai's old closed PR.
- [ ] "chat style" messages can stack. In King's Recruit you can get several differnt ones at the same time. 
- [ ] This means NPC overlay needs to handle more than one chat line
- [ ] They should probably be displayed like in the chat, one on top of the other, and each one of them having individual timeouts so they disappear one after the other, so timeouts need to be on a per-line basis, not for the entire overlay
- [ ] How handle if "chat style" messages has not yet timed out, when a "need confirmation" message arrive?

### DIFF
--- a/common/src/main/java/com/wynntils/core/chat/ChatModel.java
+++ b/common/src/main/java/com/wynntils/core/chat/ChatModel.java
@@ -92,7 +92,15 @@ public final class ChatModel extends Model {
         if (!codedMessage.contains("\n") || codedMessage.indexOf('\n') == (codedMessage.length() - 1)) {
             saveLastChat(message);
             MessageType messageType = e.getType() == ChatType.SYSTEM ? MessageType.SYSTEM : MessageType.NORMAL;
-            Component updatedMessage = handleChatLine(message, codedMessage, messageType);
+            RecipientType recipientType = getRecipientType(message, messageType);
+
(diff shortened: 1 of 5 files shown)


magicus:
Replaced by https://github.com/Wynntils/Artemis/pull/767

PR #437, opened 2022-09-23
Add full screen map

kristofbolyai:
This PR also fixes the bug where minimap rendering was blurry.


common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,270 @@
...
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyUp.key.getValue()));
+        KeyMapping.set(
+                McUtils.mc().options.keyRight.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyRight.key.getValue()));
+        KeyMapping.set(
+                McUtils.mc().options.keyJump.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyUp.key.getValue()));
+        KeyMapping.set(
+                McUtils.mc().options.keyShift.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyShift.key.getValue()));
```
Incompleteusern, line 86: What does this do?

Benj2005:
There seems to be an issue where the full screen map doesn't quite fit within it's border.
(Replicated on both full screen 1080p and full screen 1440p.)

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: When opening a map, you stop running/shifting/etc. This makes it so you don't (as it is done in legacy)

magicus:
I just tested this. Awesome!!! This was like the last major thing missing from Artemis. I think I'll pause with the chat stuff and help get labels and POIs on the map now. :)

(Now that I tested it I'm going to look at the code. Usually I do it the other way round :))

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,270 @@
...
+    private float mapWidth;
+    private float mapHeight;
+    private float centerX;
+    private float centerZ;
+
+    private float mapCenterX;
+    private float mapCenterZ;
+
+    // Zoom is the scaling of the map. The bigger the zoom, the less detailed the map becomes.
+    private static final float MIN_ZOOM = 5.0f;
```
magicus: This is a bit confusing, having "min" larger than "max". One alternative is to replace the zoomFactor below with it's inverse, i.e. 1/zoomFactor. That way min zoom would be about 0.2 and max zoom about 3. 

I'm not sure this makes more sense, but maybe. Can we "ground" the zoom factor into some understandable "truth"? E.g. if zoom factor 1 means max zoomed in or max zoomed out, or one map pixel per screen pixel, or whatever.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,270 @@
...
+                McUtils.mc().options.keyRight.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyRight.key.getValue()));
+        KeyMapping.set(
+                McUtils.mc().options.keyJump.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyUp.key.getValue()));
+        KeyMapping.set(
+                McUtils.mc().options.keyShift.key,
+                KeyboardUtils.isKeyDown(McUtils.mc().options.keyShift.key.getValue()));
+
+        McUtils.mc().keyboardHandler.setSendRepeatsToGui(true);
```
magicus, line 88: Oh, you found this! Maybe it can be used for the repeating backspace in the text edit widget as well?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: Sure, I can try to change it.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: Perhaps, I did not try it because we have the EditBox you mentioned as well. We have both of these to test out later.

kristofbolyai:
> I just tested this. Awesome!!! This was like the last major thing missing from Artemis. I think I'll pause with the chat stuff and help get labels and POIs on the map now. :)
> 
> (Now that I tested it I'm going to look at the code. Usually I do it the other way round :))

I think I already have a generalized POI renderer in my mind. Do you think you could work with Scyu on the data we collected, and create a v2 POI endpoint? Or is that too much work, as we do not have that fully automized yet in legacy..?

magicus:
Let's talk POI in Discord instead of in PR comments. I have a lot of ideas, and a few very strong opinions. :-)

Regarding this PR. It is really OK to merge, and part of me is eager to get it in and then start working from there. But there are a couple of things I think should be fixed quite soon, and maybe it's best to do it before integration.

First of all, in my mind, it kind of makes sense to have "main map" and "minimap" as separate features. From a user PoV, it is quite different beasts. 

The second, and tightly related question, is if we should allow the user to select different configuration for minimap and main map. E.g. pointer style. On the one hand, it seems liberal to allow the user to modify all kind of configuration. But on the other hand, that means that if a user wants the two maps to look the same (which I think is reasonable for most users), then they also are forced to update e.g. pointer style in two different places.

This isse ties in with the first. So if we do as currently in this PR and have a single MapFeature for both minimap and main map, then it kind of also implies that we have a single config option "pointer style". Otherwise, we will get a (non-warranted) config explosion in the single MapFeature config -- "Minimap Pointer Style" and "Main Map Pointer Style" etc, for all config options. 

So if we think there should be a single config style for both maps, then they are to be considered two aspects of the same feature. But if we think we should be able to configure them separately, then they should be two separate features.

I'm on the fence. I see pros and cons with both approaches. But I'm leaning slightly towards the "two features, different configs" approach.

common/src/main/java/com/wynntils/features/user/overlays/map/MapFeature.java
```diff
@@ -32,12 +28,24 @@
...
-import org.lwjgl.opengl.GL11;
-import org.lwjgl.opengl.GL13;
+import org.lwjgl.glfw.GLFW;
+
+public class MapFeature extends UserFeature {
+    public static MapFeature INSTANCE;
+
+    @RegisterKeyBind
+    public final KeyBind openMapKeybind = new KeyBind("Open Full Screen Map", GLFW.GLFW_KEY_M, false, () -> {
+        if (McUtils.mc().screen instanceof MainMapScreen mainMapScreen) {
```
magicus, line 32: I'm trying to figure out how this works. Are we using the repeat delay from vanilla? So if we keep M pressed, after a while vanilla is going to send this as a repeated event, and if that is done when the map is displaying, we change mode to "close on key release"? Since we are not sending in `true` to the "just first press" boolean or whatever it was called?

If my guess is correct, maybe write a few lines about it? It is kind of magic otherwise.


I assume this was how Legacy also behaved, even though I never noticed or used it. It is very cool anyway, being able to quickly see the entire map just by holding M a bit longer.

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
```diff
@@ -947,6 +951,131 @@ public static void rotatePose(PoseStack poseStack, float centerX, float centerZ,
...
+            }
+
+            float halfRenderedWidth = width / 2f * extraFactor;
+            float halfRenderedHeight = height / 2f * extraFactor;
+            float halfTextureWidth = halfRenderedWidth * scale;
+            float halfTextureHeight = halfRenderedHeight * scale;
+
+            Matrix4f matrix = poseStack.last().pose();
+
+            // TODO replace with RenderUtils after settings pr is merged
```
magicus: Is this comment still relevant? I'm not sure what it means anyway.

kristofbolyai: Not sure either

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
magicus: Why do we do that? Can you like run and look at the map at the same time in Legacy? 😮

kristofbolyai: Yes

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        return true;
+    }
+
+    private void setCompassToMouseCoords(double mouseX, double mouseY) {
+        double gameX = (mouseX - centerX) / currentZoom + mapCenterX;
+        double gameZ = (mouseY - centerZ) / currentZoom + mapCenterZ;
+        Location compassLocation = new Location(gameX, 0, gameZ);
+        CompassModel.setCompassLocation(compassLocation);
```
magicus, line 231: Bonus points for this! :) It's a clear sign that we've done stuff right when it is so easy to implement.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,280 @@
...
+                mapWidth,
+                mapHeight,
+                1f / currentZoom,
+                false,
+                false);
+    }
+
+    private void updateMapCenterIfDragging(int mouseX, int mouseY) {
+        if (dragging) {
+            float zoomScale = currentZoom;
```
magicus: Don't think that's needed any longer :-)

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -0,0 +1,280 @@
...
+                100,
+                renderWidth,
+                renderHeight,
+                Texture.FULLSCREEN_MAP_BORDER.width(),
+                Texture.FULLSCREEN_MAP_BORDER.height());
+    }
+
+    @Override
+    public boolean mouseScrolled(double mouseX, double mouseY, double delta) {
+        setZoom((float) (currentZoom + delta * ZOOM_FACTOR));
```
magicus: Aha, that's where it is used. Maybe rename to `MOUSE_SCROLL_ZOOM_FACTOR`?

common/src/main/java/com/wynntils/mc/render/RenderUtils.java
magicus: Remove it then :)

common/src/main/java/com/wynntils/features/user/overlays/map/MapFeature.java
kristofbolyai: Added the comment.

magicus (approved):

PR #438, opened 2022-09-24
Corrupted Bar Overlay

Incompleteusern:
Untested as I don't have a corrupted warrior, but I think @byBackfish does :).

Port of https://github.com/Wynntils/Wynntils/pull/595

Also there is some minor changes shoehorned, sorry.


byBackfish:
Yep, I'll test it in a few minutes

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
```diff
@@ -17,11 +17,13 @@ public class WynnBossBarUtils {
     public static final BarProgress NO_BLOOD_POOL = new BarProgress(0, 0, -1);
     public static final BarProgress NO_AWAKENED_PROGRESS = new BarProgress(0, 0, -1);
     public static final BarProgress NO_FOCUS = new BarProgress(0, 0, -1);
+    public static final BarProgress NO_CORRUPTED = new BarProgress(0, 100, -1);
 
     private static final Pattern BLOOD_POOL_PATTERN = Pattern.compile("§cBlood Pool §4\\[§c(\\d+)%§4\\]");
     private static final Pattern MANA_BANK_PATTERN = Pattern.compile("§bMana Bank §3\\[(\\d+)/(\\d+)§3\\]");
     private static final Pattern AWAKENED_PROGRESS_PATTERN = Pattern.compile("§fAwakening §7\\[§f(\\d+)/(\\d+)§7]");
     private static final Pattern FOCUS_PATTERN = Pattern.compile("§eFocus §6\\[§e(\\d+)/(\\d+)§6]");
+    private static final Pattern CORRUPTED_PROGRESS_PATTERN = Pattern.compile("§cCorrupted §4\\[§c(\\d+)%§4]§r");
```
byBackfish: The bossbar's in here are different from 1.12.2. They don't end with `§r`
```suggestion
    private static final Pattern CORRUPTED_PROGRESS_PATTERN = Pattern.compile("§cCorrupted §4\\[§c(\\d+)%§4]");
```

byBackfish (changes requested):
You also need to add the key to the language file

byBackfish:
image
is it intended that the text is bound to the right?

kristofbolyai:
> image
> 
> is it intended that the text is bound to the right?
> 
> 

I think that's just configs?

common/src/main/java/com/wynntils/wynn/utils/WynnBossBarUtils.java
```diff
@@ -111,5 +113,21 @@ public static BarProgress getFocusBar() {
...
     }
 
+    public static BarProgress getCorruptedBar() {
+        Pair<LerpingBossEvent, Matcher> pair = getLerpingBossEvent(CORRUPTED_PROGRESS_PATTERN);
+
+        if (pair.a == null) return NO_CORRUPTED;
+
+        try {
+            int currentCorruptedProgress = Integer.parseInt(pair.b.group(1));
+            int maxCorruptedProgress = 100;
```
kristofbolyai, line 123: Can we make this a const in the class?

I am sad we cannot parse this... Or can we? From what I see in the regex, isn't this like blood pool, where we can use the lerpingBossEvent to parse this "better"? Although, if it is a 0-100% range, we might not need that..

Incompleteusern: It is 0-100% range

Incompleteusern:
> > image
> > is it intended that the text is bound to the right?
> 
> I think that's just configs?

It should be centered by default

byBackfish (approved):

byBackfish:
works fine then 👍

common/src/main/java/com/wynntils/features/user/overlays/CustomBarsOverlayFeature.java
```diff
@@ -350,31 +367,13 @@ public String icon() {
...
-        }
-
         @Override
         public WynnBossBarUtils.BarProgress noProgress() {
             return WynnBossBarUtils.NO_AWAKENED_PROGRESS;
         }
     }
 
-    public static class FocusBarOverlay extends BaseBarOverlay {
+    public class FocusBarOverlay extends BaseBarOverlay {
```
magicus, line 376: @Incompleteusern Why did you turn all nested classes into inner classes? Qodana points out that it was not needed.

Incompleteusern: Oops, brain slipped.

PR #440, opened 2022-09-24
Create new top-level gui package

magicus:
I have moved `screen` in under `gui`, moved the `mc.render` package to `gui.render` and moved the old `screen.widget` up one level, so now it is `gui.widget`. I think this makes much more sense.


magicus:
I also made MapRenderer a top-level class.

Going forward, I think there is could be some work restructuring gui.screen.overlay and gui.screen.settings. Perhaps. Or they are fine as they are.

kristofbolyai (dismissed):

kristofbolyai (approved):

PR #442, opened 2022-09-24
feat: add trade market emerald amount conversion

byBackfish:
Adds the ability to use shortened prices in the trade market (`10le` -> `40960`)

Ported over from Wynntils


common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -85,6 +86,7 @@ public static void init() {
         // always on
         registerFeature(new FixSpellOverwriteFeature());
         registerFeature(new LootrunFeature());
+        registerFeature(new TradeMarketPriceConversionFeature());
```
Incompleteusern: Why is this always on?

common/src/main/java/com/wynntils/features/alwayson/TradeMarketPriceConversionFeature.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.wynntils.core.features.StateManagedFeature;
+import com.wynntils.mc.event.ChatSentEvent;
+import com.wynntils.utils.StringUtils;
+import com.wynntils.wynn.event.ChatMessageReceivedEvent;
+import com.wynntils.wynn.utils.WynnUtils;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class TradeMarketPriceConversionFeature extends StateManagedFeature {
```
kristofbolyai: Why state managed? Give the user the option to disable. StateManagedFeature is only used in really specific scenarios, not when it is not logical to disable the feature.

common/src/main/java/com/wynntils/features/alwayson/TradeMarketPriceConversionFeature.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private static final Pattern PRICE_PATTERN =
+            Pattern.compile("^§6Type the price in emeralds or type 'cancel' to cancel:$");
+
+    private boolean shouldConvert = false;
+
+    @SubscribeEvent
+    public void onChatMessageReceive(ChatMessageReceivedEvent event) {
+        if (!WynnUtils.onWorld()) return;
```
kristofbolyai: Event is not posted if we are not on world. No need for this check.

common/src/main/java/com/wynntils/features/alwayson/TradeMarketPriceConversionFeature.java
```diff
@@ -0,0 +1,39 @@
...
+
+        if (PRICE_PATTERN.matcher(event.getCodedMessage()).matches()) {
+            shouldConvert = true;
```
kristofbolyai: What if something happens and this stays true? It might be how legacy does it, but I am not sure how to fix this..

common/src/main/java/com/wynntils/utils/StringUtils.java
```diff
@@ -159,4 +169,47 @@ public static String getMaxFittingText(String text, float maxTextWidth, Font fon
...
+            while (eMatcher.find()) {
+                emeralds += Long.parseLong(eMatcher.group(1));
+            }
+
+            // account for tax if flagged
+            if (input.contains("-t")) {
+                emeralds = Math.round(emeralds / taxAmount);
+            }
+        } catch (NumberFormatException e) {
+            return ""; //
```
kristofbolyai: Comment?

byBackfish:
Should solve all cases, where the user would exit the price input of the trade market where Artemis would still think the user is in it and then replaces the message

kristofbolyai (approved):
Looks okay to me.

Incompleteusern (approved):

PR #444, opened 2022-09-24
Create a new MapModel to store map-related info

magicus:
My plan is to add POI knowledge to this model as well.


kristofbolyai:
I was just about to start working on rendering service textures on map. I guess I need to wait for this.

kristofbolyai:
I don't think that pulling and deserializing the location json or the services json is that much work though (getting the textures ready will be fun for me...)?

magicus:
No, it's not much work. I just started to create a good place to put it in. But I was actually planning to spend an hour or so tonight trying to fix it, so if you can wait a bit I can do it. You can start working on the textures. :)

magicus:
And if you are okay with it, I'd like to merge this part now (since this is a pure refactoring), and then add additional functionality in an follow-up PR.

kristofbolyai:
> And if you are okay with it, I'd like to merge this part now (since this is a pure refactoring), and then add additional functionality in an follow-up PR.

Sure. Although this code you moved is buggy, you are obviously not have to fix that right now.

kristofbolyai (approved):

magicus:
@kristofbolyai I know that it is not really up to our standards, but I did thought it worked. In what way is it buggy?

kristofbolyai:
> @kristofbolyai I know that it is not really up to our standards, but I did thought it worked. In what way is it buggy?

Sometimes the map fails to load, even though every part is cached. Happens more in prod from what I've seen.

magicus:
Oh.

I noticed problems with our WebManager / Request handling. So even though I had said "download and cache, and use cache as fallback", I still did not get one of my new json files. It almost drove me crazy. It turned out that I had used `handleJsonObject` when the top-level object was an array, and that caused `handleJsonObject` to return false before even calling my code. And this apparently caused the WebManager to delete the cached file (I tried copying it in manually), download the new file (I could verify this), and then not save it in cache. *duh*

Maybe you have been seeing something similar?

kristofbolyai:
Perhaps, only thing I know is that the map or the map json fails to load from cache.

PR #446, opened 2022-09-24
Add first kinds of POIs to MapModel

magicus:


common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -36,6 +52,60 @@ public static List<MapProfile> getMaps() {
...
     }
 
+    public static Set<Poi> getAllPois() {
+        return allPois;
+    }
+
+    public static void loadLabels() {
+        File mapDirectory = new File(WebManager.API_CACHE_ROOT, "maps");
+
+        String url = "https://raw.githubusercontent.com/Wynntils/Reference/main/locations/places.json";
```
kristofbolyai: Perhaps this could be extracted to the top of the class. Perhaps, it is fine for it to stay here.

common/src/main/java/com/wynntils/wynn/model/map/poi/Label.java
```diff
@@ -0,0 +1,55 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.model.map.poi;
+
+public class Label {
+    private String name;
+    private int x;
+    private int z;
+    private int layer;
+    private String level;
```
kristofbolyai, line 12: I am not sure why level is a string? -1 for internal representation of undefined and return an optional in the getter? Or return -1?

kristofbolyai:
By the way, I started working on textures.

common/src/main/java/com/wynntils/wynn/model/map/poi/Label.java
magicus: Because that is how it is in the original data. Most of the time it is an integer, but in some places it is like "30-40" or some other weird format. We could "fix" this manually and change these values, but currently the data is non-curated and only extracted from Wynncraft's official map, and converted to a sane json representation.

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
magicus: Yeah, no, that is crap. :-) Just wanted to get this starting, and I was kind of half sleeping in front of the computer yesterday night.

magicus:
A general note: Right now the way to process POIs is to get the set of all POIs, and filter on them. I like the generality of this approach, but it risks being too expensive performance-wise. But, let us start out this way, and only fix it if we can actually show that it is a performance problem.

The alternative is likely to kind of have an empty set of POIs to begin with, and only add those categories that the user enable. But that makes it harder to have like "show this category at zoom level X", and we are still likely to worry about "why do I have to iterate through all item identifiers in the entire world when the minimap just shows Cinfras?"

But yeah, I worry a bit, since we will be looping over a potentially very large set, in rendering.

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -109,4 +175,13 @@ public static CompletableFuture<Boolean> tryLoadMaps() {
     public static boolean isMapLoaded() {
         return !maps.isEmpty();
     }
+
+    private static class PlacesProfile {
+        List<Label> labels;
+    }
+
+    private static class ServiceProfile {
```
kristofbolyai, line 183: Do I understand correctly that these "profiles" are basically deserialization helpers?

kristofbolyai (approved):

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
magicus: Yes. We've traditionally used `*Profile` as name for stuff whose structure are forced upon us by external data. It's kind of odd, but also marks the classes clearly. But in this case I could perhaps rename them to something like `ServiceProfileJsonDeserializer` to make this more explicit.

PR #447, opened 2022-09-25
Create /locate and /compass commands

magicus:
Usage:

`/locate place|service <name>` 
shows a list of matches. Clicking on a match sets the compass. `<name>` is matched using the partial match so "dtls" matches "Detlas" and "Detlas Suburbs", while "dtlsu" only matches "Detlas Suburbs"

`/compass place|service <name>`
Similar to locate, but automatically sets the compass to the matching place, if a unique place could be determined (either by fuzzy matching only a single hit, or by having an exact match, so "Detlas" is ok and will not be considered a partial match of "Detlas Suburbs"). For services, the closest service is selected.

`/compass at <location>`
Uses vanilla coordinates and sets the compass to that exact coordinates, e.g. `/compass at ~ ~ ~` sets the compass at your current location.


common/src/main/java/com/wynntils/commands/CompassCommand.java
```diff
@@ -0,0 +1,177 @@
...
+                        .build())
+                .executes(this::syntaxError);
+    }
+
+    private int compassAt(CommandContext<CommandSourceStack> context) {
+        Coordinates coordinates = Vec3Argument.getCoordinates(context, "location");
+        Location location = new Location(coordinates.getBlockPos(context.getSource()));
+        CompassModel.setCompassLocation(location);
+
+        MutableComponent response = new TextComponent("Setting compass to ").withStyle(ChatFormatting.AQUA);
```
Incompleteusern: maybe "Set compass to x y z" is better, as it isnt really a continuous process

common/src/main/java/com/wynntils/commands/CompassCommand.java
```diff
@@ -0,0 +1,177 @@
...
+            // This really should not happen...
+            MutableComponent response = new TextComponent("Found no services of kind '" + selectedKind.getName() + "'")
+                    .withStyle(ChatFormatting.RED);
+            context.getSource().sendFailure(response);
+            return 0;
+        }
+        Poi closestService = closestServiceOptional.get();
+        CompassModel.setCompassLocation(closestService.getLocation().asLocation());
+
+        MutableComponent response = new TextComponent("Setting compass to " + selectedKind.getName() + " at ")
```
Incompleteusern: Same comment as above

common/src/main/java/com/wynntils/commands/CompassCommand.java
```diff
@@ -0,0 +1,177 @@
...
+                .withStyle(ChatFormatting.AQUA);
+        response.append(new TextComponent(closestService.getLocation().toString()).withStyle(ChatFormatting.WHITE));
+        context.getSource().sendSuccess(response, false);
+        return 1;
+    }
+
+    private int compassPlace(CommandContext<CommandSourceStack> context) {
```
Incompleteusern, line 124: This is really similar to `/territory`

common/src/main/java/com/wynntils/commands/CompassCommand.java
magicus: That sounds like an imperative to the user... Maybe "Compass set to ..."?

Incompleteusern: Yeah

common/src/main/java/com/wynntils/commands/CompassCommand.java
magicus: I'm planning to look how I can unite the territory code with the new MapModel. The difference here is the selection of POIs that this draws from; this here is place names from the Wynncraft web map, not territories.

common/src/main/java/com/wynntils/commands/LocateCommand.java
```diff
@@ -0,0 +1,176 @@
...
+                        && servicePoi.getKind().equals(selectedKind))
+                .toList());
+
+        // Only keep the closest results
+        Vec3 currentLocation = McUtils.player().position();
+        services.sort(Comparator.comparingDouble(poi -> currentLocation.distanceToSqr(
+                poi.getLocation().getX(),
+                poi.getLocation().getY(),
+                poi.getLocation().getZ())));
+        services.subList(4, services.size()).clear();
```
kristofbolyai, line 105: What is this?

common/src/main/java/com/wynntils/wynn/model/map/poi/MapLocation.java
```diff
@@ -26,4 +28,17 @@ public int getY() {
...
+        return new Location(x, y, z);
+    }
+
+    @Override
+    public String toString() {
+        return "[" + x + ", " + y + ", " + z + "]";
+    }
+
+    public String asChatCoordinates() {
+        return x + " " + y + " " + z;
```
kristofbolyai, line 42:
```suggestion
        return x + ", " + y + ", " + z;
```

Perhaps this is better?

kristofbolyai (dismissed):
Seems good overall

common/src/main/java/com/wynntils/commands/LocateCommand.java
magicus: Remove everything except the three first elements in the list. As suggested by Stack Overflow.

common/src/main/java/com/wynntils/wynn/model/map/poi/MapLocation.java
magicus: No, this is "chat coordinates", that is, how Vanilla expects a coordinate to be written. This way I can use vanilla functionality. 

When normally presenting to user, I have the toString method with commas.

I also intend to make a more forgiving input method, like in Legacy, for anything "coordinate like". I'm not just there yet.

common/src/main/java/com/wynntils/commands/LocateCommand.java
kristofbolyai: Ah.., makes sense. Maybe I would have figured out after staring for like 10 minutes. I just spent 1 hour wondering why my POI rendering trigonometry did not work... (I centered my math around the screen and not the map center.... :facepalm: )

kristofbolyai (approved):
Just as a note, compass share is a todo. I know you won't forget that though.

PR #450, opened 2022-09-25
ServiceKind: Add textures

kristofbolyai:


magicus:
@kristofbolyai powder_merchant.png looks like a "powder station" since it has the hammer and not an emerald...

kristofbolyai:
But then why is there powder merchant in Detlas? There should be one in corkus only? And where are powder stations, aka the powder combiner npcs?

magicus:
I think we sorted this out. However, the "powder station" got me curious. Was it the case previously that the Powder Master NPC was like a worsktation block?

PR #451, opened 2022-09-25
MapRenderer: Render POIs 

kristofbolyai:
Tasks:

- [x] Render ServicePois
  - [x] Fix text render overlap
- [x] Render LabelPois


common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,12 +109,137 @@ public static void renderMapQuad(
...
+
+                    float textureXPosition = (float) (centerX + distanceX / scale);
+                    float textureZPosition = (float) (centerZ + distanceZ / scale);
+
+                    boolean inMap = textureXPosition >= mapTopX
+                            && textureXPosition <= mapBottomX
+                            && textureZPosition >= mapLeftZ
+                            && textureZPosition <= mapRightZ;
+
+                    return poi instanceof ServicePoi && inMap;
```
magicus: Recommend that you a) start by testing the instanceof to quickly abort if not relevant, and b) extracting the inMap test in a method.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,12 +109,137 @@ public static void renderMapQuad(
...
+
+        poseStack.pushPose();
+
+        float renderX = textureXPosition - width / 2f;
+        float renderZ = textureZPosition - height / 2f;
+        poseStack.translate(renderX, renderZ, 0);
+
+        if (mouseX >= renderX && mouseX <= renderX + width && mouseY >= renderZ && mouseY <= renderZ + height) {
+            FontRenderer.getInstance()
+                    .renderText(
```
magicus, line 274: I think it's great if we have the possibility to have text labels for all POIs, but I also think this needs to be a setting which should probably be off by default (and in the future, perhaps setting it to on/off differently depending on what kind of POI it is, and zoom level etc).

kristofbolyai: This only works on main map, on hover. I think it is fine to leave it on.

magicus: Aha, it is on hover. I missed that part. That is very good!

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,226 @@ public static void renderMapQuad(
...
+                    poiScale,
+                    servicePoi);
+        }
+
+        if (McUtils.mc().screen instanceof MainMapScreen mainMapScreen) {
+            mainMapScreen.setHovered(hovered);
+        }
+
+        if (hovered != null && mouseCoordinates != null) {
+            float textureXPosition = hovered.getRenderX(mapCenterX, centerX, 1f / scale);
```
magicus: How can this happen? If mouseCoordinates is != null we set hovered to null first thing in the method. I can't see how this ever can be true?

magicus: Ah, I see, we set hovered in renderServicePoi. Not completely obvious.

kristofbolyai: This check might not be needed, and is there just to prevent weirdness.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,226 @@ public static void renderMapQuad(
...
+            textureZPosition = servicePoi.getRenderZ(mapCenterZ, centerZ, 1f / scale);
+        }
+
+        float width = servicePoi.getIcon().width() * poiScale;
+        float height = servicePoi.getIcon().height() * poiScale;
+
+        float renderX = textureXPosition - width / 2f;
+        float renderZ = textureZPosition - height / 2f;
+
+        if (mouseCoordinates != null) {
```
magicus: Maybe you can extract this out of renderServicePoi and do it separately when looping over all service POIs. If you do it first, you can send in a "isHovered" or whatever to say that the icon should be bigger.

common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
```diff
@@ -20,4 +20,24 @@ public MapLocation getLocation() {
     public abstract Texture getIcon();
 
     public abstract String getName();
+
+    /**
```
magicus: Not entirely sure this belongs as part of a Poi. Looks more like a RenderUtils helper function to me; all it needs is the getLocation.getX/Z from the poi. It looks like there's a lot of rendering logic here, while the rest of the Poi class is just a "pure" representation of a location.

magicus:
Mostly look good! Just fix so hovered is not sneakily set by rendering, and ideally stop polluting my pristine Poi class ;-) and you're good to go.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
kristofbolyai: Well, I would have to extract a lot of math and calculate everything twice. Not good.

kristofbolyai: All of this
```java
        float textureXPosition;
        float textureZPosition;

        if (followPlayerRotation) {
            final double rotationRadians = Math.toRadians(McUtils.player().getYRot());
            final float sinRotationRadians = (float) StrictMath.sin(rotationRadians);
            final float cosRotationRadians = (float) -StrictMath.cos(rotationRadians);

            float dX = (servicePoi.getLocation().getX() - mapCenterX) / scale;
            float dZ = (servicePoi.getLocation().getZ() - mapCenterZ) / scale;

            textureXPosition = centerX + (dX * cosRotationRadians - dZ * sinRotationRadians);
            textureZPosition = centerZ + (dX * sinRotationRadians + dZ * cosRotationRadians);
        } else {
            textureXPosition = getRenderX(servicePoi, mapCenterX, centerX, 1f / scale);
            textureZPosition = getRenderZ(servicePoi, mapCenterZ, centerZ, 1f / scale);
        }

        float width = servicePoi.getIcon().width() * poiScale;
        float height = servicePoi.getIcon().height() * poiScale;

        float renderX = textureXPosition - width / 2f;
        float renderZ = textureZPosition - height / 2f;
```

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,226 @@ public static void renderMapQuad(
...
+        // TODO: This is really basic at the moment
+        //       Add fading, and other configs
+
+        float textureXPosition;
+        float textureZPosition;
+
+        if (followPlayerRotation) {
+            final double rotationRadians = Math.toRadians(McUtils.player().getYRot());
+            final float sinRotationRadians = (float) StrictMath.sin(rotationRadians);
+            final float cosRotationRadians = (float) -StrictMath.cos(rotationRadians);
```
Incompleteusern: Should be cached

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,226 @@ public static void renderMapQuad(
...
+                double distanceZ = z - mapCenterZ;
+
+                float textureXPosition = (float) (centerX
+                        + distanceX / scale
+                        - FontRenderer.getInstance()
+                                        .getFont()
+                                        .width(labelPoi.getLabel().getName())
+                                / 2f
+                                / scale);
+                float textureZPosition = (float) (centerZ + distanceZ / scale - 4.5f / scale);
```
Incompleteusern: Can't you just change the alignment configs

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,231 @@ public static void renderMapQuad(
...
+            float textureXPosition = getRenderX(hovered, mapCenterX, centerX, 1f / scale);
+            float textureZPosition = getRenderZ(hovered, mapCenterZ, centerZ, 1f / scale);
+
+            float width = hovered.getIcon().width() * poiScale;
+            float height = hovered.getIcon().height() * poiScale;
+
+            poseStack.pushPose();
+
+            float renderX = textureXPosition - width / 2f;
+            float renderZ = textureZPosition - height / 2f;
```
Incompleteusern, line 270: Same comment

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -99,10 +116,235 @@ public static void renderMapQuad(
...
+                            CommonColors.GREEN,
+                            HorizontalAlignment.Center,
+                            VerticalAlignment.Top,
+                            FontRenderer.TextShadow.OUTLINE);
+
+            poseStack.popPose();
+        }
+    }
+
+    private static boolean isPoiVisible(
```
Incompleteusern, line 288: This should be inlined

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
kristofbolyai: Scale needs to be factored in, font renderer can't do that.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
kristofbolyai: What? Alignment to what? This is so POIs are rendered centered to their coordinates. No config needed here.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
magicus: Oh, I missed that. Hm.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
kristofbolyai: Magicus wanted this extracted

magicus: Actually I requested that he extracted it! 😆 For clarity.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -13,27 +13,44 @@
...
 
 public class MapRenderer {
+    private static Poi hovered = null;
+
     // TODO: Support circle map rendering
     public static void renderMapQuad(
             MapProfile map,
             PoseStack poseStack,
+            float mapCenterX,
+            float mapCenterZ,
```
Incompleteusern, line 39: These 2 variables can be removed in all the remaining methods by translating before rendering

kristofbolyai: I don't think so? They are used a lot in the code. Even if we could remove them, it might be clearer to leave them.

Incompleteusern: Oops, I meant to refer to `centerX` and `centerZ`

Incompleteusern:
In general, my main problem is the sheer amount of variables passed to each method

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
magicus: I have a suggestion, but I understand if you do not want to follow it. Turn this calculation into a Pair<float,float> getRenderPosition() that returns the renderX and renderZ, to the loop. Then use those to a) check the hover and b) call drawScalingTexturedRect.

I think it would be better, by still containing the maths separately, but showing that we are doing two things (hover calucation and rendering) per poi.

kristofbolyai:
> In general, my main problem is the sheer amount of variables passed to each method

I can understand that, but this is meant to be only used in the class as helper methods. We kinda have to work with a lot of things at once, so this might not be that easy to fix.

Incompleteusern:
> > In general, my main problem is the sheer amount of variables passed to each method
> 
> I can understand that, but this is meant to be only used in the class as helper methods. We kinda have to work with a lot of things at once, so this might not be that easy to fix.

I can try to make a follow up pr to try and reduce this, not a concern that much rn

magicus:
If we mostly "pass along" a lot of variables, we can also group them in an object like `MapRenderTask` (if it's  more or less self-containing) or `MapRenderContext` (if that's just partial info about what is needed). This makes sense if they are just passed along through multiple layers, but not if they are changed at several layers, or gradually built up layer by layer. I must say I did not follow this dataflow exactly.

magicus:
(I agree that it is a lot of variables. It is definitely a toll on readability. But I also understand that it can be tricky to do anything about...)

Incompleteusern (commented):
I feel that we could reasonably move some of the code to the poi itself, especially of merging service poi and normal poi rendering. However, this is fine as is

Incompleteusern:
oops, closed it

Incompleteusern (dismissed):

Incompleteusern (approved):

magicus (approved):

PR #452, opened 2022-09-26
Opt to use `WynntilsMod#Error` in place of throwing exception when possible

Incompleteusern:


kristofbolyai (commented):
I am not sure if this is a good change..

magicus:
I agree with bolyai, this is probably misguided. We already do log when we have recoverable errors. Exceptions are thrown for irrecoverable errors, basically. Removing the throwing just makes the program continue run even when we have established that there is something seriously wrong.

And, on top of that, with the new crash protection, throwing an exception in a feature will instead turn off that feature. By removing the exception you will effectively disable that mechanism.

PR #454, opened 2022-09-26
MapRenderer: Add waypoint POI rendering

kristofbolyai:


kristofbolyai:
Pushed and reverted this on main by accident..

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -57,7 +62,15 @@ public static List<MapProfile> getMaps() {
     }
 
     public static Set<Poi> getAllPois() {
```
magicus: I was vaguely thinking this should return Stream<Poi>. I think this fits even better now. If you want to, please refactor it to do that. I presume that concatting streams is a much cheaper operation than merging sets.

magicus (approved):
This was *much* better! I am positively surprised at how nice that came out. Clearly the Poi system was meant to be Stream-based. I was just to scared of the new strange thing in designing an API which returned a Stream instead of a collection to fully realize it.

PR #457, opened 2022-09-26
MapModel: Try to reduce hacks and add logging

kristofbolyai:
Out of 10 runs, this worked 10 times. I am not sure how else better test this...

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
@@ -173,6 +173,8 @@ public final class MapModel extends Model {
                         thread.join();
                     } catch (InterruptedException e) {
                         WynntilsMod.error("Exception when loading map files.", e);
+                        result.complete(false);
+                        return true;
                     }
 
                     if (syncList.size() == mapArray.size()) {
--- a/common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
@@ -167,15 +167,22 @@ public final class MapModel extends Model {
                                 .build());
                     }
 
-                    // hacky way to know when handler has finished dispatching
-                    CompletableFuture.runAsync(handler::dispatch).whenComplete((a, b) -> {
-                        if (syncList.size() == mapArray.size()) {
-                            result.complete(true);
-                            maps = syncList;
-                        } else {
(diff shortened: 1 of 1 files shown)


Incompleteusern:
`Thread#join` is blocking, this is the same as synchronized dispatching.

I still think the only good solution is code that runs on grouped request completion being natively supported

kristofbolyai:
> `Thread#join` is blocking, this is the same as synchronized dispatching.
> 
> 
> 
> I still think the only good solution is code that runs on grouped request completion being natively supported

But we need to get the map info first to know what files to download?

Incompleteusern:
> > `Thread#join` is blocking, this is the same as synchronized dispatching.
> > I still think the only good solution is code that runs on grouped request completion being natively supported
> 
> But we need to get the map info first to know what files to download?

Map info can be afforded to get downloaded synchronously in my opinion, while downloading maps should done by asynchronously as a group.

kristofbolyai:
> > > `Thread#join` is blocking, this is the same as synchronized dispatching.
> 
> > > I still think the only good solution is code that runs on grouped request completion being natively supported
> 
> > 
> 
> > But we need to get the map info first to know what files to download?
> 
> 
> 
> Map info can be afforded to get downloaded synchronously in my opinion, while downloading maps should done by asynchronously as a group.

But how is this relevant to us having issue with downloading the map info and a singular map in the first place?

Incompleteusern:
I don't see what you are saying?

magicus:
There are two parts involved in downloading maps: 

1. Getting the metadata (how many maps, which x/z-coordinates do they cover
2. Geting the png files

The first is essential; we can't really do anything without it. This is a small json file and we might consider blocking the init of MapModel until that is done. (I think this is what Incomplete is suggesting)

The second can be done in the background. Rendering code will just get a null for that part of the map until it finally arrives, at which point it starts to return non-null. If we want, we can add a "MapLoadedEvent" to signal this as well.

There is a point in doing even #1 asynchronously, so we don't block the main thread, but the drawback is that we get the current double whammy with waiting for an async event just to start *another* async event, which does not look good in code.

magicus:
I am happy with this hack for now to avoid issues, while we figure out how to redo this properly.

magicus (approved):

Incompleteusern (approved):
I guess I am fine with the `Thread#join` blocking for now until we actually get requests sorted out

PR #459, opened 2022-09-27
Merge dependencies under `ModelDependent`

Incompleteusern:
This makes it so `ManagerRegistry` only maintains one list


common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -57,105 +56,88 @@ public static void init() {
...
-        }
-    }
+        WynntilsMod.registerEventListener(manager);
 
-    private static void tryInitManager(Class<? extends Manager> manager) {
         try {
             MethodUtils.invokeExactStaticMethod(manager, "init");
         } catch (IllegalAccessException | NoSuchMethodException e) {
             WynntilsMod.warn("Misconfigured init() on manager " + manager, e);
-            assert false;
```
magicus, line 147: Why are you changing this? This was *explicitly* introduced to catch code errors with assert on.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -57,105 +56,88 @@ public static void init() {
...
+            return false;
         } catch (InvocationTargetException e) {
-            WynntilsMod.error("Exception during init of manager " + manager, e.getTargetException());
+            WynntilsMod.warn("Exception during init of manager " + manager, e.getTargetException());
```
magicus, line 118: And why are you downgrading this error to a warning? :-(

@Incompleteusern, you **need** to learn more discipline when coding. You are constantly changing a lot of side stuff that is not relevant to the issue you are trying to solve. Often these changes are controversial or misguided. When reviewing your PRs, I feel I spend more time just on the look for this kind of "what the heck is happening here" than trying to understand your *intent* and if the way you have written the patch is a proper way to solve the stated goal.

Please, please try to keep to the rule "one issue, one PR". In the PR, start by explaining what is the issue you are trying to resolve.

At times, it might be necessary to have a "collect a bunch of small fixes"-PR, since it might be too much overhead to have a single PR to e.g. demote this error to a warning. But then such a PR should not do anything else than being a bunch of minor cleanups, and it needs to be reviewed differently.

Fixing a typo or an incorrect comment that is not relevant to the PR *might* be okay to include, but even that is pushing it. (I will accept it though.) But not this kinds of changes, which seem innocuous but radically changes behavior. 

I've been trying to guide you before in trying not to mix issues, but apparently I need to be much more clear. This is not a good way of software development. It makes reviewing hard. It increases the risk that bugs are slipping into the code. And it makes it hard to understand the history of the code base. And focusing on one issue at the time forces the developer to fully engage in that particular problem.

There just is no upside to mixing concerns like this in a single PR.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -57,105 +56,88 @@ public static void init() {
...
     }
 
-    public static void removeDependency(Feature dependant, Class<? extends Model> dependency) {
-        if (PERSISTENT_CORE_MANAGERS.contains(dependency)) {
+    public static boolean removeDependency(ModelDependant dependant, Class<? extends Model> manager) {
+        if (PERSISTENT_CORE_MANAGERS.contains(manager)) {
             throw new IllegalStateException("Tried to unregister a persistent manager.");
         }
 
-        MODEL_DEPENDENCIES.putIfAbsent(dependency, new ArrayList<>());
```
magicus, line 99: What was wrong with putIfAbsent?

magicus:
+1 for making Feature ModelDependent. But please do this all the way, that is, remove the way dependencies are "appended" in init(). Instead getModelDependencies() should return an empty set in Feature, and be overridden to return a static final set of dependencies for features with dependencies.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
Incompleteusern: `computeIfAbsent` removes the get call, and in my opinion makes the logic more sound

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
Incompleteusern: I apologize for the rather severe mixing of concerns in this PR.

This pull request started as a goal to revamp `ManagerRegistry` through the desired merging dependencies under `ModelDependent`. However, this lead to attempted cleaning up of the logic, and then the idea of, instead of erroring out, failling more gracefully through returning a boolean.

I will close this pr and split it into multiple sub pull requests.

PR #460, opened 2022-09-27
Add invisible player transparency setting

Incompleteusern:
If the player is invisible due to something such vanish, they can be rendered translucently instead


magicus:
That sounds like a cheat. Is this available in Legacy? You need to verify with Wynncraft if that is really allowed.

magicus (changes requested):
Do not integrate until the permissibility is sorted out.

Incompleteusern:
> That sounds like a cheat. Is this available in Legacy? You need to verify with Wynncraft if that is really allowed.

It's only the actual player, and not any other players/party members, which should be acceptable. Will ask nonetheless.

magicus:
It changes the cost of doing vanish significantly.

Incompleteusern:
> It changes the cost of doing vanish significantly.

I've asked in the channel

Incompleteusern:
Apparently this is already in the game

PR #461, opened 2022-09-27
Clean up Feature

Incompleteusern:
This pr attempts to clean up `Feature` by removing methods when possible, which includes combining methods into `Configurable` from both `Feature` and `Overlay`


common/src/main/java/com/wynntils/core/features/Configurable.java
```diff
@@ -5,10 +5,34 @@
...
 
 import com.wynntils.core.config.ConfigHolder;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
-public interface Configurable {
-    void updateConfigOption(ConfigHolder configHolder);
+public abstract class Configurable {
```
magicus: No, don't turn this into a class. We might want to make other things configurable.

magicus: Make a common abstract base class `AbstractConfigurable` or whatever.

Incompleteusern: I think its fine for now, as that's partially future thinking. Also, off the top of my head, I don't see a use case in which we need to extend configurable along with another class, but please prove me wrong.

magicus: I disagree. There are design reasons to having an interface even if there is only a single implementation.

I want to hear @kristofbolyai's opinion on this as well. It's his design, after all.

At the very least, if you two agree on keeping this abstract class redesign, you must rename it. The `*able` naming is used in Java to signify an interface.

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -129,28 +124,13 @@ protected String getNameCamelCase() {
...
-     */
-    protected boolean onEnable() {
-        return true;
-    }
-
-    /** Called on disabling of Feature */
-    protected void onDisable() {}
-
     /** Called to activate a feature */
     public final void enable() {
```
magicus, line 135: The old code implied that it was a user responsibility to to check that you did not enable an already enabled feature. What is the rationale for changing this part of the interface?

Also, with the new state, if you call enable() on an unitialized feature, the result will be a no-op. Is this what you want?

Incompleteusern: The rationale is that most of the time we were circumventing responsibility by using `tryEnable` or `tryDisable`, effectively making the code moot. Yes, that is intended

common/src/main/java/com/wynntils/core/features/UserFeature.java
```diff
@@ -20,7 +22,6 @@ public final void updateConfigOption(ConfigHolder configHolder) {
         // if user toggle was changed, enable/disable feature accordingly
         if (configHolder.getFieldName().equals("userEnabled")) {
             // If this was changed before init, do not try to toggle
```
Incompleteusern: TODO: Update this comment to reflect uninitalized state, and try to make it more obvious

magicus:
I'm all for cleaning up Features. They have grown a bit wild. But I think we need to start by stepping back and asking ourself what the goal is. You have identified two areas which you address in this PR:

1) duplication of configuration
2) feature state (uninitialized, enabled, disabled) can be simplified.

and a third, addressed separately:

3) features should implement ModelDependant

I think it is better if this PR too is split into two, so we can discuss each change separately.

For 1) I agree that we can have a common ancestor for the Configurables, but not that we drop the interface. 

For 2), I think there's much more to discuss. There's a very fundamental issue with how enabled/disabled is handled, namely that it is split between FeatureRegistry and the features themselves. We need to go deeper down in this. Maybe a feature has not three states, but four? REGISTERED, UNINITIALIZED, ENABLED, DISABLED and it is just the FeatureRegistry (perhaps renamed to FeatureManager) that decides that?

I ran into problem with the "emergency shutdown" for disabling features when a feature throws an exception. I wanted to just disable the feature "for now", but for that to work, I had to change the "userEnabled" config. The interplay here is not clear and perhaps not even correct. I think we need to separate like "user intent to enable" and "enabled in effect". I assume this was the idea behind userEnabled, but the entire flow is muddy.

common/src/main/java/com/wynntils/core/features/Configurable.java
kristofbolyai: I can't think of a use case where we would need to extend from this and some other class, right now. But that does not mean I think we should do this. I think this seems reasonable, but the only gain is that we can remove around 30 lines of code. I am not sure that is worth it.

kristofbolyai:
Also, I have to do some Configurable interface changes for my PR... I am sorry, but it happened so I need to do this to get configs working..

common/src/main/java/com/wynntils/core/features/Configurable.java
kristofbolyai: > Make a common abstract base class `AbstractConfigurable` or whatever.

Do this. This is what MC does with widgets for example. You have the widget interface, and `AbstractWidget` with some basic impl. This way, we get to keep the good of both worlds.

common/src/main/java/com/wynntils/core/features/Feature.java
magicus: Yes, but "most of the time" is not really an answer. My interpretation of this code is that we needed *some* way to get here *without* the checking done by `tryToEnable()`. So the interesting question is: who were calling `enable()` directly, without going through `tryToEnable`? It could be that it was a use case that has disappeared, so *all* calls go via tryToEnable. If that is the case, then your change is perfectly fine. 

But due to all the shuffling of logic, I cannot see now if there were some places where `enable` was called directly before, and what is happening with them. They will change behavior, and that might need discussion or argument about why the new behavior is correct.

Incompleteusern: `enable` and `disable` are only called by `WynntilsCommand`'s `reload`, which not only is unaffected by `tryToEnable` vs `enable`, but in my opinion outdated.

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -92,8 +92,7 @@ private static void handleExceptionInEventListener(Throwable t) {
                         + "' has crashed and will be disabled")
                 .withStyle(ChatFormatting.RED));
 
-        feature.setUserEnabled(false);
-        feature.tryUserToggle();
+        feature.disable();
```
Incompleteusern: I don't what is wrong here as of now, but @magicus likely has a deeper, correct view

magicus: What do you mean is wrong? As in it does not work as expected, or as in the code does not look like you think?

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -32,7 +32,7 @@ public class ConfigHolder {
     private boolean userEdited = false;
 
     public ConfigHolder(
-            Configurable parent, Field field, FeatureCategory category, Config metadata, Type typeOverride) {
+            AbstractConfigurable parent, Field field, FeatureCategory category, Config metadata, Type typeOverride) {
```
magicus: The point in having a separation of interface and abstract base class is that in all cases except when extending the abstract base class, the functionality should be referred to using the interface.

common/src/main/java/com/wynntils/core/features/Configurable.java
```diff
@@ -6,11 +6,22 @@
...
 import java.util.List;
+import java.util.Optional;
 
 public interface Configurable {
     void updateConfigOption(ConfigHolder configHolder);
 
+    /** Registers the configurable's config options. Called by ConfigManager when loaded */
     void addConfigOptions(List<ConfigHolder> options);
 
+    /** Returns all configurable options registered that should be visible to the user */
```
magicus, line 17: Here we can actually see another benefit of doing interface/abstract base class separation. Now it is clear that to be able to implement Configurable, all these other methods are needed as well. This was not obvious before. Nice!

common/src/main/java/com/wynntils/core/features/UserFeature.java
```diff
@@ -33,13 +33,20 @@ public final void updateConfigOption(ConfigHolder configHolder) {
...
+            enable();
         } else {
-            tryDisable();
+            disable();
         }
     }
 
     public void setUserEnabled(boolean newState) {
-        this.userEnabled = newState;
+        Optional<ConfigHolder> opt = getConfigOptionFromString("userEnabled");
```
magicus: What is the gain of looking up the config instead of using the field directly? I'm really confused here.

common/src/main/java/com/wynntils/core/features/UserFeature.java
```diff
@@ -33,13 +33,20 @@ public final void updateConfigOption(ConfigHolder configHolder) {
     /** Updates the feature's enabled/disabled state to match the user's setting, if necessary */
     public final void tryUserToggle() {
```
magicus, line 32: Why do we even have this function? It just seems so messy, all this enable/disable stuff.

magicus:
I think we need to rethink how we enable/disable features. I still think the problem lies with the fact that it is not the role of features to know if they are enabled or disabled, but the role of the FeatureManager. If we move the userEnabled config from the feature itself to the FeatureManager, I think a lot will be clearer. 

Also, I think we should have a conjunction between two reasons to actually enable a feature -- i.e. a systemEnabled (or "managerEnabled") and userEnabled, and only if both are true should the feature *actually* be enabled. That means that for stuff like lootrun, we *only* have the systemEnabled functionality (since it is not a user controllable feature, and this "userEnabled" is implicitly always true). And for normal user features we can let the user turn them off by disabling userEnabled. But they can never be turned on solely by the user, if they are also not systemEnabled. This can happen e.g. if a feature crashes.

So there should be no Feature.enable(), but instead a FeatureManager.setSystemEnabled(Feature) and setUserEnabled(Feature). 

And there need to be a Feature.onEnable() so the feature can have the chance to do additional work when it is enabled (as in *actually* enabled, since both userEnabled and systemEnabled is true).

Or something like that.

As I said before, I think it is better if you use this PR to focus on Configurable changes, and do not mess with enable/disable at the same time. It is a muuuch more complex issue.

common/src/main/java/com/wynntils/core/features/UserFeature.java
Incompleteusern: It actually saves, thought this had unintended side affects, so this will be reverted soon

kristofbolyai: If you only set the field, you won't call the onConfigUpdate with the userEnabled config

common/src/main/java/com/wynntils/core/features/UserFeature.java
Incompleteusern: Dealing with enabling is best left for a separate pr, and I slightly agree with your statement.

common/src/main/java/com/wynntils/core/WynntilsMod.java
Incompleteusern: It has been reverted

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -117,36 +113,21 @@ public String getShortName() {
...
-        if (enabled) throw new IllegalStateException("Feature can not be enabled as it already is enabled");
+        if (state != FeatureState.DISABLED) return;
```
kristofbolyai, line 136: We kinda lost the exception here. Why?

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -161,11 +142,9 @@ public final void enable() {
 
     /** Called for a feature's deactivation */
     public final void disable() {
-        if (!enabled) throw new IllegalStateException("Feature can not be disabled as it already is disabled");
-
-        onDisable();
+        if (state != FeatureState.ENABLED) return;
```
kristofbolyai, line 156: same, lost exception

common/src/main/java/com/wynntils/core/features/Feature.java
Incompleteusern: Merged with tryEnable

common/src/main/java/com/wynntils/core/features/AbstractConfigurable.java
```diff
@@ -0,0 +1,44 @@
...
+package com.wynntils.core.features;
+
+import com.google.common.base.CaseFormat;
+import com.wynntils.core.config.ConfigHolder;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public abstract class AbstractConfigurable implements Configurable {
```
kristofbolyai, line 14: Why is `getConfigOptions` (unfiltered version) missing? The interface should have it too.

kristofbolyai (commented):
Fix the interface, fix the merge conflicts, and then this is good to go. Though, make sure to rename the PR "Add Configurable Interface", so we know what happened here (when the upcoming "clean up feature #2" comes, which would be smth like "feature state management refactor").

kristofbolyai:
Pinging @magicus to see this PR. Do you have any other comments beside what I listed?

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -50,7 +50,8 @@ public static void registerFeature(Feature feature) {
         registerConfigOptions(feature, featureConfigOptions);
     }
 
-    private static void registerConfigOptions(Configurable configurable, List<ConfigHolder> featureConfigOptions) {
+    private static void registerConfigOptions(
+            AbstractConfigurable configurable, List<ConfigHolder> featureConfigOptions) {
```
magicus:
```suggestion
            Configurable configurable, List<ConfigHolder> featureConfigOptions) {
```

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -183,7 +184,7 @@ private static void loadFeatureOverlayConfigOptions(FeatureCategory category, Fe
         }
     }
 
-    private static List<ConfigHolder> getConfigOptions(FeatureCategory category, Configurable parent) {
+    private static List<ConfigHolder> getConfigOptions(FeatureCategory category, AbstractConfigurable parent) {
```
magicus:
```suggestion
    private static List<ConfigHolder> getConfigOptions(FeatureCategory category, Configurable parent) {
```

The `AbstractConfigurable` class should like only be used when declaring the superclass of the configurable subclasses. Elsewhere, the interface should be used.

magicus (commented):
Apart from the changes to the method signatures, this looks good.

common/src/main/java/com/wynntils/core/features/AbstractConfigurable.java
Incompleteusern: It is not missing?

kristofbolyai: I mean we should add it. I know the interface does not have the method too.

kristofbolyai:
Merge is still bad

common/src/main/java/com/wynntils/core/features/AbstractConfigurable.java
Incompleteusern:
```java
    /** Returns all configurable options  that should be visible to the user */
    List<ConfigHolder> getConfigOptions();
```
in `Configurable`?

kristofbolyai: I am blind.

kristofbolyai (dismissed):

kristofbolyai (approved):

PR #468, opened 2022-09-27
Overlays: Add info boxes

kristofbolyai:
Supersed of #279.
image
image
image


common/src/main/java/com/wynntils/functions/MinecraftFunctions.java
```diff
@@ -12,28 +12,28 @@ public class MinecraftFunctions {
     public static class XFunction extends Function<Integer> {
         @Override
         public Integer getValue(String argument) {
-            return McUtils.player().getBlockX();
+            return McUtils.player() == null ? null : McUtils.player().getBlockX();
```
magicus: How can McUtils.player() be null? I thought it was set as soon as you logged on to the server.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -87,7 +87,13 @@
   "feature.wynntils.healthPotionBlocker.threshold.description": "The percentage of your max health that you need to have for the feature to activate",
   "feature.wynntils.healthPotionBlocker.threshold.name": "Threshold (%%)",
   "feature.wynntils.healthPotionBlocker.thresholdReached": "You are already at %s%% of your max health!",
-  "feature.wynntils.hidePotionGlint.name": "Hide Potion Glint",
+  "feature.wynntils.hidePotionGlint.name": "Hide Potion Glint Feature",
```
magicus: This seem like a merge error, old code reappearing.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -87,7 +87,13 @@
...
+  "feature.wynntils.infoBoxes.name": "Info Boxes",
+  "feature.wynntils.infoBoxes.overlay.infoBox.content.description": "What should be displayed in the info box?",
+  "feature.wynntils.infoBoxes.overlay.infoBox.content.name": "Info Box Content",
+  "feature.wynntils.infoBoxes.overlay.infoBox.name": "Info Box %d",
+  "feature.wynntils.infoBoxes.overlay.infoBox.textShadow.description": "Should the overlay for teleport scrolls be drawn with a shadow?",
```
magicus: Teleport scrolls FTW! 😆

common/src/main/java/com/wynntils/functions/MinecraftFunctions.java
kristofbolyai: For some reason, this was called on feature init. As I am writing this, because this is what the render preview used to be. And it used to be a static field..., so this was needed to be calculated. I fixed that so this is not needed anymore technically. I am not sure if there is value to keeping this?

kristofbolyai: (Also, functions do not fail gracefully :))

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+import com.wynntils.wynn.utils.WynnUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.resources.language.I18n;
+
+@FeatureInfo(category = FeatureCategory.OVERLAYS)
+public class InfoBoxesFeature extends UserFeature {
```
magicus: Just a bit on the naming. I understand that `InfoBoxesFeature` is probably more correct than `InfoBoxFeature`, but the latter sounds much better. And I don't think we've had plural for other features either, like "BossBars"?

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final Overlay infoBox1Overlay = new InfoBoxOverlay(1);
```
magicus: Oh, it's still hard-coded. :( I'd hoped you would have made this more general. Oh well, that's for another time, I guess.

kristofbolyai: This would require GUI work, and overlay systematical changes.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+            }
+
+            for (Function<?> oldDependency : functionDependencies) {
+                FunctionManager.disableFunction(oldDependency);
+            }
+
+            functionDependencies.clear();
+
+            StringBuilder builder = new StringBuilder(renderableText.length() + 10);
+            Matcher m = INFO_VARIABLE_PATTERN.matcher(renderableText);
```
magicus: There is a `FunctionManager.doFormat()` which uses the new `{function}` syntax. But I see now that I never finished `getStringFromTemplate()`, which is that you would have needed here.

I am not entirely sure I want to import the old `%info` syntax. We can possibly do this short-term if you really want to push this now, as long as we agree that this is just temporary.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+                    // \escape
+                    replacement = doEscapeFormat(m.group(2));
+                }
+                if (replacement == null) {
+                    replacement = m.group(0);
+                }
+
+                m.appendReplacement(builder, replacement);
+            }
+            m.appendTail(builder);
```
magicus: In any case, if you do keep this functionality, it should move to FunctionManager, to something like `getStringFromLegacyTemplate()`...

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
kristofbolyai: I did not even know that was a thing. This is mostly code I took from DonkeyBlaster, and made it work. Yes, I remember all the new things you want to do with function variables, but most of that code right now is TODO. So, until that is changed, we should push this.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+            return new TextRenderTask(
+                    parseColorCodes(builder.toString()),
+                    TextRenderSetting.DEFAULT
+                            .withHorizontalAlignment(this.getRenderHorizontalAlignment())
+                            .withMaxWidth(this.getWidth())
+                            .withCustomColor(CustomColor.fromChatFormatting(ChatFormatting.WHITE))
+                            .withTextShadow(textShadow));
+        }
+
+        private String parseColorCodes(String toProcess) {
```
magicus: This too, if kept, should move out of the Feature itself. Either to FunctionManager (if we think this is strictly something we will do for function templates), or to like StringUtils or ComponentUtils (which include support for coded strings as well)

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
```diff
@@ -0,0 +1,218 @@
...
+                            this.getRenderVerticalAlignment());
+        }
+
+        @Override
+        public void renderPreview(PoseStack poseStack, float partialTicks, Window window) {
+            if (!WynnUtils.onWorld()) return;
+
+            // FIXME: We do re-calculate this on render, but this is preview only, and fixing this would need a lot of
+            // architectural changes at the moment
+            TextRenderTask toRenderPreview = recalculateFunctions("&cX: %x%, &9Y: %y%, &aZ: %z%");
```
magicus: I don't get this. How does the normal toRender get updated? Only on config changes? Even if it does not need recalculating on every rendered frame, it do need some periodic update.. Or am I missing something?

kristofbolyai: Wait..., I need to do this in toRender too otherwise we are getting old data. Well..

common/src/main/java/com/wynntils/functions/MinecraftFunctions.java
magicus: I thought I actually wrote a rationale about skipping functions: They are so simple I did not expect them to fail, and if they do, the failure will propagate to the `/function` command and/or InfoBoxOverlay, which will fail gracefully.

But it might be a point in disabling individual functions. If one function is behaving bad, we do not have to bring down the entire InfoBoxOverlay. I'll fix that.

kristofbolyai: This case was more of bad use of functions than them being bad, but I think there is still value to disabling functions.

magicus: And no, if it's not needed we should not keep it. I've been trying to kill code that does null checks for player() elsewhere. If we have such code, people start thinking it's needed and add it in some places but not other. We must be able to trust this value, period.

Also, functions should perhaps be blocked by the FunctionManager if we are not connected to a wynncraft server, just like events are not posted. That way we can definitely not do anything wrong with this.

kristofbolyai: Reverted this

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxesFeature.java
magicus: May I suggest instead doing the calculation in a separate function that listens to ClientTickEvent, and cache it as a pre-calculated string ever so often?

And also perhaps expose a Config that says how often values should be recalculated, either in terms of ticks or expressed in seconds (the latter is perhaps easier for users, but requires you to convert it into ticks). I think a good default value is about 4 times per second, i.e. after 5 ticks.

magicus: Most functions are relatively cheap, but some iterate through all inventory parsing lore, and that's no good to make for each frame we're drawing.

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxFeature.java
```diff
@@ -0,0 +1,159 @@
...
+                    HorizontalAlignment.Left,
+                    VerticalAlignment.Middle);
+            this.id = id;
+        }
+
+        @Override
+        public void render(PoseStack poseStack, float partialTicks, Window window) {
+            if (!WynnUtils.onWorld()) return;
+
+            if (System.nanoTime() - lastUpdate > secondsPerRecalculation * 1e+9) {
```
kristofbolyai, line 96: @magicus I hope you do not mind this. Overlays are not registered to the event bus... This is a good enough replacement IMO.

magicus: Yes, that is perfectly fine, perhaps even better than my suggestion. (Otherwise, if we really need events, let the owning Feature listen for them. I've done so in one or two places).

magicus: Maybe do the same to preview?

kristofbolyai: Yeah, but ticking 6 different overlays is pretty messy from a feature event.

kristofbolyai: I think it is fine for previews to be updated every time, for now. If there comes some really heavy variable, we can update it. (also, the active function by nature uses some caching to prevent per render call calculation)

magicus: Ok, fine. (We should perhaps also let the preview use hard-coded functions from us, since if the user has weird strings it might now show up properly. But that's a future improvement, if we should do it at all.)

magicus (approved):

common/src/main/java/com/wynntils/features/user/overlays/InfoBoxFeature.java
kristofbolyai: > Ok, fine. (We should perhaps also let the preview use hard-coded functions from us, since if the user has weird strings it might now show up properly. But that's a future improvement, if we should do it at all.)

It does have a hard coded value?

magicus: Oh, I missed that.

PR #469, opened 2022-09-27
Gui fixes

### LINKED ISSUE #466: Overlay Management Screen overlay name rendering is incorrect
image

This is aligned Left, and Middle. This is obviously incorrect.

### LINKED ISSUE #467: Text box repeated backspace clears every text box

kristofbolyai:
Fixes #467. Fixes #466. Also fixes not being able to set alignment overrides in the GUI.

### DIFF
--- a/common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
+++ b/common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
@@ -99,10 +99,16 @@ public abstract class Overlay implements Translatable, Configurable, Comparable<
     }
 
     /** Returns the config option matching the given name, if it exists */
-    public final Optional<ConfigHolder> getConfigOptionFromString(String name) {
-        return getVisibleConfigOptions().stream()
-                .filter(c -> c.getFieldName().equals(name))
-                .findFirst();
+    public final Optional<ConfigHolder> getConfigOptionFromString(String name, boolean visibleOnly) {
+        if (visibleOnly) {
+            return getVisibleConfigOptions().stream()
+                    .filter(c -> c.getFieldName().equals(name))
+                    .findFirst();
+        } else {
+            return configOptions.stream()
+                    .filter(c -> c.getFieldName().equals(name))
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -99,10 +99,16 @@ public final List<ConfigHolder> getVisibleConfigOptions() {
     }
 
     /** Returns the config option matching the given name, if it exists */
-    public final Optional<ConfigHolder> getConfigOptionFromString(String name) {
-        return getVisibleConfigOptions().stream()
-                .filter(c -> c.getFieldName().equals(name))
-                .findFirst();
+    public final Optional<ConfigHolder> getConfigOptionFromString(String name, boolean visibleOnly) {
+        if (visibleOnly) {
```
magicus: Why do you need both cases, and not just the configOptions.stream?

kristofbolyai: When I was coding this, premature optimization.. I thought that let's loop over 6 elements, not 9...
I am not sure, perhaps there is some value? Maybe it is fine to remove that argument. What do you think?

kristofbolyai: Or are you suggesting that I should filter the stream if the boolean is set, and not call the other function?

magicus: It sounds like premature optimization, which complicates the code. Is this looped over in some very hot loop, like in rendering? (That would in itself likely be a sign of problem anyway) Otherwise it doesn't matter, but it takes a clear function and makes it more "what is this?". (Also, boolean as arguments are notoriously tricky to understand when reading code).

kristofbolyai:
New fix: Map auto close when key is not held anymore is fixed.

magicus (approved):
Great, much better! Thanks.

PR #472, opened 2022-09-27
Merge dependencies under `ModelDependent`

Incompleteusern:
This pr attempts to generalize `ManagerRegistry` to, instead of splitting between `DependantFunction` and `Feature`, to instead handle both under the common superinterface of `ModelDependant`. 

As such, this pr also implements `ModelDependant` for the `Feature` class by replacing dependencies from a list to a function


magicus (dismissed):
Much better! Thank you! The separate, clearly descriptive commits also helped in reviewing this. ❤️

Incompleteusern:
Unfortunately, I've still left a few bugs in there when requesting a reviewing, so I will try to fix that. Something to improve upon in the future

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -173,7 +171,7 @@ public final void disable() {
 
         enabled = false;
 
-        ManagerRegistry.removeAllFeatureDependency(this);
+        ManagerRegistry.removeAllDependency(this);
```
magicus: The grammar does not work out. Please rename to `removeAllDependencies` (I know this is a pre-existing problem, but since you're touching it anyway...)

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -103,7 +89,7 @@ public static void removeDependency(Feature dependant, Class<? extends Model> de
         updateManagerState(dependency);
     }
 
-    public static void removeAllFeatureDependency(Feature dependant) {
+    public static void removeAllDependency(ModelDependant dependant) {
         for (Class<? extends Manager> manager : MODEL_DEPENDENCIES.keySet()) {
```
magicus: This is not correct. The class should be of Model, and the name should reflect that.

Edit: sorry, I wrote the wrong class first.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -114,11 +100,9 @@ public static void removeAllFeatureDependency(Feature dependant) {
     }
 
     private static void updateManagerState(Class<? extends Manager> manager) {
```
magicus: same here. In fact, maybe you should search for "manager" (and "function") in this file and verify that there are no remnants.

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
Incompleteusern: I was going to update this method as well in this pr.

magicus (approved):
It looks okay. Does it seem to work okay?

PR #474, opened 2022-09-28
Clean up ManagerRegistry Logic

Incompleteusern:
This PR does the following:
- https://github.com/Wynntils/Artemis/commit/7eb4ab40a3e62c48b480a12cbc586b9445e70253 and https://github.com/Wynntils/Artemis/commit/07f43dacf907fefecbcd9cecafb5b09d8959b564 abolish `updateManagerState` and split its two cases of enabling and disabling into the distinct methods that enable and disable.
    -  These commits also moves registration and unregistration of event listeners into `tryInitManager` and `tryDisableManager` respectively.
    - These commits also introduce`addAllDependencies` and simplify `removeAllDependencies` for bulk dependency addition and removal
- https://github.com/Wynntils/Artemis/commit/f00cb7d17eaccce58c700c0106d13402d34a0685 weakens collection type from `Set` to `Collection`, as suggested by IDE
- https://github.com/Wynntils/Artemis/commit/3911a78c28a2a388cb3a676c273908f34f5f1373 is a follow up that removes a unnecessary registration call for core managers after the first commit
- https://github.com/Wynntils/Artemis/commit/5a733308e77a3a3ed053d883ff600703b5400774 implements `addAllDependencies` into `Feature` and `FunctionRegistry`
- https://github.com/Wynntils/Artemis/pull/474/commits/5e471da2ef0c0d09bfce4daf73b085a6aa0c08ec removes an `assert false` in favor of `WynntilsMod.error`


common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
```diff
@@ -59,81 +59,79 @@ public static void init() {
...
 
     private static void tryInitManager(Class<? extends Manager> manager) {
+        WynntilsMod.registerEventListener(manager);
+
         try {
             MethodUtils.invokeExactStaticMethod(manager, "init");
         } catch (IllegalAccessException | NoSuchMethodException e) {
-            WynntilsMod.warn("Misconfigured init() on manager " + manager, e);
-            assert false;
+            WynntilsMod.error("Misconfigured init() on manager " + manager, e);
```
magicus, line 126: Throw a runtime exception instead. We need to catch this design errors during development.

I'll review the rest of the PR later.

Incompleteusern: Th other branch of the try-catch does not do that, should this also be changed?

magicus: Look at the exceptions, and think of what causes them to be thrown, and why they differ. What can possibly cause any of the two first exceptions? And what can cause the last?

Incompleteusern: Technically the second is more of a "compile-time" problem

magicus: Rather the first two, right? `IllegalAccessException | NoSuchMethodException e` happens if we have not declared a public init(). This problem should be detected immediately during development, and will never happen at a user installation.

The last `InvocationTargetException e` means that the init() method threw an exception while it executed. This can be anything, and can definitely happen in a user installation. The intention here is apparently to hope that the manager might function anyway; or at least, that we should not end the entire mod because of this, so we log the error and keep running.

If we ever build a system to disable core managers piecewise, we should probably turn this into a signal that the manager needs to be disabled.

magicus: Anyway, I won't have time to do much more reviewing of all this refactoring PRs until Fridays at earliest, sorry 'bout that.

kristofbolyai (dismissed):
Seems fine to me, let's wait for magicus though.

Incompleteusern:
He is busy till friday :), so I will work on other things (Trying to make a modern gui for fun)

common/src/main/java/com/wynntils/core/managers/ManagerRegistry.java
magicus: Please either reinstate the assert, or change this into throwing a RuntimeException. Otherwise it is a regression in our ability to find structural code errors during development.

magicus (changes requested):
There's just the one thing to fix, then you're good to go.

magicus (approved):

PR #476, opened 2022-09-28
Screens: Add Custom character GUI

kristofbolyai:
image


Incompleteusern (dismissed):

common/src/main/java/com/wynntils/wynn/objects/ClassInfo.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.objects;
+
+import net.minecraft.world.item.ItemStack;
+
+public record ClassInfo(
```
magicus, line 9: I think this overlaps quite a bit with `CharacterInfo` in `CharacterManager`. Can we unify them?

magicus: Also, I think we should stick to the name "Character" for selecting the character to play, not "class" (and keep that strictly as a name for the kind of "style" the character is.) Even Wynncraft themselves have renamed the dialog to "Select a Character", not "Select a Class". (Still the command is `/class` but I assume old habits die hard.)

kristofbolyai: Does it? They are used for different things, parsed differently. We parse different things for both of them because the other info is not needed.

kristofbolyai: I am fine with renaming though.. But then, it kinda conflicts.. And I am not sure I like them merged

common/src/main/java/com/wynntils/features/user/CustomCharacterSelectionScreenFeature.java
```diff
@@ -0,0 +1,43 @@
...
+import com.wynntils.mc.event.ScreenOpenedEvent;
+import com.wynntils.mc.utils.ComponentUtils;
+import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wynn.event.WorldStateEvent;
+import com.wynntils.wynn.model.WorldStateManager;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class CustomCharacterSelectionScreenFeature extends UserFeature {
+    @Config
+    public boolean onlyOpenOnce = true;
```
magicus, line 20: How does this work? I thought this was the reason you do the openedInThisCharacterSelectionState dance, but you never check this?

kristofbolyai: Oops.

common/src/main/java/com/wynntils/gui/screens/CharacterSelectorScreen.java
```diff
@@ -0,0 +1,471 @@
...
+        final int height = Math.round(Texture.LIST_BACKGROUND.height() * currentTextureScale / 8f);
+        for (int i = 0; i < classInfoList.size(); i++) {
+            ClassInfo classInfo = classInfoList.get(i);
+            classInfoButtons.add(new ClassInfoButton(
+                    5, (5 + ((i % CHARACTER_INFO_PER_PAGE) * height)), (int) width, height, classInfo, this));
+        }
+    }
+
+    private static ClassInfo getClassInfoFromItem(ItemStack item, int slot, String className) {
+        ClassType classType = null;
```
magicus: Regarding ClassInfo/CharacterInfo: I think this is very similar to `CharacterInfo.parseCharacter`. Granted, this extracts a bit more, but that should really be done for `parseCharacter` as well. In the future, we will need the XP and finished quests info, to help us confirm that we match to a specific character.

We're ending up with CLASS_ITEM_LEVEL_PATTERN and CLASS_MENU_LEVEL_PATTERN being virtually the same, etc. This is exactly the situation we had in Legacy, where we did not design a generic "model" API which could be used by all parts of the code needing it, so each and every place reinvented it themselves.

common/src/main/java/com/wynntils/wynn/objects/ClassInfo.java
magicus: The one thing that this class contains which I'm not 100% happy about putting in CharacterInfo is the slot id and itemstack. I understand why you need it here, but it should not really be exposed later on with the CharacterInfo.

common/src/main/java/com/wynntils/gui/screens/CharacterSelectorScreen.java
kristofbolyai: But then, where would we need the XP or quest info, when those can be updated mid gameplay, and we need to get it from somewhere else otherwise then?

kristofbolyai: I always wanted to have some generic regex class.

common/src/main/java/com/wynntils/wynn/objects/ClassInfo.java
magicus: If you are okay with it, I can accept the code as it is now, and then see if I can merge them in a good way, so that the slot is exposed only where it should be.

kristofbolyai: You could technically get the itemstack from the container with the slot id.. Not sure how to remove both.

common/src/main/java/com/wynntils/gui/screens/CharacterSelectorScreen.java
```diff
@@ -0,0 +1,471 @@
...
+        int scale = this.height / 5;
+        InventoryScreen.renderEntityInInventory(
+                (int) (this.width * 0.6f), (int) (this.height * 0.85f), scale, 0, 0, McUtils.player());
+    }
+
+    private void setScrollOffset(int delta) {
+        scrollOffset = MathUtils.clamp(scrollOffset - delta, 0, classInfoButtons.size() - CHARACTER_INFO_PER_PAGE);
+    }
+
+    private void reloadQuestInfoButtons() {
```
magicus: This name is a bit odd, yes?

common/src/main/java/com/wynntils/wynn/objects/ClassInfo.java
kristofbolyai: And sure, I don't really have an idea for refactoring this anyways

common/src/main/java/com/wynntils/gui/screens/CharacterSelectorScreen.java
magicus: So in my mind, the CharacterManager would provide us with e.g. "XP" from the best available source. Before selecting a character, that is the lore from the character selection screen. After logging in, it is the vanilla experience level. But the outwards interface is the same, just "getXp()".

magicus: > I always wanted to have some generic regex class.

That sounds dangerous. :-) Do you mean like a way to help us create regexp:s locally where they belong (if so, I think it's fine but probably a bit futile to get something that is genuinely better than plain regexps), or do you mean that you want to collect a lot of regexps from all over the place in a single file? If you mean the latter, I think you're just creating chaos. Instead, I think the mentality should be: Why do you need a regexp? To extract some "wynncrafty" thing, right? So then encapsulate the result of this regexp in a Model and present it in a "programmer friendly" way.

kristofbolyai: > > I always wanted to have some generic regex class.
> 
> That sounds dangerous. :-) Do you mean like a way to help us create regexp:s locally where they belong (if so, I think it's fine but probably a bit futile to get something that is genuinely better than plain regexps), or do you mean that you want to collect a lot of regexps from all over the place in a single file? If you mean the latter, I think you're just creating chaos. Instead, I think the mentality should be: Why do you need a regexp? To extract some "wynncrafty" thing, right? So then encapsulate the result of this regexp in a Model and present it in a "programmer friendly" way.

You are probably right, it is just an easy way out :)

magicus:
I think the GUI code has just a tad to much domain knowledge. What would you say about creating e.g. CharacterSelectionManager, and moving `onContainerItemsSet` and `getClassInfoFromItem` there? Also, the `onPress` methods of the widgets also has code that could move there. So e.g. `ClassSelectionDeleteButton.onPress` does: 
```

        ContainerUtils.clickOnSlot(
                characterSelectorScreen.getSelected().getClassInfo().slot(),
                menu.containerId,
                GLFW.GLFW_MOUSE_BUTTON_RIGHT,
                menu.getItems());
```
and this should be protected knowledge inside the model in e.g. `CharacterSelectionManager.deleteCharacter(slot, menu.containerId, menu.getItems())`. This helps with clarity in the widget (now is when the character is actually deleted), and it helps if anyone else is ever going to do something with the character selection interface, like a command `/character mage 105` which goes to the class menu and selects a character for you.

common/src/main/java/com/wynntils/wynn/objects/ClassInfo.java
magicus: Let's go with this for now, and I'll put it on my mental todo-list of things I want to fix later on :)

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -196,4 +197,9 @@ private void onDisconnectPre(Component reason, CallbackInfo ci) {
         // Unexpected disconnect
         EventFactory.onDisconnect();
     }
+
+    @Inject(method = "handleParticleEvent", at = @At("HEAD"), cancellable = true)
+    private void onParticle(ClientboundLevelParticlesPacket packet, CallbackInfo ci) {
```
magicus: I assume you're going to do this properly?

Are there some particles that is problematic in the character selection screen? Do Wynncraft generate particles in the container UI and it is showing through? (Just curious why this is needed)

magicus:
Otherwise, it looks much nicer with the new manager! Thank you!

common/src/main/java/com/wynntils/gui/widgets/ClassInfoButton.java
```diff
@@ -0,0 +1,117 @@
...
+                Texture.XP_BAR.height(),
+                classInfo.xp() / 100f);
+    }
+
+    @Override
+    public void onPress() {
+        if (characterSelectorScreen.getSelected() == this) {
+            AbstractContainerMenu menu =
+                    characterSelectorScreen.getActualClassSelectionScreen().getMenu();
+            ContainerUtils.clickOnSlot(
```
magicus: What does this do? And shouldn't it be moved to the manager?

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
kristofbolyai: Ohh lol, why did this commit. I was testing smth, Wynn has a kick where it constantly kicks people because of particle packets. I will revert this though...

common/src/main/java/com/wynntils/wynn/model/CharacterSelectionManager.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+public class CharacterSelectionManager extends CoreManager {
+    private static final Pattern NEW_CLASS_ITEM_NAME_PATTERN = Pattern.compile("§a\\[+\\] Create new character");
```
mbnkwm: This pattern is missing some parts

```suggestion
    private static final Pattern NEW_CLASS_ITEM_NAME_PATTERN = Pattern.compile("§l§a\\[\\+\\] Create a new character");
```

kristofbolyai: Thanks. I could never test this as I have max class count.

common/src/main/java/com/wynntils/gui/screens/CharacterSelectorScreen.java
```diff
@@ -0,0 +1,401 @@
...
+                poseStack,
+                Texture.CHALLENGES_ICON.resource(),
+                0,
+                0,
+                0,
+                Texture.CHALLENGES_ICON.width() * currentTextureScale,
+                Texture.CHALLENGES_ICON.height() * currentTextureScale,
+                Texture.CHALLENGES_ICON.width(),
+                Texture.CHALLENGES_ICON.height());
+        // FIXME: I have no idea what to render here..
```
magicus: ? Is something still needed, or can you remove this comment?

kristofbolyai: I am not sure what the medal thing stands for.. 
image

I will check legacy.

kristofbolyai: I updated to TODO. I cannot add/test this right now as I do not have free class slots on my main, and my alt is not leveled up.

magicus (approved):

PR #477, opened 2022-09-30
Circle Minimap

Incompleteusern:
image


common/src/main/java/com/wynntils/gui/render/Texture.java
```diff
@@ -50,10 +50,11 @@ public enum Texture {
...
     // Map related
-    GILDED_MAP_TEXTURES("map/gilded_map_textures.png", 262, 524),
-    PAPER_MAP_TEXTURES("map/paper_map_textures.png", 256, 512),
-    WYNN_MAP_TEXTURES("map/wynn_map_textures.png", 256, 256),
-    FULLSCREEN_MAP_BORDER("map/full_map_border.png", 510, 254),
+    GILDED_MAP_TEXTURES("map/map_border/gilded_map_textures.png", 262, 524),
+    PAPER_MAP_TEXTURES("map/map_border/paper_map_textures.png", 256, 512),
+    WYNN_MAP_TEXTURES("map/map_border/wynn_map_textures.png", 126, 256),
+    CIRCLE("map/map_border/circle.png", 256, 256),
+    FULLSCREEN_MAP_BORDER("map/map_border/full_map_border.png", 510, 254),
```
kristofbolyai, line 57: Can you order this whole map segment alphabetically? Everything else is ordered iirc

kristofbolyai (dismissed):
Looks good!

kristofbolyai:
Is the mask responsible for not rendering POIs outside the circle?

kristofbolyai:
image
Default positioning is not good. We should move the minimap a bit.

kristofbolyai:
> image Default positioning is not good. We should move the minimap a bit.

This would also fix cardinals rendering outside of the screen.

kristofbolyai:
Sorry for hijacking the last 2 commits. I wanted this out asap. I think both of that commits are reasonable.

common/src/main/java/com/wynntils/utils/MathUtils.java
```diff
@@ -68,11 +68,10 @@ public static int integerFromRoman(String numeral) {
...
-        float deltaA = fromB - fromA;
-        float deltaB = toB - toA;
-        float scale = deltaB / deltaA;
-        float negA = -1 * fromA;
-        float offset = (negA * scale) + toA;
-        return (sourceNumber * scale) + offset;
+        return MathUtils.lerp(toA, toB, MathUtils.inverseLerp(fromA, fromB, sourceNumber));
+    }
+
+    public static float mag(float x, float y) {
```
magicus, line 74: What is `mag`? This is a normal euclidean distance?

kristofbolyai: https://en.wikipedia.org/wiki/Magnitude_(mathematics)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -315,7 +316,7 @@
   "feature.wynntils.miniMap.overlay.miniMap.followPlayerRotation.name": "Follow Player Rotation",
   "feature.wynntils.miniMap.overlay.miniMap.maskType.description": "Should the minimap be a rectangle or a circle? (Circle is unsupported as of now)?",
   "feature.wynntils.miniMap.overlay.miniMap.maskType.name": "Map Mask Type",
-  "feature.wynntils.miniMap.overlay.miniMap.name": "Mini Map Overlay",
+  "feature.wynntils.miniMap.overlay.miniMap.name": "Minimap",
   "feature.wynntils.miniMap.overlay.miniMap.poiScale.description": "How big should mini-map POIs be?",
```
magicus, line 320: We are obviously not very consistent here. "Minimap", "Mini Map" and "Mini-map" are all used.

Should we standardize on "Minimap" as a single word? If so, the class should be "Minimap" and the i18n keys should be `.minimap`, not `.miniMap`.

I'm all for a standard, I don't care which one *), but we should be consistent.

*) Except "mini-map", that's just ugly.

common/src/main/java/com/wynntils/gui/render/Texture.java
```diff
@@ -50,10 +50,11 @@ public enum Texture {
...
 
     // Map related
-    GILDED_MAP_TEXTURES("map/gilded_map_textures.png", 262, 524),
-    PAPER_MAP_TEXTURES("map/paper_map_textures.png", 256, 512),
-    WYNN_MAP_TEXTURES("map/wynn_map_textures.png", 256, 256),
-    FULLSCREEN_MAP_BORDER("map/full_map_border.png", 510, 254),
+    GILDED_MAP_TEXTURES("map/map_border/gilded_map_textures.png", 262, 524),
+    PAPER_MAP_TEXTURES("map/map_border/paper_map_textures.png", 256, 512),
+    WYNN_MAP_TEXTURES("map/map_border/wynn_map_textures.png", 126, 256),
+    CIRCLE("map/map_border/circle.png", 256, 256),
```
magicus, line 56: This CIRCLE is only used as a mask, right? I think the name (both of the enum and the file) should convey that.

common/src/main/java/com/wynntils/features/user/overlays/map/MiniMapFeature.java
```diff
@@ -108,9 +108,11 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
             // enable mask
             switch (maskType) {
                 case Rectangular -> RenderUtils.enableScissor((int) renderX, (int) renderY, (int) width, (int) height);
-                    // case Circle -> {
-                    // TODO
-                    // }
+                case Circle -> {
+                    RenderUtils.createMask(
```
magicus, line 112: Qodana is right. We don't need the `{ ... }`. Removing them will also fit with the style of the other switch branch.

common/src/main/java/com/wynntils/gui/render/Texture.java
magicus: @kristofbolyai That was never done before integration..?

common/src/main/java/com/wynntils/utils/MathUtils.java
magicus: Aha, mag as in magnitude. I was so sure it meant me. ;-)

Joking aside, I think `distance` is a much better name. Using "magnitude" in this sense is for me strictly reserved to complex numbers, and then the arguments would not have been named "x" and "y" (and we do not do any complex arithmetic).

Also, do we not already *have* a `distance` function?

common/src/main/java/com/wynntils/gui/render/Texture.java
kristofbolyai: Oh I forgot. We can just do it next time I touch the map

common/src/main/java/com/wynntils/utils/MathUtils.java
Incompleteusern: Magnitude is also used in the case of vectors.

common/src/main/resources/assets/wynntils/lang/en_us.json
Incompleteusern: I agree

common/src/main/java/com/wynntils/utils/MathUtils.java
Incompleteusern: Though I think there is already a method

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: Agree that we need a standard, or that it should be "Minimap" (and not "Mini Map"), or both of the above? :)

Incompleteusern: "Minimap" should be standard in my opinion

common/src/main/java/com/wynntils/features/user/overlays/map/MiniMapFeature.java
Incompleteusern: Best for a general Qodana/formatting pr

PR #483, opened 2022-09-30
Add /compass clear

### LINKED ISSUE #481: Add compass clear

magicus:
Fixes https://github.com/Wynntils/Artemis/issues/481


kristofbolyai (approved):
This seems fine. On the other hand, I've never understood why we need to mess with the level spawn point... This is how legacy did it, and I just ported it. Although this is not relevant to the PR..

magicus:
The messing with spawn point serves a purpose: the built-in compass points to the spawn point, so by setting it we get the additional help of the spawn point. (Nowadays Wynncraft does this themselves, I think, when you track a quest.)

I never use it since the minimap marker and the beacon beam beats the sh*t out of the tiny compass needle, but it might help some, and it looks a bit better to do it. :)

But we can't send in null, that causes a crash in Vanilla, so I had to fix that. (Perhaps this was possible in 1.12?)

PR #485, opened 2022-10-01
Add backend support for getting quests sorted by distance etc

magicus:
This is not used yet but will be in #367, and hopefully in the GUI as well :)


common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -300,11 +306,70 @@ public static List<QuestInfo> getQuests() {
...
+        return Optional.of(new Location(
+                Integer.parseInt(matcher.group(1)),
+                Integer.parseInt(matcher.group(2)),
+                Integer.parseInt(matcher.group(3))));
+    }
+
+    private static class LocationComparator implements Comparator<QuestInfo> {
+        private static final Vec3 PLAYER_LOCATION = McUtils.player().position();
+
+        private double getDistance(Optional<Location> loc) {
```
kristofbolyai, line 345: Is Qodana's suggestion relevant?

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
```diff
@@ -300,11 +306,70 @@ public static List<QuestInfo> getQuests() {
...
+        @Override
+        public int compare(QuestInfo quest1, QuestInfo quest2) {
+            Optional<Location> loc1 = quest1.getNextLocation();
+            Optional<Location> loc2 = quest2.getNextLocation();
+            return (int) (getDistance(loc1) - getDistance(loc2));
+        }
+    }
+
+    public enum QuestSortOrder {
+        NONE,
```
kristofbolyai, line 362: UNORDERED?

kristofbolyai (commented):
Looks good otherwise

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
kristofbolyai: Looking at this after a while, maybe NONE is better.

kristofbolyai:
If this get's merged, I think I can add the GUI part tomorrow.

common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java
magicus: Ah, found it:

>  'Optional' used as field or parameter type: common/src/main/java/com/wynntils/wynn/model/questbook/QuestBookManager.java#L345
`Optional` used as type for parameter 'loc'

magicus: (But not in the code review GUI, had to search in the GHA logs...)

No, I don't think that's relevant. I mean, it's kind of the point that we should take in an Optional and return 0 if it is empty. I get it why one should avoid Optional in fields, but I have a hard time understanding why they should not be passed as arguments. Seems quite valid to me.

kristofbolyai (approved):

PR #487, opened 2022-10-01
Support multiple map parts

magicus:


magicus:
Let me start by saying that this solution won't scale. We can't do `getMapForLocation()` each time a frame is rendered. However, it's a start. And I will do a follow-up where we cache the proper map part in the minimap/mapscreen, and update it when needed.

And finally, this is not the solution needed to split a map in multiple parts for tiled download.

kristofbolyai:
> And finally, this is not the solution needed to split a map in multiple parts for tiled download.

What else is needed (other than proper requests and caching)?

kristofbolyai:
I think I understand partly.. https://github.com/Wynntils/WynntilsWebsite-API/compare/40d9eb1832ae...9eb0db7b6426, const hashes.

Incompleteusern:
The reason for `maps.php` is so we don't need to store the md5 hash of the file and can just worry about the file.
This should change `maps.php` instead

kristofbolyai:
> The reason for `maps.php` is so we don't need to store the md5 hash of the file and can just worry about the file. This should change `maps.php` instead

Maps.php has coordinates hard coded. Perhaps make it so it processes the json and adds the hash values?

kristofbolyai:
https://github.com/Wynntils/WynntilsWebsite-API/commit/d53f8b19d036002439f83070aaf832ad4b844e62

Incompleteusern:
> > The reason for `maps.php` is so we don't need to store the md5 hash of the file and can just worry about the file. This should change `maps.php` instead
> 
> Maps.php has coordinates hard coded. Perhaps make it so it processes the json and adds the hash values?

Yes, exactly. `maps.php` should take in the json and add md5

Incompleteusern (changes requested):
Automatically calculate hashes and use `maps.php`

kristofbolyai:
> Automatically calculate hashes and use `maps.php`

I would not make that a blocking thing though.. I think it can be fine for a while. As said, this is not a permanent solution anyways as this cannot really scale..

Incompleteusern:
> > Automatically calculate hashes and use `maps.php`
> 
> I would not make that a blocking thing though.. I think it can be fine for a while. As said, this is not a permanent solution anyways as this cannot really scale..

I'd rather we get this resolved now than kicking it down the road.

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -33,31 +31,32 @@
...
     private static final String PLACES_JSON_URL =
             "https://raw.githubusercontent.com/Wynntils/Reference/main/locations/places.json";
     private static final String SERVICES_JSON_URL =
             "https://raw.githubusercontent.com/Wynntils/Reference/main/locations/services.json";
+    private static final String MAPS_JSON_URL =
+            "https://raw.githubusercontent.com/Wynntils/WynntilsWebsite-API/master/maps/maps.json";
+
     private static final Gson GSON = new GsonBuilder().create();
-    private static List<MapProfile> maps = new ArrayList<>();
+    private static List<MapTexture> maps = Collections.synchronizedList(new ArrayList());
```
kristofbolyai: Theoretically this is a perfect use case for `CopyOnWriteArrayList`?

kristofbolyai:
> > > Automatically calculate hashes and use `maps.php`
> > 
> > 
> > I would not make that a blocking thing though.. I think it can be fine for a while. As said, this is not a permanent solution anyways as this cannot really scale..
> 
> I'd rather we get this resolved now than kicking it down the road.

Then, can you help with that? I am not sure if magicus can/want to write PHP, and I can't either. Or do we need to ask Scyu?

Incompleteusern:
> > > > Automatically calculate hashes and use `maps.php`
> > > 
> > > 
> > > I would not make that a blocking thing though.. I think it can be fine for a while. As said, this is not a permanent solution anyways as this cannot really scale..
> > 
> > 
> > I'd rather we get this resolved now than kicking it down the road.
> 
> Then, can you help with that? I am not sure if magicus can/want to write PHP, and I can't either. Or do we need to ask Scyu?

We should probably ask Scyu

kristofbolyai (dismissed):
Code seems okay to me

Incompleteusern (dismissed):

magicus:
So, are we in agreement that we do not mix php in the process? We can change how the maps.json file is generated separately; the nice thing about having this as an API is that we separate production and consumption of that file and the png:s.

kristofbolyai:
> So, are we in agreement that we do not mix php in the process? We can change how the maps.json file is generated separately; the nice thing about having this as an API is that we separate production and consumption of that file and the png:s.

Agreed imo

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
magicus: "This is ordinarily too costly, but may be more efficient than alternatives when traversal operations vastly outnumber mutations, and is useful when you cannot or don't want to synchronize traversals, yet need to preclude interference among concurrent threads. "

I dunno. Maybe? If it makes you happy. :)

Incompleteusern:
I agree, I've also made a (breaking) pr https://github.com/Wynntils/WynntilsWebsite-API/pull/12 which should remoe the old implementation on the server side.

magicus:
> > And finally, this is not the solution needed to split a map in multiple parts for tiled download.
> 
> What else is needed (other than proper requests and caching)?

The ability to draw multiple maps/map parts at the same time. The current fix assumes each map renderer will only render a single map. So instead of "getMapForLocation(x,z)" you'd need a "getAllMapsThatFitInThisBoundingBox"...

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
kristofbolyai: In my mind, since the sync list is blocking and we render the minimap and the main map at the same time (if the main map screen is open), they would block each other in render. I am not sure about the impact though.

magicus: Nah, they won't block, rendering is single threaded. The synchronization is just for when loading the maps. But I'll change it, won't hurt.

kristofbolyai (approved):

PR #490, opened 2022-10-02
Support drawing multiple maps at the same time

magicus:
This will fix the rendering problems with Realm of Light on the main map.

I have not enabled this functionality for the minimap yet, since I do not know how to test it. (Perhaps it does not really matter either, until we implements forced tiling of the main map (if ever)).

### DIFF
--- a/common/src/main/java/com/wynntils/gui/render/MapRenderer.java
+++ b/common/src/main/java/com/wynntils/gui/render/MapRenderer.java
@@ -434,4 +434,20 @@ public class MapRenderer {
         double distanceZ = poi.getLocation().getZ() - mapCenterZ;
         return (float) (centerZ + distanceZ * currentZoom);
     }
+
+    public static List<MapTexture> getMapTextures(
+            int mapCenterX, int mapCenterZ, float renderWidth, float renderHeight, float scale) {
+        float renderDistanceX = renderWidth / 2f;
+        // scale is the conversion betweeen in-game ("map") coordinates and render coordinates
+        int mapDistanceX = (int) (renderDistanceX * scale);
+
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -434,4 +434,20 @@ public static float getRenderZ(Poi poi, float mapCenterZ, float centerZ, float c
         double distanceZ = poi.getLocation().getZ() - mapCenterZ;
         return (float) (centerZ + distanceZ * currentZoom);
     }
+
+    public static List<MapTexture> getMapTextures(
+            int mapCenterX, int mapCenterZ, float renderWidth, float renderHeight, float scale) {
+        float renderDistanceX = renderWidth / 2f;
+        // scale is the conversion betweeen in-game ("map") coordinates and render coordinates
+        int mapDistanceX = (int) (renderDistanceX * scale);
```
kristofbolyai, line 442: mapDistanceX is half render width?

kristofbolyai: i am blind.

kristofbolyai (approved):

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
magicus: I think the map rendering code is really tricky. There is no clear notion if coordinates are relative to the Minecraft world, or the screen. All types are the same, int, double or floats. Often, but not always, names indicate coordinate system, but there is no clear and consequent naming standard.

PR #494, opened 2022-10-02
Refactor bossbar code

Incompleteusern:


magicus:
Can you explain with a sentence or two what you are doing. Did you find a better way to hide boss bars?

magicus:
(Also, the term "refactor" is usually meant to imply that the structure of the code changes, but not the behavior. This more seems like a bug fix or an enhancement).

common/src/main/resources/wynntils.mixins.json
```diff
@@ -1,6 +1,5 @@
 {
   "client": [
-    "MainMixin",
```
magicus, line 3: 👍 for sorting this!!!

Incompleteusern:
> Can you explain with a sentence or two what you are doing. Did you find a better way to hide boss bars?

This removes many of the variables in `TrackedBar` by storing them inside a boss event, as well as replacing minecraft's boss event with one we have a reference to. This also gets rid of us duplicating progress info.

common/src/main/resources/wynntils.mixins.json
Incompleteusern: That was done automatically haha

Incompleteusern:
> (Also, the term "refactor" is usually meant to imply that the structure of the code changes, but not the behavior. This more seems like a bug fix or an enhancement).

Yeah, sorry

magicus (approved):

PR #500, opened 2022-10-03
It's cleanup time!

magicus:
Going through IntelliJ warnings on the "Artemis" profile I created earlier, fixing new issues that has crept up since the last cleaning time...


magicus:
This should be strict refactorings. Sorry if it is a bit of pain in the behind to review. It might be easier to go commit by commit; most of them are responses to a single type of warning/complaint from IntelliJ. Some required more complex solutions, but I've tried to make sure these include just the relevant files, so the changes can be followed easily.

Incompleteusern (approved):
LGTM

PR #504, opened 2022-10-03
Mod: Add ability to self update with /update

kristofbolyai:
image

### DIFF
--- a/common/src/main/java/com/wynntils/commands/UpdateCommand.java
+++ b/common/src/main/java/com/wynntils/commands/UpdateCommand.java
@@ -37,11 +37,12 @@ public class UpdateCommand extends CommandBase {
             switch (result) {
                 case SUCCESSFUL -> context.getSource()
                         .sendSuccess(
-                                new TextComponent("Successfully downloaded Wynntils update. It will apply on shutdown.")
+                                new TextComponent(
+                                                "Successfully downloaded Wynntils/Artemis update. It will apply on shutdown.")
                                         .withStyle(ChatFormatting.DARK_GREEN),
                                 false);
                 case ERROR -> context.getSource()
-                        .sendFailure(new TextComponent("Error applying Wynntils update.")
+                        .sendFailure(new TextComponent("Error applying Wynntils/Artemis update.")
                                 .withStyle(ChatFormatting.DARK_RED));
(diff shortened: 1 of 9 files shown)


Incompleteusern (approved):

kristofbolyai:
I think we can merge this. When @magicus is available, he can review and I can refactor in a follow-up.

common/src/main/java/com/wynntils/commands/UpdateCommand.java
```diff
@@ -0,0 +1,57 @@
...
+                    .sendFailure(new TextComponent("Development environment detected, cannot update!")
+                            .withStyle(ChatFormatting.DARK_RED));
+            WynntilsMod.error("Development environment detected, cannot update!");
+            return 0;
+        }
+
+        WynntilsMod.info("Attempting to fetch Wynntils update.");
+        CompletableFuture<UpdateManager.UpdateResult> completableFuture = UpdateManager.tryUpdate();
+
+        completableFuture.whenComplete((result, throwable) -> {
```
magicus, line 36: If this takes a long time, does it hang the minecraft chat?

kristofbolyai: I did not test, but I don't think so.

magicus: If so, it might be better to just send "Starting download" as a return for the command, and then do a "send chat message" thing with "Download complete" by the downloading thread when it is done.

kristofbolyai: We need to test this

magicus: One way to test this would be to set the URL to a server name that does not resolve. Usually DNS takes a handful of seconds before a missing name results in an error.

kristofbolyai: I could just thread sleep the request.

kristofbolyai: @magicus  https://github.com/Wynntils/Artemis/commit/738774f71e8e79bbd6abc0eb58c31b985527d582 I only tested with thread sleep so this might not be actually required, but we can have it anyways I guess.

PR #507, opened 2022-10-04
Features: Add HighlightMatchingItemsFeature

kristofbolyai:
image


kristofbolyai:
As seen, code is made to be easily extendable with only a new regex + config.

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
```diff
@@ -0,0 +1,101 @@
...
+            addSimpleSearchWidget(screen, renderX, renderY);
+        }
+
+        if (filterInGuildBank && GUILD_BANK_PATTERN.matcher(title).matches()) {
+            addSimpleSearchWidget(screen, renderX, renderY);
+        }
+
+        if (filterInGuildMemberList && MEMBER_LIST_PATTERN.matcher(title).matches()) {
+            addSimpleSearchWidget(screen, renderX, renderY);
+        }
```
Incompleteusern: Extract into a separate method that basically ANDs all three of them

```java
public boolean shouldAddSearch(String title) {
    if (filterInBank && BANK_PATTERN.matcher(title).matches()) {
        return true;
    }

    if (filterInGuildBank && GUILD_BANK_PATTERN.matcher(title).matches()) {
        return true;
    }

    if (filterInGuildMemberList && MEMBER_LIST_PATTERN.matcher(title).matches()) {
        return true;
    }

    return false;

}
```

On that note, pattern matching should be extracted into `/wynn/`

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
```diff
@@ -20,13 +22,17 @@
...
 import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(Screen.class)
-public abstract class ScreenMixin {
+public abstract class ScreenMixin implements TextboxScreen {
```
magicus, line 32: I don't think you need to add `implements` to mixins.

magicus: Ah, I see, that was not a vanilla interfance, but a way to get to the text input field. Nevermind.

common/src/main/resources/wynntils.mixins.json
```diff
@@ -37,6 +37,11 @@
     "accessors.OptionsAccessor"
   ],
   "compatibilityLevel": "JAVA_17",
+  "injected_interfaces": {
+    "net/minecraft/class_437": [
```
magicus, line 41: Do we need to use the obfuscated name here? That's a bit sad. :-( Can you at least add a comment on what class this really is?

magicus: (This is likely to break on Minecraft updates...)

kristofbolyai: Yeah this is what loom needs

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
magicus: Yes and no. We should not just move the patterns; that is pointless. But I have been thinking we'd need like a ContainerType enum in `wynn`, and a `ContainerType identifyContainer(Container c)`. It could also be given a bunch of convenience methods like `isBankPage`, `isLootChest` etc.

The ContainerIdentifier could also possibly add some additional data, like bank page number, loot chest tier, etc.

magicus: But I agree with the `shouldAddSearch`. As long as the end action is the same, and we don't adapt the  `addSimpleSearchWidget` call, that is clearer.

magicus: (Also, why `simple`? If there's only one search widget at the moment, you can skip that part of the name)

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private void highlightMatchingItems(String search, AbstractContainerScreen<?> screen) {
+        search = search.toLowerCase(Locale.ROOT);
+
+        for (ItemStack item : screen.getMenu().getItems()) {
+            if (!(item instanceof WynnItemStack wynnItemStack)) continue;
+
+            String name = ComponentUtils.getUnformatted(item.getHoverName()).toLowerCase(Locale.ROOT);
+
+            boolean filtered = !search.equals("") && name.contains(search);
```
magicus: Maybe use the fancy StringUtils more inclusive fuzzy matcher? (Forgot what it's called)

common/src/main/resources/wynntils.mixins.json
Incompleteusern: We are at the stage where a minecraft update would already be a boatload of work, so this is not that big of a concern.

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
Incompleteusern: PartialMatch iirc

magicus:
The code looks okay (barring my requests for comments etc). But to be clear, this is just "highlight matching items in a screen", not the "search" functionality of legacy. I'm not saying it's bad, but the ability in legacy to search the bank by going page by page was really useful. At some point, we'll need to include that in Artemis as well. Maybe we should name this feature differently in the meantime? `HighlightMatchingItemsFeature` rather than indicating "search"?

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
kristofbolyai: I did not use that on purpose, as I don't think it helps here. IMO it makes the search a lot wider than we need it to be

common/src/main/resources/wynntils.mixins.json
magicus: Actually, I don't think it will. I have been tempted from time to time to try and make a 1.19.2 branch, just to see how much we'd need to change.

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
magicus: Ok.

magicus (approved):

PR #517, opened 2022-10-06
GuiMixin: Remove onRenderFoodPre redirect

### LINKED ISSUE #516: Replace `onRenderFoodPre` redirect with an injection
This is one of our worst trouble making mixin. Note that the forge version has a sane fix, we only need to fix the common side.

kristofbolyai:
Fixes #516.

This redirect was hell. It is not only a redirect, but we redirect inside of a method, making it fail for random reasons and making Artemis incompatible with GUI modification mods.


kristofbolyai:
I am sure @magicus will like this :).

Incompleteusern (dismissed):

Incompleteusern (approved):

PR #520, opened 2022-10-06
CallbackInfo is not needed for events that are not cancelled

magicus:
I don't know if this has always been the case, or if Fabric has changed, but at least now we don't need to add the CallbackInfo unless we need to cancel the event. And in fact, in the debug log, we'll find tons of warning about "unused callback info".

### DIFF
--- a/common/src/main/java/com/wynntils/mc/mixin/AbstractContainerScreenMixin.java
+++ b/common/src/main/java/com/wynntils/mc/mixin/AbstractContainerScreenMixin.java
@@ -22,7 +22,7 @@ public abstract class AbstractContainerScreenMixin {
     public Slot hoveredSlot;
 
     @Inject(method = "render(Lcom/mojang/blaze3d/vertex/PoseStack;IIF)V", at = @At("RETURN"))
-    private void renderPost(PoseStack client, int mouseX, int mouseY, float partialTicks, CallbackInfo info) {
+    private void renderPost(PoseStack client, int mouseX, int mouseY, float partialTicks) {
         EventFactory.onContainerRender(
                 (AbstractContainerScreen<?>) (Object) this, client, mouseX, mouseY, partialTicks, this.hoveredSlot);
     }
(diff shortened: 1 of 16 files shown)


kristofbolyai:
Well. it might be the case we do not need it. But MInecraft Development "errors" if we do not have them. Also what logs are we talking about? I can't seem to find them?
image

magicus:
Perhaps you need to update your Minecraft Dev plugin? I think mine used to show the same, but does not any longer.

magicus:
The log file is `run/logs/debug.log`. I just discovered it by accident. :) It points out a bunch of things we're doing "less optimal". Not wrong perhaps, but not as good as we could.

kristofbolyai:
I just updated to nightly and it still shows the errors.

magicus:
> I just updated to nightly and it still shows the errors.

Weird. I'll see if I can figure out why mine doesn't.

kristofbolyai:
@Inject docs:

```
Basic usage
The simplest usage of @Inject captures no context from the target scope. This is particularly useful if the injector is targetting multiple methods with different signatures. In this case only the CallbackInfo (or CallbackInfoReturnable as appropriate) is required.
private void onSomeEvent(CallbackInfo ci)
Capture target arguments
Callbacks can also capture the arguments passed to the target method. To do so specify the target arguments before the CallbackInfo:
private void onSomeEvent(int arg1, String arg2, CallbackInfo ci)

**Comment by kristofbolyai** [comment]
Inject docs: 

```
Basic usage
The simplest usage of @Inject captures no context from the target scope. This is particularly useful if the injector is targetting multiple methods with different signatures. In this case only the CallbackInfo (or CallbackInfoReturnable as appropriate) is required.
private void onSomeEvent(CallbackInfo ci)
Capture target arguments
Callbacks can also capture the arguments passed to the target method. To do so specify the target arguments before the CallbackInfo:
private void onSomeEvent(int arg1, String arg2, CallbackInfo ci)
```

kristofbolyai:
But then, even if it builds, if it errors in the plugin, it is more a of a hassle..

magicus:
Yes, I agree with that. 

Let me get a minute to get to the bottom to this, otherwise I'll withdraw this PR.

magicus:
This seems just messed up. Don't know why we got that warning in the log file, nor why my Minecraft Devel is not complaining, but this does not seem to work. It fails during runtime due to incorrect signature. I'll just scrap this.

PR #522, opened 2022-10-07
Features: Add LobbyUptimeFeature

kristofbolyai:
image

This PR adds a new event, `ContainerSetSlotEvent`. This is very important to us, and here is why:
When opening a container, the first "page" seems to be always set by `ContainerSetSlotEvent` (1 by 1). When changing pages, we get both `ContainerSetSlotEvent` (1 by 1) and `ContainerSetContentEvent` (all items in one packet). This is quite annoying....

### DIFF
--- a/common/src/main/java/com/wynntils/core/webapi/profiles/ServerProfile.java
+++ b/common/src/main/java/com/wynntils/core/webapi/profiles/ServerProfile.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.webapi.profiles;
+
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+public class ServerProfile {
+
+    long firstSeen;
+    Set<String> players;
+
+    public ServerProfile(long firstSeem, Set<String> players) {
+        this.firstSeen = firstSeem;
+        this.players = players;
+    }
+
+    public Set<String> getPlayers() {
+        return players;
+    }
+
+    public long getFirstSeen() {
+        return firstSeen;
+    }
+
+    public String getUptime() {
+        long millis = System.currentTimeMillis() - firstSeen;
+
+        return String.format(
+                "%dh %dm",
+                TimeUnit.MILLISECONDS.toHours(millis),
+                TimeUnit.MILLISECONDS.toMinutes(millis)
+                        - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(millis)));
+    }
+
+    /**
+     * This makes the firstSeem match the user computer time instead of the server time
+     *
+     * @param serverTime the input server time
+     */
+    public void matchTime(long serverTime) {
+        firstSeen = firstSeen - (System.currentTimeMillis() - serverTime);
+    }
+}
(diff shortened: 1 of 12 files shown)


common/src/main/java/com/wynntils/features/user/LobbyUptimeFeature.java
```diff
@@ -0,0 +1,66 @@
...
+        if (matcher.matches()) {
+            String serverId = "WC" + matcher.group(1);
+            ServerProfile serverProfile = ServerListModel.getServer(serverId);
+            String uptimeString = serverProfile == null ? "Unknown" : serverProfile.getUptime();
+
+            ListTag loreTag = ItemUtils.getLoreTagElseEmpty(item);
+            loreTag.add(ItemUtils.toLoreStringTag(
+                    ChatFormatting.DARK_GREEN + "Uptime: " + ChatFormatting.GREEN + uptimeString));
+            ItemUtils.replaceLore(item, loreTag);
+        }
```
Incompleteusern: This should probably be a item stack transformer by its own merit, and in the future we'd also like to be able to display worlds with the numbers greater than 64

common/src/main/java/com/wynntils/core/webapi/profiles/ServerProfile.java
```diff
@@ -0,0 +1,46 @@
...
+                        - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(millis)));
+    }
+
+    /**
+     * This makes the firstSeem match the user computer time instead of the server time
+     *
+     * @param serverTime the input server time
+     */
+    public void matchTime(long serverTime) {
+        firstSeen = firstSeen - (System.currentTimeMillis() - serverTime);
```
Incompleteusern: I know this is a port from legacy, but I still don't think this is correct?
```
System.currentTimeMillis() - newFirstSeen = serverTime - firstSeen
```
is the equation that maintains the change in time.
This rearranges to give
```
newFirstSeen = (System.currentTimeMillis() - serverTime) + firstSeen
```

Incompleteusern (commented):
Comments

magicus:
Also, server selection is an area that is just screaming for having a custom UI from Wynntils :)

common/src/main/java/com/wynntils/core/webapi/profiles/ServerProfile.java
```diff
@@ -0,0 +1,46 @@
...
+
+        return String.format(
+                "%dh %dm",
+                TimeUnit.MILLISECONDS.toHours(millis),
+                TimeUnit.MILLISECONDS.toMinutes(millis)
+                        - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(millis)));
+    }
+
+    /**
+     * This makes the firstSeem match the user computer time instead of the server time
```
magicus:
```suggestion
     * This makes the firstSeen match the user computer time instead of the server time
```

common/src/main/java/com/wynntils/mc/EventFactory.java
```diff
@@ -240,6 +242,11 @@ public static ContainerSetContentEvent onContainerSetContent(ClientboundContaine
                 packet.getItems(), packet.getCarriedItem(), packet.getContainerId(), packet.getStateId()));
     }
 
+    public static void onContainerSetSlot(ClientboundContainerSetSlotPacket packet) {
+        post(new ContainerSetSlotEvent(
```
magicus, line 246: Is this still used? It does not seem so? Was it not needed once you started using WynnItemStacks, or is there some bug hiding here? 

And if it is not needed, should you remove it for the PR or do you think it will be needed very soon so you want to keep it?

kristofbolyai: I want to keep it for my upcoming PRs

magicus (approved):

PR #527, opened 2022-10-08
Add dry_streak and dry_boxes info variables

kristofbolyai:


common/src/main/java/com/wynntils/wynn/item/UnidentifiedItemStack.java
```diff
@@ -30,6 +30,7 @@
 public class UnidentifiedItemStack extends WynnItemStack {
     private final List<Component> tooltip;
     private ItemType itemType;
+    private ItemTier itemTier;
```
kristofbolyai, line 33: This change was not used in the end. Although, I believe this is a reasonable change, but if you want, I can revert it.

common/src/main/java/com/wynntils/mc/mixin/ChestMenuMixin.java
```diff
@@ -0,0 +1,28 @@
...
+
+@Mixin(ChestMenu.class)
+public abstract class ChestMenuMixin extends AbstractContainerMenu {
+    protected ChestMenuMixin(MenuType<?> menuType, int i) {
+        super(menuType, i);
+    }
+
+    @Inject(method = "quickMoveStack", at = @At("HEAD"))
+    public void onQuickMoveStack(Player player, int index, CallbackInfoReturnable<ItemStack> cir) {
+        EventFactory.onChestMenuQuickMove(this.containerId);
```
magicus: I think you can do a double cast like this:
```
        EventFactory.onChestMenuQuickMove((ChestMenu)(Object)this.containerId);
```
and then you don't need to extend the abstract class, and can remove the constructor.

common/src/main/java/com/wynntils/wynn/item/UnidentifiedItemStack.java
magicus: It's ok.

common/src/main/java/com/wynntils/features/statemanaged/PersistentFunctionsFeature.java
```diff
@@ -0,0 +1,72 @@
...
+import com.wynntils.mc.event.MenuEvent;
+import com.wynntils.mc.utils.ComponentUtils;
+import com.wynntils.wynn.item.parsers.WynnItemMatchers;
+import com.wynntils.wynn.utils.ContainerUtils;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+// FIXME: This feature is only needed because we do not have a way to save function data persistently.
+//        Remove this when we add persistent data storage other than configs.
+public class PersistentFunctionsFeature extends StateManagedFeature {
```
magicus: I'm thinking this is typical "model" knowledge. The LootRunManager is quite long as it is, but perhaps this is the seed of a new `LootChestModel`?

Can models have `@Config`? (I'm not sure they should be able to, but perhaps.) In any case, they should definitely have a way to store data persistently.

kristofbolyai: They cant

magicus: Ok. 

My recommendation is still to move the actual logic to a new Model, and just keep the feature for data storage. (In fact, perhaps you can rename it to just `DataStorageFeature` and we'll just cram in our need for persistent data there, regardless of origin, until we develop a better solution for that? That way we can easily discover all places that need updating when we finally do implement persistant data storage.

kristofbolyai: Then, we could just move the code to the functions and store values in the feature?

magicus: That's not really better. The point I'm trying to drive home is that we need to collect all "business logic" in a single place, the `wynn` package (aka "the models"). That is, as soon as you start doing some kind of non-obvious processing of raw Minecraft data, it should be extracted into a model.

So, for instance, matching a container title to a regexp to see what kind of container it is; while only a single line of code, is "non-trivial" enough that it is better to create a separate function, like `isLootChest` in a model, and use that instead. The exemptions are stuff like the players current xyz, which is "trivial" enough to grab directly from minecraft.

I think this lack of separation was what ultimately lead Legacy astray. There were other design problems (and mostly the lack of keeping to the conceived design ideas), but this one would have been the hardest to fix, since it was so pervasive.

And I still think it is the hardest to keep you guys from keeping up in Artemis. ;-) It is so convenient to just slip in a piece of business logic here and there. But suddenly we will find ourselves once again with doing the same time in different places and slightly different ways.

I think one reason it is problematic is that we still do not have a good unified and coherent API for the business logic, aka The One Model to Rule Them All. That was hard to do initially when we had very little actual business logic and the design where elusive. But now I think we have enough to shape a rather good API. I've been thinking about it for quite some while, but been postponing it over and over, mostly because real-life requirements have prohibited me from spending the amount of continuous time I'd need to do that.

magicus: So if you check the other Functions you'll see that they are all more or less wrappers of calls to model methods.

common/src/main/java/com/wynntils/mc/mixin/ChestMenuMixin.java
Incompleteusern: Double casts can be done

common/src/main/java/com/wynntils/features/statemanaged/PersistentFunctionsFeature.java
kristofbolyai: > So if you check the other Functions you'll see that they are all more or less wrappers of calls to model methods.

(but why are they registered to the event bus then?)

magicus (approved):

common/src/main/java/com/wynntils/features/statemanaged/PersistentFunctionsFeature.java
magicus: That is a very good question. Hm. They probably shouldn't be.

PR #530, opened 2022-10-09
Unregister keybinds

### LINKED ISSUE #514: Disabled features still have keybinds enabled

Incompleteusern:
image
image
image

closes #514

### DIFF
--- a/common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
+++ b/common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
@@ -9,7 +9,6 @@ import com.wynntils.core.config.ConfigManager;
 import com.wynntils.core.features.properties.RegisterKeyBind;
 import com.wynntils.core.features.properties.StartDisabled;
 import com.wynntils.core.keybinds.KeyBind;
-import com.wynntils.core.keybinds.KeyBindManager;
 import com.wynntils.core.managers.CrashReportManager;
 import com.wynntils.features.alwayson.FixSpellOverwriteFeature;
 import com.wynntils.features.alwayson.LootrunFeature;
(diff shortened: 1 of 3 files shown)


magicus:
Hm. Question. My thinking was that disabling a Feature with a KeyBind should mean that pressing the key of the KeyBind would result in no action; not that the entire keybinding disappears from the settings.

There are certainly pros and cons with both approaches. What do you say @kristofbolyai, do you agree with this approach? If so, you're two to one and I'll accept it.

I think the main divider here is if you think of disabling features as a short-term exceptional thing (in which case I don't think we should remove the keybinding option from the vanilla screen), or more of a permanent "i don't ever want to see this cr*p", in which removing the keybinding in full would be the proper thing to do.

Incompleteusern:
> Hm. Question. My thinking was that disabling a Feature with a KeyBind should mean that pressing the key of the KeyBind would result in no action; not that the entire keybinding disappears from the settings.
> 
> There are certainly pros and cons with both approaches. What do you say @kristofbolyai, do you agree with this approach? If so, you're two to one and I'll accept it.
> 
> I think the main divider here is if you think of disabling features as a short-term exceptional thing (in which case I don't think we should remove the keybinding option from the vanilla screen), or more of a permanent "i don't ever want to see this cr*p", in which removing the keybinding in full would be the proper thing to do.

The reason for removing the key bind is to get rid of conflicts, and to not appear when it’s inactive, which would be a misnomer

kristofbolyai:
> > Hm. Question. My thinking was that disabling a Feature with a KeyBind should mean that pressing the key of the KeyBind would result in no action; not that the entire keybinding disappears from the settings.
> 
> > 
> 
> > There are certainly pros and cons with both approaches. What do you say @kristofbolyai, do you agree with this approach? If so, you're two to one and I'll accept it.
> 
> > 
> 
> > I think the main divider here is if you think of disabling features as a short-term exceptional thing (in which case I don't think we should remove the keybinding option from the vanilla screen), or more of a permanent "i don't ever want to see this cr*p", in which removing the keybinding in full would be the proper thing to do.
> 
> 
> 
> The reason for removing the key bind is to get rid of conflicts, and to not appear when it’s inactive, which would be a misnomer

I would rather remove the keybind atm. We would need to modify the screen to say "enable x feature for this keybind to work" and also implement a keybind conflict system. Since we don't have either of those, just unregister the keybinds

kristofbolyai (approved):

PR #536, opened 2022-10-11
WynntilsQuestBookFeature: Fix a bug where quest book does not open correctly if clicked on an entity

kristofbolyai:


magicus:
Do we have the same problem in Artemis?

kristofbolyai:
> Do we have the same problem in Artemis?

This is artemis?

kristofbolyai:
I need these events anyways for middle click to view gear..

magicus:
Oh, I'm just confused. I was clicking through a long list of Legacy notifications, but each time I go back Github reorders the repositories...

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
```diff
@@ -7,29 +7,24 @@
...
-    private final BlockPos pos;
-
-    private final Direction face;
-
     private InteractionResult cancellationResult = InteractionResult.PASS;
 
-    private PlayerInteractEvent(Player player, InteractionHand hand, BlockPos pos, Direction face) {
+    private PlayerInteractEvent(Player player, InteractionHand hand) {
         super(Preconditions.checkNotNull(player, "Null player in PlayerInteractEvent!"));
         this.hand = Preconditions.checkNotNull(hand, "Null hand in PlayerInteractEvent!");
```
magicus: I never noticed these before. Why do we not trust vanilla to have non-null values?

kristofbolyai: I think this is forge event code from a few months ago.

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
```diff
@@ -94,6 +82,11 @@ public void setUseItem(Event.Result triggerItem) {
             this.useItem = triggerItem;
         }
 
+        @Nonnull
```
magicus: Also this. I thought we agreed we should not have random nullness annotations sprinkled here and there?

magicus: (Doing the "annotate for null safety" needs to be all-in or nothing, otherwise it's quite pointless. And I don't believe the payoff is worth the effort for all-in)

kristofbolyai: This is from javax iirc. So we did not block these

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
kristofbolyai: anyways i removed this and the annotations

magicus (approved):

PR #538, opened 2022-10-11
CustomNametagRendererFeature: Add ability to render used gear

kristofbolyai:
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
@@ -7,18 +7,37 @@ package com.wynntils.features.user;
 import com.wynntils.core.config.Config;
 import com.wynntils.core.features.UserFeature;
 import com.wynntils.core.managers.Model;
+import com.wynntils.core.webapi.WebManager;
+import com.wynntils.core.webapi.profiles.item.ItemProfile;
 import com.wynntils.mc.event.NametagRenderEvent;
+import com.wynntils.mc.utils.ComponentUtils;
+import com.wynntils.mc.utils.McUtils;
 import com.wynntils.wynn.model.UserInfoModel;
 import com.wynntils.wynn.objects.account.AccountType;
+import com.wynntils.wynn.utils.WynnPlayerUtils;
 import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.entity.projectile.ProjectileUtil;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.phys.AABB;
+import net.minecraft.world.phys.EntityHitResult;
+import net.minecraft.world.phys.Vec3;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class CustomNametagRendererFeature extends UserFeature {
+    private static final float RAYCAST_RANGE = 5f;
+
     @Config
     public boolean hideAllNametags = false;
 
     @Config
     public boolean showAccountType = true;
+
+    @Config
+    public boolean showGearOnHover = true;
 
     @SubscribeEvent
     public void onNameTagRender(NametagRenderEvent event) {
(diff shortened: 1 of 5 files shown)


kristofbolyai:
(it also renders held item name but ss is old)

kristofbolyai:
(also wynntils developer is now rendered after the name, where it should be)

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
```diff
@@ -7,33 +7,117 @@
...
+    private static final float RAYCAST_RANGE = 5f;
+
     @Config
     public boolean hideAllNametags = false;
 
     @Config
     public boolean showAccountType = true;
 
+    @Config
+    public boolean showGearOnHover = true;
```
magicus, line 37: Is this so popular that it should be true by default? I think I'd mostly be annoyed by it.

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
```diff
@@ -7,33 +7,117 @@
...
+
+        Vec3 start = player.getEyePosition(1f);
+        Vec3 look = player.getLookAngle();
+        Vec3 direction = start.add(look.x * RAYCAST_RANGE, look.y * RAYCAST_RANGE, look.z * RAYCAST_RANGE);
+        AABB bb = player.getBoundingBox()
+                .expandTowards(look.x * RAYCAST_RANGE, look.y * RAYCAST_RANGE, look.z * RAYCAST_RANGE)
+                .expandTowards(1, 1, 1);
+
+        EntityHitResult hitResult = ProjectileUtil.getEntityHitResult(
+                McUtils.mc().level, player, start, direction, bb, (e) -> e instanceof Player);
```
magicus: Maybe extract into a *Util class? It sounds general enough that other can be interested in it in the future.

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
kristofbolyai: Oh that might be the case. It might be off by default in legacy too.

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
kristofbolyai: I will in my upcoming PR which uses this same code.

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
Incompleteusern: I think it should be true by default. You typically don't encounter that many players on your server in the gameplay

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
```diff
@@ -7,33 +7,117 @@
...
+        if (itemName.contains("Crafted")) {
+            event.addInjectedLine(new TextComponent(itemName).withStyle(ChatFormatting.DARK_AQUA));
+            return;
+        }
+
+        ItemProfile itemProfile = WebManager.getItemsMap().get(itemName);
+        if (itemProfile == null) return;
+
+        // this solves an unidentified item showcase exploit
+        // boxes items are STONE_SHOVEL, 1 represents UNIQUE boxes and 6 MYTHIC boxes
```
magicus, line 86: Is this still not fixed at the Wynncraft side? :-o

kristofbolyai: Not sure, I did not test.

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
```diff
@@ -7,33 +7,117 @@
...
+        EntityHitResult hitResult = ProjectileUtil.getEntityHitResult(
+                McUtils.mc().level, player, start, direction, bb, (e) -> e instanceof Player);
+
+        if (hitResult == null) return;
+        if (hitResult.getEntity() != event.getEntity()) return;
+
+        Player hitPlayer = (Player) hitResult.getEntity();
+
+        if (hitPlayer.getScoreboardName().contains("§")) return; // npc
+        if (WynnPlayerUtils.isPlayerGhost(hitPlayer)) return;
```
Incompleteusern: Combine these checks, also consider making a `WynnPlayerUtils#isNPC` as the `§` check has already been used elsewhere

common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
Incompleteusern: I wonder if we could reduce it to show `Unidentified Bow` and such, but that might be stretching it.

magicus (dismissed):

PR #541, opened 2022-10-12
MapRenderer Refactoring

Incompleteusern:
This pr decreases the size of MapRenderer by inlining specialized cases. It also fleshes out POI which allows for merging of service and label rendering, as well as creating a BoundingBox construct.

It also removes `minScaleForLabels` config for now, adding preliminary alpha fading.

This also fixes the issue of the cursor not being rendered correctly.

image


kristofbolyai:
While at it, you could fix player icon thing rendering outside of main map if it is on one side :). Or we can just make an issue

common/src/main/java/com/wynntils/features/user/overlays/map/MinimapFeature.java
```diff
@@ -117,43 +127,99 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
...
                         this.renderUsingLinear);
             }
 
-            // TODO minimap icons
+            // disable rotation if necessary
+            if (followPlayerRotation) {
+                poseStack.popPose();
+            }
+
+            List<Poi> pois = MapModel.getAllPois()
```
magicus, line 173: Do you think we should have a `MapModel.getPoisForBoundingBox`, to do some initial filtering?

Incompleteusern: I don't think that such a method is necessary, as we can filter in `MainMapScreen` when calculating hovered and `MinimapFeature` has rotation, which makes filtering also annoying.

Though I agree some easy, more general checks could be good, I'm unable to think of how we could do one without false positives.

magicus (approved):
Looks good. Thanks for cleaning this up.

PR #547, opened 2022-10-13
Add Guide menu screen, Item Guides and favoriting feature

kristofbolyai:
image
image
image
image
image


common/src/main/java/com/wynntils/features/user/ItemFavoriteFeature.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+
+    @SubscribeEvent
+    public void onRenderSlot(SlotRenderEvent.Post event) {
+        ItemStack itemStack = event.getSlot().getItem();
+
+        if (isFavorited(itemStack)) {
+            renderFavoriteItem(event);
```
Incompleteusern, line 63: Could this be made a property?

kristofbolyai: Lock is not a property either atm, Maybe it could be, but we still need the feature for blocking chest closing so

Incompleteusern: Yeah, maybe this is better for a future pr.

Incompleteusern (approved):

magicus:
Oh, nice! If you're going to continue working on this, can I ask a favour? Please try to make filtering better than in legacy. I always thought the guide was quite useless, at least in the "lower" levels that I ever reached. I had a use case like "research a good bow for my level 50 archer", but I could not easily filter out so I only see bows for level 50 (+/- 10 or so), and perhaps sort them by neutral damage, total damage, rarity, etc.

The legacy concept of "clicking on" different subsets never made much sense to me. I had to "click away" all but the category I wanted to look at. And if I wanted to look at e.g. armor instead, I had to click twice; first remove the category I had displayed, and then select the new.

I'm not sure what we should do instead, but something more akin to the search system in the trade market I think. A "filter" system where you OR alternatives within each "group" (like "rare OR legendary") and AND between groups (like "(rare OR legendary) AND (helmet OR boots)". And then a free-text search. (Which technically in a way is yet another AND:ed filter)

Incompleteusern:
Legacy had a filter system similar to the one you mentioned

magicus:
I never got it to work, properly then... Maybe it just needs some UX/UI love.

PR #551, opened 2022-10-14
chore: Remove account type config

DevScyu:
As discussed in discord, we should not be able to disable donator perks.


common/src/main/java/com/wynntils/features/user/CustomNametagRendererFeature.java
```diff
@@ -104,13 +100,12 @@ private static void getItemComponent(NametagRenderEvent event, ItemStack itemSta
...
+        if (user == null) return;
+        AccountType accountType = user.accountType();
+        if (accountType.getComponent() == null) return;
 
-            if (UserInfoModel.isAccountType(event.getEntity().getUUID(), type)) {
-                event.addInjectedLine(type.getComponent());
-            }
-        }
+        event.addInjectedLine(accountType.getComponent());
     }
```
DevScyu, line 109: This was refactored due to only a single account type per account, so we don't need to loop through all of them.

PR #552, opened 2022-10-14
Fix UnidentifiedItemStack NPE

kristofbolyai:


common/src/main/java/com/wynntils/wynn/item/UnidentifiedItemStack.java
```diff
@@ -139,6 +139,6 @@ public Optional<ItemTier> getItemTier() {
     }
 
     public List<String> getPossibleItems() {
-        return itemPossibilities;
+        return itemPossibilities == null ? List.of() : itemPossibilities;
```
magicus, line 142: Is there some Qodana check we can do to verify that collection types are never null? Since they have the empty collection, they never need to be null to represent "no value". If we could get null-checking for those, without a lot of annotation overhead, I'd actually think it was good.

kristofbolyai: Probably there is none. We could check though

PR #555, opened 2022-10-14
PowderAbilityBarOverlay: Fix memory leak

kristofbolyai:
Old: 183MB / 12 sec allocation.
New: Disappeared from profiling results (< 1 MB / 12 sec)


kristofbolyai:
I will merge this without reviews as this is really messed up. If there is some changes requested, I'll do it later.

PR #560, opened 2022-10-15
BeaconBeamFeature: Heavily optimize rendering memory usage

kristofbolyai:
allocations:
old: 400mb / 12 sec 
new: 4mb / 12 sec

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
@@ -88,6 +88,6 @@ public class BeaconBeamFeature extends UserFeature {
 
         poseStack.popPose();
 
-        bufferSource.endBatch();
+        bufferSource.endLastBatch();
     }
 }


Incompleteusern (approved):

magicus:
As I said before, I'm not particularly worried about performance, with the exception of rendering code. I think this was a good example of the latter.

But. I also realize I need to modify my comment a bit. I'm not worried about general performance, but high allocation rates might be an issue, even outside rendering. We might need to keep track on that. How did you notice this?

kristofbolyai:
> As I said before, I'm not particularly worried about performance, with the exception of rendering code. I think this was a good example of the latter.
> 
> But. I also realize I need to modify my comment a bit. I'm not worried about general performance, but high allocation rates might be an issue, even outside rendering. We might need to keep track on that. How did you notice this?

This week, I've pushed 10+ commits fixing high memory allocation in Artemis. I've been doing profiling with async-profiler, profiling memory allocations. If you look at commit/PR history, you can see a lot of really big allocations which were not needed. I think I went through and fixed almost everything I could find while profiling.

magicus:
I noticed that shortly after, when I started working my way through the old notification list. :-) Great job!

(I also notice that my motto still seems to hold; most if not not all of the fixes seemed to be render code, or called from rendering code)

kristofbolyai:
> I noticed that shortly after, when I started working my way through the old notification list. :-) Great job!
> 
> (I also notice that my motto still seems to hold; most if not not all of the fixes seemed to be render code, or called from rendering code)

We still have some allocation in item stack transformers, but fixing that is not really straightforward, maybe even impossible. But yes, mostly only rendering is heavy on memory.

PR #567, opened 2022-10-16
Fix ingredient modifier text

TomaSajt:
This PR removes an accidentally duplicated word and adds 'of' after 'right' and 'left'.


kristofbolyai (approved):
Thanks! I didn't notice these when porting the screens.

PR #568, opened 2022-10-16
Features: Add AutoApplyResourcePackFeature

kristofbolyai:


common/src/main/java/com/wynntils/mc/mixin/ClientPackSourceMixin.java
```diff
@@ -0,0 +1,49 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ClientPackSource.class)
+public abstract class ClientPackSourceMixin {
+    @Shadow
+    private Pack serverPack;
+
+    @Inject(method = "clearServerPack", at = @At("HEAD"), cancellable = true)
+    private void onClearServerPackPre(CallbackInfo ci) {
+        if (serverPack == null) {
```
magicus, line 28: There's just a tad too much going on here. Don't you think the received of the event can calculate the hash, for instance?

magicus: (The point in doing it after the event is sent is that it will not affect non-Wynncraft servers.)

kristofbolyai: Sure

kristofbolyai: We always need to send the event because this happens after disconnect. Do you still want to move this?

magicus:
Just to clarify: this is the same feature as in legacy, that the resource pack will be applied directly from start, and not load when joining the Wynncraft server?

common/src/main/java/com/wynntils/features/user/AutoApplyResourcePackFeature.java
```diff
@@ -0,0 +1,70 @@
...
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class AutoApplyResourcePackFeature extends UserFeature {
+    private final File SERVER_RESOURCE_PACK_DIR = new File(McUtils.mc().gameDirectory, "server-resource-packs");
+
+    @Config(visible = false)
+    private String packHash = "";
+
+    private String appliedHash = "";
```
Incompleteusern, line 28: What is the difference between the two?

kristofbolyai:
> Just to clarify: this is the same feature as in legacy, that the resource pack will be applied directly from start, and not load when joining the Wynncraft server?

Yes, and we also disable the mc feature that unloads the pack when leaving a server. We only disable that though if we have applied the pack.

common/src/main/java/com/wynntils/features/user/AutoApplyResourcePackFeature.java
kristofbolyai: The config is to save the last pack hash Wynn used, the second is to keep track of the currently loaded pack. Needed if Wynn changes pack's mid-game, the game won't be in a weird state.

kristofbolyai: (otherwise I could use a bool)

Incompleteusern: Then what does 
```java

        for (File file : files != null ? files : new File[0]) {
            if (clientPackSource.checkHash(packHash, file)) {
                clientPackSource.setServerPack(file, PackSource.DEFAULT);
                appliedHash = packHash;
            }
        }
```
do?

kristofbolyai: Find which file to load in the server resources directory

common/src/main/java/com/wynntils/mc/mixin/ClientPackSourceMixin.java
magicus: Ah. Hm. Ok, keep it. Maybe add a comment stating what you just said, that this happens after disconnect so needs to be done always.

common/src/main/java/com/wynntils/features/user/AutoApplyResourcePackFeature.java
Incompleteusern: Isn't `appliedHash` more of the local file we have on the client?

kristofbolyai: appliedHash is the same as packHash if we have pre-loaded the pack. Otherwise it is empty.

Incompleteusern (dismissed):

kristofbolyai:
Don't merge this yet. I am working on making it load a bit before title init.

kristofbolyai:
> Don't merge this yet. I am working on making it load a bit before title init.

This is a pain in the ass to do, it will be okay as it is right now.

Incompleteusern (approved):

Issue #510, opened 2022-10-05
Guild Objective not self-destructing after completion


IngeniousCoder:
![Uploading image.png…]()

Bar stays there.

Could it be another #386?

IngeniousCoder:
![image](https://user-images.githubusercontent.com/29097583/193998747-e7f33389-f9b6-4f89-a5b3-5b58468d16c7.png)

kristofbolyai:
Stays until /class, as legacy

kristofbolyai:
Oh but it's 5/6

IngeniousCoder:
its completed already tho, as seen in the chat menu

magicus:
I don't remember how this looks with just vanilla Wynncraft scoreboard. When you have finished the weekly objectives, does it stay in the scoreboard with like "[6/6]"? Or does it disappear from the scoreboard, maybe after a while?

If it is the latter, maybe we happen to miss the scoreboard update? @kristofbolyai What happens if we get a scoreboard where an entire segment has disappeared. Do we clear out that segment, or can it be that it happens to just hang around indefinitely?

Perhaps we need more robust code to account for the fact that we might miss intermediate updates of the scoreboard (since we're only checking it periodically). An alternative fallback is to listen to the "... has finished their weekly objective" in chat, and clear out the guild objectives explicitly if that is the case.

magicus:
@IngeniousCoder If you still have this in session, how does it look like if you turn of the Wynntils objective rendering, and only use the vanilla scoreboard?

Incompleteusern:
> Perhaps we need more robust code to account for the fact that we might miss intermediate updates of the scoreboard (since we're only checking it periodically). An alternative fallback is to listen to the "... has finished their weekly objective" in chat, and clear out the guild objectives explicitly if that is the case.

Can't we just check when a specific packet is processed, if that isn't done already? Unless that too many packets are sent?

IngeniousCoder:
The side would say
```
Win Dungeons 6/6

Use /guild rewards to claim your rewards
```

Ideally it should stay at 6/6 and automatically disappear after claimed, but my case it got stuck at 5/6.

IngeniousCoder:
> Perhaps we need more robust code to account for the fact that we might miss intermediate updates of the scoreboard (since we're only checking it periodically). An alternative fallback is to listen to the "... has finished their weekly objective" in chat, and clear out the guild objectives explicitly if that is the case.

[...]

PR #569, opened 2022-10-16
Rewrite ScoreboardModel

Incompleteusern:


kristofbolyai:
Maybe this can be fixed here #510

magicus:
As a general rule, it is a better idea to keep pure refactorings and actual change in behavior (e.g. bug fixes) separated. If something changes behavior after a refactoring, we know for certain it is a problem. And if we do the refactoring properly, the bug fix should be simple in a follow-up PR.

magicus (changes requested):

PR #571, opened 2022-10-17
Materials: Use athena data

kristofbolyai:


common/src/main/java/com/wynntils/core/webapi/profiles/ingredient/IngredientInfo.java
```diff
@@ -0,0 +1,41 @@
...
+        if (materialName == null) {
+            return new ItemStack(Items.AIR);
+        }
+
+        Optional<Item> item = Registry.ITEM.getOptional(new ResourceLocation(materialName));
+        ItemStack itemStack = item.map(ItemStack::new).orElseGet(() -> new ItemStack(Items.AIR));
+
+        if (metadata != null) {
+            itemStack.setDamageValue(Integer.parseInt(metadata));
+        }
```
Incompleteusern, line 37: How does this support skulld?

kristofbolyai: Oh right. Oops

common/src/main/java/com/wynntils/core/webapi/WebManager.java
```diff
@@ -226,42 +221,6 @@ private static void tryLoadItemGuesses() {
         // Check for success
     }
 
-    private static void tryLoadItemMaterialIdMap() {
```
magicus, line 229: How do we get the data if you remove the json download? Not sure I'm following the changes here.

kristofbolyai: I made scyu update athena so the new ids are added there, and will be always up-to-date.

magicus: So prior to this fix we got info from two sources, old ids from Athena and new from the json file? And now all are merged into one in Athena?

kristofbolyai: Yes

magicus (approved):

PR #576, opened 2022-10-18
CharacterManager: Fix profession info parsing

kristofbolyai:


magicus (approved):
There can be decimal numbers now?

kristofbolyai:
> There can be decimal numbers now?

Yeah... I just noticed too

PR #578, opened 2022-10-18
QuestManager: Fix mini-quest sorting

kristofbolyai:


Incompleteusern (approved):

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
```diff
@@ -91,6 +91,12 @@ public int getLevel() {
         return level;
     }
 
+    public int getSortLevel() {
+        return !isMiniQuest || additionalRequirements.isEmpty()
+                ? level
+                : additionalRequirements.get(0).b();
```
magicus, line 97: @kristofbolyai 
So if a mini-quest is mis-tagged and do not have any addition requirement, we'll crash? :-)

I think a more robust implementation would return `level` if `additionalRequirements.isEmpty()`, and otherwise `additionalRequirements.get(0).b()` iff this `isMiniQuest`.

kristofbolyai: This is correct? Can you check again?

magicus: You are right. This is correct. Nevermind me, I'll just go back to sleep and return when I've gotten all my morning coffee down.

PR #582, opened 2022-10-18
Feature category assert

### LINKED ISSUE #478: Assert that category I18n keys are present
Does not error.

image

kristofbolyai:
Fixes #478


common/src/main/java/com/wynntils/core/features/properties/FeatureCategory.java
```diff
@@ -16,6 +16,10 @@ public enum FeatureCategory {
     REDIRECTS,
     TOOLTIPS;
 
+    FeatureCategory() {
+        assert !toString().startsWith("core.wynntils");
```
Incompleteusern, line 20: Is there a safer check?

kristofbolyai: Safer as in? We do this with features too. I think this is good enough.

Incompleteusern (approved):

PR #586, opened 2022-10-20
Implement Hades in Artemis

kristofbolyai:


common/src/main/java/com/wynntils/wynn/model/PlayerRelationsModel.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public static void disable() {
+        friends = Set.of();
+    }
+
+    @SubscribeEvent
+    public static void onChatReceived(ChatPacketReceivedEvent e) {
+        if (e.getType() != ChatType.SYSTEM) return;
+
+        if (expectingFriendMessage) {
```
magicus: We should probably test this with an account that has no friends...

magicus: (I know for certain there were a long-standing bug in legacy where we tried to detect guild members and it failed if you were not part of a guild)

kristofbolyai: I happen to have an account for this :)

Incompleteusern: I wonder if we could parse the tab list first, and if that is full only then run the command

kristofbolyai: This is more reliable imo.

magicus: The tab list cut long names short, so it will not work, unfortunately. :( Iirc, this was done in the packet sent by the server, and not only on client rendering.

Incompleteusern: True

kristofbolyai: Fixed

common/src/main/java/com/wynntils/wynn/model/PlayerRelationsModel.java
```diff
@@ -0,0 +1,262 @@
...
+    private static final Pattern PARTY_NO_LIST_MESSAGE_PATTERN = Pattern.compile("§eYou must be in a party to list.");
+    private static final Pattern PARTY_OTHER_LEAVE_MESSAGE_PATTERN = Pattern.compile("§e(.+) has left the party.");
+    private static final Pattern PARTY_OTHER_JOIN_MESSAGE_PATTERN = Pattern.compile("§e(.+) has joined the party.");
+    private static final Pattern PARTY_SELF_LEAVE_MESSAGE_PATTERN =
+            Pattern.compile("§eYou have been removed from the party.");
+    private static final Pattern PARTY_SELF_JOIN_MESSAGE_PATTERN =
+            Pattern.compile("§eYou have successfully joined the party.");
+    private static final Pattern PARTY_DISBAND = Pattern.compile("§eYour party has been disbanded.");
+
+    private static boolean expectingFriendMessage = false;
```
magicus, line 47: Do you think we should have a timeout for this? If something goes wrong, we'll be looking for these replies for ever. And then the user types `/friends` manually, and the result is eaten up here.

I'm not sure about this. A timeout complicates code, and it can instead cause the problem "what if the reply is just a bit later than the timeout?". So this is mostly me thinking loud. I'll leave the decision to do this, or not, to you.

kristofbolyai: Wynn can lag quite a bit, text messages can be delayed by seconds. I think eating up a /party list or /friend list is not too bad.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -220,12 +226,59 @@ private void renderMap(PoseStack poseStack, int mouseX, int mouseY) {
...
+        renderPlayerIcons(poseStack, textureBoundingBox, mouseX, mouseY);
+
         // Cursor
         renderCursor(poseStack);
 
         RenderSystem.disableScissor();
     }
 
+    private void renderPlayerIcons(PoseStack poseStack, BoundingBox textureBoundingBox, int mouseX, int mouseY) {
+        for (HadesUser user : HadesUserModel.getHadesUserMap().values()) {
```
Incompleteusern: Can player icons be made into a POI?

kristofbolyai: ATM Pois have a fixed location. As these are meant to be updated frequently, I chose to do this, to save a lot of allocations. I think this is candidate for a POI refactor (as we would need to update POI locations for player-defined POIs aswell)

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
```diff
@@ -0,0 +1,146 @@
...
+
+        hadesConnection.sendPacket(new HCPacketUpdateWorld(
+                WorldStateManager.getCurrentWorldName(),
+                CharacterManager.getCharacterInfo().getId()));
+    }
+
+    @SubscribeEvent
+    public static void onTick(ClientTickEvent.End event) {
+        if (!isSocketOpen()) return;
+        if (!WorldStateManager.onWorld()) return;
```
Incompleteusern: Combine the first two checks?

kristofbolyai: Sure

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
Incompleteusern: Maybe POIs could use a supplier or something. An arguement for this is that we should be able to place a compass on player icons and making them POIs would give that easily.

kristofbolyai: But you agree that this is fine as a workaround for now, so we do not have to change POI infrastructure in an already big PR?

Incompleteusern: Yeah, best for a separate pr that does suppliers + caching

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
```diff
@@ -0,0 +1,202 @@
...
+            tickCountUntilUpdate = TICKS_PER_UPDATE;
+
+            lastSentStatus = new PlayerStatus(
+                    pX,
+                    pY,
+                    pZ,
+                    ActionBarModel.getCurrentHealth(),
+                    ActionBarModel.getMaxHealth(),
+                    ActionBarModel.getCurrentMana(),
+                    ActionBarModel.getMaxMana());
```
Incompleteusern: Can't this and the class `PlayerStatus` just be removed and inlined?

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private static void sendPing() {
+        if (!isSocketOpen()) return;
+
+        hadesConnection.sendPacketAndFlush(new HCPacketPing(System.currentTimeMillis()));
+    }
+
+    @SubscribeEvent
+    public static void onFriendListUpdate(RelationsUpdateEvent.FriendList event) {
```
Incompleteusern: SocketModel implicitly depends on PlayerRelationsModel here, don't see a resolution besides moving code to SocketFeature

kristofbolyai: Or you just make sure a feature that depends on socket model also depends on relations model

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
kristofbolyai: We have a class called PlayerStatus?

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
Incompleteusern: We could have models depend on other models :)

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
Incompleteusern: A record

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
kristofbolyai: Ew. I think we should do what magicus wanted and merge all models into one and make them non-disablable.

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
kristofbolyai: lastSentStatus is used so client does not send data if it did not change. How do I do that if I remove the class?

Incompleteusern: Oh. I guess it could be used in `HadesUser` as well, but eh

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
```diff
@@ -0,0 +1,202 @@
...
+            if (lastSentStatus != null
+                    && lastSentStatus.equals(
+                            pX,
+                            pY,
+                            pZ,
+                            ActionBarModel.getCurrentHealth(),
+                            ActionBarModel.getMaxHealth(),
+                            ActionBarModel.getCurrentMana(),
+                            ActionBarModel.getMaxMana())) {
+                tickCountUntilUpdate = 1;
```
Incompleteusern: Shouldn't it reset `tickCountUntilUpdate` to `TICKS_PER_UPDATE` instead of `1`?

kristofbolyai: I made it so it resets to one so we check for updates every tick until there is one. Checking per tick is fine, sending data per tick is not. Hence why this is set to 1.

common/src/main/java/com/wynntils/sockets/model/SocketModel.java
magicus: I'm sorry I won't be able to work on that (at least for a very long time). But I whole-heartedly will support if you do it! :)

kristofbolyai:
@Incompleteusern @magicus @DevScyu 
This PR is ready for reviews and production.

common/src/main/java/com/wynntils/features/user/SocketFeature.java
```diff
@@ -0,0 +1,71 @@
...
+    public boolean getOtherPlayerInfo = true;
+
+    @Config
+    public boolean shareWithParty = true;
+
+    @Config
+    public boolean shareWithFriends = true;
+
+    @Config
+    public boolean shareWithGuild = true;
```
Incompleteusern: Comment out the config until its implemented

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -37,6 +39,21 @@ public class MapFeature extends UserFeature {
...
+    @Config(subcategory = "Remote Players")
+    public boolean renderRemotePartyPlayers = true;
+
+    @Config(subcategory = "Remote Players")
+    public boolean renderRemoteGuildPlayers = true;
+
+    @Config(subcategory = "Remote Players")
+    public HealthTexture remotePlayerHealthTexture = HealthTexture.a;
+
+    @Config(subcategory = "Remote Players")
```
Incompleteusern, line 54: What is subcategory

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
```diff
@@ -142,4 +142,14 @@ public static float getRenderZ(Poi poi, float mapCenterZ, float centerZ, float c
...
+
+    public static float getRenderX(float gameX, float mapCenterX, float centerX, float currentZoom) {
+        double distanceX = gameX - mapCenterX;
+        return (float) (centerX + distanceX * currentZoom);
+    }
+
+    public static float getRenderZ(float gameZ, float mapCenterZ, float centerZ, float currentZoom) {
+        double distanceZ = gameZ - mapCenterZ;
+        return (float) (centerZ + distanceZ * currentZoom);
+    }
```
Incompleteusern: This is identical to `getRenderX` and `getRenderZ` but not poi, make one of the methods call the other?

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -329,8 +344,9 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
                 if (hovered.hasStaticLocation()) {
                     CompassModel.setCompassLocation(new Location(hovered.getLocation()));
                 } else {
+                    Poi finalHovered = hovered;
```
Incompleteusern: Add final modifier? unnecessary, but might be better

common/src/main/java/com/wynntils/sockets/SocketClientHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    public void handleClearMutual(HSPacketClearMutual packet) {
+        HadesUserModel.removeUser(packet.getUser());
+    }
+
+    @Override
+    public void handleDisconnect(HSPacketDisconnect packet) {
+        WynntilsMod.info("Disconnected from HadesServer. Reason: " + packet.getReason());
+
+        if (WorldStateManager.onServer()) {
+            McUtils.sendMessageToClient(new TextComponent("[Wynntils/Artemis] Disconnected from HadesServer.")
```
Incompleteusern: Inconsistent format, others do not use prefix

kristofbolyai: We do use this elsewhere

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: I plan to implement them in the config gui, config annotation has it for a while now.

common/src/main/java/com/wynntils/gui/render/MapRenderer.java
kristofbolyai: I don't think this is even used now

common/src/main/java/com/wynntils/sockets/SocketClientHandler.java
kristofbolyai: Update reminder does this too.

DevScyu (dismissed):
LGTM

Incompleteusern (dismissed):

Incompleteusern (approved):

PR #587, opened 2022-10-21
InventoryEmeraldCountFeature: Fix categories showing up even though there is not enough emeralds

kristofbolyai:


common/src/main/java/com/wynntils/features/user/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -123,7 +123,7 @@ private void renderTexturedCount(PoseStack poseStack, int emeralds) {
 
         for (int i = emeraldAmounts.length - 1; i >= 0; i--) {
             String emeraldAmount = emeraldAmounts[i];
```
magicus: Why is this a string and not an integer?

kristofbolyai: It is formatted in decimal and stuff

magicus: So is the check for "0" always valid then? Maybe it is a french zero? ;-)

kristofbolyai: It should be, yes.

magicus: I think what I am saying is that we should not have two different ways to represent "no value".

magicus (approved):
Thanks!

PR #595, opened 2022-10-22
fix: Gammabright is now not changed incorrectly when opening settings screen

kristofbolyai:


common/src/main/java/com/wynntils/features/user/GammabrightFeature.java
```diff
@@ -27,26 +27,24 @@ public class GammabrightFeature extends UserFeature {
...
         if (event.getNewState() != WorldStateManager.State.WORLD) return;
-        if (!gammabrightEnabled) return;
 
-        lastGamma = McUtils.options().gamma;
-        McUtils.options().gamma = 1000d;
+        applyGammabright();
     }
 
-    private void onGammabrightKeyPress() {
+    private void applyGammabright() {
```
Incompleteusern, line 39: `applyGammabright` seems to be a toggle of `gammabrightEnabled`. As such, why are we calling it on world state change, which toggles the gamma as the world changes, and when the config is changed, which negates the change?

kristofbolyai: Right, thanks

common/src/main/java/com/wynntils/features/user/GammabrightFeature.java
```diff
@@ -50,6 +50,11 @@ private void applyGammabright() {
         ConfigManager.saveConfig();
     }
 
+    private void toggleGammaBright() {
+        gammabrightEnabled = !gammabrightEnabled;
```
Incompleteusern, line 51: You seem to have forgotten the toggling still present within applyGammabright itself

kristofbolyai: ape mode

Incompleteusern (approved):

PR #597, opened 2022-10-22
fix: Assert our overlay sections are current pre-render

### LINKED ISSUE #570: Overlays sometimes render off

kristofbolyai:
fixes https://github.com/Wynntils/Artemis/issues/570

### DIFF
--- a/common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
+++ b/common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
@@ -27,12 +27,14 @@ import net.minecraft.network.chat.TextComponent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class OverlayManager extends CoreManager {
+    private static final int OVERLAY_SECTIONS = 9; // must be a square number
+
     private static final Map<Overlay, OverlayInfo> overlayInfoMap = new HashMap<>();
     private static final Map<Overlay, Feature> overlayParent = new HashMap<>();
 
     private static final Set<Overlay> enabledOverlays = new HashSet<>();
 
-    private static final List<SectionCoordinates> sections = new ArrayList<>(9);
+    private static final List<SectionCoordinates> sections = new ArrayList<>(OVERLAY_SECTIONS);
 
     public static void registerOverlay(Overlay overlay, OverlayInfo overlayInfo, Feature parent) {
         overlayInfoMap.put(overlay, overlayInfo);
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -57,11 +59,40 @@ public static void enableOverlays(List<Overlay> overlays, boolean ignoreState) {
...
+
+        // this assumes sections are equal in size
+        Window window = McUtils.window();
+        SectionCoordinates section = sections.get(0);
+
+        // account for loss of pixels when section does not perfectly divide screen
+        int sqrt = (int) Math.sqrt(OVERLAY_SECTIONS);
+
+        if (Math.abs((section.x2() - section.x1()) * sqrt - window.getGuiScaledWidth()) >= sqrt) {
+            WynntilsMod.error("We sections with wrong width when trying render! Forcing recalculation.");
```
magicus: "This sentence no verb"...?

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -57,11 +59,40 @@ public static void enableOverlays(List<Overlay> overlays, boolean ignoreState) {
...
+            calculateSections();
+            return;
+        }
+
+        // this assumes sections are equal in size
+        Window window = McUtils.window();
+        SectionCoordinates section = sections.get(0);
+
+        // account for loss of pixels when section does not perfectly divide screen
+        int sqrt = (int) Math.sqrt(OVERLAY_SECTIONS);
```
Incompleteusern, line 81: Why do we use square roots?

kristofbolyai: 9 sections = 3 sections per row or  per column

Incompleteusern: I think that just makes it more complicated and can be simplified.

kristofbolyai: I mean it was like this before

Incompleteusern: I'd rather we go
```java
private static final int SECTION_SIDE_NUMBER = 3;

private static final List<SectionCoordinates> sections = new ArrayList<>(SECTION_SIDE_NUMBER * SECTION_SIDE_NUMBER);
```

Incompleteusern:
I think instead of asserting every time prerender we can just hook post `Window#setGuiScale` with DisplayResizeEvent

kristofbolyai:
> I think instead of asserting every time prerender we can just hook post `Window#setGuiScale` with DisplayResizeEvent

How would that fix anything? We already post the event on return

Incompleteusern:
> > I think instead of asserting every time prerender we can just hook post `Window#setGuiScale` with DisplayResizeEvent
> 
> How would that fix anything? We already post the event on return

It detects when the guiScaledWidth/guiScaledHeight are updated, but that already is called by `DisplayResizeEvent` as you mention. I guess it is slightly "closer" to the gui scale modification.

However, this raises the question of how this assertion fixes the problem when the recalculation on resize does not

Incompleteusern:
Furthermore, the titlescreeninit calculation seems unnecessary as `resizeDisplay` is also called during `Minecraft`'s initialization

PR #598, opened 2022-10-22
feat: /compass supports non-standard coordinates

kristofbolyai:


common/src/main/java/com/wynntils/mc/utils/LocationUtils.java
```diff
@@ -0,0 +1,30 @@
...
+package com.wynntils.mc.utils;
+
+import com.wynntils.mc.objects.Location;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class LocationUtils {
+    private static final Pattern COORDINATE_PATTERN =
+            Pattern.compile("(?<x>[-+]?\\d+)(\\D{0,5}(?<y>[-+]?\\d+))?\\D{0,5}(?<z>[-+]?\\d+)");
```
magicus: Is this from legacy, or a re-implementation?

(Fun fact: allowing `/compass` to take a more relaxed approach on coordinates was actually my very first contribution to Wynntils!)

kristofbolyai: I crafted this myself. Do you think the legacy one is better?

magicus: I dunno. I know I tweaked it a couple of times, whenever I ran into issues, such as not being able to copy/paste coordinates from the wiki. (Otoh, with the hopefully upcoming `/locate npc`, the need to to this is not as great).

Regexps are really write-only code. :-( There's no way I could compare your regexp and mine, and say which one is better. They do a lot of wonderful things, but they are suck like *%#$& when you should try to understand them.

kristofbolyai: This regex is basically x(5 non digits max){{y(5 non digits max)}}z

where the {{}} part is optional so x(5 non digits max)z works too.

kristofbolyai: Do you think this is fine to merge though?

magicus: Yes.

magicus (approved):

PR #601, opened 2022-10-24
fix: Fix GammabrightFeature toggle

kristofbolyai:
@Incompleteusern


common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -113,6 +113,8 @@
   "feature.wynntils.gameNotificationOverlay.overlay.gameNotification.textShadow.name": "Text Shadow",
   "feature.wynntils.gammabright.gammabrightEnabled.description": "Should gamma be set to a really high value to achieve night-vision like effect?",
   "feature.wynntils.gammabright.gammabrightEnabled.name": "Gammabright",
+  "feature.wynntils.gammabright.lastGamma.description": "Last gamma used by Minecraft",
```
magicus, line 116: Do we need i18n for non-visible config options? That does not seem to make any sense. If our assert was triggered, it needs to be adjusted to ignore non-visible configs, methinks.

kristofbolyai: This is the case for a while now. We could change this, yes.

PR #603, opened 2022-10-24
feat: Add /feature reload

kristofbolyai:


magicus (approved):

Incompleteusern:
`setUserEnabled` should not be set, this is imo a structural flaw. User Enabled should be more of a condition that is set.

kristofbolyai:
> `setUserEnabled` should not be set, this is imo a structural flaw. User Enabled should be more of a condition that is set.

Perhaps. Will fix this tmrw

Incompleteusern:
I think this is actually what the `Condition` system was made for, as a set of predicates that need to be satisfied to enable a feature.

kristofbolyai:
> I think this is actually what the `Condition` system was made for, as a set of predicates that need to be satisfied to enable a feature.

you wanna treat user enabled as a condition? otherwise i am not sure what you are getting at.

Incompleteusern:
> > I think this is actually what the `Condition` system was made for, as a set of predicates that need to be satisfied to enable a feature.
> 
> you wanna treat user enabled as a condition? otherwise i am not sure what you are getting at.

yes if possible

magicus:
I'm slightly worried that the Condition system is overly broad. It seems to bite off more than it can chow, abstraction-wise. Changing a very concrete thing like "user enabled" into a general "condition" seems like a step backwards.

PR #604, opened 2022-10-25
fix: /feature reload should not modify userEnabled

kristofbolyai:


common/src/main/java/com/wynntils/commands/FeatureCommand.java
```diff
@@ -225,8 +225,7 @@ private int reloadFeature(CommandContext<CommandSourceStack> context) {
             return 1;
         }
 
-        feature.setUserEnabled(false);
```
magicus, line 228: I still don't get it with all these different ways of enabling/disabling features. I thought Incomplete was about to refactor that? But it got stalled, or lost?

magicus (approved):

common/src/main/java/com/wynntils/commands/FeatureCommand.java
kristofbolyai: Iirc it is upcoming

PR #608, opened 2022-10-26
fix: Remove all static configs

kristofbolyai:
I would like @magicus' opinion on this. 

Since we clearly decided that features are instance based, let's make all config fields respect that. Also, I removed all instance getters, I think public INSTANCE fields are much more readable.


magicus:
+1 on removing static from the config fields.

I'm in general not really for public fields, but if we treat `INSTANCE` as a special case and not as a precedent, then I'm fine with it.

magicus (approved):

PR #618, opened 2022-10-29
fix(HighlightMatchingItemsFeature): Search now persist through page switching on the same screen

kristofbolyai:


Incompleteusern (commented):
Why was the textbox clearing if only the container contents was changing?

kristofbolyai:
> Why was the textbox clearing if only the container contents was changing?

I think because the screen re-inited.

Incompleteusern:
If I understand correctly, one container instance can switch between multiple screens?

kristofbolyai:
> If I understand correctly, one container instance can switch between multiple screens?

Yeah. I think page switching opens a new screen (and closes the old one)

Incompleteusern (approved):

PR #619, opened 2022-10-29
feat: Add auto search to bank and guild member list

kristofbolyai:
Guild bank is weird and uses a lot of set slot packets so this won't work there.

This PR depends on https://github.com/Wynntils/Artemis/pull/618


common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    private void forceUpdateSearch() {
+        Screen screen = McUtils.mc().screen;
+        if (lastSearchWidget != null && screen instanceof AbstractContainerScreen<?> abstractContainerScreen) {
+            matchItems(lastSearchWidget.getTextBoxInput(), abstractContainerScreen);
+        }
+    }
+
+    public enum ContainerType {
```
magicus: This seems too general to be put in a Feature. Move to model or a util class.

kristofbolyai: I moved the enum to not be an inner class, but imo the feature itself can keep the other code, since it is generalized enough.

kristofbolyai: (Even the enum feels like generalized enough since I renamed it to SearchableContainerType)

common/src/main/java/com/wynntils/features/user/ContainerSearchFeature.java
```diff
@@ -121,34 +121,41 @@ private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
...
+                currentSearchableContainerType.getNextItemSlot(),
                 abstractContainerScreen.getMenu().containerId,
                 GLFW.GLFW_MOUSE_BUTTON_LEFT,
                 abstractContainerScreen.getMenu().getItems());
     }
 
-    private ContainerType getCurrentContainerType(String title) {
-        if (filterInBank && ContainerType.BANK.getTitlePattern().matcher(title).matches()) {
-            return ContainerType.BANK;
+    private SearchableContainerType getCurrentSearchableContainerType(String title) {
```
magicus, line 139: Can't you move this as well to the SearchableContainerType class, and just keep the `filterInBank` etc checks here?

magicus: Also, that means you don't have to expose the pattern from the ContainerType class

magicus (approved):

PR #620, opened 2022-10-29
Add dynamic compass location support

Incompleteusern:


common/src/main/java/com/wynntils/features/user/overlays/QuestInfoOverlayFeature.java
```diff
@@ -138,6 +147,7 @@ public void render(PoseStack poseStack, float partialTicks, Window window) {
 
             if (toRender.get(0).getText() == null) {
                 // Set at first use; I18n is not available at initialization time
+                // TODO check if above comment still valid
```
kristofbolyai: It is not. This comment was the reason features were made to init after i18n

common/src/main/java/com/wynntils/features/user/overlays/QuestInfoOverlayFeature.java
```diff
@@ -61,14 +62,22 @@ public void onScoreboardSegmentChange(ScoreboardSegmentAdditionEvent event) {
...
 
-        Optional<Location> location = event.getQuestInfo().getNextLocation();
-        if (location.isEmpty()) return;
+        return location.get();
+    };
 
-        CompassModel.setCompassLocation(location.get());
+    @SubscribeEvent
+    public void onTrackedQuestUpdate(TrackedQuestUpdateEvent event) {
+        // set if valid
```
kristofbolyai, line 64: when is it not valid?

common/src/main/java/com/wynntils/features/user/overlays/QuestInfoOverlayFeature.java
```diff
@@ -61,14 +62,22 @@ public void onScoreboardSegmentChange(ScoreboardSegmentAdditionEvent event) {
         }
     }
 
-    @SubscribeEvent
-    public void onTrackedQuestUpdate(TrackedQuestUpdateEvent event) {
-        if (event.getQuestInfo() == null || !autoTrackQuestCoordinates) return;
+    private static final Supplier<Location> supplier = () -> {
```
kristofbolyai: Debatable if this should be stored here or in QuestManager. Let's move this to QuestManager and still set the compass here in the event?

common/src/main/java/com/wynntils/wynn/model/CompassModel.java
```diff
@@ -5,39 +5,73 @@
...
+        if (supplier == null) return;
+
+        Location newLocation = supplier.get();
+
+        if (newLocation == null) { // drop location
+            reset();
+        } else if (compassLocation != newLocation) { // update location
+            compassLocation = newLocation;
+
+            if (McUtils.mc().level != null) {
```
kristofbolyai, line 36: I am pretty sure client tick only happens if there is a level. But we can keep this.

kristofbolyai (commented):
IMO move the quest supplier to quest manager, then this is good to go

kristofbolyai (dismissed):

kristofbolyai:
Actually, @Incompleteusern this might not be the right step to make. Why not make Pois have a supplier for location? That way, you can turn Poi into an interface too, just like we discussed in discord earlier

common/src/main/java/com/wynntils/wynn/model/CompassModel.java
```diff
@@ -5,39 +5,73 @@
...
 
             return Optional.of(waypointPoi);
         }
 
         return Optional.empty();
     }
 
-    public static void setCompassLocation(Location compassLocation) {
-        CompassModel.compassLocation = compassLocation;
+    public static void setDynamicCompassLocation(Supplier<Location> compassSupplier) {
```
kristofbolyai, line 63: Why don't we use our new Poi interface and implement WaypointPoi as a "special" POI? That way you can get rid of this supplier weirdness and directly support it in the class.

magicus (commented):
I just skimmed through it, found nothing that looks out of the ordinary, but I have not done a full review (no time for that, sorry).

PR #621, opened 2022-10-30
feat: Add ShamanMasksOverlayFeature

kristofbolyai:
image


common/src/main/java/com/wynntils/wynn/model/ShamanMaskModel.java
```diff
@@ -0,0 +1,67 @@
...
+    public static void onTitle(SubtitleSetTextEvent event) {
+        String title = ComponentUtils.getCoded(event.getComponent());
+
+        if (title.contains("Mask of the ") || title.contains("➤")) {
+            parseMask(title);
+
+            if (ShamanMasksOverlayFeature.INSTANCE.hideMaskTitles
+                    && ShamanMasksOverlayFeature.INSTANCE.shamanMaskOverlay.isEnabled()) {
+                event.setCanceled(true);
+            }
```
Incompleteusern, line 36: Use an event and cancel it in the overlay itself

kristofbolyai: Why? We do stuff like this elsewhere I am pretty sure. I guess I could duplicate the event but it would be just code copying.

Incompleteusern: CustomBarsOverlayFeature uses an event for cancelling. This also breaks the structural barrier between features and models, which I'd like to preserve when possible.

Incompleteusern (approved):

PR #623, opened 2022-10-31
Unbind unregistered keybinds

### LINKED ISSUE #580: Unregistering keybinds will still make then "conflict"
image

Incompleteusern:
Fixes #580


kristofbolyai:
@Incompleteusern is this a draft because you did not get to test it?

Incompleteusern:
Yes, I have not gotten around to testing it yet

common/src/main/java/com/wynntils/core/keybinds/KeyBind.java
```diff
@@ -43,6 +43,10 @@ public KeyBind(String name, int keyCode, boolean firstPress, Runnable onPress, C
     public KeyBind(String name, int keyCode, InputConstants.Type type, boolean firstPress, Runnable onPress) {
         this.firstPress = firstPress;
         this.keyMapping = new KeyMapping(name, type, keyCode, CATEGORY);
+
+        // Unbound by default
+        keyMapping.setKey(InputConstants.UNKNOWN);
+        KeyMapping.resetMapping();
```
kristofbolyai: Why? Don't we want to actually register the keybind? This feels smth like "@StartDisabled" for keybinds. I think the changes in `unregisterKeybind` should be enough

Incompleteusern: This should unbind the keybind, and if the keybind is registered it should be bound by the options loading later.

Incompleteusern: Actually, registering keybind should bind it too, oops

kristofbolyai:
Technically this still needs testing/fixing? We need to see if KeyMapping constructor does weird behavior

kristofbolyai:
@Incompleteusern Can you prioritize this whenever you are free?

kristofbolyai (approved):

PR #624, opened 2022-10-31
feat: Redirect tool/crafted durability messages

kristofbolyai:


common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -246,6 +258,36 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
...
+            if (toolDurability != FilterType.KEEP) {
+                Matcher matcher = NO_TOOL_DURABILITY_PATTERN.matcher(uncoloredMsg);
+                if (matcher.find()) {
+                    e.setCanceled(true);
+                    if (toolDurability == FilterType.HIDE) {
+                        return;
+                    }
+
+                    NotificationManager.queueMessage(
+                            new TextComponent("Your tool has 0 durability!").withStyle(ChatFormatting.DARK_RED));
```
magicus, line 271: I think we've done it like this for other redirects, but these messages should really be i18n'd.

magicus (approved):

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
kristofbolyai: https://github.com/Wynntils/Artemis/issues/626

PR #625, opened 2022-11-01
Replace depth mask with stencil mask

### LINKED ISSUE #528: Artemis minimap circle mask overrides Wynnlib side indicator
Besides being a conflict, this is also indicative that our depth mask code isn't well enough.
image

Incompleteusern:
Should fix #528


fabric/src/main/java/com/wynntils/fabric/mixins/RenderTargetMixin.java
```diff
@@ -0,0 +1,57 @@
...
+            method = "createBuffers",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lcom/mojang/blaze3d/platform/GlStateManager;_texImage2D(IIIIIIIILjava/nio/IntBuffer;)V",
+                            ordinal = 0,
+                            remap = false))
+    public void init(Args args) {
+        if (RenderUtils.isStencilEnabled()) {
```
magicus: I'm not really fond of this. I fear it has a high risk of conflict with other mods. Is there some other way you can go about to achieve what you need? (This goes for both `@ModifyArgs`)

quilt/src/main/resources/wynntils.mixins.quilt.json
```diff
@@ -0,0 +1,14 @@
+{
+  "client": [
+    "RenderTargetMixin"
+  ],
+  "compatibilityLevel": "JAVA_17",
```
magicus, line 5: What was the baseline before? Or is this specified in multiple places?

Incompleteusern: This is copy pasted from other places

Incompleteusern:
image

Incompleteusern:
Still needs testing that it works on quilt and forge (mainly forge).

fabric/src/main/java/com/wynntils/fabric/mixins/RenderTargetMixin.java
Incompleteusern: I doubt there will be much conflict, as I feel any other modification would be also for enabling stencil.

kristofbolyai:
Forge works fine. 

Fabric/Quilt:
image

fabric/src/main/java/com/wynntils/fabric/mixins/RenderTargetMixin.java
kristofbolyai: We agreed to merge this on the basis that if this creates problems, we revert it

PR #630, opened 2022-11-01
fix: Fix not being able to change overlay settings + animation

kristofbolyai:


common/src/main/java/com/wynntils/core/features/AbstractConfigurable.java
```diff
@@ -31,7 +31,7 @@ public final List<ConfigHolder> getConfigOptions() {
 
     @Override
     public final Optional<ConfigHolder> getConfigOptionFromString(String name) {
-        return getVisibleConfigOptions().stream()
+        return getConfigOptions().stream()
```
kristofbolyai, line 34: Features did visible config options, overlays did all config options. Everywhere this is used, we need to filter from all config options.

Incompleteusern (approved):

PR #631, opened 2022-11-01
Stencils and support for compass distance

### LINKED ISSUE #528: Artemis minimap circle mask overrides Wynnlib side indicator

Incompleteusern:
Need to get it actually to function correctly.

Fixes #528


Incompleteusern:
Should be ready to review and be tested. If wanted I can add a config to get the old depth based impl.

kristofbolyai:
Can we cherry pick https://github.com/Wynntils/Artemis/pull/631/commits/8f4616f7a3bdc9e286325cd524a749ea5dcd1af2 to be a separate PR? If we ever need to revert this again. Also, it is just better to not mix up PRs.

kristofbolyai (commented):
The stencil thing looks good? I am not sure, we can't see any problems in this code straight away. However, I do wonder, what did you change since last time?

Incompleteusern:
I think the old impl only write the stencil buffer once and never clears it, which should? be fixed now

kristofbolyai:
> I think the old impl only write the stencil buffer once and never clears it, which should? be fixed now

Oh. Ill separate the PRs for you and merge this I guess.

fabric/src/main/java/com/wynntils/fabric/mixins/RenderTargetMixin.java
```diff
@@ -0,0 +1,53 @@
...
+            method = "createBuffers",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lcom/mojang/blaze3d/platform/GlStateManager;_texImage2D(IIIIIIIILjava/nio/IntBuffer;)V",
+                            ordinal = 0,
+                            remap = false))
+    public void init(Args args) {
+        args.set(2, GL30.GL_DEPTH32F_STENCIL8);
```
magicus, line 31: What will happen with this if we're not on Wynncraft? Can it interact badly with other mods, or plain Vanilla behavior elsewhere?

forge/src/main/java/com/wynntils/forge/WynntilsModForge.java
```diff
@@ -23,6 +24,9 @@ public WynntilsModForge() {
 
         File modFile = new File(path.toUri());
 
+        // Enable stencil support
+        Minecraft.getInstance().getMainRenderTarget().enableStencil();
```
magicus, line 28: So in Forge there's a simple call for this?! :-o What does that do? The same as what you do "manually" in Fabric?

Incompleteusern: Yeah, basically

fabric/src/main/java/com/wynntils/fabric/mixins/RenderTargetMixin.java
kristofbolyai: @Incompleteusern perhaps return if not on server here? What do you think?

Incompleteusern: This is identical to forge/burger's impl, and I don't know enough about rendering to see if it will work when returning not on server

kristofbolyai (approved):

Incompleteusern:
Going to try to merge and hope no conflicts 🙏

PR #635, opened 2022-11-02
feat: Add auto publishing

DevScyu:


kristofbolyai:
How do we parse build number from new versioning?

kristofbolyai:
Also we need to fix auto updater here

.github/.config.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+const config = require("conventional-changelog-conventionalcommits");
+
+module.exports = config({
+    types: [
+        // Unhide all types except "ci" so that they show up on generated changelog
```
magicus, line 6: I don't get it. We only want to show "feature" and "fix" on the changelog, right? All other is of no interest to users.

.github/.config.js
```diff
@@ -0,0 +1,22 @@
...
+        { type: "revert", section: "Reverts" },
+        { type: "docs", section: "Documentation" },
+        { type: "style", section: "Styles" },
+        { type: "chore", section: "Miscellaneous Chores" },
+        { type: "refactor", section: "Code Refactoring" },
+        { type: "test", section: "Tests" },
+        { type: "build", section: "Build System" },
+        { type: "ci", section: "Continuous Integration", hidden: true },
+    ],
+});
```
magicus: And the missing newline. Maybe fix your IDE settings so it is automatically added..?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,146 @@
...
+          token: ${{ secrets.PRIVATE_TOKEN }}
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm install conventional-changelog-conventionalcommits
+
+      - name: Create changelog
+        id: changelog
+        uses: DevScyu/conventional-changelog-action@v4
```
magicus, line 31: Is this an action you've written yourself, or a modified fork from some upstream project? Why not use upstream if that is the case? Otherwise, if we really need a special tool or a modified version (I hoghly recommend against it -- even more maintenance burden!), it should move into the Wynntils area on github.

magicus:
@DevScyu Please don't ever force push (rebase) to an open PR. That makes reviewing next to impossible. After a PR is opened, just use merge. Thanks.

kristofbolyai:
> @DevScyu Please don't ever force push (rebase) to an open PR. That makes reviewing next to impossible. After a PR is opened, just use merge. Thanks.

It's because of the build system.

.github/.config.js
DevScyu: It just gives us more flexibility on what we can add to the changelog

DevScyu:
> @DevScyu Please don't ever force push (rebase) to an open PR. That makes reviewing next to impossible. After a PR is opened, just use merge. Thanks.

I have to force push due to this branch auto building and making a new release version

CHANGELOG.md
```diff
@@ -0,0 +1,86 @@
+## [0.0.1-alpha.0](https://github.com/Wynntils/Artemis/compare/v0.0.0...v0.0.1-alpha.0) (2022-11-02)
```
magicus: Do we need a changelog file? To me it seems like just a burden to have it updated all the time. *Especially* if we now have commits with the markers that mean we can generate it automatically.

The one thing I can think of as an exception is if we want to include free text overviews, like we have done in Legacy. "Hi everyone, In this release we have focused on the ...." etc.

.github/workflows/pre-release.yml
DevScyu: This is a fork of https://github.com/TriPSs/conventional-changelog-action with changes to support pre-releases.

I have submitted a PR upstream here: https://github.com/TriPSs/conventional-changelog-action/pull/186

magicus:
Maybe set this as a draft until you are ready to receive reviews?

CHANGELOG.md
DevScyu: It will auto update from the build system with the last `5` version changelogs by default. Take a look at https://github.com/Wynntils/Wynntils/blob/development/CHANGELOG.md

.github/.config.js
kristofbolyai: I would rather show these than no changes tbf

PR #637, opened 2022-11-02
feat: Add auto publishing

DevScyu:
Like legacy version format follows semver.

Currently only `alpha` branch is supported, and are marked as alpha builds on Curseforge and Modrinth.

First version will be marked as: `v0.0.1-alpha.0` with each build incrementing the final number, until we make a "release" version, which would then increase the minor / patch depending on commits.

### DIFF
--- a/common/src/main/java/com/wynntils/core/managers/UpdateManager.java
+++ b/common/src/main/java/com/wynntils/core/managers/UpdateManager.java
@@ -54,8 +54,7 @@ public class UpdateManager extends CoreManager {
             InputStreamReader stInputReader = new InputStreamReader(st.getInputStream(), StandardCharsets.UTF_8);
             JsonObject jsonObject = JsonParser.parseReader(stInputReader).getAsJsonObject();
 
-            String latestMd5 =
-                    jsonObject.getAsJsonPrimitive("md5").getAsString();
+            String latestMd5 = jsonObject.getAsJsonPrimitive("md5").getAsString();
 
             String currentMd5 = getCurrentMd5();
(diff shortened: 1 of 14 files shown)


common/src/main/java/com/wynntils/core/managers/UpdateManager.java
```diff
@@ -4,143 +4,93 @@
...
-                    continue;
-                }
+        try {
+            URLConnection st = WebManager.generateURLRequest(LAST_BUILD_CHECK_PATH);
+            InputStreamReader stInputReader = new InputStreamReader(st.getInputStream(), StandardCharsets.UTF_8);
+            JsonObject jsonObject = JsonParser.parseReader(stInputReader).getAsJsonObject();
 
-                int buildNumber = Integer.parseInt(matcher.group(3));
+            String latestMd5 =
+                    jsonObject.getAsJsonArray("md5").getAsJsonPrimitive().getAsString();
```
kristofbolyai: getAsJsonPrimitive(md5) (and remove the array one)

common/src/main/java/com/wynntils/core/managers/UpdateManager.java
```diff
@@ -4,143 +4,93 @@
...
+                return future;
             }
-        }
 
-        if (foundArtifact == null) {
-            WynntilsMod.warn("Tried to update, but could not find target artifact.");
-            return new Pair<>(null, false);
-        }
+            String latestDownload =
+                    jsonObject.getAsJsonArray("url").getAsJsonPrimitive().getAsString();
```
kristofbolyai: same

PR #638, opened 2022-11-02
fix: BeaconBeamFeature not rendering correctly

kristofbolyai:


Incompleteusern:
How does this fix the issue?

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
```diff
@@ -32,14 +33,16 @@ public List<Class<? extends Model>> getModelDependencies() {
...
+        if (compassLocation.isEmpty()) return;
 
         PoseStack poseStack = event.getPoseStack();
         MultiBufferSource.BufferSource bufferSource =
                 McUtils.mc().renderBuffers().bufferSource();
 
         Vec3 camera = event.getCamera().getPosition();
-        Location location = CompassModel.getCompassLocation().get();
+        Location location = compassLocation.get();
+        location.set(location.x, 0, location.z);
```
magicus: Will this modify the underlying representation, or do we return a copy?

kristofbolyai:
> How does this fix the issue?

y is -1 otherwise, setting to 0 actually renders it.

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
kristofbolyai: It modifies the representation. Why?

Incompleteusern:
> > How does this fix the issue?
> 
> y is -1 otherwise, setting to 0 actually renders it.

Why is y -1?

kristofbolyai:
> > > How does this fix the issue?
> > 
> > 
> > y is -1 otherwise, setting to 0 actually renders it.
> 
> Why is y -1?

Because some POIs only have a x and z coordinate.

Incompleteusern:
> 

Which POIs? and shouldn't we be applying a fix somewhere else.

Incompleteusern:
```java
        if (locationSupplier != null && locationSupplier.get() != null) {
            WaypointPoi waypointPoi = new WaypointPoi(() -> {
                Location location = locationSupplier.get();

                // Make sure to always render on top
                location.set(location.x, Double.MAX_VALUE, location.z);

                return MapLocation.fromLocation(location);
            });

            return Optional.of(waypointPoi);
        }
```

I think this is responsible as it modifies the underlying representation, also the setting to max double is likely unnecessary now.

common/src/main/java/com/wynntils/wynn/model/CompassModel.java
```diff
@@ -48,9 +48,6 @@ public static Optional<WaypointPoi> getCompassWaypoint() {
             WaypointPoi waypointPoi = new WaypointPoi(() -> {
                 Location location = locationSupplier.get();
 
-                // Make sure to always render on top
-                location.set(location.x, Double.MAX_VALUE, location.z);
-
```
kristofbolyai, line 53: This was the problem, we did not even need this as rendering code already handled it. @Incompleteusern

Incompleteusern: Thanks!

Incompleteusern (approved):

common/src/main/java/com/wynntils/features/user/BeaconBeamFeature.java
magicus: I'm not entirely happy about the compass location API. It feels too open and leaky. I realize there's a difference between locations that are fixed (like NPCs) and that can change (like compass point), but the API should be more restrictive here, I think. I noticed you and Incomplete did some stuff on static vs dynamic poi's that I've not had the time to look into detail. Maybe it's the beginning on getting this more strict.

Also, we should perhaps consider having a location representation where we can actually flag the Y value as "don't know/don't care".

common/src/main/java/com/wynntils/wynn/model/CompassModel.java
magicus: I've been saying for a long time that this was fishy code. ;-)

magicus (approved):

PR #642, opened 2022-11-02
fix: Increase query timeout to make queries work when Wynncraft lags

kristofbolyai:


Incompleteusern (approved):
More testing of releases

kristofbolyai:
This is actually needed. Wynn is so laggy for me these days, no idea why.

PR #643, opened 2022-11-02
feat: Render guild territories on main map

kristofbolyai:
image

### DIFF
--- a/common/src/main/java/com/wynntils/core/webapi/profiles/TerritoryProfile.java
+++ b/common/src/main/java/com/wynntils/core/webapi/profiles/TerritoryProfile.java
@@ -149,8 +149,11 @@ public class TerritoryProfile {
             String friendlyName = territoryName.replace('’', '\'');
 
             String guild;
-            if (territory.get("guild").isJsonNull()) guild = "Unknown";
-            else guild = territory.get("guild").getAsString();
+            if (territory.get("guild").isJsonNull()) {
+                guild = "Unknown";
+            } else {
+                guild = territory.get("guild").getAsString();
+            }
 
             Date acquired = null;
             try {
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -340,7 +345,7 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
                 return true;
             }
 
-            if (hovered != null) {
+            if (hovered != null && !(hovered instanceof TerritoryPoi)) {
```
Incompleteusern, line 348: Maybe move to onClick code to the hovered poi itself?

kristofbolyai: We could eventually. Didn't want to refactor Pois even more.

kristofbolyai: I dont see that many use cases though

Incompleteusern: Seaskipper map will have custom behavior on hover click, but yeah.

common/src/main/java/com/wynntils/core/webapi/profiles/TerritoryProfile.java
```diff
@@ -164,16 +167,21 @@ public TerritoryProfile deserialize(JsonElement json, Type typeOfT, JsonDeserial
             }
 
             String guildPrefix;
-            if (territory.get("guildPrefix").isJsonNull()) guildPrefix = "UNK";
-            else guildPrefix = territory.get("guildPrefix").getAsString();
+            if (territory.get("guildPrefix").isJsonNull()) {
+                guildPrefix = "UNK";
```
Incompleteusern: Could this be an actual guild prefix? Maybe a prefix of UNKWN or something would be better

Incompleteusern: I know it differs from legacy

kristofbolyai: Fixed, shouldnt happen though.

magicus (approved):

magicus:
The use of <kbd>control</kbd> is lacking a bit in discoverability. :-) Maybe we'll need some kind of legend/help text for the map soon. Maybe a `?` icon that when hovered over shows a tooltip with help..? That can be both non-intrusive but also clearly accessible for new users. (Not requesting this as part of this PR, just throwing ideas randomly around me)

kristofbolyai:
> The use of <kbd>control</kbd> is lacking a bit in discoverability. :-) Maybe we'll need some kind of legend/help text for the map soon. Maybe a `?` icon that when hovered over shows a tooltip with help..? That can be both non-intrusive but also clearly accessible for new users. (Not requesting this as part of this PR, just throwing ideas randomly around me)

Yeah, I just ported legacy stuff. We can work on this when we add gui for map.

PR #646, opened 2022-11-03
feat: Add Discoveries Screen

kristofbolyai:
image
image
image


kristofbolyai:
image
image

common/src/main/java/com/wynntils/gui/screens/WynntilsDiscoveriesScreen.java
```diff
@@ -0,0 +1,377 @@
...
+        elements.addAll(Stream.concat(
+                        webDiscoveryInfoCache.stream()
+                                .filter(discoveryInfo -> switch (discoveryInfo.getType()) {
+                                    case TERRITORY -> showUndiscoveredTerritory;
+                                    case WORLD -> showUndiscoveredWorld;
+                                    case SECRET -> showUndiscoveredSecrets;
+                                })
+                                .filter(discoveryInfo -> DiscoveryManager.getAllDiscoveries()
+                                        .noneMatch(
+                                                discovery -> discovery.getName().equals(discoveryInfo.getName()))),
```
Incompleteusern, line 363: What is this code for?

common/src/main/java/com/wynntils/wynn/model/discoveries/DiscoveryManager.java
```diff
@@ -0,0 +1,72 @@
...
+        discoveries = newDiscoveries;
+
+        WynntilsMod.info("Discovered " + discoveries.size() + " discoveries");
+        WynntilsMod.postEvent(new DiscoveriesUpdatedEvent.Normal());
+    }
+
+    public static void setSecretDiscoveries(List<DiscoveryInfo> newDiscoveries) {
+        secretDiscoveries = newDiscoveries;
+
+        WynntilsMod.info("Discovered " + secretDiscoveries.size() + " secret discoveries");
```
Incompleteusern: Debug?

common/src/main/java/com/wynntils/wynn/model/discoveries/objects/DiscoveryInfo.java
```diff
@@ -0,0 +1,146 @@
...
+    public DiscoveryInfo(DiscoveryProfile discoveryProfile) {
+        this.name = discoveryProfile.getName();
+        this.type = DiscoveryType.valueOf(discoveryProfile.getType().toUpperCase(Locale.ROOT));
+        this.description = "";
+        this.minLevel = discoveryProfile.getLevel();
+        this.guildTerritory = TerritoryManager.getTerritories().get(name);
+        this.discovered = false;
+        this.requirements = discoveryProfile.getRequirements();
+
+        generateLore();
```
Incompleteusern: generate lazily?

common/src/main/java/com/wynntils/gui/screens/WynntilsDiscoveriesScreen.java
kristofbolyai: Api has every discovery, we make sure to not display them twice (if we have parsed it from quest book)

common/src/main/java/com/wynntils/wynn/model/discoveries/DiscoveryManager.java
kristofbolyai: I can remove them. It can also stay imo

common/src/main/java/com/wynntils/wynn/utils/DiscoveryUtils.java
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.utils;
```
Incompleteusern: This does not really belong in `wynn.utils`. `openDiscoveryOnMap`, `setDiscoveryCompass` can be inlined into `DiscoveryManager`. I feel `locateSecretDiscovery` should preferably be in `WebManager`

common/src/main/java/com/wynntils/gui/widgets/DiscoveryFilterButton.java
```diff
@@ -0,0 +1,134 @@
...
+                    1,
+                    texture.width(),
+                    texture.height(),
+                    0,
+                    0,
+                    texture.width(),
+                    texture.height(),
+                    texture.width(),
+                    texture.height());
+            ;
```
Incompleteusern: ?

common/src/main/java/com/wynntils/wynn/model/discoveries/objects/DiscoveryInfo.java
Incompleteusern: Might also help with requirements I think

common/src/main/java/com/wynntils/wynn/utils/DiscoveryUtils.java
kristofbolyai: Maybe not wynn.utils, but imo definitely worth a util class

common/src/main/java/com/wynntils/wynn/model/discoveries/objects/DiscoveryInfo.java
```diff
@@ -0,0 +1,146 @@
...
+    private final String name;
+    private final DiscoveryType type;
+    private final String description;
+    private final int minLevel;
+    private final TerritoryProfile guildTerritory;
+    private final boolean discovered;
+    private final List<String> requirements;
+    private List<Component> displayLore;
+
+    public DiscoveryInfo(DiscoveryProfile discoveryProfile) {
```
Incompleteusern, line 32: I don't see why we can't merge this with `DiscoveryProfile`

kristofbolyai: DiscoveryProfile is for GSON, merging it would be messy imo

common/src/main/java/com/wynntils/wynn/utils/DiscoveryUtils.java
Incompleteusern: I don't see much reuse of this code, unless you are planning something akin to `/discovery` In that case, I still feel `locateSecretDiscovery` should be moved

Incompleteusern (approved):

common/src/main/java/com/wynntils/mc/mixin/MinecraftMixin.java
```diff
@@ -26,6 +28,13 @@ private void setScreenPostPost(Screen screen, CallbackInfo ci) {
     @Inject(method = "tick", at = @At("HEAD"))
     private void tickPre(CallbackInfo ci) {
         EventFactory.onTickStart();
+
+        Queue<Runnable> runnableQueue = MinecraftSchedulerManager.getQueue();
```
magicus, line 32: I'm thinking this could be done by the MinecraftScheduleManager itself, by listening to onTickStart event? 

Otoh... That makes it always call, even if it does not have something to do. Was this the reason?

kristofbolyai: Reason was I needed to run this on  the game thread. Now I realize I could have used an event in the manager as well. Maybe it is better to inject this here

magicus: If there's anything that worries me about performance, then it's stuff done for each tick, or for each render call. 

So design-wise I think the manager should have had an event listener. But, that will make the event handler called at each tick, regardless of if it has anything to do (as it will be most time) or not. That could be mitigated by registering the manager if a runnable is enqueued, and unregistering it when the queue gets empty. But that is also overhead, and will not really fit our design either. 

So I guess this is for the best (even if untested assumptions about performance is the "premature optimization" that is the root of all evil...). If anything, the code should be broken out into a separate method in MinecraftSchedulerManager like `onTick()`, so we don't have to expose the queue and the internal logic.

PR #647, opened 2022-11-03
rainbow color

YouSoPugly:


common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -20,6 +21,7 @@ public final class FontRenderer {
     private final Font font;
 
     private static final int NEWLINE_OFFSET = 10;
+    private static final int CHAR_SPACING = 0;
```
kristofbolyai, line 25: I dont think this is useful? I assume you used this in testing or smth

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -48,7 +50,10 @@ public void renderText(
 
         if (text == null) return;
 
-        // TODO: Add rainbow color support
+        if (customColor == CommonColors.RAINBOW) {
```
kristofbolyai, line 57: If you call this after the render coord calculations, you can remove duplicate code

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -82,6 +87,67 @@ public void renderText(
...
+            case Center -> x - (font.width(text) / 2f);
+            case Right -> x - font.width(text);};
+
+        renderY = switch (verticalAlignment) {
+            case Top -> y;
+            case Middle -> y - (font.lineHeight / 2f);
+            case Bottom -> y - font.lineHeight;};
+
+        for (char c : text.toCharArray()) {
+            long dif = ((long) renderX * 10) - ((long) renderX * 10);
```
kristofbolyai: Why long?

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -82,6 +87,67 @@ public void renderText(
...
+
+            String s = String.valueOf(c);
```
kristofbolyai: This will break with chatformatting codes. You need to strip color codes beforehand, but you have to leave italic bold reset etc.

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -82,6 +87,67 @@ public void renderText(
...
+
+            // color settings
+            long time = System.currentTimeMillis() - dif;
+            float z = 2000.0F;
```
kristofbolyai, line 137: What is this and can this be a constant with a proper name?

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
kristofbolyai: Also what does dif stand for?

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -82,6 +87,67 @@ public void renderText(
...
+
+            int colorInt = Color.HSBtoRGB((float) (time % (int) z) / z, 0.8F, 0.8F);
+            CustomColor color = CustomColor.fromInt(colorInt);
+
+            // rendering shadows
+            float originPosX = x;
+            float originPosY = y;
+            switch (shadow) {
```
kristofbolyai, line 142: I am pretty sure you can call renderText (the original one) here so we don't need to duplicate this either

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
kristofbolyai: Reminder (this is undocumented lol): ChatFormatting color codes always override the color you pass to font renderer.

kristofbolyai (commented):
I mostly only have problems with code duplication and random variables. 

Also, you will have to handle ChatFormatting codes.

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
YouSoPugly: I think it needs to be duplicated to correct for the alignment

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
```diff
@@ -82,6 +91,101 @@ public void renderText(
...
+                }
+            }
+
+            // rendering the text
+            float charLength = font.width(c);
+            renderX += charLength + CHAR_SPACING;
+        }
+    }
+
+    public String getColorFromCode(char code) {
```
kristofbolyai, line 167: Well this is certainly messy. I am not sure how to support chatformatting colors without weird code.

kristofbolyai (commented):
Okay so I think I have an idea what you can do. Firstly clear color codes (using the method in ComponentUtils 
#stripColorFormatting iirc). Then you only have bold, italic, underline and reset to handle (might have left out smth). Handle that in in our code. That will make this significantly less messy.

common/src/main/java/com/wynntils/gui/render/FontRenderer.java
magicus: At the very least there's likely a factory method on `Character` that can be used instead.

magicus (changes requested):

magicus:
@YouSoPugly This PR has been untouched for a long time. I will close it now. If you decide to continue working on this functionality in the future, please open a new PR from your branch!

PR #650, opened 2022-11-05
feat: Auto message territory defense on attack

kristofbolyai:
image


Incompleteusern (changes requested):
We should ask wynncraft moderators if we can even automate message sending in this capacity

kristofbolyai:
> We should ask wynncraft moderators if we can even automate message sending in this capacity

avomod does this so

Incompleteusern:
> > We should ask wynncraft moderators if we can even automate message sending in this capacity
> 
> avomod does this so

I'd rather we still ask and be safe than sorry.

Incompleteusern (approved):
Still feel it could be a bit more obvious we are automating, but is fine

PR #656, opened 2022-11-06
feat: Parse defense in GuildAttackTimerOverlayFeature

kristofbolyai:
image


common/src/main/java/com/wynntils/core/objects/TimedSet.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.objects;
```
Incompleteusern: Move to utils/objects, I'd rather we not make a new dir

kristofbolyai: It already existed

common/src/main/java/com/wynntils/wynn/model/GuildAttackTimerModel.java
```diff
@@ -4,33 +4,84 @@
...
 
     private static List<TerritoryAttackTimer> attackTimers = List.of();
 
     public static void init() {}
 
+    @SubscribeEvent
+    public static void onMessage(ChatMessageReceivedEvent event) {
+        if (event.getRecipientType() != RecipientType.GUILD) return;
+
+        Matcher matcher = GUILD_DEFENSE_CHAT_PATTERN.matcher(event.getCodedMessage());
```
Incompleteusern, line 38: Can be faked, but we can't do much about that.

kristofbolyai: Yeah

common/src/main/java/com/wynntils/core/objects/TimedSet.java
Incompleteusern: ?

Incompleteusern: This is a new file?

kristofbolyai: But the package existed

kristofbolyai: Moved it though

Incompleteusern: Odd, I think its only on your client

Incompleteusern: > Moved it though

Sorry to ask again, but this is not wynn specific and would be better off in the com.wynntils.utils.objects package

kristofbolyai: np moved

Incompleteusern (approved):

PR #658, opened 2022-11-07
Soul Points Feature

PhoenixOrigin:
Adding /server soulpoint. It has an optional parameter (offset, -20 to 20). This is relatively simple and it has been attempted to be added before I am hoping athena works a bit better now. I also have to offset so if it is not correct once, you can try again but with the correct offset.


kristofbolyai (commented):
I am not sure why there are 2 duplicate classes or why the guild enum is commited here. However I do know that I would rather drop the "you need hero to switch" tooltip rather than ti parse the huge player info (and have caching issues with it like legacy). 

As for this feature, I am not convinced this will be accurate enough. If you clean up the mess in the PR, I'll give it a try

common/src/main/java/com/wynntils/core/webapi/profiles/ServerProfile.java
```diff
@@ -35,6 +35,14 @@ public String getUptime() {
                         - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(millis)));
     }
 
+    public int getUptimeMinutes() {
+        long millis = System.currentTimeMillis() - firstSeen;
+
+        return (int) Math.floor((TimeUnit.MILLISECONDS.toHours(millis) * 60)
```
kristofbolyai: This math is way overcomplicated

kristofbolyai: (and wrong i think)

PhoenixOrigin:
which duplicate classes

PhoenixOrigin:
for accuracy i added offset since then you can adjust manually

kristofbolyai:
Just remove player account and player class, no need to load it from API

common/src/main/java/com/wynntils/wynn/objects/account/PlayerClass.java
```diff
@@ -0,0 +1,521 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PlayerClass {
```
kristofbolyai: Why does this still exist

common/src/main/java/com/wynntils/core/webapi/ServerListModel.java
```diff
@@ -25,6 +25,10 @@ public static ServerProfile getServer(String worldId) {
         return availableServers.get(worldId);
     }
 
+    public static Map<String, ServerProfile> getAvailableServers() {
```
kristofbolyai, line 28: Remove this. This model does not update servers automatically, because it doesn't need to. Instead fetch new info using WebManager where this method is used (in the command)

PhoenixOrigin: I'll do it when I get home

magicus (changes requested):

PR #661, opened 2022-11-08
feat: Mob HP Formatting

P0keDev:
Abbreviates the mob health bar at the top of the screen to more "readable" values, if enabled.

image


common/src/main/java/com/wynntils/features/user/AbbreviateMobHealthFeature.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Pattern MOB_HEALTH_PATTERN = Pattern.compile("(.*§c)(\\d+)(§4❤.*)");
+
+    @Config
+    public static int precision = 1;
+
+    @SubscribeEvent
+    public void onHealthBarEvent(BossHealthUpdateEvent event) {
+        ClientboundBossEventPacket packet = event.getPacket();
+        Operation operation = ((ClientboundBossEventPacketAccessor) packet).getOperation();
```
magicus, line 27: My gut reaction is that this should be done by method calls on the BossHealthUpdateEvent, like having a `setName` method. But I realize that the BossHealthUpdatePacket is kind of special and that this really don't make sense. 

Maybe what we really should do is to have a "boss health manager" that help us make this kind of manipulations more easily. I'm not saying you need to fix that in this PR, just throwing out the idea.

common/src/main/java/com/wynntils/features/user/AbbreviateMobHealthFeature.java
```diff
@@ -0,0 +1,69 @@
...
+        M(1_000_000),
+        K(1_000);
+
+        final float value;
+
+        Abbreviation(float value) {
+            this.value = value;
+        }
+
+        static String formatNumber(int num) {
```
magicus: I'm pretty sure we already have methods like this..? For like emeralds and stuff. And if not, please move this to `StringUtils` or wherever you think you'd had looked for this kind of helper in the first place.

kristofbolyai (dismissed):
Looks good IMO.

common/src/main/java/com/wynntils/features/user/AbbreviateMobHealthFeature.java
P0keDev: yeah, the way boss packets work made it somewhat awkward to do this - i'm not sure how many other times it's useful to manipulate boss bar content though

common/src/main/java/com/wynntils/features/user/AbbreviateMobHealthFeature.java
P0keDev: oh you're totally right, i forgot about that. i will fix this tomorrow

common/src/main/java/com/wynntils/features/user/AbbreviateMobHealthFeature.java
magicus: I think we have a lot of boss bar interfacing already. And I promise you, if we make it easy to manipulate the boss bar, that we will find tons of new excuses to do it in many more cases. :-)

P0keDev: that's fair. i would rather leave that framework for a later PR however

kristofbolyai (approved):

PR #663, opened 2022-11-08
feat: Item Stat Decimal Setting

P0keDev:
Adds a config option for the number of decimals shown in item percentages. This also fixes percentage rounding so it matches legacy behavior.

Examples of increased/decreased decimal places:
image
image


common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
```diff
@@ -24,6 +24,9 @@ public class ItemStatInfoFeature extends UserFeature {
     @Config
     public boolean colorLerp = true;
 
+    @Config
+    public int decimalPlaces = 1;
```
HighCrit, line 28: Shouldn't this be zero, to preserve current functionality?

kristofbolyai (approved):

PR #664, opened 2022-11-08
feat: Player Armor Hiding

P0keDev:
Allows the user to disable armor rendering on a piece-by-piece basis (helmets, chestplates, etc.) and offers a separate option to preserve Wynn cosmetic rendering. This feature does not conflict with other armor-related features like the gear viewer or the nametag gear list. I decided to leave this feature disabled by default as it's a relatively drastic change to the vanilla Wynn experience and I'm unsure how popular it might be with the average player.

To implement this feature I needed an armor rendering event that fired once per armor piece instead of once per entity, which is how `PlayerArmorRenderEvent` functioned previously; I decided to repurpose this event instead of creating a separate event, but I can revert that change if necessary. Since certain head rendering is done separately by `CustomHeadLayer`, I had to mix into that rendering class as well, but I decided to use the same event for it as it fundamentally accomplishes the same thing.

While working on these render events, I noticed a bug in `PlayerGhostTransparencyFeature` which caused the value of the transparent armor setting to be ignored, so I fixed that as well.

In this example, boots are enabled and helmets are not, but the cosmetic override option is enabled. Also showcases that the nametag feature is unaffected.
image


kristofbolyai:
Can't we remove the setting from player ghost transparency feature? This is essentially the same thing isn't it? 

Also, how do you feel about merging this feature into pghost transparency? (just an idea, wanna hear your opinion)

kristofbolyai:
Sounds solid though

P0keDev:
removing the ghost armor setting would still limit player options if, say, they wanted to see certain pieces of armor on non-ghosts but still remove armor from ghosts. maybe it would be cleaner to move the ghost armor setting into this feature though?

kristofbolyai:
> removing the ghost armor setting would still limit player options if, say, they wanted to see certain pieces of armor on non-ghosts but still remove armor from ghosts. maybe it would be cleaner to move the ghost armor setting into this feature though?

Oh right because this works for non ghosts. Then this should be fine as is. I'll review throughly tomorrow, I am off to bed today

common/src/main/java/com/wynntils/features/user/players/PlayerArmorHidingFeature.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Config
+    public boolean showCosmetics = true;
+
+    @SubscribeEvent
+    public void onPlayerArmorRender(PlayerArmorRenderEvent event) {
+        switch (event.getSlot()) {
+            case HEAD -> {
+                if (!hideHelmets) return;
+                if (!showCosmetics) { // helmet is hidden regardless, no extra logic needed
```
kristofbolyai, line 40: Can you return early here too? Just a nit though

kristofbolyai (approved):
Looks good :)

common/src/main/java/com/wynntils/features/user/players/PlayerArmorHidingFeature.java
P0keDev: where, exactly? this `if` statement does return, it just needs to cancel the event first

kristofbolyai: if showcosmetics return;
e.cancel

P0keDev: not every helmet is a cosmetic though - that logic wouldn't match the functionality

kristofbolyai: oh, leave it then

PR #666, opened 2022-11-09
fix: Render player shadows correctly when injecting nametags

kristofbolyai:


common/src/main/java/com/wynntils/mc/mixin/PlayerRendererMixin.java
```diff
@@ -52,4 +54,23 @@ public void redirectNameTagRender(
             matrixStack.translate(0.0, 0.25875f, 0.0);
```
P0keDev, line 54: instead of doing all the work of requiring a second mixin, couldn't you just "undo" these translations to the matrix after the loop ends?

kristofbolyai: No, we need to render the actual name nametag

magicus: The assymmetry feels a bit strange. Maybe it's better to make a second `@At INVOKE` injection to push the matrices before the call?

kristofbolyai: Eh, I don't think it would "help"

magicus: More code, yes. I just don't feel convinced that the push and pop is always going to be paired. I assume you have checked this. But that is an assumption that can potentially break at every minecraft update.

magicus: But skip my suggestion if you don't want to do it. It was just a suggestion, after all.

kristofbolyai: If we upgrade versions we will have to deal with breakage anywYs

magicus (approved):

P0keDev (approved):

PR #668, opened 2022-11-10
fix: update spell cost ID ranges

P0keDev:
Fixes our ID logic to be in line with the new way spell cost IDs are handled by Wynn. `IdentificationProfile` didn't know whether its own ID was inverted (or even what its name was), so I had to pass that info into it via `calculateMinMax` - it's not the cleanest solution but I haven't thought of a better one.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/utils/WynnItemUtils.java
+++ b/common/src/main/java/com/wynntils/wynn/utils/WynnItemUtils.java
@@ -91,8 +91,9 @@ public final class WynnItemUtils {
      */
     public static ItemIdentificationContainer identificationFromValue(
             Component lore, ItemProfile item, String idName, String shortIdName, int value, int starCount) {
-        boolean isInverted = IdentificationOrderer.INSTANCE.isInverted(shortIdName);
         IdentificationProfile idProfile = item.getStatuses().get(shortIdName);
+        boolean isInverted =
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/wynn/utils/WynnItemUtils.java
```diff
@@ -91,8 +91,9 @@ public static ItemIdentificationContainer identificationFromLore(Component lore,
      */
     public static ItemIdentificationContainer identificationFromValue(
             Component lore, ItemProfile item, String idName, String shortIdName, int value, int starCount) {
-        boolean isInverted = IdentificationOrderer.INSTANCE.isInverted(shortIdName);
         IdentificationProfile idProfile = item.getStatuses().get(shortIdName);
+        boolean isInverted =
+                idProfile != null ? idProfile.isInverted() : IdentificationOrderer.INSTANCE.isInverted(shortIdName);
```
Incompleteusern, line 96: What is the difference between idProfile and the other?

P0keDev: the second case is intended to deal with items that aren't in the API for whatever reason and thus have no profile attached to them. it's not really necessary to do it like this, but it mirrors the check for the ID type in the adjacent line

Incompleteusern (approved):

kristofbolyai (approved):
Looks great, thanks!

PR #671, opened 2022-11-11
feat: Add Chat Tabs

kristofbolyai:
image
imageimage
image
image

Some RecipientType regexes needed adjusting. The system is not perfect yet. For example, party tab only shows actual party messages, not party creation/invitation messages. I assume guild tab does not show guild info messages either, although I have not tested that yet. Overall, I think this is better than nothing and we can start from here.


magicus:
Oh, nice! This was a big piece of what was missing from Legacy.

I'm thinking we should slightly redesign this compared to Legacy. As I recall, the regexp:s for matching e.g. private messages etc was left open to the user. While theoretically more powerful, it also made it much harder to filter. Now that Artemis tags message by type inherently, I think user-specified tabs need a kind of filter with a combination of recipient and regexp (so you e.g. could filter out personal communication from a single friend).

magicus:
Also, I think it is important to design this so messages can belong to several tabs. That is, a tab is not so much an "owner" of a message, but a filter on the total stream of messages. 

So if you filter on e.g. a friend's name, but do not specify recipient type (or clicks in all recipient type checkboxes), then you'll get a tab where you see everything this person sent, regardless of if it's a party message, guild message, shout, etc.

kristofbolyai:
> Also, I think it is important to design this so messages can belong to several tabs. That is, a tab is not so much an "owner" of a message, but a filter on the total stream of messages. 
> 
> 
> 
> So if you filter on e.g. a friend's name, but do not specify recipient type (or clicks in all recipient type checkboxes), then you'll get a tab where you see everything this person sent, regardless of if it's a party message, guild message, shout, etc.

Low priority exists for this reason. Low priority tabs don't "consume" messages on match.

magicus:
And the default tab should be an "All" tab, that basically looks like the chat screen is without chat tabs, with all messages appearing. That way we minimize difference to Vanilla by default.

magicus:
> Low priority exists for this reason. Low priority tabs don't "consume" messages on match.

I'm not sure this is a good concept. Let's just call this the "normal" handling, for now, and call "high priority" tabs "stealing" tabs. That means that when a stealing tab makes a match, it will not show up in any other tab. This introduce a bunch of other problems.

1) you need to define the order of the stealing tabs. Maybe this is the ordering of the tabs on-screen, but that might also lead to surprising behavior, if you drag the tabs around and they suddenly start changing content.

2) it makes things like an "All" tab impossible to implement

magicus:
I think what you possibly want is an "exclusive" or "except" part of filtering. So let's say you have a tab "Primary" that you wanted to make "stealing", containing e.g. all private, guild and party messages. And then you have a a tab "Rest" that you want to contain all other messages, that is not consumed by "Primary".

But instead of making "Primary" stealing, you can set the filter for "Rest" to be like:
```
RecipientType: *
RegExp: .*
Except-When-Matched-By: Primary
```

kristofbolyai:
> > Low priority exists for this reason. Low priority tabs don't "consume" messages on match.
> 
> 
> 
> I'm not sure this is a good concept. Let's just call this the "normal" handling, for now, and call "high priority" tabs "stealing" tabs. That means that when a stealing tab makes a match, it will not show up in any other tab. This introduce a bunch of other problems.
> 
> 
> 
> 1) you need to define the order of the stealing tabs. Maybe this is the ordering of the tabs on-screen, but that might also lead to surprising behavior, if you drag the tabs around and they suddenly start changing content.
> 
> 
> 
> 2) it makes things like an "All" tab impossible to implement

I can add priority, but I don't see why inverting low/high priority bool would cause any issues

magicus:
Also implicit in my thinking is that we have a single linear source of chat messages, which is basically monotonically increasing (except when pruning too old messages). That also brings several benefits:
* Changing the filters work retrospectively, so if you want to blacklist shouts from MrAnnoying, it will update the chat history as well.
* It will allow us to retrospectively update chat messages with new color information, which will (finally) bring the ChatManager functionality in parity with vanilla (where chat colors are hidden when an NPC dialogue is in progress, only to be colored in afterwards).

kristofbolyai:
> Also implicit in my thinking is that we have a single linear source of chat messages, which is basically monotonically increasing (except when pruning too old messages). That also brings several benefits:
> 
> * Changing the filters work retrospectively, so if you want to blacklist shouts from MrAnnoying, it will update the chat history as well.
> 
> * It will allow us to retrospectively update chat messages with new color information, which will (finally) bring the ChatManager functionality in parity with vanilla (where chat colors are hidden when an NPC dialogue is in progress, only to be colored in afterwards).

At the moment, I use different ChatComponents (MC's name for chat history gui) and render the correct one. While the tabs don't have a single source, on update, we can loop through the tab's "recent" messages, and update it.

magicus:
> I can add priority, but I don't see why inverting low/high priority bool would cause any issues

It's more of a design issue. I think we lock ourselves in on a bad design if we think that tabs "own" messages, rather than viewing tabs as "filter" or "viewpoints" into a single, coherent stream of messages.

Basically, think of it like a `stream()` interface of messages, where each tab can do a `filter()` to select which messages to display.

magicus:
Anyway, that's what I think is the best way to implement this. But I know I've been spending less and less time with Artemis, and I believe in the OSS credo of "he who writes the code get's the say", so if you want to do it differently, do it so.

kristofbolyai:
Well, I see you have a pretty good idea in your head about this. That's partly why I started working on this so "late". However, we both know that you likely won't be able to work on this other than comments, but even with them, I might not able to bring chat tabs into a state that you could. 

So my plan is to implement a legacy like functionality, with "simpler" code. And whenever you are free again, you can rewrite/refactor it. I think that would fit the users best.

kristofbolyai:
Just the be clear, I have no intention in "ignoring" your comments. The design sketch sounds really good, and will likely work really good, like every other model of yours. The only thing I am saying is that I think you are the only one who can bring it to that level of perfection.

magicus:
Sounds like a plan.

kristofbolyai:
@magicus first PR comment might be interesting for you. (https://github.com/Wynntils/Artemis/pull/671#issue-1445949060)

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent event) {
+        if (event.getScreen() instanceof ChatScreen chatScreen) {
+            int xOffset = 0;
+
+            for (ChatTab chatTab : chatTabs) {
+                chatScreen.addRenderableWidget(new ChatTabButton(xOffset + 2, chatScreen.height - 35, 40, 13, chatTab));
+                xOffset += 43;
```
magicus, line 110: Should the tab offset depend on the length of the name of the tab?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class ChatTabModel extends Model {
+    private static final ChatComponent EMPTY_CHAT_COMPONENT = new ChatComponent(McUtils.mc());
+
+    private static ChatTab focusedTab = null;
+
+    private static Map<ChatTab, ChatComponent> chatTabData = Map.of();
+
+    public static void init() {
+        chatTabData = new ConcurrentHashMap<>();
```
magicus: Why not assign this from the start? Is `init` called multiple times if this is enabled/disabled?

kristofbolyai: Yes. And we want to re-assign it so in-case we disable the model, we don't hold onto useless chat data in memory.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
kristofbolyai: Long names just overflow the button, as it is in legacy. It is rather messy to impl a gui button with a dynamic size. So for now, this should work.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
```diff
@@ -0,0 +1,52 @@
...
+        this.filteredTypes = filteredTypes;
+        this.customRegexString = customRegexString;
+
+        if (customRegexString != null) {
+            this.customRegex = Pattern.compile(customRegexString);
+        }
+    }
+
+    public boolean matchMessageFromEvent(ChatMessageReceivedEvent event) {
+        if (customRegexString != null) {
```
magicus: I think this logic could be improved. Your comment above states ` // Filters, inactive one is null` but that is not really how this is implemented. I think the filters should be ANDed, and that `null` filters should count as `true` (or a match). 

Right now, if you have both a regexp and a recipient filter, and the regexp match, you will not check the recipient type, but return true regardless. And if you pass in `null` to both filters, you will get `false`. 

I think the first example should return true iff both the regexp and the recipient filter match. And I think the latter example should return `true` (and thus making the "All" ChatTab just use two null filters).

kristofbolyai: Fixed

magicus (dismissed):

kristofbolyai:
@magicus Do you have a fix in mind for matching guild/party info messages in ChatModel? Maybe a new RecipientType called GuildInfo and PartyInfo?

IMO this is not a blocking thing for this feature, but nice to sort out eventually.

kristofbolyai:
@magicus I've noticed a problem after your review. We dismissed any "local" messages (client to player chat). So I needed to add code to fix that. However, I changed so much, I will most definitely need a review again whenever you are free.

magicus:
> @magicus Do you have a fix in mind for matching guild/party info messages in ChatModel? Maybe a new RecipientType called GuildInfo and PartyInfo?

I don't have a fix in mind. I tried to let "guild" cover all kinds of guild messages, but I have probably not seen messages for someone leaving or entering the guild. (And I did not think of testing that for Party).

So if anything, I think we should consider those messages part of guild and party. And only if that seem to make these categories too messy try to split it up afterwards instead. For guild, there are more potential "subcategories" as well; but I'm not sure that's relevant to separate.

kristofbolyai:
> > @magicus Do you have a fix in mind for matching guild/party info messages in ChatModel? Maybe a new RecipientType called GuildInfo and PartyInfo?
> 
> 
> 
> I don't have a fix in mind. I tried to let "guild" cover all kinds of guild messages, but I have probably not seen messages for someone leaving or entering the guild. (And I did not think of testing that for Party).
> 
> 
> 
> So if anything, I think we should consider those messages part of guild and party. And only if that seem to make these categories too messy try to split it up afterwards instead. For guild, there are more potential "subcategories" as well; but I'm not sure that's relevant to separate.

We should add a list of regexes for recipient types then

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
```diff
@@ -0,0 +1,60 @@
...
+            return false;
+        }
+
+        return customRegexString == null
+                || getCustomRegex().matcher(event.getCodedMessage()).matches();
+    }
+
+    public boolean matchMessageFromEvent(LocalMessageEvent event) {
+        if (customRegexString == null) {
+            return false;
```
magicus, line 45: This should be true..?

kristofbolyai: If we don't have a custom regex, we are not matching this. Local messages are command responses (and every mod to player chat message) usually, so they are not matched to RecipientTypes. So if we don't have a regex for a chat tab, we are not getting any command responses there.

magicus: I think a more "flow oriented" filtering would be more readable. Like:
 * "if we have a recipient type filter, and it does not match, return false"
 * "if we have a regexp filter, and it does not match, return false"
 * "otherwise, return true"

And for local messages, which do not have a recipient, will always match true for the first case.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
```diff
@@ -0,0 +1,60 @@
...
+    private Set<RecipientType> filteredTypes;
+    private String customRegexString;
+
+    private transient Pattern customRegex;
+
+    public ChatTab(String name, boolean lowPriority, Set<RecipientType> filteredTypes, String customRegexString) {
+        this.name = name;
+        this.lowPriority = lowPriority;
+        this.filteredTypes = filteredTypes;
+        this.customRegexString = customRegexString;
```
magicus, line 27: I think it's better to make the field a Pattern and compile it in the constructor, or assign null if no regexp string is given. The `getCustomRegex` looks too much like the failed double-locking pattern.

kristofbolyai: No, you can't do that. Gson won't use the constructor properly. This is more simple than writing a whole deserializer though

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public boolean matchMessageFromEvent(ChatMessageReceivedEvent event) {
+        if (filteredTypes != null && !filteredTypes.isEmpty() && !filteredTypes.contains(event.getRecipientType())) {
```
magicus, line 31: Actually, I think there is a conceptual difference between `filteredTypes` being `null` as "don't have this filter", and `filteredTypes` being the empty set. While admittedly a bit stupid, the latter seems to say "match no recipients ever". It's like sending in "^$" to the regexp, it would be a regexp but it would not match any real line.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
magicus: But you do ` chatTabData.clear()` in `disable`, is that not enough?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: Why would one create a chat tab like that? and more importantly, should we allow users to fall into that rabbithole?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static void disable() {
+        chatTabData.clear();
+    }
+
+    @SubscribeEvent
+    public static void onWorldStateChange(WorldStateEvent event) {
+        if (event.getNewState() == WorldStateManager.State.NOT_CONNECTED) {
```
magicus, line 34: I just realized: how well does this code play with logging in to different servers and local worlds? So if you log in to e.g. server A, then Wynncraft, then server B, does everything look fine and dandy, for all three servers?

kristofbolyai: Yes, everything should work fine. Chat tabs are only injected on Wynn.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -0,0 +1,133 @@
...
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class ChatTabsFeature extends UserFeature {
+    public static ChatTabsFeature INSTANCE;
+
+    @Config(visible = false)
+    public List<ChatTab> chatTabs = Arrays.asList(
+            new ChatTab("All", true, null, ".*"),
```
magicus: I think the regexp here should be `null` as well. `All` is kind of conceptually the tab with no filters whatsoever.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
magicus: Ah, are you using gson for the chat tabs? I must have missed that. Why's that? To save history?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: Should be, changing it.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: Configs are saved with gson? Chat Tabs are saved to config in ChatTabFeature.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
magicus: I see your point. Fine, keep it this way. But then maybe you should set the filteredTypes to the empty set in the constructur... except of course you can't do that because of json.... *sigh*

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
magicus: Hm.

Maybe we should treat these messages as a special recipient category, COMMANDS?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
magicus: Just to be clear: have you tested, or are you just asserting that "it should work"? :)

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: > I think a more "flow oriented" filtering would be more readable. Like:
> 
> * "if we have a recipient type filter, and it does not match, return false"
> * "if we have a regexp filter, and it does not match, return false"
> * "otherwise, return true"
> 
> And for local messages, which do not have a recipient, will always match true for the first case.

This doesnt capture the case correctly when we have a regex filter but do not match it.

kristofbolyai: Oh, maybe it does. Im tired

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Ah, right, it's saved by json for the configuration.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: It won't work for local messages, because it wont handle no regex case.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
kristofbolyai: > I think the regexp here should be `null` as well. `All` is kind of conceptually the tab with no filters whatsoever.

That would mean I don't have a regex so I don't match anything there? Unless I change null regex to match everything which would mean I can't filter commands from tabs.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: > Just to be clear: have you tested, or are you just asserting that "it should work"? :)

Tested on Hypixel

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
magicus: I'm a bit tired as well, so maybe I'm not thinking straight either. :) But my understanding is that we, confusingly, have two different meanings of "local" right now:

1) chat messages sent from Wynncraft, written by players nearby

2) chat messages sent client-side only, from e.g. mods and commands

This is confusing in itself. I suggest we rename one of them. Either the Wynncraft concept to "NEARBY", or the client-side one to "COMMAND" or "CLIENTSIDE" or something like that. Or, possibly, rename both and have no "local" if it can be misunderstood.

kristofbolyai: Renamed to ClientsideMessageEvent.

magicus: And then, for 2 (client-side), I think things should work just the same. If we get e.g. an `Unknown command` back as a clientside reply, and it is tagged `MessageRecipientType.CLIENTSIDE`, then it would go to a "Clientside" tab for that recipient type. And if you set up a custom filter with both CLIENTSIDE and a regexp, it would only match messages that fullfil both requirements. So I think it would work just like the rest of the messages.

(In fact, it should probably even be the responsibility of the ChatManager to catch these messages, and tag them properly)

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Yes, I mean that the null regexp should match everything. (And I think I presented a solution for the command replies)

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: Okay. I will refactor this tomorrow. Better for both of us :)

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Ok. Legacy compatiblity is a perfectly fine excuse for just about anything! :-) (I'm kind of seeing the xkcd "Compiling!" "Oh, okay, carry on then" strip in front of me...)

common/src/main/java/com/wynntils/features/user/UpdatesFeature.java
```diff
@@ -38,50 +39,51 @@ public void onWorldStateChange(WorldStateEvent event) {
...
-                }
-
-                if (updateReminder) {
-                    remindToUpdateIfExists(version);
-                }
-
-                if (autoUpdate) {
-                    if (WynntilsMod.isDevelopmentEnvironment()) {
-                        WynntilsMod.info("Tried to auto-update, but we are in development environment.");
+        CompletableFuture.runAsync(() -> UpdateManager.getLatestBuild()
```
magicus, line 42: I'm confused. Is this a merge error?

common/src/main/java/com/wynntils/mc/mixin/LocalPlayerMixin.java
```diff
@@ -38,4 +42,13 @@ private void onDropPre(boolean fullStack, CallbackInfoReturnable<Boolean> cir) {
             cir.cancel();
         }
     }
+
+    @Inject(method = "sendMessage", at = @At("HEAD"), cancellable = true)
+    private void onSendMessage(Component component, UUID senderUUID, CallbackInfo ci) {
+        if ((Object) this != McUtils.player()) return;
+
+        if (EventFactory.onLocalMessage(component).isCanceled()) {
```
magicus: Rename this Clientside as well.

common/src/main/java/com/wynntils/features/user/UpdatesFeature.java
kristofbolyai: No. Some really weird thing happened to MC because we sent a message to a player chat from ForkJoinPool. Now, we should send it from the correct, game thread

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
kristofbolyai: I think I did this. I did not mix client sided messages with chatrecievedevent, I think that is fine. LMK if you want anything else changed.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -0,0 +1,133 @@
...
+    // We do this here, and not in ChatTabModel to not introduce a feature-model dependency.
+    @SubscribeEvent(priority = EventPriority.LOWEST)
+    public void onChatReceived(ChatMessageReceivedEvent event) {
+        // Firstly, find the FIRST matching tab with high priority
+        for (ChatTab chatTab : chatTabs) {
+            if (chatTab.isLowPriority()) continue;
+
+            if (chatTab.matchMessageFromEvent(event)) {
+                ChatTabModel.addMessageToTab(chatTab, event.getMessage());
+                return;
```
P0keDev, line 61: does this mean it's not possible for two tabs that aren't low priority to receive the same message? this seems like a step backwards from legacy functionality

kristofbolyai: Doesn't legacy non low priority consume the message? as in the first matching tab gets to show the message, any other not

P0keDev: in legacy that logic only applies to tabs specifically marked as low priority - i had two tabs that were set to receive all messages (one for typing into public chat, one for typing to guild chat), and that structure couldn't work here unless i made all my tabs low priority, which doesn't feel very intuitive 

honestly it might be better to abandon the priority system altogether and implement a different hierarchy system along with "negative" filters, but that could be dealt with in a later PR

kristofbolyai: We could just make consumer tabs. If checked, the specific overlay consumes it's matching messages. How do you feel about that?

P0keDev: so there would be consumer/"high priority" tabs, regular tabs, and then low priority? that sounds like a good solution

kristofbolyai: Exactly

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
magicus: I think it looks good now.

magicus (dismissed):

PR #675, opened 2022-11-12
fix: GuildTerritoryModel should account for partial updates

kristofbolyai:


Incompleteusern (approved):

common/src/main/java/com/wynntils/wynn/model/territory/GuildTerritoryModel.java
```diff
@@ -70,10 +72,8 @@ public static void onAdvancementUpdate(AdvancementUpdateEvent event) {
 
             if (territoryProfile == null) continue;
 
-            newMap.put(entry.getKey(), new TerritoryPoi(territoryProfile, entry.getValue()));
```
magicus, line 73: Is there anyway we can remove TerritoryPoi now? Or maybe that is not relevant?

kristofbolyai: No we can't

kristofbolyai: It is a class used for rendering on map

PR #679, opened 2022-11-12
fix: Correct chat item encoding of spell cost IDs

P0keDev:
Updates chat item logic for our new handling of spell cost IDs, which was previously encoding "negative" values and breaking the decoding process

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/ChatItemModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/ChatItemModel.java
@@ -95,7 +95,8 @@ public final class ChatItemModel {
             if (Math.abs(idProfile.getBaseValue()) > 100) { // calculate percent
                 translatedValue = (int) Math.round((idValue * 100.0 / idProfile.getBaseValue()) - 30);
             } else { // raw value
-                translatedValue = idValue - idProfile.getMin();
+                // min/max must be flipped for inverted IDs to avoid negative values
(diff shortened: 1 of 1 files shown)


kristofbolyai:
How backwards compatible is this?

P0keDev:
In chat items to and from artemis and legacy, spell cost IDs will be shown as the incorrect value (and usually out of the actual range of the ID). This can't really be avoided until legacy gets updated for the spell cost changes too, but I will try to finish that today

kristofbolyai:
> In chat items to and from artemis and legacy, spell cost IDs will be shown as the incorrect value (and usually out of the actual range of the ID). This can't really be avoided until legacy gets updated for the spell cost changes too, but I will try to finish that today 

So old/new artemis is incompatible too? I know we can't avoid this, I am just curious

kristofbolyai (approved):

P0keDev:
> > In chat items to and from artemis and legacy, spell cost IDs will be shown as the incorrect value (and usually out of the actual range of the ID). This can't really be avoided until legacy gets updated for the spell cost changes too, but I will try to finish that today
> 
> So old/new artemis is incompatible too? I know we can't avoid this, I am just curious

If you mean before/after this specific commit, the chat item feature straight up wouldn't work at all on items with spell cost IDs until this fix. Artemis versions from before the spell cost ID changes at all would be the same as legacy

PR #684, opened 2022-11-14
fix: Misc. spell cost ID fixes

P0keDev:


common/src/main/java/com/wynntils/core/webapi/profiles/item/IdentificationProfile.java
```diff
@@ -32,14 +32,14 @@ public IdentificationProfile(IdentificationModifier type, int baseValue, boolean
     }
 
     public void calculateMinMax(String shortId) {
+        isInverted = IdentificationOrderer.INSTANCE.isInverted(shortId);
```
magicus, line 35: Why did you move this?

common/src/main/java/com/wynntils/wynn/objects/SpellType.java
```diff
@@ -14,7 +14,7 @@ public enum SpellType {
     ARROW_SHIELD(ClassType.Archer, 4, "Arrow Shield", 8, 1),
 
     SPIN_ATTACK(ClassType.Assassin, 1, "Spin Attack", 6, 0),
-    VANISH(ClassType.Assassin, 2, "Vanish", 2, 0),
+    VANISH(ClassType.Assassin, 2, "Dash", 2, 0),
```
magicus: Has the spell changed name in 2.0? Maybe we should update the name of the enum as well?

kristofbolyai: It did, and I agree

common/src/main/java/com/wynntils/core/webapi/profiles/item/IdentificationProfile.java
kristofbolyai: the value was used earlier than calculated

magicus: Is it, really? Am I blind? It has moved from line 41 to line 35. I see no mention of `isInverted` between those lines. It is refered to in line 43, but the old position was before that as well. There are no method calls between line 35 and 41. I can't see how this change affects anything.

kristofbolyai: Line 40 returns early, so isInverted needs to be set before

magicus: Ah, now I get it. I was just really thick. Thanks!

common/src/main/java/com/wynntils/wynn/objects/SpellType.java
P0keDev: woops, good point

kristofbolyai (approved):

PR #691, opened 2022-11-15
Adds config for max distance that the waypoint's text will render

YouSoPugly:


magicus (dismissed):
This looks good, but I'm wondering if it should be better if there were a fade-out as you get closer to the limit, slowly raising alpha for like the last 10% of the distance?

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -84,7 +88,10 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
 
     @SubscribeEvent
     public void onRenderGuiPost(RenderEvent.Post event) {
-        if (CompassModel.getCompassLocation().isEmpty() || normalizedDeviceCoordinates == null || isOnScreen()) return;
+        if (CompassModel.getCompassLocation().isEmpty()
+                || normalizedDeviceCoordinates == null
+                || isOnScreen()
+                || maxWaypointTextDistance < distance) return;
```
kristofbolyai: Would like it so if 0 would mean display every time

YouSoPugly: Added this in the most recent commit

YouSoPugly:
> 

That could be kind of awkward though because the idea of the limit is that when you're something like 10k blocks away on a quest it doesn't render and cover the screen.

kristofbolyai (dismissed):
Thanks!

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -590,6 +590,8 @@
   "feature.wynntils.userFeature.userEnabled.name": "Enabled",
   "feature.wynntils.worldWaypointDistance.backgroundOpacity.description": "What level of opacity should text background have?",
   "feature.wynntils.worldWaypointDistance.backgroundOpacity.name": "Background Opacity",
+  "feature.wynntils.worldWaypointDistance.maxWaypointTextDistance.description": "The distance at which the waypoint text should stop rendering",
```
```suggestion
  "feature.wynntils.worldWaypointDistance.maxWaypointTextDistance.description": "At what distance should the waypoint text stop rendering?",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -590,6 +590,8 @@
...
+  "feature.wynntils.worldWaypointDistance.maxWaypointTextDistance.name": "Maximum Distance of the Waypoint Text",
```
```suggestion
  "feature.wynntils.worldWaypointDistance.maxWaypointTextDistance.name": "Maximum Distance for Waypoint Text",
```

PR #692, opened 2022-11-16
fix: Fix chat items not parsing + feature crash when screenshot copy fails

kristofbolyai:


common/src/main/java/com/wynntils/features/user/ItemScreenshotFeature.java
```diff
@@ -104,7 +106,11 @@ private static void takeScreenshot(Screen screen, Slot hoveredSlot) {
         McUtils.mc().getMainRenderTarget().bindWrite(true);
 
         BufferedImage bi = RenderUtils.createScreenshot(fb);
-        RenderUtils.copyImageToClipboard(bi);
+        try {
+            RenderUtils.copyImageToClipboard(bi);
+        } catch (HeadlessException ex) {
+            WynntilsMod.error("Failed to copy image to clipboard", ex);
```
magicus, line 112: Should we send an error message to the chat as well?

kristofbolyai: We could. I'll do that later.

kristofbolyai: (this fix also introduces a model dependency, i can fix it easily later)

PR #697, opened 2022-11-17
feat: Add chat timestamps

Zatzou:
This pr adds chat timestamps simlar to 1.12 wynntils


common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
```diff
@@ -0,0 +1,43 @@
...
+import com.wynntils.core.features.properties.StartDisabled;
+import com.wynntils.wynn.event.ChatMessageReceivedEvent;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@StartDisabled
```
kristofbolyai: Why? This is useful to be enabled by default IMO

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
```diff
@@ -0,0 +1,43 @@
...
+public class ChatTimestampFeature extends UserFeature {
+    @Config
+    public String formatPattern = "HH:mm:ss";
+
+    @SubscribeEvent
+    public void onChat(ChatMessageReceivedEvent event) {
```
kristofbolyai: ChatMessageReceivedEvent is from ChatModel. Make sure to add ChatModel to this feature's model dependencies.

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
```diff
@@ -0,0 +1,43 @@
...
+        Component message = event.getMessage();
+
+        // use the user specified formatter and if that formatter fails for whatever reason handle it as gracefully as
+        // we can
+        MutableComponent timestamp;
+        try {
+            LocalDateTime date = LocalDateTime.now();
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatPattern);
```
kristofbolyai: Can this be cached and updated on config change (`onConfigUpdate` method)?

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
```diff
@@ -0,0 +1,43 @@
...
+
+            timestamp = new TextComponent("§8[§7" + date.format(formatter) + "§8]§r ");
+        } catch (Exception e) {
+            timestamp = new TextComponent("§8[§7<Invalid time format string>§8]§r ");
```
kristofbolyai: I don't like how we handle this. IMO, check correctness on config update, print the incorrect message once, and have a boolean to return early onChat, if the format is incorrect (or set the cached DateTimeFormatter to null and check for that)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -30,6 +30,9 @@
   "feature.wynntils.beaconBeam.waypointBeamColor.name": "Beacon Beam Color",
   "feature.wynntils.chatItem.name": "Chat Item",
   "feature.wynntils.chatTabs.name": "Chat Tabs",
+  "feature.wynntils.chatTimestamp.formatPattern.description": "String used for formatting the timestamps",
+  "feature.wynntils.chatTimestamp.formatPattern.name": "Time formatting string",
```
```suggestion
  "feature.wynntils.chatTimestamp.formatPattern.name": "Timestamp Format",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -30,6 +30,9 @@
   "feature.wynntils.beaconBeam.waypointBeamColor.name": "Beacon Beam Color",
   "feature.wynntils.chatItem.name": "Chat Item",
   "feature.wynntils.chatTabs.name": "Chat Tabs",
+  "feature.wynntils.chatTimestamp.formatPattern.description": "String used for formatting the timestamps",
```
```suggestion
  "feature.wynntils.chatTimestamp.formatPattern.description": "How should the timestamp look like in chat?",
```

kristofbolyai (changes requested):
I require a few changes to improve the "mechanics" of this feature. Otherwise this looks great for a first PR :)

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
Zatzou: this feature is disabled by default on 1.12 wynntils afaik so that is why I disabled it by default

kristofbolyai: Oh. Well, I'll leave it to you I guess.

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
Zatzou: would it be better to have the feature soft disable on an invalid formatstring or would it be better to have it revert to the default value with a message telling the user that the value has been reverted?

kristofbolyai: Soft disable cause reverting would be buggy with the config system. Not the greatest solution but oh well.

Zatzou:
I've made the requested changes.

common/src/main/java/com/wynntils/features/user/ChatTimestampFeature.java
```diff
@@ -0,0 +1,63 @@
...
+            formatter = fmt;
+        } catch (Exception e) {
+            formatter = null;
+
+            McUtils.sendMessageToClient(new TranslatableComponent("feature.wynntils.chatTimestamp.invalidFormatMsg")
+                    .withStyle(ChatFormatting.DARK_RED));
+        }
+    }
+
+    private DateTimeFormatter formatter = DateTimeFormatter.ofPattern(formatPattern);
```
kristofbolyai: Could you move this to either below or after the config? Fields are generally at the top of the class.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -30,6 +30,10 @@
   "feature.wynntils.beaconBeam.waypointBeamColor.name": "Beacon Beam Color",
   "feature.wynntils.chatItem.name": "Chat Item",
   "feature.wynntils.chatTabs.name": "Chat Tabs",
+  "feature.wynntils.chatTimestamp.formatPattern.description": "How should the timestamp look like in chat?",
+  "feature.wynntils.chatTimestamp.formatPattern.name": "Timestamp Format",
+  "feature.wynntils.chatTimestamp.invalidFormatMsg": "Invalid timestamp format specified in config! Timestamps will be disabled until a valid format is set",
```
```suggestion
  "feature.wynntils.chatTimestamp.invalidFormatMsg": "Invalid timestamp format specified in config! Timestamps will be disabled until a valid format is set.",
```

kristofbolyai (commented):
It's really close now :)

Zatzou:
I've fixed the ordering of the class

kristofbolyai:
@Zatzou I will fix up the problem with priorities, https://github.com/Wynntils/Artemis/pull/697/commits/36fcfcdec2ac6c0d43c527199182f25c82f2ad05 won't be good enough in the long run.

Zatzou:
Yeah I actually noticed after I made that commit that the timestamps don't appear at all with LOWEST priority LOW seems to work but I think its more a case where the filters should have a higher priority.

Zatzou:
After testing it LOW priority works perfectly

kristofbolyai:
> After testing it LOW priority works perfectly

No, it doesnt. If you create a tab with a custom regex, the event would already have a modified component.

kristofbolyai:
I have a fix in mind, I am just not free for a few hours.

kristofbolyai:
image

kristofbolyai:
@Zatzou I can't seem to push to your fork, do you know why?

kristofbolyai:
@Zatzou nevermind, I pushed my changes. Sorry for taking over, but it was easier for me to do than explain it. Hope you understand :). I will leave you to check out the final version, and let me know if you don't agree with anything.

kristofbolyai (approved):

magicus (approved):

magicus:
Looks good to me. Thank you for your contribution!

PR #698, opened 2022-11-17
fix: Fix Figura crash and black config screen

kristofbolyai:


Incompleteusern (approved):

magicus:
I think the new code looks much less intrusive, thank you! However, all the rendering work that is done in PlayerRendererMixin do no longer need to be done there. Instead, it can be done elsewhere by listening to the event, doing the rendering, and then cancel the event. (Basically, the same way as we do all other of our mixins.)

If such a listener have no other reasonable place to be put, we might need to have a new `NameTagManager` that handles additional name tags, and if any such are registered, takes over the rendering and cancels the events.

PR #705, opened 2022-11-19
fix: Fix chat tabs resetting on config reload

kristofbolyai:


common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -60,13 +61,18 @@ public static void addMessageToTab(ChatTab tab, Component message) {
     }
 
     public static void setFocusedTab(ChatTab focused) {
+        if (Objects.equals(focusedTab, focused)) {
```
kristofbolyai, line 64: Again, I didn't provide info about this PR. With the implemented equals method and this check, we do not overwrite the current focused if it is the same as the new one.

Incompleteusern (approved):

common/src/main/java/com/wynntils/core/chat/tabs/ChatTab.java
```diff
@@ -65,4 +66,21 @@ public Pattern getCustomRegex() {
...
+        return consuming == chatTab.consuming
+                && Objects.equals(name, chatTab.name)
+                && Objects.equals(filteredTypes, chatTab.filteredTypes)
+                && Objects.equals(customRegexString, chatTab.customRegexString);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, consuming, filteredTypes, customRegexString);
+    }
```
magicus, line 85: ... ChatTabs really should have been a record. Too bad it does not play nicely with gson. :-(

PR #706, opened 2022-11-19
fix: Do not scan quest book on entering world

kristofbolyai:


kristofbolyai:
Forgot to add the reasoning. Basically, we do not depend on the quest book analysis anywhere expect the quest book screens itself, where we are free to do analysis. Scanning on world join is useless therefore.

Incompleteusern (approved):

magicus:
My reason for adding it was to do it in the "background" when you logged in, to make the quest book gui come up more snappy. But maybe the container query is so efficient that is not a problem?

kristofbolyai:
> My reason for adding it was to do it in the "background" when you logged in, to make the quest book gui come up more snappy. But maybe the container query is so efficient that is not a problem?

Wynn is slow, so this blocked interactions for a few seconds on world join.

magicus:
But otoh, now there will be once again a delay when opening the Wynntils quest book? Oh well, I guess it's better than freezing your character when you log in. (I never noticed this, though, but I agree that this sounds like a worse problem, if it happens)

kristofbolyai:
> But otoh, now there will be once again a delay when opening the Wynntils quest book? Oh well, I guess it's better than freezing your character when you log in. (I never noticed this, though, but I agree that this sounds like a worse problem, if it happens)

Yeah, okay conpromise

PR #708, opened 2022-11-19
fix: Rework how overlays positions work internally (and fix up some overlay positions)

kristofbolyai:
image

NOTE: This is breaking change to local configs. The mod will load, but Bottom/Right aligned overlays won't be at the correct places. Do NOT merge without discussion.


common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -176,7 +176,7 @@ public float getRenderX(OverlayPosition position) {
         return switch (position.getHorizontalAlignment()) {
             case Left -> section.x1() + position.getHorizontalOffset();
             case Center -> (section.x1() + section.x2() - this.getWidth()) / 2 + position.getHorizontalOffset();
-            case Right -> (section.x2() + position.getHorizontalOffset() - this.getWidth());
+            case Right -> section.x2() - position.getHorizontalOffset() - this.getWidth();
```
Incompleteusern: What is the benefit of this?

kristofbolyai: I started this PR off by this change. After the work I done in OverlayPosition, I think we do not need this breaking change. Reverted this.

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -86,35 +86,35 @@ public String toString() {
 
     public static OverlayPosition getBestPositionFor(
             Overlay overlay, float oldRenderX, float oldRenderY, float offsetX, float offsetY) {
-        Vec2 topLeftOfOverlay = new Vec2(overlay.getRenderX(), overlay.getRenderY());
+        Vec2 referencePoint = new Vec2(
+                overlay.getRenderX() + overlay.getWidth() / 2f, overlay.getRenderY() + overlay.getHeight() / 2f);
 
         // 1. Get the best section (section with the top left point of overlay)
```
P0keDev: it seems like you changed this code to use the overlay's center instead of top left (a change i think is good), but didn't update the comment? unless i am confused

kristofbolyai: Yes, sorry

P0keDev (approved):

common/src/main/java/com/wynntils/core/features/overlays/OverlayPosition.java
```diff
@@ -86,35 +86,35 @@ public String toString() {
 
     public static OverlayPosition getBestPositionFor(
             Overlay overlay, float oldRenderX, float oldRenderY, float offsetX, float offsetY) {
-        Vec2 topLeftOfOverlay = new Vec2(overlay.getRenderX(), overlay.getRenderY());
+        Vec2 referencePoint = new Vec2(
```
magicus, line 89: This reference point is the center of the overlay, yes? Do you think it would be helpful in the GUI to render it, so the user can see which section the center is inside? Or will it just be confusing?

PR #709, opened 2022-11-19
fix: Fix crash loops in feature exception

kristofbolyai:


common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -90,9 +94,11 @@ private static void handleExceptionInEventListener(Throwable t) {
 
         WynntilsMod.error("Exception in feature " + feature.getTranslatedName(), t);
         WynntilsMod.warn("This feature will be disabled");
-        McUtils.sendMessageToClient(new TextComponent("Wynntils error: Feature '" + feature.getTranslatedName()
-                        + "' has crashed and will be disabled")
-                .withStyle(ChatFormatting.RED));
+        if (sendMessageToUser) {
```
magicus: I think the instanceof test should just be here. It kind of makes sense that a function named `handleExceptionInEventListener` needs to be written defensively, so it won't do MC stuff that might crash again. It is not the callers responsibility to determine this.

magicus (approved):

PR #713, opened 2022-11-20
fix: ignore duplicated bomb bell messages

P0keDev:
This should fix bomb timers getting duplicated thanks to dialogue re-sending chat - since `BombInfo#equals` ignores the time variable, two bombs of the same type, server, and thrower can't be added to the set. 

For the entirely valid scenario of a bomb getting thrown again after expiring, the check for removing old timers now occurs before we try to add the new bomb.


Incompleteusern (approved):

kristofbolyai:
Thanks. I thought I did not need to do this because ChatModel handles this.. Anyways this should work because of bomb timer delay.

PR #714, opened 2022-11-20
feat: Add command aliases feature

kristofbolyai:


magicus (commented):
Hm. I dunno. I mean, if this is a requested feature or so, I won't say no. But if feels a bit ... ad-hoc..?

I'm thinking either perhaps we should support a general method of shortening commands? Or we should shorten it by parts, that is, allow "guild" to become "gu" or "g", as an independent thing from shortening the guild sub commands.

I thought Wynncraft allowed a short form of some commands. But maybe these were excluded, so they are just added to keep up consistence?

kristofbolyai:
> Hm. I dunno. I mean, if this is a requested feature or so, I won't say no. But if feels a bit ... ad-hoc..?
> 
> I'm thinking either perhaps we should support a general method of shortening commands? Or we should shorten it by parts, that is, allow "guild" to become "gu" or "g", as an independent thing from shortening the guild sub commands.
> 
> I thought Wynncraft allowed a short form of some commands. But maybe these were excluded, so they are just added to keep up consistence?

This is an actual "feature" in legacy too. Technically this can be used to create any kinds of command macros, not just shorten actual commands.

P0keDev (approved):
i would definitely be interested in expanding this feature further and allowing users to define new aliases via the config gui (in a later PR of course)

kristofbolyai:
> i would definitely be interested in expanding this feature further and allowing users to define new aliases via the config gui (in a later PR of course)

All you have to do is implement a `ConfigOptionElement` and register it in `getWidgetFromConfigHolder`. Then the rest is just GUI work.

PR #716, opened 2022-11-20
chore: 1.19.3 port

kristofbolyai:
This PR ports the whole mod to 1.19.3.


magicus:
Okay, so the time has finally come to take the bull by the horns... upgrading Minecraft version.

Are you thinking we should have a 1.19 and an 1.18 version, or should we drop 1.18?

magicus:
By a quick glance it looked like most changes were due to Component.literal. If we had our own wrapper, which could be compiled differently on 1.18 and 1.19, I wonder how many changes there would be then. Maybe few enough that it could be worth supporting several versions? 

I don't know what kind of interest there is in keeping an 1.18 version of Artemis, but I'm guessing the controversial "chat reporting" system in 1.19 is making at least some users queasy about using 1.19 on servers.

magicus:
(And finally, regarding the PR tags -- this is a major undertaking and functionality change for the end users, and does not seems fittingly described by "chore")

kristofbolyai:
> By a quick glance it looked like most changes were due to Component.literal. If we had our own wrapper, which could be compiled differently on 1.18 and 1.19, I wonder how many changes there would be then. Maybe few enough that it could be worth supporting several versions? 
> 
> 
> 
> I don't know what kind of interest there is in keeping an 1.18 version of Artemis, but I'm guessing the controversial "chat reporting" system in 1.19 is making at least some users queasy about using 1.19 on servers.

Supporting both versions is much harder than you would expect on first look. Based on the community vote I organised, 1.19 is prefered by far.

magicus:
Did you try using the fabric migration tool? `gradlew migrateMappings --mappings "1.19.2+build.28"` It's in no way perfect, but my impression is that it does a decent job of automatically converting some refactorings Mojang has done.

kristofbolyai:
> Did you try using the fabric migration tool? `gradlew migrateMappings --mappings "1.19.2+build.28"` It's in no way perfect, but my impression is that it does a decent job of automatically converting some refactorings Mojang has done.

I did not, but I did not need to. CTRL + F solved almost every problem :). The things it wouldn't solve would require my manual attention either way.

magicus:
For future upgrades it can be worth remembering. It makes a copy in "updatedSrc" (or something like that), and then you can compare it for any changes it makes, and chose to keep them if you think they are good.

fabric/build.gradle
```diff
@@ -18,7 +18,7 @@ buildscript {
          */
 
         dependencies {
-            classpath "com.github.kristofbolyai:EventBusTransformer:3fedef0169"
```
magicus: It's just weird that a central dependency -- that we develop ourselves -- should reside in personal name space.
I forked your repo into https://github.com/Wynntils/EventBusTransformer, so you can switch to that instead.

kristofbolyai:
@magicus I see you created an official fork of the event bus transformer. My version is a bit patchy, but coehlrich is working on a bit cleaner version.

kristofbolyai:
> @magicus I see you created an official fork of the event bus transformer. My version is a bit patchy, but coehlrich is working on a bit cleaner version.

I've let them known to use the official repo now. Anyways, a centralized repo is always better.

magicus:
I picked your repo as starting point since you updated this patch to use this version instead. I thought @coehlrich did not have time to work on it anymore. Do you want me to re-create the official repo starting with coehlrich's version instead?

kristofbolyai:
> I picked your repo as starting point since you updated this patch to use this version instead. I thought @coehlrich did not have time to work on it anymore. Do you want me to re-create the official repo starting with coehlrich's version instead?

It is fine I think, their repo is in the fork tree.

magicus (changes requested):

kristofbolyai:
Today marks the point where we have a building and loading 1.19.3 version, without obvious crashes.

common/src/main/java/com/wynntils/mc/EventFactory.java
```diff
@@ -497,27 +506,34 @@ public static SetSpawnEvent onSetSpawn(BlockPos spawnPos) {
...
+            for (ClientboundPlayerInfoUpdatePacket.Action action : packet.actions()) {
+                if (action == ClientboundPlayerInfoUpdatePacket.Action.UPDATE_DISPLAY_NAME) {
+                    GameProfile profile = entry.profile();
+                    if (entry.displayName() == null) continue;
+                    post(new PlayerDisplayNameChangeEvent(profile.getId(), entry.displayName()));
+                }
             }
         }
+
+        //            else if (action == Action.REMOVE_PLAYER) {
```
magicus: @kristofbolyai  This looks like a "FIXME". I assume we still want to post PlayerLogOutEvents...

kristofbolyai: No, I just forgot to remove this. If you look at this change, logout and other actions were separated with packets. Just below this is where I handle this.

PR #718, opened 2022-11-24
feat: added mana_pct function

joshieman06:
added function for mana percent and registered it. description still needed

### DIFF
--- a/common/src/main/java/com/wynntils/functions/CharacterFunctions.java
+++ b/common/src/main/java/com/wynntils/functions/CharacterFunctions.java
@@ -151,6 +151,20 @@ public class CharacterFunctions {
         }
     }
 
+    public static class ManaPctFunction extends DependantFunction<Float> {
+        @Override
+        public Float getValue(String argument) {
+            int currentMana = ActionBarModel.getCurrentMana();
+            int maxMana = ActionBarModel.getMaxMana();
+            return ((float) currentMana / maxMana * 100.0f);
+        }
+
+        @Override
+        public List<Class<? extends Model>> getModelDependencies() {
+            return List.of(ActionBarModel.class);
+        }
+    }
+
     public static class LevelFunction extends Function<Integer> {
         @Override
         public Integer getValue(String argument) {
--- a/common/src/main/resources/assets/wynntils/lang/en_us.json
+++ b/common/src/main/resources/assets/wynntils/lang/en_us.json
@@ -640,6 +640,8 @@
   "function.wynntils.mana.name": "Mana",
   "function.wynntils.manaMax.description": "Maximum possible mana",
   "function.wynntils.manaMax.name": "Mana[max]",
+  "function.wynntils.manaPct.description": "Your current mana as percentage of max",
+  "function.wynntils.manaPct.name": "Mana[%%]",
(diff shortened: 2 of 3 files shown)


common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -151,6 +151,20 @@ public List<Class<? extends Model>> getModelDependencies() {
         }
     }
 
+    public static class ManaPctFunction extends DependantFunction<Float> {
+        @Override
+        public Float getValue(String argument) {
+            int currentMana = ActionBarModel.getCurrentMana();
+            int maxMana = ActionBarModel.getMaxMana();
+            return ((float) currentMana / maxMana * 100.0f);
+        }
```
Incompleteusern, line 160: Perhaps inline `currentMana` and `maxMana`, as 
```java
        @Override
        public Float getValue(String argument) {
            return ((float) ActionBarModel.getCurrentMana() / ActionBarModel.getMaxMana() * 100.0f);
        }
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -640,6 +640,8 @@
   "function.wynntils.mana.name": "Mana",
   "function.wynntils.manaMax.description": "Maximum possible mana",
   "function.wynntils.manaMax.name": "Mana[max]",
+  "function.wynntils.manaPct.description": "Your current mana as percentage of max",
+  "function.wynntils.manaPct.name": "Mana[%%]",
```
kristofbolyai, line 644: why double %%

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
kristofbolyai: Okay either way imo

kristofbolyai: Although, shouldn't you round the number to 1 or 2 decimals?

magicus: This returns a Float. The rounding is done in presentation?

kristofbolyai: > This returns a Float. The rounding is done in presentation?

FunctionManager handles it. Right

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: I think it would have been needed in Legacy, where % were a meta character. It is not needed here.

joshieman06: That is how health_pct is written as well so I wanted to stay consistent.

kristofbolyai: okay, sure!

kristofbolyai (approved):
Looks good, thank you!

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: Sounds like we should fix health_pct then as well...

PR #720, opened 2022-11-24
feat: Add a GUI to create custom map icons, and add helper buttons to maps

kristofbolyai:


kristofbolyai:
image
image
image

kristofbolyai:
image

kristofbolyai:
image
image
image

magicus:
I'm a bit confused here. "Place a pin", does that mean "place your compass marker"? Can there be more than one compass marker/pin?

kristofbolyai:
> I'm a bit confused here. "Place a pin", does that mean "place your compass marker"? Can there be more than one compass marker/pin?

No, and it does mean that.

magicus:
But then you can shift-middle click to edit it? I understand how you could edit Waypoints, but not compass markers. So what is this pin?

Also, the "left click to delete" sounds dangerous. Left click is typically "select", so I'd expect something more difficult to delete, maybe ctrl-left click or so?

kristofbolyai:
I will fix this i18n as this is really confusing.

magicus:
I think the set of actions is really confusing. I understand it mostly copies Legacy, but I thought that was confusing as hell, as well. :-)

I would suggest that we start over a bit, and tries to make the UI a bit more sane. My suggestions:

* Left click should always be a "select" action. If we click on an existing marker (of any kind), it should get selected. We should then be able to do some further actions on it. Press Backspace to delete it, if it is user-placed. Or press "H" to hide it (and all of the same category), if it is an imported POI. Press "C" to set our compass ("pin") to point to it. Press "E" to edit it (if it is user-placed). 

All these keyboard shortcuts should also have some icons/toolbar buttons for these actions (compass, edit, delete, hide).

* Left clicking outside an existing POI should create a "temporary" POI (compare with how Google Maps works). This temporary marker could be turned into a new Waypoint using "N", or have it's compass set pointing to it using "C".

These two keyboard shortcuts should also be accessible via toolbar buttons.

For convenience, we would also have the following shortcuts:
* Double left clicking on an existing POI, or outside a POI, would be the same as pressing "C" (setting the compass target).
* Middle clicking would be equivalent of pressing "E" (edit), if possible
* Right clicking would be equivalent of pressing "N" (create new waypoint), if possible
* Shift+right click would be equvalent of pressing "Backspace" (delete marker), if possible.

kristofbolyai:
@magicus 
image

Here is the updated i18n. This is for sure more understandable

magicus:
Maybe what I am suggesting is too much to do in this PR. If so, I am okay that we bring in this PR now, and then tries to make this a bit more logical.

kristofbolyai:
> Maybe what I am suggesting is too much to do in this PR. If so, I am okay that we bring in this PR now, and then tries to make this a bit more logical.

I agree. I fixed the current additions to be more specific. A new PR should target refactoring this and adding better binds (without making people used to the old system hate it)

magicus:
Your new i18n was definitely an improvement! I think we should use "the compass marker" and not "a compass marker", if there can only be one.

magicus:
Yeah, maybe we need a legacy compatibility mode anyway. So, let's start there then, and add a "sane" mode later on. :)

Can you make "center on your compass marker" a toolbar button? I did not even knew we had that functionality; it never occurred to me that it could be a mouse button combination. Maybe have a button that if you press it once, it centers on your current position, and if you press it once more, it centers on the compass marker, and if you press it a third time, it centers on the center of the map (and then it repeats)..?

kristofbolyai:
> Yeah, maybe we need a legacy compatibility mode anyway. So, let's start there then, and add a "sane" mode later on. :)
> 
> Can you make "center on your compass marker" a toolbar button? I did not even knew we had that functionality; it never occurred to me that it could be a mouse button combination. Maybe have a button that if you press it once, it centers on your current position, and if you press it once more, it centers on the compass marker, and if you press it a third time, it centers on the center of the map (and then it repeats)..?

Is it fine if I PR it later to not add even more stuff in this PR? I will get it on my TODO list :)

magicus:
... but maybe that is for the wishlist of a future PR. :) Sorry for getting carried away. I'll review this actual code now.

common/src/main/java/com/wynntils/gui/screens/maps/PoiCreationScreen.java
```diff
@@ -0,0 +1,483 @@
...
+        updateSaveStatus();
+        firstSetup = false;
+    }
+
+    @Override
+    public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        super.renderBackground(poseStack);
+        super.render(poseStack, mouseX, mouseY, partialTick);
+
+        FontRenderer.getInstance()
```
magicus, line 266: Is there like no `Label` renderable widget we can use and add to the GUI, instead of having to render this all the time?

magicus (approved):

common/src/main/java/com/wynntils/gui/screens/maps/PoiCreationScreen.java
kristofbolyai: Even if there is one, this is basically the same computation.

magicus:
Looks okay. Minecraft GUI code is kind of horribly verbose. I don't know if there's anything we can do about it. I think we could gain a bit by creating some more helpers that are expressed at a higher abstraction layer, but it's not the most pressing issue.

Issue #261, opened 2022-08-09
Wynncraft resends entire chat history line by line in some quests


kristofbolyai:
@magicus is this something ChatManager should handle?

magicus:
Hm. I'm not sure how or when this happens. Is this if you have NPC dialogue overlay active, and the multi-line extraction enabled?

magicus:
Nothing should really "resend" chat.

kristofbolyai:
Yes, didn't test it while disabled.

magicus:
So you mean that the +1 soul points are sent twice? What's triggering the "resend"?

kristofbolyai:
Not twice, they are basically sent again when a npc dialogue is sent.

magicus:
Was that the last message before the NPC dialog started?

I try to locate the last sent message by searching for the message text, stripped of formatting, to be able to detect if new messages arrive. This is inherently a bit hit-and-miss, where I could potentially misidentify messages. I can't really figure out how that would happen in this case, though...

kristofbolyai:
https://paste.ee/p/n6G7C

magicus:
Thanks for the log. Seems there are many issues I need to address there. Tracking a quest with a book seems to trigger the same "screen" like behavior, for instance.

magicus:
Also, some NPC dialogues seems to be not matched.

kristofbolyai:
It seems to be caused by NPC dialogue without "shift to continue". At least all of those types of NPC dialogue is broken.

magicus:
Where can I find such dialogues? Is it like the initial quest where the caravan driver starts shouting at me?

kristofbolyai:
Yeah, or the corrupted village iirc.

P0keDev:
> It seems to be caused by NPC dialogue without "shift to continue". At least all of those types of NPC dialogue is broken.

does this type of dialogue trigger the same chat replacement feature as normal dialogue? i recall it not doing so but maybe i'm wrong

magicus:
I did not think it did. I just thought these NPCs called out like normal chat messages. I'll have to look into it.

magicus:
I've done some tests and analysis on this.

1) Tracking a quest does not cause "screens". I was worried this might be the case since old-style Wynncraft tracking continuously printed the tracking info to the chat. But that was skipped when they introduced scoreboard tracking. So the quest book tracking was just a red herring.

[...]

PR #722, opened 2022-11-24
feat: Redirect "not enough mana", healing and speed effect messages

kristofbolyai:


kristofbolyai:
No background message regex because spells cannot be cast when in dialouge.

kristofbolyai:
> No background message regex because spells cannot be cast when in dialouge.

Even if there is a way it would appear, it is not worth to filter for it.

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -106,6 +108,9 @@ public class InfoMessageFilterFeature extends UserFeature {
     @Config
     private FilterType craftedDurability = FilterType.REDIRECT;
 
+    @Config
+    private FilterType notEnoughMana = FilterType.REDIRECT;
+
     @SubscribeEvent
     public void onInfoMessage(ChatMessageReceivedEvent e) {
         if (e.getRecipientType() != RecipientType.INFO) return;
```
Incompleteusern, line 130: Not strictly this pr, but we need to filter dialogue messages from this check and spamming queue

kristofbolyai: There is already an issue for it

kristofbolyai: https://github.com/Wynntils/Artemis/issues/261

P0keDev (approved):

magicus:
>    >No background message regex because spells cannot be cast when in dialouge.
>
> Even if there is a way it would appear, it is not worth to filter for it.

Just to clarify: of course spells can be cast when in dialogue. You can do whatever you want, it's just that Wynncraft will keep resending chat history until the NPC has finished talking. But your second assessment might still apply, that it is not worth it to fix.

kristofbolyai:
> > > No background message regex because spells cannot be cast when in dialouge.
> > 
> > 
> > Even if there is a way it would appear, it is not worth to filter for it.
> 
> Just to clarify: of course spells can be cast when in dialogue. You can do whatever you want, it's just that Wynncraft will keep resending chat history until the NPC has finished talking. But your second assessment might still apply, that it is not worth it to fix.

No, it did not seem to me like that I can cast anything. Weird...

magicus:
What? Is that so? I need to check that :)

edit: You are absolutely correct. Funny I never realized that.

PR #723, opened 2022-11-25
fix: emerald pouch parsing when picking emeralds up

joshieman06:
emerald pouches now register even when picking up emeralds (needed for emerald info vars)

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
+++ b/common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
@@ -104,12 +104,11 @@ public final class WynnItemMatchers {
                 && itemStack.getDamageValue() <= 6);
     }
 
-    public static boolean isEmeraldPouch(ItemStack itemStack) {
+    public static boolean isEmeraldPouch(ItemStack itemStack) {        
         return itemStack instanceof EmeraldPouchItemStack
-                || (itemStack.getItem() == Items.DIAMOND_AXE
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
```diff
@@ -107,6 +107,8 @@ public static boolean isUnidentified(ItemStack itemStack) {
     public static boolean isEmeraldPouch(ItemStack itemStack) {
         return itemStack instanceof EmeraldPouchItemStack
```
kristofbolyai: Imo lets refactor this!

1, If itemStack is EmeraldPouchStack, return true.

2, Otherwise return Item is (Golden Shovel or Diamond Axe) AND name is "...". Also, maybe add a comment that there are two possible items because of the pickup "plus texture".

joshieman06: Done! Lmk if you wanted the EmeraldPouchStack in a conditional, still using an or operator

kristofbolyai: > Done! Lmk if you wanted the EmeraldPouchStack in a conditional, still using an or operator 

I think we should return early if the instanceof matches. So I think do the "conditional" way.

kristofbolyai (approved):
Thank you for fixing this!

PR #725, opened 2022-11-25
feat: add money, em, eb and le functions

joshieman06:


common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -317,6 +317,10 @@ public static void init() {
         registerFunction(new CharacterFunctions.XpPctFunction());
         registerFunction(new CharacterFunctions.BpsFunction());
         registerFunction(new CharacterFunctions.BpsXzFunction());
+        registerFunction(new CharacterFunctions.MoneyFunction());
```
kristofbolyai, line 320: Can you preserve alphabetical order? (forgot to do this for Bps too)

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -188,6 +189,38 @@ public List<Class<? extends Model>> getModelDependencies() {
         }
     }
 
+    public static class MoneyFunction extends Function<Integer> {
```
kristofbolyai, line 192: I am not fond of this. At first sight this looks sooooo innocent. But!, what you do here is parsing every item in your container and summing up emerald count. That is LOTS of allocations. Not to even mention that if we use multiple of these functions, we do this multiple times. 

So, how do we fix this? Simple. Create a new model, `PlayerInventoryModel`. Parse emerald count there and cache it. You can use `ContainerSetContent` and `SetSlotEvent` to know when to update your cache, and `WorldStateChangeEvent` to know when to reset the cache. The functions can depend on the model (which can even parse emerald count on init, so there is no need to relog to make it work). When you done this, you can go ahead, and use this cache in `InventoryEmeraldCountFeature`, making it much better too! This way, you not only added new functions, but also improved an existing feature.

kristofbolyai: Let me know if you are having problems understanding this. I tried to be as clear as I can, but you can always ask for help :)

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -188,6 +189,38 @@ public List<Class<? extends Model>> getModelDependencies() {
...
+        @Override
+        public Integer getValue(String argument) {
+            int emeralds = ContainerUtils.getEmeraldCountInContainer(McUtils.inventoryMenu());
+            return emeralds;
+        }
+    }
+
+    public static class LeFunction extends Function<Integer> {
```
kristofbolyai, line 200: I would rename these to be LiquiedEmeraldFunction (same for Eb and Em). You can override `getAliases` and add "le", "eb" and "em" as an alias respectively.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -652,6 +658,8 @@
   "function.wynntils.mem_pct.name": "Mem[%%]",
   "function.wynntils.mem_used.description": "Current amount of memory used by the JVM",
   "function.wynntils.mem_used.name": "Mem[used]",
+  "function.wynntils.money.description": "Amount of emeralds in inventory",
+  "function.wynntils.money.name": "Total emerald cunt",
```
```suggestion
  "function.wynntils.money.name": "Total emerald count",
```

Ooopise :)

kristofbolyai (commented):
Nice additions. You do need to cache emerald count to a model, but I hope you are happy that you can learn about a core part about Artemis, `Models`. 

Models in our code are a way to abstract wynn-related knowledge. This way, our features/functions/overlays only represent the data, but models parse it. This means we can also share caches/code (exactly like the case now!).

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
joshieman06: I will get on that

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
magicus: @kristofbolyai I totally agree with what you are saying about caching this in a model, but I'm just thinking that maybe listening on the container events is not enough? If you run over emeralds and pick them up, will the `ContainerSetContent` event really be sent? I thought that was only if you have your inventory screen open.

So maybe we'll need a new mixin and event to capture when the player picks up items. And that might be a bit too complicated to ask of a newcomer to implement in what was supposed to be a simple PR.

kristofbolyai: Player inventory is a container too. I think on pickup, wynn always resends the inventory because it reorders stuff etc. Would need testing for sure though.

kristofbolyai: > So maybe we'll need a new mixin and event to capture when the player picks up items. And that might be a bit too complicated to ask of a newcomer to implement in what was supposed to be a simple PR.

I am happy to write the mixin if it turns out to be needed.

joshieman06:
will reopen when i figure out how models work

PR #726, opened 2022-11-25
fix: Allow clicks on locked emerald pouch and ability tree slots

kristofbolyai:


Incompleteusern (dismissed):

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
```diff
@@ -89,6 +96,12 @@ public void onInventoryClickEvent(ContainerClickEvent event) {
             return;
         }
 
+        // We want to allow interactions (except throwing) with emerald pouch even if locked
+        if (event.getClickType() != ClickType.THROW
+                && WynnItemMatchers.isEmeraldPouch(slotOptional.get().getItem())) {
+            return;
+        }
+
```
P0keDev: why should emerald pouches have this exception? if someone has the "block interaction" option enabled and locks an emerald pouch, i don't see why they should expect to still use it like this.

throw actions will get blocked regardless of the setting, so this makes even less sense to me

P0keDev (changes requested):

common/src/main/java/com/wynntils/features/user/ItemLockFeature.java
P0keDev: the reason i don't like this is that there's a very legitimate use case for blocking emerald pouches - you can essentially force close a chest by right clicking on a pouch, and a user might lock their pouch slot specifically to avoid doing that

kristofbolyai: I will make it a setting. People complained, as far as I know, this is what legacy did

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -287,6 +287,8 @@
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightColor.name": "0 Star Ingredient Highlight Color",
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightEnabled.description": "Enable highlighting of 0 star ingredient items",
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightEnabled.name": "0 Star Ingredient Highlight",
+  "feature.wynntils.itemLock.allowClickOnEmeraldPouchInBlockingMode.description": "If Block ALl Actions on Locked Items is enabled, should clicking on the Emerald Pouch be allowed?",
```
P0keDev: capitalization problem

```suggestion
  "feature.wynntils.itemLock.allowClickOnEmeraldPouchInBlockingMode.description": "If Block All Actions on Locked Items is enabled, should clicking on the Emerald Pouch be allowed?",
```

P0keDev (approved):

magicus:
This reminds me of a long-standing bug in Legacy: when entering creative mode in your housing plot, locked slots should not apply either.

PR #727, opened 2022-11-25
chore: Update gradle [skip ci]

kristofbolyai:


magicus:
What version of gradle is this? Can you update the PR description to mention that?

magicus (approved):

kristofbolyai:
> What version of gradle is this? Can you update the PR description to mention that?

I am not exactly sure. It seems like gradle updated itself when I was working on the project. From looking at the diff, it definitely seems as a new gradle version.

magicus:
Eh. I did not thought it could do that? Maybe we should re-do this more controlled. Let me check in my environment.

magicus:
It is supposed to only update this if you run like `gradle wrapper --gradle-version 7.3`.

magicus (changes requested):
Latest version of gradle is 7.6. I think it's better to skip this "what is this I don't even" random change and make a controlled upgrade to 7.6 using the gradle wrapper command.

magicus (approved):

PR #728, opened 2022-11-26
feat: prevent trades and duels in combat

P0keDev:
Essentially ports the functionality of legacy's equivalent feature - blocks attack and interact packets on other players while holding a weapon, so mashing buttons in combat doesn't give you unwanted trade/duel requests.

I do wish that a further description could be added to features right now, because fitting a proper explanation into just the feature name isn't really possible


common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -522,6 +522,7 @@
   "feature.wynntils.powderAbilityBarOverlay.overlay.powderAbilityBar.onlyIfWeaponHeld.name": "Only Show When Weapon Held",
   "feature.wynntils.powderAbilityBarOverlay.overlay.powderAbilityBar.textShadow.description": "What should the text shadow look like?",
   "feature.wynntils.powderAbilityBarOverlay.overlay.powderAbilityBar.textShadow.name": "Text Shadow",
+  "feature.wynntils.preventTradesDuels.name": "Prevent Trades & Duels",
```
Incompleteusern, line 531: Lacks `.description`

P0keDev: features do not have description fields - only specific config options do

Incompleteusern: Oops i might've gone crazy, sorry

Incompleteusern (approved):

magicus:
Is the idea that you manually toggle this feature on and then off when going into combat?

Otherwise it seems to me like a natural extension of this functionality would be to:

1. have a hotkey to allow toggling this behavior on and off, and
2. have a gui indicator telling if it is enabled or not

Something like the "locked item slot".

If that were added, then the user could keep the feature itself enabled at all time, and only activate the functionality when needed.

kristofbolyai:
> Is the idea that you manually toggle this feature on and then off when going into combat?
> 
> Otherwise it seems to me like a natural extension of this functionality would be to:
> 
> 1. have a hotkey to allow toggling this behavior on and off, and
> 2. have a gui indicator telling if it is enabled or not
> 
> Something like the "locked item slot".
> 
> If that were added, then the user could keep the feature itself enabled at all time, and only activate the functionality when needed.

You can just trade/duel with a non-weapon. I don't think we need to overcomplicate this

magicus:
And yes, a `<feature>.desc` addition to the i18n could probably be a good idea. It's just a bit of a chore at this time with the copywriting... Maybe we should make it optional.

magicus:
> You can just trade/duel with a non-weapon. I don't think we need to overcomplicate this

Ah, good point. :)

magicus:
So maybe just change the name to `Prevent Trades & Duels while holding weapon` and be done with it?

PR #730, opened 2022-11-26
fix: Label pois should not be behind ServicePois

kristofbolyai:


P0keDev (approved):

magicus (approved):

magicus:
Maybe we should rethink what "y" means for POIs. Some POIs have a real Y value, like NPCs. That is useful to have when calculating distance or with the `/locate` command.

But when drawing POIs on the map, the "real" Y value is definitely less useful. Sure, there might be cases where e.g. some NPC is underground, and another one above it, and it would make sense to render the latter above the former. But otherwise the "Y" value when rendering maps is more about the question "what is allowed to overwrite anything other". 

Maybe POIs should have a `priority` tag, which is what mostly determines when it gets rendered? Let's say it has a value between 0 and 10, and you basically calculate `mapRenderY = realY + priority*256`. Then a Identifier positioned above another Identifier will still be correctly ordered, but it will never shadow a label.

PR #736, opened 2022-11-27
feat: Auto create waypoints for chests

kristofbolyai:


common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -81,4 +104,66 @@ public class MapFeature extends UserFeature {
     public List<Class<? extends Model>> getModelDependencies() {
         return List.of(MapModel.class);
     }
+
+    @SubscribeEvent
+    public void onRightClick(PlayerInteractEvent.RightClickBlock event) {
+        if (!autoWaypointChests) return;
+
+        lastChestPos = event.getPos();
```
Incompleteusern: check if chest?

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -81,4 +104,66 @@ public class MapFeature extends UserFeature {
...
+
+        ChestTier tier = ChestTier.fromString(matcher.group(1));
+
+        if (tier.ordinal() < minTierForAutoWaypoint.ordinal()) return;
+
+        MapLocation location = new MapLocation(lastChestPos.getX(), lastChestPos.getY(), lastChestPos.getZ());
+        CustomPoi newPoi = new CustomPoi(
+                location, tier.getWaypointName(), CommonColors.WHITE, tier.getWaypointTexture(), Integer.MIN_VALUE);
+
+        if (MapFeature.INSTANCE.customPois.stream().noneMatch(customPoi -> customPoi.equals(newPoi))) {
```
Incompleteusern, line 137: I wonder if we could store chestPois as a hashset separately.

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: no need to imo

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: Too much complication imo

Incompleteusern: It would be better for importing/exporting chest pois maybe if they were held separately

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
Incompleteusern: Could be caused by lag to have lastChestPos not match screen

kristofbolyai: Maybe, but this is the best we can do. Legacy does the same. Ill add block checking later

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: We could filter for them ig. I don't wanna separate POIs into multiple sets. Maybe later we could have some kind of poi metadata to indicate it

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: Added the check.

Incompleteusern (approved):

PR #740, opened 2022-11-27
fix: Fix keybinding not getting bind on registration

kristofbolyai:
Needs a review by @Incompleteusern.

When we run a new instance, it does not have options.txt saved, we can't load the keybind options back from it, making every keybind save as unbind. Hence, we should not assert that options loading loads our keybinds.

I removed this "reset". I tested making a conflicting keybind, disabling the feature, the conflict resolved. Then I turned the feature back on, the keybind still worked.

I also tested the case which I think made this part of the code exist. Disabling a feature with a conflicting keybind, restarting the game, and seeing if the keybind gets registered when the feature is off. The conflict does not happen.

I also tested running a clean install, it works flawlessly.


common/src/main/java/com/wynntils/core/keybinds/KeyBind.java
```diff
@@ -49,6 +49,11 @@ public KeyBind(String name, int keyCode, boolean firstPress, Runnable onPress, C
     public KeyBind(String name, int keyCode, InputConstants.Type type, boolean firstPress, Runnable onPress) {
         this.firstPress = firstPress;
         this.keyMapping = new KeyMapping(name, type, keyCode, CATEGORY);
+
+        // Unbind keybind, bound after registration by options reload
```
kristofbolyai, line 53: Just so you are not confused this is the other constructor

Incompleteusern (approved):

PR #742, opened 2022-11-27
feat: Parse coordinates from chat and inject clickable component and add /compass share

kristofbolyai:


common/src/main/java/com/wynntils/features/user/ChatCoordinatesFeature.java
```diff
@@ -0,0 +1,107 @@
...
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onChatReceived(ChatMessageReceivedEvent e) {
+        if (!WynnUtils.onWorld()) return;
+
+        Component message = e.getMessage();
+
+        e.setMessage(insertCoordinateComponents(message));
+    }
+
+    public static Component insertCoordinateComponents(Component message) {
```
kristofbolyai, line 44: Yes, this is taken from ChatItemModel. I don't think we need to generalize this, as we don't really have any other use case for this (and it would be weird)

common/src/main/java/com/wynntils/features/user/ChatCoordinatesFeature.java
```diff
@@ -0,0 +1,107 @@
...
+        for (Component comp : components) {
+            Matcher m = LocationUtils.strictCoordinateMatcher((ComponentUtils.getCoded(comp)));
+            if (!m.find()) {
+                Component newComponent = comp.copy();
+                temp.append(newComponent);
+                continue;
+            }
+
+            do {
+                String text = ComponentUtils.getCoded(comp);
```
magicus, line 64: I'm sorry, I don't really have time to fully wrap my head around this. I hope you've thought about how the coordinates might appear in the middle of formatting and that you need to make sure you keep the formatting properly on both sides of the inserted onHover/onClick component.

common/src/main/java/com/wynntils/features/user/ChatCoordinatesFeature.java
magicus: I'm starting to think that maybe the ChatModel should have some kind of "transformers" registered on it. And "redirectors", that capture chats that should be redirected to the notification manager. Just a hand-wavy idea, not requesting that you change anything here.

magicus (approved):
Looks good. Just say "I've checked that I am doing the right thing when splitting out coordinates" and I'll believe you. :)

kristofbolyai:
> Looks good. Just say "I've checked that I am doing the right thing when splitting out coordinates" and I'll believe you. :)

This is the same logic chat items uses and that works (+ Ive fixed it already a while ago:))

common/src/main/java/com/wynntils/features/user/ChatCoordinatesFeature.java
kristofbolyai: Perhaps, although this case is rare enough not to add it for now

PR #743, opened 2022-11-27
feat: Redirect Blacksmith Sell/Scrap Messages to NotificationManager Overlay

san7890:
Hey there,

I missed this from Wynntils Legacy, so I just decided to spend time to get it ported over from there. Has essentially the same features as the old version, although with a few new things:

Now relies completely on Regex Groups, as opposed to splitting it out based on the Section Character (don't have to toy around with sentence fragments).
Now handles the message when you sell/scrap multiple items to be singular/plural instead of just relying on an `item(s)` crutch (in the photographs below).

image

image

image

image


common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onChat(ChatMessageReceivedEvent event) {
+        EnumMap<ItemTier, Integer> totalItems = new EnumMap<>(ItemTier.class);
+        // Tracks count of sold or scrapped items
+        int totalItemInteger = 0;
+        // If we get emeralds, this will be the number of emeralds we get. If we get scrap, this will be the number of
+        // scrap we get.
+        String paymentString = "";
+        Matcher messageMatcher = BLACKSMITH_MESSAGE_PATTERN.matcher(event.getCodedMessage());
```
kristofbolyai: This check should be the first thing in the event

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,102 @@
...
+                    .group(1)
+                    .charAt(0)); // find the color code, then get the ChatFormatting of the item to match it to a tier.
+            ItemTier tierToIncrease = ItemTier.fromChatFormatting(itemColor);
+            if (tierToIncrease == null) continue;
+            totalItems.put(tierToIncrease, totalItems.getOrDefault(tierToIncrease, 0) + 1);
+            totalItemInteger++;
+        }
+
+        // The final part of the message.
+        {
```
kristofbolyai: Why the extra block? Remove it

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,102 @@
...
+                sendableMessage.append(ChatFormatting.LIGHT_PURPLE + "Scrapped " + totalItemInteger);
+                sendableMessage.append(" item" + pluralizer + " for ");
+                sendableMessage.append(
+                        ChatFormatting.YELLOW + paymentString + " scrap" + ChatFormatting.LIGHT_PURPLE + ".");
+            }
+
+            String sendableMessageString = sendableMessage.toString();
+
+            // Finally, we send the message.
+            NotificationManager.queueMessage(sendableMessageString);
```
kristofbolyai: I am not happy about this whole string generation part, it is a bit of a mess. 

Let's refactor this:
1, Create a StringBuilder to create the sold items (`1/0/0/0/0` part)
2, If we sold items -> "Sold %d (%s) item(s) for %dE",
    otherwise -> "Scrapped %d (%s) item(s) for %dscrap",

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -6,6 +6,7 @@
   "core.wynntils.category.redirects": "Game Notification Redirects",
   "core.wynntils.category.tooltips": "Item Tooltips",
   "core.wynntils.category.uncategorized": "Uncategorized",
+  "feature.wynntils.BlacksmithRedirectFeature.name": "Blacksmith Message Redirect",
```
kristofbolyai: This should not even run in dev env.
```suggestion
  "feature.wynntils.blacksmithRedirect.name": "Blacksmith Message Redirect",
```

kristofbolyai:
Also, why don't we include item count for scrapped items?  Either include it or only create the item count stringbuilder if we sold the items so we don't waste resources

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,112 @@
...
+            String itemPluralizer = totalItemInteger == 1 ? "item" : "items";
+
+            // Build up the string that outlines how many items were sold in what tier (0/0/0/0/0/0/0/0).
+            StringBuilder countByTier = new StringBuilder();
+            for (ItemTier tier : ItemTier.values()) {
+                countByTier.append('/' + tier.getChatFormatting().toString() + totalItems.getOrDefault(tier, 0));
+                countByTier.append(ChatFormatting.LIGHT_PURPLE);
+            }
+            countByTier.append(")");
+            countByTier.setCharAt(0, '(');
```
kristofbolyai: Add these in the format string below

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,112 @@
...
+
+            String countByTierString = countByTier.toString();
+
+            // Sold 1 (1/0/0/0/0/0/0/0) item for 4e.
+            sendableMessage = String.format(
+                    "%sSold %d %s %s for %s%s%s%s.",
+                    ChatFormatting.LIGHT_PURPLE,
```
kristofbolyai: You can include colors int the format string itself to make this look less weird

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,112 @@
...
+                    totalItemInteger,
+                    itemPluralizer,
+                    countByTierString,
+                    ChatFormatting.GREEN,
+                    paymentString,
+                    EmeraldSymbols.EMERALDS,
+                    ChatFormatting.LIGHT_PURPLE);
```
kristofbolyai: same

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,112 @@
...
+        else {
+            Matcher itemMatcher =
+                    ITEM_PATTERN.matcher(messageMatcher.group(2)); // Second group contains all of the items.
+            while (itemMatcher.find()) {
+                totalItemInteger++;
+            }
+            String itemPluralizer = totalItemInteger == 1 ? "" : "s";
+
+            sendableMessage = String.format(
+                    "%sScrapped %d %s for %s%s scrap%s.",
```
kristofbolyai: Actually, is there a reason why we  don't have item tier breakdown here? Even if legacy didnt do it, I think it is nice to have it

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
```diff
@@ -0,0 +1,112 @@
...
+                if (tierToIncrease == null) continue;
+                totalItems.put(tierToIncrease, totalItems.getOrDefault(tierToIncrease, 0) + 1);
+                totalItemInteger++;
+            }
+            String itemPluralizer = totalItemInteger == 1 ? "item" : "items";
+
+            // Build up the string that outlines how many items were sold in what tier (0/0/0/0/0/0/0/0).
+            StringBuilder countByTier = new StringBuilder();
+            for (ItemTier tier : ItemTier.values()) {
+                countByTier.append('/' + tier.getChatFormatting().toString() + totalItems.getOrDefault(tier, 0));
```
kristofbolyai: Use chained append here, Using + defeats the purpose of using stringbuilder

kristofbolyai (commented):
much better now! A few changes and it will be ready

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
san7890: Feels more cluttered to me IMO, but I can look into adding it. The only technical concern I can think of it the message overwriting it's bounds, like something I ran into where you might see:

"Scrapped 5 items for 16 scrap."
versus
"Scrapped 5 (0/5/0/0/0/0/0/0) items for
16 scrap"

might not matter though and we could be fine, the only concern is the relative fingerprint the sendablemessage will have

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
san7890: This is what cleans it up from:

`/0/0/0/0/0/0/0/0/`

to:

`(0/0/0/0/0/0/0/0)`

Not sure how I would go about adding that to the format sting.

common/src/main/java/com/wynntils/features/user/redirects/BlacksmithRedirectFeature.java
joshieman06: could ignore 0s (so instead of 0/5/0/0/0/0/0/1 it would be 5/1 but color coded)

san7890:
Thank you!

kristofbolyai:
@Incompleteusern @magicus can one of you review this, as I have contributed to it too?

Incompleteusern (approved):

PR #744, opened 2022-11-28
feat: Remove Quilted Fabric API dependency

kristofbolyai:


kristofbolyai:
This improves build times and development environment external sources a lot.

quilt/src/main/resources/quilt.mod.json
```diff
@@ -16,10 +16,6 @@
         "id": "quilt_loader",
         "version": "*"
       },
-      {
-        "id": "quilt_base",
-        "version": "*"
-      },
```
Incompleteusern, line 22: why is this being removed?

kristofbolyai: We don't need this as a dependency now as we include it in the mod

Incompleteusern (approved):

PR #745, opened 2022-11-28
feat: Redirect Horse Despawning Messages to NotificationManager Overlay 

san7890:
Ports feature from Wynntils Legacy where game information regarding the horse not being able to spawn / despawning going to the overlay instead of staying in chat.

I also made it so the auto-mounting functionality would send its messages to the overlay manager as well.

image

image


common/src/main/java/com/wynntils/features/user/redirects/HorseRedirectFeature.java
```diff
@@ -0,0 +1,51 @@
...
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@FeatureInfo(category = FeatureCategory.REDIRECTS)
+public class HorseRedirectFeature extends UserFeature {
+    // Strings and patterns that we get from Wynncraft.
+    private static final String noRoomString = "There is no room for a horse.";
+    private static final String despawnedString = "Since you interacted with your inventory, your horse has despawned.";
+    private static final Pattern NO_ROOM_PATTERN = Pattern.compile(noRoomString);
+    private static final Pattern HORSE_DESPAWNED_PATTERN = Pattern.compile(despawnedString);
```
kristofbolyai: Just a reminder, no need to separate the strings and patterns here. 

Use colored messages in the patterns to make sure it works.

common/src/main/java/com/wynntils/features/user/redirects/HorseRedirectFeature.java
```diff
@@ -0,0 +1,51 @@
...
+import com.wynntils.wynn.event.ChatMessageReceivedEvent;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.TranslatableComponent;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@FeatureInfo(category = FeatureCategory.REDIRECTS)
+public class HorseRedirectFeature extends UserFeature {
```
kristofbolyai: Don't make this a new feature. You can "easily" fold this into InfoMessageRedirectFeature. There are plenty of examples there.

kristofbolyai (commented):
You are getting there, but you should use our existing feature for this :^)

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -321,6 +328,30 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
...
 
+            if (horse != FilterType.KEEP) {
+                Matcher roomMatcher = NO_ROOM_PATTERN.matcher(msg);
+                if (roomMatcher.matches()) {
+                    e.setCanceled(true);
+                    if (horse == FilterType.HIDE) {
+                        return;
+                    }
+                    NotificationManager.queueMessage(
+                            new TextComponent("No room for a horse!").withStyle(ChatFormatting.DARK_RED));
```
magicus, line 339: @kristofbolyai This is missing i18n. When checking the file, I notice that all of these notifications are missing i18n. We should probably do something about it.

I'll open a bug so we don't forget.

kristofbolyai: No redirect in this feature has an i18n atm.

PR #750, opened 2022-11-29
feat: Redirect Potion Stack Messages to NotificationManager Overlay

san7890:
image

I don't like un-necessary things in the subtitle, and this was one of them. I added it to the PouchRedirect (now InventoryRedirect to represent the general nature of it) code since Wynncraft sends these messages in the exact same way as the other two, and I didn't find a better place to put it. I also decided to create a new string to send to the NotificationManager overlay since the raw string's formatting looks quite ugly when it isn't so large and in your face.


Incompleteusern (approved):

kristofbolyai:
Why is the message not red?

san7890:
> Why is the message not red?

I defaulted to green since that's the message color in the subtitle event. I also think red should be reserved for "warning" messages, while this is more of a friendly message.

Are you referring to the health potion aspect? Wynncraft never said if it was a health potion in the subtitle... maybe we made that assumption in Legacy since it's the only potion you "charge"? Unsure

kristofbolyai:
> > Why is the message not red?
> 
> I defaulted to green since that's the message color in the subtitle event. I also think red should be reserved for "warning" messages, while this is more of a friendly message.
> 
> Are you referring to the health potion aspect? Wynncraft never said if it was a health potion in the subtitle... maybe we made that assumption in Legacy since it's the only potion you "charge"? Unsure

Fine this way ig. I just expected it to be red

PR #751, opened 2022-11-29
refactor: Item stack transformers as models

P0keDev:
So far, just some preliminary work for representing item stack transformation as a set of individual models that can be defined as dependencies and individually enabled/disabled.

I can copy the gist of what I've done with item stacks over to item properties without much trouble, but I wanted to make sure there were no objections to this approach in general. Unfortunately this does feel like a good bit of code duplication/bloat, but I don't think it can be avoided given the framework of the model system - correct me if I'm wrong on that though.


magicus:
What is the purpose? To be able to turn on and off individual transformations?

kristofbolyai:
> What is the purpose? To be able to turn on and off individual transformations? 

Yes. And we will create our first manager-submodel-feature dependency; something that ChatModel could use too.

magicus (changes requested):

PR #753, opened 2022-11-29
feat: Redirects Territory (Town) Entering/Leaving Messages to NotificationManager Overlay

san7890:
image

More trimming down on having these messages (that get old fast) being in the subtitle. Since there was both a subtitle and a chat message that said the exact same thing, I put it in it's own file to handle both instances of it showing up- let me know if I should just roll it elsewhere. I figure this way is probably better so you only have to edit one regex should Wynncraft change how it sends these messages and don't have to worry about desync.

Also, let me know if there are any recommendations for altering the formatting of the message itself. I went through about five different variations and this was the only one that I liked, will appreciate feedback on making it a bit better if it looks ugly.


common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,53 @@
...
+        String direction = matcher.group(1);
+        String rawTerritoryName = matcher.group(2);
+        // Want to account for weird stuff like "the Forgery" and make it "The Forgery" for the sake of our brief
+        // message (looks odd otherwise).
+        String territoryName = rawTerritoryName.substring(0, 1).toUpperCase() + rawTerritoryName.substring(1);
+
+        String enteringMessage = String.format("§7Now §o%s§r %s", direction, territoryName);
+        NotificationManager.queueMessage(enteringMessage);
+    }
+    // Handles the chat log message event, we don't want a duplicate so just cancel the event and rely on the subtitle
```
kristofbolyai, line 55:
```suggestion
    
    // Handles the chat log message event, we don't want a duplicate so just cancel the event and rely on the subtitle
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,53 @@
...
+import com.wynntils.core.notifications.NotificationManager;
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import com.wynntils.mc.utils.ComponentUtils;
+import com.wynntils.wynn.event.ChatMessageReceivedEvent;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class TerritoryMessageRedirectFeature extends UserFeature {
```
kristofbolyai, line 20: Again, why did we not fold this into InfoMessageFilter?

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,53 @@
...
+        String codedString = ComponentUtils.getCoded(event.getComponent());
+        Matcher matcher = TERRITORY_MESSAGE_PATTERN.matcher(codedString);
+        if (matcher.matches()) event.setCanceled(true);
+        String direction = matcher.group(1);
+        String rawTerritoryName = matcher.group(2);
+        // Want to account for weird stuff like "the Forgery" and make it "The Forgery" for the sake of our brief
+        // message (looks odd otherwise).
+        String territoryName = rawTerritoryName.substring(0, 1).toUpperCase() + rawTerritoryName.substring(1);
+
+        String enteringMessage = String.format("§7Now §o%s§r %s", direction, territoryName);
```
kristofbolyai: I think we could shorten this with using an arrow emoji, like friend join message does.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
kristofbolyai: Oh! This is a subtitle event. Okay. Then it is fine to leave it as is. For now. Maybe, we should create an universal feature for titles, like InfoMessageFilter feature is for chat. Not scope of this PR though.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Config
+    public boolean redirectTerritoryMessages = true;
+
+    // Handles the subtitle text event.
+    @SubscribeEvent
+    public void onSubtitleSetText(SubtitleSetTextEvent event) {
+        if (!redirectTerritoryMessages) {
+            return;
+        }
```
kristofbolyai:
```suggestion
        if (!redirectTerritoryMessages) return;
        
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
kristofbolyai: Add the \@FeatureInfo annotation though.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
san7890: See PR Body for new example with emoji. I think the italicization still "works", but let me know if you want me to axe it.

kristofbolyai: I would just straight up remove the entering/leaving part. No need for it. Arrows are condense and easy to understand. I would however, make the whole message gray, territory name included. 

As much as you like your screen/chat not the be spammed with random stuff, I like it when notifications are short and simple to understand.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@FeatureInfo(category = FeatureCategory.REDIRECTS)
+public class TerritoryMessageRedirectFeature extends UserFeature {
+    private static final Pattern TERRITORY_MESSAGE_PATTERN = Pattern.compile("§7\\[You are now (\\S+) (.+)\\]");
+
+    @Config
+    public boolean redirectTerritoryMessages = true;
```
kristofbolyai: This config is useless. If the user turns off this feature, they get the same behaviour. Remove it.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -579,6 +579,9 @@
   "feature.wynntils.statusOverlay.overlay.status.textShadow.description": "What should the text shadow look like?",
   "feature.wynntils.statusOverlay.overlay.status.textShadow.name": "Text Shadow",
   "feature.wynntils.territoryDefenseMessage.name": "Territory Defense Messages",
+  "feature.wynntils.territoryMessageRedirect.name": "Territory Message Redirect",
+  "feature.wynntils.territoryMessageRedirect.redirectTerritoryMessages.description": "Should messages about entering or leaving a Territory be redirected to the game update ticker?",
+  "feature.wynntils.territoryMessageRedirect.redirectTerritoryMessages.name": "Redirect Entering/Leaving Territory Messages",
```
kristofbolyai: Don't forget to remove the i18n as well

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+        String territoryName = rawTerritoryName.substring(0, 1).toUpperCase() + rawTerritoryName.substring(1);
+        String direction = rawDirection.substring(0, 1).toUpperCase() + rawDirection.substring(1);
+
+        String enteringMessage = String.format("§7%s §o%s§r %s", directionalArrow, direction, territoryName);
+        NotificationManager.queueMessage(enteringMessage);
+    }
+
+    // Handles the chat log message event, we don't want a duplicate so just cancel the event and rely on the subtitle
+    // text event.
+    @SubscribeEvent(priority = EventPriority.LOW)
```
kristofbolyai: This is fine to be a normal event.
```suggestion
    @SubscribeEvent
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
san7890: Updated, will show up in coming commits.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+    public void onChat(ChatMessageReceivedEvent event) {
+        if (!redirectTerritoryMessages) {
+            return;
+        }
+        String codedString = ComponentUtils.getCoded(event.getMessage());
```
kristofbolyai: No need for this. Use `event.getOriginalCodedMessage()`

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -579,6 +579,9 @@
   "feature.wynntils.statusOverlay.overlay.status.textShadow.description": "What should the text shadow look like?",
   "feature.wynntils.statusOverlay.overlay.status.textShadow.name": "Text Shadow",
   "feature.wynntils.territoryDefenseMessage.name": "Territory Defense Messages",
+  "feature.wynntils.territoryMessageRedirect.name": "Territory Message Redirect",
```
```suggestion
  "feature.wynntils.territoryMessageRedirect.name": "Territory Message Redirects",
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+    public boolean redirectTerritoryMessages = true;
+
+    // Handles the subtitle text event.
+    @SubscribeEvent
+    public void onSubtitleSetText(SubtitleSetTextEvent event) {
+        if (!redirectTerritoryMessages) return;
+
+        String codedString = ComponentUtils.getCoded(event.getComponent());
+        Matcher matcher = TERRITORY_MESSAGE_PATTERN.matcher(codedString);
+        if (matcher.matches()) event.setCanceled(true);
```
kristofbolyai:
```suggestion
        if (matcher.matches()) event.setCanceled(true);
        
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+        String directionalArrow = "";
+        switch (rawDirection) {
+            case "entering":
+                directionalArrow = "→";
+                break;
+            case "leaving":
+                directionalArrow = "←";
+                break;
+        }
+        // Want to account for weird stuff like "the Forgery" and make it "The Forgery" for the sake of our brief
```
kristofbolyai:
```suggestion

        // Want to account for weird stuff like "the Forgery" and make it "The Forgery" for the sake of our brief
```

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,66 @@
...
+    public void onSubtitleSetText(SubtitleSetTextEvent event) {
+        if (!redirectTerritoryMessages) return;
+
+        String codedString = ComponentUtils.getCoded(event.getComponent());
+        Matcher matcher = TERRITORY_MESSAGE_PATTERN.matcher(codedString);
+        if (matcher.matches()) event.setCanceled(true);
+        String rawDirection = matcher.group(1);
+        String rawTerritoryName = matcher.group(2);
+        String directionalArrow = "";
+        switch (rawDirection) {
```
kristofbolyai, line 36:
```suggestion

        switch (rawDirection) {
```

kristofbolyai (commented):
Leaving new lines between different parts of a method can go a long way in making the method look much cleaner!

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
kristofbolyai: Also, didn't you mean to return early on the case this does not match? And cancel the event otherwise?

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
san7890: PR Body Image updated

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -178,6 +179,7 @@ public static void init() {
         registerFeature(new PlayerArmorHidingFeature());
         registerFeature(new PlayerGhostTransparencyFeature());
         registerFeature(new InventoryRedirectFeature());
+        registerFeature(new TerritoryMessageRedirectFeature());
```
kristofbolyai: These 2 are not in an alphabetical order.

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,58 @@
...
+        String territoryName = rawTerritoryName.substring(0, 1).toUpperCase() + rawTerritoryName.substring(1);
+
+        String enteringMessage = String.format("§7%s %s", directionalArrow, territoryName);
+        NotificationManager.queueMessage(enteringMessage);
+    }
+
+    // Handles the chat log message event, we don't want a duplicate so just cancel the event and rely on the subtitle
+    // text event.
+    @SubscribeEvent
+    public void onChat(ChatMessageReceivedEvent event) {
```
kristofbolyai, line 58: The whole method could be just "if( TERRPATTERN.matcher(event.get..()).matches()) cancel event"

common/src/main/java/com/wynntils/features/user/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -0,0 +1,60 @@
...
+            case "leaving":
+                directionalArrow = "←";
+                break;
+            default:
+                return;
+        }
+
+        // Want to account for weird stuff like "the Forgery" and make it "The Forgery"
+        // for the sake of our brief message (looks odd otherwise).
+        String territoryName = rawTerritoryName.substring(0, 1).toUpperCase() + rawTerritoryName.substring(1);
```
magicus: Is there not already a utility function to do this? Either in our classes or like google CaseFormat. If not, consider making this a string util function.

san7890: it was in apache stringutils, whoops

magicus (approved):

kristofbolyai (approved):

kristofbolyai:
Thank you for your contribution !

PR #754, opened 2022-11-29
feat: Render player POI's on minimap

bricefrisco:
This PR introduces a new feature which renders remote friend or party members player heads onto the minimap.

Screenshot 2022-11-28 184339

Similar to the equivalent feature on the main map, this requires that both players are connected to Hades.

Changes:
- Created abstract class `PlayerPoiBase`
- Renamed `PlayerPoi` to `PlayerMainMapPoi` and extended `PlayerPoiBase`
- Created new class `PlayerMiniMapPoi` that also extends `PlayerPoiBase`
- Modified `MinimapFeature` to render `PlayerMiniMapPoi`
- Added configuration for enabling or disabling the feature
- Added configuration for player head scale


common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
```diff
@@ -234,9 +248,19 @@ private void renderPois(
...
 
             float currentZoom = 1f / scale;
-            for (Poi poi : Stream.concat(MapModel.getServicePois().stream(), MapFeature.INSTANCE.customPois.stream())
-                    .toList()) {
 
+            List<Poi> poisToRender = new ArrayList<>(MapModel.getServicePois());
+            poisToRender.addAll(MapFeature.INSTANCE.customPois);
+            List<PlayerMiniMapPoi> playerPois = HadesUserModel.getHadesUserMap().values().stream()
+                    .filter(user -> user.isPartyMember() && renderRemotePartyPlayers
+                            || user.isMutualFriend() && renderRemoteFriendPlayers)
```
kristofbolyai: Aren't you missing 2 pairs of parenthesis here?

common/src/main/java/com/wynntils/wynn/model/map/poi/PlayerPoiBase.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.wynntils.sockets.objects.HadesUser;
+
+public abstract class PlayerPoiBase implements Poi {
+    protected final HadesUser user;
+    protected float playerHeadRenderSize;
+
+    protected PlayerPoiBase(HadesUser user, float playerHeadScale) {
+        this.user = user;
+        this.playerHeadRenderSize = 20 * playerHeadScale;
```
kristofbolyai: Can you make a const for the `20` to not have magic values here (and below)?

common/src/main/java/com/wynntils/wynn/model/map/poi/PlayerPoiBase.java
```diff
@@ -0,0 +1,37 @@
...
+    public boolean hasStaticLocation() {
+        return false;
+    }
+
+    public int getWidth(float mapZoom, float scale) {
+        return (int) (playerHeadRenderSize + 20);
+    }
+
+    public int getHeight(float mapZoom, float scale) {
+        return (int) (playerHeadRenderSize + 17);
```
kristofbolyai: (same thing about constant)

common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
```diff
@@ -234,9 +248,19 @@ private void renderPois(
...
+                    .sorted(Comparator.comparing(
+                            hadesUser -> hadesUser.getMapLocation().getY()))
```
kristofbolyai, line 258: Do we need to sort this? Why are minimap pois not sorted? @Incompleteusern

common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
bricefrisco: I believe AND takes precedence over OR, but to avoid bugs/confusion in future it may be best to add them here.

Incompleteusern: Im pretty sure it is the other way around

common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
bricefrisco: This was just something I copied from the main map feature, and it made sense to leave that sort.

It appears that the sort here matters but I can test that. Unless I missed the logic somewhere I don't see any sort logic elsewhere for these.

kristofbolyai: Oh. We don't even sort any other pois on the main map now. This is fine as is then.

kristofbolyai (approved):
Looks good to me! Thank you for your contribution.

PR #757, opened 2022-12-01
feat: Redirects Potion Replaced/At-Limit Messages as notifications

san7890:
image

Let me know if I need to edit the "brief overlay message" formatting, I just went with what sounded best in my head. I also realized I put the "horse" regexes from an earlier PR in the wrong spot, so I adjusted that in this PR since I had just made that mistake again

Note: "Lesser potion replaced" and "At Potion Charge Limit!" are the only two added in this PR, "+3 Potion Charges" was added in a previous PR.

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
@@ -93,6 +93,11 @@ public class InfoMessageFilterFeature extends UserFeature {
     private static final Pattern HORSE_DESPAWNED_PATTERN =
             Pattern.compile("§dSince you interacted with your inventory, your horse has despawned.");
 
+    private static final Pattern MAX_POTIONS_ALLOWED_PATTERN =
+            Pattern.compile("§4You already are holding the maximum amount of potions allowed.");
+    private static final Pattern LESS_POWERFUL_POTION_REMOVED_PATTERN =
+            Pattern.compile("§7One less powerful potion was replaced to open space for the added one.");
+
     @Config
     private boolean hideWelcome = true;
 
@@ -131,6 +136,9 @@ public class InfoMessageFilterFeature extends UserFeature {
 
     @Config
     private FilterType horse = FilterType.REDIRECT;
+
+    @Config
+    private FilterType potion = FilterType.REDIRECT;
 
     @SubscribeEvent
     public void onInfoMessage(ChatMessageReceivedEvent e) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/user/InfoMessageFilterFeature.java
```diff
@@ -396,6 +404,35 @@ public void onInfoMessage(ChatMessageReceivedEvent e) {
...
+            if (potion != FilterType.KEEP) {
+                Matcher maxPotionsMatcher = MAX_POTIONS_ALLOWED_PATTERN.matcher(msg);
+                if (maxPotionsMatcher.matches()) {
+                    e.setCanceled(true);
+                    if (potion == FilterType.HIDE) {
+                        return;
+                    }
+
+                    NotificationManager.queueMessage(
+                            new TextComponent("At Potion Charge Limit!").withStyle(ChatFormatting.DARK_RED));
```
kristofbolyai: why is the message capitalized

san7890: in my head i thought it would play better with the already existing "+3 Potion Charges" message but this can be changed to be "At potion charge limit!"

kristofbolyai: If you look at the other message it is not consisgent

san7890: i see your point

magicus: @kristofbolyai I think at some point (maybe when doing the i18n?) we need to have a look at all the notification messages as a whole to make sure coloring, capitalization and punctuation is consistent. Right now it feels a bit random if a message ends with a period, an exclamation mark or nothing at all (which I think makes the most sense, for this short phrases). 

Maybe we can classify them as "info", "warning" or "error" and use colors based on that. "No totems nearby" would be an error, "+3 potion charges" and info and "At potion charge limit" a warning. Or something like that.

@san7890 Not saying you need to do any of this in this PR, just thinking out loud.

kristofbolyai: > @kristofbolyai I think at some point (maybe when doing the i18n?) we need to have a look at all the notification messages as a whole to make sure coloring, capitalization and punctuation is consistent. Right now it feels a bit random if a message ends with a period, an exclamation mark or nothing at all (which I think makes the most sense, for this short phrases). 
> 
> 
> 
> Maybe we can classify them as "info", "warning" or "error" and use colors based on that. "No totems nearby" would be an error, "+3 potion charges" and info and "At potion charge limit" a warning. Or something like that.
> 
> 
> 
> @san7890 Not saying you need to do any of this in this PR, just thinking out loud.

Perhaps modify the i18n issue to include this?

kristofbolyai: or make a new..

magicus:
> I also realized I put the "horse" regexes from an earlier PR in the wrong spot,

What makes this the "wrong spot"? I fail to see any consistent ordering. If you want it there I won't complain, I'm just curious what ordering you think there should be.

magicus (approved):

san7890:
> > I also realized I put the "horse" regexes from an earlier PR in the wrong spot,
> 
> 
> 
> What makes this the "wrong spot"? I fail to see any consistent ordering. If you want it there I won't complain, I'm just curious what ordering you think there should be.

It seemed all of the "BACKGROUND" message regexes had their own grouping, while the "NORMAL" message regexes were in their own group as well (mirroring how the if/else chain is set up). I accidentally put the Horse/Potion regexes in the "BACKGROUND" regex section.

magicus:
Ah, I see. It probably was sorted that way, yes. Not that it matters much as of #759...

PR #759, opened 2022-12-01
feat: Improve abstraction of redirected messages

magicus:
I have split up the old `InfoMessageFilterFeature` into two parts -- one with the original "filter" intent of this feature, that just removes certain messages. The other, new feature is `ChatRedirectFeature`. This contains a standardized and more abstracted way of expressing how to match on a regexp and convert that into a notification message.

Ideas for future improvements: Trying to make this more compact by utilizing lambdas instead of classes for each redirector. Making config values automatically created from the redirectors array, instead of the current "hardcoded" hack.


magicus:
Also, `BlacksmithRedirectFeature` and `AbilityRefreshRedirectFeature` should be folded into this class, but I do not know what MessageType they are using (this is not checked in the current code) so I couldn't do it right now. And the `getUncoloredSystemPattern()` should also be removed, but we need to register the proper, color coded strings for that. So I'm leaving that too for a future follow-up.

common/src/main/java/com/wynntils/core/features/FeatureRegistry.java
```diff
@@ -183,6 +184,7 @@ public static void init() {
         registerFeature(new PreventTradesDuelsFeature());
         registerFeature(new QuestInfoOverlayFeature());
         registerFeature(new QuickCastFeature());
+        registerFeature(new ChatRedirectFeature());
```
kristofbolyai: You forgot to maintain alphabetical order

magicus: ..! 🙄 Oh dead. I named it `RedirectFeature` originally, and placed it in the right spot. Then I used IntelliJ renaming and forgot about it...

I also noticed that I had managed to miss pushing my last updates from yesterday, which -- ironically enough -- was about alphabetic order as well. :-)

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -0,0 +1,626 @@
...
+            String unusedAbilityPoints = matcher.group(2);
+
+            return List.of(
+                    UnusedSkillPointsRedirector.getUnusedSkillPointsMessage(unusedSkillPoints),
+                    UnusedAbilityPointsRedirector.getUnusedAbilityPointsMessage(unusedAbilityPoints));
+        }
+
+        @Override
+        String getNotification(Matcher matcher) {
+            // We still need to implement this
```
kristofbolyai: At first sight, this got me really confused. I thought, why did we not just implement this one notification? Then, it hit me. This is an abstract method and you mean that we need to impl this even though it won't be used!

```suggestion
            // We still need to implement this method, but it should not be called
```

kristofbolyai (dismissed):
Looks good!

kristofbolyai:
Also, Welcome back magicus :)

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
magicus: What I really wanted to express in the top-level abstract class that you need to implement *either* getNotification, if you have just a single notification, which is by far the most common case, *or* getNotifications, if you have multiple. 

I could have solved it by having a default implementation for getNotification which just returned null, but that would mean that we would not get the compiler screaming if we do wrong when adding more "normal" redirectors.

Or I could have solved it by always returning a List<String> (basically, only keep getNotifications()) but that would have made all "normal" implementations more complicated.

Hm. I should of course make a interface and a common base class for the simple notifications. 🤦 Sometimes a night's sleep make wonders with your ability to see your code.

magicus:
Replaced by https://github.com/Wynntils/Artemis/pull/765

PR #760, opened 2022-12-02
feat: Add off-screen indicator to WorldWaypointDistanceFeature

HalcyonSuoh:


common/src/main/java/com/wynntils/wynn/model/CompassModel.java
```diff
@@ -57,18 +59,28 @@ public static Optional<WaypointPoi> getCompassWaypoint() {
         return Optional.empty();
     }
 
+    public static Texture getTargetIcon() {
+        return targetIcon;
+    }
+
     public static void setDynamicCompassLocation(Supplier<Location> compassSupplier) {
```
kristofbolyai, line 66: Why? Imo, we should still take a texture as argument. Ofc, you can have a method without texture, which defaults to Texture.WAYPOINT. 

We don't need a texture supplier here as even `DynamicIconPoi` has a static icon texture.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -233,7 +234,12 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
             if (hovered != null && !(hovered instanceof TerritoryPoi)) {
                 McUtils.playSound(SoundEvents.EXPERIENCE_ORB_PICKUP);
                 if (hovered.hasStaticLocation()) {
-                    CompassModel.setCompassLocation(new Location(hovered.getLocation()));
+                    if (hovered instanceof IconPoi) {
```
kristofbolyai:
```suggestion
                    if (hovered instanceof IconPoi iconPoi) {
```

This way, you don't need to cast again.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,170 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
+            centerRelativePosition = new Vec3(minY / m, minY, 0);
+        }
+
+        // bring the position back to normal screen space (top left origin point)
+        return new Vec2(
+                (float) (centerRelativePosition.x + centerPoint.x), (float) (centerRelativePosition.y + centerPoint.y));
+    }
+
+    private boolean isInBound(Vec3 position, Window window) {
+        return 0 < position.x
```
kristofbolyai:
```suggestion
        return position.x > 0
```

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,170 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
+                && position.x < window.getGuiScaledWidth()
+                && 0 < position.y
```
kristofbolyai:
```suggestion
                && position.y > 0
```

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
kristofbolyai: This way, position is always on the left side. This improves readability.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,170 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
-                        VerticalAlignment.Middle,
-                        textShadow);
-    }
+        float displayPositionX;
+        float displayPositionY;
+
+        Vec2 intersectPoint = getBoundingIntersectPoint(screenCoord, event.getWindow());
+        Texture icon = CompassModel.getTargetIcon();
+
+        if (intersectPoint == null) {
```
kristofbolyai, line 118:
```suggestion
       // The set waypoint is visible on the screen, so we render the icon + distance
        if (intersectPoint == null) {
```

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,170 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
+
+        Window window = McUtils.window();
+
+        return new Vec3(
+                (float) ((ndc.x + 1.0f) / 2.0f) * window.getGuiScaledWidth(),
+                (float) ((1.0f - ndc.y) / 2.0f) * window.getGuiScaledHeight(),
+                (float) ndc.z);
+    }
+
+    // tldr: draw a line from screen center to the target's screenspace coordinate
```
kristofbolyai:
```suggestion
    // draw a line from screen center to the target's screenspace coordinate
```

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,170 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
+        // (x2 - x1) is the equivalent of the x position
+        // c is the line's y-intercept, but line pass through origin, so this will be 0
+        // finalize to y = mx, or x = y/m
+
+        if (centerRelativePosition.x > 0) {
+            centerRelativePosition = new Vec3(maxX, maxX * m, 0);
+        } else {
+            centerRelativePosition = new Vec3(minX, minX * m, 0);
+        }
+        if (centerRelativePosition.y > maxY) {
```
kristofbolyai, line 249:
```suggestion

        if (centerRelativePosition.y > maxY) {
```

kristofbolyai (commented):
I have some nit-picks to improve code quality, but this looks great!

kristofbolyai (approved):
Looks good to me! Thank you, this looks really great in-game! 

Also, you mentioned in discord that math could be optimized. It sure could be, but this whole feature does not even show up on CPU/Memory profiles :)

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -82,61 +96,172 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
+        // finalize to y = mx, or x = y/m
+
+        if (centerRelativePosition.x > 0) {
+            centerRelativePosition = new Vec3(maxX, maxX * m, 0);
+        } else {
+            centerRelativePosition = new Vec3(minX, minX * m, 0);
+        }
+
+        if (centerRelativePosition.y > maxY) {
+            centerRelativePosition = new Vec3(maxY / m, maxY, 0);
```
magicus, line 250: Are we 100% sure we can't get a division by zero here? If the angle is 0 then m will be 0. But I must admit I got a bit lost on the maths here.

HalcyonSuoh: > Are we 100% sure we can't get a division by zero here? If the angle is 0 then m will be 0. But I must admit I got a bit lost on the maths here.

centerRelativePosition.y > maxY is the case where the line intersect with bottom bounding, in which the only case where slope is zero is probably where resolution's height being 0, I don't think such edge case exist in any playable situation

HalcyonSuoh: on second thought, situation might be possible if bottom bounding goes higher than screen center, so it think it's intersecting with right and bottom bounding simultaneously, I'll fix it if I can get a crash out of it

PR #511, opened 2022-10-05
WIP: Create /quest command

magicus:

PR #766, opened 2022-12-03
feat: Add /quest command

magicus:
This replaces https://github.com/Wynntils/Artemis/pull/511 but targets 'main' instead of 'alpha'.


magicus:
Yeah, this has been dragging on for far too long. It's not the most important part, but since it was mostly done I felt it was bad just throwing away, and it couldn't sit here rotting in a half-done state.

common/src/main/java/com/wynntils/commands/QuestCommand.java
```diff
@@ -0,0 +1,271 @@
...
+    @Override
+    public LiteralArgumentBuilder<CommandSourceStack> getBaseCommandBuilder() {
+        return Commands.literal("quest")
+                .then(Commands.literal("list")
+                        .executes((ctxt) -> listQuests(ctxt, "distance"))
+                        .then(Commands.argument("sort", StringArgumentType.word())
+                                .suggests(SORT_SUGGESTION_PROVIDER)
+                                .executes((ctxt) -> listQuests(ctxt, ctxt.getArgument("sort", String.class)))))
+                .then(Commands.literal("search")
+                        .then(Commands.argument("text", StringArgumentType.greedyString())
```
kristofbolyai, line 50: "text"? Is this quest name?

kristofbolyai: Oh, I see. Okay, it can stay.

common/src/main/java/com/wynntils/commands/QuestCommand.java
```diff
@@ -0,0 +1,271 @@
...
+                        .then(Commands.argument("quest", StringArgumentType.greedyString())
+                                .suggests(QUEST_SUGGESTION_PROVIDER)
+                                .executes(this::lookupOnWiki)))
+                .executes(this::syntaxError);
+    }
+
+    private int listQuests(CommandContext<CommandSourceStack> context, String sort) {
+        QuestSortOrder order = QuestSortOrder.fromString(sort);
+
+        Managers.Quest.rescanQuestBook(true, false);
```
kristofbolyai, line 71: This method is technically async, and the command doesn't "await" it?

common/src/main/java/com/wynntils/commands/QuestCommand.java
```diff
@@ -0,0 +1,271 @@
...
+
+        response = Component.literal("Matching quests:").withStyle(ChatFormatting.AQUA);
+
+        generateQuestList(quests, response);
+
+        context.getSource().sendSuccess(response, false);
+        return 1;
+    }
+
+    private void generateQuestList(List<QuestInfo> quests, MutableComponent response) {
```
kristofbolyai, line 133: Can you sort methods in some order? I usually do command methods first, then helper methods.

kristofbolyai (approved):
Other than not awaiting rescan, I think this looks good.

common/src/main/java/com/wynntils/commands/QuestCommand.java
magicus: Yes. I made some botched efforts to synchronize on this, and then gave up, and just warned the user to retry the command. 

The "proper" solution will require quite some amount of plumbing, and I could not really motivate it. If you think this is too ugly, I'd rather drop the entire /quest command. I'm not sure how useful it will be in practice. I originally started this PR to have a way to exercise the quest model, and I did not want to make a GUI.

kristofbolyai: Okay, this is fine.

magicus: Thanks.

magicus:
I have been thinking of ways to improve quest handling that would spill over into this command as well. For instance, to save a copy of the quest book analysis per character, so that when we load the character, we already have the quest book available. Of course it will need to be refreshed, to be on the safe side we don't miss anything, but in 99/100 cases it will be correct.

I am also trying to split the "QuestInfo" into two parts, one with immutable quest metadata (name, length, etc) and one with "current" information (is tracked, next task, started/can start/finished, etc). It's always a bit of a shame when you mix mutable and immutable information like in the QuestInfo. I realize it probably is my fault, but ... yeah, I was young and stupid then. Now I'm old and know better. :)

kristofbolyai:
> I have been thinking of ways to improve quest handling that would spill over into this command as well. For instance, to save a copy of the quest book analysis per character, so that when we load the character, we already have the quest book available. Of course it will need to be refreshed, to be on the safe side we don't miss anything, but in 99/100 cases it will be correct.
> 
> 
> 
> I am also trying to split the "QuestInfo" into two parts, one with immutable quest metadata (name, length, etc) and one with "current" information (is tracked, next task, started/can start/finished, etc). It's always a bit of a shame when you mix mutable and immutable information like in the QuestInfo. I realize it probably is my fault, but ... yeah, I was young and stupid then. Now I'm old and know better. :)

Yeah, we just need to fix class ids to be persistent. QuestInfo is a bit mixed, yes.

PR #767, opened 2022-12-03
feat: Handle more types of NPC dialogues

### LINKED ISSUE #334: NPC dialogue overlay should also catch "non-screen" dialogue

### LINKED ISSUE #702: ChatModel with disabled dialogue extraction throws away NPC dialouges
This bug is an interaction between **ChatTabs** and **ChatModel**. 

The reason for this bug is that only messages from `ChatMessageReceivedEvent` are injected into chat tabs, messages from `ChatPacketReceivedEvent` are not. The reason for is that I thought we handle every case with either `ChatMessageReceivedEvent` or a `NpcDialogEvent` event in `ChatModel`. We do not, but I really think we should. In the future I would prefer to NOT use `ChatPacketReceivedEvent` anywhere except `ChatModel` (this is basically the case now, the event is only used in `ChatTabFeature` to be cancelled and `ActionBarModel` which needs to use it because action bar data is sent with the packet), process the message and post an appropriate event EVERY case.

What we really need to do is either make `ChatMessageReceivedEvent` post on NPC dialogue, if it is not handled with dialogue extraction. Alternatively, we can introduce a new event, although I don't think I would really like it.

magicus:
These are the ones where the NPC message just goes to the chat as normal, not requiring you to press sneak.

There are actually two kinds of those, "disappearing" and normal chat messages. The "disappearing" ones are just like normal confirmation NPC dialogues, but they do not require sneak. I have only seen those in King's Recruit. These are handled by Wynncraft by sending out "screens", and when the chat is done, the screen is "redrawn" without it.

The second one is just like normal chat messages.

There are a slew of problems with getting these to work as you'd expect. This is just a draft.

Needs fixing in the framework:
- [ ] Disappearing messages should have a NULL sent when they are cleared. 
- [ ] This means we need an enum to separate between "chat style", "need confirmation", "need no confirmation", not just a boolean
- [ ] There is (at least one) bug when different kinds of messages are mixed, so old "chat style" messages pop up in the overlay when new "need no confirmation" messages gets removed, or something like that.

Needs fixing in the Overlay:
- [ ] "chat style" messages needs a time out. Can probably reuse code from @kristofbolyai's old closed PR.
- [ ] "chat style" messages can stack. In King's Recruit you can get several differnt ones at the same time. 
- [ ] This means NPC overlay needs to handle more than one chat line
- [ ] They should probably be displayed like in the chat, one on top of the other, and each one of them having individual timeouts so they disappear one after the other, so timeouts need to be on a per-line basis, not for the entire overlay
- [ ] How handle if "chat style" messages has not yet timed out, when a "need confirmation" message arrive?


magicus:
This PR replaces https://github.com/Wynntils/Artemis/pull/435, but targets `main` instead.

magicus:
This will solve #334.

This also adds support for "selection style" NPC dialogs. 

I have verified that this covers all kinds of different NPC dialogue styles that appear in King's Recruit.

common/src/main/java/com/wynntils/handlers/chat/RecipientType.java
```diff
@@ -9,6 +9,7 @@
 public enum RecipientType {
     INFO(null, null, "Info"),
     CLIENTSIDE(null, null, "Clientside"),
+    NPC("(?:§r)?§7\\[\\d+\\/\\d+\\] §r§2.+: §r§..*", "(?:§r)?§8\\[\\d+\\/\\d+\\] .+: §r§..*", "NPC"),
```
magicus, line 12: @kristofbolyai I just realized... how does this mix with chat tabs?

kristofbolyai: I am not sure what the question is. All tab works fine, other tabs won't have this selected. But we need to add a GUI checkmark..

kristofbolyai:
One question, does this work with NPC dialogue overlay disabled, but chat model enabled? (old bug, NPC dialogue didn't use to show up)

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -75,13 +80,32 @@ private void cancelAutoProgress() {
...
+            if (dialogueType == NpcDialogueType.CONFIRMATIONLESS) {
+                // If it was confirmationless, delay it's removal
+                removeTime = System.currentTimeMillis() + calculateMessageReadTime(msg);
+            }
+        } else {
+            currentDialogue = msg;
+            dialogueType = e.getType();
+            if (dialogueType == NpcDialogueType.CONFIRMATIONLESS) {
+                removeTime = System.currentTimeMillis() + calculateMessageReadTime(msg);
+            }
```
kristofbolyai: This part seems to be common? Extract it?

kristofbolyai (dismissed):
Looks good. I will do the GUI work in a few days.

common/src/main/java/com/wynntils/handlers/chat/RecipientType.java
magicus: The question was: "It is waaay point bedtime and I don't fully understand what I have written". :) But more seriously. Even though this is a separate RecipientType, it will not make sense to have a separate tab for this. There are just a handful of those in the game at all (and I think like half of them are in King's Recruit...) So in the chat tabs, these should be treated as INFO messages (which is where they would have ended up before this PR). 

My qustion is: do I need to make some changes in ChatTabs to get messages tagged as NPC recipient type to be treated as INFO messages? 

I don't think it makes any sense to expose this category to the user.

kristofbolyai: Are not the NPC messages canceled? I am not sure what special handling info messages have. But we need to add this so people without the overlay can see npc dialogue in chat?

common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
magicus: Yeah, about that. Code quality. It shucks. Big time. Not up to my usual standard. :-( I was hoping to have a Chat Manager update that would leave it shining and sparkling bright and easy to understand. Instead there were just more special handling, convoluted code and strange edge cases. :-(

I have pretty high confidence in the code, since I've tested it thoroughly, and also played a real bit of the "All Roads to Peace" quest, where there are a lot of odd dialogs with the Orcs.

But it looks like shite. 

I would like to improve this, but I don't know how, and I'm worried that I might break things if I poke around too much.

But I agree that the duplication sticks in your eyes. I'll see if I can do something about that.

magicus: Oh, I did not see where you commented! :) I missed this part, fixed now. I thought you were talking about the massive chunk of duplication in the ChatHandler after `  } else if (NPC_SELECT_PATTERN.matcher(firstLineCoded).find()) {`... I am not proud of that. I'll look if I can make something about it now that I'm not half falling asleep.

common/src/main/java/com/wynntils/handlers/chat/RecipientType.java
magicus: Riiiight.. I'm not sure what I'm thinking. Let me check the code again. And try without the overlay.

magicus:
NPC special handling will only happen if shouldSeparateNPC() is true. This will (in practice) only happen if Npc Dialogue Overlay is active. All look fine!

magicus:
But note that chat tabs will not work as expected during NPC dialogs ("chat screens"), if NPC overlay is disabled. This does not change with this PR. We basically have a hopeless situation in that case. The chat history is reprinted several times per second, but since we can't display the NPC in any better way, we can't try to counteract this screen behavior.

kristofbolyai:
> NPC special handling will only happen if shouldSeparateNPC() is true. This will (in practice) only happen if Npc Dialogue Overlay is active. All look fine!

Nice, thank you! And sure, we can't do much.

kristofbolyai (approved):

kristofbolyai:
Fixes https://github.com/Wynntils/Artemis/issues/334, https://github.com/Wynntils/Artemis/issues/702

PR #774, opened 2022-12-03
fix: No config value should be saved to disk unless edited

kristofbolyai:
In a perfect world, our "default", freshly generated config file should be empty. 

During my development of chat tabs, I've noticed this is not the case. I've left the problem back then, noted, but unfixed. Now, I've came back to it, in preparation of my upcoming chat tab changes... 

Only to realize the problem was fixed by me, accidentally. Turns out, I only needed to implement hashCode/equals for `ChatTab`, which I needed before. I have went ahead and tested what needs to be fixed. Turned out, we had only one outlier, `CommandAliasesFeature`. By implementing equals/hashCode, the problem is fixed, and our freshly generated config file is empty!

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/CommandAliasesFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/CommandAliasesFeature.java
@@ -81,5 +81,19 @@ public class CommandAliasesFeature extends UserFeature {
         public String getOriginalCommand() {
             return originalCommand;
         }
+
+        @Override
+        public boolean equals(Object other) {
+            if (this == other) return true;
+            if (other == null || getClass() != other.getClass()) return false;
+
+            CommandAlias that = (CommandAlias) other;
+            return originalCommand.equals(that.originalCommand) && aliases.equals(that.aliases);
+        }
+
+        @Override
+        public int hashCode() {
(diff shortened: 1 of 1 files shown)


Incompleteusern:
How does adding `equal/hashCode` fix the issue?

kristofbolyai:
> How does adding `equal/hashCode` fix the issue?

image

Incompleteusern (approved):

magicus:
So what is the requirement here? That all classes used in `@Config` annotations are "hash map safe", i.e has equals and hashcode?

Is it possible to tell from introspection in the Config class that the classes support this rule? (I don't know if it is possible to check if a class actually reimplements a method, only that it supports calling it, and that of course everything descendant from Object can...)

kristofbolyai:
> So what is the requirement here? That all classes used in `@Config` annotations are "hash map safe", i.e has equals and hashcode?
> 
> 
> 
> Is it possible to tell from introspection in the Config class that the classes support this rule? (I don't know if it is possible to check if a class actually reimplements a method, only that it supports calling it, and that of course everything descendant from Object can...)

I am not sure we can test for this. Basically the goal is to make every config object comperable.

PR #776, opened 2022-12-04
fix: Dungeon keys overlay not working properly in TM (#501)

### LINKED ISSUE #501: Dungeon Keys Purchase not showing accurately in TM
Similar to #324 ,

image

quartzexpressDEV:
small regex edit to fix #501 
image

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
+++ b/common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
@@ -22,7 +22,9 @@ import net.minecraft.network.chat.Component;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 
-/** Tests if an item is a certain wynncraft item */
+/**
+ * Tests if an item is a certain wynncraft item
+ */
 public final class WynnItemMatchers {
     private static final Pattern SERVER_ITEM_PATTERN = Pattern.compile("§[baec]§lWorld (\\d+)(§3 \\(Recommended\\))?");
     private static final Pattern CONSUMABLE_PATTERN = Pattern.compile("(.+)\\[([0-9]+)/([0-9]+)]");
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
```diff
@@ -40,7 +42,8 @@ public final class WynnItemMatchers {
             Pattern.compile(".*?§([2ebcf])([✤✦❉✹❋]) (Strength|Dexterity|Intelligence|Defence|Agility).*?");
     private static final Pattern TELEPORT_SCROLL_PATTERN = Pattern.compile(".*§b(.*) Teleport Scroll");
     private static final Pattern TELEPORT_LOCATION_PATTERN = Pattern.compile("- Teleports to: (.*)");
-    private static final Pattern DUNGEON_KEY_PATTERN = Pattern.compile("(?:§.)*(?:Broken )?(?:Corrupted )?(.+) Key");
+    private static final Pattern DUNGEON_KEY_PATTERN =
+            Pattern.compile("(?:§.)*(?:\\d*x )*(?:§.)*(?:Broken )?(?:Corrupted )?(.+) Key");
```
kristofbolyai, line 46: This doesn't fix the whole issue. In fact this is a really patchy fix that only fixes a single regex, other regexes are broken.

quartzexpressDEV: oh right the other items have that issue too, I'll fix those too

kristofbolyai: No, don't do that. We need to handle them internally, somewhere else. @magicus could share his opinion on this

quartzexpressDEV:
is there any reasoning for the `.*` in the beginning of the teleport scroll regex? It seems to just breaks stuff in the TM

kristofbolyai:
> is there any reasoning for the `.*` in the beginning of the teleport scroll regex? It seems to just breaks stuff in the TM

I am not sure.

common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
magicus: @kristofbolyai Not entirely sure what you mean. Like when in the trade market, we should prepend a number prefix to the regexp?

kristofbolyai: Maybe. I am not sure how to deal with this.

magicus: Maybe we need like a "InventoryTypeItemNameRewriter" which, when in the TradeMarket, removes the count ... eh... no, sounds bad as well. I don't know either. 

Does it affect other types of items apart from dungeon keys? Otherwise, if this modified regexp works, maybe it's an acceptable solution? And if e.g. teleportation scrolls are affected as well, maybe just update those regexps as well.

Otherwise, I don't really know either.

kristofbolyai:
Okay. First off all, let me thank you for your contribution. It is very great that you went ahead, tried to fix this problem, and submitted a PR for it. However, this PR is not the way to fix that issue. I think, we will need to handle items in that specific TM slot differently. And changing all regexes to work with that single slot is not worth it, and is not the way to go. If you find a solution to fixing a problem, systematically, feel free to submit a new PR.

Again, thank you for your time and efforts, and I hope you will be back with a new PR shortly :)

common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
kristofbolyai: This affects every item in that one specific slot. Some regexes work(ish), some don't match at all.

magicus:
> is there any reasoning for the `.*` in the beginning of the teleport scroll regex?

In general, when I write regexps for matching stuff from Wynncraft, I make sure to include a `^` at the start and a `$` at the end. This makes sure we match the entire string, and makes it clear to anyone reading the code how the pattern was intended to match. If it is possible that additional stuff (like numbers) can appear in front, then it should be "removed" using as tight a pattern as possible, e.g. `\d+` for digits.

PR #777, opened 2022-12-04
feat: Add auto commands to chat tabs + switch with "shift+tab" key

kristofbolyai:


kristofbolyai:
When we release this feature, we need to communicate to people that auto commands will only work if they reset their chat tabs settings, if they modified it (or used an old version with a bugged config saving). Alternatively, they can set auto commands themselves, in config.json.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -29,19 +30,20 @@
...
-            new ChatTab("Party", false, Sets.newHashSet(RecipientType.PARTY), null),
-            new ChatTab("Private", false, Sets.newHashSet(RecipientType.PRIVATE), null),
-            new ChatTab("Shout", false, Sets.newHashSet(RecipientType.SHOUT), null));
+            new ChatTab("All", false, null, null, null),
+            new ChatTab("Global", false, null, Sets.newHashSet(RecipientType.GLOBAL), null),
+            new ChatTab("Local", false, null, Sets.newHashSet(RecipientType.LOCAL), null),
+            new ChatTab("Guild", false, "/g ", Sets.newHashSet(RecipientType.GUILD), null),
+            new ChatTab("Party", false, "/p ", Sets.newHashSet(RecipientType.PARTY), null),
+            new ChatTab("Private", false, "/msg ", Sets.newHashSet(RecipientType.PRIVATE), null),
+            new ChatTab("Shout", false, null, Sets.newHashSet(RecipientType.SHOUT), null));
```
magicus, line 47: I've never done a shout, but is there not a command for that?

Incompleteusern: there is not a command for shouts

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -41,6 +43,14 @@ public static void onWorldStateChange(WorldStateEvent event) {
         }
     }
 
+    @SubscribeEvent
+    public static void onScreenOpened(ScreenOpenedEvent event) {
+        if (!(event.getScreen() instanceof ChatScreen chatScreen)) return;
+        if (focusedTab == null || focusedTab.getAutoCommand() == null) return;
+
+        chatScreen.insertText(focusedTab.getAutoCommand(), true);
```
magicus, line 51: Does this mean that if you switch between the tabs, the different auto commands will pop up in the chat message bar, and then be replaced with a new one when you switch to a new tab, and be removed if you select "local" etc? What if you have started to type a message and switch tab?

kristofbolyai: Then the message will get replaced. I am not fully sure what legacy does, but imo it is fine this way.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -133,6 +135,17 @@ public void onScreenRender(ScreenRenderEvent event) {
         chatScreen.commandSuggestions.render(event.getPoseStack(), event.getMouseX(), event.getMouseY());
     }
 
+    @SubscribeEvent
+    public void onChatScreenKeyTyped(ChatScreenKeyTypedEvent event) {
+        // We can't use keybinds here to not conflict with TAB key's other behaviours.
+        if (event.getKeyCode() != GLFW.GLFW_KEY_TAB) return;
```
magicus, line 142: Will this not break autocompletion..?

kristofbolyai: 🤦 It will. What keybind should we use then?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
magicus: Eww... I dunno, really. :( Just removing what a player has typed is Not Okay, I think. That's the kind of user-generated data destruction behavior you can kind of kill your computer for.

Can't you check if anything has been typed, apart from another auto command, and if so, keep it?

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Can we use ctrl-tab? Or that will interfere with window-changing in Windows, probably... If that still is ctrl-tab, and has not changed to like windows-tab or anything.

PgUp/PgDn, perhaps? Or ctrl-pgup/pgdn, or ctrl- left/right arrow? I'm not sure if any of those are used in vanilla to edit the chat message line or scroll the chat history.

kristofbolyai: Shift tab should work on windows.. I guess

magicus: An alternative might be to allow switching of tabs using the tab key, as long as the message bar is empty. If it is, the tab completetion is useless anyway (just starts inserting names of players logged in on the server), and we can use it for tab switching. If you have started typing, we can let vanilla handle the tab character for autocompletion.

The drawback is of course that you can only change tab with the mouse once you've started typing.

kristofbolyai: Set to shift-tab.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: Fixed now.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Ooookay. I was just about to suggest that "whatever you chose, perhaps adding shift should rotate in the opposite direction".

What about alt-tab? Or wait, maybe that was the window changing thing on Windows?

kristofbolyai: alt tab is a windows thing too

magicus: Sorry, I've been using a mac for so long I've forgotten. It is alt-tab that is used by Windows.

ctrl-tab is used by Firefox on mac, windows and linux to switch tab (and shift-ctrl-tab to switch backwards). It is also used in vscode. I think that's good enough of a "universal standard" for me -- i recommend we use it as well.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Ok. Out of curiousity, how to you do shouts? From a web page?

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
kristofbolyai: Do not worry I also only use linux and max daily. But, I remember windows keybinds. For now.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
kristofbolyai: /use -> bombs -> shout bomb (buy it from store)

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -29,19 +31,20 @@
...
-            new ChatTab("Guild", false, Sets.newHashSet(RecipientType.GUILD), null),
-            new ChatTab("Party", false, Sets.newHashSet(RecipientType.PARTY), null),
-            new ChatTab("Private", false, Sets.newHashSet(RecipientType.PRIVATE), null),
-            new ChatTab("Shout", false, Sets.newHashSet(RecipientType.SHOUT), null));
+            new ChatTab("All", false, null, null, null),
+            new ChatTab("Global", false, null, Sets.newHashSet(RecipientType.GLOBAL), null),
+            new ChatTab("Local", false, null, Sets.newHashSet(RecipientType.LOCAL), null),
+            new ChatTab("Guild", false, "/g ", Sets.newHashSet(RecipientType.GUILD), null),
+            new ChatTab("Party", false, "/p ", Sets.newHashSet(RecipientType.PARTY), null),
+            new ChatTab("Private", false, "/msg ", Sets.newHashSet(RecipientType.PRIVATE), null),
```
magicus: Is there not a `/r` (for "reply") that let's you answer the last private message? Perhaps that would be a better auto message for "Private"?

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -74,6 +84,13 @@ public static void setFocusedTab(ChatTab focused) {
             chatTabData.putIfAbsent(focusedTab, new ChatComponent(McUtils.mc()));
             unreadMessages.put(focusedTab, false);
             McUtils.mc().gui.chat = chatTabData.get(focusedTab);
+
+            // If chat screen is open, and current message is empty, set our auto command
+            if (McUtils.mc().screen instanceof ChatScreen chatScreen
+                    && chatScreen.input.getValue().isEmpty()) {
+                String autoCommand = focusedTab.getAutoCommand() == null ? "" : focusedTab.getAutoCommand();
```
magicus, line 94: Stop me if you think I'm too demanding here, but I think this is still not really ideal. Let's say you open the chat on a tab with no auto command. Then you tab your way to "Party" and get an autocommand of `/p`. Now you tab back to  "Local", but you will still keep the `/p`. And instead, if you were to continue switching tabs to "Guild", the `/p` would not be switched to a `/g`.

magicus: Ok, now I realized what I want you to do. :-) Whenever you set an autocommand, you should also store it as a `lastAutocommand`, and when you get here, you should check if the input field is equal to the `lastAutocommand`. If it is, you should replace it with the new tabs autocommand. But if it is not equal to that (or empty), then you should leave it be.

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
kristofbolyai: I can see both msg and r being useful

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: > Stop me if you think I'm too demanding here, but I think this is still not really ideal. Let's say you open the chat on a tab with no auto command. Then you tab your way to "Party" and get an autocommand of `/p`. Now you tab back to  "Local", but you will still keep the `/p`. And instead, if you were to continue switching tabs to "Guild", the `/p` would not be switched to a `/g`.

This didn't use to happen until i made this change, and I forgot to fix it again😅

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: Yes, but we can reasonably only select one. I would argue for `/r`, since it is a reasonable default, and it is easier to backspace a single `r` if you want to change it to `msg`. If you default to `/msg`, you'll either have to delete `msg` to type `r`, or type the name of the person you want to answer.

kristofbolyai: Changed it.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: https://github.com/Wynntils/Artemis/pull/777/commits/8c50784ed2fd4d68dfa63aa750fd9ecd214e5352 even more elegant

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -74,6 +86,14 @@ public static void setFocusedTab(ChatTab focused) {
             chatTabData.putIfAbsent(focusedTab, new ChatComponent(McUtils.mc()));
             unreadMessages.put(focusedTab, false);
             McUtils.mc().gui.chat = chatTabData.get(focusedTab);
+
+            // If chat screen is open, and current message is empty, set our auto command
```
magicus: I like your solution. But this comment is slightly misleading now. Fix it and then you can just push it.

magicus (dismissed):

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
magicus:
```suggestion
            // If chat screen is open, and current message is empty or the previous auto command, set our auto command
```

magicus (approved):

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
magicus: I fixed it for ya :)

common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
magicus: I realize I was a bit too quick with pushing this for you. I did not check this carefully enough. (Too tired, I should have gone to bed long time ago...)

I think we should change the "shift-tab" to "ctrl+tab", and also let "shift+ctrl+tab" circle backwards through the tabs. Can you fix it in a separate PR? (Sorry for messing up)

kristofbolyai: I'll later

PR #778, opened 2022-12-04
fix: Fix miss-use of CallbackInfoReturnable cancelling

kristofbolyai:
I am not sure why this mixin worked in the first place. How can we cancel a method with a return value, and not actually set a value for it. Probably, the mixin has a default return value. I am not sure. It is fixed now, this was the only instance we've done this.


magicus (dismissed):
This fix looks sane, but there are more suspicious usages of CallbackInfoReturnable... In AbstractContainerScreenMixin, we change the return value but do not actually cancel when the event is cancelled...

magicus:
Also in MultiPlayerGameModeMixin ...

kristofbolyai:
Fixed it now!

magicus (approved):

magicus:
Thanks!

PR #779, opened 2022-12-04
WIP: New web access framework

magicus:
This is very much WIP. I wanted to show what I'm doing, but it's far from where I want this to go. I'll let you know when I need input on the design.


common/src/main/java/com/wynntils/core/net/ApiResponse.java
```diff
@@ -0,0 +1,84 @@
...
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+public class ApiResponse {
+    static Map<HttpRequest, CompletableFuture<HttpResponse<InputStream>>> downloadFutures = new HashMap<>();
+    static Map<HttpRequest, CompletableFuture<Void>> processFutures = new HashMap<>();
```
Incompleteusern: what is this?

common/src/main/java/com/wynntils/core/net/Download.java
```diff
@@ -0,0 +1,120 @@
...
+        // in case of failure:
+        //        FileUtils.deleteQuietly(cacheFile);
+        downloadFutures.put(request, future);
+        return future;
+    }
+
+    protected CompletableFuture<InputStream> cacheInputStreamAsync() {
+        if (request == null) {
+            try {
+                InputStream inputStream = new FileInputStream(localFile);
```
Incompleteusern: Qodana might be right, use try with resource?

common/src/main/java/com/wynntils/core/net/athena/WynntilsAccountManager.java
```diff
@@ -86,55 +82,42 @@ public void dumpEncodedConfig(String name) {
...
+                WynntilsMod.info("Successfully connected to Athena!");
+            });
+        });
+        try {
+            // FIXME: Wait up to 5 seconds for login to complete
+            // This is not ideal, but the rest of the code assumes that login is done
+            // when further down the manager init chain.
+            future.get(5000, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            // if timeout is reached, return false
```
Incompleteusern, line 121: Log failure?

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -79,9 +82,14 @@ private int reload(CommandContext<CommandSourceStack> context) {
             feature.disable();
         }
 
-        WebManager.reset();
+        // FIXME: This probably does not do what it was intended to do.
+        // reset
         ItemProfilesManager.reset();
-        WebManager.init(); // reloads api urls as well as web manager
+        // reloads api urls as well as web manager
```
magicus, line 88: I intend to leave this more or less as-is, for this PR. I have opened https://github.com/Wynntils/Artemis/issues/824 to track how to really handle this.

common/src/main/java/com/wynntils/core/net/Download.java
magicus: How does that mix with future lambdas? 🤔

common/src/main/java/com/wynntils/core/net/athena/WynntilsAccountManager.java
magicus: If `loggedIn = true;` is not done at the time we return from this function, we will log an error so it's handled higher above.

I think this mechanism might need some work, but I opted to make minimal changes as part of this patch.

PR #780, opened 2022-12-05
fix: Place a limit on bounding distance, use camera y level for inconvenient waypoint location

HalcyonSuoh:


kristofbolyai:
I want to personally review this, sometime later. Just want to make sure how this goes with my planned patch.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -78,10 +79,19 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
-        float dz = (float) (location.z + 0.5 - cameraPos.z);
+        if (location.y <= 0 || location.y > 255) {
+            dx = (float) (location.x + 0.5 - cameraPos.x);
+            dy = (float) (cameraPos.y + 0.5 - cameraPos.y);
+            dz = (float) (location.z + 0.5 - cameraPos.z);
+        } else {
+            dx = (float) (location.x + 0.5 - cameraPos.x);
+            dy = (float) (location.y + 0.5 - cameraPos.y);
+            dz = (float) (location.z + 0.5 - cameraPos.z);
+        }
```
kristofbolyai: You can change this code so you only calculate dy in the if statements, and the other at declaration.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -264,4 +275,24 @@ private boolean isInBound(Vec3 position, Window window) {
                 && position.y < window.getGuiScaledHeight()
                 && position.z < 1;
     }
+
+    // limit the bounding distance to prevent divided by zero in getBoundingIntersectPoint
+    @Override
+    protected void onConfigUpdate(ConfigHolder configHolder) {
+        if (!configHolder.getFieldName().endsWith("BoundingDistance")) return;
```
kristofbolyai: I dont think there is a need for this if statement.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -83,6 +84,10 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
         float dy = (float) (location.y + 0.5 - cameraPos.y);
         float dz = (float) (location.z + 0.5 - cameraPos.z);
 
+        if (location.y <= 0 || location.y > 255) {
+            dy = (float) (cameraPos.y + 0.5 - cameraPos.y);
```
kristofbolyai: Wait, I do not get this line. Doesn't the `cameraPos.y`s cancel out each other?

HalcyonSuoh: you would be right, since delta y will be 0 if camera y is taken when calculating position-camera delta, guess my brain stop functioning when I wrote it

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -83,6 +84,10 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
         float dy = (float) (location.y + 0.5 - cameraPos.y);
         float dz = (float) (location.z + 0.5 - cameraPos.z);
 
+        if (location.y <= 0 || location.y > 255) {
+            dy = 0;
```
kristofbolyai, line 88: This is a bit weird IMO. What if we clamped the value between 0 and 255? So basically, if the value was -10, we get 0, but if it was 300, we get 255? Isn't that better? (Use `MathUtils#clamp` if you do so)

HalcyonSuoh: clamping the value will still result the indicator render in mostly inaccessible locations, the aim of it is to reduce the instance where indicator show up at confusing location, like underground or way high up in the sky, for player placed waypoint and LablePoi respectively. 
I might still have a misunderstanding about what value is referring to, but clamping the delta doesn't quite make sense, clamping the location.y will still result in situation described above

kristofbolyai: But, y level zero means we still render underground?

kristofbolyai: Or does 0 here mean we render at camera level?

HalcyonSuoh: 0 is the delta y, which is the difference between the location's y level and camera's y level, and then used to convert the worldspace location to it's screenspace representative

HalcyonSuoh: and that's the reason in first iteration I took the obvious approach and replace location.y with cameraPos.y without realizing

HalcyonSuoh: > Or does 0 here mean we render at camera level?

so this is correct

kristofbolyai: I see. Okay.

kristofbolyai (approved):
Looks good! Thank you!

kristofbolyai:
@HalcyonSuoh you opened a PR to alpha branch. We now use main. Please remember this next time.

PR #781, opened 2022-12-05
feat: Multiple Duplicate Messages in ChatRedirectFeature now get Squashed

san7890:
Apologies if the code is ugly/unclear/can be improved somehow, I'm quite new to how maps work in Java. I wanted it to still remain generic enough to accommodate any type of message.

Are you sick of having stuff like this clog up your notification screen?

image

I know I am, so I decided to spend a few hours and crack this out:

image

Much more flat, much less clutter, much more nicer.


magicus:
First of all, thank you for your contribution! This sounds like a good functionality to have.

Secondly, this belongs in NotificationManager, so it is applied to all notifications, not just from chat. 

Is the code really correct? It looks like if you hit a cached message but that has expired, you will not get a message at all, when instead you should have a message with count 1. Maybe I'm misreading this.

I'm also slightly worried that the cache never gets pruned. You will only remove items just when you find a new identical message but the cached one is too old. I think that instead, each time you update the cache, expired items should be removed. I know we used to have a general datastructure for this kind of "timed expiry cache" in Legacy, I don't remember if it was ported to Artemis.

Finally, maybe the time kept in cache should not correspond to a wall clock time, but due to if the message is visible in the notification overlay? If so, moving the logic to NotificationManager will be a necessary first step, but it might also require some tighter coupling with the NotificationOverlay, so the manager know what is visible.

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -86,6 +91,16 @@ private void register(Redirector redirector) {
         redirectors.add(redirector);
     }
 
+    // Data structure is String Message stores Entry of MessageContainer which stores Integers for the amount of times
+    // the message has been sent and the associated time (Long) in milliseconds
+    private Map<String, Map.Entry<MessageContainer, Map.Entry<Integer, Long>>> cachedNotifications = new HashMap<>();
```
kristofbolyai: Cant you use a Set for this? If you can, use TimesSet, so we don't fill the memory with cache, as @magicus pointed out

kristofbolyai: Perhaps store the repetition count in MessageContainer. That way, you don't need a map here at all.

san7890:
Alright, I have moved all the functionality to NotificationManager and added tracking the "iterations" that the message had appeared in MessageContainer. Cache being cleared is automatically handled by using `TimedSet`, and Time-To-Live is Five Seconds. The formatting of the message has also been updated, which is reflected in the (bottom) image in the PR body.

> Finally, maybe the time kept in cache should not correspond to a wall clock time, but due to if the message is visible in the notification overlay?

Unfortunately, I am completely unaware on the backend of the UI to track if the message is visible or not. I also do not know  how this could be accommodated in the current framework where TimedSets are used to keep freeing up cache. If either of you know a good way to do this, I would greatly appreciate being walked through it since I'm not as familiar with it.

kristofbolyai:
> Alright, I have moved all the functionality to NotificationManager and added tracking the "iterations" that the message had appeared in MessageContainer. Cache being cleared is automatically handled by using `TimedSet`, and Time-To-Live is Five Seconds. The formatting of the message has also been updated, which is reflected in the (bottom) image in the PR body.
> 
> 
> 
> > Finally, maybe the time kept in cache should not correspond to a wall clock time, but due to if the message is visible in the notification overlay?
> 
> 
> 
> Unfortunately, I am completely unaware on the backend of the UI to track if the message is visible or not. I also do not know  how this could be accommodated in the current framework where TimedSets are used to keep freeing up cache. If either of you know a good way to do this, I would greatly appreciate being walked through it since I'm not as familiar with it.

Set the cache time to 10 secs, it is the message render time. Otherwise it should be fine

san7890:
@kristofbolyai done

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,21 +9,36 @@
...
         return message;
     }
 
-    public void editMessage(String newMessage) {
-        this.message.setText(newMessage);
+    public Integer getIterations() {
+        return iterations;
+    }
+
+    public void editMessage(String newMessage, Boolean incrementIterations) {
```
kristofbolyai: Why not boolean? Why use the object?

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,21 +9,36 @@
...
+        String sendableMessage;
+        if (incrementIterations) {
+            this.iterations++;
+            sendableMessage = newMessage + " §7[x" + this.iterations + "]";
```
kristofbolyai: I dont like this. In this method, you edit the message. 

Create a new method, "incrementDuplicate", increment the counter there. Then make method to get the formatted string, appended with "[x2]".

kristofbolyai: Atm, for example, `update` doesn't even work with this code

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,6 +43,19 @@ public static MessageContainer queueMessage(TextRenderTask message) {
 
         WynntilsMod.info("Message Queued: " + message);
         MessageContainer msgContainer = new MessageContainer(message);
+        String messageText = message.getText();
+        Integer messageTextHash = messageText.hashCode();
+
+        for (Pair<String, MessageContainer> cachedMessagePair : cachedMessageSet) {
+            Integer iteratedMessageHash = cachedMessagePair.a().hashCode();
+            if (messageTextHash.equals(iteratedMessageHash)) {
+                WynntilsMod.info("Matched Message: " + message + " to existing message. Updating existing message.");
```
kristofbolyai: debug log

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,6 +43,19 @@ public static MessageContainer queueMessage(TextRenderTask message) {
 
         WynntilsMod.info("Message Queued: " + message);
         MessageContainer msgContainer = new MessageContainer(message);
+        String messageText = message.getText();
+        Integer messageTextHash = messageText.hashCode();
+
+        for (Pair<String, MessageContainer> cachedMessagePair : cachedMessageSet) {
+            Integer iteratedMessageHash = cachedMessagePair.a().hashCode();
+            if (messageTextHash.equals(iteratedMessageHash)) {
```
kristofbolyai: Why do we compare the hashes? Why not compare the message content string itself?

kristofbolyai (commented):
Looks better! I still would like a minor refactor to make this look a bit better.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
san7890: I had originally had it work on the strings themselves but was under the impression that a hashed integer would be simpler to check again. Understandable if this micro-opt isn't necessary. I thought your recommended something along the line of "hashcode equals" in Discord the other day, is this not what you meant?

kristofbolyai: > I had originally had it work on the strings themselves but was under the impression that a hashed integer would be simpler to check again. I thought your recommended something along the line of "hashcode equals" in Discord the other day, is this jot what you meant?

No, I did not mean this. I don't see why you couldn't just .equals 2 strings.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
san7890: Unsure what you mean here.

kristofbolyai: Boolean vs boolean. U used Boolean, why?

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,21 +9,36 @@
 
 public class MessageContainer {
     protected TextRenderTask message;
+    private Integer iterations;
```
kristofbolyai: This too.

Why not make this an int?

```suggestion
    private int iterations;
```

int is a primitive type.
Integer is an Object around int, which means Integer can be null, int can. We do not want this to be null.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
kristofbolyai:
```suggestion
    public void editMessage(String newMessage, boolean incrementIterations) {
```

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
magicus: To clarify, you should append the iteration "on the fly" in `getRenderTask`.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
san7890: ack, this was cruft from my own testing

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
san7890: Alright, I think I did what was recommended to me here. Incrementing the iterations is done by the method that "dispatches" the message, and `getRenderTask` calls a method to read the number of iterations and then concatenate that to the message.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,25 @@
 
 public class MessageContainer {
     protected TextRenderTask message;
+    private int iterations;
 
     public MessageContainer(String message) {
         this.message = new TextRenderTask(message, TextRenderSetting.DEFAULT);
+        this.iterations = 0;
```
kristofbolyai: shouldn't we start this from 1?

So, when we have a message like this again, we have 2 "iterations". Maybe we could rename the variable to `duplicateCount` so it is clearer?

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -29,4 +37,13 @@ public void editMessage(String newMessage) {
     public void update(MessageContainer other) {
         this.message = other.message;
     }
+
+    public void incrementIterations() {
+        this.iterations++;
+    }
+
+    private String duplicateMessageBuilder(MessageContainer container) {
+        if (this.iterations == 1 || this.iterations == 0) return "";
```
kristofbolyai: (dont forget to adapt this if)

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
san7890: right now, the iterations of a message are only incremented whenever a message matches one in the cache. i can change the name to make that more clear.

kristofbolyai: Feels a bit weird to have it start from 1 imo

san7890: yeah i come from a language where arrays start at 1 (lol it's fucked up) so i thought that way in the initial implementation of this PR but it's not hard to adjust the logic to have it make sense this way

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,25 @@
 
 public class MessageContainer {
     protected TextRenderTask message;
+    private int messageCount;
 
     public MessageContainer(String message) {
         this.message = new TextRenderTask(message, TextRenderSetting.DEFAULT);
+        this.messageCount = 0;
```
kristofbolyai: This is still 0?

So in `duplicateMessageBuilder` 0 means no duplication. Then we increment message count. 1 means no duplication. That is not good. Make this start from 1.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,6 +43,26 @@ public static MessageContainer queueMessage(TextRenderTask message) {
...
+                    sendOrEditNotification(msgContainer);
+                }
+
+                WynntilsMod.postEvent(new NotificationEvent.Edit(matchedContainer));
+                return matchedContainer;
+            }
+        }
+
+        cachedMessageSet.put(new Pair<>(messageText, msgContainer));
+        msgContainer.incrementMessageCount();
```
kristofbolyai: I see! So you default to 0, increment instantly. That is a bit of a twisted logic though, imo. Let's just start it from 1 and remove this line.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,6 +43,26 @@ public static MessageContainer queueMessage(TextRenderTask message) {
...
+        String messageText = message.getText();
+
+        for (Pair<String, MessageContainer> cachedMessagePair : cachedMessageSet) {
+            String checkableMessage = cachedMessagePair.a();
+            if (messageText.equals(checkableMessage)) {
+                MessageContainer matchedContainer = cachedMessagePair.b();
+                matchedContainer.incrementMessageCount();
+
+                // Overlay is not enabled, send in chat
+                if (!GameNotificationOverlayFeature.INSTANCE.isEnabled()) {
```
kristofbolyai: You moved this here. But!. What if we don't have cached messages? Shouldn't we handle this case after the for loop as well?

san7890: it is handled after the for loop? see lines 69-71.

kristofbolyai: Sorry! Github perfectly cut the code, so I didn't see that.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,25 @@
...
+        this.messageCount = 1;
     }
 
     public TextRenderTask getRenderTask() {
-        return message;
+        String sendableMessage = message.getText() + duplicateMessageBuilder(this);
+        return new TextRenderTask(sendableMessage, this.message.getSetting());
+    }
+
+    public Integer getMessageCount() {
```
kristofbolyai:
```suggestion
    public int getMessageCount() {
```

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,25 @@
...
 
     public MessageContainer(TextRenderTask message) {
         this.message = message;
+        this.messageCount = 1;
     }
 
     public TextRenderTask getRenderTask() {
-        return message;
+        String sendableMessage = message.getText() + duplicateMessageBuilder(this);
+        return new TextRenderTask(sendableMessage, this.message.getSetting());
```
kristofbolyai: I just noticed this! This is a no-no. This code is called every frame render. That means, we allocate a lot. Usually code like this looks just fine, but if you actually think about this, this is a perfectly hidden memory sink.

To fix this, keep a cached TextRenderTask with the duplicate message part, `renderedMessage`. Then, you can update it in `incrementMessageCount`, which is way less allocations.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -10,12 +10,26 @@
...
 import com.wynntils.wynn.event.NotificationEvent;
+import com.wynntils.wynn.event.WorldStateEvent;
 import com.wynntils.wynn.utils.WynnUtils;
+import java.util.concurrent.TimeUnit;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.TextComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class NotificationManager {
+    private static final TimedSet<Pair<String, MessageContainer>> cachedMessageSet =
```
kristofbolyai: This Set could contain only a `MessageContainer`. If you make a method, `getOriginalMessage`, in MessageContainer, you can use it to compare messages here, and don't need to store it in the set.

Note that, This change is only to make this code a bit shorter/cleaner, it does not save that much memory.

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,6 +43,25 @@ public static MessageContainer queueMessage(TextRenderTask message) {
...
+            if (messageText.equals(checkableMessage)) {
+                MessageContainer matchedContainer = cachedMessagePair.b();
+                matchedContainer.incrementMessageCount();
+
+                // Overlay is not enabled, send in chat
+                if (!GameNotificationOverlayFeature.INSTANCE.isEnabled()) {
+                    sendOrEditNotification(msgContainer);
+                }
+
+                WynntilsMod.postEvent(new NotificationEvent.Edit(matchedContainer));
```
magicus: Can you extract line 55-59 into a separate `doSendMessage` (or something like that), and also replace the code at line 66-71 with a call to the same function?

I also really don't like that we check a feature from a manager, but I have no good suggesion at hand on how to solve this otherwise. But we should at least not duplicate the code.

Other than this, the PR is starting to look really, really good!

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
san7890: alright, i think i did what you wanted. my current implementation requires a nullcheck since i'm not able to figure out a better way to do it, so i would greatly appreciate any further pointers

magicus: @san7890 You did what I asked of you. I did not think of this problem. :-( I apologize.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,29 @@
...
+        this.messageCount = 1;
+    }
+
+    public String getOriginalMessage() {
+        return message.getText();
     }
 
     public TextRenderTask getRenderTask() {
-        return message;
+        if (renderedMessage == null) return message;
```
magicus: My suggestion is:
1) Rename it `currentMessage`, since it is not "rendered" yet

2) Set it directly in the constructor (to the original message) and you can just return it here, no need for a check.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -29,4 +41,16 @@ public void editMessage(String newMessage) {
...
         this.message = other.message;
     }
+
+    public void incrementMessageCount() {
+        this.messageCount++;
+        this.renderedMessage =
+                new TextRenderTask(this.message.getText() + duplicateMessageBuilder(this), this.message.getSetting());
+    }
+
+    private String duplicateMessageBuilder(MessageContainer container) {
```
magicus: Also, at this point, you can inline this into increamentMessageCount (and then get rid of the if statement, since you always know it is >= 1 after the ++).

common/src/main/java/com/wynntils/core/notifications/NotificationManager.java
```diff
@@ -29,25 +41,41 @@ public static MessageContainer queueMessage(TextRenderTask message) {
...
     }
 
     public static void editMessage(MessageContainer msgContainer, String newMessage) {
         msgContainer.editMessage(newMessage);
 
-        WynntilsMod.postEvent(new NotificationEvent.Edit(msgContainer));
+        doSendMessage(msgContainer, true);
+    }
+
+    private static void doSendMessage(MessageContainer container, boolean isEdit) {
```
magicus: Oh, drat, I didn't see it was different events. Having booleans in a method parameter list is seldom a good design choice. In this case, it is better to either instantiate and send in the correct Event from the call site, or move the postEvent stuff out of this function and just keep the check if overlay is not enabled. If you go the latter route, maybe rename it to `sendToChatIfNeeded` (??).

magicus: I am sorry for suggesting a refactorization that did not really work out. I should have been more careful when looking at the code. :(

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
san7890: I sent you a ping in Discord on why the null-check was here and why it's not defined in the constructor because I had already tried that and was met with the following crash reason. https://hst.sh/ogusuragit Very possible that I was doing something incorrectly at the time of that comment, I believe that log is from the implementation at https://github.com/Wynntils/Artemis/pull/781/commits/ca26cf96a072d276347b35aae2d582ce006347ad.

Will rename it though.

kristofbolyai:
@san7890 I fixed it, I think

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,25 @@
...
+    public String getOriginalMessage() {
+        return message.getText();
     }
 
     public TextRenderTask getRenderTask() {
-        return message;
+        return currentMessage;
     }
 
     public void editMessage(String newMessage) {
```
magicus, line 36: Well this won't fly, will it? When do we call this? It will just edit the "original" message, not the one returned by getRenderTask().

*sigh* I was just about to press approve. Are we doing something wrong, design-wise, since this has turned out to be so difficult to get right?

magicus: I checked. This is called only once, when we pick up emeralds in the pouch in successively in short order.
I guess the best solution is to try to keep this functionality but fix this method. That is, we need to re-create the current message. Which, given the current design, is a bit messy. :(

kristofbolyai: This is fine as is, I think. Yes, it does need to update the cached message, and it does not, but we should save it to the "original" message. I am unsure why we even call it original / current message. Message is the thing we want to display, and we can edit it (for example, `10 emeralds picked up` -> `202 emeralds picked up`). What `currentMessage` currently is just a cached version of the render task that includes the "[x2]", aka the duplication string, so we do not have to allocate it.

kristofbolyai: So, IMO, call `incrementMessageCount` in edit message. Also rename `currentMessage` -> `renderedMessageTaskCache`, as it is no more than that. `getOriginalMessage` should become `getMessage`, as it does not return the __original__ message, but the current version, without the duplication part.

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -29,4 +37,10 @@ public void editMessage(String newMessage) {
     public void update(MessageContainer other) {
```
magicus, line 40: Oh no. It gets worse. :( I'm seriously starting to think we need to redesign this class and its interactions with the manager. I looked up this as well. It, too, is only called once, from the TimedMessageContainer, which in turn is called from the notification overlay. I don't fully understand the code, but it is used to take an existing message and reset its lifespan.

kristofbolyai (dismissed):
@san7890 Just to bring you up to speed. Magicus and I had a long conversation about this PR. We decided that this version will be the best for now, and once we merge this, I am going to start working on the refactoring of NotificationManager/MessageContainer/TimedMessageContainer/GameUpdatesOverlay, basically everything related.

Sorry for this PR being such a ping-pong between you and us. This is what happens when our original code is not so great, adding new changes requires a lot of work. Your PR was great overall, and thank you for following through every change request.

kristofbolyai:
Once @magicus checks this version out too, we can merge this :)

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
```diff
@@ -9,17 +9,24 @@
...
+        this.messageCount = 1;
+    }
+
+    public String getOriginalMessage() {
+        return message.getText();
     }
 
     public TextRenderTask getRenderTask() {
-        return message;
+        String messageMultiplier = " §7[x" + this.messageCount + "]";
```
magicus, line 32: You just need to add a 
```
if (messageCount == 1) return message;
```
first, and then it will 1) be correct and 2) be more efficient. ;-)

magicus:
> Your PR was great overall, and thank you for following through every change request.

I second that. Thank you @san7890 for your contribution, and for hanging in there!

common/src/main/java/com/wynntils/core/notifications/MessageContainer.java
kristofbolyai: Fixed!

magicus (approved):

magicus (approved):

PR #783, opened 2022-12-05
feat: Poi location and rendering refactors

kristofbolyai:
This PR changes a lot of things related to POIs:

- Rename `MapLocation` to `PoiLocation`
- `PoiLocation` Y level is now optional/nullable
- POIs now have render-priority:
  - This is used to keep a sane rendering order of the POIs on the maps, so we don't have to rely on dummy Y levels for POIs that have no logical Y level
  - A lower priority means rendering is done earlier. Highest priority is rendered on top of everything/last.
  - New rendering order (earlier on list means earlier rendering, last to render last):
    - ServicePOIs (Banks, Merchants, Etc)
    - CustomPOIs and PlayerPOIs (Hades Players)
    - WaypointPOIs (Compass)
    - LabelPOIs (Location Labels)
    - TerritoryPOIs (Guild Territories)
- Minor POI code fixes
  - Include `@Override` annotations in cases where it was missing
  - Simplify `TerritoryPoi` constructor

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
+++ b/common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
@@ -8,7 +8,13 @@ import com.mojang.blaze3d.vertex.PoseStack;
 
 public interface Poi {
 
-    MapLocation getLocation();
+    PoiLocation getLocation();
+
+    /**
+     * Render priority is used to determine the order in which POIs are rendered.
+     * A lower render priority means, that the POI is rendered later, so it will be on top of other POIs.
+     */
+    RenderPriority getRenderPriority();
 
     boolean hasStaticLocation();
 
--- a/common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
+++ b/common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
@@ -11,10 +11,10 @@ public interface Poi {
     PoiLocation getLocation();
 
     /**
-     * Render priority is used to determine the order in which POIs are rendered.
+     * Display priority is used to determine the order in which POIs are rendered.
      * A higher render priority means, that the POI is rendered later, so it will be on top of other POIs.
      */
-    RenderPriority getRenderPriority();
+    DisplayPriority getDisplayPriority();
 
     boolean hasStaticLocation();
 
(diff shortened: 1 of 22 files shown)


kristofbolyai:
I would like to explicitly request a review from @magicus, as he is the original developer to suggest this refactor.

magicus:
> * A higher priority means rendering is done earlier. Lowest priority is rendered on top of everything/last.

Do you really mean this? It feels backwards. Highest priority should be rendered last so it is always visible.

magicus:
I'll look more into the PR later on.

kristofbolyai:
> > * A higher priority means rendering is done earlier. Lowest priority is rendered on top of everything/last.
> 
> 
> 
> Do you really mean this? It feels backwards. Highest priority should be rendered last so it is always visible.

Feels backwards either way imo.

kristofbolyai:
> > * A higher priority means rendering is done earlier. Lowest priority is rendered on top of everything/last.
> 
> Do you really mean this? It feels backwards. Highest priority should be rendered last so it is always visible.

Reversed it.

magicus:
> Feels backwards either way imo.

lol. 

The code basically looks good, but I agree that the priority order still sounds like it could be the other way. What if you rename it "displayPriority" instead of "renderPrirority"? That makes sure it tells us that we prioritize what we actually see (and how it is rendered is an implementation detail).

Also, maybe add a comment at the two places in rendering when we do the `reverse()` calls along the lines of
```
// Reverse order to make sure higher priority is drawn later than lower priority to overwrite them
```

kristofbolyai:
> > Feels backwards either way imo.
> 
> lol.
> 
> The code basically looks good, but I agree that the priority order still sounds like it could be the other way. What if you rename it "displayPriority" instead of "renderPrirority"? That makes sure it tells us that we prioritize what we actually see (and how it is rendered is an implementation detail).
> 
> Also, maybe add a comment at the two places in rendering when we do the `reverse()` calls along the lines of
> 
> ```
> // Reverse order to make sure higher priority is drawn later than lower priority to overwrite them
> ```

Should be fixed. It sounds better-ish. Better than renderPriority for sure.

common/src/main/java/com/wynntils/wynn/model/map/poi/Poi.java
```diff
@@ -11,10 +11,10 @@ public interface Poi {
     PoiLocation getLocation();
 
     /**
-     * Render priority is used to determine the order in which POIs are rendered.
+     * Display priority is used to determine the order in which POIs are rendered.
      * A higher render priority means, that the POI is rendered later, so it will be on top of other POIs.
```
magicus:
```suggestion
     * A higher display priority means, that the POI is rendered later, so it will be on top of other POIs.
```

magicus (approved):

PR #784, opened 2022-12-05
fix: Fix unusable delete waypoint keybind

kristofbolyai:
Control shows territories, so you can't click on POIs.

### DIFF
--- a/common/src/main/java/com/wynntils/utils/KeyboardUtils.java
+++ b/common/src/main/java/com/wynntils/utils/KeyboardUtils.java
@@ -13,10 +13,14 @@ public final class KeyboardUtils {
     }
 
     public static boolean isShiftDown() {
-        return isKeyDown(GLFW.GLFW_KEY_RIGHT_SHIFT) || isKeyDown(GLFW.GLFW_KEY_LEFT_SHIFT);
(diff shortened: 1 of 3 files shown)


magicus (approved):
So where/how do we differ from Legacy?

kristofbolyai:
> So where/how do we differ from Legacy?

We have new features / keybinds. We can't really compare to legacy.

common/src/main/java/com/wynntils/utils/KeyboardUtils.java
```diff
@@ -13,10 +13,14 @@ public static boolean isKeyDown(int keyCode) {
     }
 
     public static boolean isShiftDown() {
-        return isKeyDown(GLFW.GLFW_KEY_RIGHT_SHIFT) || isKeyDown(GLFW.GLFW_KEY_LEFT_SHIFT);
+        return isKeyDown(GLFW.GLFW_KEY_LEFT_SHIFT) || isKeyDown(GLFW.GLFW_KEY_RIGHT_ALT);
```
cmkohnen, line 16: Doesn't this have to be GLFW_KEY_RIGHT_SHIFT like it used to be instead of GLFW_KEY_RIGHT_ALT?

kristofbolyai: It should be! Thanks.

PR #787, opened 2022-12-06
chore: Remove special DEV version case from build.gradle

kristofbolyai:
There is no need for "DEV" version. Also, this is fixed on 1.19.2 branch too. "DEV" version makes forge not launch. This is a good enough back port for now.


kristofbolyai:
I dont know if merging this will break the release system.

magicus (approved):
Fine, I guess. Be aware that the promote bot will probably fail to integrate this in alpha, so you might have to do a manual merge into alpha to help it along the way.

PR #788, opened 2022-12-06
fix: Use invoker for ChatScreen#insertText to fix Forge ChatTab crash

kristofbolyai:
#787 was a preparation for this. 
There is a weird forge bug that this specific AW seems to crash. I am not sure why. It probably fails at AW/AT transformation. I don't know why, but this is likely an issue upstream in Architectury/Forge. I just replaced this with an invoker, and it works on my production build.


magicus:
Does it crash during the build of forge, rather than at runtime? (I don't get how this is related to #787)

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
```diff
@@ -48,7 +49,7 @@ public static void onScreenOpened(ScreenOpenedEvent event) {
         if (!(event.getScreen() instanceof ChatScreen chatScreen)) return;
         if (focusedTab == null || focusedTab.getAutoCommand() == null) return;
 
-        chatScreen.insertText(focusedTab.getAutoCommand(), true);
+        ((ChatScreenInvoker) chatScreen).invokeInsertText(focusedTab.getAutoCommand(), true);
```
magicus: Since this is a bit ugly (and can hopefully be reverted in the future if this is fixed upstream), maybe you can wrap this in a static helper function `insertText(ChatScreen chatScreen, String text)` or maybe even better name it `replaceChatText` since that is what this is actually doing.

kristofbolyai:
> Does it crash during the build of forge, rather than at runtime? (I don't get how this is related to #787)

I needed to fix forge env to build a runnable mod

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: In idea, this folds itself nicely.

kristofbolyai:
> Does it crash during the build of forge, rather than at runtime? (I don't get how this is related to #787)

It used to crash at chat tab change.

common/src/main/java/com/wynntils/core/chat/tabs/ChatTabModel.java
kristofbolyai: Actually, it does not. But I dont really think we should change this.

magicus: Not even to rename it `replaceChatText`? (I really dislike the vanilla design of having a boolean which turns an "insert" into a "replace" -- I needed to go read the code to understand why this was not broken)

Edit: to be clear, I was a bit annoyed at this already at the autocommand PR, but forgot to say anything.

If you do not extract it, please consider adding a comment. (But my motto is that having good names usually gets rid of the need for comments)

kristofbolyai: Okay, Ill do it

magicus: Thanks!

magicus (approved):

PR #789, opened 2022-12-07
feat: mythic box scaling

Incompleteusern:
image
image
image


kristofbolyai:
Did this just become a draft for a bug or did you intend it that way originally?

common/src/main/java/com/wynntils/features/user/MythicBoxScalerFeature.java
```diff
@@ -0,0 +1,31 @@
...
+    public void onItemRendering(GroundItemEntityRenderEvent e) {
+        if (!WynnItemMatchers.isMythic(e.getItemStack())) return;
+
+        PoseStack stack = e.getPoseStack();
+
+        stack.scale(scale, scale, 1f);
+
+        // Direction is always null when passed into SkullBlockRenderer#renderSkull which causes a 0.5 translation in x
+        // which we need to cancel out due to scaling
+        stack.translate(-0.5 + 0.5 / scale, 0, 0);
```
kristofbolyai: Shouldn't we cancel this out post render?

Incompleteusern: This line of code should be obselete and removable; I need to add a post event yeah

common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
```diff
@@ -165,11 +165,15 @@ public static boolean isCraftedGear(ItemStack itemStack) {
 
     public static boolean isMythic(ItemStack itemStack) {
         // only gear, identified or not, could be a mythic
-        if (!(isUnidentified(itemStack) || isGear(itemStack))) return false;
+        if (!(isUnidentified(itemStack) || isGear(itemStack)) || !isMythicBox(itemStack)) return false;
```
kristofbolyai: This condition is a making my head hurt.

kristofbolyai: Also why did this get changed this?

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/features/user/MythicBoxScalerFeature.java
```diff
@@ -0,0 +1,32 @@
...
+    private float scale = 2f;
+
+    @SubscribeEvent
+    public void onItemRendering(GroundItemEntityTransformEvent e) {
+        if (!WynnItemMatchers.isMythicBox(e.getItemStack())) return;
+
+        PoseStack stack = e.getPoseStack();
+
+        // cancel out old scale y transform and later replace it after scaling
+        stack.translate(0f, -0.25f, 0f);
```
magicus, line 30: This magic constant 0.25, is it depending on something Vanilla did prior to this getting called by the mixin? The comment reads a bit odd -- it says cancel "scale y" but it does this by doing a translate. Don't you mean that you cancel out the translation? Also, maybe make a final static constant of the magic number?

common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
magicus: I agree, this is pretty tough to understand. And it does not seem to be used? 

Why was this method not used instead? I don't think it is likely that people go throwing out their mythical gear on the ground, but in the odd case that they actually do, I think it could just as well be scaled up, so not only boxes (unidentified mythics) but also identified mythics could be scaled. 

On the third hand, the isMythicBox implementation seems much more efficient, and I presume this will be called often when rendering items on the ground. So maybe we should go the other way around, making isMythic something like this:

```
if (isMythicBox(itemStack) return true;
if (!isGear(itemStack)) return false;

return itemStack.getHoverName().getString().contains(ChatFormatting.DARK_PURPLE.toString());
```

common/src/main/java/com/wynntils/features/user/MythicBoxScalerFeature.java
Incompleteusern: It's a constant from vanilla

common/src/main/java/com/wynntils/wynn/item/parsers/WynnItemMatchers.java
Incompleteusern: The `isMythicBox` is a check that it has the texture of a unid mythic box by checking it is stone shovel with damage value 6. The `isMythic` method does not work for items on the ground due to differences from wynncraft.

> but also identified mythics could be scaled.

This is simply different from mythic boxes since it has a different texture, and would require a redicussing with the moderators I think

common/src/main/java/com/wynntils/features/user/MythicBoxScalerFeature.java
magicus: Ok, let me be clearer:

1. I think you should extract this magic value to a constant, and make a good name or a comment about what kind of value this is, how you did find it (in case it changes in Vanilla and needs updating etc).

2. I think the comment is wrong. You are not cancelling out a scale, you are cancelling out a translation. Either explain to me why the comment is correct, or fix it.

Incompleteusern: >I think you should extract this magic value to a constant, and make a good name or a comment about what kind of value this is, how you did find it (in case it changes in Vanilla and needs updating etc).

I will do the later, but don't think there is any point in extracting when it only appears twice in quick succession

> Effectively, vanilla code plays out as
```java
        stack.translate(0f, -0.25f * model_scale, 0f); // transform y of item on ground
        ...
        stack.scale(model_scale, model_scale, model_scale); // scale model
```

We inject after the `scale` application so that it becomes

```java
        stack.translate(0f, -0.25f * model_scale, 0f); // transform y of item on ground
        ...
        stack.scale(model_scale, model_scale, model_scale); // scale model
        stack.translate(0f, -0.25f, 0f); // new code
        stack.scale(scale, scale, scale); // new code
        stack.translate(0f, -0.25f, 0f); // new code
```

which should be equivalent to
```java
        stack.translate(0f, -0.25f * model_scale * scale, 0f); // transform y of item on ground
        ...
        stack.scale(model_scale * scale, model_scale * scale, model_scale * scale); // scale model
```

magicus: I think this was an excellent explanation! Can you please add this, or a revised version of this, as a comment to this code?

Incompleteusern: Should be done

magicus (approved):

PR #791, opened 2022-12-07
refactor: In preparation of the webapi refactoring, extract managers/model out of WebManager

magicus:
This too is basically a "pure" refactoring, where I have just shuffled code around. The one exception to this is that I have created a new local GSON object for those managers which needed it, instead of sharing the original WebManager gson object.

When reviewing, it might be easier to check individual commits, where I tackle one manager at a time.


kristofbolyai (approved):
Looks good

PR #792, opened 2022-12-07
refactor: In preparation of the webapi refactoring, fixup TerritoryManager

magicus:
This patch is probably a bit more tricky to follow. Github will show it as if I deleted the old TerritoryManager and GuildTerritoryModel, and created a brand new TerritoryManager in `com.wynntils.wynn.model.territory`. Well, I didn't. Instead I renamed the old GuildTerritoryModel to TerritoryManager, promoted it to CoreManager, and moved the logic from the old `com.wynntils.core.webapi.TerritoryManager` into this class. 

After this was done, I cleaned up the new TerritoryManager somewhat. This is still basically just refactorings; no actual behavior should be modified.


common/src/main/java/com/wynntils/wynn/model/territory/TerritoryManager.java
```diff
@@ -0,0 +1,180 @@
...
+    public static Stream<String> getTerritoryNames() {
+        return territoryProfileMap.keySet().stream();
+    }
+
+    public static Set<TerritoryPoi> getTerritoryPois() {
+        return allTerritoryPois;
+    }
+
+    public static List<Poi> getTerritoryPoisFromAdvancement() {
+        // FIXME: Clarify how this differs from getTerritoryPois()...
```
kristofbolyai: So?

kristofbolyai (dismissed):
Looks fine apart from that one FIXME

common/src/main/java/com/wynntils/wynn/model/territory/TerritoryManager.java
magicus: That was intended as a future reminder to myself. I'll remove it from here and open a bug in GH instead.

magicus: https://github.com/Wynntils/Artemis/issues/793

kristofbolyai (approved):

PR #796, opened 2022-12-07
feat: Read splashes from splashes.json instead of webapi

magicus:
Once again, title says it all

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/SplashManager.java
+++ b/common/src/main/java/com/wynntils/wynn/model/SplashManager.java
@@ -4,13 +4,27 @@
  */
 package com.wynntils.wynn.model;
 
+import com.google.common.reflect.TypeToken;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import com.wynntils.core.managers.CoreManager;
 import com.wynntils.core.webapi.WebManager;
+import com.wynntils.core.webapi.request.RequestBuilder;
+import com.wynntils.core.webapi.request.RequestHandler;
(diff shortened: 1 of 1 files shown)


P0keDev (approved):

common/src/main/java/com/wynntils/wynn/model/SplashManager.java
```diff
@@ -4,13 +4,28 @@
...
+import java.util.ArrayList;
 import java.util.List;
 
 public class SplashManager extends CoreManager {
-    private static String currentSplash = "";
+    private static final String SPLASHES_URL =
+            "https://raw.githubusercontent.com/Wynntils/WynntilsWebsite-API/master/splashes.json";
+    // Fallback splash in case loading of splashes fails
+    private static final String DEFAULT_SPLASH = "The best Wynncraft mod you'll probably find!";
+    private static final Gson GSON = new GsonBuilder().create();
```
kristofbolyai, line 25: Didn't we use new Gson() some other places? Maybe use the builder everywhere?

magicus: We can certainly make some micro-optimization and create a single shared GSON instances in e.g. WynntilsMod. It might be sane from a code quality perspective as well as the small optimization benefit.

kristofbolyai: Yes. But I did not mean this. I meant that we use both new Gson() and Gsonbuilder.build() through the codebase; and we should generalize using the builder.

PR #799, opened 2022-12-08
fix: NotificationManager / MessageContainer refactors + handle editing repeated messages 

kristofbolyai:
The commits are separated into logical parts (imo). Basically, this PR does some refactoring around the above mentioned classes, fixes editing messages, and reintroduces the task render cache, as it is now much more saner to do than before.

Here is a demonstration of how overlay editing works (now correctly):

https://user-images.githubusercontent.com/49001742/206501005-0b546389-f752-49ad-8ffd-417944d3cc7a.mp4


kristofbolyai:
This is the first part of the notification manager changes we have discussed with magicus in the development discord. The other part, game update overlay refactor is unchanged here, because it deserves a separate PR.

magicus (approved):

PR #801, opened 2022-12-09
feat: Redirects "Selling Ingredients" Ingredient Pouch Message to Overlay

san7890:
image

image

The reason why I chose to do the formatted string is because there's a few moving variables (not unlike #743) in order to get a good-looking message that matches both the Wynncraft Chat Message as well as remain in-line with an already existing Overlay message format (Blacksmith Item Redirect).

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
@@ -359,11 +359,11 @@ public class ChatRedirectFeature extends UserFeature {
         @Override
         protected String getNotification(Matcher matcher) {
             Integer ingredientCount = Integer.parseInt(matcher.group(1));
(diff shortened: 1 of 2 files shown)


magicus (approved):

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -363,6 +367,33 @@ protected String getNotification(Matcher matcher) {
         }
     }
 
+    private class IngredientPouchSellRedirector extends SimpleRedirector {
+        private static final Pattern SYSTEM_PATTERN =
+                Pattern.compile("§dYou have sold §r§7(.+)§r§d ingredients for a total of §r§a(.+)§r§d\\.$");
```
kristofbolyai, line 372: same as [before](https://github.com/Wynntils/Artemis/pull/802/files#r1045116303). No ^ start but $ end?

PR #800, opened 2022-12-08
feat: Redirects "Scared Horse" messages to Overlay

san7890:
Simple PR with a message I overlooked in #745.

image

Changed chat formatting to red since it's an "alert" that the horse can't spawn rather than a "take it or leave it message". If you haven't seen it before, try and spawn a horse in an area crowded with NPCs in Rymek (that's how I discovered it).

I think the message is a bit too long, but I'm not sure what a briefer way to communicate is. Would appreciate any further thoughts.

PR #802, opened 2022-12-09
feat: Redirect "Aggressive Mobs Nearby" Scroll Teleport Message

san7890:
Incidentally ran into this while testing #800.

image

Yet another atomic PR keeping the Minecraft chat clean while leveraging the neat capabilities of the overlay.

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
@@ -473,8 +473,7 @@ public class ChatRedirectFeature extends UserFeature {
     }
 
     private class TeleportationFailRedirector extends SimpleRedirector {
-        private static final Pattern SYSTEM_PATTERN =
-                Pattern.compile("§cThere are aggressive mobs nearby...$");
+        private static final Pattern SYSTEM_PATTERN = Pattern.compile("§cThere are aggressive mobs nearby...$");
 
         @Override
         protected Pattern getSystemPattern() {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -468,6 +472,25 @@ protected String getNotification(Matcher matcher) {
...
+        }
+
+        @Override
+        public RedirectAction getAction() {
+            return teleport;
+        }
+
+        @Override
+        protected String getNotification(Matcher matcher) {
+            return ChatFormatting.DARK_RED + "Nearby mobs prevent scroll teleporation!";
```
magicus:
```suggestion
            return ChatFormatting.DARK_RED + "Nearby mobs prevent scroll teleportation!";
```

magicus (approved):

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -468,6 +472,25 @@ protected String getNotification(Matcher matcher) {
         }
     }
 
+    private class TeleportationFailRedirector extends SimpleRedirector {
+        private static final Pattern SYSTEM_PATTERN = Pattern.compile("§cThere are aggressive mobs nearby...$");
```
kristofbolyai, line 476: This seems like an incomplete regex to me? You have no `^` in the beginning but `$` at the end. This sure works but why use the only one?

kristofbolyai: (and you didn't use any in https://github.com/Wynntils/Artemis/pull/800/)

PR #805, opened 2022-12-09
fix: Fix colored string matchers for unused skill/ability points

magicus:


magicus:
I should note that I have not been able to test the Ability message in play. For some reason, Wynncraft seem to be very restrictive with sending this.

magicus:
I have now AFK'ed for more than 2 hours, and despite having no ability points assigned, I have not had a single instance of the "You have unused ability points" message. I'm starting to suspect it was never seen when it was added but just added out of symmetry...

kristofbolyai:
> I have now AFK'ed for more than 2 hours, and despite having no ability points assigned, I have not had a single instance of the "You have unused ability points" message. I'm starting to suspect it was never seen when it was added but just added out of symmetry...

It exists. I know it does, I have seen it, tested it (not this PR)

magicus:
I have not been able to provoke it, anyway. This is an educated guess on what it would look like. Are we prepared to accept it? If not, any idea on how to provoke that message to be displayed? The Skill point notice appears almost immediately after login, and every now and then after that...

kristofbolyai:
OH! Ability points, only. No, I think there are only skill point or skill + ability reminders. Sorry, I miss read this on phone.

kristofbolyai (approved):
I am fine with this. If you wanna remove `UnusedAbilityPointsRedirector` I do not mind. We can keep it though, for the future, if they add it.

PR #806, opened 2022-12-10
fix: Clean up ServerListModel and ServerCommand

magicus:
This is in preparation for the webapi refactor. It also fixes some issues with the `/server` command.

Changes: Make ServerListModel take more responsibility for its domain: Move getWynnServerTypes() here, move functionality for returning sorted server lists here.

Fix: stop using Wynncraft API directly, instead always use Athena.

Fix: introduce explicit request to refresh data from Athena before replying to /server commands.

Fix: Add "YT" to server type list.

Modernization: Use ScheduledExecutorService instead of starting a new thread.


kristofbolyai (approved):
Looks great! It will likely be even better with the new API!

PR #808, opened 2022-12-11
fix: Healing Yourself With A Potion properly redirects to overlay

san7890:
This feature was apparently implemented, but it didn't work:
image

I looked into it, and after a few potions, it turned out that this is actually a system message! How odd, I wonder if that changed at some point in time since the "Healed by Other" message has worked perfectly fine this entire time (which is a Normal/Background message).

Also, we had a `§r` there for some reason, even though Wynncraft sends the message as `§c[+4500 ?]` (pretend the question mark is a heart symbol). Two concurrent bugs, perhaps?

Here's what it looks like now that it's fixed.

image


san7890:
I think ChatFormatting.RED might look better here but that's out of scope for a "fix" PR. I might bundle that later when I'm done working on this other redirect project.

magicus (approved):

PR #809, opened 2022-12-11
fix: Ability Point Redirect Message now works with Pluralization

san7890:
Seems to have been introduced in #805.

example of this failing:

image

the `?` was matching to the `t` in `point`, rather than be a pluralization checker for potential `points` message in case someone insane has multiple unused skill points and ability points.


magicus:
Oh drat! Thanks for discovering and fixing!

magicus (approved):

PR #810, opened 2022-12-11
feat: Redirects Mage Failing to Teleport to overlay

san7890:
Adding this now since apparently the feature I implemented in PR NUMBER HERE accidentally caused some user confusion since some thought the config pertained to mage teleportation, instead of scroll teleportation. This just fixes that by adding the Mage Failure-to-Teleport case in.

I also renamed the Scroll redirector to provide clarity codeside.

I don't have a mage character readily available to test this on, apologies. Let me know if I should create a new config option for this.


magicus (commented):
I confirmed the chat message with my mage. But please use start/end of line markers.

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -424,6 +425,26 @@ protected String getNotification(Matcher matcher) {
         }
     }
 
+    private class MageTeleporationFailRedirector extends SimpleRedirector {
+        private static final Pattern SYSTEM_PATTERN =
+                Pattern.compile("§cSorry, you can't teleport... Try moving away from blocks.");
```
magicus:
```suggestion
                Pattern.compile("^§cSorry, you can't teleport... Try moving away from blocks.$");
```

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -424,6 +425,26 @@ protected String getNotification(Matcher matcher) {
...
+
+        @Override
+        protected Pattern getSystemPattern() {
+            return SYSTEM_PATTERN;
+        }
+
+        @Override
+        public RedirectAction getAction() {
+            return teleport;
```
kristofbolyai: separate this from scroll teleport config

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -40,6 +40,9 @@ public class ChatRedirectFeature extends UserFeature {
     @Config
     public RedirectAction loginAnnouncements = RedirectAction.REDIRECT;
 
+    @Config
+    public RedirectAction mage = RedirectAction.REDIRECT;
```
magicus: I think you should rename this `mageTeleport` and the `teleport` config to `scrollTeleport`.

san7890: I did it as just `mage` in case we wanted to extend the config to more mage-specific messages, but I'll change scrollTeleport now.

magicus: Are you thinking of some specific messages? One clear benefit of having one config per redirect is that the users have full control. Otoh, many configs can be overwhelming, but then again, it can be hard to determine which belong together. Right now I think we have only skill+ability points combined into one config, and horse spawning messages into another. All other are separate.

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -56,7 +59,7 @@ public class ChatRedirectFeature extends UserFeature {
     public RedirectAction speed = RedirectAction.REDIRECT;
 
     @Config
-    public RedirectAction teleport = RedirectAction.REDIRECT;
```
magicus, line 59: Also please reorder to keep alphabetical order.

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
san7890: I'll change that over then.

magicus (approved):
Thanks! This looks great now! 🎈

PR #811, opened 2022-12-11
feat: Sends an Overlay Message Whenever You Gain Combat XP

san7890:
image
image

This just ports over this Legacy feature (`GameUpdateEXPMessages`), with a lot of the code reworked to fit the new Artemis framework (as well as just be a lot easier to read/work with). It ran on every tick in Legacy, so I believe it should be fine to work here as well? Let me know if there's a better solution. A bunch of the math might be un-necessary, but it took me a while to get to something that I felt was appropriate.

Thanks for your time!


common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+import com.wynntils.core.config.Config;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.core.features.properties.FeatureInfo;
+import com.wynntils.core.notifications.NotificationManager;
+import com.wynntils.mc.event.ClientTickEvent;
+import com.wynntils.wynn.model.CharacterManager;
+import com.wynntils.wynn.utils.WynnUtils;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@FeatureInfo
```
kristofbolyai: You only need this if you want to change defaults. So you either forgot to do that or just have this here unneeded

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+public class CombatXPGainMessageFeature extends UserFeature {
+
+    @Config
+    public boolean getCombatXPGainMessages = true;
```
kristofbolyai: In Artemis, we don't need configs like this. If you disable the feature, then you wont get combat xp gain messages :)

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!WynnUtils.onWorld() || !getCombatXPGainMessages) {
+            return;
+        }
+        CharacterManager.CharacterInfo data = WynnUtils.getCharacterInfo();
+
+        newTickXP = data.getCurrentXp();
+
+        if (newTickXP == lastTickXP) {
+            return;
+        }
```
kristofbolyai:
```suggestion
        if (newTickXP == lastTickXP)  return;
```

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private static float newTickXP = 0;
+    private static float lastTickXP = 0;
+    private static int trackedPercentage = 0;
+
+    @SubscribeEvent
+    public void onTick(ClientTickEvent.End event) {
+        if (!WynnUtils.onWorld() || !getCombatXPGainMessages) {
+            return;
+        }
```
kristofbolyai:
```suggestion
        if (!WynnUtils.onWorld()) return;
```

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+
+@FeatureInfo
+public class CombatXPGainMessageFeature extends UserFeature {
+
+    @Config
+    public boolean getCombatXPGainMessages = true;
+
+    private static float newTickXP = 0;
+    private static float lastTickXP = 0;
+    private static int trackedPercentage = 0;
```
kristofbolyai: Why is this an int?

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+        CharacterManager.CharacterInfo data = WynnUtils.getCharacterInfo();
+
+        newTickXP = data.getCurrentXp();
+
+        if (newTickXP == lastTickXP) {
+            return;
+        }
+
+        int neededXP = data.getXpPointsNeededToLevelUp();
+        if (lastTickXP != 0) {
```
kristofbolyai: I am not sure if defaulting to 0 is beneificial, does this work nicely with level ups?

san7890:
current iteration of the code has it run every tick because of a false assertion i made reading legacy code, whoops. i'll mark this ready for review when i'm able to fix that.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -171,6 +171,8 @@
   "feature.wynntils.gammabright.lastGamma.name": "Last Gamma",
   "feature.wynntils.gammabright.name": "Gammabright Keybind",
   "feature.wynntils.gearViewer.name": "Gear Viewer",
+  "feature.wynntils.getCombatXPGainMessages.description": "Do you want to see updates to your XP as overlay messages?",
+  "feature.wynntils.getCombatXPGainMessages.name": "Combat XP Gain Messages",
```
kristofbolyai: So you still don't use the run configurations in your IDE. Or you just didn't test this. But, you didn't add an i18n for the feature name, which should make it fail immediately in development environment.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,59 @@
...
+        float percentGained = (float) gainedXP / neededXP;
+        // to keep the math as striaght as possible, we only multiply by 100
+        // at the very end of all percentage calculations.
+        float percentChange = (percentGained - trackedPercentage) * 100;
+
+        String message = String.format("§2+%d XP (§6%.2f%%§2)", gainedXP, percentChange);
+
+        NotificationManager.queueMessage(message);
+
+        lastTickXP = newTickXP;
```
kristofbolyai: Maybe set this before making a notification, so notification queuing is the last step of the process?

kristofbolyai (commented):
Great feature, I missed this! However, there are a few problems. When you fix them up, this will be great!

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: In fact, the i18n for the variable, which is not needed as I commented above, should not even look like this. You missed the whole feature part.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
kristofbolyai: (just to make it nicer imo)

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
san7890: ah i misunderstood that the last time you told me about it

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
san7890: The only time when experience is 0 is at the very start of a session, when you join a world. Otherwise, it will be a positive integer by some degree. There might be an edge case regarding that where you could get exactly 0 points, but that'll smooth over by the next second. Not sure what the best way to handle that edge case is, or even it's important. 

I'll test it in-game with a levelup and see what comes of it to ensure nothing's breaking.

kristofbolyai: But theoritically you could stay on 0 xp. I think. So maybe, use a different magic number for the edge case? -1?

san7890: If you "stay on" 0 XP then you get filtered by this above check:

```java
if (newTickXP == lastTickXP) return;
```

or do I misunderstand? This code only runs whenever there is an actual change in the XP from the last time we ran. The purpose for the if/else check here is only for the case for when you start at 0 XP upon joining the World, and then Wynncraft sends you your character's saved amount of XP within a few ticks of being in-world. Otherwise, you would get this type of message:

```txt
+ 50294 XP (0.0%)
```

Because, lastTickXP was 0. 0/Any Number is 0. We haven't set lastTickXP to anything since this method hasn't fully ran to the end where we actually give it a number to hold onto. This doesn't make any sense, so the stop-guard here is to account for that and ensure that you really get something like:

```txt
+ 50294 XP (54.2%)
```

Might be worthwhile to write a comment on that. Sorry I didn't clarify on this earlier, I just woke up a bit ago.

kristofbolyai: oh. okay, perhaps this is fine

kristofbolyai:
Just to note this in this PR. A new event hooked into `ClientPacketListener#handleSetExperience` would be better than on tick.

san7890:
> Just to note this in this PR. A new event hooked into `ClientPacketListener#handleSetExperience` would be better than on tick.

It would be nicer but you would have to handle tick-counting separately somehow to ensure grouping (and that you don't get flooded with a message on every instance of XP changing).

kristofbolyai:
> > Just to note this in this PR. A new event hooked into `ClientPacketListener#handleSetExperience` would be better than on tick.
> 
> 
> 
> It would be nicer but you would have to handle tick-counting separately somehow to ensure grouping (and that you don't get flooded with a message on every instance of XP changing).

You make ticks seconds, use timestamps

san7890:
I believe I've now tested all of the cases and should now have code to handle everything.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,101 @@
...
+    public void onTick(ClientTickEvent.End event) {
+        if (!WynnUtils.onWorld()) return;
+
+        tickCounter++;
+
+        if (tickCounter % tickDelay != 0) return;
+
+        tickCounter = 0;
+
+        CharacterManager.CharacterInfo data = WynnUtils.getCharacterInfo();
```
magicus: I think it could be a bit clearer if everything below this were extracted into a separate method, i.e. the things that actually are done when the ticks align.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,101 @@
...
+        float percentGained = gainedXP / neededXP;
+
+        float percentChange;
+        // The reason we do this is for the same reason as the if/else statement above. You can get cases where
+        // both trackedPercentage and percentGained are equal, which would result in a (0.0%) message.
+        if (trackedPercentage != percentGained) {
+            percentChange = percentGained * 100;
+        } else {
+            percentChange = trackedPercentage * 100;
+        }
```
magicus: My gut feeling here is that a lot of this seems like stuff that CharacterManager should be responsible for. But I can't clearly say how that should be done, so maybe it can stay here. It all seems a bit convoluted, but your comments indicate that this is since the problem is more complex than it seems, so I guess it is okay.

san7890: It would be nice if it could be more abstract, yea- the only problem is that there are so many cases I ran into while working out the kinks of this system that there really does need to be some sort of context-dependent way to keep catching and handling odd quirks.

magicus (dismissed):

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private void gatherAndDispatchMessage() {
+        CharacterManager.CharacterInfo data = WynnUtils.getCharacterInfo();
+
+        int newLevel = data.getXpLevel();
+
+        // You get division by zero errors when you're at the level cap (i.e. 106 in Wynncraft 2.0.1).
+        // This needs to be updated if the level cap is ever raised.
+        if (newLevel > 105) return;
```
kristofbolyai: I do not like this at all. It is one thing that it does not work on new versions, but it would be nice if we did not straight up crash the thing... The only reason why I accept this is that we do not crash since the feature will capture the crash and disable itself.

kristofbolyai:
Hey @magicus and @san7890. I pushed 2 commits to address the issues I found to be annoying/blocking for release. I didn't want to take away the learning opportunity to code this such text book event/mixin, but I am sure san7890 will find some other event to add after they finish their finals season. Sorry about that.

I think, this is now in a state where we can merge this safely, after reviews.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,107 @@
...
+import com.wynntils.core.config.Config;
+import com.wynntils.core.features.UserFeature;
+import com.wynntils.core.notifications.NotificationManager;
+import com.wynntils.mc.event.SetExperienceEvent;
+import com.wynntils.wynn.event.WorldStateEvent;
+import com.wynntils.wynn.model.CharacterManager;
+import com.wynntils.wynn.utils.WynnUtils;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class CombatXPGainMessageFeature extends UserFeature {
```
magicus: I think it should be `CombatXpGainMessageFeature`, and similarly for the variables below, `newTickXp`etc. Combining acronyms traditionally written in all-caps with CamelCase is always difficult, but the normal Java convention is to treat them as words and camel-case them ("Url"), and I think that's the best. That is also what we have done before wrt to XP.

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,107 @@
...
+    @Config
+    public float secondDelay = 0.5f;
+
+    private long lastExperienceDisplayTime = 0;
```
magicus:
```suggestion
    private long lastXpDisplayTime = 0;
```

perhaps?

common/src/main/java/com/wynntils/features/user/CombatXPGainMessageFeature.java
```diff
@@ -0,0 +1,107 @@
...
+    @SubscribeEvent
+    public void onWorldStateChange(WorldStateEvent event) {
+        lastExperienceDisplayTime = 0;
+        newTickXP = 0;
+        lastTickXP = 0;
+        trackedPercentage = 0;
+    }
+
+    @SubscribeEvent
+    public void onExperienceChange(SetExperienceEvent event) {
```
magicus:
```suggestion
    public void onXpChange(SetXpEvent event) {
```

I think we should stick to XP for experience points. We could use "Experience" consistently as well, but that is not as clear and concise, I think. Mixing the two randomly is the least good solution.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -157,6 +158,12 @@ private void handleSetPlayerTeamPacketPre(ClientboundSetPlayerTeamPacket packet,
         }
     }
 
+    @Inject(method = "handleSetExperience", at = @At("HEAD"))
```
magicus:
```suggestion
    @Inject(method = "handleSetExperience", at = @At("RETURN"))
```

magicus (approved):

magicus:
Thank you @san7890 for the initial implementation, and thanks @kristofbolyai for the event-based version (and for doing the experience -> xp rename).

PR #822, opened 2022-12-13
fix: ItemRendererMixin now correctly loads on every mod loader (added a specific Forge version)

kristofbolyai:


kristofbolyai:
@Incompleteusern, pinging you just so you are aware.

kristofbolyai:
Although making a common Quilt/Fabric mixin package would be nice, I am not sure if that can be achieved.

Incompleteusern (approved):

PR #825, opened 2022-12-13
feat: Introduce NetManager and UrlManager

magicus:
Here is the first part of the webapi refactoring.

This patch introduces the new managers and supporting classes.

It also includes the conversion of `RemoteWynntilsUserInfoModel` and `DiscoveryManager` to use the new API. The reason for this is to show how it will look like when it is actually used, and to make it possible to test/try out the new code. These classes where chosen since `DiscoveryManager` conveniently uses all of `callApi`, `download` and `openLink`, and `RemoteWynntilsUserInfoModel` was a good example of a post call with the new post argument handling.

I believe the error handling of the NetResult handlers can be improved; but I don't think the current state is any worse than what we currently have in the old webapi.


common/src/main/java/com/wynntils/core/net/NetManager.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private static boolean checkLocalHash(File localFile, String expectedHash) {
+        if (!localFile.exists()) return false;
+
+        try (InputStream is = Files.newInputStream(localFile.toPath())) {
+            String fileHash = DigestUtils.md5Hex(is);
+            return fileHash.equalsIgnoreCase(expectedHash);
+        } catch (IOException e) {
+            WynntilsMod.warn("Error when calculading md5 for " + localFile.getPath(), e);
```
san7890:
```suggestion
            WynntilsMod.warn("Error when calculating md5 for " + localFile.getPath(), e);
```

common/src/main/java/com/wynntils/core/net/ApiResponse.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.InputStream;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ApiResponse extends NetResult {
+    private static final Map<HttpRequest, CompletableFuture<InputStream>> DOWNLOAD_FUTURES = new HashMap<>();
```
kristofbolyai: Is Qodana relevant here? Or we just not use this because it will be used in #826?

common/src/main/java/com/wynntils/core/net/Download.java
```diff
@@ -0,0 +1,99 @@
...
+            if (a.isEmpty()) return System.currentTimeMillis();
+            return a.getAsLong();
+        } catch (InterruptedException | ExecutionException e) {
+            return System.currentTimeMillis();
+        }
+    }
+
+    protected CompletableFuture<InputStream> getInputStreamFuture() {
+        if (request == null) {
+            // File is already in downloaded, just read from the cache
```
kristofbolyai, line 58: What is we have a request and a local file? It is likely handled fine but I have not yet understand the code as a whole

common/src/main/java/com/wynntils/core/net/NetManager.java
```diff
@@ -0,0 +1,136 @@
...
+import java.nio.file.Files;
+import java.time.Duration;
+import java.util.Map;
+import net.minecraft.Util;
+import org.apache.commons.codec.digest.DigestUtils;
+
+public class NetManager extends CoreManager {
+    protected static final HttpClient HTTP_CLIENT = HttpClient.newHttpClient();
+
+    private static final int REQUEST_TIMEOUT_MILLIS = 10000;
```
kristofbolyai, line 25: In the future, I might PR this to be an exponential timeout. So we timeout on 2 second, retry 4 seconds, then 8. That might be better overall.

kristofbolyai (dismissed):
This looks okay. This is obviously harder to review, or perhaps there is nothing to change as this is good as is :). Anyways, I would hold off on merging until @Incompleteusern's review. (I also don't know if this mergable as-is or needs to go by #826)

common/src/main/java/com/wynntils/core/net/ApiResponse.java
magicus: No, it will not be used in #826 either. Qodana is right but wrong. :-) It is not read, that is true, but storing the Future in a static map like this means the Future object will be kept alive not reclaimed by the GC.

It is possible that this is not necessary, and that the Future has some mechanism to keep the instance alive as long as it is not completed, but I am not sure, and I could find no documentation on this. 

If any of you know for certain how this works, I can remove this. But oterhwise, I'd like to keep it for now. If the Future could be GC'ed before all the async processing has happened, then we might get intermittent failures that can be very tricky to track down -- the likelihood that a GC will happen just after creating the Future but before calling the async handling part is very small.

So what I do is that I store the Future in a map, when it is created, and then ask the future to remove itself as it's last operation.

I see now that it could have been made a List or Set instead of Map. I had it a Set originally, but then I actually needed to look up things and made a Map out of it. Then I could get rid of that need, but forgot to change it back to a Set.

kristofbolyai: I am fine with it, just make a tldr comment about this. Even if it is supposed to work, maybe it would be dependant on jvm version/vendor, so it is safer this way

common/src/main/java/com/wynntils/core/net/Download.java
magicus: We will always have a local file, as in a `File` object. (That does not mean that there exists a corresponding file with data content on the disk). This is where the download is/will be cached. The primary difference between Download and ApiResult is that a Download always downloads to a file first, and then reads from the file to handle the content.

If `request` is null, then the NetManager has established that there already exists a file at the location pointed to by the local file `File` object, and that it has a matching md5 sum, so we do not need to download anything. Instead, we can just read the contents of that file directly, and do the async handling/parsing of that file.

magicus: Ah, now I think I understand what you mean. It is the responsibility of the NetManager to determine if the cached file is usable, which it will be deemed to be if we know the md5 sum for that URL, and the md5 sum of the local file matches.

There were some efforts in the old implementation to use the cache "as backup", even without looking at md5, in case of network failures. I don't think it really worked as intended. But the idea is good. I'd like to implement something similar here, but it is non-trivial. You'd have to hook in at the right place to figure out there is a network failure but a local file that might do. And you should probably inform the caller that it is operating on a stale cache.

common/src/main/java/com/wynntils/core/net/NetManager.java
magicus: Maaaybe..? Because aborting and retrying again would be better than waiting for to long? Right now we do not do any reconnection retries. (Neither did the old code). That might be a point.

Overall, we do not handle very well if network downloading is shaky. I think this PR puts the foundation for a more robust approach, but to actually implement proper handling in the face of network errors will require more thorough changes to the entire code base. And serious testing of adverse situations, otherwise the code is likely to break anyway, if it never has been tested.

magicus:
> This is obviously harder to review, or perhaps there is nothing to change as this is good as is :).

I'd like to think that it is good as it is. :-) I have polished, checked and re-checked, and refactored out almost every wrinkle out of the code. I don't think I've ever spent so much time on a thing in Artemis, apart from the initial foundation/architecture building.

(Part of this reason was that we did not get this right from the beginning, but copied in the old Legacy code and started using it everywhere. Had we written this correct from the start it would have been much simpler. )

kristofbolyai:
@magicus can we give both PRs a conventional PR name (refactor perhaps, but this is too big for that?)

kristofbolyai (dismissed):

magicus:
> @magicus can we give both PRs a conventional PR name (refactor perhaps, but this is too big for that?)

I've done so. But these are clear examples of why I think it is a conceptual misunderstanding to equate PRs with changelog items. I know I've been downvoted on this, I'm just a bit grumpy nevertheless. :)

magicus:
Actually, let's call this `feat:`, even though it is a feature more for Artemis developers than an end-facing feature.

common/src/main/java/com/wynntils/core/net/ApiResponse.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.InputStream;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ApiResponse extends NetResult {
+    // By storing the Future here we assure it will not be GC'ed before we are done with it
+    private static final Map<HttpRequest, CompletableFuture<InputStream>> DOWNLOAD_FUTURES = new HashMap<>();
```
Incompleteusern: Is this really necessary?

magicus: As I said above to Bolyai, I don't really know. But I could not find any documentation that guarantees that a Future is kept alive until all async steps are done. Normally, You'd keep the Future around as a way to get the result of the operation, so this would not be a problem, but we are relying on side-effects in the handler, so the Future itself would be lost unless saved.

If I were to remove this code, and it actually were needed, this will work fine 9999 times out of 10000, and the odd failures will be very hard do understand or debug, since the async code will just not have run for no clear reason.

That's why I prefer to play it safe and keep it, unless someone can show me documentation that it is not needed.

magicus: I see it as a future improvement (pun not intended!) to research this better to see if we can get rid of this.

Incompleteusern: Surely future is referenced by `NetManager.HTTP_CLIENT` in `sendAsync`? Haven't checked explicitly, but they must be holding a reference to it to complete it later, which then runs the latter `whenComplete` and `thenApply`?

magicus: Ok, I analyzed the code again. You are correct, at least the `DOWNLOAD_FUTURES` holders in Download and ApiResult can be removed, since the future is returned into `doHandle`. 

I'm still not convinced that the `PROCESS_FUTURES` holder can be removed, though. I have been searching the net, and people say "well, yes, but the actual implementation means that...", but there is no specification guaranteeing this, so I'd rather play it safe.

magicus: It is also worth noting that if the sendAsync() operation should finish so fast that we get to processing the `  .thenAccept(wrappingHandler(onCompletion))` step in `doHandle`, this will actually not be processed asynchronously, but in the same thread. (!) This is unlikely to happen, and unlikely to cause any issues, but in theory it means that we can happen to start doing json parsing on the main thread in certain unlucky circumstances.

To work around this would require leaving sendAsync aside, and doing `send` but in an asynch context that is not kicked off until we have registered the handler. It complicates the code and I don't think it's necessary to protect against, but it can be good to know.

Incompleteusern: > Ok, I analyzed the code again. You are correct, at least the `DOWNLOAD_FUTURES` holders in Download and ApiResult can be removed, since the future is returned into `doHandle`.
> 
> I'm still not convinced that the `PROCESS_FUTURES` holder can be removed, though. I have been searching the net, and people say "well, yes, but the actual implementation means that...", but there is no specification guaranteeing this, so I'd rather play it safe.

```java
     * * CompletableFuture method xStage(...) is called from a public
     *   stage method of CompletableFuture x. It screens user
     *   arguments and invokes and/or creates the stage object.  If
     *   not async and x is already complete, the action is run
     *   immediately.  Otherwise a Completion c is created, pushed to
     *   x's stack (unless done), and started or triggered via
     *   c.tryFire.  This also covers races possible if x completes
     *   while pushing.  Classes with two inputs (for example BiApply)
     *   deal with races across both while pushing actions.  The
     *   second completion is a CoCompletion pointing to the first,
     *   shared so that at most one performs the action.  The
     *   multiple-arity methods allOf and anyOf do this pairwise to
     *   form trees of completions.
```

```java
     * Without precautions, CompletableFutures would be prone to
     * garbage accumulation as chains of Completions build up, each
     * pointing back to its sources. So we null out fields as soon as
     * possible (see especially method Completion.detach). The
     * screening checks needed anyway harmlessly ignore null arguments
     * that may have been obtained during races with threads nulling
     * out fields.  We also try to unlink fired Completions from
     * stacks that might never be popped (see method postFire).
     * Completion fields need not be declared as final or volatile
     * because they are only visible to other threads upon safe
     * publication.
```

These comment implies in `CompletableFuture` that it holds imo

common/src/main/java/com/wynntils/core/net/NetResult.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.Reader;
+import java.net.http.HttpRequest;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+public abstract class NetResult {
+    // By storing the Future here we assure it will not be GC'ed before we are done with it
```
HeyZeer0: This cannot happen at all so you can probably remove this map. The completable future has a reference inside the Executor. At the moment the executor you are using depends on the implementation of CompletableFuture (you are calling `CompletableFuture#supplyAsync`).

Would also advise being careful with the Supply Async as it can have unintended effects on some machines, usually, it will try to use the common Fork/Join pool but if somehow not available it WILL create a thread per request and this can be problematic since it offers:
1. More delay for the request to be executed
2. It will lock the previous thread for a considerate amount of time while a new thread is being created (since it calls the Thread ctor)

HeyZeer0 (commented):
Looks like a stable structure to me! Although one really important missing factor is the ability to add more headers, this is very important as some WynnAPI routes require an API key

common/src/main/java/com/wynntils/core/net/ApiResponse.java
magicus: The last paragraph is about the opposite problem: how do I know that I do not get dangling garbage.

The first paragraph says just one thing of relevance: "Otherwise a Completion c is created, pushed to x's stack (unless done)". So that means that the Future itself holds a reference to the Completion.

But it still does not mean the *Future* will survive if no-one is pointing to it.

kristofbolyai:
> Looks like a stable structure to me! Although one really important missing factor is the ability to add more headers, this is very important as some WynnAPI routes require an API key

Is there any at the moment actually or just future?

common/src/main/java/com/wynntils/core/net/NetResult.java
magicus: Note that getting the Future from CompletableFuture#supplyAsync is just one of the two pathways for getting it, used when the file is already downloaded. For API calls and not-yet-cached files the Future comes from `HTTP_CLIENT.sendAsync`.

I can't find any guarantees about this keeping the Future alive in the documentation. (Just as I still have not found any documentation about the CompletableFuture -- just claims about implementation.) In fact, all official documentation show how the future object is kept alive.

magicus: Do you have any recommendation for replacement for CompletableFuture#supplyAsync?

common/src/main/java/com/wynntils/core/net/ApiResponse.java
Incompleteusern: > The first paragraph says just one thing of relevance: "Otherwise a Completion c is created, pushed to x's stack (unless done)". So that means that the Future itself holds a reference to the Completion.

I think the pushing to stack stores the reference, and this is only replaced on completion, or if a child is pushed instead (which then the child stores a reference).

magicus:
> Is there any at the moment actually or just future?

There is none at the moment. I'm not sure we ever will have; since the design idea is that we "channel" requests to the Wynncraft API through Athena, both to provide caching (the perception is at least that Athena is more stable than Wynncraft API -- this is probably true but I can't tell), and to lower the pressure on Wynncraft servers.

magicus:
There used to be a single wynncraft call from the `/server` command, but I replaced it with an (already existing!) Athena call as part of one of the preparatory patches.

common/src/main/java/com/wynntils/core/net/NetResult.java
HeyZeer0: Gotcha, if that's the case make sure the executor being used by the `HTTP_CLIENT` is not doing the same thing. You could force it to use the Fork/Join pool or just create a fixed thread pool executor (with `Executors#newFixedThreadPool`), the way it works is that it will create all the required threads at the moment the method is called and leave them parking until a request is queued with `Executor#submit`. Tho still it's very unlikely that the fork/join pool will be unavailable, so if you just want to get this done ASAP, can definitely be skipped!

HeyZeer0:
I think on OG Wynntils the only call that required an API Key was the user request `v2/player/ign/stats`, so if you are not going to use it at all, its definitely fine 😄

magicus:
> Although one really important missing factor is the ability to add more headers, this is very important as some WynnAPI routes require an API key

It is trivial to add if we ever need it. Just add a `Map<String, String> headers` argument to `createApiResponse`, and have them added in `create(Post|Get)Request`.

I did not want to build a generic framework that can handle all kinds of http stuff that we do not use nor can test. That was kind of the problem with the old webmanager.

magicus:
> I think on OG Wynntils the only call that required an API Key was the user request `v2/player/ign/stats`, so if you are not going to use it at all, its definitely fine 😄

That's not something we use right now, at least. If we ever start needing it, I think we should seriously consider going via Athena to fetch it anyway, even if all Athena does is proxy it onwards to Wynn. And once again, if we do need the apikey header, it's a trivial addition.

HeyZeer0:
> > I think on OG Wynntils the only call that required an API Key was the user request `v2/player/ign/stats`, so if you are not going to use it at all, its definitely fine 😄
> 
> That's not something we use right now, at least. If we ever start needing it, I think we should seriously consider going via Athena to fetch it anyway, even if all Athena does is proxy it onwards to Wynn. And once again, if we do need the apikey header, it's a trivial addition.

I feel like proxying this over Athena would make us hit our Private API Key rate limit since these are personal requests and not generic ones like the territory list. But since we are not using it, can definitely be ignored for now and implemented later if required.

common/src/main/java/com/wynntils/core/net/ApiResponse.java
magicus: Well, yes, it is stored, but only on x's stack. So if you got no reference to *x* itself, it is lost. Like if I do:
```
void foo() {
  List<Integer> list = new ArrayList();
  Integer i = new Integer.ofString("2");
  list.add(i);
}
```

Sure, while foo is executing, the `i` is kept alive since it is stored in `list`, but as soon as you return out from foo, it can be garbage collected at any time. For that to work, foo would have to return `list` and the caller would need to keep the result of calling foo.

magicus: Also, just to be clear, when they say "x's stack", they are referring to the field named `stack` in the CompletableFuture, not the call stack.

HeyZeer0: As I said below, a scheduled completable future cannot be GCed until it's finished. This is because a reference is always being held inside the executor that is running the task and after the execution finishes the future callback is called.

The java future API is kinda wonky, the way completable futures work is basically by this
```java
CompletableFuture x;
executor.submit(() -> {
    // do stuff to retrieve the results
	x.complete(myResults);
});
```
As you can see on the reference code, X is held at all times inside the runnable, so no GC will ever happen until everything has finished.

magicus: I understand that this is how the implementation works. What I still yet have to see is where this is *specified* behavior. If it is not, it is an implementation detail that might change at any time in the future or (theoretically) between Java implementors.

Incompleteusern: If I understand correctly `executor` is not java implementation code but a task using the `CompletableFuture` api.

Incompleteusern:
```java
        CompletableFuture<Void> future = getInputStreamFuture()
                .thenAccept(wrappingHandler(onCompletion))
                .exceptionally(e -> {
                    // FIXME: Error handling
                    onError.accept(e);
                    return null;
                })
                .whenComplete((ignored, exc) -> PROCESS_FUTURES.remove(request));

        PROCESS_FUTURES.put(request, future);
```

This is the code that uses `PROCESS_FUTURES`. We've established earlier that `getInputStreamFuture`'s returned future can not be gced as it is held by net client. I think`thenAccept` and `exceptionally` are dependent actions on the returned future, which is either immediately ran if called with a completed future, or represented in a series of linked `Completion` object which are held in a Treiber stacks, which then prevents them from being gced.

```
getInputStreamFuture() (CompletableFuture) -> thenApply (Completion) -> exceptionally (Completion) 
```
is an overview of the structure as I see it.

```java
     * Dependent actions are represented by Completion objects linked
     * as Treiber stacks headed by field "stack". There are Completion
     * classes for each kind of action, grouped into single-input
     * (UniCompletion), two-input (BiCompletion), projected
     * (BiCompletions using either (not both) of two inputs), shared
     * (CoCompletion, used by the second of two sources), zero-input
     * source actions, and Signallers that unblock waiters. Class
     * Completion extends ForkJoinTask to enable async execution
     * (adding no space overhead because we exploit its "tag" methods
     * to maintain claims). It is also declared as Runnable to allow
     * usage with arbitrary executors.
```

```java
        CompletableFuture<Void> future = getInputStreamFuture()
                .thenAccept(wrappingHandler(onCompletion))
                .exceptionally(e -> {
                    // FIXME: Error handling
                    onError.accept(e);
                    return null;
                })
                .whenComplete((ignored, exc) -> PROCESS_FUTURES.remove(request));

        PROCESS_FUTURES.put(request, future);
```

This is the code that uses `PROCESS_FUTURES`. We've established earlier that `getInputStreamFuture`'s returned future can not be gced as it is held by net client. I think`thenAccept` and `exceptionally` are dependent actions on the returned future, which is either immediately ran if called with a completed future, or represented in a series of linked `Completion` object which are held in a Treiber stacks, which then prevents them from being gced.

```
getInputStreamFuture() (CompletableFuture) -> thenApply (Completion) -> exceptionally (Completion) 
```
is an overview of the structure as I see it.

```java
     * Dependent actions are represented by Completion objects linked
     * as Treiber stacks headed by field "stack". There are Completion
     * classes for each kind of action, grouped into single-input
     * (UniCompletion), two-input (BiCompletion), projected
     * (BiCompletions using either (not both) of two inputs), shared
     * (CoCompletion, used by the second of two sources), zero-input
     * source actions, and Signallers that unblock waiters. Class
     * Completion extends ForkJoinTask to enable async execution
     * (adding no space overhead because we exploit its "tag" methods
     * to maintain claims). It is also declared as Runnable to allow
     * usage with arbitrary executors.
```

magicus: I actually brought this up with the guys in the core-libs team, which sparked quite a debate. :-) 

So, the end result of the discussion is that yes, we can rely on the future being kept by the commonPool until they have executed, but also that the specs could definitely be better.

magicus: The Treiber stack is just a field in the Future, so it does not survive if the object itself doesn't. But the executor has a reference to the future, until it has run, which is what keeps it alive.

common/src/main/java/com/wynntils/core/net/NetResult.java
magicus: I'd like to do something like that if I should do this really properly. See also https://github.com/Wynntils/Artemis/pull/825#discussion_r1048454216; I would really like to make sure that the "supposedly" async processing happens on another thread, to not bog down the main thread. But I guess it is unlikely to be an actual problem, and not just a theoretical, so I'll do as you say and skip it for now.

magicus:
> I feel like proxying this over Athena would make us hit our Private API Key rate limit since these are personal requests and not generic ones like the territory list.

And what key should we use in the client then to do the calls directly to Wynncraft? 🤔 Do we create API-keys against Wynncraft for each installed instance? The old legacy code just has a hard-coded API key value. I was assuming this was our "private" (but not secretly kept) key for Wynncraft. Would that not just as likely hit the rate limit as the one used by Athena? (Which I hope is different...)

Incompleteusern (approved):

Issue #824, opened 2022-12-13
/wynntils reload needs some rethinking


magicus:
It is unclear what it does and what it should do. After the webapi refactoring, this will become more apparent.

Incompleteusern:
`/wynntils reload` was added in the comparatively early pr of`https://github.com/Wynntils/Artemis/commit/db5f133a424e317f6676048b47ec03d4933228cc by me, which from now I think was a way for the user to signal that setting up an account had failed and should be rerun. This has generalized into the idea of reloading as much as possible about the mod if desired by the user, in the cases of something failing.

kristofbolyai:
Does any feature do any setup that needs reloading? They mostly just respond to events. The only things that do setup is managers, but I am not sure if any of those do anything except resetting

magicus:
I can think of two potentially useful commands:
1) clear caches, and 
2) retry login to Athena+Hades, in case it has failed

I think we need to add a bit of scaffolding to be able to gracefully clear caches and redownload data. But it might be worth adding.

kristofbolyai:
This could also clear the mod update folder since it could be bugged and have problems.

magicus:
So how do we want this to work? Let's focus on splashes, they are simple. Let's say we have downloaded the splashes.json and stored it in the cache directory, and also read and parsed that json and stored the splashes in memory.

Now the user does /wynntils reload.

What should happen?

1. We delete the cached file
2. We try download the file anew from the net
3. We delete/empty the internal java structure (List<String>, in this case)
4. We replace the internal structure when we have parsed a new one
5. We notify users of the structure that it has changed

These are not individual steps, to be taken, one after each other. These are binary questions with a yes/no answer. Like, we can chose to delete the file, then re-download it to the same place, then delete the internal structure, and re-parse it from what we downloaded, and then send an event that it is updated, and fix so all users listen to that event. That is "yes" to all questions.

But we could also skip deleting, just try to download the file anew to a temp location, and if it fails, stick to the one we already have, and only replace it if it succeeds. Similarly, we can keep the internal data, never ever empty it, but if we succeeded in downloading a new file, and succeeded in reading it, we could replace it. User's would then have to rediscover themselves that the data has changed next time they read from it. That would be "yes" to 2 and 4, and no to the rest.

Etc. Basically all 2^5 combinations are possible; though most of them do not make sense.

[...]

PR #826, opened 2022-12-14
fix: NetManager conversions and removal of WebManager

magicus:
This is a PR based on top of https://github.com/Wynntils/Artemis/pull/825. (This targets the `netmanager` branch, not `main`)

I have tried to do just one simple step per commit, and explain what I do. Most are straightforward conversions, some require a bit extra job (but the commit comment explains that).

When both #825 and this has been included in main, the "webapi rewrite" is done.


magicus:
I chose `fix:` instead of `refactor:`, since some actual bug fixes are included.

common/src/main/java/com/wynntils/commands/ServerCommand.java
```diff
@@ -82,8 +82,7 @@ private int serverInfo(CommandContext<CommandSourceStack> context) {
         if (!ServerListModel.forceUpdate(UPDATE_TIME_OUT_MS)) {
             context.getSource()
                     .sendFailure(
-                            new TextComponent("Failed to get server data from API.").withStyle(ChatFormatting.RED));
-            return 1;
+                            new TextComponent("Network problems; using cached data").withStyle(ChatFormatting.RED));
```
kristofbolyai, line 85: I am pretty sure this works, but it is a bit weird that we send a failure, but then, send a "success" as well.

common/src/main/java/com/wynntils/wynn/model/SplashManager.java
```diff
@@ -34,21 +31,16 @@ public static void init() {
...
-                        currentSplash = DEFAULT_SPLASH;
-                    } else {
-                        currentSplash = allSplashes.get(Utils.getRandom().nextInt(allSplashes.size()));
-                    }
-                    return true;
-                })
-                .build());
+        Download dl = NetManager.download(UrlId.DATA_STATIC_SPLASHES);
+        dl.handleReader(reader -> {
+            Type type = new TypeToken<List<String>>() {}.getType();
```
kristofbolyai, line 36: I don't think such "simple" types cannot be handled by GSON without a Type, but this should do.

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -67,114 +57,81 @@ public static List<MapTexture> getMapsForBoundingBox(BoundingBox box) {
...
+                                NativeImage nativeImage = NativeImage.read(inputStream);
+                                MapTexture mapPartImage = new MapTexture(
+                                        fileName, nativeImage, mapPart.x1, mapPart.z1, mapPart.x2, mapPart.z2);
+                                MAPS.add(mapPartImage);
+                            } catch (IOException e) {
+                                WynntilsMod.info("IOException occurred while loading map image of " + mapPart.name);
+                            }
+                        },
+                        onError -> {
+                            WynntilsMod.info("IOException occurred while loading map image of " + mapPart.name);
```
kristofbolyai: Was this really an IOException? Can we use a different logging message here, so we know where we failed? Also, perhaps log the error itself?

common/src/main/java/com/wynntils/wynn/model/map/MapModel.java
```diff
@@ -67,114 +57,81 @@ public static List<MapTexture> getMapsForBoundingBox(BoundingBox box) {
...
+                Download dlPart = NetManager.download(URI.create(mapPart.url), "maps/" + fileName, mapPart.md5);
+                dlPart.handleInputStream(
+                        inputStream -> {
+                            try {
+                                NativeImage nativeImage = NativeImage.read(inputStream);
+                                MapTexture mapPartImage = new MapTexture(
+                                        fileName, nativeImage, mapPart.x1, mapPart.z1, mapPart.x2, mapPart.z2);
+                                MAPS.add(mapPartImage);
+                            } catch (IOException e) {
+                                WynntilsMod.info("IOException occurred while loading map image of " + mapPart.name);
```
kristofbolyai: Is it worth to log the exception here?

kristofbolyai: (I mean logging IOException e)

common/src/main/java/com/wynntils/core/net/athena/WynntilsAccountManager.java
```diff
@@ -86,55 +84,45 @@ public void dumpEncodedConfig(String name) {
...
+                        .entrySet()
+                        .forEach((k) ->
+                                encodedConfigs.put(k.getKey(), k.getValue().getAsString()));
+                loggedIn = true;
+                future.complete(true);
+                WynntilsMod.info("Successfully connected to Athena!");
+            });
+        });
+
+        // FIXME: Rewrite athena account manager dependencies to work async
```
kristofbolyai, line 116: Didn't the commit description just say this is now up to date? Or what is this about?

kristofbolyai: Is this about the users of the token does not respect loading this is async, hence making the code weird?

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -79,9 +82,14 @@ private int reload(CommandContext<CommandSourceStack> context) {
...
-        WebManager.reset();
+        // Try to reload data from the web. Likely to be broken, see
+        // https://github.com/Wynntils/Artemis/issues/824
         ItemProfilesManager.reset();
-        WebManager.init(); // reloads api urls as well as web manager
+        // reloads api urls as well as web manager
+        UrlManager.reloadUrls();
+        ItemProfilesManager.init();
+        SplashManager.init();
+        WynntilsAccountManager.init();
```
kristofbolyai: Shouldn't we reset the other managers? (I know this is bad as a whole, so maybe we can just ignore this for the time being)

kristofbolyai (commented):
I have checked every commit now. I've left comments everywhere related. Once those are resolved, I am happy to approve this.

common/src/main/java/com/wynntils/commands/ServerCommand.java
magicus: Yes, it's kind of weird. Sounds almost as weird to call `sendSuccess` as well. I looked at the implementation, both methods do more or less the same. 

This was not really part of the real fix, just something I noticed while rummaging about in the ServerCommand source and thought was an, unrelated, improvement. I can leave it out for now as well, too, and we can revisit this separately and see if there is some better way (maybe a third `sendUpdate` method?) to do this.

kristofbolyai: This is fine. No need to revert it.

common/src/main/java/com/wynntils/wynn/model/SplashManager.java
magicus: What do you mean? That I should use the GSON.fromJson(reader, Class<?> class) version instead? It clearly says "It is not suitable to use if the specified class is a generic type since it will not have the generic type information because of the Type Erasure feature of Java. Therefore, this method should not be used if the desired type is a generic type. "

kristofbolyai: I remember a GSON method without a Type/Class. Anyways, I think I miss remember. I've been debugging matrixes for an hour before writing this 🙃.

common/src/main/java/com/wynntils/core/net/athena/WynntilsAccountManager.java
magicus: > Is this about the users of the token does not respect loading this is async, hence making the code weird?

Yes, exactly. Code like Hades is assuming that when the HadesManager or whatever it was called is being init'ed, that the login is done so we have a token. If not, it complains loudly and refuses to make any further attempts at connecting.

I think we should rewrite this so that we send out like a AthenaLoginEvent when we are authenticated with Athena, and that in turn could trigger Hades to log in, etc.

But that was not something I wanted to address here.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
magicus: So, to be clear, I have copied as good as humanely possible what the old WebManager.reset() did. It held on to a few, but not all, of the "profile" data. These are the ones I have called init on.

magicus: I thought I was clear enough that the entire thing was a mess that needs to be addressed by https://github.com/Wynntils/Artemis/issues/824 :)

common/src/main/java/com/wynntils/wynn/model/SplashManager.java
magicus: No, there is not. But if you don't have generics, you can do like `GSON.fromJson(reader, FluffyProfile.class)`.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
kristofbolyai: Yes, this is bad. Just one question. Do you have plans to address this yourself, or should I be looking into it sometime?

common/src/main/java/com/wynntils/core/net/athena/WynntilsAccountManager.java
kristofbolyai: Agreed. I think we should have an issue for this though.

kristofbolyai: https://github.com/Wynntils/Artemis/issues/829

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
magicus: I can look into it. I think the entire code base could do with another sweep, which looks more closely at error handling and exceptional conditions.

However.

I'm currently quite fed up with this, I've been working on this for two weeks (not counting the time I was contemplating the design prior to that) and I'd like to take a break. So if you think it needs to be fixed "soon-ish", maybe you should to it.

kristofbolyai: Okay. That is understandable. We will see who gets to this first. I will let you know, if I start working on this (and it would be great if you could do this for me too).

magicus: Sure, I'll let you know. (In general, it is probably a good idea to assign a bug to yourself, and also perhaps add a comment, if you intend to start solving it)

kristofbolyai (approved):
Looks great to me! I'm excited about this. (I would say this PR does not need explicit approval from someone else, as it is quite trivial, and is just about porting to use the new handlers). But you decide.

magicus:
I'm also eager to have this go into main, so I think I'm satisfied with your approval only.

PR #832, opened 2022-12-15
feat: Add POI fading and POI visibility configs

kristofbolyai:
This PR semi-breaks custom poi backwards compatiblity. They will work, but old saves visiblity's will be set to default, always.


kristofbolyai:
https://user-images.githubusercontent.com/49001742/207826766-8d780066-ba0b-4383-8cf1-fed0f674ab60.mov

magicus:
Ohh, that looks epic! Lemme check the code as well.

common/src/main/java/com/wynntils/wynn/model/map/poi/CustomPoi.java
```diff
@@ -39,8 +55,10 @@ public CustomColor getColor() {
         return color;
     }
 
-    public float getMinZoom() {
-        return minZoom;
+    public Visibility getVisibility() {
+        // Visibility is null for configs that have the old format (minZoom float)
```
magicus, line 59: Are we starting to add code to handle backwards compatibility with user's configs..? Hm. I realize that an outright crash might be bad, but the point of Alpha is, among other things, to break the config files.

Going forward, I think we will need to handle the fact that configurations changes. Maybe we need to add a version to config files, and if it is too old, make a "upgrade processor" that pre-processes the file before reading.

What I *don't* want is to have the code cluttered with "oh and this looks weird but we must support some old-style version of the config". :(

I'll accept this for now, if this will cause a NPE for all users with custom POIs, and there is no other good way to remove such "old-style" NPEs. (Yeah, I realize that is hard, but hey, it's still Alpha). But we need to find a good way to handle this going forward. Adding a version to the config file is probably the first step, and bumping it whenever we make non-backward compatible changes.

kristofbolyai: We could deisgn a system for this. I am up for coding it. And yes, the NPE was the cause

magicus: Actually, is there any way we are ever going to be able to remove this code, unless we do the config-version stuff first? Hm. If not, I don't think this is acceptable. Or is it just a matter of the users loading the config once and then saving again, and it will be fine? If so, we can hope everyone has converted in a couple of month or so, and then remove it.

But if I misunderstand things and it will not work like that, then we need to fix config versioning first.

kristofbolyai: No, it won't be fixed after a run. We need to leave this though, just until I can work on config updaters

kristofbolyai: @magicus Are you fine with this if I commit to making a config "upfixer"? We probably should have one anyway for beta/release

magicus: Yes, definitely. As long as we can remove this when we get the upfixer. But yeah, I assume that would be possible.

magicus (approved):

PR #833, opened 2022-12-15
refactor: Fully convert Managers to singletons

magicus:
This is the second and final part of refactoring the managers to singletons. This is dependent on https://github.com/Wynntils/Artemis/pull/830, and the PR targets a branch which has those changes included, so it is possible to review.

The premise for this PR is the operation "make all static methods in managers non-static", and then dealing with the fall-out.

One of the major changes as a result of this is that "ManagerRegistry" is now a pure "ModelRegistry", and also that "Model" and "Manager" share no common superclass. (The old CoreManager was renamed to just Manager, and the old superclass with the same name were removed).

I think this adds a ton of clarity to the code.

I have also fixed a couple of bugs that turned up when doing this churn.


magicus:
There are two issues that I have left unresolved for this patch. One is that Functions with model dependencies do not really fly, since then we start to init models while in the midst of initing managers. The other is that NetManager and UrlManager has a circular dependency. 

I have hacked around this and will come back to this later.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -27,10 +28,14 @@
 public class WynntilsCommand extends CommandBase {
     @Override
     public void register(CommandDispatcher<CommandSourceStack> dispatcher) {
+        throw new UnsupportedOperationException("WynntilsCommand need special treatment");
```
kristofbolyai, line 31: Uh oh

kristofbolyai: What is it that we need to do?

kristofbolyai:
What is this whole managers can depend on managers thing?

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
magicus: We need to have a list of all other commands, since we want to create aliases, like `/wynntils token` for `/token`.

Personally I am not sure I like this, but that is how we currently do it.

common/src/main/java/com/wynntils/core/managers/Manager.java
```diff
@@ -4,15 +4,20 @@
...
- * <p>
- * See {@link Model} and {@link CoreManager} for the different manager types.
- * */
-public abstract class Manager {}
+ * Managers constitute the core functionality of Wynntils. They are created
+ * as singletons in the {@link Managers} holding class.
+ */
+public abstract class Manager {
+    protected Manager(List<Manager> dependencies) {
+        // dependencies are technically not used, but only required
```
kristofbolyai, line 16: Oh. I see. Well, this is kinda weird isn't it?

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
magicus: So this was just working out of pure luck before, since we sorted `WynntilsCommand` last...

magicus: Also, I think we could have like `/wynntils wynntils wynntils wynntils token` before as well, but I have not checked.

kristofbolyai: > Also, I think we could have like `/wynntils wynntils wynntils wynntils token` before as well, but I have not checked.

I think that was handled.

kristofbolyai: Okay, so what do we need to fix this?

kristofbolyai: (I want to keep the option to prefix with /wynntils)

magicus:
> What is this whole managers can depend on managers thing?

That is a way to tackle this properly and safely:
```
        // Bootstrapping order is important, take care if reordering
```

Previously it was just a bit hand-wavy in which order Managers were bootstrapped. Now it is explicit.

kristofbolyai (commented):
Seems fine after we fix WynntilsCommand. At first sight, I first thought we convert models to a singleton pattern too, but that shouldn't be the case.

common/src/main/java/com/wynntils/core/managers/Manager.java
magicus: I am using Java:s built-in syntax check to make sure we get bootstrapping in the correct order.

If we really wanted, we could of course set up a rule that no Manager might read Managers.<mgr>, and only access the managers they need from the one they get through the constructor. However, I think it just adds boilerplate, and make accesses to other Managers from within a Manager looking quite different from the rest of the code.

magicus:
I think we maybe should convert Models as well, but lets do that separately! I need to look a bit more into how I think the upcoming model refactoring will be. But I think having singletons is a prerequisite. 

But, in any case, PRs like these get quite messy, so I think we should focus on Managers here.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
kristofbolyai: > So this was just working out of pure luck before, since we sorted `WynntilsCommand` last...

It was intentional, just not noted. Can't we still do this?

magicus: I have not removed the functionality to do `/wynntils token`. But to get this to work I send the set of previously registered commands as an explicit argument to the register method in WynntilsCommand, so it does not try to go and query Managers.Command, which is not yet set up.

I could have changed the signature for all commands to get the set of hitherto registered commands, but I thought that was just ... odd.

So everything will work just fine, and we make it explicit that `/wynntils` is a bit special since it will proxy to all other commands. I think this is just fine, and a bit less magic than previously.

kristofbolyai (approved):
Looks good! I hope we didn't miss something.

PR #834, opened 2022-12-15
feat: Add config upfixers

kristofbolyai:


common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -50,9 +52,16 @@ public final class ConfigManager extends Manager {
...
-    public ConfigManager() {
-        super(List.of());
+    public ConfigManager(ConfigUpfixerManager upfixer) {
+        super(List.of(upfixer));
+
+        // First, we load the config file
         loadConfigFile();
+
+        // Now, we have to apply upfixers, before any config loading happens
+        if (upfixer.runUpfixers(configObject)) {
```
Incompleteusern, line 62: make the upfixer run lazily by storing a version in the config and comparing differences.

P0keDev:
are upfixers meant for converting config data between versions when necessary? i'm a bit confused about their purpose

magicus:
Yes, exactly. When the config data format changes in a way that would break old config files, this is needed. Happened recently with the zoom fading custom POIs.

kristofbolyai:
> are upfixers meant for converting config data between versions when necessary? i'm a bit confused about their purpose

Yes. In beta/stable builds, we want to be able to have a way to conveniently handle cases like these. From a simple config rename to a structural change, we can handle almost everything this way.

common/src/main/java/com/wynntils/core/config/upfixers/ConfigUpfixerManager.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.config.upfixers.impl.CustomPoiVisbilityUpfixer;
+import com.wynntils.core.managers.Manager;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ConfigUpfixerManager extends Manager {
+    public final String UPFIXER_JSON_MEMBER_NAME = "upfixers";
```
magicus: Feels like a bit short as a json name. All our other json config keys are like `feature>.<something>.<some more>` or suchlike. Maybe `config.upfixers`, or `wynntils.upfixers` to sort of "claim" a namespace for more general stuff related to the mod/config file as a whole.

common/src/main/java/com/wynntils/core/config/upfixers/ConfigUpfixerManager.java
```diff
@@ -0,0 +1,88 @@
...
+            appliedUpfixers.add(upfixer.getAsString());
+        }
+
+        List<ConfigUpfixer> missingUpfixers = new ArrayList<>();
+
+        for (ConfigUpfixer configUpfixer : configUpfixers) {
+            if (!appliedUpfixers.contains(configUpfixer.getUpfixerName())) {
+                missingUpfixers.add(configUpfixer);
+            }
+        }
```
magicus: I can't help but think this can be made more compact with streams. Not necessarily better, mark my words, but at least less verbose. :-) Is there not some kind of intersection operation you can use? Your goal is like "get me a list of all upfixers that is not contained in the list of applied upfixers". Sounds like `filter()` to me...

magicus (dismissed):
This was a very sweet design! The field where you saved applied upfixers is more or less what I wanted with a version field, but more flexible. Nice solution.

This looks basically OK, but see if you can streamify `getMissingUpfixers` a bit, and change the json key name. Then I'm happy!

magicus (approved):

PR #835, opened 2022-12-15
refactor: Access models through Models instance, not directly through class name

magicus:
Just as for the Manager refactoring, this is the first step towards making Models a singleton.

This too is a massive manual search+replace + verify that only method calls where affected. Tedious to do, but trivial to verify by just skimming the PR.

I also fixed a couple of bugs with incorrect dependencies (see individual commits), and a case where a local variable where shadowing a static field, which where causing issues.

Finally I also made all Model classes final.


kristofbolyai (approved):
Looks fine

kristofbolyai:
> and a case where a local variable where shadowing a static field, which where causing issues.

Where is this?

magicus:
> > and a case where a local variable where shadowing a static field, which where causing issues.
> 
> Where is this?

LootrunModel.tryLoadFile(). That meant the static field `lootrun` had to be referenced by prefixing it with the class, something I wanted to avoid.

PR #836, opened 2022-12-15
Managers should not be registered as listener until after construction

magicus:
I knew this was a bad idea but did it anyway. Shame on me.

### DIFF
--- a/common/src/main/java/com/wynntils/core/managers/Managers.java
+++ b/common/src/main/java/com/wynntils/core/managers/Managers.java
@@ -4,6 +4,7 @@
  */
 package com.wynntils.core.managers;
 
+import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.commands.ClientCommandManager;
 import com.wynntils.core.config.ConfigManager;
 import com.wynntils.core.features.overlays.OverlayManager;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/core/managers/Managers.java
```diff
@@ -52,6 +54,17 @@ public final class Managers {
     public static final DiscoveryManager Discovery = new DiscoveryManager(Net, Territory, MinecraftScheduler);
 
     public static void init() {
-        // We don't need to do anything here, but make sure the class is loaded
+        // Register all manager singletons as event listeners
+
+        FieldUtils.getAllFieldsList(Managers.class).stream()
+                .filter(field -> Manager.class.isAssignableFrom(field.getType()))
```
kristofbolyai, line 60: we would do `FieldUtils.readField(field, this, true)` in other parts of the mod. I am not sure which is better

kristofbolyai: Wrong place to comment! Talking about `                        WynntilsMod.registerEventListener(field.get(null));` **field.get(null)**

kristofbolyai: FieldUtils is basically a wrapper. Okay. But it technically does field.get(this) instead of field.get(null)

magicus: But we are reading a static field so it is prudent to send in null to indicate this..?

kristofbolyai: oh right

kristofbolyai (approved):

PR #837, opened 2022-12-15
fix: Resolve the UrlManager/NetManager circularity

magicus:
This was introduced with the NetManager rewrite, and meant that the UrlManager did not download new URLs from the net as early as it should


kristofbolyai (approved):
Looks fine

PR #838, opened 2022-12-15
refactor: Make Models into singletons

magicus:
Similar to what I did with managers. This time around it was less painful, either since I learned a method, or since models were more structured to begin with.


P0keDev (dismissed):
i was thinking of doing this while making all the custom item models, since it would've allowed a much cleaner solution for those - i'm glad it got done eventually!

P0keDev (approved):

PR #844, opened 2022-12-16
refactor: Flatten item and property models into single files

P0keDev:
Now that models no longer have to be static, we can avoid the nasty code duplication that these models previously had using the `ItemStackModel` and `PropertyModel` "template" classes.


magicus (approved):
Yay!

Issue #845, opened 2022-12-16
NoClassDefFoundError: Could not initialize class com.wynntils.core.managers.Managers Preventing Game Launch


san7890:
Using Fabric. Found by compiling 246b9eed87e008ff8cb735a61b6c3de383c0a682 (most recent commit at the time of this issue). Appears to be a different bug than what that commit fixes (alpha version 170 crashes for a different reason, presumably because it's missing the code introduced in the aforementioned commit).

![image](https://user-images.githubusercontent.com/34697715/208035933-74ce0a30-a274-45a8-b7fa-f289cfe2b68f.png)

```txt
NoClassDefFoundError: Could not initialize class com.wynntils.core.managers.Managers

at com.wynntils.core.mod.CrashReportManager.generateDetails(CrashReportManager.java:35)
at net.minecraft.class_128.handler$bml000$addWynntilsDetails(class_128.java:525)
at net.minecraft.class_128.method_555(class_128.java:77)
at net.minecraft.class_128.method_568(class_128.java:132)
at net.minecraft.class_310.method_1565(class_310.java:842)
at net.minecraft.client.main.Main.main(Main.java:207)
at net.fabricmc.loader.impl.game.minecraft.MinecraftGameProvider.launch(MinecraftGameProvider.java:461)
at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:74)
at net.fabricmc.loader.impl.launch.knot.KnotClient.main(KnotClient.java:23)
```

https://github.com/Wynntils/Artemis/blob/246b9eed87e008ff8cb735a61b6c3de383c0a682/common/src/main/java/com/wynntils/core/mod/CrashReportManager.java#L30-L42

magicus:
@san7890 Can you please confirm that this was solved by #846.

san7890:
Resolved, thank you!

PR #846, opened 2022-12-16
fix: Remove all references to Managers.CrashReport in CrashReportManager

magicus:
Thanks @P0keDev for fixing the urgent crash. The reason for the crash is that Minecraft does `CrashReport.preload()` very early on in the bootstrapping, to "oil the machinery" for generating crash reports. This ended up calling our hook long before `Managers` where loaded.

P0ke's patch fixed the crash from `preload`, but there still were a reference to `Managers.CrashReport` shortly after, which could have NPE'd if we got an actual crash after the mc instance was non-null, but Managers.CrashReport was not available.

Instead, keep local track of our instance, and use it.

### DIFF
--- a/common/src/main/java/com/wynntils/core/mod/CrashReportManager.java
+++ b/common/src/main/java/com/wynntils/core/mod/CrashReportManager.java
@@ -5,8 +5,6 @@
 package com.wynntils.core.mod;
 
 import com.wynntils.core.managers.Manager;
-import com.wynntils.core.managers.Managers;
-import com.wynntils.mc.utils.McUtils;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


magicus:
Should fix #845.

P0keDev (approved):

P0keDev:
thanks for figuring this one out fully! i *was* able to launch the game without crashing when i tested my PR, but perhaps that was just luck? either way, this is a much more comprehensive solution

magicus:
Your version will work just fine in reality. The only thing that would cause it to fail if it were an actual crash after `preload` but before our `Managers` class is setup. Then we would NPE in the Crash Handler, which is not nice, but otoh, there must have been some external cause for the crash then. This just makes the crash report looks good in that case.

P0keDev:
> Your version will work just fine in reality. The only thing that would cause it to fail if it were an actual crash after `preload` but before our `Managers` class is setup. Then we would NPE in the Crash Handler, which is not nice, but otoh, there must have been some external cause for the crash then. This just makes the crash report looks good in that case.

ah, i've just realized that my patch didn't get built into a release because squashing a single-commit PR uses the name of that commit by default (instead of the name of the PR itself), and so the commit pushed to main branch didn't have the right format. the more you know!

magicus:
Oh, yeah, that's kind of a bug in github. :-( Annoying. One just has to learn to double-check the commit message when integrating.

I've run across that some time before, I didn't just connect the dots to realize it was because it was a single-commit PR.

PR #847, opened 2022-12-16
fix: Bug fix for multi-line objective patterns

magicus:
And also clarify the code a bit.

The old pattern for multi-line start would incidentally also match the segment header, leading to repeated error output.


kristofbolyai (dismissed):
Aaah. I wondered why it behaved weirdly... Thanks

PR #848, opened 2022-12-16
feat: Snappier and more robust quest book

magicus:
There are a lot of fixes here. On the top of my mind:
* Mini-Quests scoreboard where not properly updated since they could not be found; the name started with "Mini-Quest - " from the scoreboard but we stripped that away in the miniQuests list of QuestInfo, so no match was ever found.
* If a quest was tracked without the quest book ever read by us, scoreboard updates would fail. This could happen if you used the vanilla Wynncraft quest book to start tracking a quest, instead of our own Quest Book Screen.
* We tried to track finished quests from the GUI which did not work
* Quest Book Screen now starts showing the cached copy of quests, and triggers a rescan in the background. Quest list is updated once the rescan completes.
* QuestManager has now the sole responsibility of knowing which quest is tracked. This is interpreted as "showing up on the score board".
* The Quest Book Screen now has an "intermediate" state for quests, to signal that we have requested tracking but not yet got confirmation from Wynncraft, shown in a orange hue. Once the tracking is confirmed, it reverts to the normal green for tracked quests.
* Switching between quests and miniquests is snappy, and never loses track of which quest was selected.
* When choosing to replace the vanilla quest book, the main menu was opened instead of the quest book.
* I have now consistently called the tracked quest "tracked", instead of "pinned" or "current".
* When untracking a quest, do as vanilla Wynncraft and use the `/tracking` command. That's one container query less, at least.

Overall, I have tested all manner of combinations of using our GUI and using the Wynncraft vanilla quest book, in different order, before or after we have our quest book query run, using normal quests and mini-quests. I think it works pretty fine right now.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
+++ b/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
@@ -126,28 +126,24 @@ public class QuestInfo {
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
-
-        if (o == null || getClass() != o.getClass()) return false;
-
-        QuestInfo questInfo = (QuestInfo) o;
+        if (o == null || (!(o instanceof QuestInfo questInfo))) return false;
 
         // Consider it the same quest just based on "constant" aspects of the quest
-        return new EqualsBuilder()
-                .append(level, questInfo.level)
-                .append(isMiniQuest, questInfo.isMiniQuest)
-                .append(name, questInfo.name)
-                .append(length, questInfo.length)
-                .isEquals();
+        if (level != questInfo.level) return false;
+        if (isMiniQuest != questInfo.isMiniQuest) return false;
+        if (!name.equals(questInfo.name)) return false;
+        if (length != questInfo.length) return false;
+        return additionalRequirements.equals(questInfo.additionalRequirements);
     }
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder(17, 37)
-                .append(name)
-                .append(length)
-                .append(level)
-                .append(isMiniQuest)
-                .toHashCode();
+        int result = name.hashCode();
+        result = 31 * result + length.hashCode();
+        result = 31 * result + level;
+        result = 31 * result + additionalRequirements.hashCode();
+        result = 31 * result + (isMiniQuest ? 1 : 0);
+        return result;
     }
 
     @Override
--- a/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
+++ b/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
@@ -132,12 +132,22 @@ public class QuestInfo {
         QuestInfo questInfo = (QuestInfo) o;
 
         // Consider it the same quest just based on "constant" aspects of the quest
-        return new EqualsBuilder().append(level, questInfo.level).append(isMiniQuest, questInfo.isMiniQuest).append(name, questInfo.name).append(length, questInfo.length).isEquals();
+        return new EqualsBuilder()
+                .append(level, questInfo.level)
+                .append(isMiniQuest, questInfo.isMiniQuest)
+                .append(name, questInfo.name)
+                .append(length, questInfo.length)
+                .isEquals();
     }
 
     @Override
     public int hashCode() {
-        return new HashCodeBuilder(17, 37).append(name).append(length).append(level).append(isMiniQuest).toHashCode();
+        return new HashCodeBuilder(17, 37)
+                .append(name)
+                .append(length)
+                .append(level)
+                .append(isMiniQuest)
+                .toHashCode();
     }
 
     @Override
--- a/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
+++ b/common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
@@ -13,6 +13,7 @@ import com.wynntils.wynn.objects.profiles.ingredient.ProfessionType;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
```diff
@@ -121,6 +123,33 @@ public boolean isMiniQuest() {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+
+        if (o == null || getClass() != o.getClass()) return false;
+
+        QuestInfo questInfo = (QuestInfo) o;
+
+        // Consider it the same quest just based on "constant" aspects of the quest
+        return new EqualsBuilder()
```
kristofbolyai: In every other hashcode/equals, I opted to use the regular java. Either let's make a PR to switch everything up, or just make this standard and not use the builders here too?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -133,21 +144,96 @@ public Location getCurrentQuestLocation() {
...
+        updateTrackedQuest(null);
+    }
+
+    public void updateTrackedQuestFromScoreboard(String name, String nextTask) {
+        // If our quest book has not yet been scanned, we can't update now
+        // but will do after scanning is complete
+        if (updateAfterRescan(name, nextTask)) return;
+
+        Optional<QuestInfo> questInfoOpt;
+        questInfoOpt = getQuestInfoFromName(name);
```
kristofbolyai: why spli declaration?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -133,21 +144,96 @@ public Location getCurrentQuestLocation() {
...
+        } else {
+            questInfoOpt = Managers.Quest.quests.stream()
+                    .filter(quest -> quest.getName().equals(name))
+                    .findFirst();
+        }
+        return questInfoOpt;
+    }
+
+    private boolean updateAfterRescan(String name, String nextTask) {
+        if (name.startsWith("Mini-Quest - ")) {
```
kristofbolyai: Qodana perhaps right, extract common parts?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -133,21 +144,96 @@ public Location getCurrentQuestLocation() {
...
+                afterRescanName = name;
+                rescanQuestBook(true, false);
+                return true;
+            }
+            return false;
+        }
     }
 
-    protected void setQuests(List<QuestInfo> newQuests) {
+    protected void updateQuestsFromQuery(List<QuestInfo> newQuests, QuestInfo trackedQuest) {
```
kristofbolyai, line 200: private?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -133,21 +144,96 @@ public Location getCurrentQuestLocation() {
...
         quests = newQuests;
+        maybeUpdateTrackedQuest(trackedQuest);
         WynntilsMod.postEvent(new QuestBookReloadedEvent.QuestsReloaded());
     }
 
-    protected void setMiniQuests(List<QuestInfo> newMiniQuests) {
+    protected void updateMiniQuestsFromQuery(List<QuestInfo> newMiniQuests, QuestInfo trackedQuest) {
```
kristofbolyai, line 206: private?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Removed as much duplication as I could.

Long term I think we should not consider quests and mini-quests to be separate in this way.

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: It could be "package private", that is, without any qualifier. I just hate that syntax.

magicus: (but not private, since it is used from QuestContainerQueries)

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
magicus: Fixed.

Let's skip builders.

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Because stupid? :-)

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Are you happy with keeping it protected (which, to be clear, is a wider accessibility than default/package private)? I can remove it if you really want the lowest possible visibility. But as I said, I think that syntax is problematic. (A language design wart of java, if you ask me)

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
magicus: Same as above. I do whatever you decide on both places.

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
kristofbolyai: I guess its fine as protected. Idm either way

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
```diff
@@ -121,6 +121,29 @@ public boolean isMiniQuest() {
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || (!(o instanceof QuestInfo questInfo))) return false;
+
+        // Consider it the same quest just based on "constant" aspects of the quest
+        if (level != questInfo.level) return false;
+        if (isMiniQuest != questInfo.isMiniQuest) return false;
+        if (!name.equals(questInfo.name)) return false;
+        if (length != questInfo.length) return false;
+        return additionalRequirements.equals(questInfo.additionalRequirements);
```
kristofbolyai: Sorry for the trouble, but this still not the same as other code parts. I may have mislead you though.

This is the format I use.
image

kristofbolyai: @magicus Fixed it for you. Once you check this out, we can merge.

kristofbolyai (approved):

common/src/main/java/com/wynntils/wynn/model/quests/QuestInfo.java
magicus: Thanks. (I used the IntelliJ default, but tweaked it to allow for the new pattern-matching instanceof)

I don't know how much we need to be consistent on this sorts of things, but I guess it does not hurt. What is more important however is to be sure to actually *do* implement equals and hashCode, where needed.

PR #849, opened 2022-12-16
fix: Crash in Forge at startup

magicus:
The manager refactoring caused a crash at startup for forge, due to models being registered while setting up managers (via functions). This ugly thing is now completely fixed.


kristofbolyai (approved):
Looks good

PR #852, opened 2022-12-16
feat: Redirects "Soul Point Lost" messages to Overlay

san7890:
image
image

Another small QoL thing that extends the soulPoint configurable action.

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
@@ -88,6 +88,7 @@ public class ChatRedirectFeature extends UserFeature {
         register(new PotionsMovedRedirector());
         register(new PotionsReplacedRedirector());
         register(new ScrollTeleportationFailRedirector());
+        register(new ScrollTeleportationSoulPointRedirector());
         register(new SoulPointDiscarder());
         register(new SoulPointRedirector());
         register(new SpeedBoostRedirector());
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -587,6 +588,25 @@ protected String getNotification(Matcher matcher) {
         }
     }
 
+    private class ScrollTeleportationSoulPointRedirector extends SimpleRedirector {
```
kristofbolyai: Can't we do this for x soul points? Like when we die, isn't there a 2 soul points lost message?

That would also mean renaming the redirector, adding a capture group to the regex, and adding a new config option

san7890: image

It's a completely different message (actually two messages) with different chatformatting codes. I can do that in another PR. You could also genericize the regex to account for either the dark_red/gray codes (also the has/have distinction)... but I think I'd rather that just be its own separate deal.

kristofbolyai: The soul point part is almost the same? You can use * for the color code, and s? to fix the pluralization.

You can leave the you have died part in the chat

common/src/main/java/com/wynntils/features/user/redirects/ChatRedirectFeature.java
```diff
@@ -641,6 +642,35 @@ protected String getNotification(Matcher matcher) {
         }
     }
 
+    private class SoulPointLossRedirector extends SimpleRedirector {
+        private static final Pattern SYSTEM_PATTERN =
+                Pattern.compile("^§[47](\\d+) soul points? (has|have) been lost...");
```
kristofbolyai:
```suggestion
                Pattern.compile("^§[47](\\d+) soul points? (has|have) been lost...$");
```

kristofbolyai (approved):

PR #857, opened 2022-12-16
chore: Bumps actions/setup-java to v3 for Qodana CI

san7890:
image

GitHub CI Runners are doing a lot of deprecation lately, which is easily resolved by bumping the actions version to v3. Just to make sure, I ran a test case of this on my fork at https://github.com/san7890/Artemis-bruh/actions/runs/3716411603 and it seems to work alright (in comparison to the errors at https://github.com/Wynntils/Artemis/actions/runs/3716329395).

I believe the new error is due to how my repository is set up? Unsure, but looking at this repository's caches at https://github.com/Wynntils/Artemis/actions/caches it seems we aren't really using any cache (takes 10 minutes to build for qodana, and we generate a new cache every time). Might have to look at that later.

### DIFF
--- a/.github/workflows/qodana.yml
+++ b/.github/workflows/qodana.yml
@@ -16,7 +16,7 @@ jobs:
         run: |
           cp ./.idea/misc-example.xml ./.idea/misc.xml
       - name: Set up JDK
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: "temurin"
           java-version: 17


magicus (approved):

magicus:
> Might have to look at that later.

Please do. I added caching of the decompiled minecraft source code, which cut build times in half, but there is certainly more to gain. But 10 minutes sounds extreme, that's not what I usually see.

magicus:
This run took like 2 minutes, so there is no regression caused by this change at any rate.

san7890:
> > Might have to look at that later.
> 
> Please do. I added caching of the decompiled minecraft source code, which cut build times in half, but there is certainly more to gain. But 10 minutes sounds extreme, that's not what I usually see.

I don't know so much about Qodana but that's how long it took on both runs actions linked above (even with the same exact code)

image

image

Anecdotally, I witnessed one of them take 10 minutes to setup, and 3 minutes for the actual analyzation to occur.

magicus:
I see. Your run tried to download half the world, for some reason. That's likely the explanation for the extra time.

This is not what I see on the run for this PR. Weird. I have no idea.

san7890:
@magicus 

I believe I know what the issue might be/is. The cache will only apply on a specific pull request's branch, so it'll take the ~15 minutes or so on the first commit. However, any subsequent commits to that branch (merging main, doing a code review) will be able to use that Qodana cache. For another repository I've worked on, we were able to pull the cache from the master/main branch since we ran CI on every single commit to main, but maybe the reason that doesn't work here is because we don't have a cache on main for PRs to use? I'm not sure on that, but I'm just sharing my thoughts/experience here.

I think https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache has the most information on this... maybe building Qodana on main during regular CI would help get over that "first commit" speedbump? Interested to hear your thoughts on this.

magicus:
You are welcome to experiment. I don't really have the stamina to go deep diving into GHA caches once more. If you think adding another Qodana task on main would help, please submit a PR to do that. If it turns out not to help and we just think it wastes resource, we can revert it later.

PR #858, opened 2022-12-16
fix: CrashReportManager did not actually produce a crash report

magicus:
While the previous fix did prevent the crash at mod startup, it also meant that we did not get proper hooks in the crash report. Basically revert to the old static version.


kristofbolyai:
How do we know this works? Was this the case where poke fixed it but the commit did not get built, so we used this fix instead, which is weird?

magicus:
We know it works because I did is as part of a separate change (upcoming questbook refactorings) that crashes minecraft, and I noticed that we did not get any Wynntils info. :-( So I fixed the crash report manager, reran the test that caused the crash, and checked that it looks okay.

To be honest I don't really understand why the previous solution did not work; we are instantiating this once, and only once, and we should have been able to get hold of the instance field. But we didn't. And, it was a hacky solution. Now we're back to the old tried and tested version with just static fields.

kristofbolyai (approved):

PR #859, opened 2022-12-17
feat: Add crash protection for Screens

magicus:
Wrap all our Screens in a wrapper, which catches all Throwable, and logs them, and closes the screen, instead of letting Minecraft crash in cause of failure.

Future possible improvements:
 * Disable the screen so it can't be reopened in that session. (Not sure it is a good idea)
 * Wrap more delegated methods in crash protection
 * Use the fact that we wrap all our render calls to measure performance impact


common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
```diff
@@ -0,0 +1,313 @@
...
+        McUtils.mc().setScreen(null);
+    }
+
+    @Override
+    public void init(Minecraft minecraft, int width, int height) {
+        try {
+            delegate.init(minecraft, width, height);
+        } catch (Throwable t) {
+            failure("init", t);
+        }
```
kristofbolyai: Why do we only catch stuff in init and render? mouse/keyboard handlers, could crash just like these. 
IMO also try/catch: `keyPressed`, `keyReleased`, `charTyped`, `mouseMoved`, `mouseClicked`, `mouseReleased`, `mouseDragged` and `mouseScrolled`. I hope this is a complete list.

kristofbolyai (dismissed):
Cool thing! I am just not sure why we don't catch other crash candidates..

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
```diff
@@ -0,0 +1,313 @@
...
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.util.FormattedCharSequence;
+import net.minecraft.world.inventory.tooltip.TooltipComponent;
+import net.minecraft.world.item.ItemStack;
+
+public class ScreenWrapper extends Screen {
+    private final Screen delegate;
```
kristofbolyai: How well does this work with instanceof checks? Not really well.. Is there any place we use instanceof for our custom screens? (I know we do for MC screens)

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
```diff
@@ -0,0 +1,313 @@
...
+import net.minecraft.client.gui.narration.NarratableEntry;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.TextComponent;
+import net.minecraft.util.FormattedCharSequence;
+import net.minecraft.world.inventory.tooltip.TooltipComponent;
+import net.minecraft.world.item.ItemStack;
+
+public class ScreenWrapper extends Screen {
```
kristofbolyai:
```suggestion
public class WynntilsScreenWrapper extends Screen {
```
?

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
kristofbolyai: And about this... Do we wanna try/catch this init, or the other one? This is the one that MC calls right? And the other one is overriden to inject renderables (1.19 renamed the later one to `readdRendererables` or smth like that)

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
kristofbolyai: Perhaps, make some util which checks instanceof, but if the current instance is this wrapper, it can handle it as well.

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
magicus: As I said:

>Future possible improvements:
>
>  * Wrap more delegated methods in crash protection

I tried making a regexp to match all methods and insert the "try/catch/call `failure` with method name" pattern everywhere, but gave up. And I didn't want to do all copy paste manually.

Like 80-90% of the code in all Screens are in init and render, so the likelihood of crashes is by far biggest there. But I agree in principle that we should protect the rest as well. 

But your list is not so long. If you are satisfied with that list, I can add those as well.

kristofbolyai: I never used the other methods, and I made a few screens :). So I think these should do.

magicus: We definitely want to catch this init. I had quite a bit of an adventure trying to figure out how to get this to work wrt to init. I started by just capturing the parameterless init that is overridden. Since that was protected, I could not call it. *sigh* So I opened up with an accesswidener. Then I could not build since all our overrides where protected and now the superclass where public and you are not allowed to make accessability narrower. *sigh* So I removed the accesswidener and added a method invoker instead using mixins. Now that part worked. But then screens started crashing. Since the top-level init method setup the field `minecraft`, this was done for the wrapper, and not the delegate. I could have copied that, but it seemed wrong, so I changed the override to the top-level init instead, but that means I had to remove the final from it.

But now things work.

I have checked all our screens. I have skimmed through Screens.java in vanilla and tried to verify that they don't do anything else weird that might cause problems. So I think this is fine.

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
magicus: We do that in once place, in the CharacterSelection screen. See the patch for that file. I can turn that into a util method, maybe it's a good idea.

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
kristofbolyai: Okay. Sounds good.

kristofbolyai: > But your list is not so long. If you are satisfied with that list, I can add those as well.

Trust me, I added enough screens to know where I can crash them :)

kristofbolyai: I am just working on one. Hence this PR review was done on 3-4 separate occasions. As I was writing the new screen, I realized what could go wrong here :)

common/src/main/java/com/wynntils/wynn/model/CharacterSelectionManager.java
```diff
@@ -49,11 +50,13 @@ public CharacterSelectionManager() {
 
     @SubscribeEvent
     public void onScreenOpened(ScreenOpenedEvent event) {
-        if (event.getScreen() instanceof CharacterSelectorScreen characterSelectorScreen) {
```
magicus, line 52: Here is where we do the instance check. I tried turning this into a utility function, but it was kind of pointless. In the end you still want to cast it to a specific subclass of screen, and that could not be done by a method. At least not without a bunch of generics black magic. Unless we get more instances, I suggest we skip doing a util function, ok?

kristofbolyai: Okay.

common/src/main/java/com/wynntils/gui/screens/ScreenWrapper.java
magicus: Ok, with one nights sleep I was able to figure out how to do the regexp, so I fixed all delegated calls.

kristofbolyai (approved):
Looks good

PR #866, opened 2022-12-17
refactor: sort mixins and accesswideners

magicus:


common/src/main/resources/wynntils.accessWidener
```diff
@@ -1,44 +1,44 @@
-accessWidener v2 named
```
kristofbolyai: why newline?

common/src/main/resources/wynntils.mixins.json
```diff
@@ -2,22 +2,23 @@
   "client": [
     "AbstractContainerScreenMixin",
```
kristofbolyai, line 3: I thought these are sorted by spotless... At least, they used to be

common/src/main/resources/wynntils.accessWidener
magicus: my bad, i did not check properly after sorting.

common/src/main/resources/wynntils.mixins.json
```diff
@@ -42,7 +43,6 @@
     "accessors.MinecraftAccessor",
     "accessors.OptionsAccessor",
     "invokers.ChatScreenInvoker",
```
kristofbolyai:
```suggestion
    "invokers.ChatScreenInvoker"
```

common/src/main/resources/wynntils.mixins.json
magicus: They don't seem to be properly sorted right now anyway, and spotless has not complained. However, I see that the IntelliJ sort I used seem to be case-sensitive and that the old sort was not:

```
-    "ClientPacketListenerMixin",
     "ClientPackSourceMixin",
+    "ClientPacketListenerMixin",
```

I don't know which is "right", or at least "best". I think it is convenient to use "sort lines" in IntelliJ, but non-case sensitive sorting seems saner, I think.

magicus:
Argh. I though this was more a less a no-brainer, and I got a bit too quick about it. Is the case-sensitive sort order a problem? And can we get Spotless to sort json files for us? That'd be great.

kristofbolyai:
> Argh. I though this was more a less a no-brainer, and I got a bit too quick about it. Is the case-sensitive sort order a problem? And can we get Spotless to sort json files for us? That'd be great.

I think we can. I will look into it later.

kristofbolyai (dismissed):

P0keDev:
i'm with bolyai in that i could've sworn that we had a custom spotless task for json ordering at one point

IMO case insensitive ordering makes more sense

magicus:
@P0keDev We have, but of course it cannot reorder elements in an array. There is no way for it to know that it would be safe and okay in this particular case.

PR #869, opened 2022-12-18
refactor: Use 1.19 component factories

magicus:
I figured out a way to prepare for the 1.19 port. By adding a (temporary) local copy of the Component class, with the new 1.19 style factory methods implemented in a way that is suitable for 1.18, we can create sort of a "hybrid" code base, which is 1.18 and runs on 1.18, but where all the Component constructions are made like we will need to do in 1.19.

By integrating this now, the diff between main and the 1.19 port will shrink about an order of magnitude.


magicus:
Ok, this didn't work either. But I'll keep this branch as a comparison base for the real 1.19 integration, so it is easier to review, and we could possibly integrate it in two steps (with a short breakage in-between).

magicus:
This is integrated into main_1.19. We're doing the rest of the 1.19 port review there.

PR #870, opened 2022-12-18
WIP: Reference PR -- 1.19 port except the Component changes

magicus:
This is not intended to be pushed, just as a way to review the 1.19 changes.


magicus:
Using main_1.19 instead

PR #872, opened 2022-12-18
Last changes from 1.19_port

magicus:
DO NOT MERGE. This is purely for reviewing support. This PR shows all changes in the 1.19 port, except the Component factories.


magicus:
I have basically taking this PR, and broken it down in chunks, with some effort of grouping changes how they logically belong together. This is not an exact science, but hopefully it makes some kind of sense. I have created the following PRs:

#877 -- simple updates
#878 -- packets
#879 -- chat and commands
#880 -- rendering and widget toolkit
#881 -- gamma
#882 -- changes that look broken?! These needs discussion
#883 -- the rest! Hopefully when we have merged 887-881 (and discarded/solved 882), we can take a renewed effort at the loose ends left here. This was mostly created as a reference, in case some of the changes here really belonged with any of 877-881.

PR #874, opened 2022-12-18
refactor: In preparation for 1.19.3, break out sendCommand()

magicus:
In 1.19, we need to use a special vanilla method to send commands, not just send a chat starting with "/". By creating this method already, we minimize the merge problems with the 1.19.3 port, since we then only has to update the implementation of this method.


magicus:
Duh. Already fixed in https://github.com/Wynntils/Artemis/pull/873

PR #876, opened 2022-12-18
refactor: Convert to using Button.Builder

magicus:


common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
```diff
@@ -67,8 +67,10 @@ public void onPauseScreenInitEvent(PauseMenuInitEvent event) {
         }
     }
 
-    private Button replaceButtonFunction(Button widget, Component translatableComponent, Button.OnPress onPress) {
-        return new Button(
-                widget.getX(), widget.getY(), widget.getWidth(), widget.getHeight(), translatableComponent, onPress);
+    private Button replaceButtonFunction(Button widget, Component component, Button.OnPress onPress) {
+        return new Button.Builder(component, onPress)
+                .pos(widget.getX(), widget.getY())
+                .size(widget.getWidth(), widget.getHeight())
```
kristofbolyai, line 73: I was a bit inconsistent about this. .bounds does pos and size as well. Can you make every builder use bounds?

magicus: Actually I thought pos + size was much clearer! So I'd rather use that everywhere, if it is okay with you?

kristofbolyai: Uh. Fine. I used bounds because it was easier to port. I think both are fine, so use the one you like

PR #879, opened 2022-12-18
feat: All changes related to chat and commands

magicus:


common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -49,17 +71,85 @@ public abstract class ClientPacketListenerMixin {
...
+        if (result.isCanceled()) {
+            ci.cancel();
+        }
+
+        if (!Objects.equals(string, result.getMessage())) {
+            // Note: This code is taken from the MC source code. I don't think there is a sane way of not doing this,
+            // unfortunately (in 1.19.2, there was, not in 1.19.3).
+            //       For anyone concerned, this does not mess with NoChatReport at the time of writing.
+
+            Instant instant = Instant.now();
```
magicus: This sounds a bit scary, too. Can we end up signing messages, claiming the user sent them, when they did in fact not, since we modified the message? I can't think of really how this can be mis-used, but the entire chat report thing is creepy, and I'd hate for us to make anything stupid here.

magicus: I checked where and why we need to change the sent message (cancelling is another story, that does not play bad with chat reporting). It is done in two places: in CommandAliasesFeature and TradeMarketPriceConversionFeature. As I said in discord, I wonder if the first one is broken. For the second one, I've been thinking for a long time to write a GUI to the trade market so you can type the search terms in the GUI, not the chat. If we do that, we can do price substitution in the GUI string, even before it is sent in chat.

If it ends up that we don't need to support changing the message text, I think we should remove this code.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -206,19 +296,79 @@ private void setSubtitleTextPre(ClientboundSetSubtitleTextPacket packet, Callbac
...
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lnet/minecraft/client/multiplayer/chat/ChatListener;handleSystemMessage(Lnet/minecraft/network/chat/Component;Z)V"),
+            cancellable = true)
+    private void handleSystemChat(ClientboundSystemChatPacket packet, CallbackInfo ci) {
+        if (!isRenderThread()) return;
+        ChatPacketReceivedEvent result = EventFactory.onChatReceived(
+                packet.overlay() ? com.wynntils.mc.objects.ChatType.GAME_INFO : com.wynntils.mc.objects.ChatType.SYSTEM,
```
magicus, line 328: I thought you said there were no SYSTEM? Or was it just that Wyncraft never sent any?

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -49,17 +71,85 @@ public abstract class ClientPacketListenerMixin {
...
 
     @Inject(
             method = "handleCommands(Lnet/minecraft/network/protocol/game/ClientboundCommandsPacket;)V",
             at = @At("HEAD"))
     private void handleCommandsPre(ClientboundCommandsPacket packet, CallbackInfo ci) {
         if (!isRenderThread()) return;
-        CommandsPacketEvent event = EventFactory.onCommandsPacket(packet.getRoot());
-        ((ClientboundCommandsPacketAccessor) packet).setRoot(event.getRoot());
+        RootCommandNode<SharedSuggestionProvider> root = packet.getRoot(
+                CommandBuildContext.simple(this.registryAccess.compositeAccess(), this.enabledFeatures()));
```
magicus, line 115: This looks just like a magic incantation. Did you find this on the internet, or figure it out yourself? If the latter, how? :-O Or rather, what does this do? If you've just copied it from somewhere, I'm happy to accept that too :)

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
```diff
@@ -294,9 +294,9 @@ private static void handleStyleDifference(Style oldStyle, Style newStyle, String
                     result.append(different);
                     return;
                 }
-            }
 
-            result.append(ChatFormatting.RESET);
+                result.append(ChatFormatting.RESET);
```
magicus, line 298: I thought I had commented on this, but I can't see it now. What is this? I don't understand the effect of moving this RESET append. 

The change does not seem to have anything to do with 1.19. If it is an actual bug fix, can you push it separately into main?

magicus (commented):
This is a bit scary. Apart from what I believe is a bug in CommandAliasesFeature due to lack of adaptations to the 1.19 command changes, I think this is okay. Scary, but okay. We might need some additional testing of rewriting both incoming and outgoing messages to make sure this works without issues.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
kristofbolyai: No, it is a 1.19 bug fix. ComponentUtils#getCoded decoding behaved differently in 1.19, making our regexes fail. I've looked at them, only to find an excessive amount of RESETs. Turns out, something changed internally, and with this change, we get the same results as 1.18.2.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
kristofbolyai: There were no CHAT. If you look above, the `handlePlayerChat`, that is where we would get CHAT. But we don't. I suggest we merge this as is, and do some modifications on 1.18.2, and merge it in here.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
kristofbolyai: My latest commit to this branch (and the orignal port) should resolve your concerns.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
kristofbolyai: I got this from either reading MC source code, or some FAPI source code.

kristofbolyai (approved):
LGTM if you accept my responses

common/src/main/java/com/wynntils/features/user/CommandAliasesFeature.java
```diff
@@ -33,16 +34,14 @@ public class CommandAliasesFeature extends UserFeature {
...
-        String message = e.getMessage();
-
-        if (message.startsWith("/")) {
-            final String command = message.substring(1);
-            for (CommandAlias commandAlias : aliases) {
-                if (commandAlias.getAliases().stream().anyMatch(alias -> Objects.equals(alias, command))) {
-                    e.setMessage("/" + commandAlias.getOriginalCommand());
-                    break;
-                }
+    public void onChatSend(CommandSentEvent e) {
```
magicus:
```suggestion
    public void onCommandSent(CommandSentEvent e) {
```

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -206,19 +264,79 @@ private void setSubtitleTextPre(ClientboundSetSubtitleTextPacket packet, Callbac
...
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lnet/minecraft/client/multiplayer/chat/ChatListener;handlePlayerChatMessage(Lnet/minecraft/network/chat/PlayerChatMessage;Lcom/mojang/authlib/GameProfile;Lnet/minecraft/network/chat/ChatType$Bound;)V"),
+            cancellable = true)
+    private void handlePlayerChat(ClientboundPlayerChatPacket packet, CallbackInfo ci) {
+        if (!isRenderThread()) return;
+        ChatPacketReceivedEvent result =
+                EventFactory.onChatReceived(com.wynntils.mc.objects.ChatType.CHAT, packet.unsignedContent());
```
magicus, line 278: So this entire mixin is not needed, since Wynncraft does not send ClientboundPlayerChatPacket in 1.19? Can we remove it if that is the case?

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
magicus: Ok, let's make a change in 1.18 so we stop differentiate between CHAT and SYSTEM.

common/src/main/java/com/wynntils/mc/utils/ComponentUtils.java
magicus: I see. Thanks.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
kristofbolyai: I would keep it just so we can process it in case Bukkit decides to update and use it for some reason? I know it is quite large, but I would rather have the ability to do this.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -206,19 +264,79 @@ private void setSubtitleTextPre(ClientboundSetSubtitleTextPacket packet, Callbac
...
+                    packet.signature(),
+                    signedMessageBody,
+                    packet.unsignedContent(),
+                    packet.filterMask());
+
+            this.minecraft.getChatListener().handlePlayerChatMessage(playerChatMessage, playerInfo.getProfile(), bound);
+            this.messageSignatureCache.push(playerChatMessage);
+
+            ci.cancel();
+        }
```
Incompleteusern, line 314: what does this large if statement do?

kristofbolyai: handle changed messages

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
magicus: ok

magicus:
@kristofbolyai I'll merge this into main_1.19 now. I will open a new PR to remove the distinction between chat and system in main (if you did not do it already).

kristofbolyai:
> @kristofbolyai I'll merge this into main_1.19 now. I will open a new PR to remove the distinction between chat and system in main (if you did not do it already).

I didn't. I just got home.

PR #880, opened 2022-12-18
feat: All changes related to rendering

magicus:
This include mostly the changes to JOML, but also some Vanilla widget toolkit changes (like tooltips)


common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -71,13 +71,17 @@ public void onRenderLevelPost(RenderLevelEvent.Post event) {
...
 
         // apply camera rotation
-        projection.multiply(Vector3f.XP.rotationDegrees(camera.getXRot()));
-        projection.multiply(Vector3f.YP.rotationDegrees(camera.getYRot() + 180.0F));
+        Vector3f xp = new Vector3f(1, 0, 0);
+        Vector3f yp = new Vector3f(0, 1, 0);
+        Quaternionf xRotation = new Quaternionf().rotationAxis((float) Math.toRadians(camera.getXRot()), xp);
+        Quaternionf yRotation = new Quaternionf().rotationAxis((float) Math.toRadians(camera.getYRot() + 180f), yp);
+        projection.mul(new Matrix4f().rotation(xRotation));
+        projection.mul(new Matrix4f().rotation(yRotation));
```
magicus, line 84: I think it's probably 20 years since I forgot everything I ever knew about vector algebra. :-) It looks sane, but let me just ask: have you verified that the world waypoint seem to work? (It's more like if it's broken it's likely severly broken)

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
```diff
@@ -198,7 +202,7 @@ public void onRenderGuiPost(RenderEvent.Post event) {
 
     private Vec3 worldToScreen(Vector3f delta, Matrix4f projection) {
         Vector4f clipCoords = new Vector4f(delta.x(), delta.y(), delta.z(), 1.0f);
-        clipCoords.transform(projection);
+        projection.transform(clipCoords);
```
magicus, line 205: This inversion of instance and argument looks a bit sus. Is it just that the JOML transform method switches operands compared to Mojangs old version with the same name?

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
```diff
@@ -68,16 +68,15 @@ public class OverlayManagementScreen extends Screen {
...
-    private static final List<Component> CLOSE_TOOLTIP_LINES =
-            List.of(Component.literal("Click here to stop editing and reset changes."));
+    private static final Component CLOSE_TOOLTIP_LINES =
+            Component.literal("Click here to stop editing and reset changes.");
 
-    private static final List<Component> TEST_TOOLTIP_LINES = List.of(
-            Component.literal("Click here to toggle test mode."),
-            Component.literal("In test mode, you can see how your overlay setup would look in-game,"),
-            Component.literal("using preview render mode."));
+    private static final Component TEST_TOOLTIP_LINES = Component.literal("Click here to toggle test mode.")
```
magicus: Will these tooltip still appear with one string per line, or will the append just join them together in a huge string, that is then wrapped "at will" by the tooltip renderer?

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
```diff
@@ -105,7 +105,7 @@ private void renderTooltipPost(PoseStack poseStack, ItemStack itemStack, int mou
     }
 
     @Inject(
-            method = "init(Lnet/minecraft/client/Minecraft;II)V",
+            method = "rebuildWidgets",
```
magicus, line 108: Have they renamed the top-level "init" method? :-o If so, then the WynntilsScreenWrapper would fail..?

Or maybe they only extracted the part that called our init method into a separate method?

magicus (commented):
I think this looks good overall, but there were some eye brow raising stuff that I'd like to get sorted out, as per the comments.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
kristofbolyai: Yes, I did. I even talked to the original creator of this code (as I failed to port a singular line of change). But now, it works as intended.

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
kristofbolyai: https://github.com/Wynntils/Artemis/pull/880#discussion_r1051947100 
> (as I failed to port a singular line of change).

This was it. I have no idea what whichery MC did, or what JOML does, but this is how it is done. And it works.

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
kristofbolyai: No, it is worse that that :). We have `init(Minecraft minecraft, int width, int height)`, does the same as the old one, except it does not call another init. It calls `rebuildWidgets`, which in turn calls the secondary init.

```java
public void init(Minecraft minecraft, int width, int height) {
        this.minecraft = minecraft;
        this.itemRenderer = minecraft.getItemRenderer();
        this.font = minecraft.font;
        this.width = width;
        this.height = height;
        this.rebuildWidgets();
        this.triggerImmediateNarration(false);
        this.suppressNarration(NARRATE_SUPPRESS_AFTER_INIT_TIME);
    }
```

```java
protected void rebuildWidgets() {
        this.clearWidgets();
        this.setFocused((GuiEventListener)null);
        this.init();
    }
```

```java
protected void init() {
    }
```

common/src/main/java/com/wynntils/features/user/WorldWaypointDistanceFeature.java
magicus: Glad it's not just me then :)

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
kristofbolyai: I just found another case of MC source code being incredibly dumb. Why would you create a `renderWithTooltp` only to not call `super.render` but `this.render`...? Anyways, I am working on this.

kristofbolyai: I fixed it. It looks okay, MC wraps it by default.

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
```diff
@@ -23,48 +24,79 @@ public class WynncraftPauseScreenFeature extends UserFeature {
...
                 (button) -> {
                     McUtils.mc().setScreen(null);
                     McUtils.mc().mouseHandler.grabMouse();
                     McUtils.sendCommand("hub");
                 });
 
-        renderables.set(4, hub);
+        replacedButtons.set(4, hub);
+
+        GridWidget newGridWidget = new GridWidget();
```
Incompleteusern, line 78: why is there a `GridWidget` now?

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
```diff
@@ -242,6 +240,11 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
         }
 
         super.render(poseStack, mouseX, mouseY, partialTick); // This renders widgets
+        // This renders button tooltips
+        if (this.deferredTooltipRendering != null) {
+            this.renderTooltip(poseStack, this.deferredTooltipRendering, mouseX, mouseY);
+            this.deferredTooltipRendering = null;
+        }
```
Incompleteusern, line 245: why is `deferredTooltipRendering` set to null?

common/src/main/java/com/wynntils/mc/mixin/ScreenMixin.java
magicus: ok, that was what i meant by "they only extracted the part that called our init method into a separate method". But that's fine. This is the correct fix for that.

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
```diff
@@ -71,9 +71,8 @@ public class OverlayManagementScreen extends Screen {
     private static final Component CLOSE_TOOLTIP_LINES =
```
magicus: The name of these constants made sense when they were a List<Component>, but now I think they should be just CLOSE_TOOLTIP etc.

common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
```diff
@@ -23,48 +24,79 @@ public class WynncraftPauseScreenFeature extends UserFeature {
...
     public void onPauseScreenInitEvent(PauseMenuInitEvent event) {
         PauseScreen pauseScreen = event.getPauseScreen();
-        List<Renderable> renderables = new ArrayList<>(pauseScreen.renderables);
+
+        Optional<Renderable> gridOpt = pauseScreen.renderables.stream()
+                .filter(x -> x instanceof GridWidget)
+                .findFirst();
+        if (gridOpt.isEmpty()) return;
+
+        GridWidget grid = (GridWidget) gridOpt.get();
```
magicus, line 33: I realize I skipped past this file when reviewing earlier. This solution works, and if you don't want to muck around with it, I'm okay with that, but I just wanted to raise an idea: don't you think it would be easier to intercept the creation of the GridWidget instead?

kristofbolyai: I think it's the same amount of "bad"

magicus: Ok, let's keep this way then.

common/src/main/java/com/wynntils/features/user/WynncraftPauseScreenFeature.java
magicus: Because Vanilla remade the pause menu..?

kristofbolyai: yes, vanilla is different now.

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
kristofbolyai: This is how tooltip rendering is done in vanilla with button widgets. I would have prefered to call super, but MC code is bad and doesn't let me.

common/src/main/java/com/wynntils/gui/screens/overlays/OverlayManagementScreen.java
kristofbolyai: Done

Incompleteusern (approved):

PR #881, opened 2022-12-18
feat: All changes related to gamma/brightness

magicus:


common/src/main/java/com/wynntils/features/user/GammabrightFeature.java
```diff
@@ -23,65 +20,20 @@ public class GammabrightFeature extends UserFeature {
...
         Managers.Config.saveConfig();
     }
-
-    private void resetGamma() {
-        McUtils.options().gamma = lastGamma;
-    }
-
-    private void enableGammabright() {
-        lastGamma = McUtils.options().gamma;
-        McUtils.options().gamma = 1000d;
```
magicus, line 85: So there is no longer a gamma variable, but the gamma value is returned set when you convert a Double to float in `updateLightTexture`? It seems a bit... I dunno. Maybe this is the only way, but I don't like `@Redirect`. This sounds like it will break with other mods that change gamma, like Tweakeroo.

magicus: *reading the code* Oh, I see, they had to rewrite this to be more dynamic because of the Warden.

Hm.

magicus: If I look at the correct sources, there is still a:
```
                        float v = ((Double)this.minecraft.options.gamma().get()).floatValue();
```

Is that what you are trying to target with the mixin? But I think you will get this instead, that occurs before. Or is the mixin injected in both places in such a case?
```
                float h = ((Double)this.minecraft.options.darknessEffectScale().get()).floatValue();
```

magicus: I think it would be safer and more robust to hijack
```
    public OptionInstance<Double> gamma() {
        return this.gamma;
    }
```
in Options instead. Maybe send a GammaRequestedEvent with the option to supply a replacement OptionInstance?

magicus: Or, maybe we could just do `minecraft.options.gamma().set(100.0f)`? I think that ought to work. That would probably be a straightforward port of the old code. Save the old gamma value by `minecraft.options.gamma().get()` and then update it using the setter.

magicus (commented):
Sorry, this was a bit rambling as I worked my way through this. The TL;DR: I think we should revert most of these changes, go back to the old solution, and use `McUtils.options.gamma().get/set()`. Unless this somehow does not work.

common/src/main/java/com/wynntils/features/user/GammabrightFeature.java
kristofbolyai: > Or, maybe we could just do `minecraft.options.gamma().set(100.0f)`? I think that ought to work. That would probably be a straightforward port of the old code. Save the old gamma value by `minecraft.options.gamma().get()` and then update it using the setter.

No, config options for doubles are limited between 0-1.

kristofbolyai: Hacking options is a way, I tried it. It is worse than this. I can agree with perhaps hijacking the config option itself. Although, that is hard because of the OptionsInstance wrapper. This might be the best we can do.

kristofbolyai: If you don't worry about this mixin incompatiblity, then I think we can agree that the feature is much simpler now. And since gamma/options work this way, we need to do a hack somewhere. This is the "nicest" hack I could find.

kristofbolyai (approved):
This code is now fine, since I fixed the mixin. We need to discuss whether you like it or not.

common/src/main/java/com/wynntils/features/user/GammabrightFeature.java
magicus: > No, config options for doubles are limited between 0-1.

Ok, but we can inject a direct access to the `value` field in OptionInstance so we can get around the validation in the set method. Then we can change it to whatever.

kristofbolyai: MC would complain on config load

magicus: And no, I don't agree that the situation where we send out an event each time we try to read the event variable is simpler, than updating the value when requested. It might be that this is the only reasonable way forward, but I still think the old way was better and I'd like to not let that go yet.

I will look into Tweakeroo to see how they do gamma bright on 1.19...

magicus: Yes, that is how they do it:

```
            IMixinSimpleOption<Double> opt = (IMixinSimpleOption<Double>) (Object) this.mc.options.getGamma();
            opt.tweakeroo_setValueWithoutCheck(gamma);
```

kristofbolyai: Fine. I guess we need to revert this and I'll do the mixin.

magicus: Thanks!

kristofbolyai:
@magicus Done

magicus:
Thanks! I feel much more confident in this solution.

PR #882, opened 2022-12-18
CHECK: These changes look wrong. Check that they really are correct

magicus:
Do not merge until all changes are confirmed.


common/src/main/java/com/wynntils/mc/event/ContainerSetContentEvent.java
```diff
@@ -10,7 +10,7 @@
 import net.minecraftforge.eventbus.api.Event;
 
 public abstract class ContainerSetContentEvent extends Event {
-    private final List<ItemStack> items;
```
magicus, line 13: Okay, I understand and accept these changes. They are correct.

common/src/main/java/com/wynntils/gui/screens/settings/elements/EnumConfigOptionElement.java
```diff
@@ -84,8 +84,6 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
             return false;
         }
 
-        McUtils.playSound(SoundEvents.UI_BUTTON_CLICK.value());
```
kristofbolyai, line 87: It was duplicated. If you look below, we call it.

common/src/main/java/com/wynntils/gui/widgets/TextInputBoxWidget.java
```diff
@@ -254,7 +254,7 @@ public boolean isFocused() {
     }
 
     @Override
-    public void updateWidgetNarration(NarrationElementOutput narrationElementOutput) {}
+    protected void updateWidgetNarration(NarrationElementOutput narrationElementOutput) {}
```
kristofbolyai: The method is now protected. We should change all public cases to protected.

magicus: We have this in lots of other places, which are public. I suggest we revert these two odd ones, and then make a separate change to make them protected. (And it is always legal to widen scope, so it is more an issue of style)

common/src/main/java/com/wynntils/gui/screens/settings/elements/EnumConfigOptionElement.java
magicus: So this is a bug fix? Can we do that separately directly to main?

kristofbolyai: Sure.

kristofbolyai:
@magicus I think this PR served it's purpose. Can you merge the container event changes? I opened a PR for the sound removal on main.

PR #883, opened 2022-12-18
CHECK: Remaining changes

magicus:
These are the remaining changes after I sorted out all the others. I do not think this should be integrated as-is. Rather, I think this could serve as a reference if any change here really was needed by any of the other PRs, and I missed it.


forge/src/main/java/com/wynntils/forge/WynntilsModForge.java
```diff
@@ -36,5 +35,16 @@ public WynntilsModForge() {
                         .toString(),
                 !FMLEnvironment.production,
                 modFile);
+
+        WynntilsMod.registerEventListener(this);
+    }
+
+    // FIXME: This is slightly hacky to do this but it works!
```
magicus: Do you intend to change this? Otherwise remove "FIXME".

magicus (commented):
I think all changes here look good, but I'd like to take another round here when the other PRs are merged.

common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -127,7 +130,22 @@ private void handleContainerClosePre(ClientboundContainerClosePacket packet, Cal
             cancellable = true)
     private void handleContainerContentPre(ClientboundContainerSetContentPacket packet, CallbackInfo ci) {
         if (!isRenderThread()) return;
-        if (EventFactory.onContainerSetContentPre(packet).isCanceled()) {
+        ContainerSetContentEvent event = EventFactory.onContainerSetContentPre(packet);
+        if (event.isCanceled()) {
+            ci.cancel();
+        }
+
+        if (packet.getItems() != event.getItems()) {
```
magicus: Can you really use "==" here, as opposed to "equals"?

forge/src/main/java/com/wynntils/forge/WynntilsModForge.java
```diff
@@ -36,5 +35,16 @@ public WynntilsModForge() {
...
+    @SubscribeEvent
+    public void onClientLoad(TitleScreenInitEvent.Pre event) {
+        // Enable stencil support
+        Minecraft.getInstance().getMainRenderTarget().enableStencil();
+
+        WynntilsMod.unregisterEventListener(WynntilsModForge.class);
```
kristofbolyai: Note: We should unregister this, not the class.

common/src/main/java/com/wynntils/mc/EventFactory.java
```diff
@@ -263,6 +264,10 @@ public static LivingEntityRenderTranslucentCheckEvent onTranslucentCheck(boolean
         return post(new LivingEntityRenderTranslucentCheckEvent(translucent, entity, translucent ? 0.15f : 1f));
     }
 
+    public static UpdateLightTextureEvent onUpdateLightTexture(float oldGamma) {
+        return post(new UpdateLightTextureEvent(oldGamma));
+    }
+
```
kristofbolyai: This isn't supposed to be here.

kristofbolyai (approved):
Otherwise LGTM

PR #884, opened 2022-12-19
fix: Require a newer version of FML (to match Forge version req)

kristofbolyai:
The version requirement for Forge Mod Loader is the same as Forge now. This is needed so users get notified they are using an old Forge before they crash (older Forge versions don't load the game, and people think it is our fault).


magicus (approved):

magicus:
@kristofbolyai What about the 1.19 port? Will you make the corresponding change there?

kristofbolyai:
> @kristofbolyai What about the 1.19 port? Will you make the corresponding change there?

On there, they match atm, because I haven't specified a subversion.

PR #885, opened 2022-12-19
docs: hotswap doesn't work on windows

ryanzhoudev:


kristofbolyai:
@DonkeyBlaster I think we can close this, as upstream plans to fix it?

PR #886, opened 2022-12-19
feat: Triple totem tracking

ryanzhoudev:
- Adds SpellCastedEvents


common/src/main/java/com/wynntils/core/managers/Models.java
```diff
@@ -98,6 +99,7 @@ public class Models {
     public static final SkillIconPropertyModel SkillIconProperty = new SkillIconPropertyModel();
     public static final SkillPointPropertyModel SkillPointProperty = new SkillPointPropertyModel();
     public static final SoulPointItemStackModel SoulPointItemStack = new SoulPointItemStackModel();
+    public static final SubtitleModel Subtitle = new SubtitleModel();
```
kristofbolyai: I think it might be better to make this model only care about spells. Subtitle does too many things to be a model.

ryanzhoudev: I did some changes, let me know if this is what you had in mind

common/src/main/java/com/wynntils/wynn/model/SpellModel.java
```diff
@@ -0,0 +1,62 @@
...
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import com.wynntils.mc.objects.ChatType;
+import com.wynntils.wynn.event.SpellCastedEvent;
+import com.wynntils.wynn.objects.SpellType;
+import com.wynntils.wynn.utils.WynnUtils;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public final class SpellModel extends Model {
```
kristofbolyai, line 19: I know I just asked you to make this a model... But, #909 just happened. I think does qualify as a Handler? Since we take MC stuff, chat packets, parse it for spell events. Then, models can listen to the events provided by handlers. That is the text-book case of handlers, right? @magicus

kristofbolyai: Which makes me wonder, why `ActionBarModel` was not transformed into a handler? And should this code go there, as we basically parse action bar, and send out events? (questions to magicus)

kristofbolyai: And @DonkeyBlaster, sorry for all these trouble! You got unlucky, and decided to do this while magicus is working on one of the biggest refactors of Artemis.

kristofbolyai: #914 does just this.

magicus: Yes, correct. ActionBar is definitely a text-book Handler. It was just waaaay harder to convert than I could imagine. I've spend more hours than I care to admit on #914...

kristofbolyai:
Not sure if I commented on this or not, but magicus reminden me. `QuickSpellFeature` does a LOT of spell parsing and spell related info (as does `ActionBarHandler` when it is merged). So, maybe make it so the `SpellModel` you are making, removes a bit of logic from `QuickSpellFeature`, as it does too much at the moment.

PR #888, opened 2022-12-19
refactor: Stop differentiating between SYSTEM and NORMAL messages

magicus:
This is in preparation of the 1.19 port. In 1.19, Wynncraft only sends SYSTEM messages (probably to avoid problems with chat signing), so we can't use the difference between "NORMAL" and "SYSTEM" anymore.


kristofbolyai (approved):
Looks good

PR #892, opened 2022-12-20
feat: Redirects "Housing Teleport" messages to Overlay

san7890:
Adds redirect matchers for flying to and from a housing island. I also found an edge-case scrollTeleportation failure message that triggers when you try and use a scroll on a housing island, so I added that in this PR as well (required to rename another redirector to ensure differentiation).

image


san7890:
i'm okay to wait on the 1.19 port to go through by the way, just putting this up now so i can put out fires as they may arise when that goes through

kristofbolyai (approved):

PR #896, opened 2022-12-20
feat: New approach at screen crash protection

magicus:
This uses a 100% safe methodology, but it will only protect init and render. Otoh, these are the two mehtods mostly in need of protection.


common/src/main/java/com/wynntils/gui/screens/WynntilsScreen.java
```diff
@@ -0,0 +1,50 @@
...
+            failure("init", t);
+        }
+    }
+
+    protected abstract void doInit();
+
+    @Override
+    public final void render(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        try {
+            super.render(poseStack, mouseX, mouseY, partialTick);
```
kristofbolyai: Why? Can't we decide in doRender if we want to call super? Can we even do that with this change? Since super.render will be this call, not `Screen#render`?

kristofbolyai: Not every render call in screens call super, some of them do renders.

magicus: To be able to call super.render() from the subclasses we would need to provide a pass-through, like `public void superRender(PoseStack poseStack, ...) { super.render(...) }`.

But I think a failure to call super.render is indicative of a design/inheritance problem.

kristofbolyai: But that means this is not a great solution?

magicus: Yes, sure, I'll give you that. This is not a great solution. But it is *a* solution. And it won't crash.

kristofbolyai: But, what do you do about the render calls that call super? And those which don't? This surely does not work with every screen at the moment

magicus: Hm, I realize therein lies the answer to your problem. I solved the issue half-way; I changed calls from super.render to super.doRender. But, of course, I should add a call to super.render in the top level doRender. Then everything will be just fine.

magicus: Or... at least I could have sworn I changed those call. I can't find that in the PR though. I must have missed it.

magicus: Is this acceptable now?

kristofbolyai (approved):
Looks better now. I assume you verified that screens work.

PR #902, opened 2022-12-20
fix: Lootrun screen loading is now async, and is cached

kristofbolyai:
Old version reloaded every lootrun on every search / screen init, and it caused game pauses for seconds. This fixes these issues.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/LootrunModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/LootrunModel.java
@@ -15,6 +15,7 @@ import com.wynntils.core.managers.Model;
 import com.wynntils.features.statemanaged.LootrunFeature;
 import com.wynntils.gui.render.CustomRenderType;
 import com.wynntils.mc.utils.McUtils;
+import com.wynntils.wynn.event.LootrunCacheRefreshEvent;
 import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
 import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import java.io.File;
@@ -65,6 +66,8 @@ public final class LootrunModel extends Model {
             ChatFormatting.BLUE.getColor(),
             0x3f00ff,
             ChatFormatting.DARK_PURPLE.getColor());
+
+    private static List<LootrunInstance> LOOTRUN_INSTANCE_CACHE = new ArrayList<>();
 
     private LootrunState state = LootrunState.DISABLED;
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/wynn/model/LootrunModel.java
```diff
@@ -600,7 +607,8 @@ public List<LootrunInstance> getLootruns() {
             }
         }
 
-        return lootruns;
+        LOOTRUN_INSTANCE_CACHE = lootruns;
```
Incompleteusern, line 610: perhaps a check if the files have changed using md5s might be better for performance

kristofbolyai: I didn't want to overengineer this. As this is done async, it is fast enough, and doesn't lag the GUI.

Incompleteusern: ok

Incompleteusern (approved):

common/src/main/java/com/wynntils/wynn/model/LootrunModel.java
```diff
@@ -600,7 +607,8 @@ public List<LootrunInstance> getLootruns() {
             }
```
magicus, line 607: Another question is: why do we not just list the files in the lootrun dir? It would have been blazingly fast. And read and parse the json only if we try to read in the lootrun. The code above in the loop looks just wrong.

magicus: Like, if we have malformed json in a single lootrun, that will prevent us from even listing the files.

kristofbolyai: We need to parse it to tell the user info about the lootrun. And we do not want to parse on click.

kristofbolyai: I can change the code to not fail on a single malformed file though.

magicus: What do we tell the user? I apparently need to test this. I just throught we listed the name of the lootrun.

kristofbolyai: > Like, if we have malformed json in a single lootrun, that will prevent us from even listing the files.

Actually, no. We do catch errors, log it, but we can show other files. This looks good IMO.

kristofbolyai: image

magicus (approved):

PR #904, opened 2022-12-20
fix: WynntilsAccountManager's token users are now async

### LINKED ISSUE #829: Every consumer of Wynntils Account Tokens should be able to run asynchronously
See https://github.com/Wynntils/Artemis/pull/826#discussion_r1048942877.

kristofbolyai:
Fixes #829

### DIFF
--- a/common/src/main/java/com/wynntils/sockets/model/HadesModel.java
+++ b/common/src/main/java/com/wynntils/sockets/model/HadesModel.java
@@ -27,13 +27,19 @@ import com.wynntils.wynn.event.AthenaLoginEvent;
 import com.wynntils.wynn.event.CharacterUpdateEvent;
 import com.wynntils.wynn.event.RelationsUpdateEvent;
 import com.wynntils.wynn.event.WorldStateEvent;
+import com.wynntils.wynn.model.WorldStateManager;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import net.minecraft.ChatFormatting;
 import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.Style;
 import net.minecraft.world.effect.MobEffects;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/sockets/model/HadesModel.java
```diff
@@ -47,20 +48,25 @@ public final class HadesModel extends Model {
...
+            tryCreateConnection();
+        }
     }
 
     @Override
     public void disable() {
         tryDisconnect();
     }
 
-    private void tryCreateConnection() {
```
magicus, line 58: Most of the PR looks good, but this one kind of bugged me. If Athena login works but Hades connection fails, we will not get any feedback on this. Just like the Athena firstWorld message, maybe we need to do something similar for Hades?

kristofbolyai: Fixed

common/src/main/java/com/wynntils/sockets/model/HadesModel.java
```diff
@@ -127,6 +135,22 @@ public void onPartyListUpdate(RelationsUpdateEvent.PartyList event) {
 
     @SubscribeEvent
     public void onWorldStateChange(WorldStateEvent event) {
+        if (event.getNewState() == WorldStateManager.State.WORLD && isSocketOpen() && firstWorldJoin) {
```
Incompleteusern: How does this check for failure?

kristofbolyai: isSocketOpen, socket is closed on auth fail.

common/src/main/java/com/wynntils/sockets/model/HadesModel.java
```diff
@@ -127,6 +135,22 @@ public void onPartyListUpdate(RelationsUpdateEvent.PartyList event) {
...
+            firstWorldJoin = false;
+
+            // FIXME: Use the proper reload command here, once they are reworked
+            MutableComponent failed = Component.literal("Welps! Trying to connect to Hades failed.")
+                    .withStyle(ChatFormatting.GREEN);
```
Incompleteusern: These literals should really be translatable components. Also "Welps" is a) informal and b) probably should be welp, don't know what I was doing when I wrote that.

magicus: We can fix that now or later. I think we need to do a full sweep of the entire Artemis code base and fix all non-i18n strings. (The one exception is in log messages; they don't need to be i18n)

common/src/main/java/com/wynntils/sockets/model/HadesModel.java
magicus: But this looks wrong? Don't you first need to do an if check if this is firstWorldJoin, and set the field, and then check socket? Also, the socket check looks inverted.

magicus (approved):

common/src/main/java/com/wynntils/sockets/model/HadesModel.java
magicus: I opened https://github.com/Wynntils/Artemis/issues/906 for this.

PR #907, opened 2022-12-21
refactor: Clean up ItemProfilesManager

magicus:
Give some TLC to ItemProfilesManager. Also start the first embryo of an overall ItemManager.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/objects/profiles/ingredient/IngredientProfile.java
+++ b/common/src/main/java/com/wynntils/wynn/objects/profiles/ingredient/IngredientProfile.java
@@ -6,7 +6,6 @@ package com.wynntils.wynn.objects.profiles.ingredient;
 
 import com.google.gson.annotations.SerializedName;
 import com.wynntils.core.managers.Managers;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/wynntils/wynn/objects/profiles/ingredient/IngredientProfile.java
```diff
@@ -89,22 +88,22 @@ public ItemStack asItemStack() {
...
-                CompoundTag textureEntry = new CompoundTag();
-                textureEntry.putString("Value", ingredientHeadTextures.get(name));
-                textures.add(textureEntry);
-                properties.put("textures", textures);
-                skullData.put("Properties", properties);
-
-                itemStack.getOrCreateTag().put("SkullOwner", skullData);
-            }
+            String ingredientHeadTexture = Managers.ItemProfiles.getIngredientHeadTexture(name);
+            // This will look bad, but if we don't have the data, then what should we do?
```
kristofbolyai: Not every ingredient texture is a head, so this comment is a bit weird.

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/wynn/objects/profiles/ingredient/IngredientProfile.java
magicus: This is stated in the context of `if (itemStack.getItem() == Items.PLAYER_HEAD) {`. Maybe this looks weird in the diff, check out how it looks in the updated file. 

If the item is represented by item type PLAYER_HEAD and we have no texture, it sure is going to look bad. Ideally, I'd like a "hook" to notify us that we are missing essential data.

PR #908, opened 2022-12-21
refactor: Create ObjectivesManager

### LINKED ISSUE #898: We need an ObjectiveModel

magicus:
This will fix #898. It is part of the ongoing effort to refactor the entire "wynn model".

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/objectives/ObjectivesManager.java
+++ b/common/src/main/java/com/wynntils/wynn/model/objectives/ObjectivesManager.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.model.objectives;
+
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.managers.Manager;
+import com.wynntils.wynn.model.scoreboard.ScoreboardHandler;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class ObjectivesManager extends Manager {
+    public static final ScoreboardHandler SCOREBOARD_HANDLER = new ObjectiveHandler();
+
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/wynn/model/objectives/ObjectivesManager.java
```diff
@@ -0,0 +1,82 @@
...
+package com.wynntils.wynn.model.objectives;
+
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.managers.Manager;
+import com.wynntils.wynn.model.scoreboard.ScoreboardHandler;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class ObjectivesManager extends Manager {
```
kristofbolyai, line 14: Shoudln't this be a model which depends on scoreboardmodel? Or We cant do that with models? Either way, we should have some kind of relation.

magicus: See #909 for the wider context in which this move is made. Handlers, like Managers, are always available.

While I am shuffling models, managers and handlers around, and trying to tease them apart, I prefer not to have to deal with dependencies. When things have settled, we can re-instantiate dependencies, if we decide to keep them. (I looked through the old Discord discussion to see what we agreed on there, but it was unclear.)

In the end, the "ObjectivesManager" really ought to be a "ObjectivesModel", since it is modelling something that is essential to Wynncraft. However, "Model" instead got the meaning "something which can be enabled or disabled" while "Manager" got the meaning "something which is always enabled". This is unfortunate, but while refactoring I'll roll with this and call everything persistent for "Manager".

kristofbolyai: Okay, got it. I think the ability to disable something the mod does, like modifying the scoreboard is good. I think @Incompleteusern shares the same insight.

kristofbolyai (approved):

PR #909, opened 2022-12-21
Refactor: Introduce Handlers as an intermediary between MC and Models

magicus:
This is a major part of the model refactoring. The Handlers spans the bridge between pure Minecraft code and the Wynncraft Models. Their knowledge about Wynncrafty things should be limited to "structure", like how chat messages are formatted, or how the scoreboard is partitioned into segments.

The individual models can then extract relevant data from these handlers.


magicus:
This were the "easy" handlers. Three more handlers are upcoming, ActionBarHandler, TabListHandler and ItemHandler. The last one is, of course, the black sheep, and will need most work. But the other two require enough changes to have them in a separate PR.

magicus:
This will partly conflict with #908, but I will handle that.

common/src/main/java/com/wynntils/core/managers/Handlers.java
```diff
@@ -0,0 +1,34 @@
...
+    public static final BossBarHandler BossBar = new BossBarHandler();
+    public static final ChatHandler Chat = new ChatHandler();
+    public static final ContainerQueryHandler ContainerQuery = new ContainerQueryHandler();
+    public static final ScoreboardHandler Scoreboard = new ScoreboardHandler();
+
+    public static void init() {
+        // Register all handler singletons as event listeners
+
+        FieldUtils.getAllFieldsList(Handlers.class).stream()
+                .filter(field -> Manager.class.isAssignableFrom(field.getType()))
```
kristofbolyai:
```suggestion
                .filter(field -> Handler.class.isAssignableFrom(field.getType()))
```
?

common/src/main/java/com/wynntils/features/user/overlays/ObjectivesOverlayFeature.java
```diff
@@ -87,7 +87,7 @@ protected GuildObjectiveOverlay() {
 
         @Override
         public void render(PoseStack poseStack, float partialTicks, Window window) {
-            WynnObjective guildObjective = ObjectiveHandler.getGuildObjective();
+            WynnObjective guildObjective = ObjectiveListener.getGuildObjective();
```
kristofbolyai: Will this be ObjectiveModel when it is merged?

common/src/main/java/com/wynntils/wynn/model/GuildAttackTimerModel.java
```diff
@@ -25,7 +25,7 @@ public final class GuildAttackTimerModel extends Model {
     private static final Pattern GUILD_ATTACK_PATTERN = Pattern.compile("§b- (.+):(.+) §3(.+)");
     private static final Pattern GUILD_DEFENSE_CHAT_PATTERN = Pattern.compile("§r§3.+§b (.+) defense is (.+)");
 
-    public static final ScoreboardHandler SCOREBOARD_HANDLER = new GuildAttackHandler();
+    public static final ScoreboardListener SCOREBOARD_HANDLER = new GuildAttackListener();
```
kristofbolyai:
```suggestion
    public static final ScoreboardListener SCOREBOARD_LISTENER = new GuildAttackListener();
```

common/src/main/java/com/wynntils/wynn/model/bossbar/BossBarModel.java
```diff
@@ -4,27 +4,19 @@
...
-import net.minecraft.world.BossEvent;
-import net.minecraftforge.eventbus.api.SubscribeEvent;
+import java.util.List;
 
 public final class BossBarModel extends Model {
+    public static final String BLOODPOOL = "bloodpool";
+    public static final String MANABANK = "manabank";
+    public static final String AWAKENED = "awakened";
+    public static final String FOCUS = "focus";
+    public static final String CORRUPTED = "corrupted";
```
kristofbolyai: Seems a bit weird. Can't we make this an enum? What is the use case?

common/src/main/java/com/wynntils/wynn/model/quests/QuestManager.java
```diff
@@ -26,7 +26,7 @@
 import org.apache.commons.lang3.StringUtils;
 
 public final class QuestManager extends Manager {
-    public static final ScoreboardHandler SCOREBOARD_HANDLER = new QuestScoreboardHandler();
+    public static final ScoreboardListener SCOREBOARD_HANDLER = new QuestScoreboardListener();
```
kristofbolyai:
```suggestion
    public static final ScoreboardListener SCOREBOARD_LISTENER = new QuestScoreboardListener();
```

common/src/main/java/com/wynntils/wynn/model/scoreboard/objectives/ObjectiveListener.java
```diff
@@ -5,17 +5,17 @@
...
-import com.wynntils.wynn.model.scoreboard.Segment;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public class ObjectiveHandler implements ScoreboardHandler {
+public class ObjectiveListener implements ScoreboardListener {
```
kristofbolyai: Agree with Qodana, but I can't comment on line 76. 

I know I did this, but it is weird to have static fields, and a instance based listener.

kristofbolyai: Doesn't have to be fixed here, though.

common/src/main/java/com/wynntils/features/user/overlays/ObjectivesOverlayFeature.java
magicus: Yes.

common/src/main/java/com/wynntils/wynn/model/scoreboard/objectives/ObjectiveListener.java
magicus: This was fixed as part of the ObjectiveManager PR.

common/src/main/java/com/wynntils/wynn/model/bossbar/BossBarModel.java
magicus: It used to be an enum, but I removed it. Here's the issue:
We need to make a connection between BossBars (models) and BossBarOverlays (feature). So when CustomBarsOverlayFeature gets a BossBarAddedEvent, it should create the corresponding overlay.

But, the old style made this by propagating a global enum throught the TrackedBar class. But this is mixing different abstraction levels, since now the BossBarHandler suddently gets a dependency on an enum listing known types of boss bars.

But I agree that this is not ideal either. I'll see if I can come up with something better.

magicus: @kristofbolyai Thanks for the pushback. :) I was a bit lazy. It is now much better!

kristofbolyai (approved):
Looks good. Obviously, it is harder to see through the design changes at this point, as it is a bit of a mix. But I think it looks good so far.

PR #914, opened 2022-12-21
feat: New ActionBarHandler, allowing for more and better action bar modelling

magicus:
New ActionBarHandler, allowing for more and better action bar modelling.


magicus:
As a follow up, I'd like to see:

* Hide the power special segment like how health/mana is hidden
* Add a custom sprint bar, with option to hide the segment
* Ability to always hide coordinate segment (or maybe tie this to "info bar 7" -- which maybe should be more explicit coordinates nevertheless?)

magicus:
I also think we can predict when sprint will end, even if it does not show up on the action bar (e.g. because we are casting a spell). Here are some timestamps for when the sprint bar dropped one step. It seems to be rather consistent around 3-4 seconds. Possibly there is even an exact formula, given our current speed modifiers etc.

sprint level 13 at 2022-12-22T00:02:28.826611
sprint level 12 at 2022-12-22T00:02:31.249717
sprint level 11 at 2022-12-22T00:02:35.179176
sprint level 10 at 2022-12-22T00:02:38.164891
sprint level 9 at 2022-12-22T00:02:41.216049
sprint level 8 at 2022-12-22T00:02:45.153578
sprint level 7 at 2022-12-22T00:02:48.186865
sprint level 6 at 2022-12-22T00:02:51.110474
sprint level 5 at 2022-12-22T00:02:55.089859
sprint level 4 at 2022-12-22T00:02:58.089343
sprint level 3 at 2022-12-22T00:03:01.065695
sprint level 2 at 2022-12-22T00:03:05.100501
sprint level 1 at 2022-12-22T00:03:08.151334
sprint level 0 at 2022-12-22T00:03:15.236874

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -0,0 +1,114 @@
...
+
+        String content = event.getMessage().getString();
+        if (content.equals(previousRawContent)) {
+            // No changes, skip parsing
+            if (!content.equals(previousProcessedContent)) {
+                event.setMessage(Component.literal(previousProcessedContent));
+            }
+            return;
+        }
+        ;
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -0,0 +1,113 @@
...
+        if (!matcher.matches()) {
+            WynntilsMod.warn("ActionBarHandler pattern failed to match: " + content);
+            return;
+        }
+
+        // Create map of position -> matching part of the content
+        Map<ActionBarPosition, String> positionMatches = new HashMap<>();
+        Arrays.stream(ActionBarPosition.values()).forEach(pos -> positionMatches.put(pos, matcher.group(pos.name())));
+
+        Arrays.stream(ActionBarPosition.values()).forEach(pos -> {
```
kristofbolyai, line 75: Kind of hard to get this, this code checks the 3 action bar parts, LEFT, CENTER, RIGHT and checks if there are one or more "segments" in it, which we parse? Can one "side" have multiple "segments" in it? Or, did I not get this at all?

kristofbolyai (approved):
Other than me not getting that part of parsing, this looks good.

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
magicus: Yes, correctly. I did not know what to call things. Left is technically always health and right is always mana (or, actually, right can be bubbles when you are underwater, yes?). The center part is the weird thing. It can be coordinates, sprint meter, spell cast indicator, or powder special. At least these are what I know. I called these "segments". Maybe it should rather be "SegmentHandler" or something, I dunno, but now I kind of kidnapped "handler" to mean something different and more specifically.

magicus: so we try to figure out which segment (if any) is matching which part, in practice only for center.

kristofbolyai: Okay. Should be fine then.

PR #919, opened 2022-12-22
fix: Crisper Waypoint Image (no longer odd-looking)

san7890:
image

I'm not sure where the original texture is from, because it is vastly worse than the one that I was able to extract from Legacy's icons. Looks much better now that I've re-acquired the original asset rather than some weird crusty alternate.


Incompleteusern:
They look exactly the same?

san7890:
That's really weird, I tested it out on another branch and the waypoint texture appears to be fixed? This is the crustiness I was referring to, which was definitely present earlier today:

image

(ignore the completely skewed grind spot icon, this is from a screenshot I had lying around of this issue from two weeks ago)

Maybe I had cleaned out some cache while I was fixing my workspace and that fixed it? That's really weird though.

san7890:
Alright, I compiled again using a copy of master and the weird "smearing" issue appears to be gone. I have absolutely no idea what prompted it to come on the first place, but I was toying around with fixing my workspace earlier today and that may have inadvertently fixed some issue I was having on my end? Not sure, but glad to hear it's not a widespread issue at least.

PR #920, opened 2022-12-22
fix: Updates "Version Banner" in Wynntils Settings' Book to be Longer

san7890:
Couldn't think of a better way to describe it, but the problem with having it be `v0.0.1-main-dev-SNAPSHOT` is that it is quite a long name, and this was lightly irritating me whenever I was testing out features that I had coded:
image

So, I edited the sprite (tried to keep it in form with the rest of the texture) to accommodate for this longer version name (I think having it be within the "safe bounds" that the buttons are in works as I flipped through different screens). Looks like this now:

image

The text does look a bit off-center, but at least the sprite isn't being completely trashed by the text now. Looks better to me, if you think it's off somehow let me know and I can try and fix it.


kristofbolyai:
I don't think we need this, only the main branch text is long. Production builds won't be problematic

magicus:
I'd rather say we change the dev build to "0-dev-SNAPSHOT" or something like that.

kristofbolyai:
Closing this as this is not needed. Thank you for your effort though.

PR #928, opened 2022-12-24
fix: fix item detection (blacksmith selling)

ENORMOUZ:
Seems that Wynn does not put the reset colour code in front of items


ENORMOUZ:
It appears that this fix does not work for Set and Fabled items (not testing Mythic and Crafted items because yeah), will try to fix

ENORMOUZ:
Turns out it is because it does not recognise characters like (-) like in Morph-Stardust, (') like in Gale's Force coding a fix

ENORMOUZ:
Now uses itemMatcher.lookingAt() instead of itemMatcher.matches() to hopefully account for all symbols and other stuff, should still be accurate with no false positives

ENORMOUZ:
Will fix later

magicus:
Can you describe what the problem is that you are trying to solve? Is it that items with anything but a-z in their names are not properly detected?

ENORMOUZ:
> Can you describe what the problem is that you are trying to solve? Is it that items with anything but a-z in their names are not properly detected?

Items with "-" and "'" in their name are not detected, coding a fix but busy rn

magicus:
I see you are struggling a bit with this. You can solve this issue by only updating the regexp. That is the proper solution, and the one one that can be accepted. So you need to tweak the regexp.

Try going to regex101.com, paste in the regex and a bunch of real-world strings from Wynncraft, and see so you can make it match.

magicus (approved):

PR #929, opened 2022-12-24
feat: Remove unused translation keys and duplicate functionality

Incompleteusern:


kristofbolyai:
I don't believe these are the only unused ones? Can we maybe get a script to check it?

Incompleteusern:
I haven't seen any keys that I explicitly remember are unused, but idk

magicus (dismissed):

magicus:
I'd love to see a script check this, but... eh... how? I tried to opposite, having assertions to check that we have translations, but detecting left-over translations seem hopelessly hard.

kristofbolyai:
> I'd love to see a script check this, but... eh... how? I tried to opposite, having assertions to check that we have translations, but detecting left-over translations seem hopelessly hard.

I can think of a fairly easy way. I will get back to this.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -240,16 +234,7 @@ private void renderPois(PoseStack poseStack, int mouseX, int mouseY) {
 
     @Override
     public boolean mouseClicked(double mouseX, double mouseY, int button) {
-        if (button == GLFW.GLFW_MOUSE_BUTTON_RIGHT) {
```
magicus: Why do you remove this?

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -91,13 +91,7 @@ protected void init() {
                                 .append(Component.translatable("screens.wynntils.map.help.description6")),
                         Component.literal("- ")
                                 .withStyle(ChatFormatting.GRAY)
-                                .append(Component.translatable("screens.wynntils.map.help.description7")),
```
magicus: And these are not unused -- they merele become so since you remove functionality below..!?!

magicus (changes requested):

kristofbolyai:
```
screens.wynntils.settingsScreen.featureList
screens.wynntils.settingsScreen.title
screens.wynntils.settingsScreen.uncategorized
screens.wynntils.wynntilsMenu.close
screens.wynntils.wynntilsMenu.openOverlayMenu
screens.wynntils.wynntilsMenu.openSettingMenu
```
These are the unused keys currently, I think.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
Incompleteusern: its duplicate functionality for centering

Incompleteusern:
https://github.com/Wynntils/Artemis/pull/929/commits/f450dbbbf832a1e91213e16f5b805dbebc4928e7 opts to resolve the fact that there are two keybinds for centering by placing the functionality only in the focus bar

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
kristofbolyai: How is it duplicate? I don't get it. One time we center around the player, one time, the waypoint.

kristofbolyai:
> [f450dbb](https://github.com/Wynntils/Artemis/commit/f450dbbbf832a1e91213e16f5b805dbebc4928e7) opts to resolve the fact that there are two keybinds for centering by placing the functionality only in the focus bar

I think we should have both ways to do it? Yes, we have the button. But the mouse bindings are useful too

Incompleteusern:
I don't see the point in having two ways besides being more confusing.

kristofbolyai:
> I don't see the point in having two ways besides being more confusing.

Some people might prefer using the mouse, and for beginners, there is the button.

Incompleteusern:
I still dislike it but its probably fine, will revert change.

magicus (approved):

PR #931, opened 2022-12-24
fix: Powder Special charge bar should only be reset when coordinates are shown

### LINKED ISSUE #930: Powder Ability Bar flickers due to action bar changing
This bug might be relevant in all action bar handlers. Should we do something in the handlers to handle this, or leave it to the implementations?

magicus:
Fixes #930. This was a regression due to the action bar refactoring.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
@@ -84,8 +84,10 @@ public final class ActionBarHandler extends Handler {
                             lastSegment.removed();
                         }
                         lastSegments.put(pos, segment);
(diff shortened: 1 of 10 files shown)


kristofbolyai (approved):
Looks good! I assume you tested this? (Maybe we should also clear the bar when we change held item, not sure if you wanna do it now)

magicus:
I tested briefly, so it was kept while sprinting and casting spells. That is how it was before.

magicus:
To be clear: this fixes the regression. If we should add some additional resets, such as weapon change, class change etc, it needs to be done separately. I'm not sure it's needed though, in most cases the compass segment will likely show up, and trigger this reset anyhow. But it is probably possible to trick this if you are quick and clever, and maybe we should try to fix that.

PR #933, opened 2022-12-24
chore: IntelliJ inspection profile fixes

magicus:
This will update the Artemis profile, add a new Artemis-mini profile (which basically should be clean at all times), and fixes the stuff that has regressed in Artemis-mini since I created it.


kristofbolyai (approved):
Do we have to set it manually?

kristofbolyai:
If so, maybe add docs?

magicus:
Yes, you need to run "Analyze code" manually. 

I'm thinking of seeing if I can reshape this into a Qodana profile, to maybe run Qodana with "warnings" and "errors" or something like that, so that anything that triggers in the "mini" profile will get more or less a blocker. I don't know if or how that is possible though.

magicus:
In the meantime, I'm planning to run it every now and then instead.

PR #935, opened 2022-12-25
fix: Our command manipulation was broken by #925

magicus:
Argh. This one nearly drove me crazy. But now it works, confirmed with our /toggle present, no /wcl present, /bobby and /ccalc present from Bobby and ClientCommands mod respectively.


kristofbolyai (approved):
Such small change with such huge time spent. Thanks!

PR #936, opened 2022-12-25
feat: Chat show real name

ENORMOUZ:
Don't know why it included the past 20 commits lol


ENORMOUZ:
Still need to complete guild chat colours, add a config option for the feature, and add some stuff to the simulate command

ENORMOUZ:
Tbh can't really change guild colour to default (blue) so it will remain as yellow, if anyone can do it will be appreciated

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -159,6 +162,26 @@ private int donateLink(CommandContext<CommandSourceStack> context) {
         return 1;
     }
 
+    private int simulate(CommandContext<CommandSourceStack> context) {
```
magicus: Did you intend for this to go into production, or is it debug code that is left behind?

magicus (changes requested):

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
ENORMOUZ: Originally, I meant for it to go into production but now thinking about it, there isn't really a way to implement it that would be useful to mod developers. Do you have any ideas? If not, it can be removed.

magicus: I don't even understand the purpose of this command. Please explain how you think this should be used?

common/src/main/java/com/wynntils/features/user/ChatShowRealNameFeature.java
```diff
@@ -0,0 +1,93 @@
...
+    private static String eventOriginalColorMessage;
+
+    @SubscribeEvent
+    public void onChatMessage(ChatMessageReceivedEvent event) {
+        eventOriginalMessage = event.getOriginalMessage();
+        eventOriginalColorMessage = event.getOriginalCodedMessage();
+        System.out.println("Original Coloured message: " + event.getOriginalCodedMessage());
+        if (event.getOriginalCodedMessage().contains("§3[")) {
+            for (Component siblingMessage : event.getOriginalMessage().getSiblings()) {
+                System.out.println("Sibling message: " + siblingMessage);
```
magicus: What's this? Just tons of debug output left. This code is nowhere near ready to be merged. Please set this PR as draft or withdraw it until it is ready to review.

magicus (changes requested):
You should not request review of a PR until it is in the state that you think the code is ready to be commit, as it is.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
ENORMOUZ: I thought that it could be used by developers to test features that needed a chat output, to save time (for example, they could simulate text of them completing a Raid instead of actually doing it,saving a few minutes.

magicus: I think you'll just have to write such a command yourself if you really need to test it, and then throw it away.
*Possibly* we could introduce some kind of "DebugCommands" similar to DebugFeatures. In any case this kind of stuff to not belong in WynntilsCommand.

common/src/main/java/com/wynntils/features/user/ChatShowRealNameFeature.java
```diff
@@ -0,0 +1,79 @@
...
+public class ChatShowRealNameFeature extends UserFeature {
+    // credits to avomod for the nickname detection method
+    private static Component eventOriginalMessage;
+    private static String eventOriginalColorMessage;
+
+    @SubscribeEvent
+    public void onChatMessage(ChatMessageReceivedEvent event) {
+        eventOriginalMessage = event.getOriginalMessage();
+        eventOriginalColorMessage = event.getOriginalCodedMessage();
+        addRealNameToMessage(event.getOriginalMessage(), event.getOriginalMessage());
```
magicus, line 24: Why do you pass the argument twice?

ENORMOUZ: Oh, for some reason merging overwrote my changes, I'll see what else got overwritten.

ENORMOUZ: Nevermind, this is intended as the method is recursive and will use other values for the arguments later on.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
ENORMOUZ: Removed.

common/src/main/java/com/wynntils/features/user/ChatShowRealNameFeature.java
magicus: Why do you make it recursive?! :-o We try to use an iterative style in the mod; java is not really suitable for recursive programming.

ENORMOUZ: This is because the part of the message with the name and nickname is nested deep within the message and is in different locations for different messages (for example guild messages/public chat/private messages etc)

kristofbolyai:
Closing this as this is stale (and I don't agree with the whole code being recursive, and I requested it to be changed...).

PR #942, opened 2022-12-27
feat: Detect if NPC dialogue is protected

magicus:
This is a continuation of #767. I learned from HeyZeer0 that if you got the slowness effect while in a dialogue, hostile mobs will not target you. Use this fact to improve the dialogue UI.


magicus:
The detection is now complete. Currently this is only printed in the UI, but I think we should make the NPC dialogue into a screen (so it is clickable with the mouse) if it is protected. That will be a separate PR thought.

This PR also fixes an issue with non-protected selection NPCs (like the Winter Detlas Yeti) where we were unable to select anything.

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -293,15 +303,46 @@ private Component handleChatLine(
...
-            if (dialog.size() > 1) {
-                WynntilsMod.warn("Malformed dialog [#3]: " + dialog);
-                // Keep going anyway and post the first line of the dialog
-            }
-            NpcDialogEvent event = new NpcDialogEvent(dialog, type);
             WynntilsMod.postEvent(event);
+            return;
         }
+
+        if ((System.currentTimeMillis() <= lastSlowdownApplied + SLOWDOWN_PACKET_DIFF_MS)) {
```
kristofbolyai:
```suggestion
        if (System.currentTimeMillis() <= lastSlowdownApplied + SLOWDOWN_PACKET_DIFF_MS) {
```

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -293,15 +303,46 @@ private Component handleChatLine(
...
+                List<Component> dialogToSend = delayedDialogue;
+                delayedDialogue = null;
+                // If we got here, then we did not get the slowdown effect, otherwise we would
+                // have sent the dialogue already
+                NpcDialogEvent event = new NpcDialogEvent(dialogToSend, delayedType, false);
+                lastNpcDialog = dialogToSend;
+                WynntilsMod.postEvent(event);
+                if (type == NpcDialogueType.SELECTION) {
+                    // This is a bit of a workaround to be able to select the options
+                    MutableComponent clickMsg = Component.literal("Open chat and click on the option to select it.")
```
kristofbolyai, line 338: Can we? Do we preserve clickables at this point?

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
magicus: Yes. (And I have tested this)

PR #947, opened 2022-12-29
fix: make WorldWaypointDistanceFeature also tint the displayed icon if the poi have one

HalcyonSuoh:


common/src/main/java/com/wynntils/wynn/model/CompassModel.java
```diff
@@ -21,6 +22,7 @@ public final class CompassModel extends Model {
     private Supplier<Location> locationSupplier = null;
     private Location compassLocation = null; // this field acts as a cache for the supplier
     private Texture targetIcon = null;
+    private CustomColor targetColor = null;
```
kristofbolyai, line 25: This is getting out of hand. I would rather we have a class for all compass data, and we just use it here

HalcyonSuoh: unsure if CompassModel should still act like a proxy of the new class to prevent refactor or other places or return it in it's entirety and do change on other places that use it's data

kristofbolyai: I am not sure what you mean. But l since this is model design question, I'll wait for magicus' opinion tok.

magicus: I'm not entirely sure either what was meant by that comment. But, I think the original change with adding a targetColor here is reasonable. It aligns at least with the targetIcon property. So I think this change is fine.

We might need to take a bird's eye view on the compass model to see if it is properly designed. But for this PR, I don't think anything gets worse at least.

magicus (approved):

PR #952, opened 2022-12-30
feat: Add ItemHandler, Annotators and WynnItem models

magicus:
First part of the item handler rewrite.

This part only adds the new functionality. It is not used anywhere, except in the new debug feature, just so it can be demonstrated. 

The debug feature will append information to the tooltips with what Wynntils knows about the item.

The main design of the system is as follows:

Each ItemStack has an additional field injected, called an "annotation". In that we store what we know about the item. The ItemHandler is responsible for registering "annotators", which will examine a new item stack in order until it is recognized, and an annotation can be produced.

The ItemModel has a registry of annotators. Since we always end up using the FallbackAnnotator, at least, we will always annotate all items, so we will never have to do this twice for any item.

The annotators extract relevant information from the item stack and construct a model-relevant representation of the item, which are subclasses of WynnItem. There are two major groups, GameItems which is basically everything you can throw on the ground in the game, and GuiItems, which is everything else.

### DIFF
--- a/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
+++ b/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.debug;
+
+import com.wynntils.core.features.DebugFeature;
+import com.wynntils.mc.event.ItemTooltipRenderEvent;
+import com.wynntils.utils.KeyboardUtils;
+import com.wynntils.utils.StringUtils;
+import com.wynntils.wynn.handleditems.ItemModel;
+import com.wynntils.wynn.handleditems.WynnItem;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+public class ItemDebugTooltipsFeature extends DebugFeature {
+    @SubscribeEvent(priority = EventPriority.LOWEST)
+    public void onTooltipPre(ItemTooltipRenderEvent.Pre event) {
+        Optional<WynnItem> wynnItemOpt = ItemModel.getWynnItem(event.getItemStack());
+        if (wynnItemOpt.isEmpty()) return;
+        WynnItem wynnItem = wynnItemOpt.get();
+
+        List<Component> tooltips = new ArrayList<>(event.getTooltips());
+        tooltips.addAll(getTooltipAddon(wynnItem));
+        event.setTooltips(tooltips);
+    }
+
+    private List<Component> getTooltipAddon(WynnItem wynnItem) {
+        List<Component> addon = new ArrayList<>();
+
+        addon.add(Component.literal("Wynn Item Type: ").withStyle(ChatFormatting.GREEN));
+
+        List<String> wrappedDescription = Arrays.stream(StringUtils.wrapTextBySize(wynnItem.toString(), 150))
+                .toList();
+        if (!KeyboardUtils.isKeyDown(GLFW.GLFW_KEY_RIGHT_SHIFT) && wrappedDescription.size() > 4) {
+            wrappedDescription = new ArrayList<>(wrappedDescription.subList(0, 3));
+            wrappedDescription.add("...");
+            wrappedDescription.add("Press Right Shift for all");
+        }
+
+        for (String line : wrappedDescription) {
+            addon.add(Component.literal(line).withStyle(ChatFormatting.DARK_GREEN));
+        }
+
+        return addon;
+    }
+}
--- a/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
+++ b/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
@@ -35,7 +35,7 @@ public class ItemDebugTooltipsFeature extends DebugFeature {
     private List<Component> getTooltipAddon(WynnItem wynnItem) {
         List<Component> addon = new ArrayList<>();
 
-        addon.add(Component.literal("Wynn Item Type: ").withStyle(ChatFormatting.GREEN));
+        addon.add(Component.literal("Wynn Item Type:").withStyle(ChatFormatting.GREEN));
 
         List<String> wrappedDescription = Arrays.stream(StringUtils.wrapTextBySize(wynnItem.toString(), 150))
                 .toList();
--- a/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
+++ b/common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
@@ -4,11 +4,11 @@
  */
 package com.wynntils.features.debug;
 
+import com.wynntils.core.components.Models;
 import com.wynntils.core.features.DebugFeature;
 import com.wynntils.mc.event.ItemTooltipRenderEvent;
 import com.wynntils.utils.KeyboardUtils;
 import com.wynntils.utils.StringUtils;
-import com.wynntils.wynn.handleditems.ItemModel;
 import com.wynntils.wynn.handleditems.WynnItem;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -23,7 +23,7 @@ import org.lwjgl.glfw.GLFW;
 public class ItemDebugTooltipsFeature extends DebugFeature {
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public void onTooltipPre(ItemTooltipRenderEvent.Pre event) {
-        Optional<WynnItem> wynnItemOpt = ItemModel.getWynnItem(event.getItemStack());
+        Optional<WynnItem> wynnItemOpt = Models.Item.getWynnItem(event.getItemStack());
         if (wynnItemOpt.isEmpty()) return;
         WynnItem wynnItem = wynnItemOpt.get();
 
--- a/common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
+++ b/common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
@@ -34,11 +34,13 @@ public abstract class ItemStackMixin implements AnnotatedItemStack {
     }
 
     @Override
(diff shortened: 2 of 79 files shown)


common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+public class ItemDebugTooltipsFeature extends DebugFeature {
```
kristofbolyai, line 23: What does this do exactly? How is this different from that other, item related debug feature?

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/CraftedGearAnnotator.java
```diff
@@ -0,0 +1,66 @@
...
+        if (!matcher.matches()) return null;
+
+        CappedValue durability = WynnItemMatchers.getDurability(itemStack);
+
+        List<GearIdentification> identifications = new ArrayList<>();
+        List<Powder> powders = List.of();
+
+        // Parse lore for identifications and powders
+        List<Component> lore = ComponentUtils.stripDuplicateBlank(itemStack.getTooltipLines(null, TooltipFlag.NORMAL));
+        lore.remove(0); // remove item name
```
kristofbolyai, line 44: Can this crash with some weird 0 length lores?

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
```diff
@@ -26,4 +32,14 @@ private Component redirectGetHoveredName(ItemStack instance) {
         ItemTooltipHoveredNameEvent event = EventFactory.onGetHoverName(instance.getHoverName(), instance);
         return event.getHoveredName();
     }
+
+    @Override
+    public ItemAnnotation getAnnotation() {
```
kristofbolyai, line 38: Do these "need" to be made `@Unique` too?

common/src/main/java/com/wynntils/wynn/handleditems/WynnItem.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.wynntils.handlers.item.ItemAnnotation;
+import java.util.HashMap;
+import java.util.Map;
+
+public class WynnItem implements ItemAnnotation {
+    private Map<Class<?>, Object> cache = new HashMap<>();
+    private boolean searched = false;
+
+    public <T> T getCached(Class<T> clazz) {
```
kristofbolyai, line 15: What is this used for?

kristofbolyai: (And do we GC this..?)

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/DungeonKeyAnnotator.java
```diff
@@ -0,0 +1,52 @@
...
+public final class DungeonKeyAnnotator implements ItemAnnotator {
+    private static final Pattern DUNGEON_KEY_PATTERN = Pattern.compile("(?:§.)*(?:Broken )?(?:Corrupted )?(.+) Key");
+
+    public static Matcher dungeonKeyNameMatcher(Component text) {
+        return DUNGEON_KEY_PATTERN.matcher(WynnUtils.normalizeBadString(text.getString()));
+    }
+
+    @Override
+    public ItemAnnotation getAnnotation(ItemStack itemStack) {
+        Matcher keyMatcher = dungeonKeyNameMatcher(itemStack.getHoverName());
```
kristofbolyai, line 28: Okay so, we remove the WynnItemMatchers, I assume, and put everything in item stacks. Not sure if there will be a bit of back-fire where it would be useful to have WynnItemMatchers

common/src/main/java/com/wynntils/wynn/handleditems/items/game/AmplifierItem.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.handleditems.items.game;
+
+import com.wynntils.wynn.handleditems.properties.NumberedTierItemProperty;
+
+public class AmplifierItem extends GameItem implements NumberedTierItemProperty {
```
kristofbolyai, line 9: All these item classes look like they could be a record. Can records do inheritance?

common/src/main/java/com/wynntils/wynn/handleditems/properties/CountedItemProperty.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.handleditems.properties;
+
+public interface CountedItemProperty {
```
kristofbolyai, line 7: Item properties could be functional interfaces?

kristofbolyai (commented):
Looks good

common/src/main/java/com/wynntils/features/debug/ItemDebugTooltipsFeature.java
magicus: Screen Shot 2022-12-31 at 16 15 44

magicus: It is mostly to showcase the new framework, but I have also had some use of it when debugging. It will basically show you everything Wynntils knows about the item. If it is too long, it will be cut off unless you press right shift, since gear items with lots of identification can easily fill the screen otherwise.

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/CraftedGearAnnotator.java
magicus: No, since it always contains the name as first line of the tooltip. (That's why we remove it). But I can make some defensive programming if we have interaction with bad mods.

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
magicus: I can't I fully understand the sponge documentation for the `@Unique` annotation, but I guess it does not hurt. Bit worried that they claim that they will "rename" it, how will method resolution work in that case?

common/src/main/java/com/wynntils/wynn/handleditems/WynnItem.java
magicus: It is GC:d when the ItemStack containing this annotation is GC:ed. That's fine for us. They will be GC:ed when they no longer are present in the inventory or in an open chest.

magicus: It is used to store computationally expensive stuff that is calculated from this WynnItem. Like Tooltip (not yet implemented), and TextOverlay (implemented). So it's free for features etc "downstream" to use to store stuff that is derived from the WynnItem, but that you do not want to repeatedly recalculate. Since the class is the key to the map, you can always put your "own" storage in the object.

magicus: The "searched" flag is to signal when searching for items in the bank.

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/DungeonKeyAnnotator.java
magicus: I don't think so. The idea is that the responsibility of determining what there is to know about an item is the responsbility of the ItemModel. If some feature would need to know something more from an item, it should ask the ItemModel to provide it instead.

common/src/main/java/com/wynntils/wynn/handleditems/items/game/AmplifierItem.java
magicus: No, records cannot do inheritance :-(

Otherwise I agree that they look very record-y. Unfortunately records is only half of the story about what Java is missing. *Some* cases where classes are simple data storage units fits as records, but not all. Apparently this was a concious (but contended!) design decision. I'm not sure I fully understand it, nor agree with it. But that's where we are. So the WynnItems are "old-school records"...

common/src/main/java/com/wynntils/wynn/handleditems/properties/CountedItemProperty.java
magicus: I assume you mean "tagged `@FunctionalInterface`". They already *are* functional interfaces, since by definition any interface with a single method is a functional interface.

However, that is mostly by chance. There's nothing inherently wrong if any of these would actually need two methods, perhaps a `String getUnit()` as well. That indicates that they are not functional interfaces by "heart". In contrast with e.g. `Runnable` which would be *pointless* if it had another method. 

So no, I don't think they should be tagged `@FunctionalInterface`.

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/CraftedGearAnnotator.java
```diff
@@ -0,0 +1,70 @@
...
+            String unformattedLoreLine = WynnUtils.normalizeBadString(loreLine.getString());
+
+            // Look for Powder
+            if (unformattedLoreLine.contains("] Powder Slots")) {
+                powders = Powder.findPowders(unformattedLoreLine);
+                continue;
+            }
+
+            // Look for identifications
+            Matcher identificationMatcher = GearAnnotator.ITEM_IDENTIFICATION_PATTERN.matcher(unformattedLoreLine);
```
P0keDev, line 58: crafted items are going to need their own regex for identifications, since the format is different from regular gear (in other words, this matcher will never succeed)

P0keDev (dismissed):
looks good overall, i like this design

common/src/main/java/com/wynntils/wynn/handleditems/annotators/game/CraftedGearAnnotator.java
magicus: I started with a more ambitious goal in mind than what the code currently does, to handle crafted gear like we handle normal gear. But I realized after a while that this is more work than I expected. I still want to revisit this later on, but I might need to have a clear use in mind first.

I'll add a comment that this needs to be fixed, but for the rest I'll leave it as it is.

PR #955, opened 2022-12-31
refactor: Create HorseManager, EmeraldManager and clean up ItemProfilesManager

magicus:


magicus:
Unfortunately Github does not realize that ItemProfilesManager has moved, and been changed. What I have done is:
1) Make the IdentificationOrderer singleton instead be a local field to ItemProfilesManager.
2) Wrap all calls to the IdentificationOrderer with outwards-facing calls to ItemProfilesManager instead.
3) Move some functions from utility classes into the manager instead.

Here is a gist with the diff between the old version and the new:
https://gist.github.com/magicus/0475bf702f321ff681e8ba00c294a986

kristofbolyai (dismissed):

PR #961, opened 2023-01-02
feat: Use new item model for gear tooltips

magicus:


magicus:
I renamed ChatItemModel to GearItemManager, and then added more functionality to it (basically every way to construct a GearItem from difference sources). Unfortunately Github does not recognize this move, which makes the comparison a bit more difficult. :-(

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
```diff
@@ -4,19 +4,49 @@
...
+import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.TextColor;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
 
 @FeatureInfo(stability = Stability.STABLE, category = FeatureCategory.TOOLTIPS)
 public class ItemStatInfoFeature extends UserFeature {
     public static ItemStatInfoFeature INSTANCE;
 
+    private static final NavigableMap<Float, TextColor> COLOR_MAP = new TreeMap<>();
```
kristofbolyai, line 41: There are 2 instances of this COLOR MAP. Do we need it in both of the places?

kristofbolyai (approved):
Overall looks good

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
magicus: No. It is my intention to rewrite this code further. The problem here is that the responsibility for generating the item lore is divided across different classes, which should have nothing to do with it. Instead of breaking abstraction barriers, I chose to copy the code for now.

magicus: If you are not happy with this, I can open it up from one of the places and reference it from the other, but I'd rather not.

kristofbolyai: It's fine for now then, you know the problem with it

PR #964, opened 2023-01-02
feat: Wynntils item functionality will now work in Trade Market

magicus:
Make a more lax name matching, and only extract name once, to match item names in Trade Market as well

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/item/ItemAnnotator.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemAnnotator.java
@@ -8,5 +8,5 @@ import net.minecraft.world.item.ItemStack;
 
 @FunctionalInterface
 public interface ItemAnnotator {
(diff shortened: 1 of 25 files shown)


kristofbolyai:
Is this optimization, or what does this do?

magicus:
It is mostly a fix/feature, but also an optimization. :-)

Trade Market names have these weird `À` character in the name. We need to strip them out to match. This was just done for a few stacks before (matching where it was done with the old framework). So it is a fix/feature to have this work for all items.

The optimization is that we extract and normalize the name string once for each item, and not have every annotator do that. The code gets cleaner, and there is no risk we forget to normalize.

kristofbolyai (approved):
LGTM although PR title is a bit confusing

magicus:
I tried to write the PR title so it would make sense in the changelog. Apparently I failed. :) Hope this was better.

PR #965, opened 2023-01-03
feat: Add timeout per individual NPC chat

magicus:
This way multiple confirmationless chats won't stack up


common/src/main/java/com/wynntils/features/user/overlays/NpcDialogueOverlayFeature.java
```diff
@@ -302,18 +305,15 @@ private void renderDialogue(PoseStack poseStack, List<String> currentDialogue, N
...
-            }
+            if (currentDialogue.isEmpty() && confirmationlessDialogues.isEmpty()) return;
 
-            renderDialogue(
-                    poseStack,
-                    NpcDialogueOverlayFeature.this.currentDialogue,
-                    NpcDialogueOverlayFeature.this.dialogueType);
+            List<String> allDialogues = new ArrayList<>(currentDialogue);
+            confirmationlessDialogues.forEach(d -> {
+                allDialogues.add("");
```
kristofbolyai, line 312: If there is only 1, there will be a newline?

magicus: Yes... If you can write this without it and without making a mess out of it, please suggest. I'd need something like String.join but for a list.

kristofbolyai: I dont think anything clearner could happen other than `if (allDialogues.Length > 0 && allDialouges[0].isBlank()) allDialogues.remove(0)`

kristofbolyai: In fact we dont need to check list length as it is checked before

magicus: Well that won't work. If there is a normal dialogue, and a single confirmationless dialogue, there need to be a space inbetween. The only case is if there is no normal dialogue, and just a single confirmationless dialogue.

magicus: I fixed it anyway.

PR #977, opened 2023-01-06
feat: Hide irrelevant Vanilla stuff in tooltips

magicus:


P0keDev (approved):

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
```diff
@@ -33,6 +36,29 @@ private Component redirectGetHoveredName(ItemStack instance) {
         return event.getHoveredName();
     }
 
+    @Redirect(
```
kristofbolyai, line 39: I plan to start using https://github.com/LlamaLad7/MixinExtras/ soon, it will improve mixin compatibility greatly. These 2 mixins are a good candidate for their first time use. `LlamaLad7`, the developer, is currently working on adding a way to access any local variable in the current frame, working with any injection type.

magicus: I looked at it, and I agree, it seem nice. I'll push these as-is, and then you can convert them to "mixinextras".

marek-somm:
When the option "Hide Irrelevant Vanilla Tooltip Info" is disabled, the tooltip data of weapons, armor and accesories is still not shown

magicus:
Yeah, that is an old bug. But it will be fixed by https://github.com/Wynntils/Artemis/pull/990.

PR #980, opened 2023-01-06
fix: Make spotless update the copyright year properly in headers

magicus:


kristofbolyai (approved):

kristofbolyai (dismissed):

build.gradle
```diff
@@ -96,6 +96,7 @@ allprojects {
             palantirJavaFormat('2.26.0')
             trimTrailingWhitespace()
             endWithNewline()
+            ratchetFrom('main')
```
kristofbolyai: what does this do

magicus: Making things break. :-/

This was apparently needed for Spotless to have some kind of "baseline" against which it should determine that a file was changed so it could know it needed to update copyright year. But I am not getting it to work. :( The "origin/main" worked locally for me, but not on the CI build here. I thought that a local reference might have worked, but apparently not. :(

kristofbolyai: Wynntils/main?

magicus:
I believe this works now. The Spotless check succeed in my personal fork. It will fail for this PR since the changes in the GHA file is not yet in main and are thus not applied, but that will resolve itself with the next PR.

build.gradle
```diff
@@ -96,6 +96,7 @@ allprojects {
             palantirJavaFormat('2.26.0')
             trimTrailingWhitespace()
             endWithNewline()
+            ratchetFrom('origin/main')
```
kristofbolyai, line 99: What if my main remote is not named origin? I personally hare "Wynntils" as origin..

magicus: I dunno... :-& 

This seems to be tricky to get right. I'd like to see spotless fix years in the header, but I don't know how to get there.

magicus:
I give up on this. Someone else might have more luck...

PR #986, opened 2023-01-07
fix: Do not save "enabled" state of crashed user features

### LINKED ISSUE #978: Crashed features should just get disabled for the session
We should not save the state of a feature that has been disabled due to a crash.

Possibly we need a secondary way to disable features due to crashes, so we do not set the "user enabled# flag.

kristofbolyai:
Fixes #978


magicus:
What about the part of having a command to try and re-enable a crashed feature?

magicus:
Otherwise your changes look good. I just thought it was a nice idea to have a `/feature uncrash <Feature>`, and link to that command from the message, like "[Click here](feature/uncrash) to try to re-enable the feature".

kristofbolyai:
/feature enable does work on these, I just have to add the clickable

magicus:
Do you plan do add the click as part of this PR, or add it on separately?

kristofbolyai:
> Do you plan do add the click as part of this PR, or add it on separately?

I can add it here later

magicus (approved):

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -85,24 +87,31 @@ private static void handleExceptionInEventListener(Throwable t, Event event) {
...
 
-        feature.setUserEnabled(false);
-        feature.tryUserToggle();
+        feature.crash();
 
         WynntilsMod.error("Exception in feature " + feature.getTranslatedName(), t);
         WynntilsMod.warn("This feature will be disabled");
 
-        // TODO: This is a temporary fix for a crash that occurs when an error happens in a client-side message
+        // FIXME: This is a temporary fix for a crash that occurs when an error happens in a client-side message
```
magicus, line 101: @kristofbolyai Btw, is this really temporary? How can we solve this otherwise?

kristofbolyai: I thought about it, not sure atm

PR #987, opened 2023-01-07
feat: add handling for tome and charm items

P0keDev:
- Fixes unidentified item guess tooltips so they ~~no longer conflict with advanced tooltips, and~~ don't add an empty "possibilities" line if no possibilities have been found
- Adds mastery tome and charm support to `GearBoxItem`, by creating corresponding `ItemType` entries 
- Adds `TomeItem` and `CharmItem` - these include some basic parsing, but they're mainly to achieve item highlights

The `TomeProfile` and `CharmProfile` classes are sort of dummy stand-ins right now; eventually I'd want a list of profiles to be populated via the API, just like with item profiles, but that's a separate issue. I added the tome and charm processing to `GearItemManager` because I wanted to avoid too much code duplication, but I could create separate managers that still utilize the helper methods for IDs and rerolls.


magicus:
Looks good. Thanks for doing this!

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
```diff
@@ -66,6 +67,10 @@ public List<Component> getTooltips() {
             return tooltips;
         }
 
+        public List<Component> getBaseTooltips() {
```
magicus: I think this is more properly solved by https://github.com/Wynntils/Artemis/pull/977. That will remove the advanced tooltip part from all tooltips. This fix only solves boxes, not e.g. soul point timer, and it will also make no sense/point with #977 in place.

magicus (dismissed):
I think you can revert the getBaseTooltip part, but apart from that, it looks fine.

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
P0keDev: i'm not sure i agree; #977 is a user feature which can be disabled, and the "correctness" of other features shouldn't rely on that IMO. there *are* genuine use cases for having advanced tooltips on wynn, so i think it's better for individual features to ensure their custom tooltips function properly regardless

magicus: But will this be more "correct" then? You will forcefully remove advanced tooltips for gear boxes. With "my" solution, they can still be available, just in a different location than you are used to.

magicus: I have a suggestion: leave this part out of this PR (it does not belong to tomes/charms anyway), and move it to a separate PR.

I think it is perfectly fine to have advanced tooltips appear in a different order, if you chose to keep them. But if you disagree, I can accept that. However, in that case I think:
1) We do need to *keep* them, not just remove them for one kind of item, and
2) We need to fix it for all items where we right now just append tooltip lines at the end (gear boxes, soul point and server uptime)

So a proper solution would be to locate the advanced tooltip lines and make sure we inject our stuff before those lines. I would accept such a solution, but I think it should be done in a separate PR.

magicus: I fixed this the "proper" way in https://github.com/Wynntils/Artemis/pull/990. Hope you are happy now to remove this change from this PR.

magicus (approved):

PR #989, opened 2023-01-07
feat: Introduce versioned urls.json

magicus:


kristofbolyai (approved):
LGTM, but we need to update the API version of urls.json too

magicus:
Actually, we need to wait a while before doing that. We must make sure that everyone has upgraded to a release that contains this fix. Otherwise their mod will barf at the updated online urls.json. In contrast, any version of the mod that includes this fix will continue to work, it will just prefer the bundled version (which is at version number 1) over the online (which, as missing a number, is considered version 0).

magicus:
But after that, we will be free to update the online version without ever breaking the clients, as long as we bump the version number.

kristofbolyai:
> Actually, we need to wait a while before doing that. We must make sure that everyone has upgraded to a release that contains this fix. Otherwise their mod will barf at the updated online urls.json. In contrast, any version of the mod that includes this fix will continue to work, it will just prefer the bundled version (which is at version number 1) over the online (which, as missing a number, is considered version 0).

I meant adding the online one the version 1. That should not break. But okay, we can wait.

magicus:
I'm still not sure what you mean. We can't add a version number to the online one, until everyone has upgraded their client.

PR #990, opened 2023-01-07
feat: Make sure advanced tooltip lines are always at the bottom

magicus:


P0keDev (dismissed):
looks good, but would this conflict with the "hide advanced tooltips" feature when the advanced setting is enabled? or would the checks for the info lines prevent that

magicus:
I think it will work fine together, we will just fall through, but it will cause a bit of unnecessary performance loss since we will need to scan the entire tooltip looking for the lines that will never appear. I can't find any other good way of doing this, though. :(

kristofbolyai (approved):

PR #998, opened 2023-01-08
refactor: Various cleanups

magicus:
Most of them suggested by IntelliJ. A few inspired by not so nice things pointed out by IntelliJ, but not directly suggested by them.


common/src/main/java/com/wynntils/wynn/objects/profiles/ToolProfile.java
```diff
@@ -7,14 +7,7 @@
...
 
-public class ToolProfile {
-    private final ToolType toolType;
-    private final int tier;
-
-    public ToolProfile(ToolType toolType, int tier) {
-        this.toolType = toolType;
-        this.tier = tier;
-    }
+public record ToolProfile(com.wynntils.wynn.objects.profiles.ToolProfile.ToolType toolType, int tier) {
```
kristofbolyai, line 10: Did you just convert a class that is meant to be parsed with gson into a record? That won't work. (where is the data from?)

magicus: I agree that that won't work. And that is not what I am doing. The "ToolProfile" is perhaps not entirely named according to our standards. That depends on if you interpret the `-Profile` as *how the class is used* vs *where we get the data from*. It aligns with the former, but not the latter. The Profile records are created by the `ToolAnnotator` class when a tool item is parsed.

kristofbolyai: Oh. Can we reserve the "*Profile" names for API/GSON deserialization (or alternatively come up with a some other name for it)?

magicus: Yeah... I've been thinking a bit about different names, but not found any good ones. I think "instance" is a good name for a specific item of some kind, but I keep coming back to "class" or "type" for the "general" concept, and it is so vague. "Profile" is rather odd, but it has become a bit of our nomenclature. Synonym search gives "sort", "variety", "kind" as possible alternatives. 

Hm. Maybe "Kind" would be okay? Like in "ToolKind"?

magicus: @kristofbolyai Is this question on naming a blocker for this PR?

kristofbolyai: I would like to get this fixed, but not blocking.

kristofbolyai (dismissed):

PR #999, opened 2023-01-08
fix: i18n file should never have color codes

kristofbolyai:


magicus:
That's another sanity check we should have on the i18n file... Feels like it's getting closer to the point where we need a tool to verify it.

magicus (approved):

PR #1000, opened 2023-01-09
feat: Add /wynntils status to check if Wynntils services are up

magicus:


kristofbolyai (approved):
PR this to API repo as well before merging.

magicus:
You mean with the update urls.json? Have we given users enough time to update?

magicus:
@kristofbolyai https://github.com/Wynntils/WynntilsWebsite-API/pull/29
It's been three days since the announcement. We can probably push it now, or wait a bit more. Nothing will break if we don't update the online resource.

PR #1002, opened 2023-01-09
feat: Multiple performance optimizations

kristofbolyai:
This is an umbrella PR containing multiple performance optimizations, that has been individually code reviewed. It contains at least the following fixes:

- [x] `MinimapOverlay#renderPois` array allocation and sorting in render code
- [x] `EmeraldManager#getEmeraldCountInContainer` is run in render code in `InventoryEmeraldCountFeature#onContainerRender`
- [x] `WynnUtils.normalizeBadString` takes an unproportionally large part of that time. Is it really needed? Can it be made faster?
- [x] `ItemFavoriteFeature.isFavorited` is called from rendering, and is expensive due to `getUnformattedItemName()`. A suggestion here would be to use the `WynnItem` cache instead.
- [x] `PlayerGhostTransparencyFeature.onPlayerArmorRender` and `onTranslucentCheck` are both done on rendering, and calls `isPlayerGhost`, which needs to to call `getPlayersTeam` which is expensive.
- [x] Look into batch rendering of POIs (and other frequently rendered textures, item highlights?). Is it more efficient if we only upload our buffer once we added every POI?
- [x] Map POIs spend a lot of time sorting in prority/importance each rendering. See if we can get rid of this by utilizing Z value to make sure more important icons are rendered on top of less important ones.


Jimmy-sheep:
I want to advice somethings about QoL:
- [ ] Make the mouse scroll not scroll down whole page in configuration page, maybe each scroll just scroll down 1/3 of page.(Also can prevent the cyan title appeared at the bottom of the page (Picture 1)).
- [ ] Make tooltip `Quick Guide` in main map also have `tooltip fitting` function, so it won't appear outside the screen (Picture 2).
- [ ] Add preview for `Corrupted Bar`, `Focus Bar`, `Awakened Bar`, `Blood Pool Bar` and `Mana Bank Bar` in overlays test-mode.
- [ ] Make the decimal less in functions of Info Box, like `%xp_pct%`, or `%xp_raw%` etc. (Picture 3).
- [ ] Add more textures of others bars(like `Powder Ability Bar`, so users can decide the texture that same as `Health Bar` or `Mana Bar`.

cyan_title
cut_guide
ridiculous_decimal

magicus:
@Jimmy-sheep These are all good suggestions, but they do not belong in this optimization pull request. 

Either open these as issues (you can tag the,m as enhancements if you don't think they are bugs), one separate issue per report. Or go to our feedback site: https://feedback.wynntils.com/

Jimmy-sheep:
> @Jimmy-sheep These are all good suggestions, but they do not belong in this optimization pull request.
> 
> Either open these as issues (you can tag the,m as enhancements if you don't think they are bugs), one separate issue per report. Or go to our feedback site: https://feedback.wynntils.com/

Oh, I will submit these to the feedback site, thanks.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
```diff
@@ -213,33 +212,30 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
...
 
-        pois.addAll(Models.Map.getServicePois());
-        pois.addAll(Models.Map.getCombatPois());
-        pois.addAll(Models.Map.getLabelPois());
+        pois = Stream.concat(pois, Models.Map.getCombatPois().stream());
+        pois = Stream.concat(pois, Models.Map.getLabelPois().stream());
+        pois = Stream.concat(pois, MapFeature.INSTANCE.customPois.stream());
 
-        pois.addAll(MapFeature.INSTANCE.customPois);
+        // Make sure compass and player pois are on top
```
magicus: @kristofbolyai Is this still a relevant/correct comment? I thought the order of Pois were determined by their Z (or Y or whatever) position in this branch?

kristofbolyai: The comment probably stayed after I removed sorting

magicus: Ok if I remove it?

kristofbolyai: Yes, thanks

magicus:
Saving unimplemented ideas in https://github.com/Wynntils/Artemis/issues/1033 for posterity. Not sure they should be fixed, but we should not forget them.

common/src/main/java/com/wynntils/gui/screens/maps/MainMapScreen.java
magicus: I did anyway :)

magicus (approved):
I think this looks good now. Let's do it!

PR #1005, opened 2023-01-10
feat: totem tracking

ryanzhoudev:
Mostly done! Highlights and the tracking overlay all work. The overlay is positioned just below the territory attack timer overlay.

Hoping to add a few more things though, like the distance option discussed in the dev channel. So reviews appreciated, don't merge though please


magicus:
Can you post a screenshot on how it looks in the latest iteration?

ryanzhoudev:
image

ryanzhoudev:
image

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+
+public class ShamanTotemTrackingFeature extends UserFeature {
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    private final ShamanTotemTimerOverlay shamanTotemTimerOverlay = new ShamanTotemTimerOverlay();
+
+    @Config
+    public boolean highlightShamanTotems = true;
+
+    @Config
+    public static ChatFormatting totem1Color = ChatFormatting.WHITE;
```
kristofbolyai: Okay, this might just be me. But god, I hate numbers in variable names (that are not in the end). Can you name this `firstTotemColor` and the others the following way?

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+        // Make or get scoreboard to set highlight colors
+        Scoreboard scoreboard = McUtils.mc().level.getScoreboard();
+        if (!scoreboard.getTeamNames().contains(TOTEM_HIGHLIGHT_TEAM_BASE + totemNumber)) {
+            scoreboard.addPlayerTeam(TOTEM_HIGHLIGHT_TEAM_BASE + totemNumber);
+        }
+        PlayerTeam team = scoreboard.getPlayerTeam(TOTEM_HIGHLIGHT_TEAM_BASE + totemNumber);
+        team.setColor(color);
+
+        scoreboard.addPlayerToTeam(totemAS.getStringUUID(), team);
+        totemAS.setSharedFlag(6, true); // Makes the totem glow
```
kristofbolyai: I am lost here. Why not just apply a simple glow effect to the totem entity? Why do we need all these team magic?

ryanzhoudev: Without a team, the glow will always be white

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
ryanzhoudev: Sure, will do in a sec

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @SubscribeEvent
+    public void onTotemDestroy(TotemRemovedEvent e) {
+        if (!highlightShamanTotems) return;
+
+        // Teams should be destroyed and recreated every cast to allow the user to change totem highlight colors without
+        // having to reload the feature
```
kristofbolyai, line 94: See my comments above. Even if we keep the scoreboard magic, I suggest you delete teams in `onConfigChange` (`Feature` override).

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
kristofbolyai: Minecraft... Anyways, can you extract the `6` flag as a final variable? Where did you get this from? I assume this is something along the lines of `SHOULD_TEAM_ENTITIES_GLOW`?

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+import com.wynntils.wynn.event.TotemSummonedEvent;
+import com.wynntils.wynn.objects.ShamanTotem;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.world.entity.decoration.ArmorStand;
+import net.minecraft.world.scores.PlayerTeam;
+import net.minecraft.world.scores.Scoreboard;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class ShamanTotemTrackingFeature extends UserFeature {
```
kristofbolyai, line 42: Add `@FeatureInfo(category = FeatureCategories.OVERLAYS)`

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+                                    .map(shamanTotem -> {
+                                        String prefix =
+                                                switch (shamanTotem.getTotemNumber()) {
+                                                    case 1 -> totem1Color + "Totem 1";
+                                                    case 2 -> totem2Color + "Totem 2";
+                                                    case 3 -> totem3Color + "Totem 3";
+                                                    default -> throw new IllegalArgumentException(
+                                                            "totemNumber should be 1, 2, or 3! (switch in #render in ShamanTotemTrackingFeature");
+                                                };
+                                        String suffix = " (00:" + shamanTotem.getTime() + ")";
```
kristofbolyai: If we know it will always be < 1min, why not do smth like `32s`?

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
ryanzhoudev: I found it in the method that checks if an Entity is glowing, but yeah I'll move it to the top

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,202 @@
...
+                                        String coords =
+                                                " " + shamanTotem.getLocation().toString();
+                                        // Check if we should be saying "Summoned"
+                                        if (shamanTotem.getState() == ShamanTotem.TotemState.SUMMONED) {
+                                            suffix = " Summoned";
+                                            coords = "";
+                                        }
```
kristofbolyai: It feels like there should be an else after this if, which runs the code above.

common/src/main/java/com/wynntils/mc/mixin/MultiPlayerGameModeMixin.java
```diff
@@ -90,4 +90,15 @@ private void attack(Player player, Entity target, CallbackInfo ci) {
...
     }
+
+    @Inject(
+            method = "ensureHasSentCarriedItem",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lnet/minecraft/client/multiplayer/ClientPacketListener;send(Lnet/minecraft/network/protocol/Packet;)V"))
+    private void ensureHasSentCarriedItem(CallbackInfo ci) {
```
kristofbolyai, line 103: We should note this here, for future us: "As of 1.19.3, this seems to be the only method which sends carried item update packets to the server. Please look into this and confirm this is still the case, in future versions."

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -0,0 +1,304 @@
...
+    public void onTotemSpellCast(SpellCastedEvent e) {
+        if (e.getSpell() != SpellType.TOTEM) return;
+
+        totemCastTimestamp = System.currentTimeMillis();
+        summonWeaponSlot = McUtils.inventory().selected;
+    }
+
+    @SubscribeEvent
+    public void onTotemSpawn(AddEntityEvent e) {
+        Delay.create(
```
kristofbolyai: If your goal was to run on next tick, I would use `MinecraftSchedulerManager`. It is far less costly, and we will be doing this a lot.

kristofbolyai: (Delay 1 and `MinecraftSchedulerManager` does the same thing, I think, bug the manager should be much faster in theory.)

kristofbolyai:
Okay, it is getting late, and I have yet to finish reviewing this. I am gonna have to come back to this tomorrow. Anyways, sorry for the lot of reviews, but it is a large addition. Overall, it will be really great :)

ryanzhoudev:
image

Now looks like this with the (00:14) -> (14s)

common/src/main/java/com/wynntils/wynn/event/SpellCastedEvent.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.wynn.event;
+
+import com.wynntils.wynn.objects.SpellType;
+import net.minecraftforge.eventbus.api.Event;
+
+public class SpellCastedEvent extends Event {
```
magicus: Just a grammar note. Is not the paste tense form of "cast" still "cast"? So it should be `SpellCastEvent`?

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
magicus: Actually, don't you think this categorization breaks down a bit? Maybe you're only interested in using this for the totem highlighting, not the overlay timers?

Just a random thought...

common/src/main/java/com/wynntils/wynn/model/SpellModel.java
```diff
@@ -0,0 +1,53 @@
...
+import com.wynntils.core.components.Model;
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import com.wynntils.wynn.event.SpellCastedEvent;
+import com.wynntils.wynn.model.actionbar.event.SpellSegmentUpdateEvent;
+import com.wynntils.wynn.objects.SpellType;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class SpellModel extends Model {
```
magicus, line 18: There seems to be opportunities here to move functionality from QuickCastFeature to this new Model. But maybe that is a separate follow-up project?

common/src/main/java/com/wynntils/wynn/event/TotemActivatedEvent.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.wynntils.wynn.event;
+
+import com.wynntils.mc.objects.Location;
+
+public class TotemActivatedEvent extends TotemEvent {
+    private final int time;
+    private final Location location;
+
+    public TotemActivatedEvent(int totemNumber, int time, Location location) {
```
magicus: By tradition, when we have abstract events, we have put the concrete subclasses as static inner classes. I think it is kind of a nice pattern to follow, so you'd get a `TotemEvent.Activated` instead, etc.

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -0,0 +1,301 @@
...
+        if (entityId == -1) {
+            return new net.minecraft.world.entity.decoration.ArmorStand(McUtils.mc().level, 0, 0, 0);
+        }
+
+        return null;
+    }
+
+    /**
+     * Removes the given totem from the list of totems.
+     * @param totem The totem to remove. Must be 1 or 2.
```
magicus: Totem 3 lives forever..? :-)

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -0,0 +1,301 @@
...
+    private void removeAllTotems() {
+        removeTotem(1);
+        removeTotem(2);
+        removeTotem(3);
+        nextTotemSlot = 1;
+    }
+
+    private int getNextTotemSlot() {
+        if (nextTotemSlot == 3) return 1;
+        return nextTotemSlot += 1;
```
magicus: Are you always alternating between these numbers? What if you cast one totem, wait until it disappears, and cast another. Will the first get number 1 and the second number 2? I might be missing something here...

ryanzhoudev: Totems will always fill the slot with the lowest number. So if one totem is cast, it will be in slot 1. Then it expires and frees up slot 1. Then another one is cast and it will be in slot 1 again. If two are cast together, they will fill 1 and 2.

common/src/main/java/com/wynntils/wynn/event/SpellCastedEvent.java
ryanzhoudev: Yep looks like it. I have no idea where I came up with the word "casted"

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
magicus: Perfect, that is what I wanted to hear. :) So the wraparound above is more of a defensive programming, since you can really never have more than 3 totems?

ryanzhoudev: Yep, three totems max in-game. I think it might be better to have something like `cycleTotemSlot()` though, as opposed to the current way of doing it. I will change it in a second

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -73,7 +73,7 @@ public void onTotemSpawn(AddEntityEvent e) {
                     if (inv.size() < 4 || inv.get(3).getItem() != Items.STONE_SHOVEL) return;
 
                     int totemNumber = nextTotemSlot;
-                    nextTotemSlot = getNextTotemSlot();
```
magicus: Actually, why don't you do like:
`int totemNumber = getNextTotemSlot();`
and let getNextTotemSlot do the cycling as well? If you want to clean things up, that is imho better. It encapsulates the handling of the totem slot into one function, which just provides a simple API: "call me and I guarantee that you will get a usable totem slot number in return!".

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -0,0 +1,232 @@
...
+    public static TotemTrackingDetail totemTrackingDetail = TotemTrackingDetail.COORDS;
+
+    @Config
+    public static ChatFormatting firstTotemColor = ChatFormatting.WHITE;
+
+    @Config
+    public static ChatFormatting secondTotemColor = ChatFormatting.AQUA;
+
+    @Config
+    public static ChatFormatting thirdTotemColor = ChatFormatting.RED;
```
kristofbolyai: These configs should be in the overlay itself

kristofbolyai: as well as `TotemTrackingDetail`

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -0,0 +1,303 @@
...
+public class ShamanTotemModel extends Model {
+
+    private ShamanTotem totem1 = null;
+    private Integer pendingTotem1Id = null;
+
+    private ShamanTotem totem2 = null;
+    private Integer pendingTotem2Id = null;
+
+    private ShamanTotem totem3 = null;
+    private Integer pendingTotem3Id = null;
```
kristofbolyai, line 45: How hard is it to fold these into a list? It would also reduce code duplication.

common/src/main/java/com/wynntils/wynn/model/SpellModel.java
kristofbolyai: I honestly thought this was straight up copied from there?

common/src/main/java/com/wynntils/wynn/objects/SpellType.java
```diff
@@ -39,6 +41,19 @@ public enum SpellType {
...
+
+    private static final boolean[] RLR = {SPELL_RIGHT, SPELL_LEFT, SPELL_RIGHT};
+    private static final boolean[] RRR = {SPELL_RIGHT, SPELL_RIGHT, SPELL_RIGHT};
+    private static final boolean[] RLL = {SPELL_RIGHT, SPELL_LEFT, SPELL_LEFT};
+    private static final boolean[] RRL = {SPELL_RIGHT, SPELL_RIGHT, SPELL_LEFT};
+    // Archer only
+    private static final boolean[] LRL = {SPELL_LEFT, SPELL_RIGHT, SPELL_LEFT};
+    private static final boolean[] LLL = {SPELL_LEFT, SPELL_LEFT, SPELL_LEFT};
+    private static final boolean[] LRR = {SPELL_LEFT, SPELL_RIGHT, SPELL_RIGHT};
+    private static final boolean[] LLR = {SPELL_LEFT, SPELL_LEFT, SPELL_RIGHT};
```
kristofbolyai, line 54: Pretty sure QuickCastFeature has these too

magicus: I think Donkey has been struggling enough with this feature. I suggest we accept some duplication, and leave the unification of totem tracking spell handling with quick cast spell handling as a separate issue. I can probably volunteer to handle it.

kristofbolyai: Yeah I was just making a note, similarly to yours about spell model. No need to handle it now

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
ryanzhoudev: I can move TotemTrackingDetail and it's enum, but I think xTotemColor should stay since it not only does the config for the overlay, but also the highlights on the entity

common/src/main/java/com/wynntils/wynn/objects/SpellType.java
ryanzhoudev: Yea, will refactor QuickCastFeature to use all the new Spell stuff added in this PR, after this is merged

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
kristofbolyai: Oh, well okay, they can stay then

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
kristofbolyai: @DonkeyBlaster Are you doing this?

ryanzhoudev: I can give it a shot in a day or two, don't have time at the moment

Or we could also merge and the folding into a list could be a separate PR

kristofbolyai: I don't want to merge this as there is a performance concern to me. Ill fix it this afternoon

magicus (dismissed):

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
magicus: But that is something different to the issue above, right? Code duplication is more a code hygiene concern. From my PoV, this is acceptable. So maybe if bolyai fixes the perf issue, and then we can merge?

And then follow up by de-duping the code above, and making quick spell using the new spell model.

kristofbolyai: > But that is something different to the issue above, right? Code duplication is more a code hygiene concern. From my PoV, this is acceptable. So maybe if bolyai fixes the perf issue, and then we can merge?
> 
> 
> 
> And then follow up by de-duping the code above, and making quick spell using the new spell model.

Yes, I fix perf, then we fix. Then we fix the code.

kristofbolyai:
@magicus https://github.com/Wynntils/Artemis/pull/1005/commits/ad62d713c94f77c7b5fb6b3741a307c1760ad822 if you are fine with this, we can merge.

kristofbolyai (approved):

magicus (approved):

PR #1006, opened 2023-01-10
feat: Add profiling to annotators and overlays

magicus:


magicus:
A comment before anyone asks: System.currentTimeMillis is extremely cheap, it is an intrinsic that is basically replaced by a single machine code instruction on modern CPUs. So it is not worth trying to toggle profiling on and off, we might as well just always collect profiling data.

kristofbolyai (approved):

PR #1012, opened 2023-01-10
perf: Optimize POI rendering by rendering in batches

kristofbolyai:
From my final testings, rendering POIs on the main map takes around 3x less time.


common/src/main/java/com/wynntils/gui/render/CustomRenderType.java
```diff
@@ -8,14 +8,17 @@
...
 import net.minecraft.client.renderer.RenderType;
+import net.minecraft.resources.ResourceLocation;
 
 public class CustomRenderType extends RenderType {
     // Copied from RenderType.LINE_STRIP and changed the line width from the default
     // to 3
     public static final RenderType LOOTRUN_LINE = RenderType.create(
-            "lootrun",
+            "wynntils_lootrun_line",
             DefaultVertexFormat.POSITION_COLOR_NORMAL,
```
magicus, line 22: What does the string literal do?

magicus (approved):

common/src/main/java/com/wynntils/gui/render/CustomRenderType.java
kristofbolyai: Some sort of id, I just added wynntils if there is a problem with clashing

PR #1019, opened 2023-01-11
perf: renderPois now use streams to greatly reduce allocations

kristofbolyai:


common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
```diff
@@ -218,31 +218,35 @@ private void renderPois(
...
+                                    || (user.isMutualFriend() && renderRemoteFriendPlayers))
+                            .map(PlayerMiniMapPoi::new));
 
-            poisToRender.addAll(Models.Map.getCombatPois());
+            poisToRender = Stream.concat(poisToRender, Models.Map.getCombatPois().stream());
 
             // Reverse order to make sure higher priority is drawn later than lower priority to overwrite them
-            poisToRender.sort(Comparator.comparing(Poi::getDisplayPriority).reversed());
-            for (Poi poi : poisToRender) {
+            poisToRender = poisToRender.sorted(
```
magicus, line 247: In my measurements, the sort took a fair bit of time as well. 
Do you think we could map display priority to Z value so we could use the rendering system to make sure more prioritized icons render on top of less important, instead of sorting this for each frame?

That's a separate optimization, though, so I'll accept this as the arraylist stuff.

magicus (approved):

common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
kristofbolyai: Actually, I've not tried batching and render Y levels. Can you add this into the list of optimizations and I will take a look? I have a feeling this can work.

magicus: I think I said something about the sorting, but I can make sure it is more explicit. The idea with the Z value did not strike me until just right now.

PR #1020, opened 2023-01-11
refactor: Create PlayerModel based on RemoteWynntilsUserInfoModel

magicus:


common/src/main/java/com/wynntils/wynn/model/PlayerModel.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @SubscribeEvent
+    public void onWorldStateChange(WorldStateEvent event) {
+        switch (event.getNewState()) {
+            case NOT_CONNECTED, CONNECTING -> clearUserCache();
+        }
+        if (event.getNewState() == WorldState.WORLD) {
+            clearGhostCache();
+        }
```
kristofbolyai, line 62: Why not include this in the switch above?

common/src/main/java/com/wynntils/wynn/model/PlayerModel.java
```diff
@@ -0,0 +1,122 @@
...
+        loadUser(uuid);
+    }
+
+    @SubscribeEvent
+    public void onAddPlayerToTeam(PlayerTeamEvent.Added event) {
+        UUID uuid = McUtils.mc()
+                .getConnection()
+                .getPlayerInfo(event.getUsername())
+                .getProfile()
+                .getId();
```
kristofbolyai: I am pretty sure this call chain can NPE, I know I used to null-check parts of this before.

kristofbolyai (dismissed):
Apart from the highlighted parts, it looks good.

common/src/main/java/com/wynntils/wynn/model/PlayerModel.java
magicus: Our hades info needs to be updated only when connecting to the wynncraft network, but ghost data becomes irrelevant as soon as you switch to another world.

common/src/main/java/com/wynntils/wynn/model/PlayerModel.java
magicus: We do have a connection, otherwise we cannot be connected to the server. But I can check the rest.

PR #1021, opened 2023-01-11
refactor: Create a FavoritesManager

magicus:
This will also help with performance, since it is now easy to cache the calculated value.


common/src/main/java/com/wynntils/wynn/model/FavoritesManager.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void addFavorite(String unformattedName) {
+        getFavoriteItems().add(unformattedName);
+        revision++;
+    }
+
+    public void removeFavorite(String unformattedName) {
+        getFavoriteItems().remove(unformattedName);
+        revision++;
```
kristofbolyai, line 67: positive revision means favorite, then why do we do this here?

magicus: No, the revision number is always increasing. That is a way to keep track of outdated cache values. Let's say your pouch contain Bloody Noses and Pointy Ears. It is checked for favorites, and marked -1 since the start revision is 1, and it is negative since there is no match. Then you add Pointy Ears to the set of favorites, but your pouch content to not change. But the revision goes to 2.

Now when we check the pouch we se that it is still is at revision 1, so it is stale, and needs to be recalculated, and now we find the pointy ears match and mark it as +2.

I am basically combining two variables, a cache freshness indicator and a boolean into one, by using the positive/negative trick.

kristofbolyai: Got it.

kristofbolyai (approved):

PR #1025, opened 2023-01-12
feat: Proper tick scheduler

magicus:
This will combine the MinecraftSchedulerManager and the Delay class into a single TickScheduler manager. The Delay class was unfortunately costly to create and destroy for each delay. It also waited one more tick than what the argument seemed to apply, hence I've updated the argument in the calls to not change behavior.

I have also restored old behavior to the "next tick" callback, by moving the event from RETURN to HEAD, so it will always need to wait until the next tick, if it is scheduled in this tick.


ryanzhoudev (approved):

common/src/main/java/com/wynntils/commands/LootrunCommand.java
```diff
@@ -403,7 +403,7 @@ public LiteralArgumentBuilder<CommandSourceStack> getBaseCommandBuilder() {
 
     private int screenLootrun(CommandContext<CommandSourceStack> context) {
         // Delay is needed to prevent chat screen overwriting the lootrun screen
-        Delay.create(() -> McUtils.mc().setScreen(WynntilsLootrunsScreen.create()), 1);
+        Managers.TickScheduler.scheduleLater(() -> McUtils.mc().setScreen(WynntilsLootrunsScreen.create()), 2);
```
kristofbolyai, line 406: This could have been 1 tick, I am pretty sure. Doesn't matter though

kristofbolyai (commented):
Looks good overall, thanks

common/src/main/java/com/wynntils/commands/LootrunCommand.java
magicus: I think so to, but out of an abundance of caution I did not want to change the actual behavior in a patch that was hastily merged in the middle of the night. :-) You are welcome to fix it though.

PR #1027, opened 2023-01-13
perf: Make every part of every POI render with buffered rendering

kristofbolyai:
Now we don't have to sort while rendering :). Yes, this PR is huge, for such small change. But in the end, it is worth it, and we should use only buffered rendering from now on. So, this can be seen as the first step towards that plan.


common/src/main/java/com/wynntils/gui/render/buffered/CustomRenderType.java
```diff
@@ -34,9 +34,49 @@ public class CustomRenderType extends RenderType {
...
+                    256,
+                    false,
+                    false,
+                    CompositeState.builder()
+                            .setShaderState(POSITION_TEX_SHADER)
+                            .setTextureState(new TextureStateShard(resource, false, false))
+                            .setTransparencyState(CustomRenderStateShard.SEMI_TRANSPARENT_TRANSPARENCY)
+                            .createCompositeState(false)));
+
+    private static final Function<ResourceLocation, RenderType> POSITION_COLOR_TEXTURE_QUAD =
```
Incompleteusern, line 77: why aren't the first two memoized?

kristofbolyai: We don't have to create instances of the render type, if there is no variation in them. But as soon as there is a changing value, like a texture or a different line width, we have to memoize, to cache them (unless we want to fix the values specifically).

Incompleteusern (approved):

magicus (approved):
I've made a cursory overview and couldn't find anything to object to. The rendering black magic goes over my head, but I assume it is fine.

PR #1031, opened 2023-01-15
perf: Make all overlays use buffers

kristofbolyai:


common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
```diff
@@ -110,8 +110,11 @@ protected MinimapOverlay() {
                     new GuiScaledOverlaySize(DEFAULT_SIZE, DEFAULT_SIZE));
         }
 
+        // FIXME: This is the only overlay not to use buffer sources for rendering. This is due to `createMask`
```
magicus, line 113: Sounds a bit disappointing..? Is not the minimap the most important of all to have the buffered rendering? Or am I missing something?

kristofbolyai: Pois are rendered with buffers, the actual map part and texts, borders etc are small-ish

common/src/main/java/com/wynntils/gui/render/buffered/BufferedFontRenderer.java
```diff
@@ -355,4 +357,110 @@ public void renderText(
...
+    public void renderTexts(
+            PoseStack poseStack,
+            MultiBufferSource.BufferSource bufferSource,
+            float x,
+            float y,
+            List<TextRenderTask> lines) {
+        float currentY = y;
+        for (TextRenderTask line : lines) {
+            renderText(poseStack, bufferSource, x, currentY, line);
+            // If we ask Mojang code the line height of an empty line we get 0 back so replace with space
```
magicus, line 397: I think this is copied from the other renderTexts(). Can you extract this into a common `getLineHeight()` or something?

kristofbolyai: There are other height methods which are only a bit different so it would be a bit confusing imo

common/src/main/java/com/wynntils/gui/render/buffered/BufferedFontRenderer.java
```diff
@@ -355,4 +357,110 @@ public void renderText(
...
+            PoseStack poseStack,
+            MultiBufferSource.BufferSource bufferSource,
+            float x,
+            float y,
+            List<TextRenderTask> toRender,
+            float width,
+            float height,
+            HorizontalAlignment horizontalAlignment,
+            VerticalAlignment verticalAlignment) {
+        float renderX =
```
magicus, line 371: Come to think of it, that remark probably goes for this as well. Seems this exact code is already copied twice in FontRenderer...

And the same for renderY below.

common/src/main/java/com/wynntils/features/user/map/MinimapFeature.java
magicus: Ah, ok.

common/src/main/java/com/wynntils/gui/render/buffered/BufferedFontRenderer.java
magicus: Ok, leave it then. But I think we need to revisit this code and clean it up a bit later on. I always feel like there is a ton of "renderText" which calls other "renderText" ad nauseum until we end up in a "renderText" that actually does something. :-)

common/src/main/java/com/wynntils/gui/render/buffered/BufferedFontRenderer.java
magicus: You can skip this too, then.

common/src/main/java/com/wynntils/gui/render/buffered/BufferedFontRenderer.java
kristofbolyai: Yes, agreed. Ideally, we would get rid of the regular rendering. I think. Not so sure at the moment.

PR #1032, opened 2023-01-15
fix: Check URL map in the correct place instead of too early

magicus:


kristofbolyai (commented):
What does this fix? Also, we seem to be not checking in the correct place, but we can now load without all urls?

magicus:
I ran into issues when I locally added a new URL to my file for testing. 
The problem is that if the locally bundled urls.json includes the new enum, but the cached one don't, we will parse the cached one and complain that it is broken.

But maybe that's just an aesthetic problem. And maybe I'm doing thing worse here..?

kristofbolyai:
What behavior do we want? If we add a local url/enum, we should be warned that remote is not correct,  shouldn't we?

magicus:
Yes, you are right. It's a temporary warning just during development. Somehow I got the idea that it did not work, but that is not the case. It is fine as it is.

PR #1038, opened 2023-01-15
fix: Broken tooltips with "Health" id after optimization update

magicus:


kristofbolyai (approved):
This is somewhat messy, I assume the rewrite will fix it

PR #1039, opened 2023-01-15
fix: Flag events as "posted" if they are processed

P0keDev:
This PR adds a wrapper `WynntilsEvent` class for all our custom events, which lets us add a flag denoting whether the event was posted to the event bus or not. The purpose of this flag is to keep mixins from executing code that would otherwise run regardless of the event getting posted - or in other words, regardless of the player being online Wynncraft. This prevents the possibility of mixin code breaking unrelated, non-Wynncraft things.


magicus:
As I said in discord, I think this is better resolved by just addressing the chat mixin, either by fixing it so the additional code only executes by the recipient of the event, or by adding the isPosted flag (or similar mechanism, perhaps you can return null if the event was not posted?) just to this event.

PR #1041, opened 2023-01-16
feat: Add and use MixinExtras

kristofbolyai:


common/src/main/java/com/wynntils/mc/mixin/plugin/WynntilsMixinPlugin.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.mixin.plugin;
```
magicus: Maybe this is how it is supposed to work with plugins, but my gut reaction is that this should live in `core` instead, so all classes in `mc.mixin` are real, sponge mixin classes. I vaguely remember that there could be problems otherwise. And in any case, I think it sort of breaks a "rule" we have for our code base. It is mixin-related, yes, but it is not a mixin in itself.

fabric/build.gradle
```diff
@@ -96,6 +98,10 @@ jar {
 
 shadowJar {
     configurations = [project.configurations.shadowCommon, project.configurations.shadowImplementation]
+
+    relocate("com.llamalad7.mixinextras", "com.wynntils.mixinextras.shadow")
+    exclude 'META-INF/services/javax.*'
```
magicus, line 105: what javax services do we have? 🤔

common/src/main/java/com/wynntils/mc/mixin/plugin/WynntilsMixinPlugin.java
kristofbolyai: It won't break mixin but I can move it. Where, exactly?

gradle.properties
```diff
@@ -2,6 +2,9 @@
 archives_base_name=wynntils
 enabled_platforms=quilt,fabric,forge
 
+# MixinExtras version
+mixinextras_version=0.2.0-beta.1
```
magicus, line 6: oh, beta. :-/ Are a stable release planned? Have they had a stable 0.1 released?

fabric/build.gradle
kristofbolyai: Forge discord said this is needed, with mixinextras

gradle.properties
kristofbolyai: 0.1 is out but 0.2 has the cool features

common/src/main/java/com/wynntils/mc/mixin/plugin/WynntilsMixinPlugin.java
magicus: `com.wynntils.core.mod` is my suggestion.

magicus: I see it as a "core stuff that is generally needed to support this code base as being a mod".

fabric/build.gradle
magicus: Oookay. What happens if you don't? Do we even have any files in META-INF/services?

kristofbolyai: I'll try without I guess.

kristofbolyai: image
@magicus I will add a reflecting comment.

magicus: Okay, it describes why it is bad to have META-INF/services/javax.* files hanging around. I'm still curious: do we have such a file, or is this just a general guidance to other mods that actually have those files? I tried searching but only found the Forge mods.toml in META-INF. But then again, maybe mixinextras produces it itself?

If we don't have it, we don't need to have that exclude. With that said, maybe something will start producing such files in the future, so we can keep the exclude to avoid future surprises, but if so, I think we should add a comment saying that this is strictly speaking not needed right now.

gradle.properties
magicus: Ok. Any idea when 0.2 will be properly released? I am okay with us being on a beta dependency when we still have an alpha :-), but that will be kind of a release blocker -- I am not willing to mark Artemis as release quality with a beta dependency.

fabric/build.gradle
kristofbolyai: I think mixinextras produces this file so it is needed

magicus: Can you check? If so, a comment about that being the case is also helpful.

gradle.properties
kristofbolyai: Likely soon. It is beta as in author didn't want to make it "release", but released for testing and no apparent issues were reported.

magicus: Okay, sounds good enough.

fabric/build.gradle
kristofbolyai: I could not see it. I updated the comment, but I want to keep it.

kristofbolyai:
I converted the most obvious use cases. MixinExtras seems powerful, but overall, it might be for niche cases, where as our mixins are quite different from other mixins (as we have to provide our own events).

magicus (approved):

magicus:
Will this solve P0kes problem with the chat mixin?

kristofbolyai:
> Will this solve P0kes problem with the chat mixin?
It should. @P0keDev can you try?

P0keDev:
> Will this solve P0kes problem with the chat mixin?

this should fix it yeah, since the real issue was caused by calling `equals` on the message object when it was null

PR #1042, opened 2023-01-16
fix: Use separate buffer for lootrun rendering to prevent crashes

kristofbolyai:


kristofbolyai:
Force merging this as it is trivial.

PR #1045, opened 2023-01-17
fix: Better default config for CustomCharacterSelectionScreenFeature

kristofbolyai:


common/src/main/java/com/wynntils/features/user/CustomCharacterSelectionScreenFeature.java
```diff
@@ -18,7 +18,7 @@
 
 public class CustomCharacterSelectionScreenFeature extends UserFeature {
     @Config
-    public boolean onlyOpenOnce = true;
```
magicus, line 21: Actually, you think this is better? I think the custom char selection screen has been a bit buggy, and the ability to abort it like this seemed like a bit of a saving grace.

And while the vanilla Wynncraft GUI is not the best, I have noticed when I introduce Wynntils to new players (admittedly a small sample size of 2), that the char selection screen was not at all intuitive. So that my suggestion was actually that we should start with it as disabled.

kristofbolyai: I see what you are mentioning. But, if you see this as our current default setting, which is not aimed for new users, then I would say having this on (like legacy) is better. The problem of bugs, we can fix, and profiles will help new users.

magicus: Ok. But we definitely need to move the play button out of the bottom right corner :-) Maybe have a |> triangle directly in the character list to the left? Or, contrary, a "..." to select it for modification, so just a single click starts the character?

kristofbolyai: (double click does select the class)

magicus: Yes, but that was not intuitive for 100% of my test subjects. They single clicked, and then looked confused "now what?" The play button was too far away, did not highlight or anything, etc. A very simple usability study, but still the results were extremely clear.

PR #1047, opened 2023-01-18
fix: Totem tracking improvements

ryanzhoudev:
- Fixes flying totems that are destroyed not erasing the internal totem variable.
- Effectively eliminates chances of accidentally "capturing" someone else's totem timer.
  - This still has a very small chance to occur, if both yours and the other player's totems land in the same spot on the exact same tick, within 1 block of each other.
- Fixes totem timers not being "captured" when casting totems into a wall and they bounce
- Fixes totem timers not being "captured" when casting totems from very high locations (eg. from the top of Cinfras gate)
- Fixes missing space in COORDS option

"captured" refers to when the mod looks for and stores an ArmorStand entity that has the totem's time remaining as it's nametag.

### DIFF
--- a/common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
@@ -131,6 +131,9 @@ public class ShamanTotemModel extends Model {
         Location parsedLocation = new Location(entity.position().x, entity.position().y, entity.position().z);
 
         if (getBoundTotem(entityId) == null && Math.abs(totemCastTimestamp - System.currentTimeMillis()) < 15000) {
+            // Don't unnecessarily check if we aren't even waiting for a totem
+            if (pendingTotem1Id == null && pendingTotem2Id == null && pendingTotem3Id == null) return;
+
             // Given timerId is not a totem, make a new totem (assuming regex matches and we are within 15s of casting)
             // First check if this is actually one casted by us
             List<ArmorStand> toCheck = McUtils.mc()
@@ -142,7 +145,7 @@ public class ShamanTotemModel extends Model {
                                     entity.position().y - 0.1,
                                     entity.position().z - 0.5,
                                     entity.position().x + 0.5,
-                                    entity.position().y + 0.1,
+                                    entity.position().y + 1.7,
                                     entity.position().z + 0.5));
 
             for (ArmorStand as : toCheck) {
--- a/common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
+++ b/common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
@@ -80,6 +80,7 @@ public class ShamanTotemModel extends Model {
                     ShamanTotem newTotem = new ShamanTotem(
                             totemNumber,
                             -1,
+                            totemAS.getId(),
                             -1,
                             ShamanTotem.TotemState.SUMMONED,
                             new Location(totemAS.position().x, totemAS.position().y, totemAS.position().z));
@@ -119,7 +120,7 @@ public class ShamanTotemModel extends Model {
         Logic flow for the following bits:
         - First, the given entity is checked to see if it is a totem timer
         - If the given timerId (int entityId) is not already a totem, assign it to the lowest # totem slot
-          - Additionally, assign location, state, and time
+          - Additionally, assign location, state, and time, and keep the old visibleEntityId from when the spell was cast
         - If the given timerId is already a totem, update the time and location instead
           - Location is updated because there are now totems that can move
          */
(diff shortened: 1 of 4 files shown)


magicus:
> * This still has a very small chance to occur, if both yours and the other player's totems land in the same spot on the exact same tick, within 1 block of each other.

Well, if this happens, does it really matter? 🤔 The timers will show the same value then, right?

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -104,6 +106,61 @@ public void onTotemSpawn(AddEntityEvent e) {
...
+                                // server gets laggier
+                                possibleTimer.position().y + TOTEM_SEARCH_RADIUS * 5,
+                                possibleTimer.position().z + TOTEM_SEARCH_RADIUS));
+
+        for (ArmorStand as : toCheck) {
+            // Recreate location for each ArmorStand checked for most accurate coordinates
+            Location parsedLocation = new Location(as.position().x, as.position().y, as.position().z);
+            if (pendingTotem1VisibleId != null && as.getId() == pendingTotem1VisibleId) {
+                totem1 = new ShamanTotem(
+                        1, entityId, totem1.getVisibleEntityId(), -1, ShamanTotem.TotemState.ACTIVE, parsedLocation);
```
magicus: `totem1 = new ShamanTotem(..., totem1.getVisibleEntityId())` looks a bit self-referential. I assume this means that you already have a totem1, but you want to update it with the entityId and parsedLocation of the timer entity?

I would suggest one of these two alternatives instead: (1 is simpler but I recommend 2)

1) Keep the overall idea, but extract `totem1.getVisibleEntityId()` into a variable first, like `originalTotemVisibleEntityId = totem1.getVisibleEntityId()` so you don't reference totem1 when creating the replacement.

2) Make ShamanTotem mutable on the entityId and parsedLocation field. I assume the totemState field is already mutable? If the idea is that exactly one ShamanTotem object corresponds to exactly one totem in the game, I think this is a better solution. You are just building it up piece by piece as you gather enough data, and until it has become ACTIVE, we can't be sure to have all data about it.

magicus:
I have also another request, which really is independent of this PR, but I discovered it when I wondered if you did not already listen to `AddEntityEvent`, and was thinking that maybe `onTimerSpawn` should share the same entry point instead of having two distinct in the same class. (That's a design question I'm still on the fence for).

Anyway, the old code:
```
    public void onTotemSpawn(AddEntityEvent e) {
        Managers.TickScheduler.scheduleLater(
                () -> {
                    if (Math.abs(totemCastTimestamp - System.currentTimeMillis()) > 450) return;
                    Entity entity = getBufferedEntity(e.getId());
                    if (!(entity instanceof ArmorStand totemAS)) return;
```

Creating a `scheduleLater` is somewhat costly. We don't want to do that for every new entity in the game. At the very least, you can move the `instanceof ArmorStand` check to outside the lambda. I also think you should move the timestamp check, but since there will be a difference in time if this executes directly or 3 ticks later, you could either adjust the constant to compensate for this, or if the 450 is an extremely special value, check it first with something like 450 + 3*ticks + margin, and then once again with exactly 450 in the lambda.

(And, also, 450 should be a named constant)

ryanzhoudev:
> > * This still has a very small chance to occur, if both yours and the other player's totems land in the same spot on the exact same tick, within 1 block of each other.
> 
> Well, if this happens, does it really matter? 🤔 The timers will show the same value then, right?

Yes, but if it captures the wrong totem and either totem starts moving, the coordinates captured and displayed will be wrong

ryanzhoudev:
I have tried to move as many of the checks out of the delay as possible. However, the inventory and health check must stay in the delay, as the server sets these values 1-2 ticks after the totem actually spawns.

I also deleted the eye height check and the riding offset check because it's the same for all armour stands.

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -104,6 +109,64 @@ public void onTotemSpawn(AddEntityEvent e) {
...
+                WynntilsMod.postEvent(new TotemEvent.Activated(2, parsedLocation));
+                pendingTotem2VisibleId = null;
+            } else if (pendingTotem3VisibleId != null && as.getId() == pendingTotem3VisibleId) {
+                totem3.setTimerEntityId(entityId);
+                totem3.setLocation(parsedLocation);
+                totem3.setState(ShamanTotem.TotemState.ACTIVE);
+                WynntilsMod.postEvent(new TotemEvent.Activated(3, parsedLocation));
+                pendingTotem3VisibleId = null;
+            } else {
+                // No totem slots available?
```
magicus, line 164: I think you can leave the warning on. It should not happen, but if it does, it's good to have it in the logs.

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
```diff
@@ -104,6 +109,64 @@ public void onTotemSpawn(AddEntityEvent e) {
...
+                // WynntilsMod.getLogger().warn("Received a new totem {}, but no totem slots are available", entityId);
```
magicus, line 165:
```suggestion
                WynntilsMod.getLogger().warn("Received a new totem {}, but no totem slots are available", entityId);
```

magicus (approved):
Looks good.

common/src/main/java/com/wynntils/wynn/model/ShamanTotemModel.java
ryanzhoudev: I want to keep this as a comment, as the warn is useful for debugging purposes, but at the moment, the server sends a duplicate packet for each totem. So even if we successfully capture the totem, we will get an incorrect warn for every totem casted. If the duplicate packet disappears in the future, we can uncomment again.

magicus: ok

PR #1051, opened 2023-01-19
refactor: integrate SpellModel with QuickCastFeature

ryanzhoudev:
SpellProgressEvent and SpellCastEvent inherit from abstract SpellEvent, which allows those events to contain a SpellDirection[] array, as well as where the spell came from (for future use).

SpellProgressEvent is fired upon every user input to cast a spell.
SpellCastEvent is (incorrectly) fired upon every completed spell sequence. This should be fixed to only fire upon successful casts; but we don't really have a way of detecting this yet.

QuickCastFeature now relies on the singular SpellProgressEvent for it's purposes.

SpellDirection has been moved to it's own enum, and given an invert method for comparison later in SpellType.

### DIFF
--- a/common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
@@ -4,6 +4,8 @@
  */
 package com.wynntils.features.user;
 
+import static com.wynntils.wynn.objects.SpellDirection.NO_SPELL;
+
 import com.wynntils.core.components.Managers;
 import com.wynntils.core.features.UserFeature;
 import com.wynntils.core.features.properties.FeatureInfo;
@@ -52,7 +54,6 @@ public class QuickCastFeature extends UserFeature {
             StringUtils.compileCCRegex("§([LR]|Right|Left)§-§([LR?]|Right|Left)§-§([LR?]|Right|Left)§");
     private static final Pattern INCORRECT_CLASS_PATTERN = StringUtils.compileCCRegex("§✖§ Class Req: (.+)");
     private static final Pattern LVL_MIN_NOT_REACHED_PATTERN = StringUtils.compileCCRegex("§✖§ (.+) Min: ([0-9]+)");
-    private static final SpellDirection[] NO_SPELL = new SpellDirection[0];
 
     private SpellDirection[] spellInProgress = NO_SPELL;
 
--- a/common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
+++ b/common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
@@ -146,7 +146,7 @@ public class QuickCastFeature extends UserFeature {
         if (!WynnUtils.onWorld()) return;
 
         // Clear spell after the 40 tick timeout period
-        if (spellCountdown > 0 && --spellCountdown <= 0) spellInProgress = NO_SPELL;
+        if (spellCountdown > 0 && --spellCountdown <= 0) spellInProgress = SpellDirection.NO_SPELL;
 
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/wynntils/features/user/QuickCastFeature.java
```diff
@@ -4,18 +4,20 @@
  */
 package com.wynntils.features.user;
 
+import static com.wynntils.wynn.objects.SpellDirection.NO_SPELL;
```
magicus: Please do not do static imports, instead quality `NO_SPELL` to `SpellDirection.NO_SPELL`.

common/src/main/java/com/wynntils/wynn/event/SpellCastEvent.java
```diff
@@ -4,14 +4,20 @@
...
 package com.wynntils.wynn.event;
 
+import com.wynntils.wynn.objects.SpellDirection;
 import com.wynntils.wynn.objects.SpellType;
-import net.minecraftforge.eventbus.api.Event;
 
-public class SpellCastEvent extends Event {
+/**
+ * Fired upon successful, completed spell cast.
+ * FIXME: This event fires upon every three-click sequence, no matter if the spell was actually casted or not.
```
magicus: Maybe rename it TrySpellCastEvent instead then, to signal that we are not sure we managed to cast a spell. Or perhaps, arguably, we cast a spell, but we are not sure if it is *successful* or not.

common/src/main/java/com/wynntils/wynn/model/SpellModel.java
```diff
@@ -5,49 +5,75 @@
...
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class SpellModel extends Model {
-    private static final Pattern LEVEL_1_SPELL_PATTERN =
-            Pattern.compile("§a(Left|Right|\\?)§7-§a(Left|Right|\\?)§7-§r§a(Left|Right|\\?)§r");
-    private static final Pattern LOW_LEVEL_SPELL_PATTERN = Pattern.compile("§a([LR?])§7-§a([LR?])§7-§r§a([LR?])§r");
+    private static final Pattern SPELL_TITLE_PATTERN =
+            StringUtils.compileCCRegex("§([LR]|Right|Left)§-§([LR?]|Right|Left)§-§([LR?]|Right|Left)§");
```
magicus: I'd actually prefer not to use compileCCRegex. In fact, I'd like to remove that function. IntelliJ does not understand that it is a regexp pattern, in contrast to Pattern.compile, and will not assist you. I think it is better to have exact regexps, that can be copied into e.g. regex101.com.

I also have a memory that I spent quite some effort trying to get down a correct regex for spell patterns. Hm, maybe that was for the action bar middle segment.

common/src/main/java/com/wynntils/wynn/model/SpellModel.java
```diff
@@ -21,8 +20,11 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class SpellModel extends Model {
-    private static final Pattern SPELL_TITLE_PATTERN =
-            StringUtils.compileCCRegex("§([LR]|Right|Left)§-§([LR?]|Right|Left)§-§([LR?]|Right|Left)§");
+    // If you modify please test with
```
magicus:
```suggestion
    // If you modify please test on link below
```

ryanzhoudev: image

I'm not sure why your copy is missing the latest commit, this was fixed in a52aab7

magicus (approved):

PR #1054, opened 2023-01-20
feat: Several main map UX improvements

magicus:
I have been zooming in and out to get better cutoffs, and fading, behavior for different POIs. The goal has been to have just enough information at every zoom level; not too much but not too little either. Fading should be smooth but not too extracted so everything looks transparent for too long.

I have also added the place level overlays on hover from legacy, fixed the fading scale to be multiplicative instead of additive, and some more minor fixes.


magicus:
Screen recording demo: https://www.dropbox.com/s/tuco8ov7pwy6k58/map-ux-improvements.mp4

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -51,28 +51,34 @@ public class MapFeature extends UserFeature {
     private final Type customPoisType = new TypeToken<List<CustomPoi>>() {}.getType();
 
     @Config
-    public float poiFadeDistance = 0.6f;
+    public float poiFadeAdjustment = 0.4f;
```
kristofbolyai, line 54: People's own configs will be weird.. I guess we need to make them reset it?

common/src/main/java/com/wynntils/wynn/model/map/poi/CombatPoi.java
```diff
@@ -24,6 +24,9 @@ public Texture getIcon() {
 
     @Override
     public float getMinZoomForRender() {
+        if (kind == CombatKind.CAVES) {
```
kristofbolyai, line 27: Eventually all types should be configurable

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
magicus: I don't know really how to handle this. I assume they will get a json key that will no longer be used, and then get the new default value for poiFadeAdjustment?

Did we properly implement the functionality so values are only stored in the config if the user override the default?

common/src/main/java/com/wynntils/wynn/model/map/poi/CombatPoi.java
magicus: I agree. Doing this in a way that makes it useful will be a bit of a challenge. I really disliked the system in Legacy. I want some kind of hierarchical system, were "subtypes" inherts values of supertypes, unless overridden. So you can like set a value for all "combat POI" and then make an exception for caves, etc. Or make most services appear only when really zoomed in, except for identifiers which show up "earlier", etc.

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
kristofbolyai: This was a bad place to review. This config is renamed, so it is fine. But, there are some, with changed defaults, and the same names, that are interpreted in a different way now.

PR #1055, opened 2023-01-20
feat: Party management ui

ryanzhoudev:
Adds the party management ui from legacy.


ryanzhoudev:
oh hey the skip spotless/build on drafts actually worked

common/src/main/java/com/wynntils/gui/screens/PartyManagementScreen.java
```diff
@@ -0,0 +1,479 @@
...
+                    Component.translatable("screens.wynntils.partyManagementGui.kick"),
+                    (button) -> kickFromParty(playerName))
+                    .pos(this.width / 2 - xStart + 296, this.height / 2 - 125 + i * 20 - 10)
+                    .size(50, 20)
+                    .build());
+        }
+        this.renderables.addAll(promoteButtons);
+        this.renderables.addAll(kickButtons);
+        this.children.addAll(promoteButtons);
+        this.children.addAll(kickButtons);
```
kristofbolyai: This is really bad, we can't do such things in render time.. Can't you do this in init, or have some kind of update event that updates periodically?

kristofbolyai: Also, use this.addRenderableWidget(). For removing, well, you shouldn't really remove widgets without re-initing...

kristofbolyai: Okaay, I see now. Make these a widget. And make that widget render the promote, kick, player head, etc. It results in nicer code.

kristofbolyai: (I know this is WIP, but doing things the proper way by default is less work in the future.)

common/src/main/java/com/wynntils/core/components/Models.java
```diff
@@ -56,13 +58,15 @@ public final class Models {
...
     public static final CharacterModel Character = new CharacterModel(CombatXp);
+    public static final FriendsModel Friends = new FriendsModel(WorldState);
     public static final GearItemModel GearItem = new GearItemModel(GearProfiles);
+    public static final GuildModel Guild = new GuildModel(WorldState);
     public static final HadesModel Hades = new HadesModel(Character, WorldState);
     public static final ItemModel Item = new ItemModel(GearItem, GearProfiles, IngredientProfiles);
     public static final LootChestModel LootChest = new LootChestModel(Container);
     public static final LootrunModel Lootrun = new LootrunModel(Container);
     public static final MapModel Map = new MapModel(GuildAttackTimer);
-    public static final PlayerRelationsModel PlayerRelations = new PlayerRelationsModel(WorldState);
```
kristofbolyai, line 65: I am not sure we need this? @magicus

kristofbolyai: Oh. We separated all of them. Well. Do we need that?

common/src/main/java/com/wynntils/screens/partymanagement/widgets/PartyPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+import com.wynntils.utils.render.type.HorizontalAlignment;
+import com.wynntils.utils.render.type.TextShadow;
+import com.wynntils.utils.render.type.VerticalAlignment;
+import net.minecraft.client.gui.components.AbstractWidget;
+import net.minecraft.client.gui.components.Button;
+import net.minecraft.client.gui.narration.NarrationElementOutput;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+
+public class PartyPlayer extends AbstractWidget {
```
kristofbolyai: Rename to `PartyPlayerWidget`?

common/src/main/java/com/wynntils/screens/partymanagement/widgets/PartyPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+        // hat rendering
+        RenderUtils.drawTexturedRect(poseStack, skin, this.getX(), this.getY(), 1, 16, 16, 40, 8, 8, 8, 64, 64);
+
+        // name rendering
+        CustomColor color = CommonColors.WHITE;
+        if (playerName.equals(Models.Party.getPartyLeader())) {
+            color = CommonColors.YELLOW;
+        } else if (Models.Friends.getFriends().contains(playerName)) {
+            color = CommonColors.GREEN;
+        }
```
kristofbolyai: No color for guild? Also, shouldn't friendship take priority in coloring? (Sorry I am not sure where this is used exactly, just a "blind" suggestion)

common/src/main/java/com/wynntils/screens/partymanagement/widgets/PartyPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+        McUtils.sendCommand("party kick " + playerName);
+    }
+
+    private void promoteToLeader(String playerName) {
+        McUtils.sendCommand("party promote " + playerName);
+    }
+
+    private void disbandParty() {
+        McUtils.sendCommand("party disband");
+    }
```
kristofbolyai: The widget is not a place to put this. Move it into the models.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+        offlineMembers.removeAll(McUtils.mc().level.getScoreboard().getTeamNames());
+    }
+
+    private void createParty() {
+        McUtils.sendCommand("party create");
+    }
+
+    private void leaveParty() {
+        McUtils.sendCommand("party leave");
+    }
```
kristofbolyai: Same as earlier, this does not belong here.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        super.mouseClicked(mouseX, mouseY, button);
+
+        // For some reason the (widget.isMouseOver) check does not work for partyMembersWidgets
+        partyMembersWidgets.forEach(widget -> widget.mouseClicked(mouseX, mouseY, button));
```
kristofbolyai: We have to debug this? Doesn't this mean that we click all widgets etc? Also, shouldn't we return early after we have 1 true mouseClicked? (true means we successfully clicked on something)

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+        suggestedPlayersWidgets.forEach(widget -> {
+            if (widget.isMouseOver(mouseX, mouseY)) widget.mouseClicked(mouseX, mouseY, button);
+        });
+
+        return true;
```
kristofbolyai: See comments above, technically this means we have failed to click anything, so return false. Also, why do we not call super here?

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private void kickOffline() {
+        refreshParty();
+        offlineMembers.forEach(playerName -> McUtils.sendCommand("party kick " + playerName));
+    }
+
+    private void refreshOffline() {
+        offlineMembers.clear();
+        offlineMembers.addAll(Models.Party.getPartyMembers());
+        offlineMembers.removeAll(McUtils.mc().level.getScoreboard().getTeamNames());
```
kristofbolyai: This line needs a bit of explaining.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+    private void refreshParty() {
+        Models.Party.requestPartyListUpdate();
+        reloadSuggestedPlayersWidgets();
+        reloadMembersWidgets();
+        refreshOffline();
+    }
+
+    private void kickOffline() {
+        refreshParty();
+        offlineMembers.forEach(playerName -> McUtils.sendCommand("party kick " + playerName));
```
kristofbolyai: Move to model

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+        fieldText = COMMA_REPLACER.matcher(fieldText).replaceAll(","); // semicolons and spaces to comma
+        if (fieldText.isBlank()) return;
+
+        if (!Models.Party.isPartying()) {
+            createParty();
+        }
+
+        Set<String> toInvite = new HashSet<>(List.of(fieldText.split(",")));
+        toInvite.removeAll(Models.Party.getPartyMembers());
+        toInvite.forEach(playerName -> McUtils.sendCommand("party invite " + playerName));
```
kristofbolyai: Move to model (keep the foreach etc, but command knowledge is not a screen's duty)

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+        for (int i = 0; i < suggestedPlayers.size(); i++) {
+            String playerName = suggestedPlayers.get(i);
+            if (playerName == null) continue;
+
+            suggestedPlayersWidgets.add(new SuggestionPlayer(
+                    this.width / 2 + 204, this.height / 2 - 125 + i * 20 - 10, 200, 20, playerName));
+        }
+    }
+
+    @SubscribeEvent
```
kristofbolyai: Generally we should not use `SubscribeEvent` in screens. (I know I did before, I will remove them sometime).

What I suggest you do: Have the specific models listen to these events, and if the current screen is this, then call methods of this to update it.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+                I18n.get("screens.wynntils.partyManagementGui.leader"),
+                this.width / 2 - 300,
+                this.height / 2 - 120,
+                CommonColors.YELLOW,
+                HorizontalAlignment.Left,
+                VerticalAlignment.Middle,
+                TextShadow.NORMAL);
+        fr.renderText(
+                poseStack,
+                "§m" + I18n.get("screens.wynntils.partyManagementGui.offline"),
```
kristofbolyai: Can you use ChatFormatting here? It is much more readable.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+                I18n.get("screens.wynntils.partyManagementGui.legend"),
+                this.width / 2 - 300,
+                this.height / 2 - 144,
+                CommonColors.WHITE,
+                HorizontalAlignment.Left,
+                VerticalAlignment.Middle,
+                TextShadow.NORMAL);
+        fr.renderText(
+                poseStack,
+                "§l" + I18n.get("screens.wynntils.partyManagementGui.self"),
```
kristofbolyai: ChatFormatting here too

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -0,0 +1,426 @@
...
+
+        boolean partying = Models.Party.isPartying();
+
+        createPartyButton.active = !partying;
+        leavePartyButton.active = partying;
+        kickOfflineButton.active = partying && !offlineMembers.isEmpty();
+        inviteButton.active = !inviteInput
+                .getTextBoxInput()
+                .isBlank(); // partying check not required as button automatically makes new party if not in one
+        refreshOffline();
```
kristofbolyai: We are still doing this render time. I will need to clone this whole PR to check how to work around this.

common/src/main/java/com/wynntils/screens/partymanagement/widgets/PartyPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+            color = CommonColors.YELLOW;
+        } else if (Models.Friends.getFriends().contains(playerName)) {
+            color = CommonColors.GREEN;
+        }
+        String prefix = "";
+        if (isOffline) {
+            prefix = "§m";
+        } else if (playerName.equals(McUtils.player().getName().getString())) {
+            prefix = "§l";
+        }
```
kristofbolyai: Use ChatFormatting here too

common/src/main/java/com/wynntils/screens/partymanagement/widgets/SuggestionPlayer.java
```diff
@@ -0,0 +1,78 @@
...
+        if (Models.Party.getPartyMembers().contains(playerName)) return;
+        inviteButton.render(poseStack, mouseX, mouseY, partialTick);
+    }
+
+    private void inviteToParty(String playerName) {
+        if (!Models.Party.isPartying()) {
+            McUtils.sendCommand("party create");
+        }
+        McUtils.sendCommand("party invite " + playerName);
+    }
```
kristofbolyai: Move to Model

common/src/main/java/com/wynntils/screens/partymanagement/widgets/PartyPlayer.java
ryanzhoudev: Guild is not actually suggested in legacy, nor is it highlighted in the member list. I am not sure how often people party with their guild as I am not really in a large guild, but we can look into this if there is demand.

The leader/friend coloring is used in the member list, where leader is yellow and friends are green. I think it is more important to always have a leader visible, so friends would be lower priority.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
ryanzhoudev: All the members in the party are added to offlineMembers, then the ones that have a scoreboard entry are removed. Scoreboard entry only exists for those that are online in the server.

I will also add this as a comment.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
ryanzhoudev: So the issue is the widget technically just ends halfway across the screen and doesn't actually encompass the button that is displayed

Or I guess the issue was

kristofbolyai: The containing widget was smaller than its actual elements?

ryanzhoudev: Apparently so, I missed that the width was hardcoded to be 200 instead of the correct TOTAL_WIDTH field, which is 344

kristofbolyai: But it's fixed now?

ryanzhoudev: yes

common/src/main/java/com/wynntils/features/user/PartyManagementScreenFeature.java
```diff
@@ -0,0 +1,21 @@
...
+import com.wynntils.core.keybinds.KeyBind;
+import com.wynntils.screens.partymanagement.PartyManagementScreen;
+import com.wynntils.utils.mc.McUtils;
+import org.lwjgl.glfw.GLFW;
+
+@FeatureInfo(stability = FeatureInfo.Stability.STABLE)
+public class PartyManagementScreenFeature extends UserFeature {
+    @RegisterKeyBind
+    private final KeyBind openPartyManagementScreen =
+            new KeyBind("Open Party Management Screen", GLFW.GLFW_KEY_RIGHT_BRACKET, true, () -> McUtils.mc()
```
magicus: That's a particularly annoying key to use for us with non-US keyboards, since it claims to be `[` but that is pressed using a combination of keys, and in fact I think it means I need to press the `Å` key. Maybe use `P` instead, I don't think it is used by Vanilla or any other of our functions?

common/src/main/java/com/wynntils/core/components/Models.java
magicus: Splitting it up seems fine to me. Donkey has added a lot of party functionality. With regards to models, I think the trick is to find the "right" size. Functionality that seems to belong together and forms a large-enough unit. If adding all these party stuff to a "PlayerRelations" model made it too unwieldy, it is okay to break it apart into friends and party. 

I did a brief scroll through, and it looked overall okay, but I'd like to give it a closer look as well. Sorry to say it will have to wait for tomorrow. I'm gonna post the PRs for the Gear Rewrite now, and then fall asleep...

common/src/main/java/com/wynntils/features/user/HadesFeature.java
```diff
@@ -37,14 +37,14 @@ protected void onConfigUpdate(ConfigHolder configHolder) {
             }
             case "shareWithParty" -> {
                 if (shareWithParty) {
-                    Models.PlayerRelations.requestPartyListUpdate();
+                    Models.Party.requestPartyData();
```
magicus: Since this and requestFriendData are now in separate models, they can be named just `requestData`.

common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -0,0 +1,195 @@
...
+            if (FRIEND_LIST_FAIL_1.matcher(coded).matches()) {
+                event.setCanceled(true);
+                return;
+            }
+        }
+    }
+
+    private boolean tryParseNoFriendList(String coded) {
+        if (FRIEND_LIST_FAIL_2.matcher(coded).matches()) {
+            WynntilsMod.info("Player has no friends!");
```
magicus: Ohhh... this sounds soo sad! 😿 

But seriously, maybe rephrase it to something like "Friend list is empty", so we don't trigger a lonely suicidal player to read the logs and think their life sucks.

common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -0,0 +1,195 @@
...
+        McUtils.sendCommand("friend list");
+        WynntilsMod.info("Requested friend list from Wynncraft.");
+    }
+
+    public Set<String> getFriends() {
+        return friends;
+    }
+
+    @SubscribeEvent
+    public void onFriendJoin(FriendConnectionEvent.Join e) {
```
magicus: This does not seem right, either. Now the model has hard-coded knowledge about the screen. Let me think a bit and see if I can suggest a better solution.

common/src/main/java/com/wynntils/features/user/PartyManagementScreenFeature.java
```diff
@@ -0,0 +1,21 @@
...
+import com.wynntils.screens.partymanagement.PartyManagementScreen;
+import com.wynntils.utils.mc.McUtils;
+import org.lwjgl.glfw.GLFW;
+
+@FeatureInfo(stability = FeatureInfo.Stability.STABLE)
+public class PartyManagementScreenFeature extends UserFeature {
+    @RegisterKeyBind
+    private final KeyBind openPartyManagementScreen =
+            new KeyBind("Open Party Management Screen", GLFW.GLFW_KEY_P, true, () -> McUtils.mc()
+                    .setScreen(PartyManagementScreen.create()));
```
magicus: Yes, this is the solution. Here, extract `PartyManagementScreen.create()` into a field (and `null` it when closing the screen), and then add your event subscriptions here, and call the appropriate methods on the saved screen.

common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -0,0 +1,404 @@
...
+
+    private static final Pattern PARTY_CREATE_SELF = Pattern.compile("§eYou have successfully created a party\\.");
+
+    private static final Pattern PARTY_KICK_OTHER = Pattern.compile("§eYou have kicked the player from the party.\\.");
+    // endregion
+
+    private boolean expectingPartyMessage = false; // Whether the client is expecting a response from "/party list"
+    private boolean nextKickHandled = false; // Whether the next "/party kick" sent by the client is being handled
+
+    private boolean partying; // Whether the player is in a party
```
magicus: 🍰 Party like there's no tomorrow! :-) I don't know if it is my non-native English skills that play me a trick here, but "partying" seems to me to fully indicate the "festivities" meaning of "party", not the "association" meaning.

Maybe `inParty` or `partyActive` or, I dunno...

common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -0,0 +1,404 @@
...
+        return partyMembers;
+    }
+
+    public Set<String> getOfflineMembers() {
+        return offlineMembers;
+    }
+
+    @SubscribeEvent
+    public void onPartyUpdate(RelationsUpdateEvent.PartyList e) {
+        PartyManagementScreen.reloadWidgetsIfScreenOpen();
```
magicus: Same here, this should move to the Feature.

common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    @SubscribeEvent
+    public void onSetTeam(SetPlayerTeamEvent e) {
+        if (!partying) return;
```
magicus: ... and this too.

magicus:
Overall I am very pleased to see the new models and their design. It looks great. I added a few comments, but most of them are really minor. 

There is one more thing that is nagging me though, and that is the hierarchy and naming of the events. They does not really feel spot-on, but a little bit contrived. Like when a friend joins. I would have expected  that to be like a `FriendEvent.Joined`, so the verb tense feels off, etc. I could sit down and think about it and give you a more precise description of what I want to see, but I think it is just better if I ask you to see if you can make one more pass over the events, and see if you can find anything yourself. And after that, if you can't see any more changes to be made, then I guess it's fine.

The rendering stuff I'll leave entirely to bolyai; from what I can tell it looks good. One hangup I had was maybe that I did not understand what `fr` was until I checked it up, but I understand why you'd want to shorten it and I'm okay with it.

ryanzhoudev:
The refactoring may have gotten a bit out of hand in this PR. But I think this makes a lot more sense than one single PlayerRelations model and event combination.

Quick summary - The general idea is now PlayerRelations is FriendsModel and PartyModel, and in the future, GuildModel when that is needed. The events FriendsEvent and PartyEvent correspond to these models, and the Hades events (since the format is so specific) have been renamed to HadesRelationsUpdateEvent. FriendConnectionEvent had been rolled into FriendsEvent as well.

magicus (approved):
Looks super-duper now!

PR #1075, opened 2023-01-24
feat: Add custom totem highlight color support

ryanzhoudev:
Using ChatFormatting as a config enum resulted in some unexpected problems, like the option to select BOLD as a color. It also didn't display properly in the config book because the book was just reading something like §c and displaying nothing, as it parsed the code.


ryanzhoudev:
oh hey the marking as ready to review actually did trigger a build

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -46,13 +46,13 @@ public class ShamanTotemTrackingFeature extends UserFeature {
     public boolean highlightShamanTotems = true;
 
     @Config
-    public static ChatFormatting firstTotemColor = ChatFormatting.WHITE;
+    public static TotemTrackingColors firstTotemColor = TotemTrackingColors.WHITE;
```
kristofbolyai: Why not CustomColor and pass it to font renderer?

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
ryanzhoudev: We can't set a team color to CustomColor, it has to be a ChatFormatting code

kristofbolyai: Oh, I wanna try injecting into the renderer code then to make it able to render any color. I will push that to this PR when I get it to work, alright? Or do you wanna try to so that?

ryanzhoudev: Feel free to push to here, I dunno if I can get that working anytime soon as the only mixins I've touched are for sending events

magicus:
I'm not sure I'm getting the premise here. The colors are chosen by us, right? Not by Wynncraft?

So is the core issue here just that you want people to be able to select different colors than the ones you picked?

I think it is quite easy to add all color support, and then we don't have to add a new color enum

kristofbolyai:
@DonkeyBlaster added support, but I have not had time to test it, so I will have to do that later.

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -46,16 +46,15 @@ public class ShamanTotemTrackingFeature extends UserFeature {
...
 
     @Config
-    public static ChatFormatting secondTotemColor = ChatFormatting.AQUA;
+    public static CustomColor secondTotemColor = CommonColors.BLUE;
 
     @Config
-    public static ChatFormatting thirdTotemColor = ChatFormatting.RED;
+    public static CustomColor thirdTotemColor = CommonColors.RED;
 
     private static final String TOTEM_HIGHLIGHT_TEAM_BASE = "wynntilsTH";
```
magicus: I guess this is not needed anymore?

common/src/main/java/com/wynntils/features/user/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -184,11 +162,21 @@ public void renderPreview(
...
+                                            "totemNumber should be 1, 2, or 3! (switch in #render in ShamanTotemTrackingFeature");
+                                };
+                        ;
+
                         String prefix =
                                 switch (shamanTotem.getTotemNumber()) {
-                                    case 1 -> firstTotemColor + "Totem 1";
-                                    case 2 -> secondTotemColor + "Totem 2";
-                                    case 3 -> thirdTotemColor + "Totem 3";
+                                    case 1 -> "Totem 1";
```
magicus: Maybe just `"Totem " + totemNumber`? :-D

common/src/main/java/com/wynntils/mc/mixin/LevelRendererMixin.java
```diff
@@ -55,6 +61,20 @@ private void renderLevelPre(
                 this.minecraft.levelRenderer, poseStack, partialTick, projectionMatrix, finishNanoTime, camera);
     }
 
+    @ModifyArgs(
+            method = "renderLevel",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/client/renderer/OutlineBufferSource;setColor(IIII)V"))
+    private void modifyOutlineColor(Args args, @Local Entity entity) {
+        if (entity instanceof WynntilsCustomGlowEntityProperty property) {
```
magicus: That's always going to be the case, is it not?

common/src/main/java/com/wynntils/models/entities/WynntilsCustomGlowEntityProperty.java
```diff
@@ -0,0 +1,17 @@
+/*
```
magicus, line 1: I don't think this belongs in models.

In fact, do you really need this class at all? Can't we just be happy by extending the Entity class with two new methods? Hm, no, of course not, you'll need a way to access these. 

How has we solved this before? `AnnotatedItemStack` is in `com.wynntils.handlers.item` but I'm honestly not sure that is the right place.

Maybe we need a new package in `mc`. Like `mc.extensions`? 

Or, you can put this file here for the time being, if you are okay with me moving it later on.

magicus:
@kristofbolyai I really like your idea here, but I had a handful of comments on the actual design.

common/src/main/java/com/wynntils/models/entities/WynntilsCustomGlowEntityProperty.java
kristofbolyai: Exactly. I knew this is in the wrong place, I tried to look at ScreenMixin, which does the same thing, and it is in the model (as sometimes it is used in custom screen impls, so you probably got that idea). So yes, I would like you to move this :). But, don't forget to change the mixin file to the correct class name (I have no reason why this even worked before...).

magicus: Sorry, which mixin file has the wrong name?

magicus: But yes, leave it here and I'll clean up later.

common/src/main/resources/wynntils.mixins.json
```diff
@@ -45,8 +46,11 @@
   ],
   "compatibilityLevel": "JAVA_17",
   "injected_interfaces": {
+    "net/minecraft/class_1297": [
+      "com/wynntils/models/entities/WynntilsCustomGlowEntityProperty"
+    ],
     "net/minecraft/class_437": [
-      "com/wynntils/gui/widgets/TextInputBoxWidget"
+      "com/wynntils/screens/base/TextboxScreen"
```
kristofbolyai, line 52: @magicus

kristofbolyai:
And now it works :)

kristofbolyai:
@DonkeyBlaster Sorry for hijacking this PR, but I hope it turned out good overall. Free feel to ask anything about the mixin magic I did in Discord, I am happy to help.

common/src/main/resources/wynntils.mixins.json
magicus: Yeah, look at that! :) This is not needed. That is only if you really want the actual target class to have a new interface. We are only using this as a way to get around the java compiler otherwise saying "no there is no such method in Screen/ItemStack/Entity". I'm not adding it for ItemAnnotations and it works just as fine.

magicus: But I'll clean that up as well.

magicus (approved):

PR #1077, opened 2023-01-24
feat: ItemHandler can now detect item renames

magicus:
This will allow for nice future additions in models, and showcases this use in a rudimentary framework for properly detecting spells finished casting in SpellModel.

Also fixes an old-standing bug with spell cast "popups" not showing. 

Also fixes a bug where ItemHandler re-analyzed items even when they were not changed. (I did not realize the difference between  McUtils.player().getInventory().items and McUtils.player().inventoryMenu.getItems()...)

Also removes 1 of 3 remaining "state managed" features.


magicus:
And, oh, also fixes the bug (I did not even notice until working with this!) that causes a weapon to lose it's highlight background in the hotbar after doing a spell. (Since for a brief moment, we could not match that item with a known gear since it was named "... spell cast!")

kristofbolyai (approved):

PR #1080, opened 2023-01-24
feat: Add a bunch of new functions

kristofbolyai:


magicus (approved):
Nice! Any idea on how we can make these functions more discoverable?

I'm thinking we ought to have a separate GUI for this; the config feels a bit ... I dunno. Limited? Some way do see all info box contents at the same time, with support for choosing between functions.

PR #1085, opened 2023-01-25
feat: Prohibit rename on spell cast

magicus:
Ideally, this should be an Overlay, but I'm not sure how to make an overlay that preserves the Vanilla layout as default. Hopefully someone can improve this later on.


magicus:
I can do better.

PR #1086, opened 2023-01-25
 feat: Add Spell Cast Render feature

magicus:
This will prohibit item renames by implementing a separate render for spell casting.
It will also add a spell cast vignette.

Ideally, the spell popup should be an Overlay, but I do not know how to position it in the same place as Vanilla by default.

This is dependent on https://github.com/Wynntils/Artemis/pull/1083


common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
```diff
@@ -0,0 +1,113 @@
...
+    @SubscribeEvent
+    public void onRender(RenderEvent.Post event) {
+        if (spellTimer <= 0) return;
+
+        if (renderVignette && intensity > 0f) {
+            int shownTicks = SHOW_TICKS - spellTimer;
+            int fade = FADE_TICKS - shownTicks;
+            if (fade > 0) {
+                float alpha = intensity * ((float) fade / FADE_TICKS);
+                RenderUtils.renderVignetteOverlay(event.getPoseStack(), vignetteColor, alpha);
```
kristofbolyai, line 91: How does this look like with low health vignette and this?

common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
```diff
@@ -0,0 +1,113 @@
...
+
+        int width = McUtils.mc().gui.getFont().width(spellMessage);
+        int x = (screenWidth - width) / 2;
+        int y = screenHeight - 59;
+
+        PoseStack poseStack = event.getPoseStack();
+        RenderSystem.enableBlend();
+        RenderSystem.defaultBlendFunc();
+        McUtils.mc().gui.getFont().drawShadow(poseStack, spellMessage, x, y, 0xFFFFFF + (alpha << 24));
+        RenderSystem.disableBlend();
```
kristofbolyai: This is very much ew. I am not sure we can do anything better without making some overlay or smth. So it stays.

common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        // Render it the same way vanilla renders item changes
+        int alpha = (int) Math.min((float) spellTimer * 256.0F / 10.0F, 255.0F);
+        if (alpha <= 0) return;
+
+        Window window = McUtils.mc().getWindow();
+        var screenWidth = window.getGuiScaledWidth();
+        var screenHeight = window.getGuiScaledHeight();
```
kristofbolyai: I don't think we've ever used var anywhere else. Do we want to?

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
magicus: Yeah, I matched what vanilla is doing to get the same output. 

As I said, I think the right way to go is to make an overlay, but I'm not that proficient in writing Overlays, so I'm happily leaving that conversion over to someone else. _wink wink_ 😉

common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
magicus: No, no! Absolutely not. I am working in another project which has not only accepted var, but more or less made it the normal form. It is *horrible*. You have to mouse-over everything. Code reviewing is hopeless.

I am using `var` when developing and is unsure about (or too lazy to type) the real type of variables. Then I use IDEA assist to replace it with the correct type. I do check my code if I have missed any lingering `var`s, I often have but typically discover them. I am a bit suprised that this were the first I missed. :-)

I have looked at an IntelliJ/Qodana warning to flag the use of `var`, but have not found one. I guess there is one, I just can't figure out what it is called.

kristofbolyai: I share your opinion about var.

common/src/main/java/com/wynntils/features/user/SpellCastRenderFeature.java
magicus: I wondered that myself, but I'm honestly not quite sure how to test that. I need to have a character that has a reasonable amount of spells unlocked, go to some place where I can get a lot of damage, not die, and ...

*duh*

You know what. I'll just hack the low health vignette to start appearing at 95% or something and test. :-) I'll be back.

magicus: It looks great! It gets kind of purplish, but the impression you get is that the red gets a blue flash. I captured a screen recording but it is too large to upload. But here is a still, from where both blue and red vignette is at max.

Screen Shot 2023-01-25 at 19 25 33

PR #1087, opened 2023-01-25
fix: Fix Forge compatibility of LevelRendererMixin

kristofbolyai:


magicus:
This kind of changes make me a bit scared.

How did you find out the original code did not work? Testing on Forge? What happened, a crash or just broken functionality?

Secondly, and perhaps more important, how did you come up with this solution instead? Are you inspecting the Forge version of the minecraft classes? 

I'm just worried that we have no good story to make sure this kind of problems do not happen again in the future, and I wonder if there is anything we can do differently to avoid them.

kristofbolyai:
> This kind of changes make me a bit scared.
> 
> 
> 
> How did you find out the original code did not work? Testing on Forge? What happened, a crash or just broken functionality?
> 
> 
> 
> Secondly, and perhaps more important, how did you come up with this solution instead? Are you inspecting the Forge version of the minecraft classes? 
> 
> 
> 
> I'm just worried that we have no good story to make sure this kind of problems do not happen again in the future, and I wonder if there is anything we can do differently to avoid them.

It's simple. ModifyArgs don't work on forge. So I did this instead, same concept, same work done. There are no multiple mixins or anything. Just this, working everywhere.

magicus:
Ok, ModifyArgs do not work. Like in, no crash, just don't work?

Should we check for it in Spotless and warn? Like the IntelliJ constructs?

kristofbolyai:
> Ok, ModifyArgs do not work. Like in, no crash, just don't work?
> 
> 
> 
> Should we check for it in Spotless and warn? Like the IntelliJ constructs?
> 
> 
> 
> 

ModifyArgs do not work at all in Forge. It just doesn't start. This is a known and verified issue, has been for a long time, but I had to check Forge discord to know it...

I guess we can add a spotless check.

kristofbolyai:
@magicus Here is the code, red is old `ModifyArgs`, green is new `ModifyExpressionValue`. It is not so scary after all :).

image

magicus:
Sorry if I'm unclear. It is not so much the actual change that scares me (in fact, the new code looks more comprehensible, so it's better in both that it works and is more readable :-)).

My worry was about how we should be able to keep all our current and future mixins working in the different mod loader environments. The more differences there are, the more the chance of a failure.

Sometimes Forge has patched the code, making our mixins fails. But this was obviously not the case here. But yes, sounds like we should add ModifyArg to our verboten list.

PR #1090, opened 2023-01-25
fix: Fix party sharing

kristofbolyai:


magicus (approved):
Yay, thanks for fixing. This annoyed me when I was playing several times, but forgot about it afterwards.

PR #1091, opened 2023-01-25
feat: Make the durability functions work with gathering tools

kristofbolyai:


common/src/main/java/com/wynntils/models/items/properties/DurabilityItemProperty.java
```diff
@@ -0,0 +1,11 @@
+/*
```
magicus: This is what `DurableItemProperty` does. Perhaps not the best of names, though. But can you please rename that interface instead of creating a new?

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -126,18 +126,18 @@ public List<String> getAliases() {
...
 
             Optional<WynnItem> wynnItem = Models.Item.getWynnItem(itemStack);
 
-            if (wynnItem.isPresent() && wynnItem.get() instanceof CraftedGearItem craftedGearItem) {
-                return craftedGearItem.getDurability().getCurrent();
+            if (wynnItem.isPresent() && wynnItem.get() instanceof DurabilityItemProperty durableItem) {
+                return String.valueOf(durableItem.getDurability().getCurrent());
             }
 
-            return -1;
```
magicus, line 140: It feels a bit sad that we should need to sacrifice the type information just to be able to show "no value" in a better way.
What do you say about functions being able to return `Optional<Integer>`, and if the optional is empty, we display nothing?

kristofbolyai: I think this is a task when we finish the info variable refactors of artemis.

common/src/main/java/com/wynntils/models/items/properties/DurabilityItemProperty.java
kristofbolyai: I am blind

magicus (approved):

PR #1092, opened 2023-01-25
feat: Fade the beacon beam if we are close to the target

kristofbolyai:


magicus:
I agree that the beacon beam is annoying up close. But is shrinking the right response? I was thinking (when I got annoyed at it when playing, prior to seeing this PR) that we should change the alpha to make it just fade away as we get closer.

kristofbolyai:
I didn't see a way to change alpha, and in my eyes this was a solution too. I will check the code out once again, I guess.

Incompleteusern:
Legacy behavior is fading the beacon beam, if we can get that to work a config between the two would be fine

magicus:
Looking at the vanilla beacon rendering code, I think you need to change alpha on the color.

magicus:
(Legacy did a completely original beacon implementation)

magicus:
Heh, vanilla is hard-coding alpha to 1.0F :-(

magicus:
Can we do stuff with `RenderSystem.enableBlend()`, or am just misunderstanding how rendering works?

Otherwise we probably need to do some mixin injection in vanilla to be able to change alpha. :-&

kristofbolyai:
Well, I looked at the source code.. I think there is a way to do it, but I would rather not. Can you accept this, or should I draft this?

magicus:
I need to try it out in gameplay to see what it looks like. I'll check this out and have a look.

magicus:
Honestly, it looks a bit silly with the shrinking beam. :-/

A for effort, but I think you should draft it and see if you can get the fading to work instead.

magicus (approved):

magicus:
Thanks for fixing the fade!

PR #1096, opened 2023-01-27
refactor: Models need intra-model dependencies

magicus:
Just as with Managers, Models can depend on Models and thus initialization order must be correct. (This will be needed in upcoming item tooltip changes.)


magicus:
Also, we might want to look over our design as well. Some of these dependencies look strange. At least these comes to mind:

```
    public static final LootChestModel LootChest = new LootChestModel(Container);
    public static final LootrunModel Lootrun = new LootrunModel(Container);
    public static final MapModel Map = new MapModel(GuildAttackTimer);
    public static final PlayerRelationsModel PlayerRelations = new PlayerRelationsModel(WorldState);
    public static final QuestModel Quest = new QuestModel(Character);
    public static final ShamanMaskModel ShamanMask = new ShamanMaskModel(WorldState);
    public static final ShamanTotemModel ShamanTotem = new ShamanTotemModel(WorldState);
    public static final DiscoveryModel Discovery = new DiscoveryModel(Character, Compass, Quest, Territory);
    public static final EmeraldModel Emerald = new EmeraldModel(Item);
```

kristofbolyai (approved):
I thought we had these.. Or was it managers..?

magicus:
@kristofbolyai It was just for managers.
In the old world, the dependency was upheld by a feature A that needed model X, and if model X needed model Y, then feature A said it depended on both X and Y. (So it had to know intimate details about models). 

I have in generally tried to remove inter-model dependencies. I realize we can't get rid of them entirely, but we should try to minimize them. (As with Managers)

kristofbolyai:
I see now, thanks!

PR #1098, opened 2023-01-28
feat: Add Experience Functions (xpm_raw, xpm, xppm)

kristofbolyai:


magicus:
Can you move in the other related XP functionality from the CharacterManager, so XP functionality is not split between two managers? Like LEVEL_UP_XP_REQUIREMENTS and the XP getters.

common/src/main/java/com/wynntils/models/discoveries/DiscoveryInfo.java
```diff
@@ -70,7 +70,7 @@ private List<Component> generateLore() {
 
         displayLore.add(Component.literal(name).withStyle(type.getColor()).withStyle(ChatFormatting.BOLD));
 
-        if (Models.Character.getXpLevel() >= minLevel) {
+        if (Models.Experience.getXpLevel() >= minLevel) {
             displayLore.add(Component.literal("✔")
```
magicus, line 74: The DiscoveryModel needs a new dependency as well...

common/src/main/java/com/wynntils/models/experience/ExperienceModel.java
```diff
@@ -39,21 +54,21 @@ public void onXpChange(SetXpEvent event) {
         if (!Models.WorldState.onWorld()) return;
 
         // We don't want to track XP before we've even got a packet to set our level
-        if (Models.Character.getXpLevel() == 0) return;
+        if (Models.Experience.getXpLevel() == 0) return;
```
magicus: These are local now, so you can remove the `Models.Experience`.

common/src/main/java/com/wynntils/models/quests/QuestInfo.java
```diff
@@ -163,7 +163,7 @@ public static List<Component> generateTooltipForQuest(QuestInfo questInfo) {
         tooltipLines.add(Component.literal(""));
         // We always parse level as one, so check if this mini-quest does not have a min combat level
         if (!questInfo.isMiniQuest || questInfo.additionalRequirements.isEmpty()) {
-            tooltipLines.add((Models.Character.getXpLevel() >= questInfo.getLevel()
+            tooltipLines.add((Models.Experience.getXpLevel() >= questInfo.getLevel()
                             ? Component.literal("✔").withStyle(ChatFormatting.GREEN)
```
magicus, line 167: The QuestModel needs a new dependency as well...

magicus: This is actually a bit of a pain. :-/ 

I need to think if I can do some redesign so we can get rid of this inter-model dependency stuff. I have some ideas that might work.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -159,7 +159,7 @@ public Float getValue(String argument) {
     public static class LevelFunction extends Function<Integer> {
         @Override
         public Integer getValue(String argument) {
-            return Models.Character.getXpLevel();
```
magicus, line 162: These functions should move over to the new ExperienceFunctions class

common/src/main/java/com/wynntils/models/experience/ExperienceModel.java
```diff
@@ -0,0 +1,146 @@
...
+import com.wynntils.models.experience.event.ExperienceGainEvent;
+import com.wynntils.models.worlds.WorldStateModel;
+import com.wynntils.models.worlds.event.WorldStateEvent;
+import com.wynntils.models.worlds.type.WorldState;
+import com.wynntils.utils.mc.McUtils;
+import java.util.List;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class ExperienceModel extends Model {
```
magicus: I was thinking, we have consistently used "Xp" instead of "Experience" as name for this. But XpModel might sound a bit short. I would suggest using the "proper full name" of "Combat XP", so name this `CombatXpModel` (and also `CombatXpFunctions`). Would that be ok to you?

common/src/main/java/com/wynntils/models/experience/event/ExperienceGainEvent.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.experience.event;
+
+import net.minecraftforge.eventbus.api.Event;
+
+public class ExperienceGainEvent extends Event {
```
magicus: Maybe `CombatXpGainedEvent`?

common/src/main/java/com/wynntils/models/quests/QuestInfo.java
magicus: @kristofbolyai This still need fixing.

magicus (commented):
Sorry for making so much fuss. :) I think it is super important to get Models right, it is the foundation that will make or break our ability to work with and extend Artemis easily.

magicus (approved):

PR #1099, opened 2023-01-29
feat: Add Changelog Screen

kristofbolyai:


kristofbolyai:
image

kristofbolyai:
This is basically ready for release. When the new API comes, we can easily patch it to use it.

kristofbolyai:
This is now (hope)fully working, using the new API.

common/src/main/java/com/wynntils/features/user/ChangelogFeature.java
```diff
@@ -0,0 +1,46 @@
...
+
+            response.handleJsonObject(
+                    jsonObject -> {
+                        if (!jsonObject.has("changelog")) return;
+
+                        String changelog = jsonObject.get("changelog").getAsString();
+
+                        lastShownVersion = WynntilsMod.getVersion();
+
+                        Managers.TickScheduler.scheduleNextTick(
```
magicus, line 55: I think this is a bit scary, throwing the changelog in your face when you log in, and might be in a vulnerable situation.

 Legacy had some thing where you did not immediately did not connect if you pressed the W button on the main screen, and there were an update pending, instead it showed the changelog instead of logging in.

I have two suggestions:
1) Let the user login, but if we have updates and want to show a changelog, throw them over to `/hub`, unless they already are, first.

2) See if we can show the changelog at the main screen instead

3) Intercept the wynncraft server connection with a mixin and abort it if we want to show a changelog.

common/src/main/java/com/wynntils/screens/changelog/ChangelogScreen.java
```diff
@@ -0,0 +1,155 @@
...
+
+public class ChangelogScreen extends WynntilsScreen implements WynntilsPagedScreen {
+    private final String changelog;
+    private List<List<TextRenderTask>> changelogTasks;
+    private int currentPage = 0;
+
+    public ChangelogScreen(String changelog) {
+        super(Component.translatable("screens.wynntils.changelog.name"));
+
+        this.changelog = changelog + "\n" + changelog + "\n" + changelog + "\n" + changelog;
```
magicus: This looks just odd..?

common/src/main/java/com/wynntils/features/user/ChangelogFeature.java
kristofbolyai: We do not do any updating? This feature is to show the updates between last played and current version. I get what you are saying, but /hub-ing the user is undesirable behavior too.

common/src/main/java/com/wynntils/screens/changelog/ChangelogScreen.java
kristofbolyai: Oh.. Testing..

common/src/main/java/com/wynntils/utils/StringUtils.java
```diff
@@ -130,4 +130,8 @@ public static String toSignedString(int value) {
             return Integer.toString(value);
         }
     }
+
+    public static String stripMarkdownStyling(String input) {
+        return input.replaceFirst("[# ]*", "");
```
magicus: Can't we replace this with `§` color codes? It looks very bland, and actually a bit hard to read. Would be nice if headings and bullet items had different colors.

common/src/main/java/com/wynntils/features/user/ChangelogFeature.java
magicus: No, but we throw a screen in the face of the user, at a point in time where they might not be able to handle it.

common/src/main/java/com/wynntils/utils/StringUtils.java
kristofbolyai: Fixed

common/src/main/java/com/wynntils/features/user/ChangelogFeature.java
kristofbolyai: @magicus Try it now :). I think it works neatly :).

magicus: Ah, `/class` is fine, and better than hub! Good thinking. I assume you have tested both with and without our special class selection screen?

kristofbolyai: Havent tested without our screen, but that was the one giving problems anyways. I accounted for both options though

magicus (approved):

PR #1100, opened 2023-01-29
fix: Mixed bag of fixes (gear rewrite preparation)

magicus:
* fix: Crashed Features were not properly disabled, leading to crash when trying to re-enable them
* fix: In ItemScreenshotFeature, we cannot make chat encoded message of unidentified gear. Prohibit and warn for this.
* fix: Formatting and wording of crash error messages
* refactor: Move texture position information out of GearType and into UnidentifiedItemIconFeature
* refactor: Move fancy formatting out from ItemStatInfoFeature
* refactor: Extract out some helper functions that will be needed soon
* feat: Create basic type Element, and use it in element-assocated types
* feat: Add utils that will be needed by upcoming gear rewrite


kristofbolyai (approved):
Looks good. Thanks!

PR #1101, opened 2023-01-30
feat: Make ItemScreenshotFeature disk saving optional

kristofbolyai:


magicus:
I'm not sure I'm getting this. There is no problem with saving screenshots on mac. It's just saving to the clipboard that does not work.

magicus:
I had as a mental todo item to have another look at this feature. I think it is three things baked into one:
* Save a screenshot to disk
* Make a clipboard copy of the image
* Make a clipboard copy of the chat encoding

It's kind of confusing. For starter, having a separate hotkey for just copying for chat would make sense, I think.

kristofbolyai:
> I'm not sure I'm getting this. There is no problem with saving screenshots on mac. It's just saving to the clipboard that does not work.

The point is that people don't want to save item screenshots to disk unnecessarily.

magicus (approved):
I think we need to split this up into separate functionality in the future, but lets do this for now

PR #1104, opened 2023-01-31
fix: Hades events will work now, workaround for Forge EventBus issues #2 (Thread class loader altering)

kristofbolyai:

PR #1105, opened 2023-01-31
fix: Hades events will work now, workaround for Forge EventBus issues

kristofbolyai:


magicus:
Not needed, fixed by #1104.

PR #1106, opened 2023-02-01
chore: Remove misleading comment

kristofbolyai:


kristofbolyai:
As I said in Discord, doing this fits more with the concept of only running when we are in world. EDIT: Actually, we might as well use the event here. It really doesn't matter.

PR #1111, opened 2023-02-02
feat: Completely new gear and stat models

magicus:
This is part one of the gear rewrite. In this part, I have added all the new files. Due to a handful interactions with old code, this will not compile by it's own. But I think it will make the review process simpler, by highlighting the new architecture. Part 2 will contain the changes made to existing models and features to change them to use the new model instead, and part 3 finally will remove the old model. I would also like to add a part 4 afterwards, were I rename the `models.gearinfo` package to just `models.gear`. I kept it this way during development to have a clear separation between old and new, but basically `gearinfo` is a bad name for this, and I'd like to take over `gear` as soon as the old stuff is gone.

The mejor functionality is added here to three new models:
* StatModel
* GearModel
* RewardModel

(The latter is a minor model, and handles mastery tomes and charms, which is kind-of like gear but mostly kind-of like not.)

The "Stat" model is handling all different kinds of statuses gear can have. Some are fixed, but most are variable, meaning they will need to get rolled to an actual value by an Item Identifier. This mechanism is quite complex, and is handled by the StatModel.

The GearModel handles all aspects of a gear item. Basically, it splits the description of an (identified) gear item in two parts: the GearInfo, which is more or less the old GearProfile, and contains all *unchanging* info about a gear item, that is true for all instances of this item. The GearInstance, on the other hand, contains information of everything that makes this specific instance unique -- the identified stats, but also e.g. powders.


common/src/main/java/com/wynntils/models/gearinfo/GearInfoRegistry.java
```diff
@@ -0,0 +1,420 @@
...
+                Gson gearInfoGson = new GsonBuilder()
+                        .registerTypeHierarchyAdapter(GearInfo.class, new GearInfoDeserializer(allMajorIds))
+                        .create();
+                WynncraftGearInfoResponse gearInfoResponse =
+                        gearInfoGson.fromJson(reader, WynncraftGearInfoResponse.class);
+
+                // Remove the dummy "default" entry
+                List<GearInfo> registry = gearInfoResponse.items.stream()
+                        .filter(gearInfo -> !gearInfo.name().equals("default"))
+                        .toList();
```
kristofbolyai: What do we filter here? Do we remove a singular item from the API?

common/src/main/java/com/wynntils/models/gearinfo/GearInfoRegistry.java
```diff
@@ -0,0 +1,420 @@
...
+        }
+
+        private List<Pair<DamageType, RangedValue>> parseDamages(JsonObject json) {
+            List<Pair<DamageType, RangedValue>> list = new ArrayList<>();
+            // First look for neutral damage, which has a non-standard name
+            JsonElement neutralDamageJson = json.get("damage");
+            if (neutralDamageJson != null) {
+                String rangeString = neutralDamageJson.getAsString();
+                RangedValue range = RangedValue.fromString(rangeString);
+                if (!range.equals(RangedValue.NONE)) {
```
kristofbolyai: Can you unify this so you continue "early" like the code below?

common/src/main/java/com/wynntils/models/gearinfo/GearModel.java
```diff
@@ -0,0 +1,108 @@
...
+    public GearInstance parseInstance(GearInfo gearInfo, JsonObject itemData) {
+        GearParseResult result = GearParser.parseInternalRolls(gearInfo, itemData);
+
+        return GearInstance.create(gearInfo, result.identifications(), result.powders(), result.tierCount());
+    }
+
+    public CraftedGearItem parseCraftedGearItem(ItemStack itemStack) {
+        GearParseResult result = GearParser.parseItemStack(itemStack, null);
+        CappedValue durability = new CappedValue(result.tierCount(), result.durabilityMax());
+        // FIXME: Damages and requirements are not yet parsed
```
kristofbolyai: Is this upcoming after this whole rewrite?

common/src/main/java/com/wynntils/models/gearinfo/parsing/GearParseResult.java
```diff
@@ -0,0 +1,21 @@
...
+package com.wynntils.models.gearinfo.parsing;
+
+import com.wynntils.models.concepts.Powder;
+import com.wynntils.models.gearinfo.type.GearTier;
+import com.wynntils.models.gearinfo.type.GearType;
+import com.wynntils.models.stats.type.StatActualValue;
+import java.util.List;
+
+// "tierCount" depends on what item were parsed. For identified items, it is the
+// number of rerolls. For crafted items, it is the current durability
```
kristofbolyai: Bit weird.. But okay, it's fine. I assume this is just some internal helper class.

common/src/main/java/com/wynntils/models/gearinfo/parsing/GearParser.java
```diff
@@ -0,0 +1,211 @@
...
+            String normalizedCoded = WynnUtils.normalizeBadString(coded);
+
+            // Look for powder
+            Matcher powderMatcher = POWDER_PATTERN.matcher(normalizedCoded);
+            if (powderMatcher.matches()) {
+                int usedSlots = Integer.parseInt(powderMatcher.group(1));
+                String codedPowders = powderMatcher.group(3);
+                if (codedPowders == null) continue;
+
+                String powderString = codedPowders.replaceAll("[^✹✦❋❉✤]", "");
```
kristofbolyai: Can't you extract the pattern so it is not "patternized" every time we parse? (I am pretty sure you can do smth like `X_PATTERN.matcher(string).replaceAll(..)`)

kristofbolyai (commented):
From what I have seen, this looks promising. But I am nowhere near the end, so I will have to split this up into multiple review sessions.

common/src/main/java/com/wynntils/models/gearinfo/GearModel.java
magicus: I guess so. It's not *that* hard do to. But it kind of depends on what we want to do with the data! I don't even know how much we should care about crafted gear, or what do do with it. I had some idea of a "builder", and for that to know your exact stats, it needs to fully understand crafted gear. Otherwise, I frankly don't know. We don't actually use the identified stats we're parsing out of crafted gear right now either, it was more a matter of "well, it's simple to do, so why not do it?"

common/src/main/java/com/wynntils/models/gearinfo/GearInfoRegistry.java
magicus: Yes. This was the very last item as returned in the API, with I believe all stats at 0 and the lore text "This is a rad weapon yo. Let's make the lore longer so we can test how long it can go. Hi Jumla!". It did not show up in Wynnbuilder or Wynndata. So I decided to filter it out. 

Then I was told that there are more test items in the data set, so we should probably expand this filtering to check an entire "blocklist" of items that are not for real, and put this in a json file, so we can update it easier. 

I forgot about that, though. My suggestion is that we have just the one filtering of "default" right now, and write a more general filtering mechanism as a follow-up.

common/src/main/java/com/wynntils/models/gearinfo/GearInfoRegistry.java
magicus: Like in extracting common functionality between the neutral and elemental damages? I can try and do that.

The neutral damage check is not a loop so there is no corresponding keyword to `continue` to break out of the if statement, if that was what you meant.

kristofbolyai: Oh, right. You can see I was getting tired 😅. Code is fine, no need to extract stuff to save 2 lines or so.

common/src/main/java/com/wynntils/models/gearinfo/parsing/GearParseResult.java
magicus: Yes, I agree. It is a bit weird. It used to be just `rerollCount` until I started supporting crafted gear. 

You are right, this is just an internal class bundling the results of the gear parsing. 

An alternative is of course to change this to have like both "rerollCount" and a "currentDurability", and only ever set one of them. Maybe that is clearer?

kristofbolyai: Second is a bit cleaner but it really doesn't matter

common/src/main/java/com/wynntils/models/gearinfo/parsing/GearParser.java
magicus: Done.

common/src/main/java/com/wynntils/models/gearinfo/parsing/GearParseResult.java
magicus: Fixed.

common/src/main/java/com/wynntils/models/gearinfo/type/GearInstance.java
```diff
@@ -0,0 +1,69 @@
...
+import com.wynntils.models.concepts.Powder;
+import com.wynntils.models.stats.StatCalculator;
+import com.wynntils.models.stats.type.StatActualValue;
+import com.wynntils.models.stats.type.StatPossibleValues;
+import com.wynntils.models.stats.type.StatType;
+import java.util.DoubleSummaryStatistics;
+import java.util.List;
+import java.util.Optional;
+
+// FIXME: GearInstance is missing Powder Specials...
```
kristofbolyai: How so?

common/src/main/java/com/wynntils/models/gearinfo/type/GearTier.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static GearTier fromComponent(Component component) {
+        String name = component.getString();
+
+        if (name.charAt(0) == '§') {
+            return fromChatFormatting(ChatFormatting.getByCode(name.charAt(1)));
+        }
+
+        return null;
```
kristofbolyai: You could technically return `fromFormattedString(name)` to make it a bit simpler

common/src/main/java/com/wynntils/models/rewards/type/TomeType.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package com.wynntils.models.rewards.type;
+
+import com.wynntils.utils.StringUtils;
+import java.util.Optional;
+
+public enum TomeType {
+    WEAPON(true, true),
+    ARMOUR(true, true),
+    GUILD("Allegiance", true, false),
```
kristofbolyai: Do guild tomes really have a set name? I have a few and I don't remember this line

common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
```diff
@@ -0,0 +1,243 @@
...
+import com.wynntils.models.stats.type.StatType;
+import com.wynntils.utils.ListUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public final class StatListOrderer {
+    // Legacy order was defined by Athena as a fixed list. This was missing several stat types; I have
+    // tried filling them in into "logical" places
+    private static final List<String> LEGACY_ORDER = List.of(
```
kristofbolyai, line 23: Shouldn't this look more sane if we had `List<List<String>>` where there is a delimiter after each inner list? Although this is just purely looks.

common/src/main/java/com/wynntils/models/stats/builders/MiscStatKind.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Locale;
+
+public enum MiscStatKind {
+    // Health and mana
+    HEALTH("Health", StatUnit.RAW, "healthBonus"),
+    HEALTH_REGEN_PERCENT("Health Regen", StatUnit.PERCENT, "healthRegen"),
+    HEALTH_REGEN_RAW("Health Regen", StatUnit.RAW, "healthRegenRaw"),
+    LIFE_STEAL("Life Steal", StatUnit.PER_3_S, "lifeSteal"),
+    MANA_REGEN("Mana Regen", StatUnit.PER_5_S, "manaRegen"),
+    MANA_STEAL("Mana Steal", StatUnit.PER_3_S, "manaSteal"),
```
kristofbolyai, line 17: I will have to ping @P0keDev for this. Did 2.0 make it so mana stats are not `/ x seconds`? Should we keep them this way?

common/src/main/resources/assets/wynntils/urls.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "version": 2
+    "version": 3
```
kristofbolyai: I did just realize that I did not bump this when I added changelogs. Oops.

kristofbolyai (commented):
Looked through all files now. Commented at some places, otherwise looks fine.

common/src/main/java/com/wynntils/models/gearinfo/type/GearInstance.java
magicus: The powder slots we can see only tells you about the type of powder added (and how many), not their tier. If you have I think it is two tier 4 or above, you get the Powder Special. This is listed in the tooltip. We should be able to parse that. This is data that belongs in a GearInstance, since it is for a specific instance. If we knew this, we could (for instance) send it along with the chat encoding (but that would mean having a way to modify the protocol, which we have talked about in other situations).

In any way, this is just a reminder that there is more we could possibly extract, not that we are missing something we did before. I intended to fix this originally, but since it all just grew out of hand I had to leave it out.

common/src/main/java/com/wynntils/models/rewards/type/TomeType.java
magicus: I honestly don't know. I don't have a single Tome. I have copied all this code straight out from the old code which I believe @P0keDev added.

common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
magicus: My idea was that we should allow users to provide a `CUSTOM` ordering, in addition to the three predefined orderings by us. My plan was that this should take the shape of a json file containing an array of strings, just like here, where the "" were used to mark delimiters. 

But now when I think of it, of course we could even there make it a list of lists.

Ugh. I dunno. Maybe it would be better, but ... I'm not sure. I think this also gives a nice visual indication of where the separator goes. Like you know when you set up a toolbar and you drag in a "delimiter space" between buttons. 

If you are okay with this I'd rather not change it. If nothing else because I'm starting to get sick and tired on tweaking this code.

common/src/main/java/com/wynntils/models/stats/builders/MiscStatKind.java
magicus: What do you mean? They are `PER_5_S` or `PER_3_S` meaning, "per 3 seconds" (i.e. "/3 s"), or 5. This is as Wynncraft is. I have not changed anything in this area..?

Except the old code had Athena rewrite both per 5 and per 3 s to a single "per 4 s", and then we had to re-split them in the mod again. That was just stupid, and nothing I kept. Is that what you are thinking about?

common/src/main/resources/assets/wynntils/urls.json
magicus: Should we bump current to 3 now (better late than never?!?), and make it 4 in this PR? I guess if no-one has complained then maybe it does not matter so much now.

common/src/main/java/com/wynntils/models/gearinfo/type/GearTier.java
magicus: Nice observation. I think this was old code I did not touch, but it looks much better this way.

common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
kristofbolyai: Okay, it can stay.

common/src/main/resources/assets/wynntils/urls.json
kristofbolyai: Let's leave it, I guess. Seems like it doesn't matter and the bundled urls can do the job.

common/src/main/java/com/wynntils/models/rewards/type/TomeType.java
kristofbolyai: I checked and it seems to be true. I guess I forgot..

common/src/main/java/com/wynntils/models/stats/builders/MiscStatKind.java
kristofbolyai: > What do you mean? They are `PER_5_S` or `PER_3_S` meaning, "per 3 seconds" (i.e. "/3 s"), or 5. This is as Wynncraft is. I have not changed anything in this area..?
> 
> Except the old code had Athena rewrite both per 5 and per 3 s to a single "per 4 s", and then we had to re-split them in the mod again. That was just stupid, and nothing I kept. Is that what you are thinking about?

I think Wynn changed how mana works in 2.0. Maybe I am just so outdated in Wynn mechanics... Let's just leave this.

kristofbolyai (dismissed):
Part 1 looks good!

magicus:
Good! I'll hold on integrating this until all parts are accepted, otherwise chaos will ensure :-)

common/src/main/java/com/wynntils/models/gearinfo/tooltip/GearTooltipFooter.java
```diff
@@ -0,0 +1,87 @@
...
+import net.minecraft.network.chat.MutableComponent;
+
+public final class GearTooltipFooter {
+    private static final int PIXEL_WIDTH = 150;
+
+    public static List<Component> buildTooltip(GearInfo gearInfo, GearInstance gearInstance) {
+        List<Component> footer = new ArrayList<>();
+
+        // major ids
+        // FIXME: Missing "<+Entropy: >Meteor falls..." which should be in AQUA.
```
P0keDev: Is Entropy formatted differently by the game?

P0keDev: Upon further inspection of the code, it seems to me that _every_ major ID would be missing its aqua-colored name, unless I'm missing something

common/src/main/java/com/wynntils/models/rewards/type/TomeType.java
P0keDev: Yes, all tomes of a certain type follow the same name pattern - although I will admit this code isn't the cleanest representation of that

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
```diff
@@ -0,0 +1,197 @@
...
+    public static double getPerfectChance(StatPossibleValues possibleValues) {
+        // FIXME: This is the chance of getting a *** (3 star) roll, not the chance of
+        // getting the maximum possible value. But for now, keep old behavior.
+        double perfectChance = 1 / (possibleValues.baseValue() > 0 ? 101d : 61d) * 100;
+        return perfectChance;
+    }
+
+    public static double getDecreaseChance(StatActualValue actualValue, StatPossibleValues possibleValues) {
+        // If we are in any of these situations, our stat has "invalid" values. Maybe it is old?
+        // If so, rerolling will have a certain outcome of 100% or 0% chance of decrease
```
P0keDev: An "old" roll can't give you an invalid value - the only time this would happen (bar Wynncraft mechanics fundamentally changing) is if our item API info is outdated. In that case, I don't think we should display any percentage, and it certainly shouldn't be 0 or 100

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
```diff
@@ -0,0 +1,197 @@
...
+        if (innerRollRange.low() > 0) {
+            result = (avg - 30) / 101d;
+        } else {
+            result = (130 - avg) / 61d;
+        }
+        return result * 100;
+    }
+
+    public static double getIncreaseChance(StatActualValue actualValue, StatPossibleValues possibleValues) {
+        // If we are in any of these situations, our stat has "invalid" values. Maybe it is old?
```
P0keDev: The comment on decrease chances applies here as well

common/src/main/java/com/wynntils/models/stats/builders/MiscStatKind.java
P0keDev: The stats are still displayed the same way and do still represent the "per x seconds" value - what changed is how exactly the buffs are applied to the player. I.e., having 20/5s mana regen means you'll gain 4 mana per second

P0keDev (commented):
Overall this looks good and I like the overall design and structure. I only have a few concerns about specific things which I commented on

P0keDev:
One more comment, actually: As far as I can tell the tooltip builder implementations would fail to display major IDs/variable identifications present on an item but not reflected in our API data (and thus its GearInfo). Am I correct in this? Because if so, I would strongly prefer to have that particular edge case function differently

common/src/main/java/com/wynntils/models/gearinfo/tooltip/GearTooltipFooter.java
magicus: Yeah, sorry, the comment was misleading, it should read "for example, like this:"

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
magicus: So, the thing is, in the end, I did not rewrite the calculation code, I just took the old one and refactored it step by step to make it more sane (turn it into functions instead of a weird class which copied itself etc). I did not dare change the current behavior.

But my understanding here, which is reflected in the comment I wrote, is that if you have an item, which has an actual value outside the range of possible values, is either a bug or perhaps an old item where the Wynncraft gear type has changed but your specific item has not been updated (don't know if this can really happen).

But, I think the percentage is still correct. Note that you are looking at the "get decrease chance" function, not the basic "get percent of possible rolls". So if the range is 1-10, and you have a 13, then I claim that any reroll has 100% chance of getting lower, 0% chance of getting higher, and it would have a stat percentage of 130%. This is the best way we can describe this scenario if it happens, and this is what the code currently does.

If this scenario can ever happen, I don't know. But it's a simple piece of defensive programming to have this check.

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
magicus: And my reply as well :)

magicus:
> One more comment, actually: As far as I can tell the tooltip builder implementations would fail to display major IDs/variable identifications present on an item but not reflected in our API data (and thus its GearInfo). Am I correct in this? Because if so, I would strongly prefer to have that particular edge case function differently

No, that is not the case. 

We have two ways of creating a tooltip builder: from an existing tooltip, and from scratch. 

From scratch is used for the gear guide, for items sent over chat, and when inspecting another player. In this case, we can only add lines to the tooltip that correspond to data that is present in the API. I've tried my very best to make these lines look 100% identical to what Wynncraft displays (this was not fully the case before). 

The one missing thing is the major ID which only displays the lore and not the name, as you saw above. (We use a function that splits a long line into several parts, but it only took in a String and not a formatted component, and I gave up on trying to create a counterpart for Components, and left it unsolved for now). 

If we start with an existing item stack, we split the existing lore in three parts, the pre-IDs, the IDs, and the post-IDs. The pre- and post- parts are kept unchanged (and post-ID is where MajorIDs are stored). Only the ID lines are re-created by us (since we do sorting, grouping, and adding percent numbers etc). So Set info, Powder Specials, Average DPS etc, are all kept.

If, during parsing, we run into a stat that we do not recognize as being a part of the API, we will print a warning to the console, but I think we will drop it from our parsing. It would perhaps be more defensive to include it as part of the GearInstance, and when printing it, acknowledge that we have no GearInfo about this stat, so we can't display any ranges etc, but do not drop it. I have perhaps not thought fully about this.

Also, at some point, I'd like to see us being able to calculate *all* of these aspects, since it would make a 100% perfect builder possible. But that's a future project.

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
P0keDev: What you're describing is an impossible situation given the way Wynncraft currently works - so this code is defensive against a nonexistent scenario.

If a stat range is 1-10 and its actual value is 13, that doesn't mean that it's guaranteed to decrease after a reroll; it means our stat range is wrong, and we can't calculate anything about its reroll chances. To display any percentage at all is misleading, IMO

P0keDev:
Ah, I didn't realize that about major IDs. In that case I have no issue there. However:

> If, during parsing, we run into a stat that we do not recognize as being a part of the API, we will print a warning to the console, but I think we will drop it from our parsing. It would perhaps be more defensive to include it as part of the GearInstance, and when printing it, acknowledge that we have no GearInfo about this stat, so we can't display any ranges etc, but do not drop it. I have perhaps not thought fully about this.

This isn't entirely what I meant, although this is a separate possible point of improvement.  An item's GearInstance _will_ contain every ID, assuming it exists in the API somewhere - but GearTooltipIdentifications doesn't iterate through the list of collected IDs, it uses the list from the passed GearInfo, which is potentially outdated/wrong compared to the reality of Wynn. 

This shouldn't be too hard to change, but it's also not an urgent problem at the moment; I'm willing to accept this PR as-is for now and correct it later on.

P0keDev (dismissed):

magicus:
> If, during parsing, we run into a stat that we do not recognize as being a part of the API, we will print a warning to the console, but I think we will drop it from our parsing. It would perhaps be more defensive to include it as part of the GearInstance, and when printing it, acknowledge that we have no GearInfo about this stat, so we can't display any ranges etc, but do not drop it. I have perhaps not thought fully about this.

I fixed this now. The code was *almost* correct, we collected the info correctly, but did not account for the possibility that there were no associated "possible values" when building the tooltip.

However, I am trying to understand your last comment. .... 💡 Ah, now I get it! That sounds like an easy fix, I'll do that as well.

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
magicus: Ok, I'll remove the defensive part.

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public static double getPerfectChance(StatPossibleValues possibleValues) {
+        // FIXME: This is the chance of getting a *** (3 star) roll, not the chance of
+        // getting the maximum possible value. But for now, keep old behavior.
+        double perfectChance = 1 / (possibleValues.baseValue() > 0 ? 101d : 61d) * 100;
+        return perfectChance;
+    }
+
+    public static double getDecreaseChance(StatActualValue actualValue, StatPossibleValues possibleValues) {
+        assert !possibleValues.range().isFixed();
```
P0keDev, line 138: just to confirm, this won't have any effect in a prod environment, right?

P0keDev (dismissed):

common/src/main/java/com/wynntils/models/stats/StatCalculator.java
magicus: No, unless the user for some reason adds `-ea` (enable asserts) to their command line.

PR #1112, opened 2023-02-02
feat: Completely new gear and stat models, part 2

magicus:
This is the second part, where I update all existing models and features to use the new gear/stat model. Note that this will not compile by itself, as some of the old code (which is no longer used) has a handful of references to other models that has changed. This will be resolved in part 3 were I will remove these files.

Some noteworthy changes in this PR:

GearItem now consequently consists of just a GearInfo and a GearInstance, since that is (exactly) what we need to describe a gear item. An unidintified item (named item, not a "gear box") will have a `null` GearInstance.

Most changes to features and models are just due to changed packages, and that the new types are mostly records (making getters like `name()` instead of `getName()`), but there are some larger changes in features that interact much with the gear system. Most affected of all is (of course)`ItemStatInfoFeature`, which is for some parts completely rewritten. It also have additional new configs to tweak the display, and a new identified stats suffix, by pressing left shift and control at the same time: this will show the internal roll of the item (or more precisely, the lowest and highest possible values for the internal roll, since in most cases we cannot determine it exactly).

The Guide item stacks has been changed to provide "real" tooltips, so the normal Vanilla tooltip rendering could be used. This allowed ItemStatInfoFeature to apply its settings also to Guide item stacks. This was technically only needed for the Item Guide, but I thought it was a) much better than the old system, and b) good to keep this consistent across all the guides.


magicus:
A comment: There are a handful of `FIXME`s in the code. These are either:
* FIXMEs ported over from the old, corresponding code, or
* Things I think should be fixed, but was broken before, so there is no regression, or
* Some possibility for future expansion that I think would be reasonable

There are no known new bugs or regression in this code.

magicus:
Also, the Ingredient stuff that has changes is basically me copying over the minimal set of code that is needed from the old gear package, before deleting it in the next step.

Ingredients are tightly related to gear, at last they share the same stats, so a reasonable continuation of this is for me to address ingredients next.

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
```diff
@@ -58,14 +65,20 @@ public class ItemStatInfoFeature extends UserFeature {
...
     public boolean groupIdentifications = true;
 
+    @Config
+    public boolean identificationDecorations = true;
+
     @Config
     public boolean overallPercentageInName = true;
 
+    @Config
+    public boolean showBestValueLastAlways = true;
```
kristofbolyai, line 80: I have no idea what this does. But I assume we have not changed behaviour from before? Or did we?

common/src/main/java/com/wynntils/models/ingredients/IngredientProfilesModel.java
```diff
@@ -69,6 +73,16 @@ private void tryLoadIngredientList() {
         });
     }
 
+    private void tryLoadInvertedList() {
+        Download dl = Managers.Net.download(UrlId.DATA_ATHENA_ITEM_LIST);
+        dl.handleJsonObject(json -> {
+            Type hashmapType = new TypeToken<HashMap<String, String>>() {}.getType();
```
kristofbolyai: Is this used?

common/src/main/java/com/wynntils/models/ingredients/type/IdentificationModifier.java
```diff
@@ -0,0 +1,26 @@
...
+
+    IdentificationModifier(String inGame) {
+        this.inGame = inGame;
+    }
+
+    public String getInGame(String name) {
+        if (this != FOUR_SECONDS) return inGame;
+
+        if (name.equals("manaRegen")) return "/5s";
+        return "/3s";
```
kristofbolyai: DIdn't you just say you removed this witchcraft in part 1? Am I completely losing it?

magicus: You are looking at ingredients. I have not converted them yet, just moved the code from the old "GearProfile" things over here. (But in this split-up review it looks like added code, rather than moved)

kristofbolyai: Oh, I have missed the fact this is ingredients... Got it!

common/src/main/java/com/wynntils/models/ingredients/IngredientProfilesModel.java
magicus: Yes, for ingredients. I could almost, but not completely, remove this. I think it was some additional metadata that was needed by the ingredients. See also my previous comment below for why ingredient looks like they have new code even though it doesn't.

kristofbolyai: Github reviews.. I meant the variable `hashmapType`. I can't see any use for it?

common/src/main/java/com/wynntils/screens/guides/gear/WynntilsItemGuideScreen.java
```diff
@@ -95,27 +91,7 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
...
-                    .withStyle(ChatFormatting.RED));
-            RenderUtils.drawTooltipAt(
-                    poseStack,
-                    mouseX,
-                    mouseY,
-                    0,
-                    tooltipLines,
-                    FontRenderer.getInstance().getFont(),
-                    true);
+            this.renderTooltip(poseStack, guideGearItemStackButton.getItemStack(), mouseX, mouseY);
```
kristofbolyai, line 94: Is Qodana valid here?

common/src/main/java/com/wynntils/screens/guides/ingredient/WynntilsIngredientGuideScreen.java
```diff
@@ -119,29 +116,7 @@ protected void renderTitle(PoseStack poseStack, String titleString) {
...
-                    .withStyle(ChatFormatting.RED));
-            RenderUtils.drawTooltipAt(
-                    poseStack,
-                    mouseX,
-                    mouseY,
-                    0,
-                    tooltipLines,
-                    FontRenderer.getInstance().getFont(),
-                    true);
+            this.renderTooltip(poseStack, guideGearItemStack.getItemStack(), mouseX, mouseY);
```
kristofbolyai, line 119: (Qodana here too? I would assume it is a bug in Qodana?)

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
kristofbolyai: I mostly figured this out from i18n.

kristofbolyai (commented):
Looks good after resolving the reviews.

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
magicus: In Wynncraft vanilla, there is no clear consistency which value is put where in a range, so "X to Y" could mean either that X is the worst possible value (and Y is the best), or that Y is the worst possible value (and X is the best).

If this is false, we will copy Wynncraft behavior. If this is true, we will always go from "worst" to "best".

The problems arise when you have like spell costs which has negative values (but "more negative" is better), and where a "positive" stat has a negative value (like -5 health regen). Iirc, Wynntils also makes this different depending on whether an item is in the Trade Market or in your inventory. This was too complex for me to handle to replicate, so I picked one of the two methods to use.

common/src/main/java/com/wynntils/models/ingredients/IngredientProfilesModel.java
magicus: No, you are right. I copied old code and pruned unused parts, but missed this. Fixed now.

common/src/main/java/com/wynntils/screens/guides/gear/WynntilsItemGuideScreen.java
magicus: It looks really scary, yes. But I can't see anything but that Qodana is just flat out wrong:
```
	public void renderTooltip(PoseStack poseStack, ItemStack itemStack, int mouseX, int mouseY) {
```
is what we are calling.

common/src/main/java/com/wynntils/screens/guides/ingredient/WynntilsIngredientGuideScreen.java
magicus: The method in question resides in the superclass, which is `Screen` in vanilla code. Maybe there is some kind of source mapping issue, since I assume method arguments are renamed during obfuscation.

And yes, the code is correct here too.

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
P0keDev: isn't this exclusively the case with spell cost IDs, thanks to the way they were recently changed? i'm not sure if it's worth worrying about replicating what is effectively a bug on wynn's end

P0keDev:
> Also, the Ingredient stuff that has changes is basically me copying over the minimal set of code that is needed from the old gear package, before deleting it in the next step.
> 
> Ingredients are tightly related to gear, at last they share the same stats, so a reasonable continuation of this is for me to address ingredients next.

it would be nice if this was true, but sadly it's not: ingredient IDs aren't defined by base values, but instead by arbitrary min/max values. the `IdentificationProfile` class, for instance, wouldn't apply to ingredients at all

edit: well now i'm confused because `IngredientIdentificationContainer` is still here and i don't see where `IdentificationHolder` is used

edit2: seems like it's only used for its static name conversion method? if that's the case and i'm not missing another usage, i'd say it's rather unnecessary to have the class here

P0keDev:
overall looks good besides my concerns above

common/src/main/java/com/wynntils/features/user/tooltips/ItemStatInfoFeature.java
```diff
@@ -123,15 +131,90 @@ public void onTooltipPre(ItemTooltipRenderEvent.Pre event) {
...
+                    .append(Component.literal(String.format(
+                                    Locale.ROOT,
+                                    " \u21E9%.1f%%",
+                                    StatCalculator.getDecreaseChance(actualValue, possibleValues)))
+                            .withStyle(ChatFormatting.RED));
+
+            return rerollChancesComponent;
+        }
+
+        private MutableComponent getPercentSuffix(
```
P0keDev, line 218: there doesn't seem to be any handling for values that are above/below the API range? these IDs should have the NEW tag applied and prevent the item's overall percentage from being calculated

magicus: That was the handling I kind of removed in Part 1 since you claimed that we could not lie outside the API range. ;-)

But sure, I can add something back in that replaces "illegal" values with the text "NEW".

magicus:
> it would be nice if this was true, but sadly it's not: ingredient IDs aren't defined by base values, but instead by arbitrary min/max values. the `IdentificationProfile` class, for instance, wouldn't apply to ingredients at all
> 
> edit: well now i'm confused because `IngredientIdentificationContainer` is still here and i don't see where `IdentificationHolder` is used
> 
> edit2: seems like it's only used for its static name conversion method? if that's the case and i'm not missing another usage, i'd say it's rather unnecessary to have the class here

What you are missing is that you are just seeing a PR of the partial changes. In the next step I will remove the old IngredientIdentificationContainer. So this particular case might be easier to understand if you look at the PR for the combined changes: https://github.com/Wynntils/Artemis/pull/1114/files. Scroll down to the `ingredients` package.

magicus:
And also, I have started a follow-up change where I rewrite identifications according to the same pattern. Since the entire StatModel exists, this is a much simpler process. You are correct in that ingredients have no base value, but that is not really a problem, since the baseValue from StatPossibleValues is only used in some very special cases; normally the range is all that is used, and this is shared with ingredients.

magicus:
Heh, I actually checked the "merged" PR I linked to, and it did not show the moved IdentificationProfile as a moved file. :-( But anyway, that is what has happened. I could not get rid of it fully since ingredients still used it, so I just moved it there, and made a few tweaks to get it to work with most of the old `gear` package removed. That is just life-support to keep it functioning, as I said, I am about to replace that too.

PR #1113, opened 2023-02-02
feat: Completely new gear and stat models, part 3

magicus:
Now finally the old code is removed, and Artemis builds correctly again.

This is the final part of the review. All three needs to go in together, but I believe it will be easier to review it separately.


kristofbolyai (approved):
Not much to review here...

PR #1114, opened 2023-02-02
[REFERENCE] feat: Completely new gear and stat models [all parts]

magicus:
For reference, here is a PR that contains all changes at the same time. This should not be pushed.


magicus:
One benefit of this review PR, compared to the step-wise, is that moved files show up properly here.

PR #1115, opened 2023-02-02
feat: Add textured lootrun path

HalcyonSuoh:


HalcyonSuoh:
I gave up trying to interpolate rotation, I'm grateful if someone more knowledgable could take over

common/src/main/java/com/wynntils/utils/render/buffered/CustomRenderType.java
```diff
@@ -89,10 +89,30 @@ public class CustomRenderType extends RenderType {
...
+                    256,
+                    false,
+                    false,
+                    CompositeState.builder()
+                            .setShaderState(POSITION_COLOR_TEX_SHADER)
+                            .setCullState(NO_CULL)
+                            .setTextureState(new TextureStateShard(resource, false, false))
+                            .setTransparencyState(CustomRenderStateShard.SEMI_TRANSPARENT_TRANSPARENCY)
+                            .setWriteMaskState(COLOR_WRITE)
+                            .createCompositeState(false)));
```
kristofbolyai: We don't need to use memoize here, do we? We can just hard-code the lootrun texture here and it will be a bit simpler.

common/src/main/java/com/wynntils/models/lootruns/LootrunRenderer.java
```diff
@@ -205,4 +282,75 @@ private static void renderPoint(VertexConsumer consumer, Matrix4f lastMatrix, Co
...
+        Vector3f pos2 = new Vector3f(0.5f, 0.24f, -0.5f);
+        Vector3f pos3 = new Vector3f(0.5f, 0.24f, 0.5f);
+        Vector3f pos4 = new Vector3f(-0.5f, 0.24f, 0.5f);
+
+        Vec3 direction =
+                new Vec3(endVec.x, endVec.y, endVec.z).subtract(startVec).normalize();
+
+        // rotation angle to point surface normal to end position
+        // 1.5707963267948966 is the result of Math.toRadians(90f), so the arrow point to the end position instead
+        float xAngle = (float) ((float) Math.acos(direction.y / direction.length()) - 1.5707963267948966);
```
kristofbolyai: You can use Math.PI / 2 and it will look nicer :) (and the compiler will optimize for it)

common/src/main/java/com/wynntils/models/lootruns/LootrunRenderer.java
```diff
@@ -191,6 +202,72 @@ private static void renderNonTexturedLootrunPoints(
...
+
+        for (ColoredPath locationsInRoute : locations) {
+            VertexConsumer consumer = BUFFER_SOURCE.getBuffer(renderType);
+            List<Pair<ColoredPoint, ColoredPoint>> toRender = new ArrayList<>();
+            boolean sourceBatchEnded = false;
+            BlockPos lastBlockPos = null;
+
+            boolean pauseDraw = false;
+
+            for (int i = 0; i < locationsInRoute.points().size() - 1; i += 10) {
```
kristofbolyai, line 215: Why size - 1?

common/src/main/java/com/wynntils/models/lootruns/LootrunRenderer.java
```diff
@@ -205,4 +282,75 @@ private static void renderPoint(VertexConsumer consumer, Matrix4f lastMatrix, Co
...
+        Quaternionf xRot = new Quaternionf().rotateAxis(xAngle, xRotAxis);
+
+        // apply vertex rotation
+        pos1.rotate(yRot).rotate(xRot);
+        pos2.rotate(yRot).rotate(xRot);
+        pos3.rotate(yRot).rotate(xRot);
+        pos4.rotate(yRot).rotate(xRot);
+
+        // transform position back to world space and then to position camera delta
+        pos1 = pos1.add(startVec.toVector3f()).sub(camPos.toVector3f());
```
kristofbolyai: Only allocate startVec/camPos.toVector3f once (for a minor performance gain)

common/src/main/java/com/wynntils/models/lootruns/LootrunRenderer.java
HalcyonSuoh: it pass the next point to renderTexturedPoint to get it's rotation, if the operation is performed on the last point, there won't be a next point to pass onto

kristofbolyai (approved):
Nice! Thanks.

PR #1129, opened 2023-02-07
feat: Ingredient rewrite

magicus:


common/src/main/java/com/wynntils/models/gear/parsing/GearParser.java
```diff
@@ -161,10 +161,6 @@ public static GearParseResult parseInternalRolls(GearInfo gearInfo, JsonObject i
                 // First convert loretype (e.g. DAMAGEBONUS) to our StatTypes
                 StatType statType = Models.Stat.fromInternalRollId(id);
                 if (statType == null) {
-                    // This can happen for skill point bonus, which used to be variable...
-                    Skill skill = Skill.fromApiId(id.replace("POINTS", ""));
```
P0keDev, line 165: was this removed because it wasn't necessary/effectively did nothing, or for another reason?

common/src/main/java/com/wynntils/models/items/annotators/game/IngredientAnnotator.java
```diff
@@ -25,15 +25,16 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, String name) {
...
-        int tier = Models.IngredientProfiles.getTierFromColorCode(tierColor);
 
-        IngredientProfile ingredientProfile = Models.IngredientProfiles.getIngredient(ingredientName);
-        if (ingredientProfile == null) return null;
-        if (ingredientProfile.getTier().getTierInt() != tier) {
+        int tier = Models.Ingredient.getTierFromColorCode(tierColor);
+        IngredientInfo ingredientInfo = Models.Ingredient.getIngredientInfoFromName(ingredientName);
+        if (ingredientInfo == null) return null;
+
+        if (ingredientInfo.tier() != tier) {
```
P0keDev, line 33: i'm always wary of "requiring" an item to have matching API data to be fully processed, and this is no exception. i would much prefer it if ingredients could still get basic functionality like inventory highlights even if we can't find a matching `IngredientInfo`

common/src/main/java/com/wynntils/models/items/annotators/gui/IngredientPouchAnnotator.java
```diff
@@ -30,24 +30,25 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, String name) {
...
-            int tier = Models.IngredientProfiles.getTierFromColorCode(tierColor);
 
-            IngredientProfile ingredientProfile = Models.IngredientProfiles.getIngredient(ingredientName);
-            if (ingredientProfile == null) return null;
-            if (ingredientProfile.getTier().getTierInt() != tier) {
+            int tier = Models.Ingredient.getTierFromColorCode(tierColor);
+            IngredientInfo ingredientInfo = Models.Ingredient.getIngredientInfoFromName(ingredientName);
+            if (ingredientInfo == null) return null;
+
+            if (ingredientInfo.tier() != tier) {
```
P0keDev, line 48: previous comment is especially true here - in this case you're throwing the baby out with the bathwater IMO, if one unknown ingredient can stop the entire pouch from being processed

common/src/main/java/com/wynntils/screens/guides/ingredient/GuideIngredientItemStack.java
```diff
@@ -67,9 +61,24 @@ public List<Component> getTooltipLines(Player player, TooltipFlag isAdvanced) {
...
+    }
+
+    private String getTierString(int tier) {
+        return switch (tier) {
+            case 0 -> ChatFormatting.GRAY + "[" + ChatFormatting.DARK_GRAY + "✫✫✫" + ChatFormatting.GRAY + "]";
+            case 1 -> ChatFormatting.GOLD + "[" + ChatFormatting.YELLOW + "✫" + ChatFormatting.DARK_GRAY + "✫✫"
+                    + ChatFormatting.GOLD + "]";
+            case 2 -> ChatFormatting.DARK_PURPLE + "[" + ChatFormatting.LIGHT_PURPLE + "✫✫" + ChatFormatting.DARK_GRAY
+                    + "✫" + ChatFormatting.DARK_PURPLE + "]";
+            case 3 -> ChatFormatting.DARK_AQUA + "[" + ChatFormatting.AQUA + "✫✫✫" + ChatFormatting.DARK_AQUA + "]";
```
P0keDev: seeing all these switch statements on a tier integer makes me think we ought to have an enum for this, mirroring the one for gear item tiers

P0keDev (commented):
Overall looks good, just a few small things I noticed

common/src/main/java/com/wynntils/models/gear/parsing/GearParser.java
P0keDev: ah, i see now that this functionality was moved elsewhere

common/src/main/resources/assets/wynntils/urls.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "version": 5
```
kristofbolyai, line 3: Can you make a PR for API repo?

common/src/main/java/com/wynntils/utils/JsonUtils.java
```diff
@@ -31,4 +31,12 @@ public static boolean getNullableJsonBoolean(JsonObject json, String key) {
 
         return jsonElement.getAsBoolean();
     }
+
+    public static JsonObject getNullableJsonObject(JsonObject json, String key) {
```
kristofbolyai, line 35: I expected this method to return null from the name. I can't find a better name though.

common/src/main/java/com/wynntils/screens/guides/ingredient/GuideIngredientItemStackButton.java
```diff
@@ -68,13 +69,28 @@ public void renderButton(PoseStack poseStack, int mouseX, int mouseY, float part
         }
     }
 
+    // FIXME: This should be painted by ItemHighlightFeature instead...
+    private CustomColor getHighlightColor(int tier) {
+        return switch (tier) {
+            case 0 -> new CustomColor(102, 102, 102);
+            case 1 -> new CustomColor(255, 247, 153);
+            case 2 -> new CustomColor(255, 255, 0);
+            case 3 -> new CustomColor(230, 77, 0);
```
kristofbolyai: Is this not the same memory leak concept I asked about you earlier? Can we make this static?

kristofbolyai: (not leak..., but high allocation, since this is called render time.)

common/src/main/java/com/wynntils/screens/guides/ingredient/GuideIngredientItemStack.java
kristofbolyai: It would also solve https://github.com/Wynntils/Artemis/pull/1129/files#r1116992319

common/src/main/java/com/wynntils/models/ingredients/type/IngredientTier.java
```diff
@@ -1,56 +0,0 @@
...
- * Copyright © Wynntils 2022.
- * This file is released under AGPLv3. See LICENSE for full license details.
- */
-package com.wynntils.models.ingredients.type;
-
-import com.google.gson.annotations.SerializedName;
-import com.wynntils.utils.colors.CustomColor;
-import net.minecraft.ChatFormatting;
-
-public enum IngredientTier {
```
kristofbolyai, line 11: Both me and Poke mentioned this already, but why was this removed?

kristofbolyai (commented):
This should be all. I think it looks great overall, I liked the changes to net data loading. 

The only thing I don't get is why IngredientTier was removed. I already mentioned it in the reviews though.. Other than that (and maybe Poke's comment on wrong data on a single ingredient making nothing load), this looks acceptable even in this state.

common/src/main/java/com/wynntils/models/items/annotators/game/IngredientAnnotator.java
magicus: I understand. But please note that this is not really a change compared to before; we required an IngredientProfile then, and an IngredientInfo now. It is the same thing but different name.

Handling "wynnitems" (gear and ingredients) that are not found in our database "as best we could" is certainly some kind of robustness improvement. But it also makes things more difficult, since then suddenly you don't know if your ingredient has an associated IngredientInfo or not, and need to handle that (probably very unlikely) case everywhere. And if you forget it, you get a NPE which is arguably worse. Or we have to juggle around Optionals everywhere.

I've been thinking of trying to make more elegant parsing of crafted items; possibly such code could spill over and maybe we could have like UnrecognizedGear and UnrecognizedIngredient? 

I hope you agree that we can take this code as it is, and continue this discussion afterwards. (I fully understand your concern, I'm just saying it's not clear to me how to do this so we don't pay a price that is higher than the gain.)

common/src/main/java/com/wynntils/models/items/annotators/game/IngredientAnnotator.java
```diff
@@ -25,15 +25,16 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, String name) {
...
             WynntilsMod.warn("Incorrect tier in ingredient database: " + ingredientName + " is " + tier);
             return null;
```
magicus: I did remove this return null though, it is unnecessary. We can warn for the inconsistency, but we have something to go with, it's better than nothing.

common/src/main/java/com/wynntils/models/items/annotators/gui/IngredientPouchAnnotator.java
magicus: Once again, this is not new behavior. But I can fix so we just ignore unknown ingredients in the pouch, that sounds like a better fallback option yes.

common/src/main/java/com/wynntils/screens/guides/ingredient/GuideIngredientItemStackButton.java
magicus: Yes I can, but see also the comment above:
```
 // FIXME: This should be painted by ItemHighlightFeature instead...
```

I'm not really happy about this, but did not have stamina to continue fighting with this code. I'd rather see the guide item stacks being drawn as normal item stacks, so our item stack features can do their job. Now we duplicate that code, and we lose any settings the user has made for these colors. 

But I'll make them static for now, it might be a lousy design but there's no need for it do have lousy memory footprint as well.

common/src/main/java/com/wynntils/models/ingredients/type/IngredientTier.java
magicus: I thought it mixed different areas of concern. How to represent the tier as a string was only really needed for the guide to know. The background color as well, and as I said above, my plan was to change so the guide item stack did not duplicate features with hard-coded values. But I never got around to fixing it. So, it might be that the best/simplest version is to restore this for now, and then I can see if I can make it superfluous in an upcoming PR.

common/src/main/java/com/wynntils/utils/JsonUtils.java
magicus: I thought all these JsonUtils method did the same (and returned a valid object) but I see now that the String version returns null instead of an empty string. I think I should fix that. And maybe then we can rename the methods, either just drop "Nullable" completely, or try weaving in a "safe" or something like that, e.g. `getJsonObjectSafe` or such....

common/src/main/resources/assets/wynntils/urls.json
magicus: Sure, thanks for reminding me! https://github.com/Wynntils/WynntilsWebsite-API/pull/46

PR #1134, opened 2023-02-07
fix: Item Lock's block all mode now works correctly if HEAD slot is locked

kristofbolyai:


ryanzhoudev (approved):
seems to work well in-game

magicus (approved):

PR #1139, opened 2023-02-08
chore: More IntelliJ fixes

magicus:
These are manually selected from all issues notified by the "Artemis" profile.

### DIFF
--- a/common/src/main/java/com/wynntils/core/config/ConfigManager.java
+++ b/common/src/main/java/com/wynntils/core/config/ConfigManager.java
@@ -33,9 +33,9 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.reflect.FieldUtils;
 
@@ -122,11 +122,11 @@ public final class ConfigManager extends Manager {
         }
     }
 
-    public void loadAllConfigOptions(boolean resetIfNotFound) {
-        loadConfigOptions(CONFIG_HOLDERS, resetIfNotFound);
+    public void loadAllConfigOptions() {
+        loadConfigOptions(CONFIG_HOLDERS, true);
     }
 
-    public void loadConfigOptions(List<ConfigHolder> holders, boolean resetIfNotFound) {
+    private void loadConfigOptions(List<ConfigHolder> holders, boolean resetIfNotFound) {
         if (configObject == null) {
             WynntilsMod.error("Tried to load configs when configObject is null.");
             return; // nothing to load from
@@ -260,8 +260,8 @@ public final class ConfigManager extends Manager {
         return options;
     }
 
-    public List<ConfigHolder> getConfigHolders() {
-        return Collections.unmodifiableList(CONFIG_HOLDERS);
+    public Stream<ConfigHolder> getConfigHolders() {
+        return CONFIG_HOLDERS.stream();
     }
 
     public Object deepCopy(Object value, Type fieldType) {
--- a/common/src/main/java/com/wynntils/core/config/ConfigManager.java
+++ b/common/src/main/java/com/wynntils/core/config/ConfigManager.java
@@ -33,6 +33,7 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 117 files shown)


common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -260,7 +261,7 @@ private List<ConfigHolder> getConfigOptions(Configurable parent) {
     }
 
     public List<ConfigHolder> getConfigHolders() {
-        return CONFIG_HOLDERS;
+        return Collections.unmodifiableList(CONFIG_HOLDERS);
```
kristofbolyai: We only of of this in the wrong place and we will allocate a lot memory, won't we? Why do we need this anyways?

common/src/main/java/com/wynntils/features/user/CommandAliasesFeature.java
```diff
@@ -63,35 +63,5 @@ public void onCommandPacket(CommandsPacketEvent event) {
...
-            CommandAlias that = (CommandAlias) other;
-            return originalCommand.equals(that.originalCommand) && aliases.equals(that.aliases);
-        }
-
-        @Override
-        public int hashCode() {
-            return Objects.hash(originalCommand, aliases);
-        }
-    }
+    private record CommandAlias(String originalCommand, List<String> aliases) {}
```
kristofbolyai: Isn't this used in @Config? That means GSON won't work.

kristofbolyai (commented):
I can't say I like the unmodifiable collection creation everywhere

magicus:
> I can't say I like the unmodifiable collection creation everywhere

I can leave that out of this PR, so we can discuss it separately. I think it makes sense to improve the robustness of the code, but I am willing to discuss it.

common/src/main/java/com/wynntils/features/user/CommandAliasesFeature.java
magicus: Ah, I see. Annoying that GSON cannot handle records. :( I'll have to revert it then. Good catch!

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: The API from ConfigManager was not particularly streamlines for its actual purpose. I made this return a `Stream` instead which gets rid of the possibility to accidentally modify the list. I also fixed some other API issues that I ran into while doing it. Overall, it looks much smoother now.

magicus:
I think returning `Stream` instead of `List` when we do not want the original list to be able to be modified might be a better design choice in many cases. I'll look into that as an alternative, but in a separate PR.

kristofbolyai:
> I think returning `Stream` instead of `List` when we do not want the original list to be able to be modified might be a better design choice in many cases. I'll look into that as an alternative, but in a separate PR.

Sounds good. But definitely remove the unmodifiable stuff for now.

magicus:
>  we will allocate a lot memory, won't we? 

I just realized I need to clear out a possible misconception. Calling `Collections.unmodifiableList` will not waste any memory. The new list is just a view, a wrapper backed by the underlying list, that has methods trying to modify the list throw unsupported operation. So there is no (or at least negligible) memory spent. What we get is a slight performance penalty (if this happens in hot spots, the JVM will likely optimize that away) since a get call needs to be delegated once.

But the upside you get from this is that you cannot by mistake modify an array that is still internal to another object.

If we really want to modify it, and not modify the list belonging to the object, then we need to make a copy. I have not discovered any instances that this is the case, though. But if it were, then we could definitely have a bug on our hands if we did not copy it.

kristofbolyai:
> > we will allocate a lot memory, won't we?
> 
> I just realized I need to clear out a possible misconception. Calling `Collections.unmodifiableList` will not waste any memory. The new list is just a view, a wrapper backed by the underlying list, that has methods trying to modify the list throw unsupported operation. So there is no (or at least negligible) memory spent. What we get is a slight performance penalty (if this happens in hot spots, the JVM will likely optimize that away) since a get call needs to be delegated once.
> 
> But the upside you get from this is that you cannot by mistake modify an array that is still internal to another object.
> 
> If we really want to modify it, and not modify the list belonging to the object, then we need to make a copy. I have not discovered any instances that this is the case, though. But if it were, then we could definitely have a bug on our hands if we did not copy it.

Okay. Performance is once concern, making the code look worse is another. I just wouldn't want to see us needing to add unmodifiable wrappers in half our classes. I think we should use streams where logical, and leave the rest to unmodifiable? (Or just use Kotlin :P ?)

magicus:
Ok, let's skip the unmodifiable part. I agree that it makes the code look worse. Are you happy with the rest of the PR?

kristofbolyai:
> Ok, let's skip the unmodifiable part. I agree that it makes the code look worse. Are you happy with the rest of the PR?

Yes, I didn't find anything else

magicus:
I take that as an approval :)

PR #1141, opened 2023-02-09
feat: Rework gear viewer

P0keDev:
This PR changes up the gear viewer backend to improve some functionality - the GUI is now created as an `AbstractContainer` instead of just a `Screen`, so the items rendered in it are treated as real items by the game. This means our tooltip fitting/scaling features work, users can screenshot items from the gear viewer, etc.

To fully accomplish this I did have to create an exception in the `ItemHandler` annotation code to ignore `FakeItemStack` objects - by definition these items will already have annotations and shouldn't be processed, but I'm willing to change this workaround if it's too "messy."

I also fixed an issue in `GearParser` that caused null powders to be parsed, which would break the item screenshot feature on all items (not just gear viewer ones).


common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -73,6 +74,9 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
     }
 
     private void updateAnnotation(ItemStack existingItem, ItemStack newItem) {
+        // These items will already have an annotation, no handling necessary
```
magicus, line 77: I don't think this is needed. Fake item stacks will get an annotation "at birth", and just a few lines below this case will be taken care of.

common/src/main/java/com/wynntils/models/gear/parsing/GearParser.java
```diff
@@ -79,7 +79,7 @@ public static GearParseResult parseItemStack(ItemStack itemStack, GearInfo gearI
 
                 codedPowders.chars().forEach(ch -> {
                     Powder powder = Powder.getFromSymbol(Character.toString(ch));
-                    powders.add(powder);
+                    if (powder != null) powders.add(powder);
```
magicus: Please use braces after if statements, unless it is a trivial return.

common/src/main/java/com/wynntils/screens/base/WynntilsContainerScreen.java
```diff
@@ -0,0 +1,54 @@
...
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+
+public abstract class WynntilsContainerScreen<T extends AbstractContainerMenu> extends AbstractContainerScreen<T> {
```
magicus, line 16: This duplication of functionality is a bit annoying. :( But I guess there is not much to do about it, as long as we are stuck with the class hierarchy of Minecraft, and a Java language that do no do multiple inheritance... :-/

magicus (dismissed):
Overall it looks very nice.

common/src/main/java/com/wynntils/screens/base/WynntilsContainerScreen.java
magicus: If anything, maybe you can make the failure handler in WynntilsScreen public static, and call that one, instead of duplicating it.

magicus: Nah, on second thought, it's not worth it either. Just let this be.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
P0keDev: If this return is removed, the annotations get overwritten with WynnItem ones: 
image

magicus: Hm, that looks bad. It is classified as a general, non-specific WynnItem, not a GearItem. Maybe the Fake Item Stacks are improperly setup? I'll check...

magicus: No, they should have proper annotations. That's weird. 

But let's do it like this. You check in this PR as-is, and then I can go and debug the issue in my own time, to figure out what is going wrong, since this should not have to be here.

magicus (approved):

PR #1146, opened 2023-02-09
feat: Auto update profession level and parse level/percentage from nodes/crafting

kristofbolyai:


common/src/main/java/com/wynntils/models/labels/LabelModel.java
```diff
@@ -0,0 +1,40 @@
...
+import com.wynntils.utils.mc.McUtils;
+import java.util.List;
+import java.util.Optional;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.syncher.SynchedEntityData;
+import net.minecraft.world.entity.Entity;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class LabelModel extends Model {
```
magicus: Label model? I think models should model some kind of concept present in the game. So "ProfessionModel" sounds perfectly fine. But here I think you are more trying to generalize a way to handle Minecraft data. I think you should be able to do the onEntitySetData where you need it. And if you really want to have this somewhere independent, it is more a matter of a "LabelHandler". But that would also kind of imply that there are (or are planned to soon be) more than one usage of the LabelEvent.

kristofbolyai: There will be more use case for it, area dps iirc.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -323,22 +312,7 @@ private void parseCharacterFromCharacterMenu(ItemStack characterInfoItem, ItemSt
         }
         ClassType classType = ClassType.fromName(className);
 
-        Map<ProfessionType, Integer> levels = new HashMap<>();
```
magicus, line 326: Is the idea to restore this functionality in the ProfessionModel?

common/src/main/java/com/wynntils/models/labels/LabelModel.java
magicus: Ok, if so, I think it will make more sense being a LabelHandler.

magicus:
Nice to see the profession stuff getting some love and care!

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: It's been already done

magicus: Nice, I must have just missed it. I'll let you continue working on the draft now, and stop pestering you until you are ready for review. :)

kristofbolyai: > Nice, I must have just missed it. I'll let you continue working on the draft now, and stop pestering you until you are ready for review. :)

Now is the time :)

common/src/main/java/com/wynntils/handlers/labels/LabelHandler.java
```diff
@@ -0,0 +1,36 @@
...
+        Entity entity = McUtils.mc().level.getEntity(event.getId());
+
+        if (entity == null) return;
+
+        for (SynchedEntityData.DataValue<?> packedItem : event.getPackedItems()) {
+            if (Entity.DATA_CUSTOM_NAME.getId() == packedItem.id()) {
+                Optional<Component> value = (Optional<Component>) packedItem.value();
+                if (value.isEmpty()) continue;
+
+                WynntilsMod.postEvent(new LabelEvent(value.get()));
```
magicus: So this is triggered when an entity gets a name, after spawning, or if it is renamed?

Just to check, the way we're e.g. detecting totems, by looking for armor stands and checking their name tag, it's not going to work in this case? Or would it, but you think this method is superior? 

I just find this quite opaque. Maybe because I don't fully understand this SynchedEntityData or PackedItems.

common/src/main/java/com/wynntils/handlers/labels/event/LabelEvent.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.handlers.labels.event;
+
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.Event;
+
+public class LabelEvent extends Event {
```
magicus: This is a bit non-descriptive. Maybe something like `EntityLabelChangedEvent`?

common/src/main/java/com/wynntils/models/profession/objects/ProfessionProgress.java
```diff
@@ -0,0 +1,9 @@
+/*
```
magicus: For other models, I used the package name `type` for this kind of simple record/enum type classes.

common/src/main/java/com/wynntils/handlers/labels/LabelHandler.java
kristofbolyai: We want to know when this happens, not that they exist. So event is better

magicus (approved):

PR #1147, opened 2023-02-09
fix: default info box positioning overlaps with minimap coordinates

ryanzhoudev:


magicus (approved):
Looks good to me but let @kristofbolyai have a say before integrating.

kristofbolyai:
Looks good

PR #1148, opened 2023-02-10
feat: mob totem tracking

ryanzhoudev:
image

Looks like this, but there are ( ) around the player's name. I can't give an updated real screenshot since I am out of totems for the day, but the preview button in-game is a good example.


common/src/main/java/com/wynntils/utils/mc/PacketUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import net.minecraft.network.syncher.SynchedEntityData;
+
+public final class PacketUtils {
+
+    public static String getNameFromMetadata(List<SynchedEntityData.DataValue<?>> data) {
+        // The rename stuff we're looking for is eventually something like
+        // Optional[literal{§c§l3:46}[style={}]]
+        // or
+        // Optional[literal{§c26s}[style={}]]
+        for (SynchedEntityData.DataValue<?> packedItem : data) {
```
magicus: This looks very much like the code that is done by the LabelHandler in @kristofbolyai's PR:
 
https://github.com/Wynntils/Artemis/pull/1146/files/8a20fe33f8e351421d4133432508810a4637e9be#diff-1d0f8b11a9ce3165c2c87c5c2bad669a7c51ef955e4d5a6e647eb027f6f8baae

I think you are both trying to solve the same problem, so we should find a unified way of doing this. Maybe if bolyai merges his PR first, and then you can see if you can use the LabelHandler as it is, or if it needs adapting to your needs?

common/src/main/java/com/wynntils/features/user/overlays/MobTotemTrackingFeature.java
```diff
@@ -0,0 +1,127 @@
...
+                            this.getWidth(),
+                            this.getHeight(),
+                            this.getRenderHorizontalAlignment(),
+                            this.getRenderVerticalAlignment());
+        }
+
+        void updateRenderTaskCache() {
+            renderTaskCache = Models.MobTotem.getMobTotems().stream()
+                    .map(mobTotem -> new TextRenderTask(
+                            "Mob Totem (" + mobTotem.getOwner() + ") at " + mobTotem.getLocation() + " ("
```
magicus: Just as with the shaman totems, I think the coordinates is not very helpful. I know I introduced them myself with the original mob totem implementation in Legacy :-) but I have been regretting it ever since. I suggest changing this to a distance instead, that is at least possible for the human brain to parse during a grinding session. (Or make it configurable, but I honestly don't know how the coordinates ever is going to be helpful).

common/src/main/java/com/wynntils/utils/mc/PacketUtils.java
magicus: I got a bit carried away and created https://github.com/Wynntils/Artemis/pull/1150.

If we merge that first, you should be able to use the extended `EntityLabelChangedEvent` here as well.

common/src/main/java/com/wynntils/features/user/overlays/MobTotemTrackingFeature.java
```diff
@@ -105,10 +108,35 @@ public void renderPreview(
 
         void updateRenderTaskCache() {
             renderTaskCache = Models.MobTotem.getMobTotems().stream()
-                    .map(mobTotem -> new TextRenderTask(
-                            "Mob Totem (" + mobTotem.getOwner() + ") at " + mobTotem.getLocation() + " ("
-                                    + mobTotem.getTimerString() + ")",
-                            textRenderSetting))
+                    .map(mobTotem -> {
```
magicus: Oooh! That is brilliant! Maybe you can extract this into a more general place, so it can be used for shaman totems as well? (It is okay to do that as a follow up)

ryanzhoudev:
The arrow rendering thing where it pointed in the direction of the totem is being removed because a) It appears the one arrow that was added after the others isn't actually in windows 10, and shows up for me only on my windows 11 machine. b) When it did show up, the arrow was inconsistent and aligned to the bottom, instead of the middle

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -666,8 +670,6 @@
   "feature.wynntils.shamanTotemTracking.secondTotemColor.name": "Shaman Totem 2 Color",
   "feature.wynntils.shamanTotemTracking.thirdTotemColor.description": "What color should the totem 3 highlight and tracking text be?",
   "feature.wynntils.shamanTotemTracking.thirdTotemColor.name": "Shaman Totem 3 Color",
-  "feature.wynntils.shamanTotemTracking.trackTotem.description": "Should Shaman/Skyseer totems be tracked?",
```
magicus, line 678: Should this really be removed?

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.core.Position;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.decoration.ArmorStand;
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class MobTotemModel extends Model {
+    private static final Pattern MOB_TOTEM_NAME = Pattern.compile("^§f§l(.*)'s§6§l Mob Totem$");
```
magicus, line 25: I need to check how they render the genitive form for player names ending in `s`. I'm afraid this won't match. I'll check (I'm `mag_icus` so I've fixed a lot of those issues in the past :)) and get back to you.

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
```diff
@@ -0,0 +1,130 @@
...
+        if (nameMatcher.find()) {
+            int mobTotemId = e.getEntity().getId();
+
+            if (mobTotems.containsKey(mobTotemId)) return; // If the totem is already in the list, don't add it again
+
+            mobTotems.put(mobTotemId, new MobTotem(new Location(as), nameMatcher.group(1)));
+            return;
+        }
+
+        if (mobTotems.isEmpty()) return;
```
magicus: That's not really needed since the for loop will be a no-op. If you think it adds clarity you can keep it, but its not needed.

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
```diff
@@ -0,0 +1,130 @@
...
+
+        for (MobTotem mobTotem : mobTotems.values()) {
+            // Exact equality is fine here because the totem is stationary
+            if (as.getX() == mobTotem.getLocation().x()
+                    && as.getY() == (mobTotem.getLocation().y() + TOTEM_COORDINATE_DIFFERENCE)
+                    && as.getZ() == mobTotem.getLocation().z()) {
+                Matcher timerMatcher = MOB_TOTEM_TIMER.matcher(e.getName());
```
magicus: I think you should turn this inside out. First check the timerMatcher. If it fails to match, then we don't need to do anything anyway. If it does match, then we look through the mobtotems. 

If you feel comfortable with it, or want to experiment with it, then I think that if might be rewritten as a stream filter expression rather nicely, instead of a for loop.

magicus:
Did you give up on the arrow rendering? :-( Too bad, it seemed like a cool idea. Maybe we can restore it later, perhaps as a more general functionality?

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
```diff
@@ -0,0 +1,130 @@
...
+            this.timerString = timerString;
+        }
+
+        public int getDistanceToPlayer() {
+            // y() - 4 because the entity is 4 blocks above the ground
+            return (int)
+                    Math.sqrt(McUtils.player().distanceToSqr(new Vec3(location.x(), location.y() - 4, location.z())));
+        }
+
+        public double getLookAngleDiff() {
```
magicus, line 109: I'm guessing this is not needed right now. It's up to you if you want to keep it anyway.

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
magicus: Confirmed that this works fine!

magicus:
> Did you give up on the arrow rendering? :-( Too bad, it seemed like a cool idea. Maybe we can restore it later, perhaps as a more general functionality?

I now see you wrote an explanation. I missed that before, sorry.

common/src/main/resources/assets/wynntils/lang/en_us.json
ryanzhoudev: Yeah, it was originally intended for the enable/disable config, but since Artemis just does this by disabling/enabling the feature, we don't need a specific config

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
ryanzhoudev: Would like to keep it for in the future

magicus (approved):
Looks great!

Issue #1108, opened 2023-02-01
ItemRenamedEvent is not calculated properly for crafted items


kristofbolyai:
This is likely caused by crafted item lore containing durability, and if spell cast makes the durability change, then the items won't be "similar", and item rename won't be detected. (I have not confirmed this. But this is a likely theory)

kristofbolyai:
https://github.com/Wynntils/Artemis/issues/1108

kristofbolyai:
Another item transformer related issue: ![image](https://user-images.githubusercontent.com/49001742/216811060-425d8a5d-7332-4f4e-90f7-2ee5d1543a89.jpeg)

magicus:
Fixed by  #1149

PR #1149, opened 2023-02-10
fix: Crafted items is now handled better

magicus:
Note that the title says "better", not "perfect". I started doing a more encompassing fix, but we are lacking the overall infrastructure for that. I decided not to back out the partial fixes, but to keep them in place, awaiting the proper handling of crafted items.

This will solve #1108.

Note that there are edge cases I have not been able to test. The isWeapon test for crafted gear is conservatively calling all crafted items for weapons. This might cause breakage, even I can't really say how. Also, if a crafted weapon changes it lore (due to durability) at the same time as a spell is cast, we might not be able to handle that (a new name + a new lore is interpreted as a completely new item, not a spell cast rename + a crafted item update). I have not been able to provoke this into happening so I don't know if Wynncraft does such a thing in a single update, or if the rename is sent separately from the durablity update.


magicus:
> Also, if a crafted weapon changes it lore (due to durability) at the same time as a spell is cast, we might not be able to handle that (a new name + a new lore is interpreted as a completely new item, not a spell cast rename + a crafted item update). I have not been able to provoke this into happening so I don't know if Wynncraft does such a thing in a single update, or if the rename is sent separately from the durablity update.

I figured out a way to test this using a 1 durability weapon. So I can confirm this works. Wynncraft sends first the spell rename, and when it is renamed "back", the durability changes. This is handled perfectly by this code.

kristofbolyai (approved):
Looks good, thanks

PR #1150, opened 2023-02-10
feat: Extend EntityLabelChangedEvent

magicus:


common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -176,21 +176,22 @@ public void onTimerSpawn(AddEntityEvent e) {
...
-    public void onTotemRename(SetEntityDataEvent e) {
+    public void onTotemRename(EntityLabelChangedEvent e) {
         if (!Models.WorldState.onWorld()) return;
 
-        int entityId = e.getId();
-        Entity entity = getBufferedEntity(entityId);
+        Entity entity = e.getEntity();
         if (!(entity instanceof ArmorStand)) return;
 
-        String name = getNameFromMetadata(e.getPackedItems());
```
kristofbolyai, line 186: Is `getNameFromMetadata` used anywhere else? Can we remove that too?

kristofbolyai (dismissed):
Otherwise looks good

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
ryanzhoudev: It will be used for mob totem tracking, but at the moment, this should be it (just checking from e.getPackedItems())

PR #1154, opened 2023-02-11
fix: Crafted and signed items aren't parsed correctly

ryanzhoudev:
Previous regex could be broken if the player signed and item. Example: (Crafted Bow) replaced with (Crafted by v8j).

Moved type parsing to other regex, which means it also works for non-crafted weapons now.


magicus:
The fixed regexp is good, thank you for that. However, I'm not sure about the gear type changes. I'd like to see a proper gear type analyzer for crafted gear, but that needs to handle armor as well. I think the upshot of that is that we should look at the itemstack instead. 

We're not actually using the gear type as from the parsing as it is right now anyway, so unless you plan to use it in the near future, I don't think its worth it.

kristofbolyai:
What was the original issue?

magicus:
Crafted gear did not match the regex pattern. I assumed it was like "Crafted Wand", but it could be "Crafted by foo_123_bar", and then the A-Za-z regexp did not match. This is a real issue, which is fixed by this.

ryanzhoudev:
probably should leave gear type parsing to other pr then, this is just for the fix

magicus (approved):

PR #1155, opened 2023-02-11
fix: Improve gear box guesses

magicus:


common/src/main/java/com/wynntils/models/items/annotators/game/GearBoxAnnotator.java
```diff
@@ -45,6 +45,7 @@ private static RangedValue getLevelRange(ItemStack itemStack) {
         if (!matcher.matches()) return null;
         int low = Integer.parseInt(matcher.group(1));
         int high = Integer.parseInt(matcher.group(2));
-        return RangedValue.of(low, high);
+        // Wynncraft "lies" to us, it says like "range 8-12" but in reality this means "9-12".
+        return RangedValue.of(low + 1, high);
```
kristofbolyai, line 50: Does this work for `1-4` level guesses?

magicus: According to the old data, they are presented as "0-4". (This in itself should have been a clue...) So yes, this will work. Or, if not, then the old code could never have worked for level <= 4, and I believe that would have been detected by now.

magicus: But it was a good catch! It could have been a bug.

kristofbolyai (dismissed):

PR #1156, opened 2023-02-11
fix: Changelog no longer forces Character Selection by default

Rafii2198:
I see many people on Discord being confused why that happens, and they don't know that it can be turned off. I believe that turning it off by default would be beneficial. Especially that it completely negates Wynncraft's `autojoin` toggle after updating Wynntils.


magicus:
I am not sure this is a good idea. I am not fond of the idea of pushing a changelog in the face of folks when they log into a world. You might find yourself immediately attacked by a mob, or maybe you were standing in lava when you logged out, and now you have a changelog you need to get rid off.

In Legacy, we displayed the changelog instead of connecting to Wynncraft when pressing the W button. Maybe that was better?

Or maybe we should just give a link on login: "Click here to read the changelog"?

Rafii2198:
I think some other way would be better than current, for me it breaks auto join and people on Discord ask why is it a thing so I think there is something in it.

The additional screen works for Legacy because W button had more functionality there. Here it just connects to Wynncraft so it would add some unnecessary *blockage* when someone knows changes.

I these two ideas should be fine for everyone and could be enabled by default: 
- When you update Wynntils, a clickable chat message will appear saying "You have updated Wynntills to XXX version. Show Changes" or similar message.
- In the book, the version of Wynntils  becomes clickable and it would show the changelog. Additionally, if you haven't checked it out after the update some symbol (probably exclamation point in a triangle) would appear next to it to indicate new changes.

The current system on the other hand could be merged into one thing and be disabled by default i.e it would show changelog automatically when joining server and would put you in Character Selection in case of autojoin for safety.

P0keDev (approved):

P0keDev:
I'm in favor of this one, personally. Having the /class menu opened is pretty jarring, and it's potentially a lot more catastrophic than opening the changelog GUI mid-gameplay if the player manages to input something like the drop key in the split second they're on a class before getting booted back to the selection menu

magicus:
Ok, let's do this for now then. I still don't really like this, though. We need to figure out a better way to do this. Maybe only show a small marker on the screen "Wynntils updated, read changelog in menu" and add a new button in the Wynntils menu?

Rafii2198:
As a suggestion, taking damage should instantly close the changelog then there shouldn't be much of a problem then

PR #1157, opened 2023-02-12
feat: Add "how to obtain item" info

magicus:
This will show up in guides. And it will make the gear box possible item prediction much better!


magicus:
There is a corresponding update of the API site urls.json in https://github.com/Wynntils/WynntilsWebsite-API/pull/43

magicus:
Screenshot 2023-02-12 at 16 36 02

magicus:
The Wynndata "htoi" data included a wide range of additional information, with locations, screenshots of mobs, etc. I opted not to include it here, at least not right now. This provides the basic info. Additional data needs some thinking on to present properly.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -95,38 +101,51 @@ private void loadRegistry() {
...
-                    }
-                }
+                itemObtainMap = obtainGson.fromJson(obtainReader, obtainType);
 
-                // Make the result visisble to the world
-                gearInfoRegistry = registry;
-                gearInfoLookup = lookupMap;
-                gearInfoLookupApiName = altLookupMap;
+                // Now we can do the gear DB
+                downloadGearRegistry();
```
kristofbolyai, line 114: What if the one calling this fails

magicus: Then we can't build up the gear database properly in any case. :( You could possibly argue that it would be better to have a partially correct gear DB in that case. But then you have to start drawing lines. Is it OK to not have major IDs? How much extra complexity do we accept to have code that would handle an only partially complete gear DB? Etc.

kristofbolyai (approved):

PR #1158, opened 2023-02-13
feat: Chat mentions

Zatzou:
This pr adds chat mentions similar to the ones 1.12 wynntils.


common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,116 @@
...
+                "(?<!\\[)\\b(" + McUtils.mc().getUser().getName()
+                        + (aliases.length() > 0 ? "|" + aliases.replace(",", "|") : "") + ")\\b(?!:|])",
+                Pattern.CASE_INSENSITIVE);
+    }
+
+    @SubscribeEvent()
+    public void onChat(ChatMessageReceivedEvent e) {
+        Component message = e.getMessage();
+
+        Matcher looseMatcher = pattern.matcher(ComponentUtils.getUnformatted(message));
```
magicus: This all looks very complicated. Can't you just convert it to a coded string and look for a regexp in that? That's `ComponentUtils.getCoded()` instead. And then you do not need to match twice.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,116 @@
...
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.sounds.SoundEvents;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class ChatMentionFeature extends UserFeature {
+    @Config
+    public String aliases = "";
+
+    private Pattern pattern = Pattern.compile(
+            "(?<!\\[)\\b(" + McUtils.mc().getUser().getName()
```
magicus: I don't understand this regex. It looks like it is starting a named capture group with `(?<` but then it just goes away in a completely different direction. Can you give an example of the kind of text it is supposed to capture, or better yet, add a test suite at regexr.com, which is what we tend to do with new regexes. See https://regexr.com/76ijo for an example.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: The first match is to check if the message should be processed at all. Then the components are modified in recurseMark in a way which should preserve the raw components as well as possible in order to not edit what doesn't need to be edited.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: The first group is an negative lookbehind. It and the negative lookahead at the end are used to make sure the regex doesn't match every message sent by the user. Here is a regexr with tests [https://regexr.com/788em](https://regexr.com/788em).

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus: Ah, I missed the fact that you actually *rewrite* the chat message, and not only look for a pattern to play a sound. Ok, then I realize why you do this. 

Some suggestions:
1) These are two different behaviors -- pling on your name mentioned, and rewriting the chat to color your name. Please add config options for these two. They can be true by default, but users should be able to opt-out of that which they might not want.

2) The "recurseMark" is not a very descriptive name. Please name the method after what it is trying to achieve, not after how it is implemented. In this case, perhaps "rewriteComponentWithHighlight", or something like that. (Not the best of names, if you can get the message across with a more succinct name, feel free to use that instead).

3) Add some comment at the place where you actually rewrite the message, to highlight the fact that the message is changed at that point. All the rest is just boilerplate for processing the Component tree.

4) Maybe the user should be able to select what color code the highlight use?

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus: Oh, I see. I tend to stay away from negative lookbehinds. You can easily get into highly complicated regexps which are computationally extremely expensive. At one extreme, you can [find prime numbers](https://www.noulakaz.net/2007/03/18/a-regular-expression-to-check-for-prime-numbers/) with regexp. But that will of course have a killer big-Oh.

It might be that this case is benign, but can you please check if there is some what to solve this without using this construct?

Zatzou: It could certainly be done so that the checking of whenever its an messsage sent by the player themselves is done in code rather than with regex. However that would involve a lot more code for detecting those cases.

Overall I kind of doubt this would be that exploitable. From what I found online I at least couldn't find anything that would break this. I think realistically the lookbehind won't be too much since its only matching a single character which means it can't cause that much checking.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: Alright I've implemented all of these but the fourth. I'm not sure how changing the color could be done nicely as the highlight has to be styled with §e rather than using propper styling to not break guild chat.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,129 @@
...
+
+            // NOTE: Message is modified here!
+            while (match.find()) {
+                // Mention found -> split component and mark the
+                String before = text.substring(nextStart, match.start());
+                String name = text.substring(match.start(), match.end());
+
+                comps.add(Component.literal(before).withStyle(comp.getStyle()));
+                // styling is not used as it breaks guild chat because wynn in their infinite wisdom decided to make
+                // guild chat not use the propper styling
```
magicus:
```suggestion
                // guild chat not use the proper styling
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,129 @@
...
+
+                nextStart = match.end();
+            }
+
+            MutableComponent afterComp =
+                    Component.literal(text.substring(nextStart)).withStyle(comp.getStyle());
+
+            // process any siblings
+            List<Component> sib = comp.getSiblings();
+            if (sib.size() != 0) {
```
magicus: the if check is unecessary, the for loop will not do anything if there are no siblings.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,129 @@
...
+                e.setMessage(rewriteComponentWithHighlight(message));
+            }
+            if (dingMention) {
+                McUtils.playSound(SoundEvents.NOTE_BLOCK_PLING.value());
+            }
+        }
+    }
+
+    private Component rewriteComponentWithHighlight(Component comp) {
+        MutableComponent curr = MutableComponent.create(comp.getContents()).withStyle(comp.getStyle());
```
magicus, line 72: This function is still unnecessarily hard to read. I wonder if you can extract the sibling descent into a common part? If not, you can at least more clearly split it into two halves, the "have no text" and "have text" parts.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,129 @@
...
+            List<Component> sib = comp.getSiblings();
+            if (sib.size() == 0) {
+                // no siblings -> this is an end component return self
+                return comp;
+            } else {
+                // some siblings -> check them
+                for (Component c : sib) {
+                    curr.append(rewriteComponentWithHighlight(c));
+                }
+            }
```
magicus: >  clearly split it into two halves, the "have no text" and "have text" parts.

To do that, move the `return curr` here, and drop the `else {`.

Also, I'm thinking maybe you could even drop the `if (sib.size() == 0)` check and just have a for loop and a `return curr`. If I read the code correctly, `curr` is at this point essentially the same as `comp`.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus, line 72: Also, why are you just not assigning `comp` to  `curr`?

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,129 @@
...
+    public String aliases = "";
+
+    private Pattern pattern = Pattern.compile(
+            "(?<!\\[)\\b(" + McUtils.mc().getUser().getName()
+                    + (aliases.length() > 0 ? "|" + aliases.replace(",", "|") : "") + ")\\b(?!:|])",
+            Pattern.CASE_INSENSITIVE);
+
+    @Override
+    protected void onConfigUpdate(ConfigHolder configHolder) {
+        pattern = Pattern.compile(
```
magicus: This complicated pattern is duplicated twice. That's not a good design.

Please extract the pattern creation to a method, and call it in the constructor and in this method. You can do it either as `updatePattern()` or `pattern = buildPattern()`; whichever you think is the clearest.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus:
```
@Config
ChatFormatting rewriteColorCode;
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: assigning comp to curr would mean that curr now has comps siblings while doing it this way copies the body of the component but doesn't copy its siblings.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: So overall what should be done with this case? I think having the checks on the regex shouldn't be that big of a concern. However its also not that hard to do the checks in code as they are done on legacy wynntils.

So which approach should be taken here?

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,124 @@
...
+    @Config
+    public boolean dingMention = true;
+
+    @Config
+    ChatFormatting rewriteColorCode = ChatFormatting.YELLOW;
+
+    @Config
+    public String aliases = "";
+
+    private Pattern pattern = buildPattern();
```
magicus: Please make the assignment in a constructor.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,124 @@
...
+                e.setMessage(rewriteComponentWithHighlight(message));
+            }
+            if (dingMention) {
+                McUtils.playSound(SoundEvents.NOTE_BLOCK_PLING.value());
+            }
+        }
+    }
+
+    private Component rewriteComponentWithHighlight(Component comp) {
+        // Create a new component with only the body and style of comp
```
magicus:
```suggestion
        // Make a copy of the component without the siblings
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,124 @@
...
+    ChatFormatting rewriteColorCode = ChatFormatting.YELLOW;
+
+    @Config
+    public String aliases = "";
+
+    private Pattern pattern = buildPattern();
+
+    @Override
+    protected void onConfigUpdate(ConfigHolder configHolder) {
+        // rebuild pattern incase it has changed
```
magicus:
```suggestion
        // rebuild pattern in case it has changed
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,124 @@
...
+
+        MutableComponent afterComp =
+                Component.literal(text.substring(nextStart)).withStyle(comp.getStyle());
+
+        // process any siblings
+        for (Component c : comp.getSiblings()) {
+            afterComp.append(rewriteComponentWithHighlight(c));
+        }
+
+        // throw all of it together
```
magicus: This also seem very convoluted. Why reversing and stuff?

I think the basic recursive descent here is:
1) check if current node has text that needs to be updated, and
2) process all siblings recursively.

The current text needs to be updated iff a) it has a non-empty text part, and b) it matches the username regexp.

If this is true, you should be able to replace the Component in question with a new Component containing basically Component[text: "", siblings: [prior-to-match,highlighted-match,after-match]]. (And more of course a slightly more complex if you want to keep supporting multiple mentions in the same Component, with after-first-match-and-before-second-match, etc).

And then you can just append all the siblings, recursively processed.

magicus: And to make the functionality clearer, you can extract the "Turn this Component, ignoring the siblings, into a new Component, which has the name highlighted" into a separate function.

Zatzou: I've cleaned up the rewriting function a lot now but I'm not sure of how extracting the rewriting should be done if it needs to be done still. I tried to extract it but either I'm failing to understand how java passes references or I just did something wrong but anyways the code is a lot cleaner now.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,111 @@
...
+        // because wynn still uses legacy coloring for it.
+        String text = curr.getString();
+
+        // if current component has text check it for mentions
+        if (text != "") {
+            Matcher match = pattern.matcher(text);
+
+            // if we match then rewrite the component
+            if (match.find()) {
+                // get the start of the string before any mention and set curr to be it
```
magicus: Ok, the code is starting to look comprehensive now. The `rewriteComponentTree` method still needs some work though. It is a complex function; that is okay, it is doing a complex thing, but it is also not designed to handle that complexity. 

More to the point: This part, when you have found a match and processes the component, can -- and should! -- be extracted into a separate method.

A good rule of thumb is a method should not be longer than it fits on a single screen. It is not always possible to follow, but here it definitely is. If you extract the "replace text in this component" into a function, you get one screen with the overview of rewriteComponentTree, on a high level, traversing the Component tree and replacing the text when a match is found, and then you can have a `rewriteComponentNode` (or whatever) that focuses on the one single objective of replacing the text in that particular node.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,111 @@
...
+                String before = text.substring(0, match.start());
+                curr = Component.literal(before).withStyle(comp.getStyle());
+
+                int nextStart = 0;
```
magicus: But, also, after you have moved this code into a separate method, you also need to simplify it further. The looping logic with a mutable nextStart int, a do while loop (which are much harder to read than while loops), and the loop starting with a check for nextStart which can never be != 0 when you start, etc. It's just a mess.

Think like this: It is not enough for the code to **have no obvious faults**. The code should be written so it **obviously have no faults**. Unfortunately, this code falls in the former category, rather than the latter. It looks like it might work, but it is not immediately clear that it does.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,111 @@
...
+    @Config
+    public boolean dingMention = true;
+
+    @Config
+    ChatFormatting rewriteColorCode = ChatFormatting.YELLOW;
+
+    @Config
+    public String aliases = "";
+
+    private Pattern pattern;
```
magicus: You seem to have forgotten the constructor.

Zatzou: My java knowledge is a little lacking so what exactly do you mean by this?

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: My patience is starting to run dry trying to simplify this function more and more. In my opinion its already clear enough what it does. It has a decent name, there are comments explaining what it does, it only has a few major code paths that can occur.

I'll see if I have the patience to try clean it up further tomorrow but at this point I'm heavily considering just abandoning this in hopes that someone else might come along and finish the job.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus: I asked you to move the initialization to the constructor from the field declaration. You removed it from the field, but you never added it to the constructor. You need to add a no-args constructor and add a `pattern = buildPattern` to it.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: Alright I've done my best to clean up things as much as possible and at this point I'm all out of ideas of how to clean things up further. If someone else wants to take a crack at cleaning things up more they can go ahead but I'm done with this code at least for now.

magicus (dismissed):
I think this looks fine now. Thanks for hanging in there, it's been a process!

@kristofbolyai Can you have a look as well?

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private Pattern buildPattern() {
+        return Pattern.compile(
+                "(?<!\\[)\\b(" + McUtils.mc().getUser().getName()
+                        + (aliases.length() > 0 ? "|" + aliases.replace(",", "|") : "") + ")\\b(?!:|])",
+                Pattern.CASE_INSENSITIVE);
+    }
+
+    @SubscribeEvent()
```
kristofbolyai:
```suggestion
    @SubscribeEvent
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private Component rewriteComponentTree(Component comp) {
+        // Make a copy of the component without the siblings
+        MutableComponent curr = MutableComponent.create(comp.getContents()).withStyle(comp.getStyle());
+        // .getString() is used here as it gives formattingchars when those exist. It is needed for guild messages
+        // because wynn still uses legacy coloring for it.
+        String text = curr.getString();
+
+        // if current component has text check it for mentions and rewrite the component if necessary
+        if (text != "") {
```
kristofbolyai:
```suggestion
        if (!text.isEmpty()) {
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+    @Config
+    public boolean dingMention = true;
+
+    @Config
+    ChatFormatting rewriteColorCode = ChatFormatting.YELLOW;
+
+    @Config
+    public String aliases = "";
+
+    private Pattern pattern;
```
kristofbolyai:
```suggestion
    private Pattern mentionPattern;
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+    private MutableComponent rewriteMentions(MutableComponent curr, String text, Style style) {
+        Matcher match = pattern.matcher(text);
+
+        // if we match then rewrite the component if there are no matches the function will just return
+        if (match.find()) {
+            // get the start of the string before any mention and set curr to be it
+            curr = Component.literal(text.substring(0, match.start())).withStyle(style);
+
+            // do the name of the first mention
+            curr.append(Component.literal(rewriteColorCode + match.group()));
```
kristofbolyai:
```suggestion
            curr.append(Component.literal(rewriteColorCode + match.group(0)));
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+            int lastEnd = match.end();
+
+            // Handle any additional mentions within the message
+            while (match.find()) {
+                // get the bit in between of matches
+                curr.append(Component.literal(text.substring(lastEnd, match.start())))
+                        .withStyle(style);
+
+                // get the name and recolor it
+                curr.append(Component.literal(rewriteColorCode + match.group()));
```
kristofbolyai:
```suggestion
                curr.append(Component.literal(rewriteColorCode + match.group(0)));
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -43,6 +43,15 @@
...
   "feature.wynntils.chatCoordinates.name": "Chat Coordinates",
   "feature.wynntils.chatItem.name": "Chat Item",
+  "feature.wynntils.chatMention.aliases.description": "Any aliases you wish to get mentioned for in addition to your username. Names should be seperated with an single comma.",
+  "feature.wynntils.chatMention.aliases.name": "Aliases",
+  "feature.wynntils.chatMention.dingMention.description": "Should there be a sound played when you get mentioned?",
+  "feature.wynntils.chatMention.dingMention.name": "Ding On Mention",
+  "feature.wynntils.chatMention.markMention.description": "Should mentions be highlighted in chat?",
+  "feature.wynntils.chatMention.markMention.name": "Highlight Mentions",
+  "feature.wynntils.chatMention.name": "Chat Mentions",
+  "feature.wynntils.chatMention.rewriteColorCode.description": "What color should be used for highlighting",
```
```suggestion
  "feature.wynntils.chatMention.rewriteColorCode.description": "What color should be used for highlighting?",
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
```diff
@@ -0,0 +1,123 @@
...
+
+public class ChatMentionFeature extends UserFeature {
+    @Config
+    public boolean markMention = true;
+
+    @Config
+    public boolean dingMention = true;
+
+    @Config
+    ChatFormatting rewriteColorCode = ChatFormatting.YELLOW;
```
kristofbolyai:
```suggestion
    public ChatFormatting mentionColor = ChatFormatting.YELLOW;
```

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
Zatzou: This should actually use .group() instead of .group(0) as we want the specific group which was matched that time and not the first time.

so if we for example have a message
hi name hows it going alias

where name is the playername and alias is an set alias then this code would change the message to
hi name hows it going name

kristofbolyai:
```
public String group() {
        return this.group(0);
    }
```

Zatzou: well in that case I must have misunderstood the documentation

kristofbolyai (approved):
Looks okay to me. We **need** to make the Component Converter though. This is unnecessarily complex.

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus: Oops! I did not see that 😊

magicus:
> Looks okay to me. We **need** to make the Component Converter though. This is unnecessarily complex.

Ha! Now I got you to agree. ;-)

magicus (approved):

common/src/main/java/com/wynntils/features/user/ChatMentionFeature.java
magicus: I am sorry you feel that way. Maybe we should have communicated better at the start that the PR needed some heavy rework to reach the quality levels needed for integration.

magicus:
This was broken. :-( I did not verify that it even started:

[11:45:59] [Render thread/ERROR] (wynntils) Config displayName i18n is missing for feature.wynntils.chatMention.mentionColor.name
[11:45:59] [Render thread/ERROR] (wynntils) Failed to initialize feature ChatMentionFeature

magicus:
@Zatzou When was the last time you verified that the code in the PR even worked?

Zatzou:
@magicus oh that issue with the translation string actually happened since bolyai suggested a change to one of the variable names and I forgot to rename it in the translations file.

Other than that I've tested it pretty much between every change and at least the last time I tested it everything seemed to work however I obviously didn't actually look into the settings menu to notice the missing translation string.

magicus:
The missing detection is checked at startup by the framework, which then disables the entire feature, so it didn't work at all. But I understand that you don't normally think you need to re-test after renaming a variable.

Zatzou:
seems a little odd that I didn't catch the issue in that case but yeah the missing translation strings were definitely my fault just quickly trying to rename the variables. I still tested all of the highlighting itself quite thoroughly making sure that things like hovers and links were preserved.

I actually ended up finding in testing that wynncraft itself handles some stuff like links quite weirdly but mostly everything in my testing seemed to work as well with the mentions feature on and off.

PR #1162, opened 2023-02-17
fix: Remove player nametag from gear viewer GUI

P0keDev:
This is a pretty small issue, but I found it bothersome and wanted to fix it; the vanilla nametag rendering function does weird z-level stuff which would make it overlap with tooltips in the gear viewer. The solution here is pretty messy, but it's the best I could do without adding a custom entity render method (I think).

image

### DIFF
--- a/common/src/main/java/com/wynntils/screens/gearviewer/GearViewerScreen.java
+++ b/common/src/main/java/com/wynntils/screens/gearviewer/GearViewerScreen.java
@@ -31,15 +31,35 @@ import net.minecraft.world.inventory.ClickType;
 import net.minecraft.world.inventory.Slot;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
+import net.minecraft.world.scores.PlayerTeam;
+import net.minecraft.world.scores.Scoreboard;
+import net.minecraft.world.scores.Team.Visibility;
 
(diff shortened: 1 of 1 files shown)


magicus:
I've seen this as well, and was annoyed by it, so I'm glad its getting fixed. However, I agree that this solution is not the cleanest... If you say you won't put any more time and effort into this problem, I'll accept you solution, but otherwise I'd like to see something better.

I'm thinking the viewer GUI should be rendered in such a way that it is drawn on top of everything else. Why is that not so? Or, rephrased somewhat, does the name tag appear like this for other GUIs as well, such as the inventory, or our Wynntils main menu? 

It could be that it does, but I haven't noticed it, and the gear viewer is special since the player stays in the same place relative to the GUI. If that is the case, then this is probably the only solution.

But if not, what makes those GUIs different compared to the Gear viewer? Are we initializing it wrong?

P0keDev:
Other GUIs don't show a nametag because they're all rendering your own player model, and there's separate logic to avoid rendering your own nametag.

magicus:
Ah, *now* I get it! It's the miniature avatar that gets the nametag; I always thought it was the real player's nametag "shining through" the GUI. Riiiight. Well, then this is okay, I think.

magicus (approved):

PR #1163, opened 2023-02-17
fix: Fix guild map time held order

kristofbolyai:


common/src/main/java/com/wynntils/utils/SimpleDateFormatter.java
```diff
@@ -9,17 +9,19 @@
 import java.text.ParsePosition;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.TimeZone;
 
 public class SimpleDateFormatter extends DateFormat {
-    private static final Map<Integer, String> MAPPINGS = Map.of(
```
magicus, line 16: Why is the old version broken? To me, it looks about the same as the new...?

magicus: Ah, I see, it's a linked hash map...

magicus: Can't you create a LinkedHashMap seeded from a Map? Like 
`new LinkedHashMap<>(Map.of( .... ))`?

I know Java map "constants/literals" suck, but if possible, I still would like to avoid adding elements in the constructor. That's almost the last resort..

magicus (approved):

common/src/main/java/com/wynntils/utils/SimpleDateFormatter.java
magicus: Thanks!

PR #1164, opened 2023-02-17
feat: Assign stable Character GUIDs to classes

kristofbolyai:


kristofbolyai:
Note: Do not merge this, only review it. We need to do an announcement, we break compatibility with old item lock configs.

common/src/main/java/com/wynntils/models/players/hades/HadesModel.java
```diff
@@ -239,8 +239,9 @@ public void onTick(TickEvent event) {
     public void tryResendWorldData() {
         if (!isConnected()) return;
 
-        hadesConnection.sendPacket(
-                new HCPacketUpdateWorld(Models.WorldState.getCurrentWorldName(), Models.Character.getId()));
+        hadesConnection.sendPacket(new HCPacketUpdateWorld(
```
magicus, line 242: Will this mess with Hades if not all users have the same version?

kristofbolyai: No, this pretty much does not get used anywhere. It was just an old concept, we never got to use it.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -80,7 +84,7 @@ public final class CharacterModel extends Model {
     // This field is basically the slot id of the class,
     // meaning that if a class changes slots, the ID will not be persistent.
     // This was implemented the same way by legacy.
-    private int id;
+    private String id;
```
magicus, line 87: Ideally, I'd like to see us correlate the value from the soul point tooltip with the full uuid from the API. But I guess it is "room for future improvements".

magicus (approved):

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: I am not fully sure if that is needed though.

magicus: Neither am I, but I'm fairly certain it's still a good idea. :-D

My point being, I can't figure out good reasons right now for having an UUID instead of this truncated string, but if you were to ask me to design this, if I did not know I was given just a partial UUID from the soul points, I would have chosen UUID as the representation.

And one reason to have the UUID is to match it with the data provided by Wynntils API. And if we have to match it with the UUID anyway, then we might just as well store it as UUID fully.

PR #1166, opened 2023-02-18
fix: Consuming chat tabs don't duplicate messages

kristofbolyai:


common/src/main/java/com/wynntils/features/user/ChatTabsFeature.java
```diff
@@ -51,6 +51,9 @@ public class ChatTabsFeature extends UserFeature {
     // We do this here, and not in Models.ChatTab to not introduce a feature-model dependency.
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public void onChatReceived(ChatMessageReceivedEvent event) {
+        // We are already sending this message to every matching tab, so we can cancel it.
+        event.setCanceled(true);
```
kristofbolyai, line 55: It is hard to notice in reviewing, but in line 63 we return; making it so we don't cancel the event, when in reality, we should.

magicus (approved):

PR #1168, opened 2023-02-18
chore: Debug improvements

magicus:


common/src/main/java/com/wynntils/core/net/NetManager.java
```diff
@@ -53,6 +53,11 @@ public Download download(URI uri, File file) {
     }
 
     public Download download(URI uri, File file, String expectedHash) {
+        // For debugging, always return cached files if requested
+        if (WynntilsMod.isDevelopmentEnvironment() && new File(CACHE_DIR, "keep").exists()) {
```
kristofbolyai: This is kinda weird though? Why not have a bool in NetManager which can be set true before compiling?

HighCrit: You might accidentally push that 😌

magicus: @HighCrit Exactly. I wanted to have a marker on the file system when you want to meddle with the cache files.

kristofbolyai (approved):

PR #1171, opened 2023-02-19
fix: Chat Mentions were DOA

magicus:


magicus:
I verified that the functionality works. I don't really trust it though, so let's start it as disabled by default.

kristofbolyai (dismissed):

PR #1174, opened 2023-02-19
fix: Party management scaling improvements

ryanzhoudev:
Looks like this now:

image

Also scales with gui scale setting somewhat well. Elements will overlap at very large gui scales (4 or auto), but this is much better than them going off the screen previously. Everything is still accessible when overlap happens. There also isn't really a good fix for the overlap, since stuff like the buttons have a minimum vertical height, and the text can only be so small. Heads are also static size.


common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -126,174 +150,145 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
 
         super.render(poseStack, mouseX, mouseY, partialTick);
 
+        // region debug dividers
+        // to make positioning easier, we will split the screen into GRID_DIVISIONS parts horizontally and vertically
+        // uncomment when changing gui elements
```
ryanzhoudev, line 154: This section contains the commented debug grid I created for positioning elements. I think leaving it here would be helpful for anyone that tries to move things around in the future, but if anyone thinks this is too messy, feel free to lmk and I'll remove it

kristofbolyai: We never left any code like this in main, I think that is how we should do it. Maybe we can introduce a `stash` folder in the repo, to have debug/whatever code like this. (I know we have DebugFeatures but those are vastly different)

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -126,174 +150,145 @@ public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTic
...
         // region Invite field header
         FontRenderer.getInstance()
                 .renderText(
                         poseStack,
-                        I18n.get("screens.wynntils.partyManagementGui.inviteFieldHeader1"),
-                        this.width / 2 - X_START,
-                        this.height / 2 - 206,
+                        // Yes this is kind of abusive of the formatting system, and I should probably do
+                        // another .renderText call, but this makes aligning these two texts significantly easier
+                        // (especially with changing gui scales and resolutions)
```
kristofbolyai, line 163: It is fine IMO, we render MC colored text in a lot of places.

kristofbolyai (commented):
The system looks good. 
Though.. To be honest, I've always wanted this to be a generic system, something we can use anywhere and is not bound to a specific screen. Obviously there is a lot of ways to clear this grid screen system up and add a lot of fancy stuff, if we abstract it away. What do you think? Would you be interested in working in something like this?

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
magicus: `stash` folder or `stash` branch? I don't think either is a super good idea.

I think it can be reasonable to save debug code, in place, like here. In fact, I'm not sure you even need to comment it out; if we have a method that is never called, but a comment saying "don't remove, used when debugging" I think it's probably better.

magicus:
A generic system for writing minecraft screens would be awesome, and probably a huuuuge amount of work. :) When the next screen that needs working like this comes up, let's see how we can reuse and generalize this code.

magicus:
I'm really missing this :) I think the one thing left is that @kristofbolyai and I agree what to do with the debug code. I propose, as I said, to restore it as real code (not commented out), but add a comment that it is OK that the function is not called so we don't remove it as dead code during cleanup.

ryanzhoudev:
> Would you be interested in working in something like this?

Yea sure. I think the waypoint creation screen could also move to this system since the buttons on the bottom disappear if the gui scale is 4.

kristofbolyai (approved):

PR #1175, opened 2023-02-19
feat: Function Refactor Part 1: Basic Expression Parsing (FunctionExpression)

kristofbolyai:
Part 1: Basic parsing support for one handed expressions (FunctionExpression at the moment)


common/src/main/java/com/wynntils/utils/type/ErrorOr.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.type;
+
+public final class ErrorOr<T> {
```
magicus, line 7: The class itself is nice, I'm just thinking a bit about the name. What would you think of "ConditionalValue" instead?

kristofbolyai: I think it read fine as `ErrorOr<UsefulType`. Maybe I wrote too much C++... I am fine with renaming, but `ConditionalValue<OtherLongType>` is getting long, isn't it?

magicus (approved):
I think this looks good. Not fully managing to wrap my head around it, but maybe that's just me being a bit sleepy right now. If you want to, you can integrate it now, or wait until I get another look tomorrow or later this week.

kristofbolyai:
> I think this looks good. Not fully managing to wrap my head around it, but maybe that's just me being a bit sleepy right now. If you want to, you can integrate it now, or wait until I get another look tomorrow or later this week.

This depends on how you want to review it. My "parts" are basically growing, so part 2 = part 1 + part 2 changes, targeting the full refactor branch. If we merge this, we just make part 2 look reviewable.

common/src/main/java/com/wynntils/utils/type/ErrorOr.java
magicus: Ok, fine. Keep ErrorOr.

common/src/main/java/com/wynntils/core/functions/expressions/parser/ExpressionParser.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.function.Function;
+
+public final class ExpressionParser {
+    private static final List<Function<String, Optional<Expression>>> registeredExpressions =
+            List.of(FunctionExpression::tryParse);
+
+    private ExpressionParser() {}
+
+    public static ErrorOr<Expression> tryParse(String rawExpression) {
+        for (Function<String, Optional<Expression>> expression : registeredExpressions) {
```
magicus, line 21: Feels like a `stream().filter()`..?

magicus (approved):
Looks fine.

common/src/main/java/com/wynntils/core/functions/expressions/parser/ExpressionParser.java
kristofbolyai: This is bit complex to convert to a stream, and readability would not get better, maybe it would be even worse.

PR #1177, opened 2023-02-19
feat: Function Refactor Part 0: Functions are just facades for model logic

kristofbolyai:
Removes `ActiveFunction`. There should be no case where a function needs to be disabled, enabled or have any kind of event logic. All of the above should be kept in the respective model. I kept crash protection, if for whatever reason we still crash (likely a formatting error).


kristofbolyai:
This should be merged before the whole refactor, to have this structure going on main.

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
```diff
@@ -38,10 +40,19 @@ public class CombatXpModel extends Model {
 
     private boolean firstJoinHappened = false;
 
+    private final TimedSet<Float> lastMinuteXpGainRawSet = new TimedSet<>(1, TimeUnit.MINUTES, true);
```
magicus: I understand what you mean, but "last minute" sounds like it is almost too late, [cf](https://www.dictionary.com/browse/at-the-last-minute).

Maybe "latestMinute"? Maybe we need assistance from a native English speaker. @P0keDev ?

magicus (dismissed):
Change itself look good but we need better name for "last minute".

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
kristofbolyai: @magicus I renamed it. Do you accept it so I can merge this (and come back to it if needed)?

magicus (approved):

PR #1179, opened 2023-02-19
feat: Function Refactor Part 2: Function Arguments

kristofbolyai:
This PR focuses on adding arguments to functions, along with some smaller systematic changes.


kristofbolyai:
Arguments are now treated as expressions. This means, adding generic functions, like `IF` is fairly easy. It also means, we can nest functions, which can be great for advanced users (when supporting functions are added).

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -101,44 +103,61 @@ public List<String> getAliases() {
     }
 
     public static class ClassFunction extends Function<String> {
-        // FIXME: original had upper/lower case versions. Make a upper/lower function instead.
         @Override
-        public String getValue(String argument) {
-            return Models.Character.getActualName();
+        public String getValue(FunctionArguments arguments) {
+            Boolean showReskinnedName =
+                    (Boolean) arguments.getArgument("showReskinnedName").getValue();
```
magicus: I think we can take a page out of GSON's book here. If you add a `asBoolean()` to the argument class, we can get rid of the casting when using it. (It needs to check for type and throw an exception in case of type mismatch).

And probably you'll need `asString`, `asInt`, and `asDouble` as well...

kristofbolyai: Sounds good, yes.

magicus (approved):
Part 2 looks good as well. I have reviewed this via https://github.com/Wynntils/Artemis/pull/1180 to see the diffs between part 2 and part 1.

PR #1180, opened 2023-02-19
DUMMY: Just for comparing functions part 2

magicus:


common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -101,44 +103,61 @@ public List<String> getAliases() {
     }
 
     public static class ClassFunction extends Function<String> {
-        // FIXME: original had upper/lower case versions. Make a upper/lower function instead.
         @Override
-        public String getValue(String argument) {
-            return Models.Character.getActualName();
+        public String getValue(FunctionArguments arguments) {
+            Boolean showReskinnedName =
+                    (Boolean) arguments.getArgument("showReskinnedName").getValue();
```
magicus, line 109: I think we can take a page out of GSON's book here. If you add a `asBoolean()` to the argument class, we can get rid of the casting when using it. (It needs to check for type and throw an exception in case of type mismatch).

And probably you'll need `asString`, `asInt`, and `asFloat` as well...

common/src/main/java/com/wynntils/core/functions/expressions/ConstantExpression.java
```diff
@@ -0,0 +1,78 @@
...
+    private static Optional<Object> doubleParser(String rawString) {
+        try {
+            return Optional.of(Double.parseDouble(rawString));
+        } catch (NumberFormatException e) {
+            return Optional.empty();
+        }
+    }
+
+    private static Optional<Object> booleanParser(String rawString) {
+        if (rawString.equalsIgnoreCase("true") || rawString.equalsIgnoreCase("false")) {
```
magicus, line 70: maybe also allow more variants? like `t` or `f` or `1` and `0`?

kristofbolyai: I fear that adding these will somehow make the parser less reliable. I noted this to myself, and I will get back to this. But until the full system is emerged, I will hold off on doing this.

PR #1182, opened 2023-02-20
fix: Fix parsing of weapons with gear skins

P0keDev:
Weapons with skins would fail to parse if Wynncraft updated their skin at the same time a spell was cast - it would be detected as a new item, but not as a gear item because it would have the spell cast line as its name. On any further updates it would be detected as the same item and the annotation would never get updated.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
@@ -21,13 +21,16 @@ import java.util.Map;
 import java.util.Optional;
 import net.minecraft.ChatFormatting;
 import net.minecraft.core.NonNullList;
-import net.minecraft.nbt.ListTag;
 import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
 import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class ItemHandler extends Handler {
+    private static final List<Item> WILDCARD_ITEMS = List.of(Items.DIAMOND_SHOVEL, Items.DIAMOND_PICKAXE);
+
     private final List<ItemAnnotator> annotators = new ArrayList<>();
     private Map<Class<?>, Integer> profilingTimes = new HashMap<>();
     private Map<Class<?>, Integer> profilingCounts = new HashMap<>();
@@ -89,16 +92,14 @@ public class ItemHandler extends Handler {
         }
 
         // Check if item type, damage and count matches, if not, it's definitely a new item
-        if (!similarStack(existingItem, newItem)) {
+        // Wildcard items are exempt from this check due to the possibility of gear skins
+        if (!similarStack(existingItem, newItem) && !isWildcardItem(existingItem) && !isWildcardItem(newItem)) {
             annotate(newItem);
             return;
         }
 
         // This might be just a name update. Check if lore matches:
-        ListTag existingLore = LoreUtils.getLoreTag(existingItem);
-        ListTag newLore = LoreUtils.getLoreTag(newItem);
-
-        if (!LoreUtils.isLoreEquals(existingLore, newLore)) {
+        if (!LoreUtils.loreSoftMatches(existingItem, newItem)) {
             // This could be a new item, or a crafted item losing in durability
             annotate(newItem);
             return;
@@ -158,6 +159,11 @@ public class ItemHandler extends Handler {
         }
 
         return firstItem.getDamageValue() == secondItem.getDamageValue();
+    }
+
+    private boolean isWildcardItem(ItemStack itemStack) {
+        // This checks for gear skin items, which are a special exception for item comparisons
+        return WILDCARD_ITEMS.contains(itemStack.getItem());
     }
 
     private ItemAnnotation calculateAnnotation(ItemStack item, String name) {
--- a/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
@@ -99,7 +99,7 @@ public class ItemHandler extends Handler {
         }
 
         // This might be just a name update. Check if lore matches:
-        if (!LoreUtils.loreSoftMatches(existingItem, newItem)) {
+        if (!LoreUtils.loreSoftMatches(existingItem, newItem, 3)) {
             // This could be a new item, or a crafted item losing in durability
             annotate(newItem);
             return;
--- a/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
@@ -13,6 +13,7 @@ import com.wynntils.mc.extension.ItemStackExtension;
 import com.wynntils.utils.mc.ComponentUtils;
 import com.wynntils.utils.mc.LoreUtils;
 import com.wynntils.utils.mc.McUtils;
+import com.wynntils.utils.wynn.WynnItemMatchers;
 import com.wynntils.utils.wynn.WynnUtils;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -149,15 +147,22 @@ private String getBaseName(String name) {
         return bracketIndex == -1 ? name : name.substring(0, bracketIndex);
     }
 
-    private boolean similarStack(ItemStack firstItem, ItemStack secondItem) {
-        if (!firstItem.getItem().equals(secondItem.getItem())) return false;
+    private boolean sameGearWithSkin(ItemStack existingItem, ItemStack newItem) {
+        if (WynnItemMatchers.isGearSkin(existingItem) || !WynnItemMatchers.isGearSkin(newItem)) return false;
```
magicus: I've tried hard to keep "specific" item knowledge out of the ItemHandler. Before you merge this, let me check a bit closer if there is a way to solve this which does not mean that the ItemHandler has to know about gear and skins...

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -88,20 +89,17 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
             return;
         }
 
-        // Check if item type, damage and count matches, if not, it's definitely a new item
```
magicus, line 91: This is a very handy shortcut. Why did you remove this? Did any of item type, damage or count change when applying a skin?

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -88,20 +89,17 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
...
-        if (!similarStack(existingItem, newItem)) {
-            annotate(newItem);
-            return;
-        }
-
         // This might be just a name update. Check if lore matches:
         ListTag existingLore = LoreUtils.getLoreTag(existingItem);
         ListTag newLore = LoreUtils.getLoreTag(newItem);
 
         if (!LoreUtils.isLoreEquals(existingLore, newLore)) {
```
magicus: ~~I'm kind of thinking loud here, but what if you just turned `LoreUtils.isLoreEquals` into a `LoreUtils.isLoreEqualsIgnoringSkin`? Would that not solve the entire problem? And that means that the "knowledge" of the skin line is kind of contained in the LoreUtils function. And all it needs to do is filter out `^\\[Active .* Skin: \\]$` from both lore line lists before comparing.~~

~~It could be made even more general by turning it into like `isLoreEqualsIgnoringBracketLines` and just filter out `^\\[.*\\]$`, but maybe that is too generic for no good reason.~~

edit: Bad idea, ignore

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: Hm, I realize now that this line is strictly not necessary. What the code below checks is really "is the new lore identical to the old, except that it has an added line". That is a more general approach.

Sorry for all the confusing comments in random order here, but I think this approach is actually quite good, but you can drop the "isGearSkin" test and just dive directly into comparing the lore. You can also start by checking not only size inequality, but if new is exactly one line longer than existing, otherwise abort quickly.

And name the method not `sameGearWithSkin` but like `isLoreAlmostEquals` (and maybe move it to LoreUtils?)

magicus: ... and then you can actually replace the original call to `isLoreEquals` with `isLoreAlmostEquals`, you don't need to check this twice.

P0keDev: This is a great point actually! I don't know off the top of my head if there are any other instances where lore would get "added to" like this, but it's probably safe to assume that if they exist, they can be handled in the same way. I will try to implement this later today

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
P0keDev: The item has to fundamentally change for a skin to work, so yes. I'm unsure if there's cases where the item itself will change or if it's only the item damage, but I'd say it's safest to remove the check altogether for this context

magicus: The `similarStack` methods looks for three things: if the Item of the two ItemStacks are different. If they are, they can not be the same item. If they are the same, then the count is checked. If the count is different, it cannot be the same item. And finally, if the item and count matches, it also checks the "damage" value. If it is different, it cannot be the same value.

I think I need to log in and check weapon skins myself, but I have a hard time imagining that any of these change when you apply a skin. Or, oh, hm... does the damange value change maybe? The item looks differently in the hotbar with a skin applied?

I need to log in and experiment a bit.

Anway, it's kind of sad if we need to remove this. It's a quick shortcut that makes us skip reading and comparing lore texts for most item changes. Like each time you switch pages in the bank, then 99% of all items get caught in this check and we don't have to do the expensive lore reading.

At the very least we should be able to keep the count check. I hope we can also keep the item type, but we might have to skip the damage check.

P0keDev: weapon skins change the model of the weapon; that can't be achieved without at the very least changing the item's damage value

you're right about the stack count though, that can definitely be kept regardless

magicus: Ok, I have checked a bit. As far as I could tell, *all* weapon skins are `diamond_shovel`. And head-wear skins are `diamond_pickaxe`. That means we could accept that these two items are "wildcard" items that match anything. So in similarStack, the short-circuit logic will return false if the two Items are different, but not if one of the two is a "wildcard" item.

something like this:
```
    private boolean similarStack(ItemStack firstItem, ItemStack secondItem) {
        if (!firstItem.getItem().equals(secondItem.getItem()) &&!isWildcardItem(firstItem) && !isWildcardItem(secondItem)) return false;
       
```
and `isWildcardItem` checks for equality with `Items.DIAMOND_PICKAXE` or `Items.DIAMOND_SHOVEL`.

magicus: I realize this hard-codes a bit Wynncraft-y behavior, but I'm willing to sacrifice that to keep the high-performing short-circuit logic of `similarStack`.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: Yes, if this kind of behavior would appear anywhere else, I think it will be reasonable to consider it the "same" item.

common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
```diff
@@ -242,6 +242,25 @@ public static boolean isLoreEquals(ListTag existingLore, ListTag newLore) {
         return existingLoreString.equals(newLoreString);
     }
 
+    /**
+     * This checks if the lore of the second item contains the entirety of the first item's lore.
```
magicus: I'm thinking this should be symmetric, so it also returns true if the first item contains all of the second items lines.

Then it will catch the case of disabling a weapon skin as well.

Or maybe I'm missing something in my thinking here.

magicus (commented):
Overall I think this looks very good. And it works, I hope? :-)

While this solves the transition "no skin" -> "has skin", I think the opposite would be nice as well, since the weapon you left will have its skin removed, and it would be nice to not have to re-parse the GearItem for that.

common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
P0keDev: It shouldn't be possible for a skin getting disabled to cause any conflicts, since it can't coincide with a spell cast - but I might as well make this method symmetric anyway

magicus: It will save us a re-parse in GearAnnotator when you switch weapon. So yeah, it's not a bug fix, but a (ridiculous) micro-optimization.

magicus (approved):

PR #1183, opened 2023-02-20
DUMMY: Just for comparing functions part 3

magicus:
This should not be integrated.

### DIFF
--- a/common/src/main/java/com/wynntils/core/functions/FunctionManager.java
+++ b/common/src/main/java/com/wynntils/core/functions/FunctionManager.java
@@ -6,6 +6,7 @@ package com.wynntils.core.functions;
 
 import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Manager;
+import com.wynntils.core.functions.templates.parser.TemplateParser;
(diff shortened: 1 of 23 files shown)


common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -119,21 +120,36 @@ public Component getSimpleValueString(
...
 
-        return format(value.get());
+        return format(value.get(), formatted, decimals);
     }
 
-    private String format(Object value) {
+    private String format(Object value, boolean formatted, int decimals) {
+        if (value instanceof Number number) {
+            if (formatted) {
+                // French locale has NBSP
```
magicus, line 141: Swedish locale, too. And a decimal comma instead of a decimal point. Overall, I think we need to think a bit whether we want to support locale formatting here, or just use `Locale.ROOT`. Maybe it should even be a setting.

kristofbolyai: Yes. I thought about that. Or only using US locale or smth.. Not sure atm.

PR #1184, opened 2023-02-20
feat: Function Refactor Part 4: Argument validation and Generic Functions

kristofbolyai:
This PR changes how expression values are calculated in a way we do not convert function expressions to strings, in case we need to use them as their original values later on. This came up as a problem with argument validation and when nesting functions, both of which should be now functional.


kristofbolyai:
This is ready now. I do not want to introduce any more generic functions in this PR. Instead, I will make part 5 be only about adding new functions, and adapting old ones. I will leave MathFunctions here though, as an example of how these should be used.

common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
```diff
@@ -0,0 +1,161 @@
...
+        public List<String> getAliases() {
+            return List.of("div");
+        }
+    }
+
+    public static class ModuloFunction extends GenericFunction<Double> {
+        @Override
+        public Double getValue(FunctionArguments arguments) {
+            return arguments.getArgument("first").getDoubleValue()
+                    % arguments.getArgument("second").getDoubleValue();
```
magicus, line 91: Kind of weird doing module a non-integer? Is that even well-defined?

common/src/main/java/com/wynntils/core/functions/expressions/FunctionExpression.java
```diff
@@ -93,21 +99,12 @@ public static ErrorOr<Optional<Expression>> tryParse(String rawExpression) {
 
         // Handle argument parsing
 
-        FunctionArguments.Builder argumentsBuilder = function.getArguments();
-
-        if (matcher.groupCount() < 3) {
-            return ErrorOr.of(Optional.of(new FunctionExpression(
-                    rawExpression, function, argumentsBuilder.buildWithDefaults(), isFormatted, decimals)));
-        }
+        FunctionArguments.Builder argumentsBuilder = function.getArgumentsBuilder();
```
magicus, line 102: Just checking here. This is done at parse time, right? Which for info boxes is done only on config update time, not at render time..? (Sorry, I'm being a bit lazy and try to skip having to delve into the code to check this...)

common/src/main/java/com/wynntils/core/functions/arguments/parser/ArgumentParser.java
```diff
@@ -16,7 +16,17 @@ public final class ArgumentParser {
...
+            // 1, If there are no arguments, return early.
+            if (argumentsBuilder.getArgumentCount() == 0) {
+                return argumentsBuilder.buildWithValues(List.of());
+            }
+
+            // 2, If there are required arguments, return an error, otherwise return the default arguments.
+            if (argumentsBuilder instanceof FunctionArguments.OptionalArgumentBuilder optionalArgumentBuilder) {
+                return ErrorOr.of(optionalArgumentBuilder.buildWithDefaults());
+            } else {
+                return ErrorOr.error("Missing required arguments: (%s)".formatted(argumentsBuilder.getArgumentNames()));
```
magicus, line 28: This assumes its some kind of RequiredArgumentBuilder we got then..? I probably missed the full understanding of this in the previous PRs.

common/src/main/java/com/wynntils/core/functions/expressions/FunctionExpression.java
kristofbolyai: Yes, updated on parse time, configurable for info boxes

common/src/main/java/com/wynntils/core/functions/arguments/parser/ArgumentParser.java
kristofbolyai: Builder is required, OptionalRequirementBuilder is not

kristofbolyai: (it is not on per argument basis, rather in function basis)

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -26,38 +25,59 @@ public Argument getArgument(String name) {
         return this.lookupMap.get(name);
     }
 
-    public static final class Builder {
+    // A builder for functions with required arguments
```
magicus: Here was the answer to my previous question (if it's confusing, it's because I ended up reviewing from bottom to top... don't ask why 🤷 )

Maybe make Builder abstract and have a second concrete subclass RequiredArgumentBuilder, to clarify the difference?

magicus (approved):
Looks good. I left a few comments, feel free to address them if you agree, or ignore them if you think you have good reason to. :)

common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
```diff
@@ -0,0 +1,161 @@
...
+                    List.of(new FunctionArguments.Argument<>("value", Number.class, null)));
+        }
+
+        @Override
+        public List<String> getAliases() {
+            return List.of("sqrt");
+        }
+    }
+
+    public static class RoundFunction extends GenericFunction<Double> {
```
magicus, line 146: I think you also need an `int` function that rounds to the nearest integer, and returns the result as an `Integer`. So if you e.g. force the modulo function to take int as the second argument, the user can use the `int` function to make sure they send valid values.

magicus: (At first I thought that round should return int but then I realize this has a slightly different meaning, which is totally fine, doing it an int is a different task)

common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
kristofbolyai: Fully defined within Java specs. Only C/C++ can't handle it. https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.17.3

common/src/main/java/com/wynntils/core/functions/expressions/FunctionExpression.java
kristofbolyai: Oh. I just realized. Well, it is done on calculation time, which is "render" time (every 5 ticks). Yes, I know that is not so optimal. I would wait with optimizing though. If we profile this, it might not even show up (or be significant).

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
kristofbolyai: Done.

common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
kristofbolyai: I added it.

common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
magicus: Goodness gracious... But, even if Java allows it, I think it makes no sense. I think we will help avoid problems for users if we limit the second argument to integers.

PR #1193, opened 2023-02-23
feat: Multiple improvements to Wynncraft command expansion

magicus:
This PR adds more commands from Wynncraft, marks aliases as such, and makes it possible to select all or just a few short-hand aliases to be included in expansion. It also makes it possible to hide deprecated commands. All in all, this makes for a much cleaner UX where only the relevant commands are presented as expansions.

Furthermore, I have made sure commands that take arguments get proper suggestions, as far as possible.

And finally I also sneaked in a fix for ChatTabs where the auto command will now present a proper tooltip instead of an error message, by working around an issue in Vanilla.


kristofbolyai (dismissed):
Looks good. I think all command renames, including the next PR should be merged at the same time and announced to not cause (a lot of) confusion.

magicus:
I'll merge both PRs into a common branch, and then integrate once, then.

PR #1194, opened 2023-02-24
feat: Wynntils commands overhaul

magicus:
Giving our commands some well needed TLC

Noteworthy changes: 

`/territory` is now `/compass territory`
`/token` is now `/wynntils token`
`/update` is now `/wynntils update`

All commands are listed using `/wynntils help`, and this is now enforced by the Command class. Descriptions in `/wynntils help` are updated to be more to the point and more consistent.


kristofbolyai:
> `/territory` is now `/compass territory`

This is such a "breaking" change to me, and many users. I fully understand why you would group the command under the `/compass` command, but everyone using this command is so used to it, that we can't change this (sadly). (You have to understand we are talking about hard core guild players..). Rather than making it more complicated, do we have `/terr` as an alias? I think we should.

kristofbolyai (approved):
LGTM

PR #1195, opened 2023-02-24
feat: Custom cape and elytra cosmetics

P0keDev:
Essentially ports legacy's cape and elytra handling. Custom `RenderLayer` objects are attached to the vanilla player render engine to handle our versions of the respective cosmetics, and any capes or elytras being rendered through the regular layer will be disabled if ours is active.

I had to change the way that `PlayerModel` loaded players, since the original `PlayerJoinedWorldEvent` was implemented in such a way that it never triggered on the user's own account.

image


common/src/main/java/com/wynntils/features/user/players/WynntilsCosmeticsFeature.java
```diff
@@ -0,0 +1,196 @@
...
+                float netHeadYaw,
+                float headPitch) {
+            if (!parent.shouldRenderCosmetic(player, Type.CAPE)) return;
+
+            ResourceLocation texture = parent.getTexture(player);
+            if (texture == null) return;
+
+            poseStack.pushPose();
+            poseStack.translate(0.0f, 0.0f, 0.125f);
+            double d = Mth.lerp(partialTick, player.xCloakO, player.xCloak)
```
kristofbolyai: Is this math copied from MC source?

common/src/main/java/com/wynntils/mc/mixin/ClientLevelMixin.java
```diff
@@ -6,13 +6,19 @@
...
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(ClientLevel.class)
 public abstract class ClientLevelMixin {
+    @Inject(method = "addPlayer(ILnet/minecraft/client/player/AbstractClientPlayer;)V", at = @At("HEAD"))
+    private void addPlayer(int id, AbstractClientPlayer player, CallbackInfo ci) {
+        EventFactory.onPlayerJoinedWorld(player);
```
kristofbolyai, line 19: To be sure, this is the change so we load our own user's info too?

kristofbolyai (dismissed):
Looks good overall. Do you want to review this or are we good to merge? @magicus

common/src/main/java/com/wynntils/mc/mixin/ClientLevelMixin.java
P0keDev: It is, yes. The method we used to mix into delegated to this one, actually - but it was the packet listener class, so it was only handling other players

common/src/main/java/com/wynntils/features/user/players/WynntilsCosmeticsFeature.java
P0keDev: yupp

i can try to clean it up if you want? this code is for sure a huge mess

kristofbolyai: I think this is different from other cases, where we want to fully preserve mc code. So yes, cleanup please.

magicus:
Let me have a look later tonight.

magicus (dismissed):
I think it looks good. You have a trivial-ish merge conflict from my changes in PlayerModel incoming. I can't read and understand the rendering code, but if bolyai wants cleaner gibberish, then you should probably do that :-)

magicus (approved):

PR #1197, opened 2023-02-24
feat: Add keybind for waypoint creation

mbnkwm:
https://user-images.githubusercontent.com/71361356/221200969-8caf932c-079b-45b9-886d-7aa8b2a69cd5.mp4


common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -131,6 +136,15 @@ private void openMainMap() {
         McUtils.mc().setScreen(MainMapScreen.create());
     }
 
+    private void openWaypointSetup() {
+        var location = new PoiLocation(
+                McUtils.player().getBlockX(),
+                McUtils.player().getBlockY(),
+                McUtils.player().getBlockZ());
+
+        McUtils.mc().setScreen(PoiCreationScreen.create(null, location));
```
kristofbolyai, line 145: I would like you to add a null-check in `PoiCreationScreen#onClose` to handle passing `null` as the old screen value.

kristofbolyai (commented):
Otherwise LGTM, thank you :)

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
mbnkwm: A null check there would make it impossible to leave the waypoint screen, it has to pass through to the setScreen call

kristofbolyai: Right. I completely forgot that setScreen(null) means to close the screen.. This is why you don't do code reviews after a whole day of competitions..

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/features/user/map/MapFeature.java
```diff
@@ -131,6 +136,15 @@ private void openMainMap() {
         McUtils.mc().setScreen(MainMapScreen.create());
     }
 
+    private void openWaypointSetup() {
+        var location = new PoiLocation(
```
kristofbolyai:
```suggestion
        PoiLocation location = new PoiLocation(
```

We do not use var anywhere in Artemis. And we won't.

kristofbolyai (approved):

magicus (approved):

magicus:
@mbnkwm Thank you for your contribution to Wynntils!

PR #1199, opened 2023-02-24
fix: Map data were missing on dev version

magicus:
Event signalling loading of services failed to be delivered

### DIFF
--- a/common/src/main/java/com/wynntils/core/WynntilsMod.java
+++ b/common/src/main/java/com/wynntils/core/WynntilsMod.java
@@ -74,6 +74,10 @@ public final class WynntilsMod {
         }
     }
 
+    public static void postEventOnMainThread(Event event) {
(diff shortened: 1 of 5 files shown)


P0keDev (approved):

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -74,6 +74,10 @@ public static boolean postEvent(Event event) {
         }
     }
 
+    public static void postEventOnMainThread(Event event) {
```
kristofbolyai, line 77: We could/should have replaced all other occurrences of this. Doesn't really matter though.

magicus: Most places should *not* call this, I assume.

PR #1202, opened 2023-02-25
fix: At least print the major ID name, even if formatting is off

magicus:


kristofbolyai (approved):
Why is it so hard to do this in the first place?

magicus:
@kristofbolyai Feel free to address the core issue! The problem is that we use `RenderedStringUtils.wrapTextBySize()`, and it takes in a string, not a component. Rewriting it to work on components instead was, well, more work than I have been eager to get on to fixing...

It might be possible of course, I realize now, that there is a Vanilla corresponding function for Components already.

PR #1206, opened 2023-02-26
feat: Introduce the new functions system

kristofbolyai:


common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+package com.wynntils.core.functions.arguments;
+
+import com.wynntils.utils.type.ErrorOr;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+// FIXME: Expose this to user and add i18n
```
magicus: I think this is done..?

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -0,0 +1,155 @@
...
+    public abstract static class Builder {
+        protected List<Argument> arguments;
+
+        protected Builder(List<Argument> arguments) {
+            this.arguments = arguments;
+        }
+
+        public ErrorOr<FunctionArguments> buildWithValues(List<Object> values) {
+            if (values.size() != this.arguments.size()) {
+                return ErrorOr.error("Invalid number of arguments.");
```
magicus:
```suggestion
                return ErrorOr.error("Invalid number of arguments");
```

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            for (int i = 0; i < this.arguments.size(); i++) {
+                Argument argument = this.arguments.get(i);
+
+                if (!argument.getType().isAssignableFrom(values.get(i).getClass())) {
+                    return ErrorOr.error("Invalid argument type: \"%s\" is not a %s."
```
magicus:
```suggestion
                    return ErrorOr.error("Invalid argument type: \"%s\" is not a %s"
```

common/src/main/java/com/wynntils/core/functions/templates/ExpressionTemplatePart.java
```diff
@@ -0,0 +1,45 @@
...
+import com.wynntils.utils.type.ErrorOr;
+
+public class ExpressionTemplatePart extends TemplatePart {
+    private final String expressionString;
+
+    public ExpressionTemplatePart(String part) {
+        super(part);
+
+        if (!this.part.startsWith("{") || !this.part.endsWith("}")) {
+            throw new IllegalArgumentException("Expression was not wrapped in curly braces.");
```
magicus, line 18: I'm getting a bit lost in the data flow here, sorry. If the user enters a format string like `unmatched: {mana_max`, will we end up here then? Or is this supposed to be caught before by the parser, so this is just a sanity check? And if we end up here, how does this exception propagate to something nicely for the user to see?

kristofbolyai: This is more of a "there is no way this can happen at the moment since the callers are safe but I left this here".

It is fine to remove. But it is also fine to keep

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -99,132 +105,115 @@ private Optional<Object> getFunctionValueSafely(Function<?> function, String arg
...
-                consumer.accept(mapper.apply(format.substring(index, indexStartOfNextVariable)));
-            }
+        for (int i = 0; i < templateString.length(); i++) {
+            char c = templateString.charAt(i);
+            if (c == '\\') {
+                if (i + 1 < templateString.length()) {
+                    char nextChar = templateString.charAt(i + 1);
 
-            String toMatch = format.substring(indexStartOfNextVariable + 1, indexEndOfNextVariable);
+                    resultBuilder.append(doEscapeFormat(nextChar));
```
magicus, line 197: I tend to agree with Qodana here. Assigning to the loop variable is not nice. I know I have already approved this so I can look the other way and let you push this. But if you think it is possible to get away without doing this, it would be nice. This is a quite complex loop; it is possible it will be just worse if you try to do it. But otoh, just since it is complex, it helps to keep it simple by e.g. not violating commonly held assumptions about loop variables.

Now I said it at least. You decide if it is possible to make this cleaner, or if you just want to press Resolved and get on with your life. :)

magicus (dismissed):
This has already been review in parts. I noticed a few odds and ends that I missed the previous reviews, but overall this looks like a great improvement!

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
```diff
@@ -247,27 +236,79 @@ private String parseColorCodes(String toProcess) {
...
-            case "E" -> EmeraldUnits.EMERALD.getSymbol();
-            case "B" -> EmeraldUnits.EMERALD_BLOCK.getSymbol();
-            case "L" -> EmeraldUnits.LIQUID_EMERALD.getSymbol();
-            case "M" -> "✺";
-            case "H" -> "❤";
+            case '\\' -> "\\\\";
+            case 'n' -> "\n";
+            case '{' -> "\\[\\";
+            case '}' -> "\\]\\";
+            case '§' -> "&";
```
magicus: Just as a reminder, at least allow `\&` as an escape for `&`, and possibly remove this odd § one.

Also, better fallback than null for unknown escape characters.

common/src/main/java/com/wynntils/core/functions/FunctionManager.java
kristofbolyai: I don't think anything good will come out of this. I share your concern, but I think we might have to do it here.

common/src/main/java/com/wynntils/core/functions/templates/ExpressionTemplatePart.java
kristofbolyai: So, what do you think @magicus ?

magicus: That was basically what I wanted to hear, that is just a fail-safe. Thanks.

magicus (approved):

PR #1207, opened 2023-02-26
feat: Function Refactor Part 5: Lots of generic functions and function adoptations

kristofbolyai:


common/src/main/java/com/wynntils/functions/CombatXpFunctions.java
```diff
@@ -76,10 +73,10 @@ public String getValue(FunctionArguments arguments) {
         }
     }
 
-    public static class XpRawFunction extends Function<Float> {
+    public static class XpRawFunction extends Function<Double> {
         @Override
-        public Float getValue(FunctionArguments arguments) {
-            return Models.CombatXp.getCurrentXp();
+        public Double getValue(FunctionArguments arguments) {
+            return (double) Models.CombatXp.getCurrentXp();
```
magicus, line 79: Is this cast really needed?

common/src/main/java/com/wynntils/functions/generic/ConditionalFunctions.java
```diff
@@ -0,0 +1,59 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.functions.generic;
+
+import com.wynntils.core.functions.GenericFunction;
+import com.wynntils.core.functions.arguments.FunctionArguments;
+import java.util.List;
+
+public class ConditionalFunctions {
+    public static class IfStringFunction extends GenericFunction<Object> {
```
magicus: Hm. This looks like it could do with a bit of generics. Create an abstract IfFunction<T> and just let IfStringFunction extend IfFunction<String>. Relace the String.class with T in the argument builder. And then you can also replace the return value of getValue to T.

magicus: Or, as things look rn, I don't even see why you have two version. Both have getValue that returns an object. Maybe you should instead allow for a general argument parser that can take a string as well as a number, and just returns an object?

kristofbolyai: > Or, as things look rn, I don't even see why you have two version. Both have getValue that returns an object. Maybe you should instead allow for a general argument parser that can take a string as well as a number, and just returns an object?

I am quite happy with how it is right now, except for this singular function... I don't think I should design the whole parser around one of the most "edge case" functions. So, I can generify this, will you accept that?

common/src/main/java/com/wynntils/functions/CombatXpFunctions.java
kristofbolyai: Yes, I would rather have all functions return a double, but the combat xp model is heavily using floats at the moment. I would rather not change every float to double, or do you want me to?

common/src/main/java/com/wynntils/functions/generic/ConditionalFunctions.java
```diff
@@ -0,0 +1,59 @@
...
+                    new FunctionArguments.Argument<>("condition", Boolean.class, null),
+                    new FunctionArguments.Argument<>("ifTrue", String.class, null),
+                    new FunctionArguments.Argument<>("ifFalse", String.class, null)));
+        }
+
+        @Override
+        public List<String> getAliases() {
+            return List.of("if_str");
+        }
+    }
```
kristofbolyai, line 36: @magicus I know this is not what you imagined. But I can't use `T.class` and I need that in `getRequiredArgumentsBuilder`. So I resorted to this smaller base class. I hope it is fine.

common/src/main/java/com/wynntils/functions/CombatXpFunctions.java
magicus: No, don't change type. I thought the value could be widened automatically but maybe that does not work when it also needs boxing.

common/src/main/java/com/wynntils/functions/generic/ConditionalFunctions.java
magicus: Yeah, okay. I guess this is due to type erasure. :-( I will not hold limitation in Java's type system against you. :)

magicus:
@kristofbolyai Do you intend to do more work on the full refactor branch, or is it time to bring it into mainline?

kristofbolyai:
> @kristofbolyai Do you intend to do more work on the full refactor branch, or is it time to bring it into mainline?

There are more work to do. Mainly command updates, actual infobox impl, removing legacy, adding argument i18n.

PR #1210, opened 2023-02-26
chore: Make it clearer for developers when i18n is missing

magicus:
I've been going mad over this more than once...


ryanzhoudev (approved):
Thank you for this, I have also been caught a bunch of times by i18n

PR #1214, opened 2023-02-27
feat: Option to only prevent trade/duel while fighting (damage dealt in the last N seconds)

magicus:
I think this gives us the best of two worlds: no mis-clicks while grinding or fighting in a party, but still allows easy trading when things are calm. A notification tells you what's wrong if you get blocked and did not expect it, saying how long you must wait (and if it is too long, you have to switch away your weapon, just as before).


common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
```diff
@@ -14,25 +16,39 @@
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @FeatureInfo(category = FeatureCategory.PLAYERS)
 public class PreventTradesDuelsFeature extends UserFeature {
+    private static final int FIGHTING_CUTOFF = 30; // 30 seconds
```
kristofbolyai: Why not a config?

common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
```diff
@@ -14,25 +16,39 @@
...
 
     @SubscribeEvent
     public void onPlayerLeftClick(PlayerAttackEvent event) {
-        if (!shouldBlockClick(event.getPlayer(), event.getPlayer().getMainHandItem(), event.getTarget())) return;
+        handlePlayerClick(event, event.getPlayer(), event.getPlayer().getMainHandItem(), event.getTarget());
+    }
 
-        // stops attack packet from going out
+    private void handlePlayerClick(Event event, Player player, ItemStack itemStack, Entity target) {
+        int timeSinceLastFight = (int) ((System.currentTimeMillis() - Models.CombatXp.getLastXpGainTimestamp()) / 1000);
```
kristofbolyai: This is not fully what the PR title means. There are many cases where you don't gain any XP, yet you are still fighting. Like raid bosses. Those are places you REALLY don't want to trade, but you still do a lot of shift-clicking to trigger powder specials. 

Preferably we should implement the DPS tracker from legacy (and add `area_damage_per_second` function, as it is one the most requested one at the moment.) and use it here. Until that happens, can we add a FIXME here (and somehow rename the PR to reflect this)?

magicus: Yes, you are correct. I used "time since last combat xp" as a proxy to mean "is fighting". Most of the cases when you fight a boss there will be minion mobs that are killed in the process, so I think this will work fine in most cases. But I did consider the possibility of a single, powerful boss as a way that this could fail. 

Note that the i18n description clarifies that "is fighting" counts as being less than 30 seconds since the last killed mob. I know that RTFM is a bad excuse for creating lousy programs, but at least it is documented.

I was toying with the possibility of tracking spells and melee attacks, but it was just too hard. Spells could be teleport spells, melee attacks differs between archer and others, and maybe you are just swinging your fist in the air to say hi, etc. 

If we could track damage dealt it would be an improvement, yes. How did we do that in legacy? Count the heart popup AS texts?

common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
magicus: I did not see the need to change this on a per-user basis. If the value is bad, I think we should adjust it for all users. I did some simple testing with me and my alt account and I thought this seemed like a good tradeoff between not having to wait too long but still keep the prevention while actively engaged. Too many config just makes everything confusing.

Are you happy with that reply or do you want me to turn it into a config?

common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
kristofbolyai: > If we could track damage dealt it would be an improvement, yes. How did we do that in legacy? Count the heart popup AS texts?

AFAIK, we tracked the pop-up "labels". You would have to check to make sure though.

The second, more important question is what we should do with the PR. Imo, accept it. Even if you want to do damage tracking, there is no need to be in this PR, and it can be a direct follow up.

common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
kristofbolyai: Okay. 30 seconds sounds reasonable. Let's leave it.

common/src/main/java/com/wynntils/features/user/players/PreventTradesDuelsFeature.java
```diff
@@ -14,25 +16,41 @@
...
 import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.Event;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @FeatureInfo(category = FeatureCategory.PLAYERS)
 public class PreventTradesDuelsFeature extends UserFeature {
+    private static final int FIGHTING_CUTOFF = 30; // 30 seconds
+
+    @Config
+    public boolean onlyWhileFighting = true;
```
ryanzhoudev: I think this should be off by default, or maybe have the timer be a user config. 30 seconds is quite a bit during raids, where you can finish a room pretty quickly, then be teleported to the buff room, where you might need to trade potions or other consumables to another player.

magicus: So you're worry is that it is too high? I was just being conservative here. This feature is on by default (which I question the logic in, at least in the state it is without this PR), which will disable all trade, ever, if you hold a weapon. 

But I will make it a config (requested by both you and bolyai now), and lower it to 15 seconds. 

I'm also thinking if checking for boss bars would be a good sign of a fight in progress. I just need to differentiate them from "informative" boss bars.

ryanzhoudev: You could check for highlights nearby that correspond with a boss bar, most bosses have the glowing effect applied by the server now

magicus: I think actually the bossbar handler can be easily extended to catch if we're currently dealing damage to a mob. I'll hold on integrating this for a while, until I've seen what I can do of that...

magicus:
I'm awaiting https://github.com/Wynntils/Artemis/pull/1218 before continuing on this.

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
```diff
@@ -162,4 +162,8 @@ public TimedSet<Float> getRawXpGainInLastMinute() {
     public TimedSet<Float> getPercentageXpGainInLastMinute() {
         return percentageXpGainInLastMinute;
     }
+
+    public long getLastXpGainTimestamp() {
```
magicus, line 166: Even though it is not needed here now, I'm keeping this for future use.

magicus:
I think this works pretty nicely now. There was a bug in the damage model, where I presumed damages would come in a certain order. This was most often correct, but not always... 

The damage dealt (damage labels) functionality is very precise, so I lowered the default value. This means you can trade with your buddies almost directly after a fight, but as long as you are fighting, trading is blocked. If you are in a fight but have not done any damage whatsoever for 10 seconds, you're not trying very hard. :-)

kristofbolyai (approved):
LGTM, I see we bundle some damage model changes too. It used to assume element order, which seems to not be static, right?

magicus:
> I see we bundle some damage model changes too. It used to assume element order, which seems to not be static, right?

Yes. It is all very weird. Most of the time it followed this order, but sometimes the "heart" damage was first, and sometimes it was last. I'm still not sure I understand the Wynncraft damage system enough to understand what this means. Is it that I inflicted neutral damage, or is it just a summary of how much health a mob lost? Or maybe they heart is used for the two different cases, and that is why it sometimes is first and sometimes is last? In any case, this solution works and is arguable more robust as well.

PR #1217, opened 2023-02-27
fix: Fixed typos in en_US.json

Rafii2198:
As I said multiple times, there were just a few and small typos here! It is really great job from you all! In total, only 4 words were changed!


magicus:
Thanks for the fixes! The incorrect keys for functions worries me a bit, I think we need to add a check that we have proper i18n there...

Rafii2198:
Only horse functions were standing out for some reason, all other functions are working properly, it seems.

Rafii2198:
Just as I sent that, I found that memory functions are affected in the same way.

magicus:
Actually, we're missing a ton of i18n for functions! I'll open a separate bug on that...

edit: See https://github.com/Wynntils/Artemis/pull/1220. You don't have to start fixing the incorrect function keys here if you don't want to.

Rafii2198:
Alright, I'll just concentrate this PR for typos for now, as this was its original purpose.

magicus (approved):

PR #1218, opened 2023-02-27
feat: Introduce DamageModel

magicus:


magicus:
Right now I only need the timestamp, but I'm sure this can form the basis of a DPS counter.

magicus:
Also an additional future enhancement is to send `DamageReceived` events, where we track the health bar label of nearby mobs. So if it goes from `[||||1000||||]` to `[||||800||||]` we know the mob received 200 damage between updates. I think this is what is requested by the "area DPS", seeing how much AoE damage you deal (if you are alone), or how much damage a party deals collectively. (Note: one-hit kills will not be counted, but I don't know how we could ever do that...)

To do this we need to detect when these labels appear, track each label individually how it lower its value, and notice when it is removed (presumably because health dropped to 0). I started doing this but it was a bit too much work, in what was a tangent for what I really wanted to do. Mostly, I got stuck trying to compose a meaningful regex to match the health bar labels...

For the record: I spotted some really weird health bar labels with a trailing skull: `§4[§c|||||§419946§c|||||§4] §5☠§736`. I did not see these in the game myself, they just registered on my function. I don't know what these means.

kristofbolyai:
> For the record: I spotted some really weird health bar labels with a trailing skull: `§4[§c|||||§419946§c|||||§4] §5☠§736`. I did not see these in the game myself, they just registered on my function. I don't know what these means.

Some mobs have immunity to some effects, those symbols track that.

kristofbolyai (dismissed):
Can you add commands regarding what the 2 new events do? After that, LGTM

PR #1224, opened 2023-02-28
feat: Show item level when pressing shift

magicus:
Also adds level extraction to several item annotators


P0keDev (changes requested):
Binding this functionality to the shift key makes it pretty obtrusive, especially since there's no toggle for it. Ideally this should be a regular keybind that can be edited by the user; at the very least it should be changed to left control, to match the default key in legacy

magicus:
It breaks for stacking items; need to fix that first.

magicus:
We don't have any precedence of changing key bindings for when just checking if modifier keys are pressed. You have a point in that it would be nice to have this configurable, but I think it needs a bit more general rework. I changed it to control.

magicus:
It turned out we needed to disable vanilla count rendering more properly than to change the actual stack count (that one always made me squirm anyway when I saw it...).

But. There is a nice way to hook into vanilla to draw a random text instead of the stack count. The code could be made much nicer if we were to use it. But the limitation then is that the count text must be white. I started out doing the count white, but P0ke changed it so the skill point books on the compass menu gets colored. @P0keDev, would you be willing to sacrifice this? If so, we can just send in the count value to override right into the mixin, and can skip this turn on/turn off business.

P0keDev:
I'm not too sure about this new solution; if we're adding a new mixin, why not create one that allows us to cancel the item text drawing instead? That would make things much cleaner and wouldn't have the drawback of giving up color

> We don't have any precedence of changing key bindings for when just checking if modifier keys are pressed

The level overlay button is configurable in legacy (though I'm not sure if the same method would really be applicable in Artemis atm)

magicus:
> I'm not too sure about this new solution; if we're adding a new mixin, why not create one that allows us to cancel the item text drawing instead?

That is indeed what I am doing. 

Unfortunately the vanilla code is quite convoluted, and the only sane way I could find out to "cancel" the drawing is to:
1) flag in the WynnItem cache that we should cancel drawing
2) when about to draw the count in vanilla, hook in a mixin that checks this flag
3) if the flag is set, ask vanilla to draw the text "" instead
4) then draw our own text
5) and finally clear out the flag

I mean, it works. And you do get your colored texts. :-) And it does not mess with the item stack count, like the old solution. But we substituted like one line of code for a very complicated system. :-(

When I had done all this, I realized we could have instead done:
1) when about to draw the count in vanilla, hook in a mixin that sends an event
2) listen to this event, and set the count value in the event to what we want to show instead

It would have been much cleaner, and more in line with how the rest of Artemis is working. But, this is not what is proposed in this PR, since I could not find a way to change the color then. :-(

magicus:
Hm, maybe I can pull off the second solution, with color change included, after all...

P0keDev:
> > I'm not too sure about this new solution; if we're adding a new mixin, why not create one that allows us to cancel the item text drawing instead?
> 
> That is indeed what I am doing. 
> 
> Unfortunately the vanilla code is quite convoluted, and the only sane way I could find out to "cancel" the drawing is to:
> 1) flag in the WynnItem cache that we should cancel drawing
> 2) when about to draw the count in vanilla, hook in a mixin that checks this flag
> 3) if the flag is set, ask vanilla to draw the text "" instead
> 4) then draw our own text
> 5) and finally clear out the flag
> 
> I mean, it works. And you do get your colored texts. :-) And it does not mess with the item stack count, like the old solution. But we substituted like one line of code for a very complicated system. :-(
> 
> When I had done all this, I realized we could have instead done:
> 1) when about to draw the count in vanilla, hook in a mixin that sends an event
> 2) listen to this event, and set the count value in the event to what we want to show instead
> 
> It would have been much cleaner, and more in line with how the rest of Artemis is working. But, this is not what is proposed in this PR, since I could not find a way to change the color then. :-(

my suggestion is that instead of using the whole convoluted flag system, the mixin should post its own event, and then cancelling the event would stop the text from being drawn (by setting the text to an empty string, or whatever other applicable method)

listeners for that event can thus cancel it and do any sort of custom text rendering they'd like

magicus:
I think I got it pretty nice now. In fact, possibly almost better than before, since we now use Vanilla rendering.

P0keDev (dismissed):

common/src/main/java/com/wynntils/mc/event/ItemCountOverlayEvent.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.Event;
+
+public final class ItemCountOverlayEvent extends Event {
```
kristofbolyai:
```suggestion
public final class ItemCountOverlayRenderEvent extends Event {
```

What do you think?

kristofbolyai (dismissed):
Otherwise LGTM. I think the @WrapOperation is a bit miss-used(, if that can even happen with a mixin :)). I can't suggest any better alternatives though, so it is fine.

common/src/main/java/com/wynntils/mc/event/ItemCountOverlayEvent.java
magicus: Sure, it's clearer.

PR #1226, opened 2023-02-28
fix: Fix cape texture loading issues

P0keDev:


common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -192,8 +194,13 @@ private void clearUserCache() {
     }
 
     private void clearTextureCache() {
-        cosmeticTextures.forEach((uuid, array) -> Arrays.stream(array)
-                .forEach(l -> McUtils.mc().getTextureManager().release(l)));
+        for (ResourceLocation[] locations : cosmeticTextures.values()) {
+            // Make sure texture unloading is done on client thread
+            Arrays.stream(locations).forEach(l -> {
+                Managers.TickScheduler.scheduleNextTick(
```
magicus: Do you need to schedule one call per location, or could you move the forEach into the scheduled lambda?

P0keDev: whoops, good point

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -192,8 +194,13 @@ private void clearUserCache() {
     }
 
     private void clearTextureCache() {
-        cosmeticTextures.forEach((uuid, array) -> Arrays.stream(array)
-                .forEach(l -> McUtils.mc().getTextureManager().release(l)));
+        for (ResourceLocation[] locations : cosmeticTextures.values()) {
```
magicus, line 197: Just checking here, do the outermost for loop as need to happen when called? Or phrased differently, could not the entire clearTextureCache() be executed on the main loop?

(Maybe I'm missing some race condition on cosmeticTextures here)

kristofbolyai: Wouldn't this create lag spikes in the first place? Why do we need to run on next tick?

P0keDev: if it doesn't run on the client thread, it has a chance of causing texture corruption issues (as far as i know)

this PR is meant to fix the reports of bugged out cape textures that we've been receiving

magicus (approved):
I think it's fine.

PR #1228, opened 2023-03-01
fix: minor tab list spacing issue

ryanzhoudev:
Previously looked like this:
image

Now looks like this:
image

The first image had two spaces between the heart and Windy Feet. Also added the test suite from legacy.


magicus (dismissed):

ryanzhoudev:
Fix now also works with boiling blood. We can't change the regex to accommodate for boiling blood because the space is in the middle of a symbol and color code. (Looks like this `§8⬤ §7Boiling Blood §a(00:02)`). Everything up to and including the 7 will be parsed as part of the prefix. In an effort to make sure the name field is actually just the name and doesn't have strange colours, we shouldn't allow the name capturing group to capture the `§7`

kristofbolyai (approved):

PR #1229, opened 2023-03-01
fix: Strange ordering in party mgmt screen

ryanzhoudev:
Happened when resolving merge conflicts


ryanzhoudev:
Request throttling is because doInit() is spam called in the management screen when the user resizes the window. This also helps prevent a user from spamming the refresh button.

kristofbolyai (approved):

PR #1232, opened 2023-03-01
fix: Move (almost) all Features into categories

magicus:
This makes navigating the settings screen so much easier


magicus:
There were two enums in the feature/maps directory. I inlined one in the class it was used, and moved the other to util, close to other texture stuff. (Perhaps not ideal, but better than having it in features).

The BombBellFeature was not used and were removed.

A handful of features were hard to categorize and were left uncategorized.

common/src/main/java/com/wynntils/features/user/chat/ChatTimestampFeature.java
```diff
@@ -18,6 +21,8 @@
 import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
+@StartDisabled
```
kristofbolyai: Hidden change (or I failed to read something)? Anyways, iirc legacy has this disabled by default too

kristofbolyai (dismissed):

common/src/main/java/com/wynntils/features/user/chat/ChatTimestampFeature.java
magicus: Good catch! I did not intend to sneak this in; I changed it when I was processing that feature and thought to myself "remember to put this in a separate change", and then promptly forgot about it.

I still think it should be done, but I'll do it separately.

PR #1235, opened 2023-03-01
feat: Potion improvement, incl health potion blocker fixes

magicus:
This gives all potions (and crafted consumables) the ability to parse effects. This means the HealthPotionBlockerFeature can use the WynnItem itself to determine if the potion should be blocked.

Also fixes to HealthPotionBlockerFeature:
 * Did not block if clicking on an entity
 * Shorter notification message
 * Do not show a misleading health percentage in notification

 Also refactored GearParser to WynnItemParser, since it is more general than just gears.


common/src/main/java/com/wynntils/models/items/items/game/CraftedConsumableItem.java
```diff
@@ -30,8 +46,19 @@ public GearTier getGearTier() {
         return GearTier.CRAFTED;
     }
 
+    public boolean isHealing() {
+        return getEffects().stream()
+                .filter(e -> e.type().equals("Heal"))
```
kristofbolyai: Qodana is right :)

kristofbolyai (dismissed):
LGTM

common/src/main/java/com/wynntils/models/items/items/game/CraftedConsumableItem.java
magicus: Thanks for noticing this. I just habitually ignore Qodana... :-/

PR #1238, opened 2023-03-01
WIP: Config in models

magicus:


magicus:
Main has diverged too much, and I never really got anywhere here.

PR #1247, opened 2023-03-03
fix: Remove use of @Redirect in mixins

magicus:
This is almost all instances of @Redirect. Some command stuff is trickier and remains to be fixed.

Also contains some related mixin cleanups, like renaming our injected fields to always start with `wynntils` to avoid mod collisions.


magicus:
Verified with Fabric and Forge.

kristofbolyai:
> Also contains some related mixin cleanups, like renaming our injected fields to always start with `wynntils` to avoid mod collisions.

https://gist.github.com/apple502j/48883911bffc4b10bc2975335f2fd5d5#mixin-0123. As a heads up, in the future, there won't be a need for this. (But we still "need" this for current loader versions, many people won't use the new loader for 1.19.3 if not forced to.)

kristofbolyai (approved):
We should use beta 4 of MixinExtras.

magicus:
> We should use beta 4 of MixinExtras.

Yes. Feel free to update. :)

PR #1248, opened 2023-03-03
feat: Functions Refactor Part 7: Fully support arguments in commands and i18n

kristofbolyai:


kristofbolyai:
image
image
image

common/src/main/java/com/wynntils/commands/FunctionCommand.java
```diff
@@ -47,17 +54,20 @@ public String getDescription() {
     @Override
     public LiteralArgumentBuilder<CommandSourceStack> getCommandBuilder() {
         return Commands.literal(getCommandName())
-                .then(Commands.literal("list").executes(this::listFunctions))
+                .then(Commands.literal("list")
+                        .then(Commands.argument("type", StringArgumentType.word())
```
magicus, line 64: Maybe allow also just `/function list` as an alias of `/function list all`? You can do that either by handling the exception in the function on `context.getArgument` if there is no argument, or by calling `listFunction` with either "all" or `context.getArgument("type")` as argument, instead of using a direct method reference. I think we have used both methods about 50% of the time...

common/src/main/java/com/wynntils/commands/FunctionCommand.java
```diff
@@ -34,6 +38,9 @@ public class FunctionCommand extends Command {
                             .map(Function::getName),
                     builder);
 
+    private static final SuggestionProvider<CommandSourceStack> FUNCTION_LIST_TYPES_SUGGESTION_PROVIDER =
```
magicus, line 41: It might be nicer do create an enum and let the provider return `EnumType.values().toLowerCase()`. Or possibly there was even a vanilla type for handling enums, I don't really remember.

magicus (approved):
Except for the `/function list` part this looks good.

PR #1249, opened 2023-03-04
fix: More mixin fixin' 

magicus:
Need to test with Forge first...


magicus:
Verified succesfully on Fabric and Forge.

kristofbolyai (approved):

PR #1250, opened 2023-03-04
fix: Fix multiple issue with command mixin

magicus:
* Make a much leaner mixin implementation
* Use events just as everywhere else
* Get rid of all unchecked warnings
* Rename events and group them together in EventFactory


magicus:
Verified with Fabric and Forge.

kristofbolyai (approved):

kristofbolyai (approved):

PR #1252, opened 2023-03-04
WIP: The Fallen Factory

magicus:
This is how it would look like if we remove the EventFactory. What do you think?


kristofbolyai (commented):
I think this looks fully reasonable.

PR #1258, opened 2023-03-04
fix: Context-aware argument parsing should also know if we are in a string

kristofbolyai:


magicus (approved):
My head is spinning a bit round this. I assume it is OK but I do not fully understand this, but you get a free rubber-stamp :)

PR #1260, opened 2023-03-04
feat: Add more aliases for GreaterThanFunction and GreaterThanOrEqualsFunction

kristofbolyai:


common/src/main/java/com/wynntils/functions/generic/LogicFunctions.java
```diff
@@ -151,7 +151,7 @@ public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
 
         @Override
         public List<String> getAliases() {
-            return List.of("mt");
+            return List.of("mt", "more_than", "gt");
```
magicus, line 154:
```suggestion
            return List.of("mt", "more_than", "gt",">");
```

or do function names have to be a-z?

magicus: Also, just checking, is less than already `lt`?

magicus (approved):

common/src/main/java/com/wynntils/functions/generic/LogicFunctions.java
kristofbolyai: Technically not, but this looks really bad imo.

kristofbolyai: > Also, just checking, is less than already `lt`?

yes

PR #1261, opened 2023-03-04
fix: Fix pouch functions and held item function returns int

kristofbolyai:


common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -107,15 +107,15 @@ public Integer getValue(FunctionArguments arguments) {
             Optional<WynnItem> wynnItem = Models.Item.getWynnItem(itemStack);
 
             if (wynnItem.isPresent() && wynnItem.get() instanceof IngredientPouchItem pouchItem) {
-                return 27 - pouchItem.getCount();
+                return 27 - pouchItem.getIngredients().size();
```
magicus, line 110: What is the difference? Or rather, this is something quite different, right? getCount says how many ingredients we have in total, and this returns the number of different ingredients. Hm, right, so this is a bug fix. I see.

magicus (approved):

PR #1262, opened 2023-03-04
feat: Add AreaDamagePerSecondFunction and AreaDamageAverageFunction

kristofbolyai:


common/src/main/java/com/wynntils/models/damage/DamageModel.java
```diff
@@ -29,6 +31,8 @@ public final class DamageModel extends Model {
 
     private final DamageBar damageBar = new DamageBar();
 
+    private final TimedSet<Integer> areaDamagePerSecondSet = new TimedSet<>(1, TimeUnit.SECONDS, true);
```
magicus: You don't think we should have like a 10 second buffer and calculate the average (ie divide by 10)? This sounds like it can be very unstable.

kristofbolyai: I copied the legacy, requested impl. I can extend it, if you give me a second.

magicus (approved):

PR #1265, opened 2023-03-04
fix: Do not return null in horse functions

kristofbolyai:


common/src/main/java/com/wynntils/functions/HorseFunctions.java
```diff
@@ -15,7 +15,9 @@ public static class HorseLevelFunction extends Function<Integer> {
         @Override
         public Integer getValue(FunctionArguments arguments) {
             HorseItem horse = Models.Horse.getHorse();
-            if (horse == null) return null;
+
+            if (horse == null) return -1;
```
magicus, line 19: I could have sworn I wrote a comment here, but I can't see it now. Anyway, I wondered if it would be better if we returned 0 instead of -1. At least for stuff like tier which can't be 0 anyway, I think 0 makes for a better "n/a" value.

kristofbolyai: We always returned -1 before, I think it is better to keep it universal.

magicus (approved):

PR #1269, opened 2023-03-05
refactor: Cleanups after the EventFactory removal

magicus:


magicus:
I really like how this turned out. The mixin code is now very consistent, self-contained and easy to read. It is a bit of duplication in each mixin, yes, but they all follow a simple pattern:

```
if <not the right conditions> return;

event = new event;
post(event);
if (event.cancel/changed) 
   cancel the mixin parent/change some variables
```

kristofbolyai (approved):

PR #1270, opened 2023-03-05
fix: Setting screen now properly handles config changes 

kristofbolyai:
I hope this does not conflict too badly with #1238 @magicus.

### DIFF
--- a/common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
+++ b/common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
@@ -74,6 +74,8 @@ public final class WynntilsBookSettingsScreen extends WynntilsScreen implements 
 
     @Override
     protected void doInit() {
+        reloadConfigButtons();
+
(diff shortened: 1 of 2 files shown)


magicus:
I might drop #1238, and retry it from a different approach. So don't worry about that.

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
```diff
@@ -54,9 +55,7 @@ public final class WynntilsBookSettingsScreen extends WynntilsScreen implements
     private ScrollButton configurableListScrollButton;
     private ScrollButton configListScrollButton;
 
-    // FIXME: Suboptimal to have both selected variables, but we need big changes to fix that
-    private Feature selectedFeature = null;
-    private Overlay selectedOverlay = null;
+    private AbstractConfigurable selected = null;
```
magicus: Overall this looks good, but you should use the `Configurable` interface to define the type everywhere (except when it is actually instantiated).

magicus (approved):

PR #1271, opened 2023-03-05
feat: Add HeldItemTypeFunction

kristofbolyai:


common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -178,4 +178,28 @@ public List<String> getAliases() {
             return List.of("max_held_durability");
         }
     }
+
+    public static class HeldItemTypeFunction extends Function<String> {
```
magicus, line 182: This is showcasing the WynnItem system, but is it really useful? What is the use case?

I could rather see it more useful to check what actions are possible with the currently held item. Like, if it is a weapon, show it's damage types (I always forget and need to go in and check...), or if it is a consumable, highlighting the number of charges left, etc.

kristofbolyai: Well, crafted held weapon vs non crafted, etc. I think there is value in keeping this.

magicus: Okay, if you say so. It won't hurt.

magicus (approved):

PR #1273, opened 2023-03-05
feat: Overlay Refactor Part 1: Introduce TextOverlay

kristofbolyai:
This PR exists to show what I plan to do with every overlay. Before I start the "heavy" work of creating all of the functions, I want all of us to agree on if this is the way to go.


common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
```diff
@@ -0,0 +1,88 @@
...
+            VerticalAlignment verticalAlignmentOverride) {
+        super(position, size, horizontalAlignmentOverride, verticalAlignmentOverride);
+    }
+
+    @Override
+    public void render(
+            PoseStack poseStack, MultiBufferSource.BufferSource bufferSource, float partialTicks, Window window) {
+        if (!Models.WorldState.onWorld()) return;
+
+        if (System.nanoTime() - lastUpdate > secondsPerRecalculation * 1e+9) {
```
magicus: I think nanoTime is, at least potentially, much more expensive than currentTimeMillis, so if that precision is not needed, its better to use cTM.

magicus: Ok, I see now,  this was just moved. So you can leave it like that for now if you want. But I think it should be changed at some point.

common/src/main/java/com/wynntils/features/overlays/InfoBoxFeature.java
```diff
@@ -98,38 +86,6 @@ protected InfoBoxOverlay(
...
-                                this.getRenderHorizontalAlignment(),
-                                this.getRenderVerticalAlignment(),
-                                this.textShadow);
-
-                renderY += FontRenderer.getInstance().getFont().lineHeight;
-            }
-        }
-
         @Override
         public void renderPreview(
```
magicus: Maybe the preview can move to TextOverlay as well?

magicus:
I think this looks good, but it feels a bit early to tell. The real test is when we start converting something that is already a separate overlay. 

The trick part is, I think, how to handle `content`. For "hard-coded" overlays, it would seem more reasonable to give the content in the constructur, rather than as a `@Config` field. 

And maybe it is really the InfoBoxOverlay that is special, it might be that its the only kind of overlay that really has the expression in the `content` field. So maybe really even here we should have TextOverlay take the content as a constructor, and that can be updated with a setter, and let the InfoBoxOverlay constructor call `super(content)` and also call the super class `setContent(content)` in `onConfigUpdate`.

common/src/main/java/com/wynntils/features/overlays/InfoBoxFeature.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
```diff
@@ -0,0 +1,91 @@
...
+import com.wynntils.utils.render.type.TextShadow;
+import com.wynntils.utils.render.type.VerticalAlignment;
+import net.minecraft.client.renderer.MultiBufferSource;
+
+/**
+ * An overlay, which main purpose is to display function templates.
+ */
+public abstract class TextOverlay extends Overlay {
+    @Config
+    public TextShadow textShadow = TextShadow.OUTLINE;
```
magicus, line 27: I forgot how inherited configs work. Do you need to update i18n for these? How does it work when we have multiple overlays inheriting this?

But I'm guessing at least some i18n changes is needed for this patch to work..?

common/src/main/java/com/wynntils/features/overlays/InfoBoxFeature.java
magicus: Hm. I was more thinking like rather having a `getPreviewTemplate()`, and move the `renderPreview` method to the superclass as well. That way the subclasses only need to provide the template and do not really have to bother with the rendering business at all. They just say: "here is my string, go and make it look nice on the screen".

magicus: (With that said, having the common functionality extracted to `renderTemplate`)

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
kristofbolyai: Fixed it up, but it is only needed for the future.

common/src/main/java/com/wynntils/features/overlays/InfoBoxFeature.java
kristofbolyai: Done

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
```diff
@@ -0,0 +1,101 @@
...
+import com.wynntils.utils.render.type.TextShadow;
+import com.wynntils.utils.render.type.VerticalAlignment;
+import net.minecraft.client.renderer.MultiBufferSource;
+
+/**
+ * An overlay, which main purpose is to display function templates.
+ */
+public abstract class TextOverlay extends Overlay {
+    @Config(key = "overlay.wynntils.textOverlay.textShadow")
+    public TextShadow textShadow = TextShadow.OUTLINE;
```
magicus, line 27: My understanding is that this will toggle shadow for *all* our upcoming text overlay widgets, right? I'm trying to decide if this is a problem or not. (Frankly, I think it has been too many knobs to turn for every individual text setting, so I guess this is not just fine but actually an improvement, but maybe someone will protest..?)

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
```diff
@@ -0,0 +1,101 @@
...
+                    .renderAlignedTextInBox(
+                            poseStack,
+                            bufferSource,
+                            line,
+                            renderX,
+                            renderX + this.getWidth(),
+                            renderY,
+                            renderY + this.getHeight(),
+                            0,
+                            CommonColors.WHITE,
```
magicus, line 86: I assume implementors can select the color using coded strings, so I guess this does not really matter...?

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
kristofbolyai: per overlay

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
kristofbolyai: Yes, I assume

common/src/main/java/com/wynntils/core/features/overlays/TextOverlay.java
kristofbolyai: So basically settings are inherited from the base class, but not shared

magicus: Ok, this sounds good, so I'll accept the PR, but please explain to me then what `key = "overlay.wynntils.textOverlay.textShadow"` means? Aha, that's for the i18n description, not the config save key. Duh. Now I get it. I mixed up the two.

magicus (approved):

PR #1278, opened 2023-03-06
feat: Overlay Refactor Part 2: Port all compatible overlays to TextOverlay

kristofbolyai:
Depends on part 1, #1273


kristofbolyai:
And this should be ready now! I am really happy with how this is turning out, and in the process I made some improvements too.

common/src/main/java/com/wynntils/features/overlays/ShamanTotemTrackingFeature.java
```diff
@@ -78,29 +69,22 @@ public void onTotemSummoned(TotemEvent.Summoned e) {
...
         @Config
-        public TextShadow textShadow = TextShadow.OUTLINE;
+        public static ChatFormatting firstTotemTextColor = ChatFormatting.WHITE;
 
-        private TextRenderSetting textRenderSetting;
+        @Config
+        public static ChatFormatting secondTotemTextColor = ChatFormatting.BLUE;
+
+        @Config
+        public static ChatFormatting thirdTotemTextColor = ChatFormatting.RED;
```
kristofbolyai, line 83: This is the only "regression" we have. This feature uses CustomColor for totem colors, but having multiple CustomColors per TextOverlay is not possible at the moment. I toyed with a few ways to implement it, but I did not like either of the implementations. So, for the time being, this was added. Hopefully, we can remove it sooner or later.

common/src/main/java/com/wynntils/features/overlays/ArrowShieldTrackingFeature.java
```diff
@@ -69,67 +38,22 @@ protected ArrowShieldTrackerOverlay() {
...
-                return;
-            }
-
-            renderTaskCache = new TextRenderTask(getOverlayText(charges), textRenderSetting.withCustomColor(textColor));
-        }
-
-        private static String getOverlayText(int charges) {
-            return "Arrow Shield:" + ARROW_SYMBOL.repeat(charges);
+        public String getTemplate() {
+            return "{IF_STRING(GT(ARROW_SHIELD_COUNT; 0); CONCAT(\"Arrow Shield: \"; REPEAT(\"%s\"; ARROW_SHIELD_COUNT)); \"\")}"
```
magicus: Maybe extract as a constant?

common/src/main/java/com/wynntils/features/overlays/AuraTimerOverlayFeature.java
```diff
@@ -90,42 +67,23 @@ protected AuraTimerOverlay() {
...
-                            this.getRenderY() + this.getHeight(),
-                            0,
-                            textColor,
-                            this.getRenderHorizontalAlignment(),
-                            this.getRenderVerticalAlignment(),
-                            textShadow,
-                            this.getHeight() / 15f);
+        @Override
+        public String getTemplate() {
+            return "{IF_STRING(MTE(AURA_TIMER; 0); CONCAT(CONCAT(\"Aura: : \"; STRING(AURA_TIMER:1)); \"s\"); \"\")}";
```
magicus: Constant here as well. Also, it took me a while to realize what `MTE` is. I suggest we try to use `GT/GTE` which to me feels more standard. (In fact, if it were up to me we would remove MT/MTE, but if you like it I guess we can keep it)

common/src/main/java/com/wynntils/features/overlays/MobTotemTrackingFeature.java
```diff
@@ -63,65 +38,27 @@ protected MobTotemTimerOverlay() {
...
-                            bufferSource,
-                            this.getRenderX(),
-                            this.getRenderY(),
-                            renderTaskCache,
-                            this.getWidth(),
-                            this.getHeight(),
-                            this.getRenderHorizontalAlignment(),
-                            this.getRenderVerticalAlignment());
+        public String getTemplate() {
+            return IntStream.rangeClosed(1, Models.MobTotem.getMobTotems().size())
```
magicus, line 45: A bit sad that we can't express this fully as a wynntils function expression. Oh well, one step at a time.

common/src/main/java/com/wynntils/functions/SpellFunctions.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public static class ShamanMaskFunction extends Function<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            ChatFormatting color = arguments.getArgument("isColored").getBooleanValue()
+                    ? Models.ShamanMask.getCurrentMaskType().getColor()
+                    : ChatFormatting.WHITE;
+
+            if (arguments.getArgument("useShortName").getBooleanValue()) {
```
magicus: I think this would read better as 

String name = useShort ? getAlias : getName;
return color + name;

common/src/main/java/com/wynntils/functions/SpellFunctions.java
```diff
@@ -0,0 +1,139 @@
...
+        public Double getValue(FunctionArguments arguments) {
+            int totemNumber = arguments.getArgument("totemNumber").getIntegerValue();
+
+            ShamanTotem shamanTotem = Models.ShamanTotem.getTotem(totemNumber);
+
+            if (shamanTotem == null) {
+                return 0d;
+            }
+
+            return McUtils.player()
```
magicus, line 129: As a future improvement, it would be cool to have a Location or Coord type, and then you could apply a `distance` function to it. Then we could use that to show distance to compass beam etc as well. And we could introduce stuff like `travel_time`, taking into account distance and current avg bps, etc :-)

common/src/main/java/com/wynntils/functions/WorldFunctions.java
```diff
@@ -57,4 +57,56 @@ public List<String> getAliases() {
             return List.of("world_uptime", "uptime");
         }
     }
+
+    public static class MobTotemCountFunction extends Function<Integer> {
```
magicus, line 61: I don't really think this fits in this file, but I have no better suggestion. We can let them live here for the time being, but I think long-term they need a better home.

magicus (commented):
As a general remark, I am a bit scared for race conflicts with the totem (mob and shaman) functions that take a totem number. Are we sure totems cannot disappear while parsing expressions?

kristofbolyai:
> As a general remark, I am a bit scared for race conflicts with the totem (mob and shaman) functions that take a totem number. Are we sure totems cannot disappear while parsing expressions?

Well, we can't really process packets mid-game, can we? Functions are just an abstract way of doing the same logic (so if this was not race prune before, it should not be now)

common/src/main/java/com/wynntils/models/mobtotem/MobTotem.java
```diff
@@ -35,9 +35,9 @@ public void setTimerString(String timerString) {
         this.timerString = timerString;
     }
 
-    public int getDistanceToPlayer() {
+    public double getDistanceToPlayer() {
         // y() - 4 because the entity is 4 blocks above the ground
-        return (int) Math.sqrt(McUtils.player().distanceToSqr(new Vec3(location.x(), location.y() - 4, location.z())));
+        return Math.sqrt(McUtils.player().distanceToSqr(new Vec3(location.x(), location.y() - 4, location.z())));
```
magicus, line 40: This is old code, but I think there is a `McUtils.player().distanceTo` that does the square root -- the distanceToSqr is just a optimized version if you can work with the squared distance instead (e.g. for sorting distances or checking a spherical boundary).

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
```diff
@@ -74,6 +75,16 @@ public void onWorldChange(WorldStateEvent e) {
     }
 
     public List<MobTotem> getMobTotems() {
-        return mobTotems.values().stream().toList();
+        return mobTotems.values().stream()
```
magicus, line 79: This will reorder mob totems when the player moves around. Is this intentional? (I don't think that is how the old code works, but then again I don't think I've ever seen it with two concurrent mob totems)

common/src/main/java/com/wynntils/models/territories/AuraTimerModel.java
```diff
@@ -0,0 +1,42 @@
...
+package com.wynntils.models.territories;
+
+import com.wynntils.core.components.Model;
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import java.util.List;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+// NOTE: As the package suggests, this is model is related to guild wars' Aura.
+//       Not to be confused with the aura spell.
+public class AuraTimerModel extends Model {
```
magicus: I don't really know what this is, but the title string indicates it could be called "tower aura". Maybe we can rename the class TowerAuraTimerModel?

common/src/main/java/com/wynntils/models/mobtotem/MobTotem.java
kristofbolyai: The method exists, with the wrong signature. It only takes an entity.

common/src/main/java/com/wynntils/features/overlays/MobTotemTrackingFeature.java
kristofbolyai: Yes, I thought the same.

common/src/main/java/com/wynntils/functions/SpellFunctions.java
kristofbolyai: Agree.

common/src/main/java/com/wynntils/functions/WorldFunctions.java
kristofbolyai: Yeah, I too did not know what do with this.

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
kristofbolyai: Oh, yes, this changes stuff. But that case is rare, if not disallowed by Wynn. Anyways, I changed this to `LinkedHashMap` to always have some kind of stable order, what I wanted to achieve with this sort.

common/src/main/java/com/wynntils/models/territories/AuraTimerModel.java
kristofbolyai: It is guild tower aura yes, so I can rename this.

common/src/main/java/com/wynntils/models/mobtotem/MobTotemModel.java
magicus: I was more thinking like if you sort on the totem's creator's name, or its location, then it would be stable independent on the player's current location.

magicus: (Totem owner is probably easiest to sort on, has a well-defined natural order, and each player can only have a single totem spawned at a time)

common/src/main/java/com/wynntils/functions/generic/StringFunctions.java
```diff
@@ -52,8 +52,9 @@ public String getValue(FunctionArguments arguments) {
 
         @Override
         public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
-            return new FunctionArguments.RequiredArgumentBuilder(
-                    List.of(new FunctionArguments.ListArgument<>("values", String.class)));
+            return new FunctionArguments.RequiredArgumentBuilder(List.of(
```
magicus: Merge error?

common/src/main/java/com/wynntils/functions/generic/StringFunctions.java
```diff
@@ -69,8 +70,8 @@ public Boolean getValue(FunctionArguments arguments) {
         @Override
         public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
             return new FunctionArguments.RequiredArgumentBuilder(List.of(
-                    new FunctionArguments.Argument<>("first", String.class, null),
```
magicus: Looks wrong here too

common/src/main/java/com/wynntils/models/territories/TowerAuraTimerModel.java
```diff
@@ -0,0 +1,42 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.territories;
+
+import com.wynntils.core.components.Model;
+import com.wynntils.mc.event.SubtitleSetTextEvent;
+import java.util.List;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+// NOTE: As the package suggests, this is model is related to guild wars' Aura.
```
magicus: With the new name, this comment is superfluous.

magicus (approved):

PR #1281, opened 2023-03-06
feat: Introduce capped functions

magicus:
I had some crash I need to fix, but the general structure is ready for review.


common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -151,6 +152,15 @@ public Double getDoubleValue() {
             return (Double) this.getValue();
         }
 
+        public CappedValue getCappedValue() {
+            if (this.type == Number.class) {
+                int val = ((Number) this.getValue()).intValue();
+                return new CappedValue(val, val);
+            }
```
kristofbolyai: Should this ever happen? This is like checking if an int value is 0 or 1 in getBooleanValue, when the actual value is a string.

common/src/main/java/com/wynntils/models/character/PlayerInventoryModel.java
```diff
@@ -5,23 +5,37 @@
...
 import com.wynntils.utils.wynn.InventoryUtils;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class PlayerInventoryModel extends Model {
-    private int openSlots = 0;
+    private static final int MAX_INVENTORY_SLOTS = 28;
```
kristofbolyai, line 24: Do we count offhand as a slot or how does this work?

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
```diff
@@ -48,7 +50,7 @@ public CombatXpModel(WorldStateModel worldStateModel) {
     }
 
     @SubscribeEvent
-    public void onExperienceGain(CombatXpGainEvent event) {
+    public void onXpGain(CombatXpGainEvent event) {
```
kristofbolyai: Unless you changed all of these names in this PR, I don't think this is generalized

common/src/main/java/com/wynntils/models/items/ItemModel.java
```diff
@@ -102,6 +102,15 @@ public <T extends WynnItem> Optional<T> asWynnItem(ItemStack itemStack, Class<T>
         return Optional.of((T) wynnItem);
     }
 
+    public <T> Optional<T> asWynnItemPropery(ItemStack itemStack, Class<T> clazz) {
```
kristofbolyai, line 105: I missed this :) Great addition. We should use this over the old version in all places possible (as a follow up PR)

common/src/main/java/com/wynntils/models/items/ItemModel.java
```diff
@@ -102,6 +102,15 @@ public <T extends WynnItem> Optional<T> asWynnItem(ItemStack itemStack, Class<T>
...
+        Optional<ItemAnnotation> annotationOpt = ItemHandler.getItemStackAnnotation(itemStack);
+        if (annotationOpt.isEmpty()) return Optional.empty();
+        if (!(annotationOpt.get() instanceof WynnItem wynnItem)) return Optional.empty();
+        if (wynnItem.getClass().isAssignableFrom(clazz)) return Optional.empty();
```
kristofbolyai: Shouldn't this be the other way around? Is this the reason for your crashes :)?

kristofbolyai: (technically the cast below is the one crashing but you see the point...)

common/src/main/java/com/wynntils/models/players/hades/HadesModel.java
```diff
@@ -201,38 +201,28 @@ public void onTick(TickEvent event) {
...
-                            Models.Character.getMaxHealth(),
-                            Models.Character.getCurrentMana(),
-                            Models.Character.getMaxMana())) {
-                tickCountUntilUpdate = 1;
-                return;
+            PlayerStatus newStatus =
+                    new PlayerStatus(pX, pY, pZ, Models.Character.getHealth(), Models.Character.getMana());
+
+            if (lastSentStatus != null) {
+                if (lastSentStatus.equals(newStatus)) {
```
kristofbolyai: Objects.equals would be better here (I know I did this..), so I am not forcing you to fix this.

common/src/main/java/com/wynntils/utils/type/CappedValue.java
```diff
@@ -7,12 +7,41 @@
 public record CappedValue(int current, int max) {
     public static final CappedValue EMPTY = new CappedValue(0, 0);
 
-    public int getPercentage() {
-        return Math.round((float) current / max * 100.0f);
+    public CappedValue withCurrent(int newCurrent) {
+        return new CappedValue(newCurrent, max);
+    }
+
+    /** Return true iff the current value is equal to the max value */
```
kristofbolyai, line 14:
```suggestion
    /** Return true if the current value is equal to the max value */
```

kristofbolyai (commented):
Few notes I took while reviewing:
- We should check with asserts if we have any aliases/names that are duplicated
- This PR kind of happened to be a mixed bag of fixes alongside functions, but that is needed because of the CappedValue changes in Models

Overall this is looking better. 

One thing I would like to confirm/change though. How do CappedValues look like, when evaluated and printed? `[current/max]`? If so, I would like to change that to `CappedValue[current/max]` or something shorter, but to signify the class.

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
magicus: I don't know, I tried to copy the pattern. I'll remove it then...

common/src/main/java/com/wynntils/models/character/PlayerInventoryModel.java
magicus: No, this is pure inventory slots. Same code basically as before, only the `28` was hard-coded in the calculations. When the current slots == max slots, then you cannot pick up any more items.

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
magicus: Hm, what do you mean? This is really an unrelated fix. I think we should use "xp" for "xp", not "experience", and have been trying to enforce that everywhere. I just noticed that the callback function (where the name really is not used) was not following the rules.

common/src/main/java/com/wynntils/models/items/ItemModel.java
magicus: I fixed it everywhere, actually, as part of this PR. Maybe a bit messy, but, well... 🤷

common/src/main/java/com/wynntils/models/items/ItemModel.java
magicus: Yes, indeed, this was the problem :)

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
kristofbolyai: I meant that I don't think all methods use Xp instead of Experience. But we don't need to fix it now.

common/src/main/java/com/wynntils/models/players/hades/HadesModel.java
magicus: In what way? I think that records per definition implicitly implements an equals method that is identical to Objects.equals.

kristofbolyai: We can get rid of the nested ifs - null check that way

common/src/main/java/com/wynntils/utils/type/CappedValue.java
magicus: Actually, that is not a typo. It's a shorthand for "if and only if". https://en.wikipedia.org/wiki/If_and_only_if

kristofbolyai: Never seen that :) Good to know.

magicus:
> One thing I would like to confirm/change though. How do CappedValues look like, when evaluated and printed? `[current/max]`? If so, I would like to change that to `CappedValue[current/max]` or something shorter, but to signify the class.

Currently what is printed is  `[current/max]`, but my plan is to do the opposite of what you are suggesting -- changing it to `current/max`. That is the common way this kind of relations are expressed, in e.g. objective left, etc. The point is that if you enter just `{mana}`, you should get something that makes sense in and of itself. If I were to prefix `CappedValue`, then we could *never* print just such a value since it would look butt ugly.

kristofbolyai:
> > One thing I would like to confirm/change though. How do CappedValues look like, when evaluated and printed? `[current/max]`? If so, I would like to change that to `CappedValue[current/max]` or something shorter, but to signify the class.
> 
> Currently what is printed is `[current/max]`, but my plan is to do the opposite of what you are suggesting -- changing it to `current/max`. That is the common way this kind of relations are expressed, in e.g. objective left, etc. The point is that if you enter just `{mana}`, you should get something that makes sense in and of itself. If I were to prefix `CappedValue`, then we could _never_ print just such a value since it would look butt ugly.

Looking at it this way, it sounds good. I did not think this fully through.

magicus:
I realize this PR became a bit messy. Do you think I should split it up into separate parts to get a clearer commit history?

kristofbolyai:
> I realize this PR became a bit messy. Do you think I should split it up into separate parts to get a clearer commit history?

I mostly got through it, I don't think there is need to do extra work for this.

common/src/main/java/com/wynntils/models/experience/CombatXpModel.java
magicus: I checked, and at least in CombatXpModel all use XP instead of experience.

common/src/main/java/com/wynntils/models/players/hades/HadesModel.java
magicus: I see. I did that by just swapping the order of the comparands.

kristofbolyai (approved):

PR #1282, opened 2023-03-06
fix: Fix Integer-Double constant parsing order in functions

kristofbolyai:


kristofbolyai:
Force merging as this is a trivial, but important bugfix.

PR #1283, opened 2023-03-06
feat: Introduce ListArgument type to all functions that can make use of it

kristofbolyai:


common/src/main/java/com/wynntils/functions/generic/MathFunctions.java
```diff
@@ -12,30 +12,35 @@ public final class MathFunctions {
...
         public Double getValue(FunctionArguments arguments) {
-            return arguments.getArgument("first").getDoubleValue()
-                    - arguments.getArgument("second").getDoubleValue();
+            List<Number> values = arguments.getArgument("values").asList().getValues();
+
+            if (values.size() == 1) {
+                return values.get(0).doubleValue();
+            }
+
+            return values.get(0).doubleValue()
```
magicus: So you are saying A - B - C - D should be interpreted as A - (B + C + D)? It could just as well have been A - (B - (C - D))). Since subtraction is not commutative, it is not clear how to interpret it with multiple operands; this is typically not done and I recommend against it.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -863,16 +863,13 @@
...
   "function.wynntils.generic.add.description": "Adds two numbers together",
   "function.wynntils.generic.add.name": "Add",
-  "function.wynntils.generic.and.argument.first": "First boolean",
-  "function.wynntils.generic.and.argument.second": "Second boolean",
+  "function.wynntils.generic.and.argument.values": "Values to and together",
   "function.wynntils.generic.and.description": "Checks if both conditions are true",
   "function.wynntils.generic.and.name": "And",
-  "function.wynntils.generic.concat.argument.first": "First string",
-  "function.wynntils.generic.concat.argument.second": "Second string",
+  "function.wynntils.generic.concat.argument.values": "Values to concatenate",
```
```suggestion
  "function.wynntils.generic.concat.argument.values": "Strings to concatenate",
```

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -33,13 +34,42 @@ protected Builder(List<Argument> arguments) {
         }
 
         public ErrorOr<FunctionArguments> buildWithValues(List<Object> values) {
-            if (values.size() != this.arguments.size()) {
+            if (arguments.stream()
+                            .filter(argument -> argument instanceof ListArgument<?>)
+                            .count()
```
magicus, line 39: Not that it really matters, but I think you can use like a findFirst or anyMatch or something like that, to short-circuit as soon as you find a ListArgument, instead of first counting up all of them. Not that it is likely to ever be more than one list argument here :-) but I think it generally makes a better pattern to try to follow. You can change this or skip this suggestion, it's up to you.

magicus: Ah, right, you want it to be stricter greater than 1. Of course. Forget what I said.

kristofbolyai: But the point is to check if there is more than two? I could get the first ones index and compare it to the size, depending on if the filter is going from 0..n in the list, but this is just a bit more readable

common/src/main/java/com/wynntils/core/functions/arguments/FunctionArguments.java
```diff
@@ -154,5 +188,23 @@ public Double getDoubleValue() {
...
+            return (ListArgument<U>) this;
+        }
+    }
+
+    public static class ListArgument<T> extends Argument<T> {
+        public ListArgument(String name, Class<T> type) {
+            super(name, type, null);
+        }
+
+        public void setValues(List<Object> values) {
```
magicus, line 202:
```suggestion
        public void setValues(List<T> values) {
```
?

kristofbolyai: setValue itself uses object too, i followed that

magicus: Or maybe not, I see that you don't know formally the type. Nevermind me, I'm just throwing bad suggestions around me tonight. :)

magicus (approved):

magicus:
And I realized it is not commutativity that is the problem with list arguments, but associativity. "concat" is not commutative, "foo"+"bar" is not the same as "bar"+"foo", but it is associative: ("foo" + "bar") + "baz" is the same as "foo" + ("bar" + "baz").

Just be formally correct :-)

PR #1296, opened 2023-03-07
refactor: Change renderButton to renderWidget in our WynntilsButton

magicus:
Change renderButton to renderWidget in our WynntilsButton, to minimize difference with 1.19.4


kristofbolyai (approved):
Looks good, If this looks good ingame

PR #1297, opened 2023-03-07
feat: Overlay Refactor Part 3: Dynamic overlays

kristofbolyai:


kristofbolyai:
This should be working now. GUI will follow, advanced users who take use of these can/and should use a simple command. 

I know the structure is not the greatest; I am actually quite fed up with how our Feature/Overlay managers (and the base classes) work (and obviously ConfigManager, as these three cannot be separated). I think I am going to refactor them into something hopefully much nicer.

I did not add any upfixers as it was not needed. I tested, and this works well with old configs.

And finally, I made the info box 7 coordinates it's own overlay, as it should be. Now all 7 info boxes are usable. I had to make 7 the default  value to preserve compatibility.

kristofbolyai:
Note: This is not to be merged. Needs an announcement.

common/src/main/java/com/wynntils/features/overlays/InfoBoxFeature.java
```diff
@@ -8,77 +8,52 @@
...
-            "{x:0} {y:0} {z:0}",
-            new OverlayPosition(
-                    160, 20, VerticalAlignment.Top, HorizontalAlignment.Left, OverlayPosition.AnchorSection.TopLeft),
-            HorizontalAlignment.Center,
-            VerticalAlignment.Middle,
-            0);
+    @OverlayGroup(instances = 7, renderType = RenderEvent.ElementType.GUI, renderAt = RenderState.Pre)
+    private final List<InfoBoxOverlay> infoBoxOverlays = new ArrayList<>();
+
+    //    @OverlayInfo(renderType = RenderEvent.ElementType.GUI, renderAt = OverlayInfo.RenderState.Pre)
```
magicus: Maybe remove this commented-out code?

common/src/main/java/com/wynntils/features/overlays/OverlayGroup.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.overlays;
```
magicus: I think this should live in core.overlay.annotations?

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
```diff
@@ -0,0 +1,96 @@
...
+                new GuiScaledOverlaySize(100f, 20f),
+                HorizontalAlignment.Center,
+                VerticalAlignment.Middle);
+        this.id = id;
+    }
+
+    @Override
+    public String getTranslatedName() {
+        return I18n.get(
+                "feature.wynntils." + getDeclaringFeatureNameCamelCase() + ".overlay." + getNameCamelCase() + ".name",
```
magicus, line 55: What does the `,id` achieve?

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
```diff
@@ -0,0 +1,96 @@
...
+                id);
+    }
+
+    @Override
+    public String getConfigJsonName() {
+        return super.getConfigJsonName() + id;
```
magicus, line 61:
```suggestion
        return super.getConfigJsonName() + "." + id;
```
maybe?

kristofbolyai: I think `"infoBoxFeature.infoBoxOverlay9.userEnabled": true` is better than `"infoBoxFeature.infoBoxOverlay.9.userEnabled": true`. (and this way we also preserve compatibility)

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
kristofbolyai: The I18n strings are format strings, the en_us file has `Info Box %d`. meaning `Info Box %d`.formatted(id) will be run to display the overlay name.

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -72,6 +80,46 @@ public final void initOverlays() {
         }
     }
 
+    public final void initOverlayGroups() {
+        // Unregister old instances
+        for (Overlay groupedOverlay : groupedOverlayInstances) {
```
magicus, line 86: How can this happen?

kristofbolyai: We recreate all overlay instances, even if we only added a new one. But this can also happen if we remove an overlay.

common/src/main/java/com/wynntils/commands/ConfigCommand.java
```diff
@@ -192,6 +235,81 @@ private LiteralCommandNode<CommandSourceStack> buildReloadConfigNode() {
...
+        return overlayGroupArgBuilder.build();
+    }
+
+    private int addOverlayGroup(CommandContext<CommandSourceStack> context) {
+        String featureName = context.getArgument("feature", String.class);
+        String groupName = context.getArgument("group", String.class);
+
+        OverlayGroupHolder overlayGroupHolder = getOverlayGroupHolderFromArguments(context, featureName, groupName);
+
+        if (overlayGroupHolder == null) return 0;
```
magicus, line 260: Error message?

kristofbolyai: Handled in the getter like every other method here.

common/src/main/java/com/wynntils/commands/ConfigCommand.java
```diff
@@ -607,6 +725,24 @@ private Overlay getOverlayFromArguments(
         return overlayOptional.get();
     }
 
+    private static OverlayGroupHolder getOverlayGroupHolderFromArguments(
+            CommandContext<CommandSourceStack> context, String featureName, String overlayGroupName) {
+        Feature feature = getFeatureFromArguments(context, featureName);
+        if (feature == null) return null;
```
magicus, line 730: Ok, my bad, the error message should go here, not where you called this function

common/src/main/java/com/wynntils/commands/ConfigCommand.java
```diff
@@ -192,6 +235,81 @@ private LiteralCommandNode<CommandSourceStack> buildReloadConfigNode() {
...
+
+        OverlayGroupHolder overlayGroupHolder = getOverlayGroupHolderFromArguments(context, featureName, groupName);
+
+        if (overlayGroupHolder == null) return 0;
+
+        int id = Integer.parseInt(idName);
+
+        overlayGroupHolder.removeId(id);
+
+        Managers.Config.loadConfigOptions(true, false);
```
magicus, line 295: Is this part of the load/save dance you described elsewhere?

kristofbolyai: Yes, I put it at `loadConfigOptions` method, that was the most generalized place for it.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -66,18 +74,44 @@ public ConfigManager(ConfigUpfixerManager upfixer) {
...
     }
 
+    private void registerOverlayGroups(Feature feature) {
+        List<OverlayGroupHolder> holders = Stream.of(feature.getClass().getDeclaredFields())
+                .filter(f -> f.isAnnotationPresent(OverlayGroup.class))
+                .map(field -> new OverlayGroupHolder(
+                        field, feature, field.getAnnotation(OverlayGroup.class).instances()))
+                .toList();
+
+        OVERLAY_GROUP_FIELDS.addAll(holders);
```
magicus, line 79: I think a short-circuiting "if holders is empty return" would be nice; it does not change semantics but makes the reader realize we're done here if there are no holders.

kristofbolyai: Done

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
magicus: Ah, of course. I knew about the format strings but mistakenly thought that there should be a `%d` in the i18n key. I'm probably too tired to review this properly right now...

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
magicus: I'm not sure it's really better but I am swayed by the compatibility argument.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -212,15 +221,13 @@ public Set<Overlay> getOverlays() {
...
-
     public Feature getOverlayParent(Overlay overlay) {
         return overlayParent.get(overlay);
     }
 
     public boolean isEnabled(Overlay overlay) {
         return enabledOverlays.contains(overlay);
     }
+
+    private record OverlayRenderInfo(RenderEvent.ElementType elementType, RenderState renderState) {}
```
magicus, line 232: This is like a super-class to OverlayGroup and OverlayInfo? Maybe we should express it as a such? Or annotations cannot have a hierarchy?

kristofbolyai: This is a "data" class to hold info about overlay render info. Annotations cannot have custom class fields not have hierarchy. 

I just mentioned it in discord at the same time you reviewed this :) : `the overlay annotations are a problem too. I think I might remove those too`

common/src/main/java/com/wynntils/core/features/overlays/annotations/OverlayGroup.java
```diff
@@ -0,0 +1,23 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface OverlayGroup {
+    // How many instances of this overlay should exist by default
+    int instances() default 1;
```
magicus, line 18: In fact, what would happen if you were to merge OverlayGroup and OverlayInfo? Moving the `instances()` method into OverlayInfo, and perhaps add a `boolean isGroup() default no`?

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
```diff
@@ -74,7 +74,7 @@ public static Screen create() {
 
     @Override
     protected void doInit() {
-        reloadConfigButtons();
+        reloadConfigurableButtons();
```
magicus, line 78: I don't understand what this change does, neither the `selected` stuff below? Is it related to the core fix here, or just some extra stuff you fix when you discovered it?

common/src/main/java/com/wynntils/core/features/overlays/annotations/OverlayGroup.java
kristofbolyai: Well, we would have to change reflections, and save having a class. I think I am going to refactor this either way, so I can merge this now, and remove it later, or just keep it as is, and remove it later.

common/src/main/java/com/wynntils/core/config/OverlayGroupHolder.java
```diff
@@ -0,0 +1,118 @@
...
+        return field.getName();
+    }
+
+    public Class<?> getOverlayClass() {
+        return overlayClass;
+    }
+
+    public void initGroup(List<Integer> ids) {
+        try {
+            List overlays = ReflectionUtils.createListWithType(overlayClass);
```
magicus: The `overlays` List is untyped..? So why the ReflectionUtils wrapper? I assume this is something to do with casting for correct generics?

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
magicus: Ok. I can see the need for some refactoring here, but I guess it'll have to do for now.

common/src/main/java/com/wynntils/core/features/overlays/annotations/OverlayGroup.java
magicus: Ok, let's keep it this way and refactor later.

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
kristofbolyai: This is basically a fix which is needed here, but I did on main too (and I was a bit wrong). The point here is that when we open the settings screen, we should try to preserve the last selected configurable (as this screen is kind of a singleton).

`reloadConfigButtons` is called in `setSelected` which is called just below. We have to filter the new selected configurable because instances might have changed. 

So basically I only added new stuff here, Github diff is just a bit weird.

kristofbolyai: Although I did discover a bug here.. I am fixing it.

common/src/main/java/com/wynntils/commands/ConfigCommand.java
magicus: That's fine, I was just checking so I understood correctly.

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
magicus: Ah, ok, it is about getting back to where we were before in the settings screen. I think I get it now.

common/src/main/java/com/wynntils/core/config/OverlayGroupHolder.java
kristofbolyai: Fixed it now. This is not even needed technically, I think, because of type-erasure.

magicus: Yes, I don't think this does anything. With the cast in place, you can safely replace this call with just a new ArrayList().

magicus (approved):

PR #1303, opened 2023-03-08
feat: Update to Minecraft 1.19.4

magicus:
This is not intended for pushing, at least not now. Let's have a discussion after 1.19.4 is released. But in the long run, these changes are inevitable, since they will also be included in 1.20. 

But for now, I intend to keep the 1.19.4 branch as a long-running 1.19.4 port of main. This PR is mostly to showcase the changes that has gone into 1.19.4; we can close it when it does not serve any purpose longer.


magicus:
This branch is now working properly for 1.19.4 for Fabric, Quilt and Forge. Forge loader is still in beta and displays an angry message to that effect on the main screen. 

I think we should initiate a discussion with the other Wynncraft mods about a plan for switching to 1.19.4. Perhaps also check with our user base in Discord. Since this is the basis for 1.20, it is not so much a question of "if" but of "when" we do the transition.

kristofbolyai (approved):
Looks good to me!

magicus:
We need to prepare an announcement, and sync up with VoW, but then we can push this whenever seems good.

kristofbolyai:
> We need to prepare an announcement, and sync up with VoW, but then we can push this whenever seems good.

https://github.com/Team-VoW/WynncraftVoiceProject/pull/11

PR #1304, opened 2023-03-08
feat: Show overlay for Seaskipper destinations and prices

magicus:


kristofbolyai (approved):
I hope we add the map soon :)

magicus:
Personally, I never liked the map. It was much more trouble scrolling around trying to locate your destination than clicking on an icon. And if you were to start more zoomed out with the map (this would have been better), then the smallest destinations became hard to click.

PR #1306, opened 2023-03-09
feat: Introduce proper persistent data storage

magicus:
I have only converted a single instance as an example. Converting all "invisible" configs etc will be done as a follow-up.


magicus:
Functionality that became shared between config and storage moved to a new JsonManager. In this process, ConfigManager got a little bit more love. Most noticeable, config files are now stored in sorted order.

common/src/main/java/com/wynntils/core/json/JsonManager.java
```diff
@@ -0,0 +1,131 @@
...
+        return GSON.fromJson(GSON.toJson(value), fieldType);
+    }
+
+    public void savePreciousJson(File jsonFile, JsonObject jsonObject) {
+        // create file if necessary
+        if (!jsonFile.exists()) {
+            FileUtils.createNewFile(jsonFile);
+        }
+
+        // FIXME: Make backup first!
```
kristofbolyai: Did we lose the ability to backup our old config if something went wrong? Is this a future improvement or a thing  you wanted to do in this PR?

common/src/main/java/com/wynntils/core/json/JsonManager.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        return null;
+    }
+
+    public Object deepCopy(Object value, Type fieldType) {
+        return GSON.fromJson(GSON.toJson(value), fieldType);
+    }
+
+    public void savePreciousJson(File jsonFile, JsonObject jsonObject) {
```
kristofbolyai, line 71: I am sure there is a reason for the name, but I have not figured it out (yet)

common/src/main/java/com/wynntils/core/json/JsonManager.java
kristofbolyai: I now read the code below. I am not fully sure what this comment means though. I assume we want to not corrupt our config and/or have a way to revert if something goes wrong here? (in which case I am not sure I can think of a legit case of happening)

common/src/main/java/com/wynntils/core/storage/StorageManager.java
```diff
@@ -0,0 +1,119 @@
...
+        readFromJson();
+    }
+
+    @SubscribeEvent
+    public void onWynncraftDisconnect(WynncraftConnectionEvent.Disconnected event) {
+        // Always save when disconnecting
+        writeToJson();
+    }
+
+    void persist(Storage<?> storage) {
```
kristofbolyai, line 80: So basically we save our storage every 10 secs, if there needs to be save (so a storage was changed)?

common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
```diff
@@ -35,7 +35,7 @@ public class CommandAliasesFeature extends UserFeature {
             new CommandAlias("guild territory", List.of("gu t", "gu terr", "guild t", "guild terr"))));
 
     @TypeOverride
-    private final Type aliasesType = new TypeToken<List<CommandAlias>>() {}.getType();
+    private final Type aliasesType = new TypeToken<ArrayList<CommandAlias>>() {}.getType();
```
kristofbolyai, line 38: I had a months old branch cleaning these up, never got around finishing other parts of it. Thanks!

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
```diff
@@ -38,7 +39,7 @@ public class ItemFavoriteFeature extends UserFeature {
     public Set<String> favoriteItems = new HashSet<>();
 
     @TypeOverride
-    private final Type favoriteItemsType = new TypeToken<Set<String>>() {}.getType();
+    private final Type favoriteItemsType = new TypeToken<TreeSet<String>>() {}.getType();
```
kristofbolyai, line 41: Why TreeSet (when the type above is HashSet)? Which one needs changing?

kristofbolyai: (I know they can differ technically, but still)

common/src/main/java/com/wynntils/features/inventory/ItemLockFeature.java
```diff
@@ -46,7 +48,7 @@ public class ItemLockFeature extends UserFeature {
     private final Map<String, Set<Integer>> classSlotLockMap = new HashMap<>();
 
     @TypeOverride
-    private final Type classSlotLockMapType = new TypeToken<HashMap<String, Set<Integer>>>() {}.getType();
+    private final Type classSlotLockMapType = new TypeToken<TreeMap<String, TreeSet<Integer>>>() {}.getType();
```
kristofbolyai, line 49: Same question about types as above

kristofbolyai (commented):
I could not see any upfixers related to moving dryCount and dryStreak to storage? Is that intentional? Do upfixers even work with storage..?

common/src/main/java/com/wynntils/core/json/JsonManager.java
magicus: It is a future improvement. I wanted to do that as part of this PR but forgot about it. It's rather trivial (i just want to rename our old json file to .bak before writing a new one), so I can fix it now.

common/src/main/java/com/wynntils/core/json/JsonManager.java
magicus: The code is more complicated than just saving a straightforward file. If the file is corrupted we store it as an `invalid` file.

magicus: (The name was inspired not by Lord of the Ring, but how you can mark CI artifacts as "precious" and make them persist past normal retention rules.)

common/src/main/java/com/wynntils/core/storage/StorageManager.java
magicus: Yep. Or to be more precise, we never store more often than every tenth second. If no change has been made for more than 10 seconds, the data is stored immediately (or, more or less immediately, it is done in a separate thread, to not cause lag spikes on the rendering thread). But if data has been saved less than ten seconds ago, we schedule a future write 10 seconds after the last.

So if we write value A, and then 8 seconds later get a request to write B, we will wait 2 second. If a request to write value C comes in after an additional 1 seconds, it will see that a write is scheduled and is happy with that, so what we get is:
T = 0: Write A
T = 10: Write B and C

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
magicus: We need TreeSet for persisting to json. Gson treats sets as json arrays. If you have a SortedSet, then the elements of the array will be sorted, otherwise they will appear in an arbitrary order in the json file. 

Internally, we don't care if the set is sorted or not, so we can use HashSet just as well, it does not really matter. If you think it looks confusing, I can update the internal type as well to be TreeSet.

magicus: This is all really not related to the data storage, but was a result of me getting annoyed that the config json output were not stable.

magicus: Actually, this was one of the hacks that should be resolved by the storage impl. I decided to handle it properly as well.

magicus:
No, upfixers do not currently work with storage. In this case, it would mean moving data between config and persistent storage. That is really a unique one-time event happening now since storage was not available. We can announce that users will be scr*wed, but I will not spend time doing a converter.

If we ever want to change data storage format in the future, then I am prepared to port the upfixer to storage. But I'll save that for when that day arrives.

common/src/main/java/com/wynntils/features/inventory/ItemLockFeature.java
magicus: Fixed

magicus:
Also, to be clear, since we will start mucking about with users data, this should be announced too before integration.

kristofbolyai (approved):
LGTM. Let's integrate this with the overaly refactor and share the announcement?

PR #1307, opened 2023-03-10
feat: Replace config with storage

magicus:
This fixes stuff that definitely is/should be Storage instead of HiddenConfig.


kristofbolyai:
This needs to be announced, @magicus. I don't think we should auto-merge.

kristofbolyai (approved):
Can you do the announcement (about nuking dry streaks)? If not, let me know :+1:

magicus:
I am busy tonight so please do the announcement if you can. You can merge the PR when you think it is suitable.

PR #1308, opened 2023-03-10
refactor: Refactor ChatTabManager, in preparation for Storage

magicus:
We need ChatTabManager to take on responsibility for owning the chat tab, since it will soon do. It also needs to wrap any write access, since that will need to be persisted.


kristofbolyai:
So we are moving chat tabs to storage? I think the two of us has a bit of a different idea on what Storage vs Config is. Let's discuss this further on Discord, whenever we can.

magicus:
Do you agree that this refactor is useful, regardless of how Storage/Config ends up?

kristofbolyai (approved):

PR #1313, opened 2023-03-10
feat: Optional display of zero values in Emerald Displays

Rafii2198:
Changes the feature as well as the function.
What it means is that when enabled, it will show you that you have 0 currency, for example `2LE 0EB 34E`. Currently, the only option was to completely skip 0 rendering, meaning the same amount would appear as `2LE 34E`. Everything is optional and by default everything stays the same as was.

I also wanted to include this with the Textured display, but I don't know how to do it properly. 😅


common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -46,13 +46,13 @@ public EmeraldModel(ItemModel itemModel) {
         super(List.of(itemModel));
     }
 
-    public String getFormattedString(int emeralds) {
+    public String getFormattedString(int emeralds, boolean zeros) {
```
kristofbolyai:
```suggestion
    public String getFormattedString(int emeralds, boolean appendZeros) {
```

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -43,6 +43,9 @@ public class InventoryEmeraldCountFeature extends UserFeature {
     @Config
     public boolean showContainerEmeraldCount = true;
 
+    @Config
+    public boolean showZerosInEmeraldText = false;
```
kristofbolyai: Generalize this config value for the rendered version too (at least in name and i18n)

kristofbolyai (changes requested):
You can easily make the rendered version work like this too by editing `renderTexturedCount` line 138

kristofbolyai (approved):

PR #1314, opened 2023-03-11
refactor: Clean up Feature, FeatureManager, related managers and overlay groups

kristofbolyai:
Actions taken:
- Remove any state related code from `Feature`, move to `FeatureManager` instead.
- Move any init related code to `FeatureManager`. Namely `enable`, `disable`, `initOverlays`, `initOverlayGroups`.
- Replace `onEnable`, `onDisable`, `postEnable` with events
- Remove the condition system. It seemed a good idea, but now at 100 features, we use it in a single place.
- Remove `setupEventListener`. There is no use of it.
- Move keybind initialization code to it's proper place.
- General clean up related overlay groups, and all overlay code is now contained in it's manager
- `FeatureManager#initializeFeature` is now bearable to look at
- Some slight performance improvements:
  - Overlay groups are only discovered once with reflection; instead of on every config reload
  - Keybinds are loaded once, instead of after every! feature registration


common/src/main/java/com/wynntils/core/components/Managers.java
```diff
@@ -43,12 +43,13 @@ public final class Managers {
     public static final TranslationManager Translation = new TranslationManager();
 
     // Managers with dependencies, ordered alphabetically as far as possible
-    public static final ConfigManager Config = new ConfigManager(ConfigUpfixer, Json);
```
magicus, line 46: Not that it matters much, but as the comment says I have tried to keep the managers sorted alphabetically, if possible. In this case, I think StorageManager could stay below SplashManager, since no dependency change require a shuffle.

kristofbolyai: I have kind of wanted config and storage to be as close as possible. But I can change it

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -37,61 +37,51 @@ public final class ConfigManager extends Manager {
...
-                .toList();
+        // Finish off the config init process
 
-        if (holders.isEmpty()) return;
+        // Load configs for all features
+        Managers.Config.reloadConfiguration();
 
-        OVERLAY_GROUP_FIELDS.addAll(holders);
-        feature.addOverlayGroups(holders);
+        // Save config file after loading all configurables' options
```
magicus, line 70: Is it still needed to both load and then save configs directly thereafter? How come?

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -37,61 +37,51 @@ public final class ConfigManager extends Manager {
...
-            holder.initGroup(
-                    IntStream.rangeClosed(1, holder.getDefaultCount()).boxed().toList());
-
-            List<ConfigHolder> overlayHolders = holder.getOverlays().stream()
-                    .map(this::getConfigOptions)
-                    .flatMap(List::stream)
-                    .toList();
-
-            holder.getOverlays().forEach(overlay -> overlay.addConfigOptions(this.getConfigOptions(overlay)));
+        // Create default config file containing all configurables' options
```
magicus, line 73: I'm honestly not sure what the default config file does anymore. I think I/we had some kind of idea about this, but it kind of never panned out, right?

And if we do want to keep it, don't you have to save the default file from the Configs as they are before you load in the users config file?

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: Nothing directly changed about overlay group config loading, it was just a cleanup. The config cleanup will be separate, and I have yet to fully figure it out. But basically this code was moved here from featuremanager, as we now load configs after feature manager.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -37,61 +37,51 @@ public final class ConfigManager extends Manager {
...
+    }
 
-            OVERLAY_GROUP_CONFIG_HOLDERS.addAll(overlayHolders);
+    public void registerFeature(Feature feature) {
+        for (Overlay overlay : Managers.Overlay.getFeatureOverlays(feature).stream()
+                .filter(overlay -> !overlay.isDynamic())
+                .toList()) {
+            registerConfigOptions(overlay);
         }
+
```
magicus: This is perhaps a bit irrational, but I think it would make more sense to start with the registration of the feature itself, and then do the overlays, perhaps with a comment `// also register the overlays of the feature` or so. (I know this is the same order as it was before, but I was always like flummoxed by that function as well before).

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -169,13 +152,22 @@ public void loadConfigOptions(boolean resetIfNotFound, boolean initOverlayGroups
         }
 
         // Newly created group overlays need to be enabled
-        for (OverlayGroupHolder holder : OVERLAY_GROUP_FIELDS) {
-            holder.getParent().enableOverlays();
+        for (OverlayGroupHolder holder : Managers.Overlay.getOverlayGroups()) {
```
magicus, line 155: Is it really the responsibility of ConfigManager to enable overlays? Or does `enableOverlay` just do config-related stuff and has a misleading name?

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -169,13 +152,22 @@ public void loadConfigOptions(boolean resetIfNotFound, boolean initOverlayGroups
...
+            Managers.Overlay.enableOverlays(holder.getParent());
         }
     }
 
     private static List<ConfigHolder> getConfigHolderList() {
-        return Stream.concat(CONFIG_HOLDERS.stream(), OVERLAY_GROUP_CONFIG_HOLDERS.stream())
+        // FIXME: This breaks the concept of "manager holds all config holders at all times". Instead we get the group
```
magicus: This is a FIXME since you indent to address this in a follow up? Or do you really accept this design but want to point out that it is not ideal? If so, you can keep the comment but remove the FIXME.

common/src/main/java/com/wynntils/core/features/DebugFeature.java
```diff
@@ -4,20 +4,8 @@
...
-    }
-
-    public static class DevelopmentCondition extends Condition {
-        @Override
-        public void init() {
-            setSatisfied(WynntilsMod.isDevelopmentEnvironment());
-        }
-    }
-}
+@StartDisabled
```
magicus, line 10: We can "inline" this class by just pushing @StartDisabled to the DebugFeature classes. I'm guessing you are saving that change to do separately? (I can do it as well)

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -34,119 +19,19 @@
...
-        return overlays.get(overlay);
+    public void setCategory(Category category) {
+        this.category = category;
     }
 
-    public final void enableOverlays() {
-        Managers.Overlay.enableOverlays(this.getOverlays(), false);
+    /** Whether a feature is enabled */
+    public final boolean isEnabled() {
+        return Managers.Feature.getFeatureState(this) == FeatureState.ENABLED;
```
magicus: I'm thinking we should even have `Managers.Feature.isEnabled(this)`. If a feature is enabled is perhaps not as easy as looking at a state, I guess it can be "user enabled" but "crashed" which will in effect make it "disabled". Or did we chose to encode crashness in the state? In any case, I think a `isEnabled` would be nice and clear.

Edit: I'm confused, the state is a new thing in this PR. Still, I vote for a convenience isEnabled.

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -273,7 +263,7 @@ private void registerFeature(Feature feature) {
     private void initializeFeature(Feature feature) {
         Class<? extends Feature> featureClass = feature.getClass();
 
-        // instance field
+        // Instance field
         try {
             Field instanceField = FieldUtils.getDeclaredField(featureClass, "INSTANCE", true);
```
magicus, line 276: Just a note: I'd like to get rid of this in the future. If we really *do* need to get access to a Feature object, we should be able to get it through something like `Managers.Features.getInstance(SoulPointTimerFeature.class)`. But hopefully we can get rid of accessing feature instances altogether if we fix our config/storage story.

kristofbolyai: Agreed.

common/src/main/java/com/wynntils/core/features/Feature.java
kristofbolyai: The state existed long before the refactor, it was moved. I will add the method.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: I intend to fix this up shortly, but I removed the FIXME anyways.

common/src/main/java/com/wynntils/core/features/DebugFeature.java
kristofbolyai: I did not want to do any changes to the features itself, just the backend.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: This does not enable overlays, rather reads the config file and creates the neccessary amount of instances. This is something I would like to fix in the follow-up, but this needs a lot of thinking.

kristofbolyai: (overall this still improved from what it used to be)

common/src/main/java/com/wynntils/core/config/ConfigManager.java
kristofbolyai: The default config file is a config file, containing all keys, with their default values, as examples. And no, the config holders know the default value, because it is needed for reseting configs. So this is fine here.

common/src/main/java/com/wynntils/core/components/Managers.java
magicus: Ok, keep it. It might that I merge then later on anyway.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: Ah, when you create the config holder from the annotation you read the default value? Clever.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: Well, the `enableOverlays` method seems it does thing that I should describe as "enable the overlays", but I accept that we might have to return here a second time around to fix this better.

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -291,58 +281,101 @@ private void initializeFeature(Feature feature) {
...
+    public void enableFeature(Feature feature) {
+        if (!FEATURES.containsKey(feature)) {
+            throw new IllegalArgumentException("Tried to enable an unregistered feature: " + feature);
+        }
+
+        FeatureState state = FEATURES.get(feature);
+
+        if (state != FeatureState.DISABLED && state != FeatureState.CRASHED) return;
+
+        WynntilsMod.postEvent(new FeatureStateChangeEvent.Enabled(feature));
```
magicus: I think this would, after all, be better as a callback into the feature, `feature.onEnable()` than an event. The critical point here is if anyone is listening to the event for any other reason than to handle its own feature. I'll check the code and come back here.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -45,35 +55,143 @@ public OverlayManager(CrashReportManager crashReportManager) {
...
         overlayInfoMap.remove(overlay);
-        overlayParent.remove(overlay);
         enabledOverlays.remove(overlay);
     }
 
-    public void disableOverlays(List<Overlay> overlays) {
-        enabledOverlays.removeIf(overlays::contains);
-        overlays.forEach(
+    public void disableOverlays(Feature parent) {
+        overlayParentMap.getOrDefault(parent, Collections.emptyList()).forEach(this::disableOverlay);
```
magicus: As a style issue, we have mostly used `List.of()` instead of `Collections.emptyList()`.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -45,35 +55,143 @@ public OverlayManager(CrashReportManager crashReportManager) {
...
-    public void enableOverlays(List<Overlay> overlays, boolean ignoreState) {
-        List<Overlay> enabledOverlays = ignoreState
-                ? overlays
-                : overlays.stream().filter(Overlay::isEnabled).toList();
+    public void enableOverlays(Feature parent) {
+        overlayParentMap.getOrDefault(parent, Collections.emptyList()).forEach(this::enableOverlay);
+    }
+
+    public void enableOverlay(Overlay enableOverlay) {
+        if (!enableOverlay.isEnabled()) return;
```
magicus: This looks suspicious. Did you not mean
```suggestion
        if (enableOverlay.isEnabled()) return;
```
?

magicus: I'm actually feeling this "how can this ever worked?" right now. Has enableOverlay ever been called, and done something useful?

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -45,35 +55,143 @@ public OverlayManager(CrashReportManager crashReportManager) {
...
+
+    public void createOverlayGroupWithIds(OverlayGroupHolder holder, List<Integer> ids) {
+        recreateGroupOverlaysWithIds(holder, ids);
+    }
+
+    public int extendOverlayGroup(OverlayGroupHolder holder) {
+        List<Integer> ids = holder.getOverlays().stream()
+                .map(overlay -> ((DynamicOverlay) overlay).getId())
+                .collect(Collectors.toList());
+        ;
```
magicus:
```suggestion
```

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -45,35 +55,143 @@ public OverlayManager(CrashReportManager crashReportManager) {
...
+
+        int newId = 1;
+        while (ids.contains(newId)) {
```
magicus, line 158: My CS knee-jerk reaction tells me "it's growing quadratically!" but I guess it's fine in this case...

common/src/main/java/com/wynntils/core/keybinds/KeyBindManager.java
```diff
@@ -25,16 +28,34 @@
 import net.minecraft.client.KeyMapping;
 import net.minecraft.client.Options;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.apache.commons.lang3.reflect.FieldUtils;
 
 /** Registers and handles keybinds */
 public final class KeyBindManager extends Manager {
-    private final Set<KeyBind> keyBinds = ConcurrentHashMap.newKeySet();
-    private Map<KeyBind, String> keyBindId = new ConcurrentHashMap<>();
+    private final Set<KeyBind> enbaledKeyBinds = ConcurrentHashMap.newKeySet();
```
magicus:
```suggestion
    private final Set<KeyBind> enabledKeyBinds = ConcurrentHashMap.newKeySet();
```

common/src/main/java/com/wynntils/core/storage/StorageManager.java
```diff
@@ -39,11 +40,18 @@ public final class StorageManager extends Manager {
...
+    public StorageManager(JsonManager jsonManager, FeatureManager feature) {
+        super(List.of(jsonManager, feature));
         userStorageFile = new File(STORAGE_DIR, McUtils.mc().getUser().getUuid() + FILE_SUFFIX);
     }
 
+    public void init() {
+        // Register all storageables
+        Managers.Feature.getFeatures().forEach(this::registerStorageable);
+
+        restorePersisted();
```
magicus: restorePersisted can be made private now.

magicus: ... and in fact just inlined here.

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
```diff
@@ -112,8 +113,8 @@ public void onChatScreenKeyTyped(ChatScreenKeyTypedEvent event) {
         Managers.ChatTab.setFocusedTab(Managers.ChatTab.getNextFocusedTab());
     }
 
-    @Override
-    protected void postEnable() {
+    @SubscribeEvent
+    public void onEnable(FeatureStateChangeEvent.Enabled event) {
         Managers.ChatTab.resetFocusedTab();
```
magicus, line 127: This will reset the focused tab any time any feature whatsoever gets enabled...

common/src/main/java/com/wynntils/core/features/FeatureManager.java
magicus: Ok, back after checking. It is used in two places. One is indeed checking for "this". The other one did not, but should have. 

Conclusion: This should not be an event. The old `onEnable()` and `onDisable()` callbacks on the Feature itself is better suited for this purpose. 

Also, there is no need for an `onCrash()`. Features should not crash, so coding for special handling if it happens is kind of weird. It should be enough that `onDisable()` is called in case of a crash.

magicus:
Overall this is a very nice cleanup! I added some nitpick comments that needs addressing, and the Event thing needs to go back to how it was before, but otherwise this looks really nice!

If you can fix this within a day or two I suggest you push this first, and then I handle the merge conflict with my Config<> wrapper rewrite. (There will be some but hopefully mostly trivial).

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
kristofbolyai: I renamed this to better reflect what it does.

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
kristofbolyai: Fixed with the "revert"

common/src/main/java/com/wynntils/core/features/FeatureManager.java
kristofbolyai: Fixed.

magicus (approved):

PR #1315, opened 2023-03-11
feat: Wynntils Crash Telemetry

magicus:
Sorry for the obnoxious PR title, but I want this to be clearly visible in the change log.

The change itself is rather straightforward. We just send an event when we intercept a crash (unhandled exception) in one of our components.

The only thing that made this slightly messier was that we had no good way of accessing the name of a keybind, so I added support for storing the owning feature + name of the keybind field.

Do not merge this PR until I have announced that it is coming.


HighCrit:
I think such a feature should either be an opt-in, a request on every crash whether or not to send it, or an opt-out with a prompt on first boot to do so.

magicus:
As long as we are alpha, I insist that we should have this opt-out. We will need these reports. Being an alpha tester means you need to actively support bug fixing. 

When we leave alpha, we can reconsider how to handle this. Perhaps ask the user at the very first run? That seem to be a common approach these days.

magicus:
@HighCrit Also, have a look at what kind of data is shared: https://hst.sh/qunibituni There are no personal data in there, just a version, timestamp and a stack dump, no variables, etc.

common/src/main/resources/assets/wynntils/urls.json
```diff
@@ -16,6 +16,14 @@
       "version"
     ]
   },
+  {
```
magicus, line 19: Reminder to self: Also post this on api website.

common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -121,6 +122,8 @@ private static void handleExceptionInEventListener(Throwable t, Event event) {
 
         WynntilsMod.error("Exception in feature " + feature.getTranslatedName(), t);
         WynntilsMod.warn("This feature will be disabled");
+        WynntilsMod.postEvent(
+                new WynntilsCrashEvent(feature.getClass().getName(), WynntilsCrashEvent.CrashType.FEATURE, t));
```
kristofbolyai: I just wanted to add these in the function refactor as something like `FeatureStateChangeEvent.Crashed` (disabled and enabled too..). I guess let's not forget to handle this at merge.

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -132,8 +133,8 @@ public final void setupEventListener() {
      * Adds a keyBind to the feature. Called from the registry.
      * @param keyBind KeyBind to add to the feature
      */
-    public final void setupKeyHolder(KeyBind keyBind) {
-        keyBinds.add(keyBind);
+    public final void setupKeyHolder(KeyBind keyBind, String fieldName) {
```
kristofbolyai:
```suggestion
    public final void setupKeyHolder(String fieldName, KeyBind keyBind) {
```

common/src/main/java/com/wynntils/core/WynntilsMod.java
kristofbolyai: Oh, so this works with overlay and screens too. I guess we might need to have some level of duplication. But that is a problem for the future.

common/src/main/java/com/wynntils/features/wynntils/TelemetryFeature.java
```diff
@@ -0,0 +1,45 @@
...
+    @Config
+    public boolean crashReports = true;
+
+    @SubscribeEvent
+    public void onCrash(WynntilsCrashEvent event) {
+        String title = "Crashed " + event.getType().toString().toLowerCase(Locale.ROOT) + ": " + event.getName() + "\n";
+        String trace = ExceptionUtils.getStackTrace(event.getThrowable());
+
+        ApiResponse apiResponse =
+                Managers.Net.callApi(UrlId.API_ATHENA_TELEMETRY_CRASH, Map.of("trace", title + trace));
```
kristofbolyai, line 42: I assume this is async?

common/src/main/java/com/wynntils/features/wynntils/TelemetryFeature.java
```diff
@@ -0,0 +1,45 @@
...
+
+        apiResponse.handleJsonObject(json -> {
+            String response = JsonUtils.getNullableJsonString(json, "message");
+            String hash = JsonUtils.getNullableJsonString(json, "hash");
+            if (response.equals("Crash report logged successfully.")) {
```
kristofbolyai: Qodana is right here. This is the only "outstanding" `getNullableJson*` method. Maybe let's change it?

kristofbolyai (approved):
Looks good overall. Is this being a draft not to be merged or because it is undergoing refactors?

magicus:
I thought about things a bit more, and I found a way to actually have people opt-in or opt-out. Now when the user logs in and have not decided, the following prompt will show up:

Screenshot 2023-03-12 at 00 17 10

Clicking on the links will change the value from unconfirmed to either true or false, and once that is done, the user will not be bugged again. Until that is done, the question will be repeated each time they log in.

magicus:
I set it to draft when I realized the opt-out did not work, since I did not check the variable (!), and that it was not too hard to actually ask users to opt-in or opt-out, with a reasonable and not too intrusive UX, so I thought it was important to ignore.

I'll go through your comments now.

common/src/main/java/com/wynntils/core/WynntilsMod.java
magicus: Yeah, it is a bit of duplication. I'm guessing we might perhaps add a `WynntilsMod.reportCrash`, if we can tease out enought commonality and still keep the reporting useful.

Are you OK with saving that for the future, or do you want me to make it in this PR?

common/src/main/java/com/wynntils/features/wynntils/TelemetryFeature.java
magicus: Yes, all Manager.Net calls are.

common/src/main/java/com/wynntils/features/wynntils/TelemetryFeature.java
magicus: Maybe we should, yes, but that is a bit scary. I am using it to return null in the Wynncraft gear API json, so that needs to be rewritten. I do not want to fix that in this PR.

common/src/main/resources/assets/wynntils/urls.json
magicus: https://github.com/Wynntils/WynntilsWebsite-API/pull/47

I will push that at the same time that this goes in (it's not that important to sync them, but it seems reasonable to push them roughly at the same time)

common/src/main/java/com/wynntils/core/WynntilsMod.java
magicus: I did it. It was also triggered by the fact that I realized we also have crash protection for annotators.

magicus:
So, it looks a bit different now, with the unification of the crash reporting, but I think this will be much nicer. There is a slight risk of merge conflicts wrt to the crash handling that moved into FeatureManager, but it should be trivial to resolve.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -177,7 +178,7 @@ public Object tryParseStringValue(String value) {
         try {
             Class<?> wrapped = ClassUtils.primitiveToWrapper(((Class<?>) fieldType));
             if (wrapped.isEnum()) {
-                return Enum.valueOf((Class<? extends Enum>) wrapped, value);
+                return EnumUtils.getEnumIgnoreCase((Class<? extends Enum>) wrapped, value);
```
kristofbolyai, line 181: Kind of unrelated to the PR? Does this improve anything? `/config set`?

common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -36,7 +37,7 @@
 public abstract class Feature extends AbstractConfigurable implements Storageable, Translatable, Comparable<Feature> {
     private ImmutableList<Condition> conditions;
     private boolean isListener = false;
-    private final List<KeyBind> keyBinds = new ArrayList<>();
+    private final List<Pair<String, KeyBind>> keyBinds = new ArrayList<>();
```
kristofbolyai, line 40: This conflicts a bit with my PR, but should not be hard to fix.

kristofbolyai (approved):
Looks good

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
magicus: Yes, otherwise `/config set TelemetryFeature crashReport true` did not work, but had to be `/config set TelemetryFeature crashReport TRUE`, once I made an enum out of it.

magicus: So while it is sort of unrelated, it is necessary for this PR.

PR #1316, opened 2023-03-12
feat: Random Function

Rafii2198:


Rafii2198:
I have problems with Spotless and I have no idea why

Rafii2198:
Ah I see, IntelliJ converted import to include *, and it doesn't like it. Apologies for not catching that earlier.

common/src/main/java/com/wynntils/functions/generic/RandomFunctions.java
```diff
@@ -0,0 +1,53 @@
...
+                    new FunctionArguments.Argument<>("max", Integer.class, null)));
+        }
+
+        @Override
+        public List<String> getAliases() {
+            return List.of("rand", "rnd");
+        }
+    }
+
+    public static class RandomMaxFunction extends GenericFunction<Integer> {
```
kristofbolyai: Is this a copy paste of random with max being inclusive? (While random is min exclusive and max inclusive, as it should be)

kristofbolyai: I don't really see any value in keeping both. Just have the default random implementation, like every language/etc does.

common/src/main/java/com/wynntils/functions/generic/RandomFunctions.java
```diff
@@ -0,0 +1,53 @@
...
+        @Override
+        public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
+            return new FunctionArguments.RequiredArgumentBuilder(List.of(
+                    new FunctionArguments.Argument<>("min", Integer.class, null),
+                    new FunctionArguments.Argument<>("max", Integer.class, null)));
+        }
+
+        @Override
+        public List<String> getAliases() {
+            return List.of("rand", "rnd");
```
kristofbolyai:
```suggestion
            return List.of("rand");
```

rnd might as well be round

common/src/main/java/com/wynntils/functions/generic/RandomFunctions.java
magicus: I agree. And it might just as well live in Math functions, no need to create a new top-level class for that.

Rafii2198: I did mainly random with range because there is probably no use of just a random number between 0 and 1, and it would just make functions much, much longer when it would be required to write min/max implementation manually every time.
How about, then, one function that returns a double value but with max exclusive, and a second one that return an integer value with max inclusive? The thing is that I don't know what use cases it would have, but some people on discord requested it, meaning they have some ideas and I don't know what version would suite them the most, so I think there is no harm in keeping both.

magicus: Can you explain the use case for a random function in the first place? I am not sure I understand why it is even needed.

Rafii2198: I just saw 3 people in total request it, so I decided to implement it. Some time ago someone told me they would use it for flashing text, for example in like changing colors.

Rafii2198:
I applied every change you suggested, but I kept min max aspect, I think it would be the most common thing for people to implement it themselves in their, so I think making functions shorter and thus easier for the users would be better overall.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1003,6 +1003,10 @@
   "function.wynntils.generic.power.argument.exponent": "Exponent",
   "function.wynntils.generic.power.description": "Returns the first number raised to the power of the second number",
   "function.wynntils.generic.power.name": "Power",
+  "function.wynntils.generic.random.argument.max": "Maximium value",
+  "function.wynntils.generic.random.argument.min": "Minimum value",
```
```suggestion
  "function.wynntils.generic.random.argument.max": "Maximium value (exclusive)",
  "function.wynntils.generic.random.argument.min": "Minimum value (inclusive)",
```

kristofbolyai (approved):

PR #1321, opened 2023-03-13
fix: Fix NPE in ItemLockFeature

### LINKED ISSUE #1317: NPE: Models.Character.getId() returns null in ItemLockFeature
https://athena.wynntils.com/crash/view/d57289626be316ad3f4b0bdd8cd5d3db

magicus:
This fixes #1317.


kristofbolyai (commented):
I don't think we should set this to an empty string, I am not sure it even works with GSON. Perhaps create a default string instead?

magicus:
Ok, changed to "-" which is used in getId(). But the proper fix is to stop having `hasCharacter` as different from having an ID. I intend to start fixing this soon. The CharacterModel assumed that we did not have to call the container query (since people were afraid of fake chests from legacy), but now we do that all the time, and the model should be redesigned around this fact.

kristofbolyai (approved):

PR #1325, opened 2023-03-13
fix: Validate downloaded update hash before applying

### LINKED ISSUE #1324: Verify hash of downloaded jar when upgrading
We should verify that the download succeeded before replacing the old jar when doing an in-game upgrade.

kristofbolyai:
Fixes #1324


kristofbolyai:
(This PR also includes a small code duplication cleanup)

magicus (approved):

Issue #1318, opened 2023-03-12
ChatTab regex needs to be validated


magicus:
Otherwise we risk a crash when later trying to match the chat message

https://athena.wynntils.com/crash/view/058ca1a1b0e2a442e5ca5ce95236f605

PR #1326, opened 2023-03-13
feat: Check Chat Tab regex for syntax errors

magicus:
If a previously stored regex is invalid, this is detected and the regex is ignored (but it is stored in the string, so if you edit the chat tab the old regex will be available for editing).

If an invalid regex is entered, the error message will be displayed, and the save button will be made unclickable.


magicus:
This will solve #1318

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -273,6 +273,7 @@ private void registerFeature(Feature feature) {
                 System.exit(1);
             }
         } catch (Throwable exception) {
+            // FIXME!!!
```
kristofbolyai: I assume a crashing init did fail to init all other features. I saw this behavior too.

kristofbolyai (approved):
LGTM (TextWidget now can be used in other screens)

common/src/main/java/com/wynntils/core/features/FeatureManager.java
magicus: Oops, I forgot this! That should not go in this PR; it was supposed to be a reminder to me to report this to the crash dump telemetry in a separate PR.

PR #1330, opened 2023-03-14
Convert Config to wrapper

magicus:
Ok, should we do this? I think this is actually working now. Further cleanup of ConfigManager and unification with Storage is left as a future exercise; this is just the minimal amount of work needed to change all annotation to wrapper classes.

I still think it is a bit annoying to do `.get()` everywhere, but on the other hand it is easy to spot where someone tries to write to the config (and with this change we can in fact prohibit that for pure configs, if we want). It also makes it easy to have `Storage` and `Config` both inherit a common superclass, and we can add more such subclasses for the hybrid config/storage data that we had a hard time agree on.


magicus:
One thing that worries me slightly is that we might introduce bugs. I found one bug (it was very obvious, we got into an infinite loop at startup) where we did not call `.get()` on a config. Most of these places had been found and fixed by me, but I relied on compiler errors to find where these `.get()` calls were needed. In this case, we did a null comparison, which always succeeded now, since the wrapper object was always there... 

It might be that we introduce more subtle bugs elsewhere by this PR. :( And the only way I can think of to prevent this is to check how each and every one of the quadrillion Config variables are being used... I'm not sure I'm up for that.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -103,27 +103,39 @@ private String getDeclaringFeatureNameCamelCase() {
...
     }
 
     public Object getValue() {
+        return getConfig().get();
+    }
+
+    private Config getConfig() {
         try {
-            return FieldUtils.readField(field, parent, true);
+            return (Config) FieldUtils.readField(field, parent, true);
```
kristofbolyai, line 138: Why do we need any reflection at all when we have the wrapper? Couldn't we convert all of this to not use reflection (other than for discovering configs)?

magicus:
Well, I did look through all 306 Config values anyway. And I'm glad I did, I found several bugs.

I have also changed the system so null initial values are allowed, but then we must provide a `@TypeOverride`, due to type erasure.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
magicus: We probably can. As I said:

> Further cleanup of ConfigManager and unification with Storage is left as a future exercise; this is just the minimal amount of work needed to change all annotation to wrapper classes.

magicus: This change is scary enough as it is. I don't want to do any changes that are not strictly necessary.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -221,14 +222,41 @@ public void saveDefaultConfig() {
...
         List<ConfigHolder> options = new ArrayList<>();
 
-        for (Field configField : FieldUtils.getFieldsWithAnnotation(parent.getClass(), Config.class)) {
-            Config metadata = configField.getAnnotation(Config.class);
+        Field[] annotatedConfigs = FieldUtils.getFieldsWithAnnotation(parent.getClass(), RegisterConfig.class);
+        for (Field field : annotatedConfigs) {
+            try {
+                Object fieldValue = FieldUtils.readField(field, parent, true);
+                if (!(fieldValue instanceof Config)) {
+                    throw new RuntimeException("A non-Config class was marked with @ConfigInfo annotation: " + field);
```
kristofbolyai: Let's update the annotation names in this method

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
```diff
@@ -6,61 +6,71 @@
...
+import java.lang.reflect.Type;
 import net.minecraft.client.renderer.MultiBufferSource;
 import net.minecraft.client.resources.language.I18n;
 import net.minecraft.world.phys.Vec2;
 
 public abstract class Overlay extends AbstractConfigurable implements Translatable, Comparable<Overlay> {
-    @Config(key = "overlay.wynntils.overlay.position", visible = false)
-    protected OverlayPosition position;
+    @RegisterConfig(key = "overlay.wynntils.overlay.position", visible = false)
+    public final Config<OverlayPosition> position = new Config<>(null);
```
kristofbolyai: It is debatable if configs should be public or private, but let's not change it here.

kristofbolyai (approved):
I reviewed every part which was not just `.get`. Looks good so far. You have merge conflicts to handle now though :)

common/src/main/java/com/wynntils/core/config/ConfigManager.java
magicus: Thanks for spotting. IntelliJ keeps asking "update all references?" and I keep answering "only code" since it usually messes things up otherwise... This time it would have been wise to accept its help.

common/src/main/java/com/wynntils/core/features/overlays/Overlay.java
magicus: On the one hand it really doesn't matter since we do magical stuff and force our way past protections using reflection.

On the other hand, I think they should be public since we are in effect exposing them to the world using the config files.

Prior to this fix, it was a bit more scary to expose them since in theory anyone could have modified them. Now we can make them essentially read-only, which alleviates that problem.

But I changed them back to protected for now.

PR #1332, opened 2023-03-14
refactor: Sort out position <-> location

magicus:
We are now consistently using "location" to refer to integer coordinates (like on a map, or a BlockPos), and "position" to refer to real number coordinates (typically entity positions).

I have replaced as many instaces as possible by the Vanilla interface "Position". I have redesigned our "Location" so it only deals with integer coordinates.


magicus:
Next up I will try to see if PoiLocation and Location can be merged, they seem very similar now. I'd also like to see if we can introduce some abstraction instead of Vec3 for the remaining places; basically this is due to the use of `distanceTo` which is only exported from Vec3.

kristofbolyai (approved):

PR #1335, opened 2023-03-14
refactor: Clean up RegisterConfig etc

magicus:


common/src/main/java/com/wynntils/features/wynntils/ChangelogFeature.java
```diff
@@ -24,8 +25,8 @@
 @ConfigCategory(Category.WYNNTILS)
 public class ChangelogFeature extends Feature {
     // "v0.0.2-alpha.2" is the first version with a changelog on GitHub
-    @RegisterConfig(visible = false)
-    public final Config<String> lastShownVersion = new Config<>("v0.0.2-alpha.2");
+    @RegisterConfig
+    public final HiddenConfig<String> lastShownVersion = new HiddenConfig<>("v0.0.2-alpha.2");
```
kristofbolyai, line 29: Since we allow null default configs, we should populate this with the first version the user ran. Obviously, separately.

kristofbolyai (approved):
Looks good so far :)

common/src/main/java/com/wynntils/features/wynntils/ChangelogFeature.java
magicus: I'd like to take a sweep through all configs and make sure they make sense, and has a good default value.

PR #1336, opened 2023-03-14
fix: Try to avoid NPE in mount horse if user logs out

magicus:


kristofbolyai (commented):
I am not quite sure I get how we can disconnect while running this method, since this is theoretically on the same thread as the packet listener thread.

common/src/main/java/com/wynntils/features/combat/HorseMountFeature.java
```diff
@@ -60,19 +61,20 @@ public void onUseItem(UseItemEvent event) {
     private void mountHorse() {
         if (!Models.WorldState.onWorld()) return;
 
-        if (McUtils.player().getVehicle() != null) {
+        LocalPlayer player = McUtils.player();
```
kristofbolyai, line 64: add a player null check here?

magicus:
The problem is in `trySummonAndMountHorse`, where we schedule `searchForHorseNearby` to run 6 ticks later. But I realize now that I was a bit too tired when I wrote the PR, this is not really a great solution. I've updated it.

common/src/main/java/com/wynntils/features/combat/HorseMountFeature.java
magicus: That won't be necessary. I checked the access paths to this point and it can only end up here in active game play. The problem were, as I said, in the tick delayed code.

kristofbolyai (approved):

PR #1337, opened 2023-03-15
fix: Fix faulty dynamic logic causing some overlay configs to be not collected

kristofbolyai:
There was some (seemingly not) faulty logic behind DynamicOverlay, which in turn lead to non-dynamic TextOverlays (so every text overlay except info boxes) to have their configs not collected.

This fixes the problem, but we still have the weird inheritance issue, where we both want a `DynamicOverlay` and `TextOverlay` base class, but we can't extend from both. So now some of our overlays are `DynamicOverlays` even though they are actually not. I am not sure what we can do with this, since it is not easy to convert either of these classes to an interface.


magicus (approved):

magicus:
I'll have a look at the Overlay inheritance and see if I can think up some solution. But it'll have to wait a bit, I'm still too focused on the Config and Feature changes.

PR #1347, opened 2023-03-16
fix: Fix not being able to track first mini-quest

### LINKED ISSUE #1205: Can't select first item on mini-quest book
Artemis version : v0.0.2-alpha.237

Hello,
When I open the quest book and select mini-quests, I can't select the first mini-quest. It doesn't add waypoint on the map and doesn't display the Quest Info.

The Middle click to open the map and the Right click to open the wiki work but the Left click to select the quest doesn't work only for the first item of the list. 

Best regards

kristofbolyai:
Fixes https://github.com/Wynntils/Artemis/issues/1205


magicus (approved):
Aha, so that was the reason! :) I thought the bug report just sounded so strange.

PR #1355, opened 2023-03-17
feat: Introduce Custom Bar Overlays

kristofbolyai:
DO NOT MERGE WITHOUT ANNOUNCEMENT!

This was supposed to be part 1, with part 2 being refactoring all existing bars. But I think we can release this as-is, and convert the existing ones later.

I renamed the old `CustomBarsFeature` to `GameBarsFeature`, and made a new `CustomBarsFeature`, so the diff is large. 

Overall I think this is a definite step forward with functions, especially capped values.


kristofbolyai:
image
image
image

common/src/main/java/com/wynntils/functions/EnvironmentFunctions.java
```diff
@@ -20,6 +20,12 @@ public static class CappedMemFunction extends Function<CappedValue> {
         public CappedValue getValue(FunctionArguments arguments) {
             return new CappedValue(SystemUtils.getMemUsed(), SystemUtils.getMemMax());
         }
+
+        @Override
+        public List<String> getAliases() {
+            // FIXME: These aliases are a bit backwards, let's clean it up in the future
```
kristofbolyai, line 26: I plan to fix this in an upcoming PR, fixing all these weird names. I think this is fine to leave.

common/src/main/java/com/wynntils/utils/render/type/HealthTexture.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.render.type;
+
+public enum HealthTexture {
```
kristofbolyai, line 7: All 3 of these were moved out of CustomBarsOverlayFeature. I am not fully sure if this is the right place for them, but hopefully it is close :)

magicus:
Oohhh, the screenshots look very nice indeed! I'll review it when I'm feeling a bit more sharp than tonight.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
```diff
@@ -1,237 +1,95 @@
...
 
         @Override
-        public float textureHeight() {
-            return healthTexture.get().getHeight();
+        public Texture getTexture() {
+            return Texture.HEALTH_BAR;
         }
 
         @Override
-        public String icon() {
```
magicus, line 217: Do we lose the heart icon?

kristofbolyai: All of this feature was moved. Diff is weird. Check description for more info

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
```diff
@@ -391,253 +145,156 @@ protected void renderBar(
...
+                    182,
+                    objectivesTexture.get().getTextureYOffset() + 10,
+                    progress);
         }
 
         @Override
-        public boolean isActive() {
-            return Models.BossBar.corruptedBar.isActive();
+        protected BarOverlayTemplatePair getActualPreviewTemplate() {
+            return new BarOverlayTemplatePair("{capped_ingredient_pouch_slots}", "capped_ingredient_pouch_slots");
```
magicus, line 257: I'm not sure if I'm reading the diff incorrectly, but it looks like the BubbleBar is based on ingredient pouch slots?

kristofbolyai: Yes, the preview is. I had to make a random example, let me know if I should change it.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
magicus: Ok, I understand now.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
magicus: I now think the CustomBubbleBar is really like "CustomBarWithBubbleTexture". Is that a correct understanding? Then the example makes sense. But that was not really clear to me.

magicus: Was it too tricky to have just a single type of CustomBar in which you could select the texture to use? Or are there other differences that I don't understand between the different kinds of custom bars?

common/src/main/java/com/wynntils/features/overlays/GameBarsOverlayFeature.java
```diff
@@ -0,0 +1,580 @@
...
+        }
+
+        @Override
+        protected void renderBar(
+                PoseStack poseStack,
+                MultiBufferSource.BufferSource bufferSource,
+                float renderY,
+                float renderHeight,
+                float progress) {
+            if (progress > 1) { // overflowing mana
```
magicus, line 396: Why do we need to implement renderBar here? We're doing something the superclass can't handle?

common/src/main/java/com/wynntils/features/overlays/GameBarsOverlayFeature.java
```diff
@@ -0,0 +1,580 @@
...
+
+        @Override
+        public BossBarProgress progress() {
+            CappedValue health = Models.CharacterStats.getHealth();
+            return new BossBarProgress(health, (float) health.getProgress());
+        }
+
+        @Override
+        protected void renderBar(
+                PoseStack poseStack,
```
magicus, line 241: Same question here (sorry for the reverse order)

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
kristofbolyai: You got it right. And we need the different subclasses because health/mana/experience/bubble have different texture types themselves. And we have the universal as colorable, but a single texture.

common/src/main/java/com/wynntils/features/overlays/GameBarsOverlayFeature.java
kristofbolyai: It's overflowing code, I did not invent these, so I did not touch it.

kristofbolyai: (remember that i just copied these)

common/src/main/java/com/wynntils/functions/EnvironmentFunctions.java
magicus: I used to think we should keep parity with the old names, but they kind of suck all of them, so I'm more leaning towards giving stuff new, good, names. We're changing the syntax away from %% anyway, so there is not much to convert over.

For functions returning capped values, I don't think we should have any "uncapped" variants. So there should just be a "mana", etc.

kristofbolyai: I agree. We need to fix it in a singular, breaking PR

common/src/main/java/com/wynntils/utils/render/type/HealthTexture.java
magicus: It is close, I agree :) We have some kind of Texture/Resource file somewhere around here as well. I'm not fully sure how we really should handle this, but the current approach is almost good enough, so it can probably stay that way for a while, even if it is not perfect.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
magicus: Let's say I hope for a future improvement in which we can select texture as well for a custom bar :-)

Maybe you can rename the classes a bit to make this clear, like `CustomBarWithBubbleTexture` or something like that?

common/src/main/java/com/wynntils/features/overlays/GameBarsOverlayFeature.java
magicus: Ok. One more thing we can have a look at for future improvements, see if this code can be moved upwards.

magicus:
It's mostly ok. I'd prefer if you could rename the texture custom subclasses, and then you're good to go.

magicus (approved):

magicus:
(And sorry for the delay in reviewing, been dragged away a bit from Wynntils development)

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
kristofbolyai: > Let's say I hope for a future improvement in which we can select texture as well for a custom bar :-)
> 
> Maybe you can rename the classes a bit to make this clear, like `CustomBarWithBubbleTexture` or something like that?

I am not quite sure `CustomBarWithBubbleTextureOverlay` sounds better. Or does it..?

kristofbolyai:
@magicus I forgot to include the upfixer in this PR. Please check out the last commit, and make sure to let me know if I can merge this.

kristofbolyai:
@magicus I forgot to include the upfixer in this PR. Please check out the last commit, and make sure to let me know if I can merge this.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
kristofbolyai: Also let me know what you think here.

magicus: No, I don't think it sounds better. But I think the other name is confusing, like the bar did something *more* than just have a different texture. Especially for Health and Mana. As I said, I hope we can refactor this in the future so there is only one type of custom overlay with a selectable texture. Until then, I think it needs to be clear that the different subclasses only exists to provide different textures. Maybe `BubbleTexturedCustomBarOverlay` then?

common/src/main/java/com/wynntils/core/config/upfixers/impl/GameBarOverlayMoveUpfixer.java
```diff
@@ -0,0 +1,104 @@
...
+package com.wynntils.core.config.upfixers.impl;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.wynntils.core.config.upfixers.ConfigUpfixer;
+import java.util.List;
+
+public class GameBarOverlayMoveUpfixer implements ConfigUpfixer {
+    private static final List<String> KEYS_TO_CHANGE = List.of(
+            "customBarsOverlayFeature.awakenedProgressBarOverlay.flip",
```
magicus, line 14: And it's not like `customBarsOverlayFeature.awakenedProgressBarOverlay.*` could work? Are there configs that should not be converted? The list just looked very repetetive...

Are we thinking we should keep these upfixers forever and ever, or that we should remove them after a while? If it is the latter, I can accept this code, but it it is here to stay forever, I think it needs a bit more shaping up.

kristofbolyai: I would assume we can reset these at release? Or at beta.. I am not sure.

common/src/main/java/com/wynntils/features/overlays/CustomBarsOverlayFeature.java
kristofbolyai: Renamed it!

magicus (approved):

common/src/main/java/com/wynntils/core/config/upfixers/impl/GameBarOverlayMoveUpfixer.java
magicus: Ok.

magicus:
Oops, almost forgot your comment! It's OK now, you merge it when you are ready!

PR #1357, opened 2023-03-21
feat: Inventory watcher, MiscItem, label visibility etc

magicus:
Preparations for an upcoming feature, including:

* Add InventoryWatcher, and convert old empty slots to use it
* Add label visibility detection to LabelHandler
* Introduce a MiscItem with annotator
* Allow BarOverlays to specify size
* Improvements in i18n asserts, to display the missing key
* Some bug fixes and robustness fixes in CappedValue and TimedSet

### DIFF
--- a/common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
+++ b/common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
@@ -43,6 +43,10 @@ public abstract class BarOverlay extends DynamicOverlay {
         this.size.updateConfig(overlaySize);
     }
 
+    protected BarOverlay(OverlayPosition position, OverlaySize size) {
+        super(position, size, 1);
+    }
+
     @Override
     public void render(
             PoseStack poseStack, MultiBufferSource.BufferSource bufferSource, float partialTicks, Window window) {
@@ -78,6 +82,8 @@ public abstract class BarOverlay extends DynamicOverlay {
 
         String textValue = templateCache.a();
         CappedValue value = valueOrError.getValue();
+        if (value.equals(CappedValue.EMPTY)) return;
+
         renderText(poseStack, bufferSource, renderY, textValue);
 
         float progress = (float) ((flip.get() ? -1 : 1) * value.getProgress());
--- a/common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
+++ b/common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
@@ -29,7 +29,7 @@ public class InventoryWatcher {
     }
 
     // This should only be called by the model
-    public boolean include(ItemStack itemStack) {
+    public boolean shouldInclude(ItemStack itemStack) {
         return checker.test(itemStack);
     }
 
--- a/common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
+++ b/common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.containers.type;
+
+import java.util.function.Predicate;
+import net.minecraft.world.item.ItemStack;
+
+public class InventoryWatcher {
+    private int slots;
+    private int totalCount;
+    private final Predicate<ItemStack> checker;
+
+    public InventoryWatcher(Predicate<ItemStack> checker) {
+        this.checker = checker;
+    }
+
+    public int getSlots() {
+        return slots;
+    }
+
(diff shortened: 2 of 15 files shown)


common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
```diff
@@ -78,6 +82,8 @@ protected void render(
 
         String textValue = templateCache.a();
         CappedValue value = valueOrError.getValue();
+        if (value.equals(CappedValue.EMPTY)) return;
```
kristofbolyai, line 85: I don't think I agree with this? This should be a config, but we already have a way of setting visibility with templates.

common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
```diff
@@ -0,0 +1,44 @@
...
+    public int getTotalCount() {
+        return totalCount;
+    }
+
+    protected void onUpdate(int oldSlots, int oldTotalCount) {
+        // Called after an update has been made, override if needed
+    }
+
+    // This should only be called by the model
+    public boolean include(ItemStack itemStack) {
```
kristofbolyai:
```suggestion
    public boolean includes(ItemStack itemStack) {
```
?

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
```diff
@@ -0,0 +1,44 @@
...
+        boolean questItem = false;
+
+        for (Tag line : loreTag) {
+            String coded = ComponentUtils.getCoded(line.getAsString());
+            if (coded.equals(UNTRADABLE)) {
+                untradable = true;
+            }
+            if (coded.equals(QUEST_ITEM)) {
+                questItem = true;
+            }
```
kristofbolyai, line 35: Afaik only either of these can happen, so we can break early.

kristofbolyai: (`getCoded` is quite heavy from profiling, so this is not without basis)

kristofbolyai (commented):
I overall agree with this, but BarOverlay needs some configs (and i forgot to hide it when not on world)

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
magicus: So what do you want to display for something that returns an EMPTY capped value? Note that this is essentially "impossible", it is 0/0. No normal data will have this, so it is used to signal essentially "n/a".

magicus: Also note that this is not so much a config, as that a bar that has no value to display should not be drawn. As soon as you get proper data (max != 0), then it should of course be drawn again.

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
magicus: I am like 99% sure that you can have both. It is not common, but possible. I think I encountered some such items when doing the gear model.

kristofbolyai: Okay.

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
kristofbolyai: Okay, let's go with this and I will see if I want to change it later.

kristofbolyai (approved):

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
magicus: I also remember strongly that you were worried about item annotation performance, which infected me as well, so I measured very carefully. In the end, it was almost impossible to measure, except when like speed-clicking through the Trade Market pages. 

This is just before the fallback annotator. I don't think you have to be that worried about performance.

common/src/main/java/com/wynntils/models/containers/type/InventoryWatcher.java
magicus: I think `includes` just sounded off, like it was something that was included already. I renamed it `shouldInclude`, I hope you think this is okay.

magicus: If you accept, please resolve, otherwise we can think over the naming a bit more. Maybe it needs to be something more along the lines of it being a predicate...

kristofbolyai: Sounds fine

PR #1358, opened 2023-03-21
feat: Token Gatekeeper Tracker

magicus:
The Token Gatekeeper Tracker keeps track of how many tokens you have collected, showing the progress as a bar, and playing a sound when you have collected enough tokens.

Note that the functions added currently do not handle multiple token gatekeepers at the same time very well.

This depends on https://github.com/Wynntils/Artemis/pull/1357.


common/src/main/java/com/wynntils/models/token/TokenModel.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    // This happens if we log out, switch class, etc
+    @SubscribeEvent
+    public void onEntitiesRemoved(RemoveEntitiesEvent event) {
+        for (int id : event.getEntityIds()) {
+            if (activeGatekeepers.containsKey(id)) {
+                removeGatekeeper(id, activeGatekeepers.get(id));
+            }
+            if (invisibleGatekeepers.containsKey(id)) {
```
kristofbolyai: Qodana is right, no need for the contains

common/src/main/java/com/wynntils/models/token/TokenModel.java
```diff
@@ -0,0 +1,252 @@
...
+        private String type;
+        private CappedValue value;
+        private int valueEntityId;
+        private boolean confirmed;
+
+        private BakingTokenGatekeeper(Position position) {
+            this.position = position;
+        }
+
+        protected TokenGatekeeper toGatekeeper() {
```
kristofbolyai: 2 Qodana suggetions in 1 PR. Is that a new record :)?

common/src/main/java/com/wynntils/models/token/TokenModel.java
```diff
@@ -0,0 +1,252 @@
...
+            BakingTokenGatekeeper baking = getBaking(event.getEntity().position());
+            baking.value = tokens;
+            baking.valueEntityId = event.getEntity().getId();
+            checkAndPromoteBaking();
+            return;
+        }
+
+        // This does not provide any new information for the TokenGatekeeper, but it is a good
+        // extra check that we really match only on real gatekeepers
+        if (name.equals(VERIFICATION_STRING)) {
```
kristofbolyai, line 117:
```suggestion
        if (VERIFICATION_STRING.equals(name)) {
```

common/src/main/java/com/wynntils/models/token/TokenModel.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        private static boolean isToken(String type, ItemStack itemStack) {
+            Optional<MiscItem> miscItemOpt = Models.Item.asWynnItem(itemStack, MiscItem.class);
+            if (miscItemOpt.isEmpty()) return false;
+
+            MiscItem miscItem = miscItemOpt.get();
+            if (!miscItem.isUntradable()) return false;
+
+            return (miscItem.getName().contains(type));
```
kristofbolyai:
```suggestion
            return miscItem.getName().contains(type);
```

kristofbolyai (approved):
LGTM after the minor style changes

magicus:
> You have successfully added a new QDJVMC configuration `project/qodana`. As part of the setup process, we have scanned this repository and found 2 existing alerts. Please check the repository [Security tab](/Wynntils/Artemis/security/code-scanning?query=pr%3A1358+tool%3AQDJVMC+is%3Aopen) to see all alerts.

@kristofbolyai Any idea why this suddenly appeared? Did I do something weird to trigger this?

common/src/main/java/com/wynntils/models/token/TokenModel.java
magicus: I guess it is just me that think strange to rely on remove to work properly even for non-existing items. Oh well, I'll fix it if you want.

kristofbolyai:
I am not sure, we are gonna have to see in the next Pr. It is not so bad though.. But maybe it will be annoying eventually

common/src/main/java/com/wynntils/models/token/TokenModel.java
magicus: And then it was 3... and now 0. :)

common/src/main/java/com/wynntils/models/token/TokenModel.java
magicus: Why??!? We know for sure name is not null. I distinctly unlike with a passion close to hate the Yoda-style comparison ("if 2 == var"), and can hardly even stand it even if it would save a null check...

kristofbolyai: For some reason my eye prefered the suggestion but reading the code "name is X" is better than "X is name"

common/src/main/java/com/wynntils/functions/WorldFunctions.java
```diff
@@ -58,6 +60,129 @@ public List<String> getAliases() {
...
+        }
+    }
+
+    public static class TokenGatekeeperDepositedFunction extends Function<CappedValue> {
+        @Override
+        public CappedValue getValue(FunctionArguments arguments) {
+            List<TokenGatekeeper> gatekeepers = Models.Token.getGatekeepers();
+            if (gatekeepers.isEmpty()) return CappedValue.EMPTY;
+
+            return gatekeepers.get(0).getDeposited();
```
kristofbolyai: Why don't these take an argument for index, like other functions (others are required, but you could make this default 0 if you want)?

kristofbolyai (approved):

common/src/main/java/com/wynntils/functions/WorldFunctions.java
magicus: Hm, I did not think of that...

magicus: There are two sets of all functions, one which defaults to index 0, and one which takes an index argument. But you are right that having 0 as default would be better. I'll fix.

PR #1360, opened 2023-03-22
feat: Trade market bulk sell

ryanzhoudev:
https://user-images.githubusercontent.com/57310593/226846296-47ffb93e-ac54-4c26-b5d5-cd4d0c2f649c.mp4


common/src/main/java/com/wynntils/features/overlays/TradeMarketBulkSellFeature.java
```diff
@@ -0,0 +1,145 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.OVERLAYS)
+public class TradeMarketBulkSellFeature extends Feature {
+
+    @RegisterConfig
+    public Config<Integer> bulkSell1Amount = new Config<>(64);
+
+    @RegisterConfig
+    public Config<Integer> bulkSell2Amount = new Config<>(1728);
```
magicus: These numbers seem a bit random. I assume they come from Legacy? (I always thought these numbers were a bit arbitrary, too).

What is the need for a "sell <fixed amount>" button? I understand "Sell all", that is great. But the others? Do we need 3 fixed amount buttons? It seems overkill.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -771,6 +771,18 @@
...
   "feature.wynntils.tooltipVanillaHide.name": "Hide Irrelevant Vanilla Tooltip Info",
   "feature.wynntils.tradeMarketAutoOpenChat.name": "Market Auto Open Chat",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell1Amount.description": "How many items should be sold when the bulk sell 1 button is pressed?",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell1Amount.name": "Bulk Sell 1 Amount",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell2Amount.description": "How many items should be sold when the bulk sell 2 button is pressed?",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell2Amount.name": "Bulk Sell 2 Amount",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell3Amount.description": "How many items should be sold when the bulk sell 3 button is pressed?",
+  "feature.wynntils.tradeMarketBulkSell.bulkSell3Amount.name": "Bulk Sell 3 Amount",
+  "feature.wynntils.tradeMarketBulkSell.description": "Should there be bulk sell buttons in the trade market UI?",
+  "feature.wynntils.tradeMarketBulkSell.enableSellAll.description": "Should there be a sell all button in the trade market UI?",
```
magicus: This was a good idea, but it is not implemented..?

common/src/main/java/com/wynntils/features/overlays/TradeMarketBulkSellFeature.java
```diff
@@ -0,0 +1,145 @@
...
+    private int amountToSend = 0;
+
+    @SubscribeEvent
+    public void onSellDialogueOpened(ScreenOpenedEvent e) {
+        if (!(e.getScreen() instanceof ContainerScreen cs)) return;
+        if (!ComponentUtils.getUnformatted(cs.getTitle()).equals(SELL_DIALOGUE_TITLE)) return;
+
+        sellButtons.clear();
+        sellButtons.add(new SellButton(cs.leftPos - BUTTON_WIDTH, cs.topPos, 0, true));
+        sellButtons.add(new SellButton(cs.leftPos - BUTTON_WIDTH, cs.topPos + 21, bulkSell1Amount.get(), false));
```
magicus: Maybe you can make additional buttons optional, so if the amount is 0 they are not drawn? And perhaps have the defaults 64, 0, 0, so there is the default of selling a stack, and selling all, and then people can tweak their config if they are prolific sellers?

common/src/main/java/com/wynntils/features/overlays/TradeMarketBulkSellFeature.java
ryanzhoudev: Supposedly the reasoning from legacy was that 64 does one stack, 1728 does your inventory, and 6399 does your inventory+bank

common/src/main/resources/assets/wynntils/lang/en_us.json
ryanzhoudev: It originally existed in legacy just to enable/disable the feature, but now that we have that...

magicus (approved):

common/src/main/java/com/wynntils/features/overlays/TradeMarketBulkSellFeature.java
magicus: Ah, that'd explain the numbers at least. Thanks. Hard to imagine anyone filling your entire bank with a single type of item, though :)

PR #1361, opened 2023-03-22
fix: missing i18n for token gatekeeper functions

ryanzhoudev:


magicus:
Ah, oops. But this is the wrong solution... Sorry, will fix.

magicus:
Full fix is in https://github.com/Wynntils/Artemis/pull/1362

PR #1364, opened 2023-03-22
feat: Allow features to register commands

magicus:
This PR is driven by two thoughts:

1) It is too hard to get command support to a feature.  Many times I've thought that "oh, it would be nice if we could do this by a command", but it would break abstraction, and it was too tricky to set up.

2) I think many (mostly? nearly all?) existing commands really belong with a Feature. I'm seeing a future where we can move the command logic to features, and then if we want something simpler than `/execute LootrunFeature save` we can extend the alias system so we can make an alias `/lootrun` to `/execute LootrunFeature` and we'll get to keep the current short name, but it will all be more flexible.


common/src/main/java/com/wynntils/features/debug/LogItemInfoFeature.java
```diff
@@ -25,6 +35,10 @@ public class LogItemInfoFeature extends Feature {
     private final KeyBind logItemInfoKeyBind = new KeyBind(
             "Log Item Info", GLFW.GLFW_KEY_INSERT, true, this::onLogItemInfoPress, this::onLogItemInfoInventoryPress);
 
+    @RegisterCommand
+    LiteralCommandNode<CommandSourceStack> commandNode =
```
kristofbolyai:
```suggestion
    private final LiteralCommandNode<CommandSourceStack> commandNode =
```

common/src/main/java/com/wynntils/core/features/FeatureCommands.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.reflect.Field;
+import net.minecraft.commands.CommandSourceStack;
+import net.minecraft.commands.Commands;
+import org.apache.commons.lang3.reflect.FieldUtils;
+
+public class FeatureCommands {
+    private final LiteralCommandNode<CommandSourceStack> commandNode;
+
+    public FeatureCommands() {
+        commandNode = Managers.Command.registerCommand(Commands.literal("execute"));
```
kristofbolyai, line 20: is this /wynntils execute? /execute is taken likely (by mc too)

magicus: No, it is `/execute`. Just like `/function`, you are correct that this is a vanilla command, but it is not available when you are logged in to Wynncraft. And conversly, our `/execute` will only be available when logged in to Wynncraft. So there is no collision.

common/src/main/java/com/wynntils/features/debug/LogItemInfoFeature.java
magicus: Oops! 8-o

kristofbolyai (approved):

kristofbolyai:
You can't easily type insert on a Mac, can you :)?

magicus:
No, but I have remapped the key. In fact, I chose the log feature just as a way to prove that the code worked. (In fact I wanted to use it in a separate branch to create some commands to use during development of the overlay groups).

And when I actually did use it, I noticed it needed some love :)

PR #1365, opened 2023-03-22
feat: death coordinates

ryanzhoudev:
image


common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.core.Vec3i;
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.CHAT)
+public class DeathCoordinatesFeature extends Feature {
+
```
kristofbolyai, line 20:
```suggestion
```

common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,60 @@
...
+                        HoverEvent.Action.SHOW_TEXT,
+                        Component.translatable("feature.wynntils.deathCoordindates.clickToCompass"))));
+
+        e.setCanceled(true);
+        McUtils.player().sendSystemMessage(deathMessage.append(coordMessage));
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent e) {
+        if (!Models.WorldState.onWorld()) return;
```
kristofbolyai: We only send this event on world/lobby, this is just "extra" code for nothing. I would remove it for simpleness.

common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,60 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.CHAT)
+public class DeathCoordinatesFeature extends Feature {
+    private Vec3i lastLocation;
+
+    @SubscribeEvent
+    public void onChatReceived(ChatMessageReceivedEvent e) {
+        if (!Models.WorldState.onWorld() || lastLocation == null) return;
+        if (!e.getCodedMessage().contains("§r §4§lYou have died...")) return;
```
kristofbolyai: Please extract this to a constant (private static final)

common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent e) {
+        if (!Models.WorldState.onWorld()) return;
+
+        lastLocation = new Vec3i(
+                McUtils.player().getX(),
+                McUtils.player().getY(),
+                McUtils.player().getZ());
```
kristofbolyai: I would change lastLocation to `Vec3` and use `McUitls.player().position()` so we don't allocate on tick.

kristofbolyai (approved):
Resolve my last comment and you are good to merge.

I don't want to force doing this on you for now, but this feature felt so much like legacy code.. It took me a few minutes to figure this out. But now I understand!

You are doing work in a feature, that should really be in `CharacterStatsModel` (where we handle health related stuff, name is a bit bad imo if we extend it with this). Basically, in Artemis, in the future, try to aim to do everything Wynn-related in models (and broadcast with events), everything user representation in Feature/Overlay/Function and everything mod-processing in Managers. 

Again, you are free to fix this, and you are also feel to merge and when we need a death event, someone will fix it.

kristofbolyai:
Also, thank you for working on Artemis again :). This was along the first few features I have ever added in legacy :)... It feels so long ago.

ryanzhoudev:
All this refactoring might be a bit out of scope, but I think this is maybe better than just cramming it into the feature

kristofbolyai (approved):
LGTM. @magicus what do you think of the new model name?

common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,40 @@
...
+
+@ConfigCategory(Category.CHAT)
+public class DeathCoordinatesFeature extends Feature {
+
+    @SubscribeEvent
+    public void onChatReceived(CharacterDeathEvent e) {
+
+        MutableComponent deathMessage = Component.translatable("feature.wynntils.deathCoordinates.diedAt")
+                .withStyle(ChatFormatting.DARK_RED);
+        MutableComponent coordMessage = Component.translatable(
```
magicus: We have some utility function to create a component with a build-in on click... lemme check...

magicus: I thought we had, but I can't find it now. Nevertheless, the coordinate format is quite non-standard; we use `[123, 4, 567]` everywhere else. And that does not need to be a i18n string.

magicus: Found it! In ChatCoordinatesFeature.createLocationComponent. You can move it to ComponentUtils and re-use it instead, that'd be much better than to reimplement it.

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -2,32 +2,41 @@
...
+public final class CharacterStateModel extends Model {
     private final CoordinatesSegment coordinatesSegment = new CoordinatesSegment(this::centerSegmentCleared);
     private final HealthSegment healthSegment = new HealthSegment();
     private final ManaSegment manaSegment = new ManaSegment();
     private final PowderSpecialSegment powderSpecialSegment = new PowderSpecialSegment();
     private final SprintSegment sprintSegment = new SprintSegment();
 
-    public CharacterStatsModel(CombatXpModel combatXpModel) {
+    private static final String WYNN_DEATH_MESSAGE = "§r §4§lYou have died...";
+    private Vec3 lastLocation = new Vec3(0, 0, 0);
```
magicus: Please use `Position` instead of `Vec3` unless you need to do vector operations.

common/src/main/java/com/wynntils/models/characterstate/event/CharacterDeathEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.characterstate.event;
+
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.eventbus.api.Event;
+
+public class CharacterDeathEvent extends Event {
+
+    private final Vec3 deathLocation;
```
magicus: Same here, overall in the event, use Position.

common/src/main/java/com/wynntils/models/characterstate/event/CharacterDeathEvent.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public CharacterDeathEvent(Vec3 deathLocation) {
+        this.deathLocation = deathLocation;
+    }
+
+    public Vec3 getDeathLocation() {
+        return deathLocation;
+    }
+
+    public int getDeathX() {
```
magicus: These are not necessary, the caller can do `getLocation.getX()` instead. Unless there are very good reasons, it is better to keep these event classes concise and to the point.

common/src/main/java/com/wynntils/models/characterstate/event/CharacterDeathEvent.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class CharacterDeathEvent extends Event {
+
+    private final Vec3 deathLocation;
+
+    public CharacterDeathEvent(Vec3 deathLocation) {
+        this.deathLocation = deathLocation;
+    }
+
+    public Vec3 getDeathLocation() {
```
magicus: Also, I think you can shorten the `deathLocation` to just `location` everywhere. It is the `CharacterDeathEvent` after all, so location cannot reasonably mean anything else than the death location...

common/src/main/java/com/wynntils/features/chat/DeathCoordinatesFeature.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.CHAT)
+public class DeathCoordinatesFeature extends Feature {
+
+    @SubscribeEvent
+    public void onChatReceived(CharacterDeathEvent e) {
```
magicus:
```suggestion
    public void onCharacterDeath(CharacterDeathEvent e) {
```

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -105,4 +114,16 @@ public int getTicksToNextSoulPoint() {
     private void centerSegmentCleared() {
         powderSpecialSegment.replaced();
     }
+
+    @SubscribeEvent
+    public void onChatReceived(ChatMessageReceivedEvent e) {
+        if (!e.getCodedMessage().contains(WYNN_DEATH_MESSAGE)) return;
```
magicus: I am not found of matching partial strings in chat. I think you can do `equals` instead, right? If not, check what the entire string is and make `equals` on that. If it is not constant but varies, make a regex instead.

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -105,4 +114,16 @@ public int getTicksToNextSoulPoint() {
...
+        WynntilsMod.postEvent(new CharacterDeathEvent(lastLocation));
```
magicus: I would actually recommend that you also save a copy of lastLocation to a new field e.g. `lastDeathLocation` before sending out the event. It is a slight extra cost, and allows for future queries to the model, e.g. if someone wants to add a function `lastDeathCoords` or whatever.

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -105,4 +114,16 @@ public int getTicksToNextSoulPoint() {
...
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent e) {
+        if (McUtils.player() == null) return;
```
magicus: I guess this works, but I would personally have looked for the player teleport event instead, and only save the last location before being teleported. Then if we get the death message, we know we were teleported since we died.

This is not expensive, but I do want to avoid adding stuff to all ticks, if reasonably possible.

You can keep it this way for now if you just want to get this pushed, or you can implement a mixin for teleportation, if you want to take the opportunity of learning Artemis mixins.

magicus: Oh, I see that there is already a PlayerTeleportEvent! It is called as "Post" rather than "Pre", though, which means the players current position will be overwritten when you get it. So the mixin should be changed from "RETURN" to "HEAD", and then you can listen to that instead of the TickEvent.

magicus:
> what do you think of the new model name?

Honestly I don't think it is a good idea.. I think the "character" model stuff is a kind of "misc bag", which we have had a hard time finding a good separation for. My thinking in introducing "CharacterStats" was that it should contain just things you'd likely call "stats", basically health, mana, combat and profession xp/level. I thought it made a relatively clear separation.

But then I put SoulPoints in there as well, which is kind of different. Or maybe not. 

But the death location is more ... eh, I dunno. I don't think it is a stat. And I don't think it is really helpful to make a "state" separation.

I'd rather suggest that we move the death location stuff to the original "misc bag", that is, the `CharacterModel`, instead of trying to shoehorn it into CharacterStatsModel and renaming it.

The CharacterModel deals mostly with login and stuff, and dying kinda, sorta, fits in with that pattern.

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -105,4 +114,16 @@ public int getTicksToNextSoulPoint() {
...
+        lastLocation = McUtils.player().position();
```
magicus: Actually, come to think of it, you are probably better of using a `Location` than a `Position`. The difference is that the former are integer coordinates only (BlockPos), while the latter have float coordinates (like the actual position of an entity). Minecraft will give you a position, but you really only care about the closest block. 

Sorry for not realizing this before, but even better then is to replace all instances of `Position` (or `Vec3`) in this PR with `Location`, and store the lastLocation using `lastLocation = Location.containing(McUtils.player().position());`

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
```diff
@@ -33,8 +35,11 @@ public final class CharacterStateModel extends Model {
     private final PowderSpecialSegment powderSpecialSegment = new PowderSpecialSegment();
     private final SprintSegment sprintSegment = new SprintSegment();
 
-    private static final String WYNN_DEATH_MESSAGE = "§r §4§lYou have died...";
-    private Vec3 lastLocation = new Vec3(0, 0, 0);
+    // we need a .* in front because the message may have a custom timestamp prefix (or some other mod could do
+    // something weird)
+    private static final Pattern WYNN_DEATH_MESSAGE = Pattern.compile(".*§r §4§lYou have died\\.\\.\\.");
+    private Vec3 lastLocationBeforeTeleport;
```
magicus:
```suggestion
    private Position lastPositionBeforeTeleport;
```

ryanzhoudev: If this is going to be (possibly) used for other things in the future, wouldn't it make sense to keep it as the furthest class down the tree?

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -105,4 +119,17 @@ public int getTicksToNextSoulPoint() {
     private void centerSegmentCleared() {
         powderSpecialSegment.replaced();
     }
+
```
magicus: Can you move all these additions from CharacterStatsModel to CharacterModel?

common/src/main/java/com/wynntils/models/characterstate/CharacterStateModel.java
magicus: I'd say no, a better design principle is actually the exact opposite: never declare a field (or a method argument) as a more concrete type than you have to. That is, better to call something `List<String>` than `LinkedList<String>`. Then you can still instantiate it as a LinkedList, but you do not give any guarantees that it is anything but a List. The only reason you would ever need to declare it as a LinkedList if you needed to access methods that are only available on LinkedList, like `removeFirst()` or whater it is called.

ryanzhoudev: Ah alright

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -45,7 +46,7 @@ public final class CharacterModel extends Model {
     // we need a .* in front because the message may have a custom timestamp prefix (or some other mod could do
     // something weird)
     private static final Pattern WYNN_DEATH_MESSAGE = Pattern.compile(".*§r §4§lYou have died\\.\\.\\.");
-    private Vec3 lastLocationBeforeTeleport;
+    private Position lastLocationBeforeTeleport;
```
magicus: Can you also please rename it so the name is "Position" instead of "Location"? I've tried to make sure the code base follows this pattern to help minimize confusion (once you learn about the difference, at least...)

magicus: Then you are done and we can integrate. Thanks for hanging in there, I realize there is a lot of coding differences between Legacy and Artemis!

PR #1366, opened 2023-03-22
feat: profession function improvements

ryanzhoudev:


magicus (approved):
Ah, this is a great cleanup! And nice twist with the xp per minute stuff as well.

magicus:
At some point, I think we should consider trying to use more CappedValues, e.g. on profession levels. That would make it possible to e.g. draw a bar on how far you have come on your way to maximum level. Don't expecting you do to it as part of this PR, just throwing out the idea.

PR #1370, opened 2023-03-24
feat: Text box enhancements

ryanzhoudev:
Adds
- Clicking somewhere in the box to set cursor position
- Click and drag to highlight
- Ctrl+A to highlight
- Typing over highlighted replaces text
- Home and End buttons work now
- Shift+Arrow to highlight, Ctrl+Shift+Arrow to highlight all in one direction
- Cursor reworked to not be an actual character, is just a rectangle now

https://user-images.githubusercontent.com/57310593/227700157-20019107-45e3-4d63-b127-e8ec28ce9905.mp4


magicus:
I agree that our current implementation is limited, but it is not better to see if we can use the vanilla implementation? It has stuff like selection, copy/paste, etc.

kristofbolyai:
> I agree that our current implementation is limited, but it is not better to see if we can use the vanilla implementation? It has stuff like selection, copy/paste, etc.

I tried once already, it is a very limited API for what Mojang needed (this is a trend for all of MC sadly).

magicus:
@kristofbolyai What is is missing, then? Since it does work a lot better from the user perspective, with the in-line editing and stuff...

kristofbolyai:
> @kristofbolyai What is is missing, then? Since it does work a lot better from the user perspective, with the in-line editing and stuff...

I tried using it (although I did not get it to work in the 10 minutes I spent trying..), but I realized: We might be better off reinventing the wheel and just creating our own version. Yes, we have to maintain it. But imagine a world of where I can  create a syntax highlighted text box for functions. That is just one example which is pretty much only possible with a custom implementation (or so many mixins we would not accept it). So I believe I support this PR. But, please let me know what you think.

kristofbolyai:
What I found during my limited testing:

- [x] Shift + Arrows -> highlight not implemented

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -41,19 +50,88 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
 
         boolean defaultText = Objects.equals(textBoxInput, "") && !isFocused();
 
-        String renderedText = getRenderedText(this.width - 18);
+        Pair<String, Integer> renderedTextDetails = getRenderedText(this.width - 18, true);
```
kristofbolyai: Let's unbox this pair so it is much easier to read the code (like you did with `renderedText`)

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -24,6 +31,8 @@ public class SearchWidget extends TextInputBoxWidget {
     public SearchWidget(
             int x, int y, int width, int height, Consumer<String> onUpdateConsumer, TextboxScreen textboxScreen) {
         super(x, y, width, height, Component.literal("Search Box"), onUpdateConsumer, textboxScreen);
+        textPadding = 5;
```
kristofbolyai, line 30: Is this specifically needed here or should we do this in super too?

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -41,19 +50,88 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
+            highlightedEnd = renderedText.length();
+        } else if (highlightedStart < renderedTextDetails.b() && highlightedEnd <= renderedTextDetails.b() + renderedText.length()) {
+            // The highlighted text starts outside of the rendered text, but ends within it
+            highlightedStart = 0;
+            highlightedEnd -= renderedTextDetails.b();
+        } else {
+            // The highlighted text is not within the rendered text
+            highlightedStart = 0;
+            highlightedEnd = 0;
+        }
```
kristofbolyai: Let's refactor this. 

Let `renderedTextDetails.b()` be `shift`.
Let `renderedText.a().length()` be `length`.

Let's define two intervals: 
Let `renderedInterval` be `[0, length]`
Let `highlightedInterval` be `[highlightStart, highlightEnd]`

We can get the intersection of these two intervals to know which part to highlight.
```
Let outputInterval be 
  if highlightedInterval.end > renderedInterval.start or renderedInterval.start > highlightedInterval.end 
     then [0, 0]
  else
    then [max(renderedInterval.start, highlightedInterval.start) + shift, min(renderedInterval.end, 
  highlightedInterval.end) + shift]
```

This way we guarantee that `outputInterval` is in bounds of the rendered string, and can be used to get the highlighted sub-string along with the pre-highlight and post-highlight parts.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -41,19 +50,88 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
+        String lastNormalPortion;
+        try { // There is a rare race condition where if the user spams drag, the highlighted portion can be out of bounds
+            firstNormalPortion = renderedText.substring(0, highlightedStart);
+            highlightedPortion = renderedText.substring(highlightedStart, highlightedEnd);
+            lastNormalPortion = renderedText.substring(highlightedEnd);
+        } catch (StringIndexOutOfBoundsException ignored) {
+            firstNormalPortion = renderedText.substring(0, highlightedStart-1);
+            highlightedPortion = renderedText.substring(highlightedStart-1, highlightedEnd-1);
+            lastNormalPortion = renderedText.substring(highlightedEnd-1);
+        }
```
kristofbolyai: I don't think this is needed anymore with the above change. But don't forget that `substring(start, end)` interval is half-closed (`[s, e)`) while our interval is closed (`[s, e]`), so you have to account for that.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -41,19 +50,88 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
-                        this.getX() + this.width - 5,
+                        defaultText ? DEFAULT_TEXT.getString() : lastNormalPortion,
+                        this.getX() + textPadding + FontRenderer.getInstance().getFont().width(firstNormalPortion) + FontRenderer.getInstance().getFont().width(highlightedPortion),
+                        this.getX() + this.width - textPadding,
                         this.getY() + 6.5f,
-                        this.width,
+                        0,
                         defaultText ? CommonColors.LIGHT_GRAY : CommonColors.WHITE,
                         HorizontalAlignment.Left,
                         TextShadow.NORMAL);
```
kristofbolyai: All of the font width calls need to be extracted into a variable. You should also return early if `defaultText` is true (or handle that in any nice way, as you wish).

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -85,88 +98,221 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
         RenderUtils.drawRect(poseStack, CommonColors.BLACK, 0, 0, 0, this.width, this.height);
         RenderUtils.drawRectBorders(poseStack, CommonColors.GRAY, 0, 0, this.width, this.height, 0, 2);
 
-        String renderedText = getRenderedText(this.width - 8);
+        Pair<String, Integer> renderedTextDetails = getRenderedText(maxTextWidth, false);
```
kristofbolyai: You can easily remove the duplication here. The super method does 3 things: Posestack translation, background rendering, text rendering. You need to create methods for these. You can override the background method in `SearchWidget` and just not call the posestack translation method.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -85,88 +98,221 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
-
-            int stringPosition = cursorPosition - 1;
+        String cursorChar = getRenderCursorChar(forceUnfocusedCursor);
+        final int cursorWidth = font.width(cursorChar);
 
-            while (font.width(builder.toString()) < maxTextWidth - cursorWidth && stringPosition >= 0) {
-                builder.append(textBoxInput.charAt(stringPosition));
+        String entireText = textBoxInput.substring(0, cursorPosition) + cursorChar + textBoxInput.substring(cursorPosition);
+        if (font.width(entireText) < maxTextWidth) {
+            return new Pair<>(entireText, 0);
```
kristofbolyai: Pair.of is a bit shorter (also below)

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
kristofbolyai: Also, everything I commented in SearchWidget applies here too (as the code is copy-pasted)

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -85,88 +98,221 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
+    public boolean mouseDragged(double mouseX, double mouseY, int button, double dragX, double dragY) {
+        if (isDragging) {
+            setCursorPosition(getIndexAtPosition(mouseX));
+        }
+        return super.mouseDragged(mouseX, mouseY, button, dragX, dragY);
+    }
+
+    protected int getIndexAtPosition(double mouseX) {
+        mouseX -= this.getX(); // mouseX is actually just the x position of the mouse relative to the screen, not the textbox
+        Font f = FontRenderer.getInstance().getFont();
```
kristofbolyai:
```suggestion
        Font font = FontRenderer.getInstance().getFont();
```

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -85,88 +98,221 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
+        Pair<String, Integer> renderedTextDetails = getRenderedText(maxTextWidth, false);
```
kristofbolyai: Extract `renderedTextDetails.b()` to a var

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -85,88 +98,221 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
...
+        // Width so far at index i
+        List<Float> widths = new ArrayList<>();
+        for (int i = 0; i < renderedText.length(); i++) {
+            // we are using stringWidth here because we need precision; if we use width, it will round to the nearest
+            // integer, which will cause strange behaviour when clicking on letters
+            widths.add(f.getSplitter().stringWidth(renderedText.substring(0, i)));
+        }
+
+        // get nearest width that's in the map
+        // FIXME: this is probably really slow and bad, but I am just a first year cs student and I have not taken
```
kristofbolyai: We have `com.wynntils.utils.mc.RenderedStringUtils#getMaxFittingText` which does all the work you need here. You need to just make the `...` optional, or create a new method.

kristofbolyai: I think `getMaxFittingText(renderedText, mouseX, font).length() - 1 + renderedTextDetails.b()`, but I might not understood the code fully.

magicus:
Let me be clear: If Donkey is putting in the work for this, I will not reject it. I'm just thinking it feels like a better path to start with Vanilla and then tweak it to our needs. I'll have a look at the Vanilla code and see if I think I can find a way to reuse it.

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
```diff
@@ -146,6 +152,59 @@ public void renderAlignedTextInBox(
...
+                backgroundColor,
+                renderX,
+                switch(verticalAlignment) {
+                    case Top -> renderY - 2;
+                    case Middle -> renderY - (getFont().lineHeight / 2f) - 2;
+                    case Bottom -> renderY - getFont().lineHeight + 2;
+                },
+                0,
+                getFont().width(text),
+                getFont().lineHeight + 2
```
kristofbolyai: Use the font field here

kristofbolyai:
The actual feature is working really nicely! I will definitely create some kind of syntax highlight for functions with this.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
ryanzhoudev: Needed here. My idea was that if we wanted to change the padding, we could just change this variable. The only place padding is used in the superclass is to check the offset for mouse clicks to select. In this class, it's used for the font rendering.

ryanzhoudev:
> What I found during my limited testing:
> 
> * [x]  Control + Arrows -> highlight not implemented

Is this supposed to be like that? In most windows (discord, chrome, etc), ctrl+arrow just acts the same as home/end

kristofbolyai:
> > What I found during my limited testing:
> > 
> > * [ ]   Control + Arrows -> highlight not implemented
> 
> Is this supposed to be like that? In most windows (discord, chrome, etc), ctrl+arrow just acts the same as home/end

Sorry! I meant shift + arrows for highlight. Control + Arrows jumps to the end and start of the current word.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
ryanzhoudev: (Same as mentioned in Discord)

Can't use this method, as it rounds to the leftmost character when between characters

magicus:
Just a question: Have you accounted for the fact that Ctrl-<key> typically should be Cmd-<key> on macOS?

ryanzhoudev:
> Just a question: Have you accounted for the fact that Ctrl- typically should be Cmd- on macOS?

Yeah it looks like it. I used Screen.hasControlDown(), which checks for osx.

kristofbolyai:
@DonkeyBlaster I will clean this up tomorrow, since I can't fully put my words into suggestions here. Feel free to do any work though, I am heading to sleep soon.

ryanzhoudev:
After some more extensive testing, I have determined this is not really in a state for merging. There's some strange behaviour when config book text boxes contain text longer than the box, working on it now.

kristofbolyai:
This is mostly working now. The only thing that is needed to be done is changing how we select the rendered text. If you check Chrome's behavior, it is clear that the text does ONLY move with the cursor, if it is needed - compared to how we do it - where we move the rendered text every time if we go backwards. I did not implement this, for now. (This actually causes a small weirdness with selecting text with a mouse). @DonkeyBlaster if you want to fix this, go ahead, but I think we should merge this as-is. I will review this, and request a review from you as well. Feel free to merge any time you feel like.

kristofbolyai (approved):

ryanzhoudev:
This feels good in-game now, minus the weirdness with the selecting stuff when text is overflowing. Though I agree that we should just merge now, I can try a fix for that later

PR #1372, opened 2023-03-24
feat: Introduce Overlay container, and start using it for Token overlay

magicus:


magicus:
Currently the only operations allowed by ContainerOverlay is to add a single child at the time, or remove all. It is possible to implement other operations as well, but it quickly starts getting tricky, and I did not see the need to support that right now.

magicus:
This code was really 80/20 rule -- the basic wrapped rendering of the children were done by almost no work, but then getting the alignment, preview, resizing, etc right was tons of work...

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
```diff
@@ -93,7 +94,8 @@ public void enableOverlays(Feature parent) {
     public void enableOverlay(Overlay enableOverlay) {
         if (!enableOverlay.shouldBeEnabled()) return;
 
-        this.enabledOverlays.add(enableOverlay);
+        enabledOverlays.add(enableOverlay);
+        WynntilsMod.registerEventListener(enableOverlay);
```
kristofbolyai, line 98: This singular change opens up a whole lot of refactoring that can be done in features. Obviously I don't require you to do it, but if you want to :)

kristofbolyai (commented):
Glanced over this but I will have to take a look tomorrow. Too tired to fully understand this, and I also want to test it to see what it can do.

common/src/main/java/com/wynntils/core/features/overlays/OverlayManager.java
magicus: Overall, I'm perfectly happy of leaving the overlays to you. :-) I was hoping I could trick you into writing this ContainerOverlay as well ;-) but I did not manage to express what I was going for. On the flip side, I learned a lot about the overlays in the process.

kristofbolyai: I will clean them up, and review this, but after that I will have to do some projects of my own. User experience and scoreboard handler rewrite.

magicus:
The basic principle is that the container overlay will contain a list of child overlays, which it will dictate the position (and partially the size) for, and each render call to the container will be delegated to the children.

That was a very nice and clear, 10-line-ish long implementation. And then I needed to add support for propagating alignment, windows resize, etc, which complicated things a bit.

The container can grow in one of the cardinal directions. This means new children are placed in that direction. If children are growing in a vertical direction, then each childs height is determined by the "inherent" height of the child, and they are stacked upon each other. The width of each child is instead determined by the width of the container. It feels very natural so it probably is more complicated to explain than to use it. (And of course, the vice versa applies if the grow direction is horizontal.)

kristofbolyai:
> The basic principle is that the container overlay will contain a list of child overlays, which it will dictate the position (and partially the size) for, and each render call to the container will be delegated to the children.
> 
> That was a very nice and clear, 10-line-ish long implementation. And then I needed to add support for propagating alignment, windows resize, etc, which complicated things a bit.
> 
> The container can grow in one of the cardinal directions. This means new children are placed in that direction. If children are growing in a vertical direction, then each childs height is determined by the "inherent" height of the child, and they are stacked upon each other. The width of each child is instead determined by the width of the container. It feels very natural so it probably is more complicated to explain than to use it. (And of course, the vice versa applies if the grow direction is horizontal.)

Sounds good! I was worried about "we can only delete all overlays" but then I realized this is completely fine and I just did not think this through. But the review is gonna have to wait, I had a (too) long day today.

magicus:
The handling of the children list is very simplistic right now, since that was all that was currently needed. If we should be able to delete individual children we need to handle either holes in the ID sequence, or a re-id:ing. Since I do not need any such functionality right now (clearing all and re-adding those that are needed) were enough for me, and I did not want to second-guess a potential future extension  to this functionality, I just kept it at the basic minimum needed. 

If we need more advanced handling in the future, let's add it then.

kristofbolyai (approved):
LGTM! We could technically use this for stopwatch too.

PR #1373, opened 2023-03-25
fix: item counts hiding stacked items

ryanzhoudev:
Noticed with the event fireworks. [1/1] items can be stacked, and this fixes it just saying `1` in the hotbar.


magicus (approved):
Nice! I noted this too, but could not figure out how to make a decent way to display both. But I assume that multiplying stack count with the wynnitem count makes sense.

PR #1374, opened 2023-03-25
feat: Add stopwatch

ryanzhoudev:
By default, yellow when paused, grey when 0, white when running. Can be modified in a custom info box with provided functions


common/src/main/java/com/wynntils/models/stopwatch/StopwatchModel.java
```diff
@@ -0,0 +1,63 @@
...
+    private long startTimeMillis = -1;
+    private long currentTimeMillis = -1;
+    private boolean running = false;
+
+    public StopwatchModel() {
+        super(List.of());
+    }
+
+    public int getHours() {
+        update();
```
magicus: If you are going to call cTM() at each getter, there is no need to store the value in a field..?

common/src/main/java/com/wynntils/features/overlays/StopwatchFeature.java
```diff
@@ -0,0 +1,67 @@
...
+        if (Models.Stopwatch.isRunning()) {
+            Models.Stopwatch.stop();
+        } else {
+            Models.Stopwatch.start();
+        }
+    }
+
+    public static class StopwatchOverlay extends TextOverlay {
+        private static final String TEMPLATE =
+                "{STOPWATCH_HOURS}:{STOPWATCH_MINUTES}:{STOPWATCH_SECONDS}.{STOPWATCH_MILLISECONDS}";
```
magicus: We usually have the function names in lower case

common/src/main/java/com/wynntils/features/overlays/StopwatchFeature.java
```diff
@@ -0,0 +1,67 @@
...
+public class StopwatchFeature extends Feature {
+    @RegisterKeyBind
+    private final KeyBind toggleStopwatchKeybind =
+            new KeyBind("Toggle Stopwatch", GLFW.GLFW_KEY_KP_0, true, this::toggleStopwatch);
+
+    @RegisterKeyBind
+    private final KeyBind resetStopwatchKeybind =
+            new KeyBind("Reset Stopwatch", GLFW.GLFW_KEY_KP_DECIMAL, true, Models.Stopwatch::reset);
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
```
magicus, line 63: This might be a good place to use the new command support:
```
    @RegisterCommand
    private final LiteralCommandNode<CommandSourceStack> startCommand =
            Commands.literal("start").executes(Models.Stopwatch::start).build();

    @RegisterCommand
    private final LiteralCommandNode<CommandSourceStack> stopCommand =
            Commands.literal("stop").executes(Models.Stopwatch::stop).build();

    @RegisterCommand
    private final LiteralCommandNode<CommandSourceStack> resetCommand =
            Commands.literal("reset").executes(Models.Stopwatch::reset).build();
```

That's all there is to it, I think.

common/src/main/java/com/wynntils/features/overlays/StopwatchFeature.java
ryanzhoudev: Okay, I will change the aura timer one as well then

magicus (approved):
Nice! Thanks for going through the remaining legacy functionality we're missing!

PR #1376, opened 2023-03-25
feat: Stop playing token bell sound after a while

magicus:
feat: Stop playing token bell sound after you have passed a few tokens above the needed amount

As requested by @P0keDev


kristofbolyai:
Looks good to me, although 5 pings might be a little extensive. I will leave accepting this to @P0keDev.

magicus:
It's a balance between not missing the ping if you are actually collecting the tokens, and turning it off if you obviously are ignoring them. And, after all, if you never run dungeons for real, then you can just turn off the sound completely in the Feature. This is just a way to be able to have it turned on at all times, without getting too annoying if you chose to ignore the tokens. Hearing 5 dings for that is acceptable, I think.

P0keDev (approved):

PR #1381, opened 2023-03-26
fix: Hide custom overlay bars if not on world by default & allow disabling wynn bars even if the default overlay is not enabled

kristofbolyai:


magicus:
Just checking, this is for hiding when on hub or in character selection, yes? If we're not on the server, the overlays are hidden as they should?

kristofbolyai:
> Just checking, this is for hiding when on hub or in character selection, yes? If we're not on the server, the overlays are hidden as they should?

Yes.

magicus (approved):

PR #1382, opened 2023-03-26
feat: Waypoint management GUI

ShadowCat117:
All works as far as I can tell. No import/export feature or filtering by icon


common/src/main/java/com/wynntils/screens/maps/PoiCreationScreen.java
```diff
@@ -65,7 +65,8 @@ public final class PoiCreationScreen extends WynntilsScreen implements TextboxSc
     private CustomPoi.Visibility selectedVisiblity = CustomPoi.Visibility.DEFAULT;
     private CustomColor colorCache = CommonColors.WHITE;
 
-    private final MainMapScreen oldMapScreen;
+    private MainMapScreen oldMapScreen;
+    private PoiManagementScreen managementScreen;
```
kristofbolyai: You can just merge this and save these as `Screen`

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1293,6 +1293,21 @@
   "screens.wynntils.map.share.description3_1": "Shift",
   "screens.wynntils.map.share.description3_2": " to share your compass pin location instead.",
   "screens.wynntils.map.share.name": "Share Location",
+  "screens.wynntils.map.manager.name": "Manage Waypoints",
+  "screens.wynntils.map.manager.description1_1":  "List, Delete or Create",
+  "screens.wynntils.map.manager.description1_2":  " all your preview set",
+  "screens.wynntils.map.manager.description1_3":  " waypoints!",
```
kristofbolyai: You can just make these description1, description2, description3, as there are no subcategories

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1293,6 +1293,21 @@
...
+  "screens.wynntils.poiManagementGui.close":  "X",
+  "screens.wynntils.poiManagementGui.next":  ">",
+  "screens.wynntils.poiManagementGui.previous":  "<",
```
kristofbolyai: No need to be excessive with translations. These can be made literals, they will never change in any language.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1293,6 +1293,21 @@
...
+  "screens.wynntils.poiManagementGui.group":  "Group",
+  "screens.wynntils.poiManagementGui.icon":  "Icon",
+  "screens.wynntils.poiManagementGui.name":  "Name",
+  "screens.wynntils.poiManagementGui.x":  "X",
+  "screens.wynntils.poiManagementGui.y":  "Y",
+  "screens.wynntils.poiManagementGui.z":  "Z",
```
kristofbolyai: I would remove these from here too

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,132 @@
...
+    private void renderIcon(PoseStack poseStack, CustomPoi poi, int row) {
+        float[] color = CustomColor.fromInt(poi.getColor().asInt()).asFloatArray();
+        RenderSystem.enableBlend();
+        RenderSystem.blendFunc(GlStateManager.SourceFactor.SRC_ALPHA, GlStateManager.DestFactor.ONE_MINUS_SRC_ALPHA);
+        RenderSystem.setShaderColor(color[0], color[1], color[2], 1);
+
+        float centreZ = 64 + spacingMultiplier * row;
+
+        RenderUtils.drawTexturedRect(
+                poseStack, POI_ICONS.get(POI_ICONS.indexOf(poi.getIcon())), this.width / 2f - 151 - (poi.getIcon().width() / 2f), centreZ - (poi.getIcon().height() / 2f));
```
kristofbolyai: `poi.getIcon()` returns the texture? Why do we look up the index and get it?

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,132 @@
...
+            Texture.FARMING,
+            Texture.FISHING,
+            Texture.MINING,
+            Texture.WOODCUTTING);
+
+    private CustomPoi poi;
+    private Button editButton;
+    private Button deleteButton;
+    private int row;
+    private int colour;
```
kristofbolyai:
```suggestion
    private CustomColor color;
```

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,132 @@
...
+    private int spacingMultiplier = 20; //Small 14, medium 20, large 25
```
kristofbolyai: Are these planned to be changeable or what is the purpose?

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
kristofbolyai: (Don't forget to remove `POI_ICONS` when you simplify this)

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,132 @@
...
+        drawCenteredString(poseStack, McUtils.mc().font, y.isPresent() ? String.valueOf(y.get()) : "", this.width/2 + 40, 60 + spacingMultiplier * row, colour);
+        drawCenteredString(poseStack, McUtils.mc().font, String.valueOf(poi.getLocation().getZ()), this.width/2 + 80, 60 + spacingMultiplier * row, colour);
+
+        editButton.render(poseStack, mouseX, mouseY, partialTick);
+        deleteButton.render(poseStack, mouseX, mouseY, partialTick);
+
+        System.out.println("X: " + this.getX() + "\nY: " + this.getY() + "\nWidth: " + this.width + "\nHeight: " + this.height);
+    }
+
+    private void renderIcon(PoseStack poseStack, CustomPoi poi, int row) {
```
kristofbolyai: What is stopping you from using `IconPoi#renderAt`?

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,132 @@
...
+        this.row = row;
+        this.managementScreen = managementScreen;
+
+        colour = 0xFFFFFF;
+
+        if (poi.getVisibility() == CustomPoi.Visibility.HIDDEN) {
+            colour = 0x636363;
+        }
+
+        int groupShift = group == ungroupedIndex ? 20 : 0;
```
kristofbolyai: Honestly I have no idea what any of these 3 variables do

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.List;
+
+public class PoiManagementScreen extends WynntilsScreen {
+    private final MainMapScreen oldMapScreen;
+    private Button nextButton;
+    private Button previousButton;
+    private List<CustomPoi> waypoints;
+    private int pageHeight;
+    private int page;
+    private int spacingMultiplier = 20; //Small 14, medium 20, large 25. Make customisable
```
kristofbolyai: Left a comment in the widget itself but I don't think this is a good idea. If we make this a constant 20, does it not work on a specific GUI scale?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,199 @@
...
+                        .size(20, 20)
+                        .build());
+
+        onWaypointChange();
+    }
+
+    @Override
+    public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        renderBackground(poseStack);
+        super.doRender(poseStack, mouseX, mouseY, partialTick);
```
kristofbolyai, line 114: Does this super call do anything (same goes for `super.doInit`)?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,199 @@
...
+                    0,
+                    0,
+                    this.width,
+                    this.height,
+                    poi,
+                    i,
+                    this));
+        }
+
+        poiManagerWidgets.forEach(widget -> widget.render(poseStack, mouseX, mouseY, partialTick));
```
kristofbolyai: Doing anything like adding widgets during render time is strictly bad design (legacy :)). Make this a method, do this on init and whenever we delete (edit will reopen this screen so init will handle it)

kristofbolyai: No wonder why your mouse clicks are not getting registered. I do not know how legacy did this, but here you need to use `addRenderableWidget` to register for widgets to be rendered, and all events will also be called automatically (like mouse events). Use `removeWidget` to clear widgets using your `poiManagerWidgets`.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: Sorry, this is likely used to render your widgets, when you register them. However, the question still stands for doInit.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,253 @@
...
+        this.poiManagerWidgets.clear();
+
+        for (int i = 0, lim = Math.min(pageHeight, waypoints.size() - pageHeight * page); i < lim; i++) {
+            CustomPoi poi = waypoints.get(page * pageHeight + i);
+
+            if (poi == null) {
+                continue;
+            }
+
+            PoiManagerWidget newWidget = new PoiManagerWidget(
```
JamieCallan117, line 208: This is the only part I'm unsure of if it's done well, as far as I can tell the widget renders fine but I would prefer it with actual values.

I did try inputting what I thought was correct but it then ended up moving each row to the very left of the window but the Y value was correct so seemingly the X was wrong which I was trying with `this.width/2 - 151` which is the top left position of the icon as that's the leftmost part of the widget.

I've tried resizing my window and different GUI scales and it looks good so it doesn't seem to be a problem but if there's a scenario where it would be a problem then I'm not certain on how to do this correctly

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,253 @@
...
+                    0,
+                    0,
+                    this.width,
+                    this.height,
+                    poi,
+                    i,
+                    this);
```
JamieCallan117, line 215: This is the only part I'm unsure of if it's done well, as far as I can tell the widget renders fine but I would prefer it with actual values.

I did try inputting what I thought was correct but it then ended up moving each row to the very left of the window but the Y value was correct so seemingly the X was wrong which I was trying with `this.width/2 - 151` which is the top left position of the icon as that's the leftmost part of the widget.

I've tried resizing my window and different GUI scales and it looks good so it doesn't seem to be a problem but if there's a scenario where it would be a problem then I'm not certain on how to do this correctly

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,253 @@
...
+
+            if (poi == null) {
+                continue;
+            }
+
+            PoiManagerWidget newWidget = new PoiManagerWidget(
+                    0,
+                    0,
+                    this.width,
+                    this.height,
```
ShadowCat117: This is the only part I'm unsure of if it's done well, as far as I can tell the widget renders fine but I would prefer it with actual values.

I did try inputting what I thought was correct but it then ended up moving each row to the very left of the window but the Y value was correct so seemingly the X was wrong which I was trying with `this.width/2 - 151` which is the top left position of the icon as that's the leftmost part of the widget.

I've tried resizing my window and different GUI scales and it looks good so it doesn't seem to be a problem but if there's a scenario where it would be a problem then I'm not certain on how to do this correctly

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Override
+    public void renderButton(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        MultiBufferSource.BufferSource bufferSource =
+                McUtils.mc().renderBuffers().bufferSource();
+
+        float centreZ = 64 + spacingMultiplier * row;
+
+        poi.renderAt(poseStack, bufferSource, this.width / 2f - 151, centreZ, false, 1f, 1f);
```
ShadowCat117: When there is only 1 widget being rendered on the page this doesn't display the icon.

kristofbolyai:
@JamieCallan117 If this is ready, please mark it as "ready for review". Also, if you think you have completed a request, please mark the conversation as resolved.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1311,6 +1311,10 @@
   "screens.wynntils.map.help.description8": "Right click to center on player.",
   "screens.wynntils.map.help.description9": "Shift-Right click to center on the compass marker.",
   "screens.wynntils.map.help.name": "Quick Guide",
+  "screens.wynntils.map.manager.description1": "List, Delete or Create",
+  "screens.wynntils.map.manager.description2": " all your preview set",
```
magicus: This does not really make sense?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -99,6 +99,28 @@ protected void doInit() {
...
+                        - 6),
+                16,
+                16,
+                Texture.MAP_MANAGER_BUTTON,
+                (b) -> McUtils.mc().setScreen(PoiManagementScreen.create(this)),
+                List.of(
+                        Component.literal("[>] ")
+                                .withStyle(ChatFormatting.RED)
+                                .append(Component.translatable("screens.wynntils.map.manager.name")),
+                        Component.translatable("screens.wynntils.map.manager.description1")
```
magicus: Did you intend for the descriptions to come on separate lines? Then you have to send them as separate item in the list. By using append they will all become a single line -- this may be fine, but if that is the intention you can have just a single i18n key and put it all on a single line.

kristofbolyai: While this sounds logical, it might not be the case, I am pretty sure this is a copy of what I did above, and it works.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright © Wynntils 2022.
```
magicus: It's 2023 now...

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+
+        this.addRenderableWidget(
+                previousButton = new Button.Builder(Component.literal("<"), (button) -> {
+                            previousPage();
+                        })
+                        .pos(this.width / 2 - 22, this.height - 45)
+                        .size(20, 20)
+                        .build());
+
+        this.addRenderableWidget(
```
magicus, line 82: This sounds very dangerous and drastic. Are we sure this is even needed? If someone mess up their waypoints that much, it is probably better to edit the json config files.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+        this.poiManagerWidgets.clear();
+
+        if (Math.min(pageHeight, waypoints.size() - pageHeight * page) == 0 && page != 0) {
+            previousPage();
+        }
+
+        for (int i = 0, lim = Math.min(pageHeight, waypoints.size() - pageHeight * page); i < lim; i++) {
+            CustomPoi poi = waypoints.get(page * pageHeight + i);
+
+            if (poi == null) {
```
magicus: How can this happen?

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,174 @@
...
+        this.poi = poi;
+        this.row = row;
+        this.managementScreen = managementScreen;
+
+        pois = Managers.Feature.getFeatureInstance(MapFeature.class).customPois.get();
+
+        color = CustomColor.fromInt(0xFFFFFF);
+
+        if (poi.getVisibility() == CustomPoi.Visibility.HIDDEN) {
+            color = CustomColor.fromInt(0x636363);
```
magicus: Please extract magic numbers as a constant

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,174 @@
...
+    public PoiManagerWidget(
+            float x, float y, int width, int height, CustomPoi poi, int row, PoiManagementScreen managementScreen) {
+        super((int) x, (int) y, width, height, Component.literal(poi.getName()));
+        this.poi = poi;
+        this.row = row;
+        this.managementScreen = managementScreen;
+
+        pois = Managers.Feature.getFeatureInstance(MapFeature.class).customPois.get();
+
+        color = CustomColor.fromInt(0xFFFFFF);
```
magicus: We have CommonColors.WHITE for this purpose

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
kristofbolyai: CommonColors.GRAY would be fine, no need to have this specifically set

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
```diff
@@ -0,0 +1,174 @@
...
+        this.editButton = new Button.Builder(
+                        Component.translatable("screens.wynntils.poiManagementGui.edit"),
+                        (button) -> McUtils.mc().setScreen(PoiCreationScreen.create(managementScreen, poi)))
+                .pos(this.width / 2 + 85 + 20, 54 + spacingMultiplier * row)
+                .size(40, 20)
+                .build();
+
+        this.deleteButton = new Button.Builder(
+                        Component.translatable("screens.wynntils.poiManagementGui.delete"), (button) -> {
+                            Managers.Feature.getFeatureInstance(MapFeature.class)
```
magicus, line 72: Having a delete without either a confirmation, or a way to undo, is a bit unforgiving.

(I'd prefer a "Undo last delete" rather than a confirmation)

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: Huh? This comment seems out of place. What was the purpose of it?

ShadowCat117: Sorry let me be more clear. When a page is only displaying 1 poi, so the final page, the icon for that poi doesn't render. This code is definetly being ran as the name and all other information about the poi render but for some reason the icon doesn't.

common/src/main/resources/assets/wynntils/lang/en_us.json
ShadowCat117: I just copied it from legacy originally and noticed that too, just forgot to change it to something better

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
ShadowCat117: Yes, I just copied what the other buttons did and it works

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: It does have to be pressed 3 times so while dangerous I think most people will not accidently trigger it. It was a feature in legacy so I ported it, but if you think it's not worth having then I can easily remove it

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
ShadowCat117: I could make the delete button require 2 presses, similar to the delete all button, as well as the undo last delete, or would that be overkill?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
magicus: Okay but then you don't need three different i18n keys, you can just have a single `description` key.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
magicus: Just because it was that way in legacy does not mean it was a good feature. :-) Apart from the risk of e.g. an autoclicker getting astray, there is also the risk that we somehow ourselves call such a function. I have a hard time seeing how this could ever be useful, compared to the risk that something goes wrong. 

Let's hear @kristofbolyai's opinion, but otherwise I'd recommend that you remove it, yes.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: Ok, now I understand: you have a bug that you cannot resolve, and need help fixing it.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: In theory it shouldn't, it's just a safety precaution in case it somehow does. Can remove it if you think it's pointless having it but I just thought better to be on the safe side

magicus: I think it is pointless. Even if this happened and you skipped one poi, the index would be out of phase and stuff will likely break anyway. If you get a NPE, we will register this as a crashed screen and send a report, so we can investigate.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: I think that's overkill. Look at how Github does when you delete a branch after merging a PR. I think that's pretty nifty. You can delete it with just a single click, but if you mis-clicked, you get the chance to undo. So maybe do something like this:
when deleting a poi, save it as "last deleted" in the screen, and in the screen, if there is a "last deleted" poi, show a "undo delete" button (which is otherwise hidden, or possibly visible but greyed out), and if that is pressed, just re-add the poi. And don't bother trying to persist the "last deleted poi", if you exit the screen well then you're out of luck.

I think that'd work nicely.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: Nothing really sticks out as being wrong. The one thing I react to is that in this case `row` is 0, so you don't get `spacingMultiplier` in work at all. Maybe some maths is wrong, so it is technically being rendered, but off-screen? Maybe try this hypothesis by offsetting row with one? Then all icons should be rendered one step down.

ShadowCat117: That sounds like it could be it, I'll give that a try later

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: You can reset pois via a command, if you really want to. So I don't think we need this.

common/src/main/java/com/wynntils/utils/render/Texture.java
```diff
@@ -86,6 +86,7 @@ public enum Texture {
     MAP_HELP_BUTTON("map/map_help_button.png", 10, 16),
     MAP_POINTERS("map/map_pointers.png", 256, 256),
     MAP_SHARE_BUTTON("map/map_share_button.png", 16, 14),
+    MAP_MANAGER_BUTTON("map/map_manager_button.png", 12, 16),
```
kristofbolyai: Please keep these groups sorted.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+
+        // In case doInit called again, won't say press 0 times
+        if (deleteCount <= 0) {
+            deleteAllButton.setMessage(Component.literal("DELETED!"));
+        }
+
+        waypoints =
+                Managers.Feature.getFeatureInstance(MapFeature.class).customPois.get();
+
+        if (!waypoints.isEmpty() && page * pageHeight > waypoints.size() - 1) {
```
kristofbolyai: Checking `waypoints.isEmpty()` seems pretty redundant

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+                        this.width / 2f + 80,
+                        43,
+                        CommonColors.WHITE,
+                        HorizontalAlignment.Left,
+                        VerticalAlignment.Top,
+                        TextShadow.NORMAL);
+
+        RenderUtils.drawRect(poseStack, CommonColors.WHITE, this.width / 2 - 165, 52, 0, 355, 1);
+
+        poiManagerWidgets.forEach(widget -> widget.render(poseStack, mouseX, mouseY, partialTick));
```
kristofbolyai: You don't need this as you call super.render.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+            this.removeWidget(widget);
+        }
+
+        this.poiManagerWidgets.clear();
+
+        if (Math.min(pageHeight, waypoints.size() - pageHeight * page) == 0 && page != 0) {
+            previousPage();
+        }
+
+        for (int i = 0, lim = Math.min(pageHeight, waypoints.size() - pageHeight * page); i < lim; i++) {
```
kristofbolyai: Why set lim to a variable? Just inline it in the comparasion

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+        if (delta < 0.0 && nextButton.active) {
+            nextPage();
+        } else if (delta > 0.0 && previousButton.active) {
+            previousPage();
+        }
+
+        return true;
+    }
+
+    public void populatePois() {
```
kristofbolyai, line 185: Shouln't you update `waypoints` here to the new value?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+        for (AbstractWidget widget : poiManagerWidgets) {
+            this.removeWidget(widget);
+        }
+
+        this.poiManagerWidgets.clear();
+
+        if (Math.min(pageHeight, waypoints.size() - pageHeight * page) == 0 && page != 0) {
+            previousPage();
```
kristofbolyai, line 196: Shouldn't you `page` to `waypoints.size() / pageHeight` to make sure you are on the last page, properly?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+
+public class PoiManagementScreen extends WynntilsScreen {
+    private final MainMapScreen oldMapScreen;
+    private Button nextButton;
+    private Button previousButton;
+    private Button deleteAllButton;
+    private int deleteCount = 3;
+    private List<CustomPoi> waypoints;
+    private int pageHeight;
+    private int page;
```
kristofbolyai: Rename this to `currentPage`

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -0,0 +1,243 @@
...
+    private int spacingMultiplier = 20;
```
kristofbolyai: If you say this value works for all GUI scales, then inline it.

common/src/main/java/com/wynntils/utils/render/Texture.java
ShadowCat117: Is it meant to be alphabetical? Because if so these should be swapped too
`MAP_DEFENSE_FILTER_BUTTON("map/map_defense_filter_button.png", 16, 16),
     MAP_BUTTONS_BACKGROUND("map/map_buttons_background.png", 144, 26),`

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: I don't think so, `waypoints.size() - pageHeight * page` calculates how many more pois there are to be displayed on this and the following pages. So with my GUI scale it displays 13 per page and if my final page only has 2 pois on the penultimate page then it the value will be 15 and on the final page it will be 2.

So if I deleted the 2 elements on the final page the min of those 2 values would be 0 and it will send you to the previous (now last) page.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
ShadowCat117: Well I know that's not the problem but I do know what is. It happens when all icons on the same page are the same type.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: I've tried that but it seems to make some icons randomly dissappear 

https://user-images.githubusercontent.com/8337467/228920141-9e7d07c0-902a-419e-9846-34b7bfa838bb.mp4

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
ShadowCat117: I've added the undo last delete button, but it only stores 1 deletion, do you think it would be a good idea to instead store a list of deleted pois so then you can undo everything you've deleted in the current "session".

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: Then, there is something really wrong..

magicus: Yes, that is worrisome. Overall, the PR looks pretty okay now, but the root cause of the disappearing icons must be found.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: No, I think that is enough. It's a bit of a safety, but we're not overdoing it.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
magicus: There might be a problem with the poi.renderAt() method. Hm, interesting...

magicus: I don't think you can use poi.renderAt() here. It is not supposed to work like that. For instance, it takes the mapZoom value and checks the transparency of the icon at that zoom level.

magicus: You are hard-coding a zoom level of 1f, I guess T3 loot chests are invisible at that zoom level...

kristofbolyai: > There might be a problem with the poi.renderAt() method. Hm, interesting...

I suggested that.. We used it at world distance feature. Maybe it works differently..

magicus: I'm not sure what a good solution would be. Either add a renderForGui() in Poi.java, or if your pois are only IconPoi's, call getIcon() and render it yourself.

kristofbolyai: > I'm not sure what a good solution would be. Either add a renderForGui() in Poi.java, or if your pois are only IconPoi's, call getIcon() and render it yourself.

You have to handle colors, so method in IconPoi.

ShadowCat117: For now I've just changed it back to using the same icon rendering method that PoiCreationScreen uses. I did try making a renderForGui in IconPoi where the alpha was always 1 and setting z to 1 for all pois but it just displayed exactly the same as renderAt.

If this is acceptable for now and you don't see any other areas I can improve I guess this'll be ok. Also not sure if you've seen the comment I have above this one, I'm not 100% sure if that's ok to have the way it is

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: Why did you resolve this, but not remove it?

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
kristofbolyai: Here, and below.. resolved but not changed?

common/src/main/java/com/wynntils/utils/render/Texture.java
kristofbolyai: Yes, please sort that too. It was overlooked in the PR it was added in.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: I have removed it?

https://github.com/Wynntils/Artemis/pull/1382/commits/ff6e943cb14eedd26a690b5d88abf33027b0105a#diff-908a218ec807dddcff8aababd8ececfbd7171a996e737521ef43a7fffb7b9ea6L182

kristofbolyai: @JamieCallan117 Sorry, my github was high..

kristofbolyai:
Pressing down button crashes. That down button should also be disabled.
image

kristofbolyai:
I have no idea what happened with the merges.. Seems like you are out of sync with main and the build fails. Please fix it.

common/src/main/java/com/wynntils/screens/maps/widgets/PoiManagerWidget.java
kristofbolyai: Looks good in-game now.

kristofbolyai (approved):
LGTM

ShadowCat117:
Looks like my gradle is updating, that might be the issue, didn't let it finish beforming commiting those changes

magicus:
> Pressing down button crashes. That down button should also be disabled. image

Was this resolved? It looked okay in the code, but I'd like to get it confirmed.

magicus (approved):
Code looks good. Thanks @JamieCallan117 for your contribution, and for having the patience of polishing it to match our code standards and requirements! This will be a welcome addition to Artemis.

ShadowCat117:
> > Pressing down button crashes. That down button should also be disabled. image
> 
> Was this resolved? It looked okay in the code, but I'd like to get it confirmed.


Yes!

ShadowCat117:
> Code looks good. Thanks @JamieCallan117 for your contribution, and for having the patience of polishing it to match our code standards and requirements! This will be a welcome addition to Artemis.

No problem! Honestly should be thanking you for reviewing all my code. It's my first time contributing to an open source project and a Minecraft mod so a lot of simple mistakes, hopefully if I add anything else it won't have so many

magicus:
I'm glad you've started your learning curve towards contributing to open source project. You're more than welcome to keep adding functionality to Wynncraft in the future!

PR #1383, opened 2023-03-26
feat: Introduce UnknownGearAnnotator

magicus:
This will make as much information as possible available about new items that has been added to the game, but not the Wynncraft API.

Also fixes some old limitations of CraftedGear, and a bad trinket regex causing Festival of the Hero trinkets to fail to parse.


common/src/main/java/com/wynntils/models/gear/type/GearType.java
```diff
@@ -5,34 +5,65 @@
...
+            Items.LEATHER_BOOTS,
+            0,
+            List.of(Items.CHAINMAIL_BOOTS, Items.IRON_BOOTS, Items.GOLDEN_BOOTS, Items.DIAMOND_BOOTS)),
     MASTERY_TOME(null, Items.ENCHANTED_BOOK, 0),
     CHARM(null, Items.CLAY_BALL, 0);
 
     private final ClassType classReq;
     private final Item defaultItem;
     private final int defaultDamage;
+    private final List<Item> otherItems;
```
kristofbolyai, line 56: Well.. How well does this work with reskinned items? Did we ever fix the problem with reskinned items not parsing (which is arguably much more important than this IMO)

common/src/main/java/com/wynntils/models/items/annotators/game/UnknownGearAnnotator.java
```diff
@@ -0,0 +1,32 @@
...
+import com.wynntils.handlers.item.ItemAnnotation;
+import com.wynntils.handlers.item.ItemAnnotator;
+import com.wynntils.models.gear.type.GearTier;
+import com.wynntils.models.gear.type.GearType;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.world.item.ItemStack;
+
+public final class UnknownGearAnnotator implements ItemAnnotator {
+    private static final Pattern UNKNOWN_GEAR_PATTERN = Pattern.compile("^§[5abcdef](.*)$");
```
kristofbolyai, line 17: I hope this is strict enough. Probably is knowing all the checks we do before this annotator

common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
```diff
@@ -37,20 +36,11 @@ public static void sendInventorySlotMouseClick(int slotNumber, MouseClickType mo
     }
 
     public static boolean isWeapon(ItemStack itemStack) {
-        Optional<GearItem> gearItemOpt = Models.Item.asWynnItem(itemStack, GearItem.class);
-        if (gearItemOpt.isPresent()) {
-            return gearItemOpt.get().getGearInfo().type().isWeapon();
-        }
+        Optional<GearTypeItemProperty> gearItemOpt =
+                Models.Item.asWynnItemPropery(itemStack, GearTypeItemProperty.class);
+        if (!gearItemOpt.isPresent()) return false;
```
kristofbolyai:
```suggestion
        if (gearItemOpt.isEmpty()) return false;
```

kristofbolyai (approved):
If you can confirm this works with reskinned items (or that fact that we still don't support them, so this is not a regression. If this is the case, I would reaaaaaaally like that to be fixed) then you can merge.

common/src/main/java/com/wynntils/models/gear/type/GearType.java
magicus: I'm not sure what you mean that reskinned items do not parse? For me, normal items that are reskinned parse just fine.

And no, the combination of an unknown gear (that is not in Wynncraft's API) and reskinning will make it impossible for us to detect, since it changes type to diamond_shovel, regardless of what gear type it is... So then we are just shit out of luck.

common/src/main/java/com/wynntils/models/items/annotators/game/UnknownGearAnnotator.java
magicus: That is the reason this is the antepenultimate annotator we call...

magicus:
> If you can confirm this works with reskinned items (or that fact that we still don't support them, so this is not a regression. If this is the case, I would reaaaaaaally like that to be fixed) then you can merge.

There is no regression here.

What is added is a way to deal with items that are not present in the database. This will not work with reskinned items, but otoh, these do not work without this PR, neither reskinned or not . So it is a limitation in this PR, but not a regression. Unfortunately, this is a limitation we cannot work around. :( And basically, not adding items to the API is a problem from Wynncraft's side. They ought to add new items to their DB. We are limited in several ways until they do so.

PR #1388, opened 2023-03-28
fix: Tentative fix for isWeapon crash on reskinned crafted weapons

magicus:


ryanzhoudev (approved):
From my limited testing in-game, seems like this works for skinned + signed + renamed items

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -100,6 +100,16 @@ public CraftedGearItem parseCraftedGearItem(ItemStack itemStack) {
         WynnItemParseResult result = WynnItemParser.parseItemStack(itemStack, null);
         CappedValue durability = new CappedValue(result.durabilityCurrent(), result.durabilityMax());
         GearType gearType = GearType.fromItemStack(itemStack);
+        if (gearType == null) {
+            // If it is crafted, and has a skin, then we cannot determine weapon type from item stack
+            // Maybe it is possible to find in the string type, e.g. "Crafted Wand"
+            gearType = GearType.fromString(result.itemType());
+            if (gearType == null) {
+                // ... but if the item is signed, this will not work either. We're out of luck,
```
kristofbolyai, line 108: I am confident skins only apply in appropriate slots. So armor won't have skins in your hand. I think..

ryanzhoudev: Yea, looks like this is the case in-game

magicus: We can probably get around that too, yes. Including checking the player's class to determine the kind of weapon. That would need us to break through several abstractions, though, and I did not like that. Currently, the only thing we need the gearType for anyway is to determine if we should start a spell cast sequence, or to display like the power charge overlay, so there is very little risk in assuming that a unknown item is a weapon. 

(In fact, if you were to move your armor to your hotbar, the skin would disappear)

PR #1389, opened 2023-03-28
fix: The config UI reset button has stopped working

magicus:
This is a regression, most likely caused by the text input widget changes


ryanzhoudev (approved):

kristofbolyai:
I am pretty sure the fix masks the actual problem, I will check this in the upcoming days

magicus:
I don't think so. Or rather, if anything, the problem is with the text edit widget, that needs adjusted mouse coordinates, rather than accepting the coordinates as passed into the method. That seem to break the Minecraft practice of what these coordinates refer to.

PR #1390, opened 2023-03-28
feat: Enum improvements

magicus:
This PR contains two related but yet sort of different changes. Sorry, but it was too hard to untangle them.

1) All enums are renamed to use the UPPER_CAMEL standard.

2) Our "import/export" of enums have been improved. That is, when reading and writing config files. in the `/config` command and in the settings screen. New behavior:

* In the json config file, and the `/config` command, the enums will be converted to lowerCamel, to better fit in with the identifiers in the surrounding context.
* When printing the value of an enum with `/config`, or in the toggle-button in the config GUI, a "nice" name is shown instead. This is the value of `toString()` if it is overridden (we have that for some enums, but currently not for any enum that is a Config), or the UPPER_CAMEL name is automatically translated to Title Case.
* Enums (and booleans) now get tab completion for `/config set`.

An Upfixer has been written to transition UPPER_CAMEL config to lowerCamel. Some changes to the Upfixer framework were necessary. I also made some not strictly necessary changes, wherein the upfixer name is derived automatically (and so that too is lowerCamel, which fits better in the config file). That means that existing upfixers will run again, since they now have a new name. I inspected the code and found that to be okay for all but one, which has therefore kept its old name.


magicus:
Note that this depends on #1399, which is incorporated in this PR. It might be easier to review this once that has been integrated.

magicus:
I realize this is a big PR, but it touches several interrelated things and I find it hard to see a way how to split this in several PRs. Much of the "meat" of the PR is concentrated to a few files, and most changes are just enum renames that can be trivially skipped over.

kristofbolyai (approved):
Looks good, I could not find anything strange.

PR #1391, opened 2023-03-28
feat: Embellishment feature: Wybel sounds

magicus:
Are you tired of just *seeing* your cute wybels purr in chat? Fret no more! Now you can actually hear them purr and squeak, for real.


ryanzhoudev (approved):
Tried with /kill, works great

magicus:
@DonkeyBlaster Unfortunately, there is currently no special sound for "sad squeak". I'm no sound effect expert, these four were all I could muster to design. But it would be nice if we could expand the wybel sound repertoire a bit. 

For future reference, here is a list of all pet sounds send to chat: https://wynncraft.fandom.com/wiki/Pets/Dialogue

PR #1393, opened 2023-03-29
feat: More container scrolling

ryanzhoudev:
Works for bank, guild bank, misc bucket, housing block bank, and trade market. Also ensures strange behaviour (buying new pages) will not happen with scrolling.


common/src/main/java/com/wynntils/features/ui/ContainerScrollFeature.java
```diff
@@ -0,0 +1,82 @@
...
+            slot = up ? tradeMarketPreviousSlot : tradeMarketNextSlot;
+            if (slot == tradeMarketPreviousSlot
+                    && ComponentUtils.getCoded(gui.getMenu()
+                                    .getSlot(tradeMarketPreviousSlot)
+                                    .getItem()
+                                    .getHoverName())
+                            .equals("§bReveal Item Names")) return;
+        } else {
+            return;
+        }
```
kristofbolyai: This `if` tree is really hard to read. I suggest using `if ... return` so you reduce right shift by one. For the item slot checks, maybe using screen name is better? Or keep it as is, BUT extract it to ContainerModel methods.

common/src/main/java/com/wynntils/features/ui/ContainerScrollFeature.java
```diff
@@ -0,0 +1,71 @@
...
+import com.wynntils.utils.mc.McUtils;
+import com.wynntils.utils.wynn.ContainerUtils;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@ConfigCategory(Category.UI)
+public class ContainerScrollFeature extends Feature {
+    private static final int abilityTreePreviousSlot = 57;
```
magicus, line 24: please use UPPER_CAMEL for constants (static final)

common/src/main/java/com/wynntils/features/ui/ContainerScrollFeature.java
```diff
@@ -0,0 +1,71 @@
...
+    @SubscribeEvent
+    public void onInteract(MouseScrollEvent event) {
+        Screen screen = McUtils.mc().screen;
+
+        if (!(screen instanceof AbstractContainerScreen<?> gui)) return;
+        boolean up = event.isScrollingUp() ^ invertScroll.get();
+
+        int slot;
+
+        if (Models.Container.isAbilityTreeScreen(gui)) {
```
magicus, line 48: I recommend you store the prev/next slot numbers as a pair, then you can separate this into two actions: first get the proper pair depending on the container, and then the a() or b() part of the pair, depending on the scroll direction. Also, please consider moving the slot constants to ContainerModel.

magicus (approved):

PR #1394, opened 2023-03-29
feat: Functions for mob totem x, y, z

ryanzhoudev:


magicus:
I think we should introduce Location as a primitive type for our functions, then MobTotem can return the location, and we can have that printed nicely if we accept our formatting, or extrac specific values like so `x(mobtotem(1)) - z(mobtotem(1))`. 

But we can do this in the meantime.

magicus (approved):

PR #1395, opened 2023-03-29
fix: NPC dialogue feature crashes when enabled after starting disabled

ryanzhoudev:


kristofbolyai (commented):
Why not make `currentDialogue` start as an empty array just like `confirmationlessDialogues`?

magicus (approved):

PR #1396, opened 2023-03-29
feat: Auto guild bank search

ryanzhoudev:
Has a recursive delay mechanism, as the guild bank is the only one that has partial inventory data sent sometimes.


common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -125,6 +135,15 @@ public void onInventoryKeyPress(InventoryKeyPressEvent event) {
 
     private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
         if (!autoSearching) return;
+        if (currentSearchableContainerType == SearchableContainerType.GUILD_BANK) {
+            if (System.currentTimeMillis() - guildBankLastSearch < GUILD_BANK_SEARCH_DELAY) {
+                Managers.TickScheduler.scheduleLater(
+                        () -> tryAutoSearch(abstractContainerScreen),
+                        (int) (GUILD_BANK_SEARCH_DELAY - (System.currentTimeMillis() - guildBankLastSearch)) / 20);
```
kristofbolyai: `System.currentTimeMillis() - guildBankLastSearch` needs to be extracted to prevent weird races where you could get - ticks.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -125,6 +135,15 @@ public void onInventoryKeyPress(InventoryKeyPressEvent event) {
 
     private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
         if (!autoSearching) return;
+        if (currentSearchableContainerType == SearchableContainerType.GUILD_BANK) {
+            long diff = System.currentTimeMillis() - guildBankLastSearch;
+            if (diff < GUILD_BANK_SEARCH_DELAY) {
+                Managers.TickScheduler.scheduleLater(
+                        () -> tryAutoSearch(abstractContainerScreen), (int) (GUILD_BANK_SEARCH_DELAY - diff) / 20);
```
magicus: A tick is 50 ms, so there are 20 ticks per second. So to get the amount of ticks represented by the time, you need to divide by 50, not 20.

If you have tested this code with the constant 200, what you have likely tested then is 10 ticks. This means that when you fix the division, you need to change the constant to 500.

magicus (approved):

PR #1398, opened 2023-03-29
feat: More robust handling of null in config file

magicus:
If the config file accidentally contains a null value, it will now be replaced by the default value, instead of causing a crash later on.


magicus:
I will fold this into #1390 since they overlap.

PR #1399, opened 2023-03-29
feat: Remove @TypeOverride

magicus:
Replace this by proper detection of generic types. This cannot be done in runtime on an object due to type erasure, but it can be extracted from the declaring field, which we actually check.

Also wrap the type so declared abstract collection types are replaced by concrete implementations suitable for json serialization.


kristofbolyai:
I am way tooo tired to review this.. This is gonna have to wait for tomorrow.

magicus:
No stress.

kristofbolyai (approved):

PR #1405, opened 2023-03-30
refactor: Clean up ContainerScrollFeature and ContainerModel

kristofbolyai:


common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -114,4 +115,30 @@ public Matcher lootChestMatcher(Screen screen) {
         return LOOT_CHEST_PATTERN.matcher(
                 WynnUtils.normalizeBadString(ComponentUtils.getUnformatted(screen.getTitle())));
     }
+
+    public Pair<Integer, Integer> getScrollSlots(AbstractContainerScreen<?> gui, boolean scrollUp) {
```
magicus: Since you get the scrollUp input, you don't need to return a pair but can just return the requested scroll slot itself ...

magicus:
Overall nice cleanup, thank you so much. Maybe I was a bit too trigger-happy to approve Donkey's PR, but then again, I think it is a fairly okay development process of not stalling PRs too much and then refactor afterwards.

I think we could/should let ContainerModel grow quite a bit. We have a lot of "distributed" container knowledge throughout the code, with pieces of code checking container titles with regexes, etc. That should all move into ContainerModel. I've been thinking about it for a while but haven't gotten around to fixing it.

magicus (approved):

PR #1406, opened 2023-03-30
feat: Make reset button grayed-out if no change has been made

magicus:
This makes it easy to spot if a setting is left at default or has been changed by the user


magicus:
This is still a bit wonky, though. The click sound is played, even though the button does not do anything. To solve this, we have to use the `active` attribute in AbstractWidget. While this is certainly the "right" way to go, a lot of other functionality needs to be adapted on our side for that, so I took the easy route out.

kristofbolyai (approved):

PR #1409, opened 2023-03-30
feat: Remove secondsPerRecalculation for text/bar overlays, update on tick

kristofbolyai:


kristofbolyai:
There was no real usage of secondsPerRecalculation, instead it created weird behaviour in many places. Performance wise, this is the same thing.

kristofbolyai:
(Maybe we should not cache at all.. But one step at a time)

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
```diff
@@ -52,20 +52,20 @@ public void render(PoseStack poseStack, MultiBufferSource bufferSource, float pa
...
         BarOverlayTemplatePair template = getTemplate();
-        render(poseStack, bufferSource, template.textTemplate, template.valueTemplate);
+        render(poseStack, bufferSource, templateCache);
     }
 
     @Override
     public void renderPreview(PoseStack poseStack, MultiBufferSource bufferSource, float partialTicks, Window window) {
-        BarOverlayTemplatePair template = getPreviewTemplate();
-        render(poseStack, bufferSource, template.textTemplate, template.valueTemplate);
+        BarOverlayTemplatePair previewTempalte = getPreviewTemplate();
```
magicus: typo

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
```diff
@@ -85,13 +85,19 @@ protected void render(
...
     }
 
-    protected void updateCachedLines(String textTemplate, String valueTemplate) {
-        if (System.currentTimeMillis() - lastUpdate > secondsPerRecalculation.get() * 1000) {
-            lastUpdate = System.currentTimeMillis();
-            templateCache = Pair.of(
-                    String.join(" ", Managers.Function.doFormatLines(textTemplate)),
-                    Managers.Function.tryGetRawValueOfType(valueTemplate, CappedValue.class));
-        }
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
```
magicus: Why the priority? Are we racing with others listening on the tick?

kristofbolyai: No, I just thought this was worth to be moved up the order. It can be removed.

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
kristofbolyai: fixed

common/src/main/java/com/wynntils/core/features/overlays/BarOverlay.java
magicus: Yes, please, remove it then (on both places), so we only have priority ordering where it is strictly needed, for performance and/or correctness.

magicus (approved):

PR #1410, opened 2023-03-30
chore: Update dependencies to latest versions

magicus:
Also stop depending on unclear "SNAPSHOT" versions

This depends on #1408


gradle.properties
```diff
@@ -8,14 +8,14 @@ minecraft_version=1.19.3
 
 # Architectury
 # Check for latest at https://maven.architectury.dev/architectury-plugin/architectury-plugin.gradle.plugin
-architectury_plugin_version=3.4-SNAPSHOT
+architectury_plugin_version=3.4.145
 # Check for latest at https://maven.architectury.dev/dev/architectury/loom/dev.architectury.loom.gradle.plugin
-architectury_loom_version=1.0-SNAPSHOT
+architectury_loom_version=1.0.312
```
kristofbolyai: I really don't like this. It is almost recommended by architectury that you always use snapshots. You wouldn't want to update these manually either.. IMO at least.

magicus: I never liked the fact that we depend on an unknown external identity, that can change between two different builds of the same source code. That is a recipe for hard-to-trace bugs.

I have a suggestion: what if I add a github action bot that tracks architectury, and updates our property file to track the latest version? That way we will get the bonus of always using the latest version (as with SNAPSHOT), but also get the stability that the code we depend on will not change under our feet?

kristofbolyai: Sounds good

kristofbolyai (approved):

PR #1419, opened 2023-04-01
feat: Preload our resource pack earlier to skip reload

magicus:
This annoyed me to no end. We did not preload the Wynncraft resource pack until too late, so all textures had to be reloaded when starting Minecraft.

It turned out that to be able to inject our code early enough, a Feature was not enough, since Features are not loaded until i18n are available, which is *after* resource packs have been loaded. Catch 22. So a Manager was needed. Then it turned out that we do not populate storage until after Features has been loaded, so the Manager did not have access to the persisted value until after Features had loaded anyway... So it was a bit of a chase, getting everything in place.


common/src/main/java/com/wynntils/mc/mixin/DownloadedPackSourceMixin.java
```diff
@@ -32,4 +33,14 @@ private void onClearServerPackPre(CallbackInfoReturnable<CompletableFuture<Void>
...
+
+    @Override
+    public void setServerPack(Pack pack) {
+        serverPack = pack;
+    }
+
+    @Override
+    public Pack getServerPack() {
+        return serverPack;
+    }
```
kristofbolyai: In any other place we would have used an AW to make this public and write/read the field. I am not sure what is different here.

kristofbolyai (approved):
Looks good, but if there is nothing holding back us from using an AW, IMO we should keep doing things like we used to.

common/src/main/java/com/wynntils/mc/mixin/DownloadedPackSourceMixin.java
magicus: Yeah, good point. I went through like 3-6 different solutions on how to achieve what I wanted, so when in the end I found a working solution, I did not think this through fully.

PR #1420, opened 2023-04-01
feat: Rewrite ScoreboardHandler

kristofbolyai:
The new handler is more complete than the previous version, while maintaining better code quality. This should fix some long existing issues with scoreboard related features, while also being much faster (thanks to removing the 250 MS delay, which was huge).


common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -4,346 +4,361 @@
...
-                }
-            } else {
-                ScoreboardLine line = new ScoreboardLine(processed.lineText(), processed.lineIndex());
-                scoreboardLineMap.put(line.index(), line);
-                changedLines.add(processed.lineText());
-            }
-        }
+    @SubscribeEvent
+    public void onSetObjectiveDisplay(ScoreboardSetDisplayObjectiveEvent event) {
+        event.setCanceled(true);
```
magicus, line 82: I'm still a bit lost at the minecraft scoreboard packets. What that this poor one do, that causes it to be always canceled?

kristofbolyai: This sets the scoreboard to be seen on the screen. But we want to display ours. So we have to ignore Wynn.

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardSegment.java
```diff
@@ -4,88 +4,65 @@
...
 
-    public boolean isChanged() {
-        return changed;
+    public boolean isVisible() {
+        return visible;
     }
 
-    public void setContent(List<String> content) {
-        this.content = content;
+    public void setVisibility(boolean visible) {
```
magicus, line 39: Just checking here. Had it not been for visible, this could have been an immutable record?

common/src/main/java/com/wynntils/handlers/scoreboard/event/ScoreboardSegmentAdditionEvent.java
```diff
@@ -4,13 +4,11 @@
  */
 package com.wynntils.handlers.scoreboard.event;
 
-import com.wynntils.core.events.EventThread;
 import com.wynntils.handlers.scoreboard.ScoreboardSegment;
 import net.minecraftforge.eventbus.api.Cancelable;
 import net.minecraftforge.eventbus.api.Event;
 
 @Cancelable
-@EventThread(EventThread.Type.WORKER)
```
magicus, line 13: Oooh, this is nice. Getting rid of one more source of races.

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardSegment.java
kristofbolyai: Probably. Does it really change anything?

common/src/main/java/com/wynntils/models/objectives/DailyObjectiveScoreboardPart.java
```diff
@@ -0,0 +1,42 @@
...
+                Models.Objectives.updatePersonalObjective(objective);
+            }
+        }
+
+        // filter out deleted objectives
+        Models.Objectives.purgePersonalObjectives(objectives);
+    }
+
+    @Override
+    public String toString() {
```
magicus, line 39: Do we need a toString for this and the other scoreboard part classes?

kristofbolyai: Not strictly. It is more like a debugging measurement. Just like how many item related classes do.

common/src/main/java/com/wynntils/models/players/scoreboard/PartyFinderScoreboardPart.java
```diff
@@ -0,0 +1,38 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.players.scoreboard;
+
+import com.wynntils.handlers.scoreboard.ScoreboardPart;
+import com.wynntils.handlers.scoreboard.ScoreboardSegment;
+import com.wynntils.handlers.scoreboard.type.SegmentMatcher;
+
+public class PartyFinderScoreboardPart extends ScoreboardPart {
+    private static final SegmentMatcher PARTY_FINDER_MATCHER = SegmentMatcher.fromPattern("Party Finder:");
```
magicus, line 12: Is there a Party Finder scoreboard part? I've never seen that? How, when does it appear?

kristofbolyai: Whenever you queue for a party finder queue. 
image

magicus (approved):
Basically looks fine. I had a handful of questions, if we sort them out I think this is fine.

common/src/main/java/com/wynntils/models/objectives/DailyObjectiveScoreboardPart.java
magicus: Since it has no fields, the default toString will still print the class name, I think that's fine, and it makes the code a bit more compact.

kristofbolyai: It will print the fully qualified name though.. So `com.wynntils.models.objectives.DailyObjectiveScoreboardPart`.

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
magicus: Ah, I see. I think I see, at least. For those scoreboard parts that are not handled by others, do we render them ourselves? I thought we modified the scoreboard but still let Vanilla render it, but I might be misunderstanding what I'm seeing.

kristofbolyai: We can render any part without processing. The only processing we do is setting visibility. If no visibility was changed, we basically clone the wynn scoreboard, and display it.

kristofbolyai: And we basically let vanilla render it, we just create a different display.

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardSegment.java
magicus: No, I guess not. It's nice to have immutable holder classes as records, if possible. When a situation occurs that a class is mostly immutable but has a mutable field, it is worth thinking once more if that field actually belongs in the class. In this case I think it's fine. 

(As an example of the opposite is QuestInfo, which should really be split into two parts -- an immutable info about the quest per se, and one mutable QuestProgress part. I have a branch doing this but it got stuck and has probably bit-rotted beyond repair by now)

common/src/main/java/com/wynntils/models/players/scoreboard/PartyFinderScoreboardPart.java
magicus: Oh, I guess I have never been queuing, only joining existing parties. At least not enough to think about that.

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
magicus: Ok, I think I get it now. Mostly. :) I think my main problem here is that I don't fully understand the vanilla mechanics, since I've only seen how Wynncraft mis-uses it. But I'm satisfied with my current level of understanding. :)

common/src/main/java/com/wynntils/models/objectives/DailyObjectiveScoreboardPart.java
magicus: Ok, fine, let's keep it.

PR #1423, opened 2023-04-01
fix: Fix translations feature crashes

kristofbolyai:
The bug is caused by design that I could not fully understand, so I patched it in a sane, but maybe not so design friendly way. The `TranslationService` classes used to return null to signal that the translation should not happen, and the original text should be used. However, none of the users of translation manager handled this (I think the method should return a boolean indicating success/fault, or have a separate error consumer..).


magicus (approved):
Yeah, the translation was ported straight from legacy, and I haven't given it as much love and attention as it has needed.

I think we need to plugin "translatability" into more components instead of just trying to blindly replace text. So things like the chat should "request" translation, rather than having it stuffed in their face.

PR #1431, opened 2023-04-02
fix: Clean up overlay enabled state checks

kristofbolyai:
When I renamed `isEnabled` to `shouldBeEnabled` in `Overlay`, I should have done this, but I did not notice. But it is clear with the new name, that the method is not something for state checks. The overlay itself does not know it's state. Now we use the manager to check for state (as we should).

### DIFF
--- a/common/src/main/java/com/wynntils/features/overlays/GuildAttackTimerOverlayFeature.java
+++ b/common/src/main/java/com/wynntils/features/overlays/GuildAttackTimerOverlayFeature.java
@@ -6,6 +6,7 @@ package com.wynntils.features.overlays;
 
 import com.mojang.blaze3d.platform.Window;
 import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.components.Managers;
 import com.wynntils.core.components.Models;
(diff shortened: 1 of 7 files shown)


magicus:
I'm not sure I fully understand. You are replacing one check with another, is that correct? Since the check we did was incorrect. So this is a bug fix?

kristofbolyai:
> I'm not sure I fully understand. You are replacing one check with another, is that correct? Since the check we did was incorrect. So this is a bug fix?

The old check tells if the overlay should be enabled **by default**. The new check returns the actual state, having accounted user enabled, crashes, parent disabled state etc.

kristofbolyai:
I also agree that the naming is not so great, but these are the correct checks to use. Fixing the naming is a different problem.

magicus:
So `overlay.shouldBeEnabled()` is like "this overlay is requesting, from its PoV, to be enabled"? I think I understand.

magicus (approved):

kristofbolyai:
> So `overlay.shouldBeEnabled()` is like "this overlay is requesting, from its PoV, to be enabled"? I think I understand.

Basically yes.

PR #1433, opened 2023-04-02
feat: Config search now properly finds overlays by name, and search only shows relevant configurables

kristofbolyai:


common/src/main/java/com/wynntils/core/features/Feature.java
```diff
@@ -97,10 +97,14 @@ private void tryUserToggle() {
     }
 
     @Override
-    public int compareTo(Feature other) {
+    public int compareTo(Object other) {
+        if (!(other instanceof Feature feature)) {
+            return 0;
```
magicus: I thought 0 means "is equal to"? This looks wrong. Should you not sort stuff so everything that is not a feature will come before, or after, all features?

common/src/main/java/com/wynntils/core/features/overlays/DynamicOverlay.java
```diff
@@ -78,15 +78,19 @@ public int getId() {
     }
 
     @Override
-    public int compareTo(Overlay other) {
+    public int compareTo(Object other) {
+        if (!(other instanceof Overlay overlay)) {
```
magicus: ... same here...

common/src/main/java/com/wynntils/core/features/Feature.java
kristofbolyai: I fixed the logic itself, this change was dumb to begin with.

magicus (approved):

PR #1434, opened 2023-04-02
feat: Add StyledText, rewrite ChatMentionsFeature and enable it

kristofbolyai:


common/src/main/java/com/wynntils/core/chat/transcoder/CodedStyle.java
```diff
@@ -0,0 +1,175 @@
...
+        // Rules of converting a Style to a String:
+        // 0. Every style is prefixed with a §.
+        // 1. Style color is converted to a color segment.
+        //    A color segment is the prefix and the chatFormatting char.
+        //    If this is a custom color, a hex color code is used.
+        //    Example: §#FF0000 or §1
+        // 2. Formatting is converted the same way as in the Style class.
+        // 3. Click events are wrapped in angle brackets, and is reprenseted as an id.
+        //    The parent of this style's owner is responsible for keeping track of click events.
+        //    Example: §<1> -> (1st click event)
```
magicus: Just for the sake of my sanity, can we swap click and hover? I'm thinking of `[1]` like a URL,e.g. like in markdown, so I think of that as an onclick event, rather than a hover.

kristofbolyai: Sure.

common/src/main/java/com/wynntils/core/chat/transcoder/CodedString.java
```diff
@@ -0,0 +1,118 @@
...
+
+        parts.forEach(part -> component.append(part.getComponent()));
+
+        componentCache = component;
+
+        return component;
+    }
+
+    public CodedStringPart getPartBefore(CodedStringPart codedStringPart) {
+        int index = parts.indexOf(codedStringPart);
```
magicus: This depends on two CodedStringPart with the same content still not being "equal", like if you have a string: `§cHey §cHey §6Yo!`. Sounds a bit scary. I'm still trying to figure out where this is used, and if it can be solved differently.

kristofbolyai: Why though? Every part is in the list, I am not sure how this is a problem.

magicus: I think you should construct the CodedStringPart with its index instead, and use that. It seems more robust.

kristofbolyai: But the part index can change when we modify the whole string. I might as well save the component before this part directly, but that can still change.. And I think it would be a pain to keep track.

common/src/main/java/com/wynntils/core/chat/transcoder/CodedStyle.java
```diff
@@ -0,0 +1,238 @@
...
+        this.bold = bold;
+        this.italic = italic;
+        this.underlined = underlined;
+        this.strikethrough = strikethrough;
+        this.obfuscated = obfuscated;
+        this.clickEvent = clickEvent;
+        this.hoverEvent = hoverEvent;
+    }
+
+    static CodedStyle fromStyle(Style style, CodedStringPart owner, CodedStringPart partBefore) {
```
magicus: Actually, I think you don't need a CodedStringPart here. All you need is a `CodedStyle accumulatedStyle`. I think it will be clearer if you limit it this way.

magicus: Or `previousStyle`, depending on how you look at it.

common/src/main/java/com/wynntils/core/chat/transcoder/CodedString.java
kristofbolyai: How we do this depends on the type of "modifying" operations we want to implement. Let's discuss that on discord.

common/src/main/java/com/wynntils/core/chat/transcoder/CodedStyle.java
```diff
@@ -0,0 +1,238 @@
...
+
+        return styleString.toString();
+    }
+
+    public Style getStyle() {
+        if (styleCache != null) {
+            return styleCache;
+        }
+
+        Style reconstructedStyle = Style.EMPTY
```
magicus: An alternative is to use AW to get to the private Style constructor. But maybe this is good enough, perhaps even better...

common/src/main/java/com/wynntils/core/chat/transcoder/CodedStyle.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        styleCache = reconstructedStyle;
+
+        return reconstructedStyle;
+    }
+
+    private boolean isCompatibleWith(CodedStyle other) {
+        // A style is compatible with another style if:
+        // 1. If the color is non-null in this style, it is the same as the other style or other style's color is null.
```
magicus: I'm not sure I get this logic. If we have `"red".withStyle(RED).append("blue".withStyle(BLUE))`, then you would say they are not "compatible", and thus you would issue a `§r`. But that will not be necessary..? In fact, I think you *only* need to issue a reset (basically, what you call "incompatible") if a style goes from true to false. Color can always be replaced.

(This comment ties in with how this is used in asString as well)

kristofbolyai: I just fixed this before you types this out :). But yes, you are right.

kristofbolyai: And no, there is a case where you have whatever color, then none color. Then you need to reset.

common/src/main/java/com/wynntils/core/chat/transcoder/CodedStyle.java
kristofbolyai: This is likely better.

kristofbolyai:
This is basically ready. It is missing unit tests; I will likely add those separately (I think this will make it into main faster). 

Only discussing that needs to be done is about removing the `FULL` style component hack, it seems like we are now on parity with ComponentUtils, using the correct implementation. I started discussions about this in Discord.

common/src/main/java/com/wynntils/core/chat/tab/ChatTab.java
```diff
@@ -2,7 +2,7 @@
  * Copyright © Wynntils 2022.
  * This file is released under AGPLv3. See LICENSE for full license details.
  */
-package com.wynntils.core.chat;
+package com.wynntils.core.chat.tab;
```
magicus: I think these can move back, for now. We have unfinished business with chat, but lets save that for another day.

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -0,0 +1,298 @@
...
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.Style;
+
+public final class PartStyle {
+    private static final String STYLE_PREFIX = "§";
+
+    private final StyledTextPart owner;
+
+    private CustomColor color;
```
magicus: It seems you made this immutable now, yes? If so, these can be final.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -61,69 +61,42 @@ private Pattern buildPattern() {
...
-        // if current component has text check it for mentions and rewrite the component if necessary
-        if (!text.isEmpty()) {
-            curr = rewriteMentions(curr, text, comp.getStyle());
-        }
+        StyledTextPart partFinding = styledText.getPartFinding(mentionPattern, PartStyle.StyleType.NONE);
 
-        // process and append siblings
-        for (Component c : comp.getSiblings()) {
-            curr.append(rewriteComponentTree(c));
+        // If the part is null, it means that the mention is in two different parts, we likely don't want to mark it.
```
magicus: .. or it means the chat message did not mention the user..? :)

kristofbolyai: Yes :). This code was shuffled around quite a bit, the comment was left a bit outdated.

fabric/src/test/java/TestStyledText.java
```diff
@@ -0,0 +1,413 @@
...
+    public void blankText_shouldProduceBlankString() {
+        StyledText styledText = StyledText.fromString("    ");
+
+        Assertions.assertTrue(styledText.isBlank(), "StyledText.isBlank() returned an unexpected value.");
+    }
+
+    @Test
+    public void contains_shouldProduceCorrectResult() {
+        final String testString = "§c§oitalicred§9§oblue§cnonitalic§oinherited§lbold§rafter";
+        final String[] expectedParts = {"§c§oitalicred", "§9§oblue", "§cnonitalic", "§oinherited", "§lbold", "§rafter"};
```
magicus: It would be interesting to test like `"icred§9§obl"`, and not just at part breaks.

fabric/src/test/java/TestStyledText.java
```diff
@@ -0,0 +1,413 @@
...
+                    "StyledText.contains(NONE) could not find a part matching: " + expectedPart);
+        }
+    }
+
+    @Test
+    public void startsWith_shouldProduceCorrectResult() {
+        final String testString = "§c§oitalicred§r§9§oblue§r§cnonitalic§oinherited§lbold§rafter";
+        final String expectedStart = "§c§oitalicred";
+
+        StyledText styledText = StyledText.fromString(testString);
```
magicus, line 211: Same here and for endsWith, we want to make sure it also works across part breaks, so like "after" and "old§rafter" as well. And perhaps, what should "" return? true or false? Not terribly important, but looking for edge cases is a great way of honing your test writing skills...

kristofbolyai: I can add the special cases. About "", it will return what the string impl does. So we get the result we would like to "follow".

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -0,0 +1,323 @@
...
+    public String asString(PartStyle previousStyle, StyleType type) {
+        // Rules of converting a Style to a String:
+        // Every style is prefixed with a §.
+        // 0. Every style string is fully qualified, meaning that it contains all the formatting, and reset if needed.
+        // 1. Style color is converted to a color segment.
+        //    A color segment is the prefix and the chatFormatting char.
+        //    If this is a custom color, a hex color code is used.
+        //    Example: §#FF0000 or §1
+        // 2. Formatting is converted the same way as in the Style class.
+        // 3. Click events are wrapped in square brackets, and is reprenseted as an id.
```
magicus:
```suggestion
        // 3. Click events are wrapped in square brackets, and is represented as an id.
```

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -0,0 +1,323 @@
...
+
+                if (chatFormatting.isPresent()) {
+                    styleString.append(STYLE_PREFIX).append(chatFormatting.get().getChar());
+                } else {
+                    styleString.append(STYLE_PREFIX).append(color.toHexString());
+                }
+            }
+
+            // 2. Formatting
+            if (bold) {
```
magicus, line 151: I think we should follow the alphabetical ordering used in Vanilla:

```
	OBFUSCATED("OBFUSCATED", 'k', true),
	BOLD("BOLD", 'l', true),
	STRIKETHROUGH("STRIKETHROUGH", 'm', true),
	UNDERLINE("UNDERLINE", 'n', true),
	ITALIC("ITALIC", 'o', true),
```

The order we define here will be canonical for all future, and I think this order makes more sense.

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -0,0 +1,323 @@
...
+        StringBuilder styleString = new StringBuilder();
+
+        boolean skipFormatting = false;
+
+        // If the color is the same as the previous style, we can try to construct a difference.
+        // If colors don't match, the inserted color will reset the formatting, thus we need to include all formatting.
+        // If the current color is NONE, we NEED to try to construct a difference,
+        // since there will be no color formatting resetting the formatting afterwards.
+        if (previousStyle != null && (color == CustomColor.NONE || previousStyle.color.equals(color))) {
+            StringBuilder stringBuilder = this.tryConstructDifference(previousStyle);
```
magicus: It's kind of odd to return a stringBuilder, just to append it to another stringBuilder :)

I recommend you send in the styleString stringbuilder as an argument to tryConstructDifference, and return a boolean saying if you did find a difference or not.

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public StyledText getParent() {
+        return parent;
+    }
+
+    public PartStyle getPartStyle() {
+        return style;
+    }
+
+    public void setPartStyle(PartStyle style) {
```
magicus: I do not like this. Why do you think our parts should be mutable? I'd rather see a "wither". 

I can possibly live with our StyledText being mutable, so we can swap out parts etc (like a list of immutable objects), but I don't think the parts or the style should be mutable.

And I'm frankly not sure that StyledText should be mutable as well. There is a case to be made for saying that if we want to modify a StyledText, we should replace it with a new, modified StyledText instead. I'm not sure about that one yet...

common/src/main/java/com/wynntils/core/text/PartStyle.java
kristofbolyai: I know, but doing that would require changes to the method, which would not turn out so great (since you append, and find return cases later). This means that we would lose the clarity of that method.

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
kristofbolyai: StyledText is immutable. This is the only part where we have some kind of mutability. I can't see any way of creating an API for "I want to change this parts style without actually leaking this part". Well, we could have `findMatchingPartAndReplaceStyleWith` but that is kind of weird..

common/src/main/java/com/wynntils/core/text/PartStyle.java
kristofbolyai: I think I fixed it in every location.

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
```diff
@@ -0,0 +1,139 @@
...
+    StyledTextPart(StyledTextPart part, StyledText parent) {
+        this.text = part.text;
+        this.style = part.style;
+        this.parent = parent;
+    }
+
+    // This factory is used to create a StyledTextPart from a string that has formatting codes
+    // It is separate from the constructor because this only needs to be applied in cases there the text could have
+    // formatting codes
+    static StyledTextPart fromStyledString(String text, Style style, StyledText parent, Style parentStyle) {
```
magicus: So maybe I'm missing something here since I have not looked at StyledText itself yet, but this sounds like it is on the wrong level. If you are creating a part of a string like "§cfoo", then I guess this works, but then you only need to check if the first character is §. But if you have a string like "§afoo§cbar" or even "foo§cbar", then you would in fact need to create two different StyledTextParts for that single component literal.

And it looks like this function sort-of tries to do that, but not correctly. I think it is the responsibility of StyledText to make sure it only tries to instantiate StyledTextParts that has the very basic minimal "atom" of a stretch of text -- without any formatting codes!  -- and a style that applies to the entire part.

magicus: If this comment seem totally wrong to you, ignore it and I'll continue reading the code tomorrow. But if it makes sense, feel free to go ahead and fix.

kristofbolyai: Well, this is the "handle if the component has formatting" code. You are correct in the regard that once we not find a format, we can "break" out of the loop.

This is NOT the code which is called from `StyledText#fromString`. There, it is handled correctly, and this is skipped, since there is no case where we have styles in any part.

But when parsing from components, or after doing string operations, like join, we need to check if we added new styles as string.

kristofbolyai: I pushed the bit more optimized, and more understandable code. I also renamed it `fromComponentWithPossibleFormattingCodes` so it is not confused anymore.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+    public StyledText getNormalized() {
+        return new StyledText(
+                parts.stream().map(StyledTextPart::asNormalized).collect(Collectors.toList()),
+                temporaryWorkaround,
+                clickEvents,
+                hoverEvents);
+    }
+
+    public StyledText trim() {
+        return new StyledText(
```
magicus: This is not correct. trim means removing trailing and leading whitespace for the entire text, not for each part.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+                parts.stream().map(StyledTextPart::trim).collect(Collectors.toList()),
+                temporaryWorkaround,
+                clickEvents,
+                hoverEvents);
+    }
+
+    public boolean isEmpty() {
+        return parts.stream().allMatch(StyledTextPart::isEmpty);
```
magicus: It's empty if there are no parts.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+
+    public boolean contains(String string, PartStyle.StyleType styleType) {
+        return getString(styleType).contains(string);
+    }
+
+    public boolean contains(StyledText styledText, PartStyle.StyleType styleType) {
+        return contains(styledText.getString(styleType), styleType);
+    }
+
+    public boolean startsWith(String string) {
```
magicus: Can you rename all these `String string` to  `String codedString` to clarify that this is what is expected here?

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: If the stream is empty, true is returned.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+
+public final class StyledText {
+    private final Component temporaryWorkaround;
+
+    private final List<StyledTextPart> parts;
+
+    private final List<ClickEvent> clickEvents;
+    private final List<HoverEvent> hoverEvents;
+
+    StyledText(Component component) {
```
magicus: I'm not fond of having complicated constructors. Can't you move this logic into the factory method instead, and when you have finished parsing the component, just call the "thin" StyledText constructor? (And similarly for the String)

magicus: (This is not just superstition; constructors are specially treated by the JLS, the object is not fully constructed but you can still leak out `this` which can cause very difficult bugs, etc. For a static method, it is verboten to use `this` so that cannot happen, the compiler will make sure you write correct code.)

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+        temporaryWorkaround = component;
+
+        parts = new LinkedList<>();
+        clickEvents = new LinkedList<>();
+        hoverEvents = new LinkedList<>();
+
+        // Walk the component tree using DFS
+        // Save the style of the parent component so we can inherit it
+        Deque<Pair<Component, Style>> deque = new LinkedList<>();
```
magicus, line 72: Why are we not using Component.visit()?

kristofbolyai: It behaves weirdly, has some unexpected cases. This is a much stable way of walking the tree (visit is using `Component#getString` which we know causes unexpected behavior)

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+
+        deque.add(new Pair<>(component, Style.EMPTY));
+
+        while (!deque.isEmpty()) {
+            Pair<Component, Style> currentPair = deque.pop();
+            Component current = currentPair.key();
+            Style parentStyle = currentPair.value();
+
+            String componentString =
```
magicus, line 82: Ok, this is wrong like I suspected but could not really say for sure before. What you really need to do here is basically call `StyledText.fromString(current.getString())`. Since this individual component element can have a style like RED, and a string like "warning: §kDANGER§r ahead", and in such a case, it will need to be replaced with not one, but three StyledTextParts.

magicus: (And prectically, you probably want to solve this by an additional static method like `getStyledTextParts(String codedString)`, which just gives you a List<StyledTextParts>, and that you can use both here and in `StyledText.fromString`.

kristofbolyai: No, this is not `Component#getString`. `MutableComponent.create(current.getContents())` gives us the special treatment of only getting THIS components string.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+                    MutableComponent.create(current.getContents()).getString();
+
+            StyledTextPart styledTextPart = StyledTextPart.fromComponentWithPossibleFormattingCodes(
+                    componentString, current.getStyle(), this, parentStyle);
+
+            List<Pair<Component, Style>> siblingPairs = current.getSiblings().stream()
```
magicus, line 91: This looks odd, too. Are you trying to do here what I describe above? But current cannot have any siblings, right? You just created it from current.getContents()).getString()...

magicus: Ah, now I understand. This is for your DFS traversal.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -241,11 +242,17 @@ public StyledText getNormalized() {
     }
 
     public StyledText trim() {
-        return new StyledText(
-                parts.stream().map(StyledTextPart::trim).collect(Collectors.toList()),
-                temporaryWorkaround,
-                clickEvents,
-                hoverEvents);
+        if (parts.isEmpty()) {
+            return new StyledText(List.of(), temporaryWorkaround, clickEvents, hoverEvents);
```
magicus: You can just return this, if it is empty it is already trimmed...

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -241,11 +242,17 @@ public StyledText getNormalized() {
...
+        }
+
+        List<StyledTextPart> newParts = new ArrayList<>(parts);
+        newParts.set(0, newParts.get(0).trim());
```
magicus: Unfortunately, it is not that simple :( This will also trim trailing spaces of the first segment. You need to have a "trimLeadingSpacedOnly" functionality here. Unless Apache COmmons has something for that, you most likely need to reimplement Java's trim function.

kristofbolyai: Java 11 for the win!

common/src/main/java/com/wynntils/core/text/StyledText.java
magicus: Let's take a step back here. My point was that isEmpty is true iff (if and only if) we have no parts.

But I now realize your code checks if we have an empty part somewhere in our part list. 

I think that should not be allowed. If we can have empty parts in the list, then there is no longer a 1-to-1 relationship with e.g. a coded string and a StyledText, since the styled text can have an arbitrary number of "hidden" empty parts in it.

So we need to make sure this cannot happen. And then this function once again reduced down to checking if we have no parts, since then by definition if we have any parts, they cannot be empty.

common/src/main/java/com/wynntils/core/text/StyledText.java
magicus: Please add a comment about why we are not using the vanilla provided functionality.

common/src/main/java/com/wynntils/core/text/StyledText.java
magicus: Yes, I understand that. You are getting THIS component's string. And now let's assume that this string is "warning: §kDANGER§r ahead". Then we need to create *three* StyledTextParts from it, right?

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: The check for no empty parts is already in place in suitable locations, when converting from string. I will add the check for components. Are you happy with `isBlank`?

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: I see what you mean now.. Yes, most likely..

common/src/main/java/com/wynntils/core/text/PartStyle.java
magicus: I see. I missed all `return null` except the first...  Maybe at least make tryConstructDifference return a String..?

Also, in general, for modern Java, StringBuild does not provide any performance gains compared to normal string operations (because of the string indyfy optimizations), so it only makes sense to use it if the code gets clearer from it. In this case I think it does for the "outer" asString() method, at least.

common/src/main/java/com/wynntils/core/text/PartStyle.java
magicus: Was there more locations than here that define the order?

kristofbolyai: tryConstructDifference iirc

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
magicus: But why should you change an existing part? You can create a new part based on an existing part + the new style that you want.

And then you can create a new StyledText based on the old one, but where the old part of interest was, is now the new part you have created.

(This is actually the one place where I might think mutability can be allowed -- that you can, like a List, be allowed to change what parts a StyledText consists of). 

That way you could do like:
```
int index = 0;

do {
index = styledText.findMatch(myRegexp, index);
if (index == -1) return;
StyledTextPart matchingPart = styledText.getPart(index);
StyledTextPart modifiedPart = fixup(matchingPart);
styledText.setPart(index, modifiedPart);
} while (true);
```

magicus: That is, basically let StyledTextPart implement Collection (not sure if we should *actually* do this, but at least those operations that makes sense).

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,487 @@
...
+
+    public MutableComponent getComponent() {
+        if (parts.isEmpty()) {
+            return Component.empty();
+        }
+
+        MutableComponent component = parts.get(0).getComponent();
+
+        for (int i = 1; i < parts.size(); i++) {
+            component.append(parts.get(i).getComponent());
```
magicus, line 137: You are setting parts to a LinkedList. This means this loop will be O(n^2). 

But I think this implementation is the correct one. What you should do is to not use LinkedList. It is worse than ArrayList in 99 cases out of 100, unless you *really* need to be able to pop of the head of the list, and make sure you know what you are doing.

common/src/main/java/com/wynntils/core/text/StyledText.java
magicus: Yeah, isBlank is a different story. We can have a sequence of differently colored spaces, and thus we need to check all of them for isBlank.

(In fact, stuff like this and the trim() issue I pointed out would make awesome unit tests)

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
magicus: I'll resolve this thread since we are discussion the same problem elsewhere.

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: I handled this, and also merged StyledText fromString with the StyledTextPart#fromString. I know the location is semi-ideal for the method, but this is the most sane way, with the least code duplication.

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: I will continue to add unit tests, but we can never have enough..

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: Using ArrayList everywhere now.

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
kristofbolyai: I think I did this now. I used `replacePart` instead of implementing custom methods. Please check if you are happy.

common/src/main/java/com/wynntils/core/text/PartStyle.java
kristofbolyai: In that sense, I think append is much cleaner here than formats += ChatFormatting.RED.

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
```diff
@@ -35,53 +37,71 @@ public final class StyledTextPart {
     // This factory is used to create a StyledTextPart from a component that has formatting codes
     // It is separate from the constructor because this only needs to be applied in cases there the text could have
     // formatting codes
-    static StyledTextPart fromComponentWithPossibleFormattingCodes(
-            String text, Style style, StyledText parent, Style parentStyle) {
+    static List<StyledTextPart> fromComponentWithPossibleFormattingCodes(
```
magicus: I think this can be promoted to just `fromCodedString`.

common/src/main/java/com/wynntils/core/text/PartStyle.java
magicus: Yes, it is cleaner in the methods themselves. but tryConstructDifference should return a string, so we get to do 
  `styleString.append(styleDifference);` where styleDifference is a String, not a StringBuilder.

magicus: StringBuilder is this kind of "dirty" implementation detail that should in general not be shared between methods.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,460 @@
...
+        List<StyledTextPart> parts = new ArrayList<>();
+        List<ClickEvent> clickEvents = new ArrayList<>();
+        List<HoverEvent> hoverEvents = new ArrayList<>();
+
+        final int length = texts.length;
+        for (int i = 0; i < length; i++) {
+            StyledText text = texts[i];
+            parts.addAll(text.parts);
+
+            if (i != length - 1) {
```
magicus, line 153: I don't understand this part. Is it still correct? Parts cannot have formatting now. 
edit: Ah, it's the separator you are talking about.

kristofbolyai: It is, the comment is missleading. I removed the comment.

common/src/main/java/com/wynntils/core/text/PartStyle.java
kristofbolyai: Did it.

common/src/main/java/com/wynntils/core/text/StyledText.java
magicus: It's maybe better to let it actually take a StyledText as a separator. And possibly even easier to implement. And then we can a join(String separator, StyledTexts...) wrapper that just calls join(StyledText.fromCodedString(separator), ...)

kristofbolyai: The implementation would not really change I think. But I can add support for StyledText.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,460 @@
...
+
+    public static StyledText join(String separator, StyledText... texts) {
+        List<StyledTextPart> parts = new ArrayList<>();
+        List<ClickEvent> clickEvents = new ArrayList<>();
+        List<HoverEvent> hoverEvents = new ArrayList<>();
+
+        final int length = texts.length;
+        for (int i = 0; i < length; i++) {
+            StyledText text = texts[i];
+            parts.addAll(text.parts);
```
magicus, line 151: Note that the StyledTexts are not allowed to be empty here. Maybe you should have a private method for adding parts that checks if the part is empty, and if so, skips adding it. We need to be careful when manipulating the parts list so we uphold this invariant.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -0,0 +1,460 @@
...
+
+        return new StyledText(parts, clickEvents, hoverEvents);
+    }
+
+    public static StyledText join(String separator, Iterable<StyledText> texts) {
+        return join(separator, Iterables.toArray(texts, StyledText.class));
+    }
+
+    public static StyledText concat(StyledText... texts) {
+        List<StyledTextPart> parts = new ArrayList<>();
```
magicus, line 180: It can be worth noting that a concat is just a join with an empty separator string. I don't know if it would be worth implementing it in terms of a join, but maybe. That will help us make join correct when the separator is empty, as well...

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: I will make it so the constructor skips empty parts. That way we can't possibly construct a StyledText with an empty part.

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: It could be changed, but I don't really see the point. If you want to make sure it works, unit tests can be used. (Although I think we should add tests after/before your PR, to test everything changed there)

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: Added the impl.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -61,69 +61,44 @@ private Pattern buildPattern() {
...
-        MutableComponent curr = MutableComponent.create(comp.getContents()).withStyle(comp.getStyle());
-        // .getString() is used here as it gives formattingchars when those exist. It is needed for guild messages
-        // because wynn still uses legacy coloring for it.
-        String text = curr.getString();
-
-        // if current component has text check it for mentions and rewrite the component if necessary
-        if (!text.isEmpty()) {
-            curr = rewriteMentions(curr, text, comp.getStyle());
-        }
+        StyledTextPart partFinding = styledText.getPartFinding(mentionPattern, PartStyle.StyleType.NONE);
```
magicus: Focusing on the usage of the new StyledText here, I'm not entirely happy. Githubs view of the PR is less than ideal, but I'll try to comment somewhere close to the code...

First of all, from what I can see you are not using partFinding here, but are really just checking if the entire message matches the mention pattern. Why not use the normal match() function for that?

magicus: Or `getMatcher` as you called it. (We can discuss naming separately)

You are calling it in the "should highlight text" part anyway, I see now.

kristofbolyai: The comments just below answer this. This is because we don't really want to go through the edge case of marking a ping, when it is in 2 separate components (there is no logical way that this can happen). So instead of checking the whole string, we try to find the first part.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -61,69 +61,44 @@ private Pattern buildPattern() {
...
+                        "Found a part matching the mention pattern, but couldn't find a match in the string.");
+                return;
+            }
 
-            // Handle any additional mentions within the message
-            while (match.find()) {
-                // get the bit in between of matches
-                curr.append(Component.literal(text.substring(lastEnd, match.start())))
-                        .withStyle(style);
+            StyledText newText = styledText.splitAt(matcher.start()).splitAt(matcher.end());
```
magicus: This logic seems frail, and not general enough. Getting matcher start position seems scary and probably incorrect -- that matching is done using the coded string, so matching "§4b" for b will give position 2, even though the entire string is just 1 character long.  You will also fail to highlight multiple mentions in the same message. Perhaps not strictly needed, but does look like a bug.

kristofbolyai: Matching is done on a not coded string. Highlighting multiple mentions was not something I planned to do; there is no point (unlike chat items where I will add it.)

magicus: Instead, I think you should have a iterating callback interface to StyledText, eg. like this:
```
styledText.process(part -> {
if (matches(part, pattern)) {
List<Part> splitted = part.splitInThreeAndHighlighMatch(part, patttern);
return splitted;
} else {
return null; // no change to part
}})
```

and let the callback to process return either `null` == no change to the part, a new part == replace it with the new part, List<Part> == replace it with this list of parts (don't forget to check for empty parts!), or finally `StyledTextPart.EMPTY` (a new constant with an empty string), which means we want to delete this part.

magicus: (an alternative implementation is to just return the unchanged part if it should not be changed, and return null if it should be deleted)

magicus: That also  mean I think we should remove the `splitAt`, `getPartMatching` and `replacePart` functions on StyledText, since I think they are likely to get used in a way that introduce bugs, i.e. they are not helpful to us for writing correct code.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -61,69 +61,44 @@ private Pattern buildPattern() {
...
-            // finally add any text after the mentions back onto the component tree
-            curr.append(Component.literal(text.substring(lastEnd))).withStyle(style);
+            newText = newText.replacePart(
+                    partMatching, partMatching.withStyle(style -> style.withColor(mentionColor.get())));
+
+            MutableComponent newComponent = newText.getComponent();
+            e.setMessage(newComponent);
         }
 
-        return curr;
```
magicus, line 127: If you move this above the markMention.get() check, you can flatten out the rest of the method by short-cutting out a return if markMention is not true. Less symmetry, perhaps, but also fewer nesting levels. You decide.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
kristofbolyai: But this 3 way split is needed for multiple features. Should I make it an util?

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
kristofbolyai: Some people might want the sound but no mark.

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -0,0 +1,335 @@
...
+        this.obfuscated = obfuscated;
+        this.bold = bold;
+        this.strikethrough = strikethrough;
+        this.underlined = underlined;
+        this.italic = italic;
+        this.clickEvent = clickEvent;
+        this.hoverEvent = hoverEvent;
+    }
+
+    PartStyle(PartStyle partStyle, StyledTextPart owner) {
```
magicus, line 52: Just checking, this is not private since it is used from another class?

kristofbolyai: Yeah, everything that is package private is used somewhere

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
magicus: I think this looks fine as now. We can make a util later on if we need to split parts like this in more places.
(Also, I think there is a regexp "split" functionality in Java that can probably be used to make it a tad more elegant)

magicus (approved):
Ship it!

PR #1437, opened 2023-04-03
feat: Introduce JUnit tests

kristofbolyai:


build.gradle
```diff
@@ -39,7 +39,6 @@ subprojects {
 
     loom {
         silentMojangMappingsLicense()
-        clientOnlyMinecraftJar()
```
kristofbolyai, line 42: Had to remove this. Apparently this should not have been used for a long time..

kristofbolyai:
This is ready now.

build.gradle
magicus: Ok. The downside is that now we will get 3 more run configurations that will be not used :(

magicus (approved):

PR #1442, opened 2023-04-04
fix: Add separators between scoreboard segments, log suspicious states for objective scoreboard part

kristofbolyai:


common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -344,6 +349,14 @@ private void createScoreboardFromSegments() {
                 scoreboard.getOrCreatePlayerScore(line, wynntilsObjective).setScore(currentScoreboardLine);
                 currentScoreboardLine--;
             }
+
+            if (i != segments.size() - 1) {
+                scoreboard
```
magicus, line 354: Is the number of À:s increasing for each part? :-O I have never noticed that. Oh, I see, this is to create a unique name.

magicus (approved):

PR #1443, opened 2023-04-04
fix: Don't show duplicate values in WynntilsBookSettingsScreen

kristofbolyai:


magicus (approved):

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
```diff
@@ -349,19 +350,19 @@ private void reloadConfigurableButtons() {
...
+                        || feature.getVisibleConfigOptions().stream().anyMatch(this::configOptionContains))
+                .map(feature -> (Configurable) feature)
+                .sorted()
+                .collect(Collectors.toList());
+
+        configurableList.addAll(Managers.Overlay.getOverlays().stream()
+                .filter(overlay -> !configurableList.contains(Managers.Overlay.getOverlayParent(overlay)))
+                .filter(overlay -> searchMatches(overlay)
+                        || overlay.getVisibleConfigOptions().stream().anyMatch(this::configOptionContains))
+                .sorted()
```
magicus, line 364: Just out of intellectual curiosity, I wonder if this could have been done purely with streams? If you had concatenated the streams and then sorted them?

kristofbolyai: I don't think so, since I needed the features to be collected first, so I can check for not adding duplicates (as adding features will add all of its overlays, but adding an overlay will not add the holder feature..)

kristofbolyai: (Which might not be the smartest decision to do, but we are going through iterations of this :) and it is getting better..)

PR #1448, opened 2023-04-05
fix: Many fixes to Token overlay

kristofbolyai:
With the alignment fix (I would assume this is how it was supposed to look, based on it's default position, top left):

image

### DIFF
--- a/common/src/main/java/com/wynntils/core/features/overlays/ContainerOverlay.java
+++ b/common/src/main/java/com/wynntils/core/features/overlays/ContainerOverlay.java
@@ -6,6 +6,7 @@ package com.wynntils.core.features.overlays;
 
 import com.mojang.blaze3d.platform.Window;
 import com.mojang.blaze3d.vertex.PoseStack;
+import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.config.Config;
 import com.wynntils.core.config.ConfigHolder;
 import com.wynntils.core.config.RegisterConfig;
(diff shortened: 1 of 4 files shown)


kristofbolyai:
This great feature was broken in so many ways with the recent changes to overlays.. Hopefully, this should fix all of them :)

common/src/main/java/com/wynntils/core/features/overlays/ContainerOverlay.java
```diff
@@ -33,24 +34,39 @@
...
+            GrowDirection growDirection,
+            HorizontalAlignment horizontalAlignment,
+            VerticalAlignment verticalAlignment) {
+        this(position, size, growDirection, DEFAULT_SPACING, horizontalAlignment, verticalAlignment);
     }
 
     public void addChild(T overlay) {
         inherentSize.put(overlay, overlay.getSize().copy());
         children.add(overlay);
+        WynntilsMod.registerEventListener(overlay);
```
magicus, line 63: You're making all overlays listeners? I guess that is fine, but why? I assumed that I had manually registered the TokenOverlay as a listener and that you would remove that as part of this fix, but I could not find it. Are there any other overlays that are "manually" registered now that does not need to be? If not, why is this needed? 🤔

kristofbolyai: It was you that made all overlays event listeners afaik. But yes, every overlay that works with functions uses events. And some other overlays too, so the holder feature does not have to hook into the overlay. So this is needed.

kristofbolyai: (Registering all overlays caused the unintended issue of https://github.com/Wynntils/Artemis/pull/1446)

magicus: Ok, I'm apparently pretty confused about this issue. :) Let me rephrase: This seems like you are adding a registration. If you have moved this from somewhere else, there should have been a corresponding delete somewhere in the patch.

Oh. Now I think the penny dropped. This is for Container children; other overlays get registered by the overlay framework. I see.

magicus (approved):

PR #1453, opened 2023-04-05
feat: Parse profession level up messages and xp gains with multipliers

kristofbolyai:


common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -92,18 +103,27 @@ public void resetValueFromItem(ItemStack professionInfoItem) {
...
+
+        professionProgressMap.put(type, new ProfessionProgress(oldValue.level(), newPercentage));
+
         rawXpGainInLastMinute.get(type).put(xpGain);
     }
 
+    public void updateLevel(ProfessionType type, int newLevel) {
+        ProfessionProgress oldValue = professionProgressMap.getOrDefault(type, ProfessionProgress.NO_PROGRESS);
+
+        // We don't know the progress, so we set it to 0.
```
magicus: Is it not better to not reset the progress percentage here? It will be set by updatePercentage, yes? So read the existing percentage, but update the level.

kristofbolyai: No sadly, become we get a level up message first, then a +1000xp [100%]. (So we can only set this to 0 in both cases of level and progress update).

kristofbolyai: But I guess this is not required here yes.

magicus (approved):

PR #1454, opened 2023-04-05
feat: Major ID Range Visualizer

ryanzhoudev:
image

I have decided against having a first person circle because a) it doesn't make sense, since the circle will be decently large, but also, if you are actively trying to keep people within your radius, you will probably be in 3rd person. And b), it is kind of difficult. The posestack from the ItemInHandRenderer has bobbing and is not very consistent.

Colors:
Pink - Ignis (heart of the pack)
White - Hero (saviour's sacrifice)
Red - Guardian (guardian)


common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -0,0 +1,148 @@
...
+
+        // Major IDs that we can visualize:
+        // Taunt (12 blocks)
+        // Saviour's Sacrifice (8 blocks)?
+        // Heart of the Pack (8 blocks)?
+        // Guardian (8 blocks)?
+        // Marked with a ? needs additional confirmation
+
+        for (GearMajorId majorId : majorIds) {
+            if (majorId.name().equals("Taunt")) {
```
magicus: I'd recommend you check the majorId.id() instead, which is like `TAUNT`. This is supposed to be more stable. In theory, the name can change (perhaps if we introduce i18n of gear?)

magicus:
Maybe I'm coming here and questioning your hard-earned rendering code :) but perhaps we can make it look a bit more snazzy than just a compact circle? I was thinking, what if you e.g. were to leave out every other segment, and try to make it look "dashed", like `- - - -`. Would that work? Does it look better? Maybe by offsetting the segments based on system time, we can get the dashes to rotate around you? I think that would make it look more integrated in the Wynncraft feeling, without being too difficult to implement.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private static final MultiBufferSource.BufferSource BUFFER_SOURCE =
+            MultiBufferSource.immediate(new BufferBuilder(256));
+    private static final int SEGMENTS =
+            128; // number of straight lines to draw when rendering circle, higher = smoother but more expensive
+    private static final float HEIGHT = 0.1f;
+
+    @SubscribeEvent
+    public void onPlayerRender(PlayerRenderEvent e) {
+        if (!Models.Player.isLocalPlayer(e.getPlayer())) return; // Don't render for ghost/npc
```
magicus: These are a lot of checks to make on each render. I worry it might be too expensive. Maybe you should do these checks once per tick instead, and just store radius and color for the rendering code to pick up and just render the circle from?

common/src/main/java/com/wynntils/mc/mixin/PlayerRendererMixin.java
```diff
@@ -60,4 +61,25 @@ private void onNameTagRenderPre(
...
+            float entityYaw,
+            float partialTicks,
+            PoseStack matrixStack,
+            MultiBufferSource buffer,
+            int packedLight,
+            CallbackInfo ci) {
+        PlayerRenderEvent event =
+                new PlayerRenderEvent(entity, entityYaw, partialTicks, matrixStack, buffer, packedLight);
+        MixinHelper.post(event);
+        if (event.isCanceled()) {
```
magicus: There is never any point in making a "post" mixin cancellable, since the method has already done everything it should. So you can just remove the `cancellable = true` from the annotation, and this code.

(Also, if you really wanted to make this event cancellable, you'd have needed to make the new event `@Cancellable` as well)

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -0,0 +1,148 @@
...
+
+@ConfigCategory(Category.COMBAT)
+public class RangeVisualizerFeature extends Feature {
+
+    private static final MultiBufferSource.BufferSource BUFFER_SOURCE =
+            MultiBufferSource.immediate(new BufferBuilder(256));
+    private static final int SEGMENTS =
+            128; // number of straight lines to draw when rendering circle, higher = smoother but more expensive
+    private static final float HEIGHT = 0.1f;
+    private final Map<Player, CustomColor> renderData = new HashMap<>();
```
magicus: If you make this a `Map<Player, Pair<CustomColor, Integer>>` you can store both the color and the radius.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -0,0 +1,148 @@
...
+    private final List<Player> players = new ArrayList<>();
```
magicus: I think the purpose of this gets clearer if you rename it `detectedPlayers` or something like that.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @SubscribeEvent
+    public void onPlayerRender(PlayerRenderEvent e) {
+        if (players.contains(e.getPlayer())) return;
+        players.add(e.getPlayer());
+
+        if (!renderData.containsKey(e.getPlayer())) return;
+
+        for (Map.Entry<Player, CustomColor> entry : renderData.entrySet()) {
```
magicus: This will render all circles for all renderData, each time you come here to render one player.

Instead, do something like this:
```
Player player = e.getPlayer();
Pair<CustomColor, Integer> thisRenderData = renderData.get(player);
renderCircle(e.getPoseStack(), thisRenderData.b(), player.position(), thisRenderData.a());
```

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
ryanzhoudev: Radius isn't needed anymore, I decided to drop the TAUNT major id. It's possible for accessories to have this id, so we can't reliably detect it. I think we should just do the one that actually matter to other players first anyway.

ryanzhoudev:
The basic feature should be good now, though I have not looked into doing the rotating dashed circle yet.

magicus (approved):
There are some cleanups needed, but let's get this into mainline now and we can fix that later.

PR #1455, opened 2023-04-06
feat: Add ping function

kristofbolyai:
I am not quite happy with the packages.. This is really an edge-case for us currently. But the implementation seems to work greatly :)

### DIFF
--- a/common/src/main/java/com/wynntils/models/ping/PingModel.java
+++ b/common/src/main/java/com/wynntils/models/ping/PingModel.java
@@ -51,6 +51,7 @@ public class PingModel extends Model {
     }
 
     private void sendPingPacket() {
+        // We use -1 as the id, as it is not used by Minecraft.
         McUtils.mc().getConnection().send(new ServerboundCommandSuggestionPacket(-1, ""));
         lastPingSent = System.currentTimeMillis();
     }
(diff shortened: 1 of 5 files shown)


magicus:
As in this "model" not being a "model" of the Wynncraft world?

kristofbolyai:
> As in this "model" not being a "model" of the Wynncraft world?

Yeah.. But it needs world states model, so it cannot really be a manager. Also, I am not sure if ping is a world function..

common/src/main/java/com/wynntils/models/ping/PingModel.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    // We are specifically looking for the packet itself, not it's processing. This is why we are using the PacketEvent.
+    @SubscribeEvent
+    public void onCommandSuggestions(PacketEvent.PacketReceivedEvent<ClientboundCommandSuggestionsPacket> event) {
+        if (event.getPacket().getId() != -1) return;
+
+        lastPing = (int) (System.currentTimeMillis() - lastPingSent);
+
+        // We also need to cancel the packet, as it would be unexpected to Minecraft.
```
magicus, line 49: Are you sure it is safe to always cancel the event? Is the thinking that we will only get the packetevent if we are connected to a wynncraft server, and if we are, then the "real" command suggestion packet has already been sent by the server to the client, and it is only sent once per connection?

kristofbolyai: No packets with the id of -1 will be sent from the client (it sends from 0...POS_INF).

magicus: Ah, I see. Please add this as a comment either when sending or receiving the packet.

magicus (approved):

PR #1458, opened 2023-04-06
feat: Wrap coded string access inside CodedString

magicus:
This wraps all access to "coded strings", that is, any `String` which is supposed to, or might, contain Minecraft formatting codes, like "§a".


common/src/main/java/com/wynntils/models/discoveries/DiscoveryContainerQueries.java
```diff
@@ -165,12 +168,12 @@ private void processDiscoveryPage(ContainerContent container, int page, int last
...
-    private static String getDiscoveryPageTitle(int pageNum) {
+    private static String getDiscoveryPageTitleRegex(int pageNum) {
         // FIXME: We ignore pageNum, as we do not have a valid way of only querying dynamic amounts of pages
         return "^§0\\[Pg. \\d+\\] §8.*§0 Discoveries$";
     }
 
-    private String getNextPageButtonName(int nextPageNum) {
-        return "[§f§lPage " + nextPageNum + "§a >§2>§a>§2>§a>]";
+    private CodedString getNextPageButtonName(int nextPageNum) {
+        return CodedString.fromString("[§f§lPage " + nextPageNum + "§a >§2>§a>§2>§a>]");
```
kristofbolyai, line 177: It might be worth PRing this separately. The brackets here come from not using itemStack.getHoverName (like you should), but getDisplayName, which appends the brackets. This will break with because of the new formatting (and the fact that you parse it from fromString). 

You do you. If you see the quest book broken, let's hope you don't forget this :).

kristofbolyai: (like you did with 1.19.4 PRs.. "In preparation for ...")

kristofbolyai (approved):

common/src/main/java/com/wynntils/models/discoveries/DiscoveryContainerQueries.java
magicus: Sorry, I don't understand really what you are saying here. The quest book works fine with this PR. Is this a suggestion to improve how quest book does container queries?

kristofbolyai: Yes. I might as well just PR this tomorrow. It will save you trouble later on.

kristofbolyai: (it is a 2 liner change)

magicus: Please do.

PR #1460, opened 2023-04-07
fix: Chat mentions should not mark own name in header

kristofbolyai:


kristofbolyai:
Fixes bugs like
image

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -63,7 +65,13 @@ public void onChat(ChatMessageReceivedEvent e) {
 
         StyledText styledText = StyledText.fromComponent(message);
 
-        StyledText modified = styledText.iterate((part, changes) -> {
+        StyledText modified = styledText.iterateBackwards((part, changes) -> {
```
magicus, line 68: Ok, but now you are only marking the *last* mention of your name. I thought just marking the first, instead of all, was borderline buggy. But this is just weird. Please mark all instances, basically start at the end and iterate backwards and mark all instances until you reach the header.

kristofbolyai: Done.

kristofbolyai:
Bit of accident with the force pushes..

magicus:
force

magicus (approved):

magicus (approved):

PR #1461, opened 2023-04-07
refactor: Port ScriptedContainerQuery CodedString usages

kristofbolyai:
This ports every CodedString usage in container queries (except where we use `LoreUtils#getLore`).

This PR brings up a question. Should we use equals between StyledText? Obviously, comparing the output strings would be less faster, but more safe. I think equals should work if we parse the string correctly, which we hopefully do.


magicus:
I think we must trust in our StyledText enough to be able to use equals.

magicus (approved):

PR #1462, opened 2023-04-07
refactor: In preparation for using StyledText for item annotators, remove trimming from WynnUtils#normalizeBadString

kristofbolyai:
I checked all usages, I could not find any location it really matters...

PR #1463, opened 2023-04-07
refactor: Port ItemAnnotator (and the fallout) to use StyledText

kristofbolyai:
Depends on #1462


common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -148,12 +149,8 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
...
     }
 
-    private CodedString getBaseName(CodedString name) {
-        int bracketIndex = name.getInternalCodedStringRepresentation().lastIndexOf('[');
-        return bracketIndex == -1
-                ? name
-                : CodedString.fromString(
-                        name.getInternalCodedStringRepresentation().substring(0, bracketIndex));
+    private StyledText getBaseName(StyledText name) {
+        return name.getPartCount() == 1 ? name : name.iterate((part, changes) -> IterationDecision.SKIP);
```
kristofbolyai: This is most likely the most intrusive change in this PR. I tested this with consumables, and it seems to work. I think this captures the intent on the old method, in general.

magicus: Hm. The IterationDecision SKIP means that just the first part will be kept, and the rest ignored? So it's a way of saying "give me the first part of the styled text"? We don't have a simple like `styledText.getPart(index)`? Maybe we should? And, also, the iterate would work just as well even if partcount is 1, yes? Or is that a correctness thing rather than an optimization thing?

As for the overall functionality, I guess it would make sense to break at the first format change, but I'm not 100% sure. The current solution that was looking for `[` was not ideal either. We'll probably just have to try it and see if people start reporting breakage on item overlays.

magicus:
We need to check performance on this. I don't think it is an integration blocker; if there are regressions we can fix it up later, but we would need to go to the TM, and scroll past a decent amount of pages (I usually clicked next page 50-100 times), while gathering performance data. In general, Item annotation is not a performance problem, but in that particular case it can be, so we need to make sure the StyledText operations does not add too much performance cost.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
kristofbolyai: I think we removed `getPart(index)`, it also feels a bit weird. Iterate works on any length, yes. And yes, we will see the reports, but from testing this should work.

kristofbolyai:
> We need to check performance on this. I don't think it is an integration blocker; if there are regressions we can fix it up later, but we would need to go to the TM, and scroll past a decent amount of pages (I usually clicked next page 50-100 times), while gathering performance data. In general, Item annotation is not a performance problem, but in that particular case it can be, so we need to make sure the StyledText operations does not add too much performance cost.

Do you plan to do that? Or should I look into it?

magicus (commented):
Have you looked at the individual item annotator regexes and deemed them to be okay, within reasonable judgement? And/or did you check any of them ingame, to see that they still work?

kristofbolyai:
> Have you looked at the individual item annotator regexes and deemed them to be okay, within reasonable judgement? And/or did you check any of them ingame, to see that they still work?

Did look through all regexes, also tested most types in game.

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
```diff
@@ -40,6 +41,6 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, CodedString name) {
         // e.g. a GUI element
         if (!untradable && !questItem) return null;
 
-        return new MiscItem(name, untradable, questItem);
+        return new MiscItem(CodedString.fromStyledText(name), untradable, questItem);
```
magicus, line 44: Was it too much fallout to change MiscItem as well?

common/src/main/java/com/wynntils/models/items/annotators/game/UnknownGearAnnotator.java
```diff
@@ -18,15 +19,15 @@ public final class UnknownGearAnnotator implements ItemAnnotator {
...
+    public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
         GearType gearType = GearType.fromItemStack(itemStack);
         if (gearType == null) return null;
 
         Matcher matcher = name.getMatcher(UNKNOWN_GEAR_PATTERN);
         if (!matcher.matches()) return null;
 
         String gearName = matcher.group(1);
-        GearTier gearTier = GearTier.fromFormattedString(name);
+        GearTier gearTier = GearTier.fromFormattedString(CodedString.fromStyledText(name));
```
magicus: This appears multiple times. It might be that basically all calls to this function is from annotators. You might want to check if it is possibly to convert it to use StyledText instead. (Even if you only put the odedString.fromStyledText(name)) call inside fromFormattedString, it's an improvement.

common/src/main/java/com/wynntils/utils/type/IterationDecision.java
```diff
@@ -6,5 +6,6 @@
 
 public enum IterationDecision {
     CONTINUE,
-    BREAK
+    BREAK,
```
magicus: I'm a bit confused now. You add this here, but don't you have to use it in the styled text iterator method implementation as well?

kristofbolyai: I should, yes.. I dont know how it worked..

magicus: Also, the comment is not so clear. "next elements". I assume you mean "all remeaining elements", giving how this is used in the item handler. But the comments make it sound almost like you just skip the next part, and that the plural -s is just a typo.

common/src/main/java/com/wynntils/models/items/annotators/game/UnknownGearAnnotator.java
magicus: If so, also please rename it, to `fromStyledText` or similar.

magicus:
> Do you plan to do that? Or should I look into it?

Not planning, so feel free to check. Otherwise I guess I can do that but it will have to be later, I'm still bound up with other stuff.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: Maybe we could have a getFirstPart() then? It feels like a complicated code to just get the first part of a string, and it was (and still is) not really clear to me that this is really what is happening here.

kristofbolyai: Okay. Let's do that.

kristofbolyai:
@magicus

PR after warmup: 
image

Main (missleading because class loading):
image


Main after warmup:

image

kristofbolyai:
Performance seems identical, some annotators are slower, and we cannot test against identical sets...

magicus:
> Performance seems identical, some annotators are slower, and we cannot test against identical sets...

Perfect! All I wanted to know was that we did not make a clear and measurable regression.

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
magicus: I'm not sure what you mean by resolving this? Let me phrase it like this:
I think we should change MiscItem so the first argument to the constructor (and the field and getters) are StyledText, instead of CodedString.

And I would like to see that change in this PR, if it is possible.

common/src/main/java/com/wynntils/models/gear/type/GearTier.java
```diff
@@ -39,18 +40,26 @@ public static GearTier fromString(String typeStr) {
         }
     }
 
-    public static GearTier fromFormattedString(CodedString name) {
-        if (name.getInternalCodedStringRepresentation().charAt(0) == '§') {
-            return fromChatFormatting(ChatFormatting.getByCode(
-                    name.getInternalCodedStringRepresentation().charAt(1)));
+    public static GearTier fromStyledText(StyledText text) {
```
magicus, line 43: Hm. This seems overly complicated, and possibly incorrect as well. Basically, what you need to check is if the first part has a style that has a color. 

Oh, hm, I see. That is kind of what you are doing, but... I think PartStyle needs to have a getChatFormatting. Or, you should skip relying on fromChatFormatting() here, so you don't have to do this conversion. Looking through all colors to determine if an item is colored as "rare" or "unique" seems, well, suboptimal.

kristofbolyai: I am not sure if there is a more reliable way, and this is not that slow as it looks like.

magicus: Ok, I'll accept this now. Possibly I'll get back to it and try to clean it up.

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
magicus: Or, sure, we can fix it later.

magicus (approved):

common/src/main/java/com/wynntils/models/items/annotators/game/MiscAnnotator.java
kristofbolyai: I kind of forgot to do this.. Sorry.

PR #1464, opened 2023-04-07
feat: ChatMessageReceivedEvent now sends StyledText too

kristofbolyai:


common/src/main/java/com/wynntils/features/redirects/TerritoryMessageRedirectFeature.java
```diff
@@ -54,8 +54,6 @@ public void onSubtitleSetText(SubtitleSetTextEvent event) {
     // text event.
     @SubscribeEvent
     public void onChat(ChatMessageReceivedEvent event) {
-        if (event.getOriginalCodedMessage()
-                .getMatcher(TERRITORY_MESSAGE_PATTERN)
-                .matches()) event.setCanceled(true);
+        if (event.getOriginalCodedString().getMatcher(TERRITORY_MESSAGE_PATTERN).matches()) event.setCanceled(true);
```
magicus: I realize this is not new, but since you touch this anyway. Please add braces to the if statement.

magicus (approved):
I look forward to the day when ChatMessageReceivedEvent *only* contains a StyledText...

common/src/main/java/com/wynntils/features/redirects/TerritoryMessageRedirectFeature.java
magicus: Or, whatever, I can fix this separately. Better get this in since other PRs depend on it.

PR #1465, opened 2023-04-07
refactor: Port ChatItemFeature to use StyledText

kristofbolyai:
Depends on #1464


common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -72,83 +70,54 @@ public void onKeyTyped(KeyInputEvent e) {
...
+            StyledTextPart partToReplace = part;
+            Matcher matcher = Models.Gear.gearChatEncodingMatcher(partToReplace.getUnformattedString());
 
-        e.setMessage(decodeMessage(message));
-    }
+            while (matcher.find()) {
+                GearItem decodedItem = Models.Gear.fromEncodedString(matcher.group());
+                if (decodedItem == null) continue;
+
+                String unformattedString = partToReplace.getUnformattedString();
```
magicus: The name confused me here for a while. Does really `getUnformattedString` make any sense for a StyledTextPart? I mean, by definition it will be a string without any formatting codes. Do we have a `getString` and a `getUnformattedString`? If so, what does the latter do that the former does not? And if they are equal, we can remove one. And if this is the only way to retrieve the string from a STP, I think it should be called just `getString`.

kristofbolyai: getUnformattedString is just getString(StyleType.NONE)

kristofbolyai: I can remove it, if needed.

magicus: Ok, now I understand, the getString method returns the string with formatting codes. Hm. 

I think it would be better if "getUnformattedString" were just "getString" -- it just retrieves the string literal from a Part, and the current "getString" would be like "getCodedString", or "getFormattedString", or something like that.

We can leave that to a separate PR.

common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -72,83 +70,55 @@ public void onKeyTyped(KeyInputEvent e) {
...
+            Matcher matcher =
+                    Models.Gear.gearChatEncodingMatcher(partToReplace.getString(null, PartStyle.StyleType.NONE));
 
-        e.setMessage(decodeMessage(message));
-    }
+            while (matcher.find()) {
+                GearItem decodedItem = Models.Gear.fromEncodedString(matcher.group());
+                if (decodedItem == null) continue;
+
+                String unformattedString = partToReplace.getString(null, PartStyle.StyleType.NONE);
```
magicus, line 84: Can you update the variable to match?

magicus (approved):

PR #1467, opened 2023-04-07
fix: Fix color enums showing up as an empty string, disallow non color chatformatting enums

kristofbolyai:
This PR indirectly fixes ChatFormatting enum showing up as "" buttons (toString returns the formatting code..)


magicus (approved):

magicus:
Oops. Has this always been broken but never noticed? Or did something break this?

kristofbolyai:
> Oops. Has this always been broken but never noticed? Or did something break this?

Whenever you introduced EnumUtils, that broke it, most likely

PR #1471, opened 2023-04-07
feat: Add config to hide all/player nametags, hide nametag backgrounds

kristofbolyai:


magicus (approved):
I assume the non-player nametags is like NPCs etc, that gets their tags using an armor stand?

kristofbolyai:
> I assume the non-player nametags is like NPCs etc, that gets their tags using an armor stand?

It is.

PR #1475, opened 2023-04-10
feat: Add AbilityTreeDataDumpFeature

kristofbolyai:
I wonder what this is going to be used for...


common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -397,6 +398,11 @@ private StyledTextPart getPartBefore(StyledTextPart part) {
         return parts.get(index - 1);
     }
 
+    @Override
+    public Iterator<StyledTextPart> iterator() {
```
kristofbolyai, line 402: I did not end up using this, but I think it is still reasonable to have this.

kristofbolyai:
Example: https://gist.github.com/kristofbolyai/c3459b073e5e98869fb83dd53304b441

common/src/main/java/com/wynntils/models/abilities/AbilityTreeModel.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.abilities;
+
+import com.wynntils.core.components.Model;
+import java.util.List;
+
+public class AbilityTreeModel extends Model {
```
kristofbolyai, line 10: I have decided to move all parsing to `ParsedAbilityTree`. This will make it so this model is not cluttered, and mostly focuses on runtime manipulation of ability trees, while parsing will reside in the parser class.

kristofbolyai:
Just as a note. The PR adding the GUI also modifies this, making it work much more widely (you dont have to manually set to first page, and can parse not fully reset trees as well). I stopped backporting every change since it took too long. This works fine in itself though, so it can be merged.

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeInfo.java
```diff
@@ -0,0 +1,237 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.abilities.type;
```
magicus, line 5: I think this class does too much to reside in `type`. The intention is that is like enums and records, simple data-holding classes.

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeInfo.java
```diff
@@ -0,0 +1,237 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+
+/**
+ * This class contains all relevant info to a specific class' ability tree.
+ */
+public class AbilityTreeInfo {
+    private static final File SAVE_FOLDER = WynntilsMod.getModStorageDir("debug");
```
magicus, line 33: This class seem somewhat general, but then has parts that are highly specialized for dumping to a file. 

Is your intention to reuse this class later on to e.g. read the current players actual ability tree? If so, you might want to separate perhaps the stuff done for dumping to a json file, and the stuff needed when reading up the players actual tree.

magicus (approved):

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeInfo.java
kristofbolyai: This class is what I use to represent a class' ability tree. Ideally only 5 instances of these ever exist, one per class. I have a `ParsedAbilityTree` which is a record of map of Ability Nodes - States. I plan to use that for storing "current" (modified, parsed, loaded, shared, etc) states. 

In fact, to answer your other comment. I realized that as well, and moved the relevant node parsing code to the node record itself, making the file much smaller, and the code fit in more. However, that did not end up being backported.

kristofbolyai: @magicus Fine to merge this as is, and it is obviously going to be addressed in the next part?

magicus: Yes

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeInfo.java
magicus: You can have a shot at this in the follow up PR instead.

PR #1476, opened 2023-04-10
fix: Use Wynntils' WynnData fork in the mod

kristofbolyai:


kristofbolyai:
https://github.com/Wynntils/WynntilsWebsite-API/pull/108

magicus:
You don't need to bump the version number to update an URL, just if you want to add a new url instance.

kristofbolyai:
Okay, got it. But it might be nicer overall..

magicus:
I dunno, really. If we increase it every time we change an URL, it will start approaching very high numbers. I think it is better to try and keep the version number changes to actually mean breaking changes to the format.

PR #1477, opened 2023-04-10
Add Default value in TradeMarketBulkSellFeature

ShadowShift:
6399 is the maximum number of items you can put into a single market slot, so i think it makes sense to have it as one of the default options.


ryanzhoudev:
We specifically removed this actually

https://github.com/Wynntils/Artemis/pull/1360#discussion_r1144528543

ShadowShift:
alright. thank you anyways.

PR #1478, opened 2023-04-11
feat: Start working on custom ability tree

kristofbolyai:
Depends on #1475


common/src/main/java/com/wynntils/features/ui/CustomAbilityTreeFeature.java
```diff
@@ -0,0 +1,44 @@
...
+        Optional<AbilityTreeItem> abilityTreeItem = Models.Item.asWynnItem(event.getItemStack(), AbilityTreeItem.class);
+
+        if (abilityTreeItem.isEmpty()) return;
+
+        event.setCanceled(true);
+        McUtils.player().closeContainer();
+
+        // Wait for the container to close
+        Managers.TickScheduler.scheduleNextTick(
+                () -> Models.AbilityTree.ABILITY_TREE_CONTAINER_QUERIES.queryAbilityTree(
```
magicus: Feels like this entire mouthful should be a stored in a method in the AbilityTreeModel.

kristofbolyai: Most likely it will be. Along with the currently non existant ability tree modifying queries.

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeSkillNode.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final Pattern NODE_POINT_COST_PATTERN = Pattern.compile("§.. §7Ability Points: §f(\\d+)");
+    private static final Pattern NODE_BLOCKS_ABILITY_PATTERN = Pattern.compile("§c- §7(.+)");
+    private static final Pattern NODE_REQUIRED_ABILITY_PATTERN = Pattern.compile("§.. §7Required Ability: §f(.+)");
+    private static final Pattern NODE_REQUIRED_ARCHETYPE_PATTERN =
+            Pattern.compile("§.. §7Min (.+) Archetype: §c(\\d+)§7/(\\d+)");
+    private static final Pattern NODE_ARCHETYPE_PATTERN = Pattern.compile("§.§l(.+) Archetype");
+    private static final Pattern NODE_BLOCKED = Pattern.compile("§cBlocked by another ability");
+    private static final Pattern NODE_UNLOCKED = Pattern.compile("§eYou already unlocked this ability");
+
+    public static Pair<AbilityTreeSkillNode, AbilityTreeNodeState> parseNodeFromItem(
```
magicus: When the classes start getting this much complex logic I do not think they belong in `type`. Either move the parsing to another class, like AbilityModel, or a new AbilityParser, or whatever, or move the class to just abilities, not abilities.type.

common/src/main/java/com/wynntils/models/abilities/type/AbilityTreeSkillNode.java
```diff
@@ -0,0 +1,174 @@
...
+    public static boolean isNodeItem(ItemStack itemStack, int slot) {
+        StyledText nameStyledText = StyledText.fromComponent(itemStack.getHoverName());
+        return itemStack.getItem() == Items.STONE_AXE
+                && slot < 54
+                && nameStyledText.getMatcher(NODE_NAME_PATTERN).matches();
+    }
+
+    public record ArchetypeRequirement(String name, int required) {}
+
+    public record ItemInformation(int itemId, int damage) {
```
magicus: It could be worth considering moving these classes to their own files, at least now when the ItemInformation got more complex.

common/src/main/java/com/wynntils/utils/render/RenderUtils.java
```diff
@@ -956,7 +956,7 @@ public static void rotatePose(PoseStack poseStack, float centerX, float centerZ,
     }
 
     // Basically this is ItemRenderer#renderGuiItem, but we can modify the poseStack
-    private static void renderGuiItem(ItemStack itemStack, int x, int y, float scale) {
+    private static void renderGuiItem(PoseStack poseStack, ItemStack itemStack, int x, int y, float scale) {
```
magicus: I don't get this. What does this do?

kristofbolyai: We can now pass a pose stack. In many screens, we use pose stack translations to set a drawing origin. The old impl did not allow us to use an existing pose to draw an item, so we had to figure out x, y, which is quite messy. Now, we can support pose translations (but not scaling)

magicus: Yeah, but you are not using it..?

magicus: Also, this is completely rewritten in the 1.19.4 port. Something to consider.

kristofbolyai: Line 974. If it is rewritten, I can adapt the code accordingly.

magicus: Ah, I missed that.

kristofbolyai:
I'll merge this whole PR into a single commit to a separate branch and do the update process there.

PR #1484, opened 2023-04-13
feat: Add CustomPoiProviders

kristofbolyai:
This is something we wanted for a while now. No, it is most likely not the best version we can make. But this is a way to start, get opinions, etc. 

This version allows arrays of custom POIs to be fetched, displayed, toggled, removed. Adding a quest book GUI is pending for this.


kristofbolyai:
I would like to merge this with a curated list of the current scavenger hunt, to display the feature :). So, do not merge.

PR #1486, opened 2023-04-15
chore: Use non-lerping position method instead

magicus:


kristofbolyai (approved):
It's not a getter... I could not find this method. Either way, this is the same thing.

magicus:
I understand, it is tricky to find if you don't know what it is called. It used to be just a public field, and when Mojang encapsulated it, they did not follow standard java bean naming patterns...

PR #1496, opened 2023-04-15
feat: Add sprint function

magicus:


magicus:
The logic was there, it was just never exposed.

kristofbolyai (approved):

PR #1498, opened 2023-04-15
feat: Add HadesPartyOverlayFeature

kristofbolyai:


kristofbolyai:
image
image
imageimage

magicus:
The priority is for sorting party members in the overlay..? Otherwise I'm not sure I get this.

magicus:
I was surprised that this did not use the party scoreboard. I think that maybe we should try to get what data we can even for non-hades players, i.e. try to parsing the health number.

common/src/main/java/com/wynntils/models/players/hades/objects/HadesUser.java
```diff
@@ -30,6 +30,25 @@ public HadesUser(HSPacketUpdateMutual packet) {
         this.updateFromPacket(packet);
     }
 
+    // Dummy constructor for previews
+    public HadesUser(String name, CappedValue health, CappedValue mana) {
+        this.uuid = UUID.randomUUID();
```
magicus: This feels wrong. Can't we feed it the dummy values from outside?

Also, UUID.randomUUID() for preview seems scary. You could get anything, including NSFW skins. Better off hard-coding Alex/Steve.

magicus:
In fact, I'm not sure I understand the relationship between this feature and the built-in party scoreboard...

magicus (approved):

PR #1503, opened 2023-04-15
feat: Improve range visualizer

magicus:
Unfortunately, I have not been able to test this since I do not own any items with these major IDs on them. :(


common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -80,25 +91,22 @@ private CustomColor getCircleColor(Player player) {
...
-
-        return switch (gearInfo.fixedStats().majorIds().get(0).name()) {
-            case "HERO" -> CommonColors.WHITE;
-            case "ALTRUISM" -> CommonColors.PINK;
-            case "GUARDIAN" -> CommonColors.RED;
-            default -> null;
-        };
+        return gearInfo.fixedStats().majorIds().stream()
+                .map(majorId -> switch (majorId.name()) {
+                    case "TAUNT" -> Pair.of(CommonColors.ORANGE, 12);
```
ryanzhoudev: I don't think we should have taunt, there is actually no item at all that we can hold that has the major ID taunt

(determined by going to WynnAtlas on WynnBuilder, filtering all weapons and rarities >= rare, and ctrl+f TAUNT)

magicus: Oh, I see, taunt is only on armor. 

But should we not then check armor as well..?

ryanzhoudev: Taunt is also on an accessory, which is impossible to check, so that's why I decided to remove it, then it would at least be consistent

ryanzhoudev:
Oh yeah, as a side note, for the HERO id, since the effect only applies when the wielder's HP is >= 50%, we can integrate with Hades and only render the circle when that criteria is met. But I think that should wait for bolyai's party overlay first.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
magicus: (Taunt is available on "Signal Flare", legendary boots, and "Strobelight", a fabled necklace (with Warrior req))

magicus: (Taunt is available on "Signal Flare", legendary boots, and "Strobelight", a fabled necklace (with Warrior req))

ryanzhoudev: Hm oh actually... the new event chestplate has the hero id as well...

magicus: But we can check accessories for ourselves, at least. Just because we can't cover all cases does not mean we need to exclude cases we *can* cover, right?

magicus:
I think we also need to parse the party scoreboard to get health data for all party members, even non-Wynntils players. But let's take that discussion in bolyai's PR instead.

kristofbolyai (approved):

kristofbolyai:
https://athena.wynntils.com/crash/view/4750e5f5097543212aac4438887e19a6

magicus:
Kapture 2023-04-16 at 12 22 42

magicus:
I think this makes it look a bit more snazzy.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
magicus: I added check for armor on all players, and accessories for ourselves. We can unfortunately never get full detection on other players. :(

magicus:
I made some major adjustments since the last round of approvals, so please have another look.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -14,91 +14,132 @@
...
+        if (!Models.Player.isLocalPlayer(player)) return;
+
+        List<GearInfo> validGear;
+
+        if (player == McUtils.player()) {
+            // This is ourselves, rendered from outside
+
+            // Don't render for preview in inventory or character selection screen
+            if (McUtils.mc().screen instanceof InventoryScreen) return;
+            if (McUtils.mc().screen instanceof CharacterSelectorScreen) return;
```
kristofbolyai, line 91: Check for the wynntils book screen as well.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -14,91 +14,132 @@
...
+                    // We cannot verify class or level :(
+                    validGear.add(mainHandGearInfo);
+                }
+            }
+
+            // Check armor slots
+            player.getArmorSlots().forEach(itemStack -> {
+                GearInfo armorGearInfo = getOtherPlayerGearInfo(player.getMainHandItem());
+                if (armorGearInfo != null) {
+                    if (armorGearInfo.type().isArmour()) {
```
kristofbolyai: `isArmour`? I thought we used `isArmor`, the American english version. But since this is not added in this PR, no need to change.

common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
```diff
@@ -20,6 +24,13 @@ public final class InventoryUtils {
     public static final int SOUL_POINTS_SLOT_NUM = 8;
     public static final int INGREDIENT_POUCH_SLOT_NUM = 13;
 
+    private static final int RING_1_SLOT_NUM = 9;
+    private static final int RING_2_SLOT_NUM = 10;
+    private static final int BRACELET_SLOT_NUM = 11;
+    private static final int NECKLACE_SLOT_NUM = 12;
+    private static final int[] ACCESSORY_SLOTS = {RING_1_SLOT_NUM, RING_2_SLOT_NUM, BRACELET_SLOT_NUM, NECKLACE_SLOT_NUM
+    };
```
kristofbolyai: Unfortunate formatting

kristofbolyai (approved):
Looks good after you add the screen to the render blacklist.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
magicus: Good point, I will.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
magicus: Yes, I reacted to that, too. We need to fix it in a separate PR.

common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
magicus: Yes, I agree, but it is upheld by Spotless. :-( I tried fixing it but Spotless restored it to this look again. I need to make some code changes to force Spotless to change the formatting -- is it worth it?

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
magicus: Ah, what the heck. I fixed it in this.

PR #1504, opened 2023-04-15
fix: Bulk sell buttons incorrectly appear in blacksmith

ryanzhoudev:


kristofbolyai (commented):
I am not sure if this is the way to go, the check is rather lose.

magicus:
I agree, this looks imprecise. Why is the `SELL_DIALOGUE_TITLE` match not enough? Do the TM and Blacksmith sell dialog really have the exact same title?

ryanzhoudev:
image
image

They are the exact same

magicus:
I see. I'd rather see a solution where we check for the hover text of the "+" (set amount) button. (After all, we are counting on that existing if we are to be able to press it).

common/src/main/java/com/wynntils/features/overlays/TradeMarketBulkSellFeature.java
```diff
@@ -48,33 +48,28 @@ public class TradeMarketBulkSellFeature extends Feature {
...
+        buttonsAdded = false;
     }
 
     @SubscribeEvent
     public void onSellDialogueUpdated(ContainerSetSlotEvent e) {
         if (!(McUtils.mc().screen instanceof ContainerScreen cs)) return;
+        if (!ComponentUtils.getUnformatted(
+                        cs.getMenu().getSlot(AMOUNT_ITEM_SLOT).getItem().getHoverName())
+                .equals("Click to Set Amount")) return;
         if (!ComponentUtils.getUnformatted(cs.getTitle()).equals(SELL_DIALOGUE_TITLE)) return;
```
magicus: Can you please check the title first? I'm worried that you might end up in a completely different container screen which might not even have a slot 11, and that this check would crash.

magicus (approved):

PR #1505, opened 2023-04-16
fix: Custom loading screen looked bad without the resource pack loaded

magicus:


kristofbolyai (approved):

kristofbolyai:
There is a second issue. You ever seen "endless connecting" to a MC server before in the regular server selector (which seem to happen if you select a server too fast after opening screen)? The same seems to happen for the custom screen as well, and it gets stuck. Refreshing in the server selector seems to fix it. Third issue. I think this screen does not handle errors when connecting, and gets stuck.

magicus:
Can you open github issues, and please elaborate a bit, on how to repro etc.

kristofbolyai:
> Can you open github issues, and please elaborate a bit, on how to repro etc.

I don't have a specific way to reproduce this atm, but when I have, I will open an issue.

PR #1506, opened 2023-04-16
feat: Add a discreet marker to the nametag if the user is running Wynntils

magicus:


common/src/main/java/com/wynntils/features/players/CustomNametagRendererFeature.java
```diff
@@ -135,11 +144,38 @@ private static MutableComponent getItemComponent(ItemStack itemStack) {
 
     private void addAccountTypeNametag(PlayerNametagRenderEvent event, List<CustomNametag> nametags) {
         WynntilsUser user = Models.Player.getUser(event.getEntity().getUUID());
-        if (user == null) return;
+        if (user == null) {
+            if (!nametags.isEmpty()) {
+                // We will cancel vanilla rendering, so we must add back the normal vanilla base nametag
```
kristofbolyai, line 149: `onPlayerNameTagRender` checks for this?

```java
// need to handle the rendering ourselves
        if (!nametags.isEmpty()) {
            event.setCanceled(true);
            drawNametags(event, nametags);
        }
 ```

kristofbolyai (approved):
I think that check is not needed.

common/src/main/java/com/wynntils/features/players/CustomNametagRendererFeature.java
magicus: Ok, the problem here (that I found no way to resolve nicely) is that I want to -- in some cases -- modify the "base" (vanilla) name tag, not just add additional name tags.

But our logic says: "if the nametags list is empty, use vanilla rendering, otherwise cancel vanilla rendering and draw the base nametag ourselves".

So if we have put other nametags in the list, we must take care of rendering the base nametag ourselves, since the event will be cancelled. But if we do not have a WynntilsUser, we will not draw the base nametag with the Wynntils marker later on.

I agree that this is a bit wonky logic, but I could not figure out a way to make it better. The one thing should be to always cancel vanilla drawing and always draw the base nametag ourselves. Maybe that's an okay price to pay to get this cleaner?

kristofbolyai: Okay, I now see this. It works "fine", for now.

kristofbolyai (approved):

PR #1512, opened 2023-04-16
fix: Fix compatibility with modmenu

kristofbolyai:


kristofbolyai:
If you have a better fix @magicus please do change this.

PR #1514, opened 2023-04-17
fix: Stupid workaround for improbable NPE in translation

magicus:
There is no way this argument could be null, yet it happens. I have no idea how this can happen, but let at least avoid the crash.


magicus:
This was already fixed. I am stupid.

PR #1519, opened 2023-04-17
fix: Arrows in chat screen don't focus chat tab buttons

kristofbolyai:


kristofbolyai:
Force merging because this is important.

PR #1522, opened 2023-04-19
fix: Do not post events if player is null

magicus:
I am not clear at how this could happen, but if it does, we should not post any events

### DIFF
--- a/common/src/main/java/com/wynntils/core/events/MixinHelper.java
+++ b/common/src/main/java/com/wynntils/core/events/MixinHelper.java
@@ -6,6 +6,7 @@ package com.wynntils.core.events;
 
 import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Managers;
+import com.wynntils.utils.mc.McUtils;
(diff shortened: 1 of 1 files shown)


kristofbolyai:
Are we sure this is the right step? I am afraid we are going to be masking issues. What if we try to send a player instance in the events that they are needed in? This might be messy too...

Secondly, have you checked that there is no such event that needs to be always posted, regarding this change?

magicus:
The idea is that evens should only be posted while connected to the Wynncraft server. (With the exception of the few "alwaysPost":ed events).

I think this also means we have implicitly assumed that all parts of the vanilla Minecraft class will be properly setup for playing on a server, including the fact that McUtils.mc().player() should always return a valid Player object.

I still seriously do not understand how we can happen in a situation where this is *not* the case, but I'm guessing it is some kind of edge case related to connecting or disconnecting. That is the reason for this patch: I would have assumed that our "onServer" detection would only be true for as long as we have a local Player object, but apparently it is not, so I am also adding that as explicit check.

kristofbolyai (approved):

PR #1528, opened 2023-04-29
fix: initial opening of container incorrectly counts emeralds

ryanzhoudev:
Seems to be fine with the bugged case mentioned in the thread on Discord

### DIFF
--- a/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
+++ b/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
@@ -201,12 +201,12 @@ public final class EmeraldModel extends Model {
         }
 
         for (int i = 0; i < containerStop; i++) {
-            ItemStack is = items.get(i);
-            adjustBalance(is, 1, false);
-            // We need to make a map of the items in the container with emerald values for when the duplicate SetSlot
-            // event is received
-            if (getAmountForItemStack(is) > 0) {
-                handledEmeraldStacks.put(i, is);
+            ItemStack itemStack = items.get(i);
+            adjustBalance(itemStack, 1, false);
+            // We need to make a map of the items in the container with emerald values
+            // When the duplicate SetSlot event is received, this is read and processed properly
+            if (getAmountForItemStack(itemStack) > 0) {
+                handledEmeraldStacks.put(i, itemStack);
             }
         }
         for (int i = containerStop; i < items.size(); i++) {
--- a/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
+++ b/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
@@ -156,13 +156,13 @@ public final class EmeraldModel extends Model {
         if (pouchContainerId != -1 && !isInventory) return;
 
         // Subtract the outgoing object from our balance
-        ItemStack outgoingItemStack = event.getContainer().getItem(event.getSlot());
+        ItemStack oldItemStack = event.getContainer().getItem(event.getSlot());
         if (handledEmeraldStacks.containsKey(event.getSlot())) {
-            // We've already handled this stack, replace the outgoingItemStack with the one we've already seen
+            // We've already handled this stack, replace the oldItemStack with the one we've already seen
             // to force the balance to subtract correctly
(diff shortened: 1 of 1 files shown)


kristofbolyai:
Thanks for fixing this. Before approving though, I want to debug this myself so I understand what happened in 1.19.4

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -191,7 +201,13 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
         }
 
         for (int i = 0; i < containerStop; i++) {
-            adjustBalance(items.get(i), 1, false);
+            ItemStack is = items.get(i);
```
kristofbolyai:
```suggestion
            ItemStack itemStack = items.get(i);
```

We like to use itemStack name, instead of is, item, stack, etc so the code is overall more readable in Artemis.

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -191,7 +201,13 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
...
+            adjustBalance(is, 1, false);
+            // We need to make a map of the items in the container with emerald values for when the duplicate SetSlot
+            // event is received
```
kristofbolyai: Can you manually split this long comment into 2 lines that are not this weird to read?

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -153,7 +156,13 @@ public void onSetSlot(SetSlotEvent.Pre event) {
...
 
         // Subtract the outgoing object from our balance
-        adjustBalance(event.getContainer().getItem(event.getSlot()), -1, isInventory);
+        ItemStack outgoingItemStack = event.getContainer().getItem(event.getSlot());
+        if (handledEmeraldStacks.containsKey(event.getSlot())) {
+            // We've already handled this stack, replace the outgoingItemStack with the one we've already seen
+            // to force the balance to subtract correctly
+            outgoingItemStack = handledEmeraldStacks.get(event.getSlot());
+        }
+        adjustBalance(outgoingItemStack, -1, isInventory);
```
kristofbolyai: what? We remove it then add it the next line? Shouldn't this be in the if block?

Is `event.getItemStack()` the same as `event.getContainer().getItem(event.getSlot())`? If so, please use the first one.

kristofbolyai: `event.getContainer().getItem(event.getSlot())` is the OLD item.. Okay. Can you rename `outgoingItemStack` because I could not figure this out on reading

ryanzhoudev: Removing and adding it in the next line was the original behaviour. If you put a print statement in the adjustBalance method at the end of the file, there is actually a lot of adding, removing, setting to zero, adding again, etc. If we want to fix that, I think that would be better in a separate PR.

And no, the two event methods are not the same. .getItemStack() returns the new item being inputted into that slot, where the getContainer() way returns the old item being replaced.

kristofbolyai: > `event.getContainer().getItem(event.getSlot())` is the OLD item.. Okay. Can you rename `outgoingItemStack` because I could not figure this out on reading

Okay, I see we did not change the original code in this way.. But still, please rename this,  then we are good to go.

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -180,6 +189,7 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
 
         inventoryEmeralds = 0;
         containerEmeralds = 0;
+        handledEmeraldStacks = new HashMap<>();
```
kristofbolyai, line 193: I think you should do this on `onWorldChange` too.

kristofbolyai (approved):

PR #1529, opened 2023-04-29
feat: social function improvements

ryanzhoudev:
Fixes inaccurate function names, where "online_friends" and "online_party_members" returned counts with offline members included. Also adds party_leader.


common/src/main/java/com/wynntils/functions/SocialFunctions.java
```diff
@@ -7,19 +7,38 @@
...
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.OptionalArgumentBuilder(
+                    List.of(new FunctionArguments.Argument<>("includeOffline", Boolean.class, true)));
+        }
+    }
+
+    public static class PartyLeaderFunction extends Function<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            return Models.Party.getPartyLeader();
```
kristofbolyai: What does this return when not in a party/we don't know the leader (didnt parse it yet)?

kristofbolyai (approved):

common/src/main/java/com/wynntils/functions/SocialFunctions.java
ryanzhoudev: It appears to become "??" since the returned value is null and the FunctionManager deals with that for us

kristofbolyai: Can you check if we return null in other places? I feel like that "??" is more of error handling. I think maybe return an empty string (since we return 0 or -1 in cases where we can't figure out a number)

kristofbolyai (approved):

PR #1530, opened 2023-04-29
feat: Message filter refactor, add party finder filter

ryanzhoudev:
Does a bunch of refactors to clean up duplicated code. Also fixes one of the regexes that were changed.

Fixes the welcome message filter not working when the NPC Dialogue feature is disabled. Apparently, events were not posted to the filter when that is disabled, and now there is an independent check for the welcome message.

Also adds the party finder filter.


kristofbolyai:
The welcome to wynncraft messages dont trigger that event since they are multi-line, and are filtered within chat handler (iirc).

ryanzhoudev:
How did this feature previously work then? The regexes were previously all private (still are after this pr) and there seems to be no other occurrence of them

If the only event in that feature wasn't triggered, we just had dead regexes sitting here?

kristofbolyai:
> How did this feature previously work then? The regexes were previously all private (still are after this pr) and there seems to be no other occurrence of them
> 
> 
> 
> If the only event in that feature wasn't triggered, we just had dead regexes sitting here?

That case we might send out the event when we detect multiline, but not quest dialogue.

We might have broken this with CodedString/StyledText? I am not sure.

ryanzhoudev:
I will try to figure this out after lunch then

common/src/main/java/com/wynntils/features/chat/MessageFilterFeature.java
```diff
@@ -0,0 +1,120 @@
...
+    // List of Pair<Foreground, Background>
+    // Ensures we only try relevant regexes for any given message
+
+    private static final List<Pair<Pattern, Pattern>> WELCOME = List.of(
+            Pair.of(Pattern.compile("^§7Loading Resource Pack\\.\\.\\.$"), null),
+            Pair.of(Pattern.compile("^§6Thank you for using the WynnPack\\. Enjoy the game!$"), null),
+            Pair.of(
+                    Pattern.compile(
+                            "^§cSelect a character! Each character is saved individually across all servers, you can come back at any time with /class and select another character!$"),
+                    null),
```
kristofbolyai, line 31: This can be useful and you disable it for everyone by default. Maybe we should make this feature disabled by default?

kristofbolyai: Also this is not a "welcome" message but something other... So I am not sure what to do with this

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -111,7 +113,7 @@ public void onChatReceived(ChatPacketReceivedEvent e) {
                 && codedMessage.getInternalCodedStringRepresentation().indexOf('\n')
                         != (codedMessage.getInternalCodedStringRepresentation().length() - 1)) {
             // This is a "chat screen"
-            if (shouldSeparateNPC()) {
+            if (shouldSeparateNPC() || isWelcomeMessage(codedMessage)) {
```
kristofbolyai: I am afraid of touching chat handler this way, I don't think this was intended to be used this way. Do we really need to do this now? I would rather not filter these messages. Also, these messages are Wynncraft's self-promo, and most likely we should keep them

ryanzhoudev: I am pretty sure this bug existed for as long as these features also existed. There is no other way for the MessageFilter Feature to get the event that it actually filters, and as long as you had NPC Dialogue disabled, that event would never be posted for multiline messages.

The welcome message is a multiline message, so it would never have been filtered with NPC Dialogue disabled.

Also, I only added filters for the non-self-promo stuff, like the "Welcome to Wynncraft!" and the website link/ip. The event promos like the plushie stuff still exists.

common/src/main/java/com/wynntils/features/chat/MessageFilterFeature.java
ryanzhoudev: I will disable by default but I feel like it is a welcome message, since it always comes with the multiline welcome after you select a class. You can't actually see this message with the custom character selector until *after* you already selected a character,

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
magicus: I agree with bolyai's hesitation. The idea with the "Handler" abstraction, is to be a kind of "dispatcher", that has knowledge about the format of Wynncraft multiplexed data, and separates it out to send it further down. It should not dabble with concrete content.

I am not sure I fully understand the problem (I have just skimmed through this), but if we really want to do something like this, then we should have a more general approach, where the filter feature should know about the welcome messages, and plug in the ChatHandler in a suitable way to have them processed.

common/src/main/java/com/wynntils/features/chat/InfoMessageFilterFeature.java
```diff
@@ -1,119 +0,0 @@
-/*
```
magicus, line 1: I assume you have renamed the file, rather than actually deleting and creating a new? It makes it hard to see the diff. Was there a particular reason for the rename, or was it just that you felt that it filters messages that could not be described as "info"?

Maybe you can revert the rename for now, so the diff is easier to examine, and then redo it, either as the last step of this PR after it is approved, or in a separate PR?

ryanzhoudev: Yeah I renamed it, since the filter also does welcome and level up messages. I will revert for now.

ryanzhoudev:
Do not merge until InfoMessageFilterFeature is back to MessageFilterFeature, the i18n is not changed for it since it is just there for ease of review

common/src/main/java/com/wynntils/features/chat/InfoMessageFilterFeature.java
```diff
@@ -12,108 +12,109 @@
...
-            if (hideLevelUp.get()) {
-                if (msg.getMatcher(BACKGROUND_LEVEL_UP_1).find()
-                        || msg.getMatcher(BACKGROUND_LEVEL_UP_2).find()) {
-                    e.setCanceled(true);
-                    return;
-                }
-            }
+        if (hidePartyFinder.get()) {
+            // §5Party Finder:§r§d Hey v8j, over here! Join the §r§bThe Nameless Anomaly§r§d queue and match up with
+            // §r§e3 other players§r§d!
```
kristofbolyai: Remove this or put a regex test link

common/src/main/java/com/wynntils/features/chat/InfoMessageFilterFeature.java
```diff
@@ -12,108 +12,109 @@
...
+            if (msg.getMatcher(pattern).find()) {
+                return true;
             }
         }
+        return false; // Failed to match any patterns
+    }
+
+    /**
+     * Returns the pattern by the given messageType. If that pattern does not exist (null), returns the pattern for the
+     * other messageType instead.
```
kristofbolyai: Why would this be the behaviour? Just return null and handle it.

common/src/main/java/com/wynntils/features/chat/InfoMessageFilterFeature.java
```diff
@@ -12,108 +12,100 @@
...
-                if (msg.getMatcher(BACKGROUND_LEVEL_UP_1).find()
-                        || msg.getMatcher(BACKGROUND_LEVEL_UP_2).find()) {
-                    e.setCanceled(true);
-                    return;
-                }
-            }
+        if (hidePartyFinder.get()) {
+            // Matches until the username in the party finder message
+            e.setCanceled(msg.startsWith(ChatFormatting.DARK_PURPLE + "Party Finder:" + ChatFormatting.RESET
+                    + ChatFormatting.LIGHT_PURPLE + " Hey "));
```
kristofbolyai: Actually I didnt notice this before but why is this not just a regular regex?

ryanzhoudev: Do we need it to be a regular regex? Imo this is strict enough to only detect the pfinder messages and is faster than a normal regex

kristofbolyai: When you have a feature containing 10+ regexes it is weird for a random one not to be a regex, there is no point in doing it. As for performance, it basically doesn't change a thing, you will never notice this.

kristofbolyai (approved):

PR #1534, opened 2023-05-01
fix: Refactor map rendering to fix map rendering issues with Sodium (black lines)

kristofbolyai:
This should also make map rendering faster in theory..


kristofbolyai:
From profiling performance seems to be the same. Either way, the issue is now fixed.

PR #1538, opened 2023-05-02
fix: Refactor emerald count handling to simplify and improve correctness

kristofbolyai:
This basically works (fixes what broke in 1.19.4).

Overall, this is the old system, but we solely rely on set slot event, which should be enough. This also makes the code cleaner. I also refactored the old item stack handling to be much clearer.

However, sometime recently Wynn's emerald stacking broke, causing tooltips (and thus item stacks) not to update. We relied on this, so our count breaks when we modify item stacks. I am fixing this in this PR too for the sake of completeness (TODO).


kristofbolyai:
This now works, but is using a "hack". I think it might be fine to keep it, either way, I am merging this to fix issues.

kristofbolyai:
This now works, but is using a "hack". I think it might be fine to keep it, either way, I am merging this to fix issues.

PR #1541, opened 2023-05-09
feat: Fully complete ability tree parsing

kristofbolyai:
This PR contains parts of the ability screen PR that only affect parsing/processing, and is fine to go into the mod, without mixing it up with the screen elements.


common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
```diff
@@ -0,0 +1,174 @@
...
+        public void processPage(ContainerContent content, int page) {
+            List<ItemStack> items = content.items();
+
+            for (int slot = 0; slot < items.size(); slot++) {
+                ItemStack itemStack = items.get(slot);
+
+                unprocessedTree.processItem(itemStack, page, slot, true);
+            }
+
+            boolean lastPage = page == Models.AbilityTree.ABILITY_TREE_PAGES;
```
magicus: Using both = and == in the same line is always a bit confusing. In this case, the variable lastPage is strictly not needed, and I think the comparison is better inlined in the if test. If you want to explain that you are looking for the last page, maybe to so in a short comment instead?

common/src/main/java/com/wynntils/models/abilitytree/parser/AbilityTreeParser.java
```diff
@@ -0,0 +1,188 @@
...
+
+            matcher = text.getMatcher(NODE_UNLOCKED);
+            if (matcher.matches()) {
+                state = AbilityTreeNodeState.UNLOCKED;
+                continue;
+            }
+        }
+
+        if (state == AbilityTreeNodeState.UNLOCKABLE || state == AbilityTreeNodeState.UNLOCKED) {
+            // Skip empty line + "click here to unlock" / "unlocked already"
```
magicus, line 118: Are you sure these are always the two last lines?

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
kristofbolyai: Yes I can refactor this back, this boolean used to be used for many stuff before.

common/src/main/java/com/wynntils/models/abilitytree/parser/AbilityTreeParser.java
kristofbolyai: Yes, otherwise we wouldnt parse the unlockabel/unlocked state. I know it seems weird, but it should work, and this is only needed for data dumping.

kristofbolyai:
Is the PR looking good after that review you left, or you just peeked at it and couldn't go in depth?

kristofbolyai:
@magicus Forgot to mention ;)

magicus:
Sorry, I just peeked at it. I'll finish the review. Sorry for the delay.

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final int PREVIOUS_PAGE_SLOT = 57;
+    private static final int NEXT_PAGE_SLOT = 59;
+    private static final int DUMMY_SLOT = 76; // This is the second archetype icon
+    private static final StyledText NEXT_PAGE_ITEM_NAME = StyledText.fromString("§7Next Page");
+
+    public void dumpAbilityTree(Consumer<AbilityTreeInfo> supplier) {
+        queryAbilityTree(new AbilityTreeContainerQueries.AbilityPageDumper(supplier));
+    }
+
+    public void updateParsedAbilityTree() {
```
magicus, line 42: Where is this called from? I can't find it.

magicus (approved):
I think it looks good. I just don't understand how the processed tree is populated. And also please fix the =/== thing. Otherwise it's good to go.

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
kristofbolyai: This is only called when the tree parsing is actually used, in the next PR. This just made it into this PR (accidentally).

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
```diff
@@ -0,0 +1,174 @@
...
+
+            for (int slot = 0; slot < items.size(); slot++) {
+                ItemStack itemStack = items.get(slot);
+
+                unprocessedTree.processItem(itemStack, page, slot, true);
+            }
+
+            boolean lastPage = page == Models.AbilityTree.ABILITY_TREE_PAGES;
+
+            if (lastPage) {
```
kristofbolyai:
```suggestion
            if (page == Models.AbilityTree.ABILITY_TREE_PAGES) {
```

PR #1544, opened 2023-05-13
feat: auto-connect on startup

re-vc:
hi, this is a simple "auto join to wynncraft on startup" feature.
configurable - off by default.
tested and working on forge, fabric and quill - jdk 17.0.7
i considered splitting it off into a separate class but let me know what you think.


common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -38,15 +38,20 @@
 public class WynncraftButtonFeature extends Feature {
     private static final String GAME_SERVER = "play.wynncraft.com";
     private static final String LOBBY_SERVER = "lobby.wynncraft.com";
+    private static boolean initialBoot = false;
```
ryanzhoudev: Does this boolean mark that it is currently the initial boot, or that the initial boot has passed? Change the name to be a bit more clear please

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +71,22 @@ private void addWynncraftButton(TitleScreen titleScreen) {
         titleScreen.addRenderableWidget(wynncraftButton);
     }
 
+    /* connects to the server on "startup" - off by default */
+    private void autoConnect() {
+        if (!initialBoot && autoConnect.get()) {
```
ryanzhoudev: Can this config check be moved to where autoConnect() is called, so it forms a natural flow? Also, we don't need to add the wynncraft button if we are immediately joining the server, so maybe put that in some check as well.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +71,22 @@ private void addWynncraftButton(TitleScreen titleScreen) {
...
+            ServerData wynncraftServer =
+                    new ServerData("Wynncraft", connectToLobby.get() ? LOBBY_SERVER : GAME_SERVER, false);
+            wynncraftServer.setResourcePackStatus(ServerData.ServerPackStatus.ENABLED);
```
ryanzhoudev: Has this been checked in both cases where the mod loads the resource pack, and where the mod does not?

common/src/main/resources/assets/wynntils/lang/zh_cn.json
```diff
@@ -686,6 +686,8 @@
   "feature.wynntils.worldWaypointDistance.textShadow.name": "文本阴影",
   "feature.wynntils.worldWaypointDistance.topBoundingDistance.description": "离屏幕时，航路点指示器最近对屏幕的顶部距离是多少？",
   "feature.wynntils.worldWaypointDistance.topBoundingDistance.name": "顶部提示器距离",
+  "feature.wynntils.wynncraftButton.autoConnect.description": "translate: Should you auto-connect to Wynncraft on startup?",
```
ryanzhoudev: Also, I am not sure we need these. I believe most features are merged with only an entry for en_us.json.

kristofbolyai: Remove these, only keep en-us

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai: Moving the check is enough, no need to complicate it further

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai:
```suggestion
    private boolean initialBoot = false;
```

Feature variables musn't be static

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +71,22 @@ private void addWynncraftButton(TitleScreen titleScreen) {
...
+
+            Minecraft minecraft = Minecraft.getInstance();
+            Screen backScreen = minecraft.screen;
```
kristofbolyai: McUtils.mc().screen

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +71,22 @@ private void addWynncraftButton(TitleScreen titleScreen) {
...
+
+            ConnectScreen.startConnecting(
+                    backScreen, minecraft, ServerAddress.parseString(wynncraftServer.ip), wynncraftServer);
+        }
+        initialBoot = true;
```
kristofbolyai: Just set this before you start connecting so this only happens once?

re-vc:
Thank you for the feedback. I appreciate your input, and I will do my best to incorporate the suggested changes this week.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: if you refer to whether or not it causes any unexpected behavior ive tested it with and without mod loaded resource pack and it seems to be fine.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: its called "initBootPassed" now - and it marks if the TitleScreen was "loaded?" at least once.
feel free to suggest a different name for the variable

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -53,7 +58,12 @@ public void onTitleScreenInit(TitleScreenInitEvent.Post e) {
     public void onTitleScreenInit(ScreenInitEvent e) {
         if (!(e.getScreen() instanceof TitleScreen titleScreen)) return;
 
-        addWynncraftButton(titleScreen);
+        if (!initBootPassed && autoConnect.get()) {
+            autoConnect();
+        } else {
+            addWynncraftButton(titleScreen);
+        }
+        initBootPassed = true;
```
ryanzhoudev: What is the purpose of this? Doesn't the autoConnect() function already set this to be true?

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +76,23 @@ private void addWynncraftButton(TitleScreen titleScreen) {
         titleScreen.addRenderableWidget(wynncraftButton);
     }
 
+    /** connects to the server on "startup".
+     *  requires the WynncraftButtonFeature to be enabled.
```
ryanzhoudev:
```suggestion
     *  Requires the WynncraftButtonFeature to be enabled.
```

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +76,23 @@ private void addWynncraftButton(TitleScreen titleScreen) {
         titleScreen.addRenderableWidget(wynncraftButton);
     }
 
+    /** connects to the server on "startup".
```
ryanzhoudev:
```suggestion
    /** Connects to the server on "startup".
```

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: Having inside of the onTitleScreenInit() is crucial - it ensures that if the user goes back to the Title-Screen by any means - they will not get autoConnected to the server.

technically we can take it out in the autoConnect() method.
kristofbolyai said "Just set this before you start connecting so this only happens once?" 
i think it may or may not ensure to not try to connect multiple times at once or in rapid succession.

ryanzhoudev: I see. Then, we could put it before that if statement and remove the one inside autoConnect()?

re-vc: I think the best thing would be, removing it from autoConnect() and leaving it in the onTitleScreenInit() but after the if statement - if put above. it would never trigger autoConnect().

Another option would be splitting the if statement which consists of the config-check && initBootPassed and move the initBootPassed check to autoConnect()

in either case the variable will be updated to true each time if on titleScreen

ryanzhoudev (approved):
Looks fine to me, though I wouldn't be opposed to inverting initBootPassed to isInitBoot and so on if you wanted to.

re-vc:
I think ill leave it as is for now and thanks for working on this with me!

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +76,21 @@ private void addWynncraftButton(TitleScreen titleScreen) {
         titleScreen.addRenderableWidget(wynncraftButton);
     }
 
+    /** Connects to the server on "startup".
+     *  Requires the WynncraftButtonFeature to be enabled.
+     */
```
kristofbolyai: There is no reason to have this comment. More is not always better, and that is the case here.  All the info you describe here is true but can be collected from the feature's class itself.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -66,6 +76,21 @@ private void addWynncraftButton(TitleScreen titleScreen) {
...
+    private void autoConnect() {
+        ServerData wynncraftServer =
+                new ServerData("Wynncraft", connectToLobby.get() ? LOBBY_SERVER : GAME_SERVER, false);
+        wynncraftServer.setResourcePackStatus(ServerData.ServerPackStatus.ENABLED);
```
kristofbolyai, line 82: Since this is straight up copied from above, can you extract this into a method and get the `ServerData` instance from there?

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: removed.
it was a wip comment anyway but the second part the "Requires the WynncraftButtonFeature to be enabled." 
Ive actually spent easily an hour or two to find out.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -53,19 +58,40 @@ public void onTitleScreenInit(TitleScreenInitEvent.Post e) {
     public void onTitleScreenInit(ScreenInitEvent e) {
         if (!(e.getScreen() instanceof TitleScreen titleScreen)) return;
 
-        addWynncraftButton(titleScreen);
+        if (!initBootPassed && autoConnect.get()) {
+            autoConnect();
+        } else {
+            addWynncraftButton(titleScreen);
```
magicus: Why is this an else? I presume that you will want the button added even if you autoconnect, so it is there after you log out from Wynncraft.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -53,19 +58,40 @@ public void onTitleScreenInit(TitleScreenInitEvent.Post e) {
...
 
         WynncraftButton wynncraftButton = new WynncraftButton(
                 titleScreen, wynncraftServer, titleScreen.width / 2 + 104, titleScreen.height / 4 + 48 + 24);
         titleScreen.addRenderableWidget(wynncraftButton);
     }
 
+    private void autoConnect() {
+        ServerData wynncraftServer = getWynncraftServer();
+
+        ConnectScreen.startConnecting(
```
magicus: This is shared with the `onPress` method in the button class. Please extract it to a common private method.

And after doing that, `autoConnect` becomes only two lines, and is not needed (can be inlined in place).

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: The "addWynncraftButton()" method is only necessary for rendering the button on the initial start of the game. If you want to autoconnect during the initial start, you don't need to include it. The button is added through the "onTitleScreenInit(TitleScreenInitEvent.Post e)" event, which triggers after logging out or returning to the titlescreen from any other menu.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -38,10 +39,14 @@
 public class WynncraftButtonFeature extends Feature {
     private static final String GAME_SERVER = "play.wynncraft.com";
     private static final String LOBBY_SERVER = "lobby.wynncraft.com";
+    private boolean hasInitBootPased = false;
```
kristofbolyai: This variable has a spelling mistake and it is impossible to figure out what it means (without reading usages of the variable).

Please rename this to `firstTitleScreenInit` or something with the same meaning

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -868,6 +868,8 @@
   "feature.wynntils.wybelSound.hideText.description": "Should we hide the chat text if we play a sound?",
   "feature.wynntils.wybelSound.hideText.name": "Hide Chat Text",
   "feature.wynntils.wybelSound.name": "Wybel Sounds",
+  "feature.wynntils.wynncraftButton.autoConnect.description": "Should you auto-connect to Wynncraft on startup?",
```
```suggestion
  "feature.wynntils.wynncraftButton.autoConnect.description": "Should the mod auto-connect to Wynncraft on startup?",
```

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -53,29 +58,44 @@ public void onTitleScreenInit(TitleScreenInitEvent.Post e) {
...
         if (!(e.getScreen() instanceof TitleScreen titleScreen)) return;
 
-        addWynncraftButton(titleScreen);
+        if (!hasInitBootPased && autoConnect.get()) {
+            ServerData wynncraftServer = getWynncraftServer();
+            connectToServer(wynncraftServer);
+        } else {
+            addWynncraftButton(titleScreen);
+        }
+        hasInitBootPased = true;
```
kristofbolyai: Refactor this into:

```
if firstboot and autoconnect:
then set firstboot to false;
         connecttoserver;
         return

addwynncraftbutton
```

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: got it, using firstTitleScreenInit now, its a good name.
also sorry i just missed that spelling error somehow.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
re-vc: i see now i get what DonkeyBlaster meant by inverting the firstTitleScreenInit!

kristofbolyai (approved):

kristofbolyai:
Thanks for working with us on this.

PR #1545, opened 2023-05-16
feat: Seaskipper Departure Board Screen

ShadowCat117:
Tried to make it consistent across GUI scales, the destination buttons shift a little bit but I'm happy enough with it.

The map isn't intractable at all in this, it only updates the coordinates based on the mouse position so you can't scroll or drag to move it. I felt like it was unnecessary for this screen as if people want to do that they would probably rather use the other screen anyway, but if you disagree then should be easy enough to add.

https://github.com/Wynntils/Artemis/assets/8337467/f372df30-b43c-4f09-9c18-dbe493666796


kristofbolyai:
Looks good as a start! I will review the code in the upcoming days, maybe change the travel button texture and try to fix up the scaling issues (not that there is any basically, but I would just make the buttons larger on larger scales, but if you want to do that, go ahead).

Thanks for working on this :)

common/src/main/java/com/wynntils/features/ui/CustomSeaskipperScreenFeature.java
```diff
@@ -21,6 +22,9 @@
 @ConfigCategory(Category.UI)
 public class CustomSeaskipperScreenFeature extends Feature {
 
+    @RegisterConfig("feature.wynntils.customseaskipperscreen.screentype")
```
kristofbolyai: Why the custom key? If you rename the variable to `screenType` the name won't conflict (if that was the problem originally).

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,414 @@
...
+        for (MapTexture map : maps) {
+            float textureX = map.getTextureXPosition(mapCenterX);
+            float textureZ = map.getTextureZPosition(mapCenterZ);
+
+            MapRenderer.renderMapQuad(
+                    map, poseStack, centerX, centerZ, textureX, textureZ, mapWidth, mapHeight, 1f / currentZoom);
+        }
+
+        RenderUtils.disableScissor();
+    }
```
kristofbolyai: Did you change anything here or why was this overridden? (I can't look at source right now but this feels really familiar)

common/src/main/java/com/wynntils/features/ui/CustomSeaskipperScreenFeature.java
ShadowCat117: Probably not needed, I think wherever I copied it from originally had it so thought it need it

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,414 @@
...
+
+            destinationButtons.add(newButton);
+        }
+    }
+
+    private void zoomToDestination(Poi destination) {
+        updateMapCenter(
+                destination.getLocation().getX(), destination.getLocation().getZ());
+
+        float zoomLevel = McUtils.mc().getWindow().getGuiScaledHeight() / 650f;
```
kristofbolyai: What is the magic constant of 650 mean? How did you get it? Please extract it into `private static final` to the top of the class.

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,414 @@
...
+                break;
+            }
+        }
+
+        return super.doMouseClicked(mouseX, mouseY, button);
+    }
+
+    @Override
+    public boolean mouseScrolled(double mouseX, double mouseY, double delta) {
+        return true;
```
kristofbolyai, line 328:
```suggestion
        // Disallow zooming
        return true;
```

common/src/main/java/com/wynntils/screens/maps/widgets/SeaskipperBoatButton.java
```diff
@@ -0,0 +1,63 @@
...
+                Texture.BOAT_BUTTON.width(),
+                Texture.BOAT_BUTTON.height() / 2,
+                Texture.BOAT_BUTTON.width(),
+                Texture.BOAT_BUTTON.height());
+
+        if (isHovered) {
+            RenderUtils.drawTooltipAt(
+                    poseStack,
+                    mouseX,
+                    mouseY - TooltipUtils.getToolTipHeight(TooltipUtils.componentToClientTooltipComponent(TOOLTIP)),
```
kristofbolyai, line 56: Marker for myself to check it out whatever this does / if it can be simplified

common/src/main/java/com/wynntils/screens/maps/widgets/SeaskipperTravelButton.java
```diff
@@ -0,0 +1,70 @@
...
+import com.wynntils.utils.render.FontRenderer;
+import com.wynntils.utils.render.RenderUtils;
+import com.wynntils.utils.render.Texture;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+
+public class SeaskipperTravelButton extends WynntilsButton {
+    private List<Component> tooltip;
+    private SeaskipperDestinationPoi selectedPoi = null;
```
kristofbolyai: This doesn't belong here. Store the selected poi in the screen class, and pass the screen to this widget. Then you can call a method in the screen, like `travelToDestination`.

common/src/main/java/com/wynntils/screens/maps/widgets/SeaskipperTravelButton.java
```diff
@@ -0,0 +1,70 @@
...
+                selectedPoi == null ? Texture.TRAVEL_BUTTON.height() / 2 : 0,
+                Texture.TRAVEL_BUTTON.width(),
+                Texture.TRAVEL_BUTTON.height() / 2,
+                Texture.TRAVEL_BUTTON.width(),
+                Texture.TRAVEL_BUTTON.height());
+
+        if (isHovered && selectedPoi != null) {
+            tooltip = List.of(Component.translatable(
+                            "screens.wynntils.seaskipperMapGui.travelToDestination", selectedPoi.getName())
+                    .withStyle(ChatFormatting.DARK_GRAY));
```
kristofbolyai: You can probably drop the %s, and just use `Click here to travel`, it will be obvious enough what is selected without this tooltip.

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
ShadowCat117: Just checking GuiScaledHeight for different sized windows and gui scales and finding a constant to get them all to a similar zoom level, couldn't think of a better way

kristofbolyai:
This is more of a fast review than an actual in-depth review. I didn't get to try this yet, so I am going to be back when I find time. Overall it looks mostly good, nothing outstandingly bad.

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,402 @@
...
+    public void onClose() {
+        McUtils.player().closeContainer();
+        super.onClose();
+    }
+
+    @Override
+    protected void doInit() {
+        currentTextureScale = (float) this.height / (Texture.DESTINATION_LIST.height() + 49);
+
+        reloadDestionationPois();
```
Incompleteusern: Typo here

common/src/main/java/com/wynntils/features/ui/CustomSeaskipperScreenFeature.java
```diff
@@ -21,6 +22,9 @@
 @ConfigCategory(Category.UI)
 public class CustomSeaskipperScreenFeature extends Feature {
 
+    @RegisterConfig
+    private final Config<SeaskipperScreenType> screenType = new Config<>(SeaskipperScreenType.DEPARTURE_BOARD);
```
kristofbolyai:
```suggestion
    public final Config<SeaskipperScreenType> screenType = new Config<>(SeaskipperScreenType.DEPARTURE_BOARD);
```

Configs are public.

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,402 @@
...
+                destinationListWidth,
+                Texture.DESTINATION_LIST.height() * currentTextureScale,
+                Texture.DESTINATION_LIST.width(),
+                Texture.DESTINATION_LIST.height());
+
+        renderWidgets(poseStack, mouseX, mouseY, partialTick);
+    }
+
+    @Override
+    protected void renderCoordinates(PoseStack poseStack, int mouseX, int mouseY) {
```
kristofbolyai: I would remove coordinate rendering, since having these 3 cases makes it a bit confusing to know what is displayed imo.

ShadowCat117: Could cut it down to just the mouse pos, or entirely remove?

kristofbolyai: I am not sure if that would be useful at all?

common/src/main/java/com/wynntils/screens/maps/SeaskipperDepartureBoardScreen.java
```diff
@@ -0,0 +1,404 @@
...
+                    selectedDestination = destinationButton.getDestination();
+
+                    zoomToDestination(selectedDestination);
+                }
+
+                break;
+            }
+        }
+
+        return super.doMouseClicked(mouseX, mouseY, button);
```
kristofbolyai: Why is this super call needed here? I am pretty sure we don't want to do anything the regular map does, do we?

kristofbolyai (approved):
Looks good overall after resolving my last 2 comments. It looks good IMO, it is a shame @magicus is not here to see it.

ShadowCat117:
I may do some more scaling tweaks in the future, but since it works and looks fine as is no point delaying it till I do that

kristofbolyai:
Nice! Thanks for working on this. I agree the scaling is not 100% correct, maybe we should make text have more space for larger scales, etc, but overall it is good enough.

PR #1549, opened 2023-05-19
feat: Emerald Pouch Filled Arc

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/3991742c-0bb7-4cbe-a9e0-0d2b64c3b64d

Tested with tier 1, 4 and 9 pouches.


kristofbolyai (changes requested):
Instead of reviewing this whole PR bit by bit, I only have a general suggestion that will simplify this greatly. Revert the capacity changes to `EmeraldPouchItem` and figure out the capacity by using the already parsed tier value. `GuideEmeraldPouchItemStack` has the maths to calculate the capacity from tier (it might not work with tier 10 because it was written before that existed but that case let me know and I will fix up the math).

kristofbolyai:
However, the feature itself is looking good, we just need to refine the code a bit :)

ShadowCat117:
@kristofbolyai Not sure if this is exactly what you meant so let me know, also looking at it I think it should work with tier 10 but I don't have one to test with

ShadowCat117:
Actually, getting it to print the capacity when viewing the guide screen gives the right number so should work

kristofbolyai:
> @kristofbolyai Not sure if this is exactly what you meant so let me know, also looking at it I think it should work with tier 10 but I don't have one to test with

It is getting closer. I want you to copy the maths from the guide item stack, and use it in the emerald pouch item constructor to calculate the max value. So basically copy what you do in the guide item stack (and revert the guide item stack changes.)

common/src/main/java/com/wynntils/features/inventory/EmeraldPouchFillArc.java
```diff
@@ -0,0 +1,64 @@
...
+import com.wynntils.utils.colors.CustomColor;
+import com.wynntils.utils.render.RenderUtils;
+import com.wynntils.utils.type.CappedValue;
+import java.util.Optional;
+import net.minecraft.util.Mth;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.INVENTORY)
+public class EmeraldPouchFillArc extends Feature {
```
kristofbolyai: This should be called `EmeraldPouchFillArcFeature`

kristofbolyai (approved):
Looks good after you rename the feature

ShadowCat117:
I've clearly missed something as I keep getting `Fix i18n for EmeraldPouchFillArcFeature
 java.lang.AssertionError: Fix i18n for feature.wynntils.emeraldPouchFillArc.name` but can't see what it is

kristofbolyai:
> I've clearly missed something as I keep getting `Fix i18n for EmeraldPouchFillArcFeature
> 
>  java.lang.AssertionError: Fix i18n for feature.wynntils.emeraldPouchFillArc.name` but can't see what it is

Check if en_us.json has the correct key

ShadowCat117:
It is `"feature.wynntils.emeraldPouchFillArcFeature.name": "Emerald Pouch Fill Arc",`

kristofbolyai:
> It is `"feature.wynntils.emeraldPouchFillArcFeature.name": "Emerald Pouch Fill Arc",`

But you don't need to add Feature there. Only the class name, the key was fine

kristofbolyai:
Thanks for working on this! I am happy you are getting familiar with other parts of the mod other than the screens!

PR #1554, opened 2023-05-21
Improved colours from hash

Kepler-17c:
Instead of using the hash's value for the colours directly, this uses it to seed a PRNG and generate an HSV colour with saturation and "lightness" requirements.


kristofbolyai:
Internally, we agreed to do this server-side, closing this. Thanks regardless.

PR #1556, opened 2023-05-21
feat: Territory owner function

ryanzhoudev:
current_territory_owner(prefixOnly), aliased to territory_owner.
prefixOnly is false by default. When true, outputs guild prefix instead of guild name.


kristofbolyai (approved):

kristofbolyai:
Nice, thanks!

PR #1557, opened 2023-05-21
feat: Chat guild rank replacement

ryanzhoudev:
Looks exactly like legacy from my testing.


common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,62 @@
...
+
+    // Test suite: https://regexr.com/7e5gr
+    private static final Pattern RANK_STARS_PATTERN = Pattern.compile("(§3\\[(?:§b)?)(★{0,5})((?:§3)?.{1,16}])");
+
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onChatMessageReceived(ChatMessageReceivedEvent e) {
+        if (rankType.get() == RankType.STARS) return; // default wynn behavior is stars
+
+        Matcher m = RANK_STARS_PATTERN.matcher(
+                StyledText.fromComponent(e.getMessage()).getString());
```
kristofbolyai: This is not how you are meant to use `StyledText` (`getString` would be private in a better world, Check out `ChatMentionsFeature`). TLDR use `StyledText#getMatcher`.

kristofbolyai: This comment is overall true for the other parts of the replacement, I think you can get away by changing only the correct `StyledTextPart` instead of making a whole new component.

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,98 @@
...
+    public void onChatMessageReceived(ChatMessageReceivedEvent e) {
+        if (rankType.get() == RankType.STARS) return; // default wynn behavior is stars
+
+        StyledText styledText = e.getStyledText();
+
+        Matcher m = styledText.getMatcher(RANK_STARS_PATTERN);
+        if (!m.find()) return;
+
+        if (rankType.get() == RankType.NONE) {
+
```
kristofbolyai, line 57:
```suggestion
```

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,97 @@
...
+        StyledText styledText = e.getStyledText();
+
+        Matcher m = styledText.getMatcher(RANK_STARS_PATTERN);
+        if (!m.find()) return;
+
+        if (rankType.get() == RankType.NONE) {
+            StyledText modified = styledText.iterate((part, changes) -> {
+                if (part.getString(null, PartStyle.StyleType.NONE).contains("★")) {
+                    changes.remove(part);
```
kristofbolyai: return BREAK here otherwise you will also remove stars in the user messages itself

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,97 @@
...
+@ConfigCategory(Category.CHAT)
+public class GuildRankReplacementFeature extends Feature {
+    @RegisterConfig
+    public final Config<RankType> rankType = new Config<>(RankType.STARS);
+
+    // Test suite: https://regexr.com/7e5gr
+    private static final Pattern RANK_STARS_PATTERN = Pattern.compile("(§3\\[(?:§b)?)(★{0,5})((?:§3)?.{1,16}])");
+
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onChatMessageReceived(ChatMessageReceivedEvent e) {
```
kristofbolyai, line 41: Can we refactor this method into something like:

```
getModifiedComponent()
if (oldC == newC) return;

e.setMessage(newC)
```

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,97 @@
...
+                changes.remove(part);
+
+                String rankName = "";
+                switch (stars) {
+                    case 1 -> rankName = "Recruiter ";
+                    case 2 -> rankName = "Captain ";
+                    case 3 -> rankName = "Strategist ";
+                    case 4 -> rankName = "Chief ";
+                    case 5 -> rankName = "Owner ";
+                }
```
kristofbolyai: Maybe this whole thing should be an enum? It doesn't fit here. Maybe in player relations model?

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,97 @@
...
+                changes.add(new StyledTextPart(rankName, part.getPartStyle().getStyle(), null, Style.EMPTY));
+                return IterationDecision.BREAK;
+            }
+
+            if (part.getString(null, PartStyle.StyleType.NONE).contains("[")) {
+                // We've already gone through the rest of the string and did not find any stars, this must be a recruit
+                // Since the [ and the username are the same part, we need to do some surgery:
+                // Find username from the original part, then remove it, and add our new reconstructed one
+                String username = part.getString(null, PartStyle.StyleType.NONE).substring(1);
```
kristofbolyai: Writing code like this is waiting to be broken. Your check doesn't guarantee that the substring will give a meaningful username part. 

I would rather you make a regex for the name part (with colors), and make the username a group. That way even if the regex breaks/format changes we won't error out and worst case we don't add "Recruit".

kristofbolyai: (I generally like to avoid substring, it is hard to gurantee that it will actually work and regexes are much better in many cases (and the cost is trivial))

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -0,0 +1,34 @@
...
+import com.wynntils.core.components.Model;
+import java.util.List;
+
+public class GuildModel extends Model {
+
+    public GuildModel() {
+        super(List.of());
+    }
+
+    public enum GuildRank {
```
kristofbolyai: Put this in `com.wynntils.models.players.type` and then there is no need for a new model

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!m.find()) return;
+
+        MutableComponent modifiedComponent;
+        if (rankType.get() == RankType.NONE) {
+            modifiedComponent = getModifiedRemovedComponent(styledText);
+        } else {
+            modifiedComponent = getModifiedNamedComponent(styledText);
+        }
+
+        if (modifiedComponent.equals(styledText.getComponent())) return; // no changes
```
kristofbolyai: The thing is, comparing components might not be stable enough. Please return the styled text and compare that (where we most likely end up with a correct equals check). Sorry for the miscommunication from my earlier comment.

kristofbolyai (approved):
Thanks for working on this!

kristofbolyai (approved):
Thanks for working on this!

PR #1560, opened 2023-05-25
refactor: Make more use of McUtils and other fixes

Mcrtin:


kristofbolyai (commented):
Hey there!

Thank you for this PR. I understand the reasoning behind these changes, however I have to ask you to only include the changes I list here, otherwise this change becomes a mixed bag of a lot of things that in the end just bring bugs.

I want you to keep:
- `McUtils.mc().containerMenu/intentoryMenu/window/options` simplifications (these do not include `Minecraft.getInstance()` -> `McUtils.mc()` changes, since those are the way they are because of class loading order)
- `McUtils.sendPacket` simplification

Now, I will list the changes I don't want you to do, with the added explanation behind every one of them, so you can hopefully understand my decision.

I want you to drop these changes:
- `@ApiStatus/Nullable/Nonnull` annotations: We don't use these (and explicitly disallow them) since they are only useful if they are used everywhere. Since they are not, it is useless to have only a few hanging around. If you want to explicitly signal to null check a return value, use `Optional`.
-  Stream API ports: While they are great and can simplify code, sometimes using them makes the code less clear/telling and ends up being harder to understand. Secondly, porting regular code to streams API is just asking for adding bugs, since you can't really do method flow control when using streams (you actually introduce a fair amount of bugs in chat tab feature if you check). So, TLDR, it is fine to use Streams, but don't overuse it.
-  `Minecraft.getInstance()` -> `McUtils.mc()` changes: Already talked about this, class loading in premature entry points is not good
- We only use `Collections.unmodifiable*` in places where the method is a model API. This is actually not so refined in the code base, but overall I don't think we need to go over the code base and add it in every place, unless it fixes a bug.

I appreciate the work you put in looking through the code base to find all of these "issues", and thank you for fixing them!

kristofbolyai (commented):
You didn't revert the optional changes I asked for? Please do, they are unnecessary and are just a way to introduce bugs.

common/src/main/java/com/wynntils/models/items/ItemModel.java
```diff
@@ -97,11 +97,7 @@ public ItemModel(
     }
 
     public Optional<WynnItem> getWynnItem(ItemStack itemStack) {
-        Optional<ItemAnnotation> annotationOpt = ItemHandler.getItemStackAnnotation(itemStack);
-        if (annotationOpt.isEmpty()) return Optional.empty();
-        if (!(annotationOpt.get() instanceof WynnItem wynnItem)) return Optional.empty();
-
-        return Optional.of(wynnItem);
+        return asWynnItem(itemStack, WynnItem.class);
```
kristofbolyai: Drop this please

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -79,13 +79,12 @@ public void init() {
...
-                                overlayGroupHolder.getOverlays().contains(overlay)))
-                .toList()) {
-            registerConfigOptions(overlay);
-        }
+        List<OverlayGroupHolder> featureOverlayGroups = Managers.Overlay.getFeatureOverlayGroups(feature);
+        Managers.Overlay.getFeatureOverlays(feature).stream()
+                .filter(overlay -> featureOverlayGroups.stream()
+                        .map(OverlayGroupHolder::getOverlays)
+                        .noneMatch(overlays -> overlays.contains(overlay)))
+                .forEach(this::registerConfigOptions);
```
kristofbolyai: Revert please

common/src/main/java/com/wynntils/models/items/ItemModel.java
kristofbolyai: Marked as completed, but this was not changed back.

common/src/main/java/com/wynntils/core/config/ConfigManager.java
```diff
@@ -132,11 +132,6 @@ public void loadConfigOptions(boolean resetIfNotFound, boolean initOverlayGroups
                 }
             }
 
-            List<ConfigHolder> overlayHolders = holder.getOverlays().stream()
-                    .map(this::getConfigOptions)
-                    .flatMap(List::stream)
-                    .toList();
-
```
kristofbolyai, line 139: Forgot to readd the code?

kristofbolyai: Oh, so this is not used.. Okay..

kristofbolyai (approved):
Looks good now! Thanks

kristofbolyai (approved):
Looks good now! Thanks

PR #1562, opened 2023-05-27
refactor: Migrate ItemAnnotator and EntityLabelChangedEvent to StyledText

Mcrtin:
someone has to check the Patterns

### DIFF
--- a/common/src/main/java/com/wynntils/core/text/PartStyle.java
+++ b/common/src/main/java/com/wynntils/core/text/PartStyle.java
@@ -15,6 +15,9 @@ import net.minecraft.network.chat.HoverEvent;
 import net.minecraft.network.chat.Style;
 
 public final class PartStyle {
+
+    public static final PartStyle NONE =
+            new PartStyle(StyledTextPart.EMPTY, CustomColor.NONE, false, false, false, false, false, null, null);
     private static final String STYLE_PREFIX = "§";
 
     private final StyledTextPart owner;
@@ -213,6 +216,10 @@ public final class PartStyle {
                 owner, newColor, obfuscated, bold, strikethrough, underlined, italic, clickEvent, hoverEvent);
     }
 
+    public PartStyle withColor(CustomColor color) {
+        return new PartStyle(owner, color, obfuscated, bold, strikethrough, underlined, italic, clickEvent, hoverEvent);
+    }
+
     public boolean isBold() {
         return bold;
     }
@@ -231,6 +238,10 @@ public final class PartStyle {
 
     public boolean isItalic() {
         return italic;
+    }
+
+    public CustomColor getColor() {
+        return color;
     }
 
     public PartStyle withBold(boolean bold) {
--- a/common/src/main/java/com/wynntils/core/text/PartStyle.java
+++ b/common/src/main/java/com/wynntils/core/text/PartStyle.java
@@ -18,6 +18,7 @@ public final class PartStyle {
 
     public static final PartStyle NONE =
             new PartStyle(StyledTextPart.EMPTY, CustomColor.NONE, false, false, false, false, false, null, null);
+            
     private static final String STYLE_PREFIX = "§";
 
     private final StyledTextPart owner;
--- a/common/src/main/java/com/wynntils/core/text/PartStyle.java
+++ b/common/src/main/java/com/wynntils/core/text/PartStyle.java
@@ -18,7 +18,7 @@ public final class PartStyle {
 
     public static final PartStyle NONE =
             new PartStyle(StyledTextPart.EMPTY, CustomColor.NONE, false, false, false, false, false, null, null);
-            
+
     private static final String STYLE_PREFIX = "§";
 
     private final StyledTextPart owner;
--- a/common/src/main/java/com/wynntils/core/text/StyledText.java
+++ b/common/src/main/java/com/wynntils/core/text/StyledText.java
@@ -6,9 +6,19 @@ package com.wynntils.core.text;
 
 import com.google.common.collect.Iterables;
 import com.wynntils.utils.mc.ComponentUtils;
+import com.wynntils.utils.mc.McUtils;
 import com.wynntils.utils.type.IterationDecision;
 import com.wynntils.utils.type.Pair;
+import net.minecraft.nbt.Tag;
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.Style;
+import org.apache.commons.lang3.ArrayUtils;
+
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.LinkedList;
@@ -18,12 +28,6 @@ import java.util.function.BiFunction;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import net.minecraft.network.chat.ClickEvent;
-import net.minecraft.network.chat.Component;
-import net.minecraft.network.chat.HoverEvent;
-import net.minecraft.network.chat.MutableComponent;
-import net.minecraft.network.chat.Style;
-import org.apache.commons.lang3.ArrayUtils;
 
 public final class StyledText {
     public static final StyledText EMPTY = new StyledText(List.of(), List.of(), List.of());
@@ -105,6 +109,10 @@ public final class StyledText {
         return new StyledText(parts, temporaryWorkaround, clickEvents, hoverEvents);
     }
 
+    public static StyledText fromComponent(Component component, boolean onlyCodes) {
+        return onlyCodes ? fromString(component.getString()) : fromComponent(component);
+    }
+
     public static StyledText fromString(String codedString) {
         return new StyledText(
                 StyledTextPart.fromCodedString(codedString, Style.EMPTY, null, Style.EMPTY), List.of(), List.of());
@@ -112,6 +120,10 @@ public final class StyledText {
 
     public static StyledText fromCodedString(CodedString codedString) {
         return fromString(codedString.getInternalCodedStringRepresentation());
+    }
+
+    public String getString() {
+        return getString(PartStyle.StyleType.DEFAULT);
     }
 
     // We don't want to expose the actual string to the outside world
@@ -235,7 +247,7 @@ public final class StyledText {
     }
 
     public boolean contains(StyledText styledText) {
-        return contains(styledText.getString(PartStyle.StyleType.DEFAULT), PartStyle.StyleType.DEFAULT);
+        return contains(styledText.getString(), PartStyle.StyleType.DEFAULT);
     }
 
     public boolean contains(String codedString, PartStyle.StyleType styleType) {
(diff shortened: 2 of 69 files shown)


kristofbolyai (commented):
Did you not test it or you couldn't figure this out?

Mcrtin:
i already did all changes and just want to introduce them slowly and I also missed 2 or 3 classes

kristofbolyai:
Then what does "someone has to check patterns" mean?

Mcrtin:
I've marked every pattern where I'm not sure whether I removed all escapes or not

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -105,6 +107,10 @@ public static StyledText fromComponent(Component component) {
         return new StyledText(parts, temporaryWorkaround, clickEvents, hoverEvents);
     }
 
+    public static StyledText fromComponent(Component component, boolean onlyCodes) {
+        return onlyCodes ? fromString(component.getString()) : fromComponent(component);
+    }
+
```
kristofbolyai: What is this and why is this needed? I don't event think this properly works. Either way, we should not need to use this and if there are places where we depend on only using color codes from components and not actual styles, then we need to update those places to be correct.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -114,6 +120,10 @@ public static StyledText fromCodedString(CodedString codedString) {
         return fromString(codedString.getInternalCodedStringRepresentation());
     }
 
+    public String getString() {
+        return getString(PartStyle.StyleType.DEFAULT);
+    }
+
```
kristofbolyai, line 129: We didn't need to add this or any change related to adding this. As the comment states on the original `getString` method, in an ideal world that method would be private..

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +373,30 @@ int addHoverEvent(HoverEvent hoverEvent) {
         return hoverEvents.size();
     }
 
+    public int getWidth() {
+        return McUtils.mc().font.width(getString(PartStyle.StyleType.NONE));
+    }
+
+    public StyledText[] split(String s) {
+        return Arrays.stream(getString().split(s)).map(StyledText::fromString).toArray(StyledText[]::new);
+    }
```
kristofbolyai: This is not even used. Plus, this is a dangerous operation which might not be the way to go about things.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +373,30 @@ int addHoverEvent(HoverEvent hoverEvent) {
...
+
+    public StyledText subtext(int start, int end) {
+        return StyledText.fromString(getString().substring(start, end));
+    }
+
+    public StyledText subtext(int start) {
+        return StyledText.fromString(getString().substring(start));
+    }
```
kristofbolyai: I don't like these kinds of operations in StyledText, just like `split`, this is not the way to do things. Let's remove this and find an alternative way for their usages.

kristofbolyai (commented):
I did not get to review everything, here is my comments so far.

common/src/main/java/com/wynntils/core/text/StyledText.java
Mcrtin: This is used in some places i haven't pushed yet. e.g.
```java
private long calculateMessageReadTime(List<StyledText> msg) {
        int words = StyledText.join(" ", msg).split(" ").length;
        long delay =
                dialogAutoProgressDefaultTime.get() + ((long) words * dialogAutoProgressAdditionalTimePerWord.get());
        return delay;
    }
```
or
```java
StyledText[] colored = description.split("\n");
```

I just realized why this is so dangerous...

common/src/main/java/com/wynntils/core/text/StyledText.java
Mcrtin: this is the replacement of fromComponentIgnoringComponentStylesAndJustUsingFormattingCodes as it has such a long name I thought it must be very important ^^. i just copied the behavior to not introduce new bugs.

common/src/main/java/com/wynntils/core/text/StyledText.java
Mcrtin: so, should i move the logic to StyedText if possible?
what should I do here:
```java
WynntilsMod.info("Message Edited: " + msgContainer.getRenderTask() + " -> " + newMessage.getString());
```
wrap everything in StyledText?

common/src/main/java/com/wynntils/core/text/StyledText.java
Mcrtin: what do we do here then:
```java
private StyledText getBaseName(StyledText name) {
        int bracketIndex = name.getString().lastIndexOf('[');
        return bracketIndex == -1 ? name : name.subtext(0, bracketIndex);
    }
```

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: In these tricky cases just use getString and I will eventually fix them up later in the codebase.

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: This can easily implemented with StyledText. Most likely the `Item Name [97%]` will be 2 parts: `Item Name` and `[97%]`. I just have to think about how to not make it look weird (we would have to use iterateBackwards as of right now).

Again, use getString and I will worry about it later.

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -15,6 +15,9 @@
 import net.minecraft.network.chat.Style;
 
 public final class PartStyle {
+
```
kristofbolyai, line 18:
```suggestion
```

common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -15,6 +15,9 @@
...
+    public static final PartStyle NONE =
+            new PartStyle(StyledTextPart.EMPTY, CustomColor.NONE, false, false, false, false, false, null, null);
```
kristofbolyai, line 20:
```suggestion
    public static final PartStyle NONE =
            new PartStyle(StyledTextPart.EMPTY, CustomColor.NONE, false, false, false, false, false, null, null);
            
```

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +377,51 @@ int addHoverEvent(HoverEvent hoverEvent) {
...
+            return Optional.of(JsonParser.parseString(json).getAsJsonObject());
+        } catch (JsonSyntaxException e) {
+            return Optional.empty();
+        }
+    }
+
+    public PartStyle getStyleAt(int index) {
+        for (StyledTextPart part : parts) if ((index -= part.length()) < 0) return part.getPartStyle();
+        return PartStyle.NONE;
+    }
```
kristofbolyai, line 395: This is logically wrong I think, but I also don't see any case where this should be used. I checked the usages and I think I can find a better solution for this later

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +377,51 @@ int addHoverEvent(HoverEvent hoverEvent) {
...
+
+    public StyledText subtext(
+            int start, int end) { // todo: doesn't work with color codes because the parts don't keep their color
+        return StyledText.fromString(getString().substring(start, end));
+    }
+
+    public StyledText subtext(
+            int start) { // todo: doesn't work with color codes because the parts don't keep their color
+        return StyledText.fromString(getString().substring(start));
+    }
```
kristofbolyai: The todo is part of the reason why this need to be removed...

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +377,51 @@ int addHoverEvent(HoverEvent hoverEvent) {
...
+
+    public StyledText replaceAll(String regex, String replacement) {
+        return StyledText.fromString(getString().replaceAll(regex, replacement));
+    }
+
+    public StyledText replace(String target, String replacement) {
+        return StyledText.fromString(getString().replace(target, replacement));
+    }
```
kristofbolyai: Implement these in a way that you can only replace text parts, not formatting codes. You also need to do this for every part and construct a new styled text that way, because this way you lose some crucial info.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -355,6 +377,51 @@ int addHoverEvent(HoverEvent hoverEvent) {
...
+
+    public Optional<JsonObject> toJsonObject() {
+        String json = StringUtils.substringBeforeLast(getString(), "}") + "}"; // remove extra unnecessary info
+        try {
+            return Optional.of(JsonParser.parseString(json).getAsJsonObject());
+        } catch (JsonSyntaxException e) {
+            return Optional.empty();
+        }
+    }
```
kristofbolyai, line 390: What is even this and why is it needed?

common/src/main/java/com/wynntils/core/text/StyledTextPart.java
```diff
@@ -159,6 +161,14 @@ boolean isBlank() {
         return text.isBlank();
     }
 
+    public int length() {
+        return text.length();
+    }
+
+    public int width() {
+        return McUtils.mc().font.width(text);
+    }
```
kristofbolyai: Remove everything that has to do with width from StyledText.
Instead create a method in `FontRenderer` where you can pass StyledText and get its width.

common/src/main/java/com/wynntils/features/inventory/BulkBuyFeature.java
```diff
@@ -39,8 +37,7 @@ public class BulkBuyFeature extends Feature {
     @RegisterConfig
     public final Config<Integer> bulkBuyAmount = new Config<>(4);
 
-    // Test suite: https://regexr.com/7998g
```
kristofbolyai, line 42: Why did you remove this..?

common/src/main/java/com/wynntils/handlers/item/ItemAnnotator.java
```diff
@@ -4,10 +4,10 @@
...
 package com.wynntils.handlers.item;
 
-import com.wynntils.core.text.CodedString;
+import com.wynntils.core.text.StyledText;
 import net.minecraft.world.item.ItemStack;
 
 @FunctionalInterface
 public interface ItemAnnotator {
-    ItemAnnotation getAnnotation(ItemStack itemStack, CodedString name);
+    ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name);
```
kristofbolyai, line 12: This single change was big enough for a PR. (For future reference).

common/src/main/java/com/wynntils/core/text/StyledText.java
Mcrtin: this is from LoreUtils. but due to this being only used once we can also drop the change

common/src/main/java/com/wynntils/features/inventory/BulkBuyFeature.java
Mcrtin: i changed the pattern

PR #1563, opened 2023-05-29
feat: Add StyledText editing methods

kristofbolyai:


common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -333,6 +345,146 @@ public StyledText prependPart(StyledTextPart part) {
         return new StyledText(newParts, temporaryWorkaround, clickEvents, hoverEvents);
     }
 
+    /**
+     * Splits this {@link StyledText} into multiple {@link StyledText}s at the given index.
+     * <p> Note that {@link PartStyle.StyleType.NONE} is used when splitting.
+     * @param regex the regex to split at
+     * @return the split {@link StyledText}s
+     */
+    public StyledText[] split(String regex) {
```
Mcrtin, line 354: This split method doesn't actually work like the String split method: when splitting a string it omits empty strings if they are at the last index of the result so if you have something like this "foo\n\n§0bar" splitting on "\n" this method would return an array of size 2(or 3 idk) and the string method one of size 4. This would be an issue in QuestInfo.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -333,6 +345,146 @@ public StyledText prependPart(StyledTextPart part) {
...
+    }
+
+    /**
+     * Replaces the first occurrence of the given regex with the given replacement.
+     * <p> Note that {@link PartStyle.StyleType.NONE} is used when replacing.
+     * @param regex the regex to replace
+     * @param replacement the replacement
+     * @return the new {@link StyledText}
+     */
+    public StyledText replace(String regex, String replacement) {
```
Mcrtin: you should clarify that it does NOT match "foo§0bar" with regex foobar

kristofbolyai: It is already mostly clarified in line 427

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -333,6 +345,146 @@ public StyledText prependPart(StyledTextPart part) {
...
+
+            currentIndex += part.length();
+        }
+
+        return new StyledText(includedParts, clickEvents, hoverEvents);
+    }
+
+    /**
+     * Replaces the first occurrence of the given regex with the given replacement.
+     * <p> Note that {@link PartStyle.StyleType.NONE} is used when replacing.
```
kristofbolyai:
```suggestion
     * <p> Note that {@link PartStyle.StyleType.NONE} is used when matching and replacing.
```

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -333,6 +345,146 @@ public StyledText prependPart(StyledTextPart part) {
...
+
+            newParts.add(part);
+        }
+
+        return new StyledText(newParts, clickEvents, hoverEvents);
+    }
+
+    /**
+     * Replaces all occurrences of the given regex with the given replacement.
+     * <p> Note that {@link PartStyle.StyleType.NONE} is used when replacing.
```
kristofbolyai:
```suggestion
     * <p> Note that {@link PartStyle.StyleType.NONE} is used when matching replacing.
```

kristofbolyai:
```suggestion
     * <p> Note that {@link PartStyle.StyleType.NONE} is used when matching and replacing.
```

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -333,6 +345,146 @@ public StyledText prependPart(StyledTextPart part) {
...
+
+            newParts.add(part);
+        }
+
+        return new StyledText(newParts, clickEvents, hoverEvents);
+    }
+
+    /**
+     * Replaces all occurrences of the given regex with the given replacement.
+     * <p> Note that {@link PartStyle.StyleType.NONE} is used when matching replacing.
```
kristofbolyai:
```suggestion
     * <p> Note that {@link PartStyle.StyleType.NONE} is used when matching and replacing.
```

common/src/main/java/com/wynntils/core/text/StyledText.java
kristofbolyai: I think this should fix it.

PR #1566, opened 2023-05-29
port: Port a few models to StyledText

kristofbolyai:


kristofbolyai:
StyledText#split is wrong, and needs fixing.

PR #1574, opened 2023-06-02
feat: Item Guide with Multi-Term Search and Item Guess integration

re-vc:
This PR includes the following changes:

+ **Multi-Term Search in Item Guide:** 
  Users can now filter for multiple items at the same time by separating the individual search-terms with a comma or semicolon. 
  For example, "Discoverer, Obolus, Diamond Dust" will now display all three items.

+ **Open Item Guess in Item Guide:** 
  Users can open the Item Guess of an unidentified item in the Item Guide by hovering over the unidentified item in the inventory and pressing the "Open (in) Guides List" Keybind, which is defaulted to "I".

+ **Assignable Keybind for "Dialogue History":** 
  Users now have the ability to assign a Keybind for "Dialogue History".

- **Removed Keybind for Emerald and Powder Guide:** 
  This has been removed to unclutter the settings, as it will not be frequently used from a user experience perspective.


re-vc:
Noticed that the string matching need a "strict" mode

re-vc:
Not quite happy with my Implementation yet, ill keep this as a draft for the time being, ill keep on working on this.

PR #1576, opened 2023-06-03
refactor: Port ChatMessageReceivedEvent to StyledText

Mcrtin:


common/src/main/java/com/wynntils/core/chat/ChatTabManager.java
```diff
@@ -214,8 +212,7 @@ private boolean matchMessageFromEvent(ChatTab chatTab, ChatMessageReceivedEvent
         }
 
         Optional<Pattern> regex = chatTab.getCustomRegex();
-        return regex.isEmpty()
-                || event.getOriginalCodedString().getMatcher(regex.get()).matches();
+        return regex.isEmpty() || event.getStyledText().matches(regex.get());
```
kristofbolyai:
```suggestion
        return regex.isEmpty() || event.getOriginalStyledText().matches(regex.get());
```

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -197,9 +196,9 @@ private void processNewLines(LinkedList<Component> newLines) {
         LinkedList<Component> newChatLines = new LinkedList<>();
         LinkedList<Component> dialogue = new LinkedList<>();
 
-        CodedString firstLineCoded = ComponentUtils.getCoded(newLines.getFirst());
-        boolean isNpcConfirm = firstLineCoded.getMatcher(NPC_CONFIRM_PATTERN).find();
-        boolean isNpcSelect = firstLineCoded.getMatcher(NPC_SELECT_PATTERN).find();
+        StyledText firstLineCoded = StyledText.fromComponent(newLines.getFirst());
+        boolean isNpcConfirm = firstLineCoded.matches(NPC_CONFIRM_PATTERN);
+        boolean isNpcSelect = firstLineCoded.matches(NPC_SELECT_PATTERN);
```
kristofbolyai: This is not equivalent. The old code uses `find`, and you use `matches` which greatly differs.

common/src/main/java/com/wynntils/handlers/chat/type/RecipientType.java
```diff
@@ -4,30 +4,27 @@
...
-            "^§.\\[(Lv. )?\\d+\\*?/\\d+/..(/[^]]+)?\\]§r.*$",
-            "^(§r§8)?\\[(Lv. )?\\d+\\*?/\\d+/..(/[^]]+)?\\]( \\[(§k\\|)?§r§.[A-Z+]+§r§.(§k\\|§r§7)?\\])?(§r§7)? (§r§8)?.*$",
+            "^§.\\[(Lv. )?\\d+\\*?/\\d+/..(/[^]]+)?\\].*$",
+            "^(§8)?\\[(Lv. )?\\d+\\*?/\\d+/..(/[^]]+)?\\]( \\[(§k\\|)?§r§.[A-Z+]+.(§k\\|§r§7)?\\])?(§7)? (§8)?.*$",
             "Local"),
-    GUILD("^(§r)?§3\\[(§b★{0,5}§3)?.*§3]§. .*$", "^(§r§8)?\\[(§r§7★{0,5}§r§8)?.*]§r§7 .*$", "Guild"),
-    PARTY("^§7\\[§r§e[^➤]*§r§7\\] §r§f.*$", "^(§r§8)?\\[§r§7[^➤]*§r§8\\] §r§7[^§]*$", "Party"),
-    PRIVATE("^§7\\[.* ➤ .*\\] §r§f.*$", "^(§r§8)?\\[.* ➤ .*\\] §r§7.*$", "Private"),
-    SHOUT("^§3.* \\[[A-Z0-9]+\\] shouts: §r§b.*$", "^(§r§8)?.* \\[[A-Z0-9]+\\] shouts: §r§7.*$", "Shout"),
-    PETS("^§2(.*): §r§a(.*)$", "^§8(.*): §r§7(.*)$", "Pets");
```
kristofbolyai: Did you just remove all resets or did you manually check these patterns (there are cases where the resets need to stay, you only need to remove when if they are followed by a **color** (not bold/italic/etc) code)?

kristofbolyai (commented):
Close to being able to approve.

common/src/main/java/com/wynntils/handlers/chat/type/RecipientType.java
Mcrtin: i did check. there are multiple §k and i didn't remove the following §r, but I've just noticed that some of the §k are optional and I didn't consider that

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
Mcrtin: there is a ^ and a $. Doesn't that mean, that it always matches the whole string?

kristofbolyai: I didn't check the regexes for this info. However, please still use find (you can add StyledText methods) so it doesn't look confusing in the git history. If you are really bothered you can come back and change the regex etc.

kristofbolyai (approved):

PR #1578, opened 2023-06-03
refactor: Port LoreUtils to StyledText

Mcrtin:


common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
```diff
@@ -116,18 +76,11 @@ public static Matcher matchLoreLine(ItemStack itemStack, int startLineNum, Patte
...
+     * @return a {@link StyledText} representation of the lore
      */
-    public static CodedString getStringLore(ItemStack itemStack) {
-        StringBuilder toReturn = new StringBuilder();
-        for (CodedString x : getLore(itemStack)) {
-            toReturn.append(x);
-        }
-        return CodedString.fromString(toReturn.toString());
+    public static StyledText getStringLore(ItemStack itemStack) {
+        return StyledText.join("\n", getLore(itemStack));
```
kristofbolyai: This method doesn't do the same thing it did?  I also saw you changed how this method is used. Can you explain? (If there is not a good reason I might want this reverted, any unnecessary change in these kinds of PRs can lead to a big mess since this scale of ports is already bug-prone enough and hard to follow)

Mcrtin: you are right. It made more sense to me but could lead to many unforeseen interactions because it is also used in LoreUtils#getJsonFromIngameLore. I will drop this change.

common/src/main/java/com/wynntils/features/debug/LogItemInfoFeature.java
```diff
@@ -70,7 +69,7 @@ private static String getDescription(ItemStack itemStack) {
 
         return "[Logging Item]\nName: "
                 + itemStack.getHoverName().getString() + "\nLore:\n"
-                + CodedString.join("\n", LoreUtils.getLore(itemStack)) + "\nItem Type: "
```
kristofbolyai, line 73: Revert this as well?

common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
```diff
@@ -116,18 +76,11 @@ public static Matcher matchLoreLine(ItemStack itemStack, int startLineNum, Patte
     }
 
     /**
-     * Get the lore from an item, combined into one string.
-     * Relies on {@link #getLore(ItemStack)}. See the aforementioned
-     * for a list format
-     *
-     * @return a {@link String} containing all item lore
+     * Joins the lines of the lore of the given itemStack into a single one with new line characters.
```
kristofbolyai: Is this comment accurate?

kristofbolyai (commented):
Mostly looks good!

kristofbolyai (approved):
Looks good! Thanks.

PR #1583, opened 2023-06-04
fix: Change events so chat tabs behave less weird

kristofbolyai:
These fixes were made during a bug hunt in chat tabs which turned out to be a non-issue (translation feature was duplicating messages..). However during the hunt I changed up a few methods/events which all seem reasonable, so I think we should keep it.


magicus (approved):
Looks good, I think.

PR #1586, opened 2023-06-05
fix: Fix PartyManagementScreen NPE for party leader

kristofbolyai:


common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -144,8 +144,7 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
         createLeaveButton.render(poseStack, mouseX, mouseY, partialTick);
         kickOfflineButton.active = inParty
                 && !Models.Party.getOfflineMembers().isEmpty()
-                && Models.Party.getPartyLeader()
```
magicus, line 147: Maybe we should fix so getPartyLeader never returns null..?

magicus: I'm kind of leaning towards the idea that models should never return null. If a value could be missing, it should return an Optional.

kristofbolyai: Yeah, I feel that way too, and I've kind of designed most models that way anyways.

PR #1591, opened 2023-06-06
refactor: Port everything to StyledText, remove CodedString, and deprecated ComponentUtils methods

kristofbolyai:


magicus:
Woho! 🎉 Finally! Great work, grinding away at all those CodedString instances.

PR #1596, opened 2023-06-07
fix: Fix SpellSegment pattern so spells appear on the action bar

kristofbolyai:


common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -102,6 +102,8 @@ private void processPosition(ActionBarPosition pos, Map<ActionBarPosition, Style
                 } else {
                     segment.update(m);
                 }
+
+                break;
```
kristofbolyai, line 106: Is this a sane change @magicus? I believe it should be correct since a position can only be 1 type of segment?

kristofbolyai: (This ended up being unrelated to the fix, but it caught my attention..)

magicus: Yes, it looks good. Since we have 3 segments, it is really a micro-optimization :-) but sure, I think it is correct.

PR #1605, opened 2023-06-11
fix: disable xp overlay when player hits max level

Mcrtin:


kristofbolyai:
This at very best should be a config, there are reasons why people would still want to see their gains.

Mcrtin:
can you even see that? I don't think you can fill you xp bar when you are level 106. At least I'm stuck at 0/2.1B
image

kristofbolyai:
Okay, I thought this was something else. This is okay.

kristofbolyai (approved):

PR #1615, opened 2023-06-13
feat: Add LastHarvest* profession functions

kristofbolyai:


common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -50,15 +64,33 @@ public ProfessionModel(CharacterModel characterModel) {
         }
     }
 
+    @SubscribeEvent
+    public void onContainerSetSlot(ContainerSetSlotEvent.Post event) {
+        Optional<MaterialItem> materialItem = Models.Item.asWynnItem(event.getItemStack(), MaterialItem.class);
+
+        if (materialItem.isEmpty()) return;
+
+        if (lastHarvestLabel + MAX_HARVEST_LABEL_AGE >= System.currentTimeMillis()) {
```
magicus, line 73: I'm not sure I understand this. If you pick up harvestable materials less than 1000 seconds since the last harvest label, this will be set as the new last harvest?

kristofbolyai: If you pick up a harvest item in the next 1000 MS of the label appearing then it is a "confirmed" harvest

magicus: Ok, I think I got it now. We don't know for sure the material until it is added to our inventory, so you have basically flagged "expect to see some material" and here it ... eh... materializes :)

If I'd seen this before the push I'd have requested a comment about how this logic works. Now it is maybe too much effort to setup a new PR to do this. Your call.

PR #1618, opened 2023-06-13
feat: Add local server bomb tracking, fix duplicate bombs showing up on /bombbell, rename to /bomb

### LINKED ISSUE #1430: Bomb Bell Ignores All Information But Prefixed Chat Alerts
There are three types of bomb information:
1. Champion Bomb Bell
2. general announcement in chat (when in the same world)
3. boss bar status messages

While ⑴ works flawlessly, neither ⑵ nor ⑶ are used to complement the information. Scenario where this matters:  
Players often quickly switch servers when they get the Bomb Bell notification. While they're between servers, another bomb might be thrown (e.g. combat/loot or prof-xp/speed pairs). This means, that only the first one is registered, as types ⑴ and ⑵ cannot be received between worlds. Additionally Bomb Bell has varying delay, so channel ⑶ could be used to improve timer accuracy.

The message shown when you're in the same world as the thrower could be parsed similarly to the Bomb Bell. It just doesn't have this nice prefix and adds some information about the bomb type.

In the boss bar info time is in remaining minutes, with seconds cut off (19:59 shows as 19 min). This creates the under-estimate `min` and over-estimate `min + 1`. With this a new timer with estimate `min + 0.5` can be created, or an existing timer be adjusted with `newTime = clamp(min,  min+1, curTime)`.

With these additional data channels, the feature can benefit all players, rather than just Champions. At that point it's probably justified to create an overlay for the registered bombs.

kristofbolyai:
Fixes #1430

### DIFF
--- a/common/src/main/java/com/wynntils/commands/BombCommand.java
+++ b/common/src/main/java/com/wynntils/commands/BombCommand.java
@@ -14,6 +14,7 @@ import com.wynntils.models.worlds.type.BombInfo;
 import com.wynntils.models.worlds.type.BombType;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import net.minecraft.ChatFormatting;
@@ -23,13 +24,18 @@ import net.minecraft.commands.SharedSuggestionProvider;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.MutableComponent;
 
-public class BombBellCommand extends Command {
+public class BombCommand extends Command {
     private static final SuggestionProvider<CommandSourceStack> BOMB_TYPE_SUGGESTION_PROVIDER = (context, builder) ->
             SharedSuggestionProvider.suggest(Arrays.stream(BombType.values()).map(Enum::name), builder);
 
     @Override
     public String getCommandName() {
-        return "bombbell";
+        return "bomb";
+    }
+
+    @Override
+    public List<String> getAliases() {
+        return List.of("bombbell");
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


kristofbolyai:
Fixes #1430

common/src/main/java/com/wynntils/commands/BombCommand.java
```diff
@@ -81,14 +87,15 @@ private int listBombs(CommandContext<CommandSourceStack> context) {
...
 
         if (bombBells.isEmpty()) {
             response.append(Component.literal(
-                                    "There are no active bombs at the moment! This might be because you do not have the ")
+                                    "There are no active bombs at the moment! This might be because there are no bombs currently, or you do not have the ")
                             .withStyle(ChatFormatting.RED))
                     .append(Component.literal("CHAMPION").withStyle(ChatFormatting.YELLOW))
-                    .append(Component.literal(" rank on Wynncraft, which is necessary to use this feature.")
+                    .append(Component.literal(
+                                    " rank on Wynncraft, which is necessary to use recieve bomb alerts from other servers.")
```
kristofbolyai:
```suggestion
                                    " rank on Wynncraft, which is necessary to receive bomb alerts from other servers.")
```

common/src/main/java/com/wynntils/models/worlds/BombModel.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @SubscribeEvent
+    public void onWorldStateChange(WorldStateEvent event) {
+        CURRENT_SERVER_BOMBS.clear();
+    }
+
+    public void addBombInfo(BombType bombType, BombInfo bombInfo) {
+        BombInfo old = CURRENT_SERVER_BOMBS.put(bombType, bombInfo);
+
+        // Only add to the set if it's not already there
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/commands/BombCommand.java
```diff
@@ -81,14 +87,15 @@ private int listBombs(CommandContext<CommandSourceStack> context) {
     }
 
     private static MutableComponent getBombListComponent(Set<BombInfo> bombBells) {
-        MutableComponent response = Component.literal("Bomb Bells: ").withStyle(ChatFormatting.GOLD);
+        MutableComponent response = Component.literal("Bombs: ").withStyle(ChatFormatting.GOLD);
 
         if (bombBells.isEmpty()) {
             response.append(Component.literal(
```
magicus, line 93: We should  be able to query the users rank from our Character model... Future improvement, I guess...

magicus:
Nice functionality! I also liked that you simplified `BombBell` to `Bomb`. It was about time. :)

PR #1623, opened 2023-06-17
fix: Text box bottom align highlight broken

ryanzhoudev:
Fixes bottom aligned text having broken highlights:
image

Also does some refactoring to improve cursor and highlight alignments on all alignment settings. Cursors and highlights match positions 1:1 now (you can see when you press shift+arrow to highlight). Also removed arbitrary CURSOR_PADDING and VERTICAL_OFFSET constants that were used for hacky alignments.

Tested on the info box config input, the config book search, the quest list search, and the bank search. (Not every single alignment on every single text input, just most of them).


ryanzhoudev:
Actually this breaks with larger boxes that are top/bottom aligned. I will make a real fix later.

PR #1624, opened 2023-06-17
fix: Dry count resets when getting a mythic tome

Mcrtin:
only prevents the dry chest count and the dry boxes count from resetting when getting a mythic tome, but the mythic found event still fires and the mythic found sound also still plays.

### DIFF
--- a/common/src/main/java/com/wynntils/models/containers/LootChestModel.java
+++ b/common/src/main/java/com/wynntils/models/containers/LootChestModel.java
@@ -54,7 +54,7 @@ public final class LootChestModel extends Model {
     }
 
     @SubscribeEvent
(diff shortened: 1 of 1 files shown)


kristofbolyai (commented):
Merging this PR would break the feature. I would suggest finding ways test what you are doing even in these situations (like testing with unique rarity), since this PR was not tested. 
If you read the event's comment, you will understand how I know this doesn't work.

Mcrtin:
can't we use the Post event instead?

kristofbolyai:
I think you can. Again, testing will reveal this, I can't see why it would not work though.

kristofbolyai (approved):
Seems to work fine.

PR #1626, opened 2023-06-17
fix: Remove redundant sever messages when loading resources pack or switching characters

Mcrtin:


kristofbolyai:
This is already implemented in `MessageFilterFeature` (which is opt-in, not on by default).

PR #1628, opened 2023-06-17
fix: LoadingScreen doesn't show up when joining world

Mcrtin:


kristofbolyai (commented):
You should make the Pre event always post, and check all usages of it and handle null checking if it uses McUtils().player

kristofbolyai:
I am not actually sure if this is the best step to take. What do you think @magicus? Should we make the whole event postAlways, not just `Pre`?

magicus:
I am not sure how/why this is a problem? Can you please specify in what circumstances this would be a problem.

As far as I can tell, the normal post of the `Pre` event will happen, since we cant have `ProgressScreen` or `ReceivingLevelScreen` set unless we are connected to the server.

magicus:
I have never noticed this as a problem. If it happens intermittently, it could be a race with the on server detection, and then this should be fixed instead.

Mcrtin:
The Problem is that the player is null when joining a world and the event only gets fired when the player is not null (and on Wynn)

magicus:
What do you mean by "joining a world" in this context? Can you provide a step by step reproducer?

Mcrtin:
Click the button to join Wynn raft in the main menu then the loading screen will appear then disappear and the default joining world screen shows up and then the character selector screen opens

kristofbolyai:
I am not happy with this fix. Loading screen has many issues with getting stuck too, so perhaps it needs special handling unrelated from world state.

Mcrtin:
maybe it is also better to not fix this right now because there is also no queuing screen and if this gets fixed you wouldn't be able to see you position.
image

magicus:
I have now looked into the matter. You are entirely correct. When I added the check that player should not be null to `post()`, this broke the CustomLoadingScreen, but I did not notice that.

I believe this fix is a safe and proper way to resolve this.

magicus (approved):

Mcrtin:
@magicus what about the problem with the queue? correct me if I'm wrong but I don't think you can see your position or that you are in a queue with this fix.

magicus:
I don't know; I have actually never seen that queue screen. But yes, I guess that you will not see that. If it is important to you, you will likely have to disable this functionality for now.

A better way forward would be to parse the queue information and show it on the custom loading screen. I just need to figure out how to trigger this to happen so I can check how to decode this information.

Mcrtin:
I got in a queue because every wc was full. I was like 8th in the queue. When I got on the world I did /hub and saw that most of the servers were full and some were almost.

magicus:
@Mcrtin I have made a best-effort attempt to solve this in https://github.com/Wynntils/Artemis/pull/1679. Please let me know if you ever encounter a queue situation again if this actually solved the problem.

PR #1631, opened 2023-06-18
fix: Fix welcome message filter

Mcrtin:


kristofbolyai (commented):
Why is the regex wrong? I don't like this fix because it eats messages without any info on what it is. Perhaps @DonkeyBlaster can help with the regex?

Mcrtin:
the welcome message is not a foreground message. You can fix that easily but there are always 4 lines after that: the IP, an empty line, and two random ones with for example the youtube channel or other promotion

kristofbolyai:
> the welcome message is not a foreground message. You can fix that easily but there are always 4 lines after that: the IP, an empty line, and two random ones with for example the youtube channel or other promotion

That is easily subject to change in Wynn, and this implementation does nothing to handle that.

ryanzhoudev:
I thought we agreed that we shouldn't be removing their promotional lines, such as the event specific ones or whatever else they like to post. If it's not a foreground message, why not just copy over the regex? I am certain they were foreground at some point or with some configuration. If performance is a problem, we can maybe look at restricting running these regexes to x number of seconds after a class is selected?

magicus:
@DonkeyBlaster is correct. It is a design choice to keep the promotional content from Wynncraft. Basically, we don't want to hide stuff where Wynncraft might ask for money, even if it is annoying. We do not want to pull the rug out from under their feet.

If anything, we could document in the source code that this is deliberately left out from filtering.

kristofbolyai:
Filtering this is not something we should do, and this implementation is not great.

PR #1632, opened 2023-06-18
fix: Fix map moving when left-clicking

Mcrtin:


kristofbolyai (commented):
I don't get what this fixes. I also don't see the fix itself?

Even if this fixes the "issue" it is because you converted the mouse coordinates to be integer, so very small movements are not detected. So I would not even count it as a proper fix.

Mcrtin:
the problem is that the doRender method is called with interger coords of the current mouse position so when you just click on the map it moves a bit to the rounded coods. because I didn't want to change the params of the doRender method I thought I just adapt to it, but ofcourse I could also do that instead. What do you think is better?

magicus:
I think this looks okay. @kristofbolyai Do you still have objections?

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -259,6 +234,13 @@ public boolean keyReleased(int keyCode, int scanCode, int modifiers) {
         return false;
     }
 
+    @Override
+    public boolean mouseDragged(double mouseX, double mouseY, int button, double dragX, double dragY) {
+        if (button == 0)
+            updateMapCenter((float) (mapCenterX - dragX / currentZoom), (float) (mapCenterZ - dragY / currentZoom));
```
kristofbolyai: Use braces

kristofbolyai (approved):

PR #1633, opened 2023-06-19
fix: Major ID range incorrectly renders in cosmetic preview

ryanzhoudev:


common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -88,6 +89,7 @@ private void checkMajorIdCircles(Player player) {
 
             // Don't render for preview in inventory or character selection screen
             if (McUtils.mc().screen instanceof InventoryScreen) return;
+            if (McUtils.mc().screen instanceof ContainerScreen) return;
             if (McUtils.mc().screen instanceof CharacterSelectorScreen) return;
             if (McUtils.mc().screen instanceof WynntilsMenuScreen) return;
```
kristofbolyai: I wonder if we should render this at all if a screen is opened?

magicus: That's a good point. I think that if you enter the inventory you would not like to hide it from the "background" world view, but I don't know if this code even does that. It is a reasonable assumption that we should not render it at all if a screen is opened.

magicus (approved):

kristofbolyai (approved):

PR #1634, opened 2023-06-19
fix: Cursor/highlight alignment improvements

ryanzhoudev:
Fixes bottom aligned text having broken highlights:
image

Highlights are also consistent with 1px padding on either side of a letter now. Previously, letters would have 0px on the left and 2px on the right. Or maybe the other way around, don't remember.

Also does some refactoring to improve cursor and highlight alignments on all alignment settings. Cursors and highlights match positions 1:1 now (you can see when you press shift+arrow to highlight). Also removed arbitrary CURSOR_PADDING and VERTICAL_OFFSET constants that were used for hacky alignments.

All vertical aligns are tested on TextInputBoxWidget, and some of them are also tested on SearchWidget, but that shouldn't matter since it inherits from the former anyway.


kristofbolyai:
Hey, just wanted to let you know that I saw this PR, however it is something I have to sit down and think through, and I will be leaving that for another day. Sorry for the wait.

magicus (commented):
I'll leave the review of this to bolyai. It looks ok to me but I don't have very detailed knowledge about this code.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -76,10 +75,12 @@ protected void renderText(
                         StyledText.fromString(defaultText ? DEFAULT_TEXT.getString() : firstPortion),
                         this.getX() + textPadding,
                         this.getX() + this.width - textPadding - lastWidth - highlightedWidth,
-                        this.getY() + VERTICAL_OFFSET,
+                        this.getY() + this.getHeight() / 2,
+                        this.getY() + this.getHeight() / 2,
```
kristofbolyai, line 79: Why is this there twice? We render between this too coordinates because of `VerticalAlignment.MIDDLE`. So This could be reverted to just `this.getY() + this.getHeight() / 2` and no other change?

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -90,24 +91,26 @@ protected void renderText(
...
-                        this.getY() + VERTICAL_OFFSET,
-                        this.getY() + VERTICAL_OFFSET,
+                        this.getY() + this.getHeight() / 2,
+                        this.getY() + this.getHeight() / 2,
                         0,
                         CommonColors.BLUE,
                         CommonColors.WHITE,
                         HorizontalAlignment.LEFT,
-                        VerticalAlignment.TOP);
+                        VerticalAlignment.MIDDLE);
```
kristofbolyai, line 100: Same here

kristofbolyai: (I know it was like this before, not sure if this was me...)

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
kristofbolyai: Or you can do `this.getY()` and `this.getY() + this.getHeight()`, but I think that is just unnecessarily longer.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
kristofbolyai, line 1: I don't fully understand the changes to the constructor since they are not used anywhere?

kristofbolyai: Also, why allow different alignments other than the specified one?

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
ryanzhoudev: The method that this used to call defaults to VerticalAlignment.TOP. We could add another "shortcut" method, but  I am in favour of removing some of those. I think it creates unnecessary confusion sometimes, and it only saves like 1-2 lines max.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
ryanzhoudev: This was in preparation for some other features I was looking at. At the very least, I think we should be allowing TOP alignments for multiline textboxes when that materializes.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
kristofbolyai: Yes but isn't `MIDDLE` -> `TOP` with (y1 + y2 / 2), which is equivalent to removing the second argument?

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
kristofbolyai: To be honest it looks kind of unreadable as is, but with these changes, it is just a new type of guessing game. So since this is not used, I think it might be the best to remove this and figure out a better font renderer.

magicus:
To be 100% clear: I have no opinions on this. I'll leave the decision when/if to merge this to bolyai.

ryanzhoudev:
Not a fan of tacking on more things to get rid of after FontRenderer refactor. We should probably just wait.

PR #1636, opened 2023-06-19
refactor: merge onScreenInit into onTitleScreenInit in WynncraftButtonFeature

Commander07:
This also fixes the root cause of the issue where two buttons show up so the jank fix for that is also removed.


kristofbolyai:
I believe I tested this when I made the fix, and this doesn't work every case? 

How did you test this?

There are 2*2 cases:
- First screen init / resize reinit
- Artemis only / with ModMenu

Please confirm that all of these cases work.

Commander07:
My bad forgot to test all cases. Thought `ScreenInitEvent` and `TitleScreenInitEvent` had the same execution flow due to their naming.

kristofbolyai:
> My bad forgot to test all cases. Thought `ScreenInitEvent` and `TitleScreenInitEvent` had the same execution flow due to their naming.

No problem, the idea is almost correct, I had this same idea when  I started looking at the fix :)

PR #1641, opened 2023-06-22
fix: Horse hotkey breaks when no space or too many mobs

ryanzhoudev:
Fixes the hotkey being very persistent in trying to get the horse even when it's not possible. This previously ended up in a weird state where the server internally had your hotbar selected on the horse, but visually was on whatever you were holding before.


common/src/main/java/com/wynntils/features/combat/HorseMountFeature.java
```diff
@@ -105,6 +124,11 @@ private void trySummonAndMountHorse(int horseInventorySlot, int attempts) {
         if (attempts <= 0) {
             postHorseErrorMessage(MountHorseStatus.NO_HORSE);
             return;
+        } else if (cancelMountingHorse != null) {
+            McUtils.sendPacket(new ServerboundSetCarriedItemPacket(prevItem));
+            postHorseErrorMessage(cancelMountingHorse);
+            cancelMountingHorse = null;
+            return;
```
kristofbolyai: The else is not needed.

Secondly, doesn't this duplicate messages? Since you redirect it in `ChatRedirectFeature` and post the error messages here too? I am not sure what the cleanest solution would be, but perhaps leaving it to `ChatRedirectFeature` and not posting errors in this case might be the best (but still parsing the errors of course).

kristofbolyai (commented):
This issue has been bothering me in the last few days while I was playing, thank you very much for looking into it!

common/src/main/java/com/wynntils/features/combat/HorseMountFeature.java
ryanzhoudev: Yeah it does. We can leave it, but if the player has the redirect for horse off, the error messages when the horse can't be found by us and the messages from wynn will be inconsistent. I'll remove it anyway

kristofbolyai: Well, yeah, that is on them. But if they turn off `GameNotificationOverlayFeature` all of it will go to the chat. So I think it is fine.

kristofbolyai (approved):

PR #1643, opened 2023-06-22
feat: Silencer - hotkey-toggleable assignable Master Volume

re-vc:
Adds hotkey-toggleable for Master Volume that allows users to switch between the original volume value and a configurable volume value. This functionality is similar to PUBG's 'Instant Volume Reducer.'


kristofbolyai:
Hey, this kind of feature is out of scope of Artemis, since this has no clear connection to Wynncraft.

magicus:
@kristofbolyai I am not sure that is a valid reason for not accepting functionality into Wynntils. I think the bar should be something like "does this provide valuable functionality to a large enough portion of the people playing on Wynncraft". 

Otherwise, we could as well argue that Gamma Bright or Chat Mentions has no clear connection to Wynncraft.

But I do agree that this feels somewhat more peripheral. @re-vc Can you expand a bit on why you think this fits in Wynntils? Is there some special play situations where you need to adjust the volume? Maybe we can even assist in finding these situations, if it is e.g. when playing raids or whatever.

re-vc:
The core idea behind my proposal is to optimize the player experience during grinding sessions / parties in zones like Scrapyard, Waterfall, and Herb. As it currently stands, these areas are overwhelmingly loud due to numerous spells and the mobs amplified by Totem use. The result is that many players end up reducing or completely disabling their game sound.

Furthermore, **grinding sessions of any kind** often lend themselves to multitasking; players might listen to music or watch shows in the background. Hence, a hotkey-toggleable Master Volume functionality would allow seamless audio transition between the game and other entertainment, enhancing the overall gaming experience.

edit:
Ive also not been able to find a mod that does anything similar - wich is understandable in normal Minecraft you do not need to toggle the sound that often. But on Wynncraft its a different story.

kristofbolyai:
I will reopen this and consider again and review.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -782,6 +782,11 @@
   "feature.wynntils.shamanTotemTracking.secondTotemColor.name": "Shaman Totem 2 Color",
   "feature.wynntils.shamanTotemTracking.thirdTotemColor.description": "What color should the totem 3 highlight and tracking text be?",
   "feature.wynntils.shamanTotemTracking.thirdTotemColor.name": "Shaman Totem 3 Color",
+  "feature.wynntils.silencer.disabled": "Silencer disabled",
+  "feature.wynntils.silencer.enabled": "Silencer enabled",
+  "feature.wynntils.silencer.name": "Silencer",
+  "feature.wynntils.silencer.silencerVolume.description": "How loud should the game be",
```
```suggestion
  "feature.wynntils.silencer.silencerVolume.description": "How loud should the game be when the silencer is enabled?",
```

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,70 @@
...
+public class SilencerFeature extends Feature {
+    @RegisterConfig
+    public final Config<Double> silencerVolume = new Config<>(0.01);
+
+    @RegisterKeyBind
+    private final KeyBind silencerKeyBind =
+            new KeyBind("Toggle Silencer", GLFW.GLFW_KEY_UNKNOWN, true, this::toggleSilencer);
+
+    private final Storage<Double> lastVolume = new Storage<>(1.0);
+    private boolean isSilencerEnabled = false;
```
kristofbolyai, line 46:
```suggestion

    private boolean isSilencerEnabled = false;
```

kristofbolyai: This boolean should be a config or storage. Please check what gammabrightfeature does and you can almost copy it as a whole, since you want to achieve the exact same behavior.

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -307,6 +308,7 @@ public void init() {
         registerFeature(new GammabrightFeature());
         registerFeature(new TerritoryDefenseMessageFeature());
         registerFeature(new TranslationFeature());
+        registerFeature(new SilencerFeature());
```
kristofbolyai: Keep alphabetical order

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: I initially started out that way, but my understanding is that using a config or storage is primarily for reapplying the settings from the "last session". I argue this isn't the optimal approach for a feature typically activated for specific tasks, such as a grind session.

A common scenario could be players enable the Silencer, grind, and then abruptly log off when finished (ALT+F4). If the Silencer was reapplied automatically upon their return, it might lead to confusion, prompting them to manually reset the volume to a "normal" level.

With this consideration, I deemed the reapplication of settings unnecessary and therefore the use of config or storage. This is why I simplified to a regular Boolean. However, we can still use config or storage, but without reapplying the settings. Or I could make the reapplication a toggleable option but this further complicates a really simple Feature.

Id like to know what you think before I proceed with updating the code.

kristofbolyai: Okay, I get what you mean, but I want to think about how to refactor this in a way that is both logical and somewhat signals that this is a "session only" setting.

kristofbolyai: The problem is with reseting on quiting, how can you be so sure you correctly reset the volume when a crash/alt-f4/exit happens? I feel like this has potential of being buggy which is worse than people not remembering that it is turned on (which can be solved by an overlay like text or smth)

re-vc: In this case id suggest we instead trigger the reset onConnect or onTitleScreenInit, this way we solve both issues at the same time.

kristofbolyai: Reset on `onTitleScreenInit`, but even that might not work with autoconnect, etc.

re-vc: Ill do some testing, i believe it will work, especially if i use the .Pre flag, as far as i remember autoconnect uses .Post

re-vc:
We need to keep the `onDisconnect` to handle scenarios where the user disconnects from the server but doesn't return to the TitleScreen. The feature operates as expected, consistently resetting the volume back to the original value when the "session ends."

I've tested the following scenarios:

- AutoConnect, both on and off
- Closing the game via the "x" button, ALT+F4, and the "quit game" button
- Disconnecting from the server through the disconnect button, and by interrupting internet access

ryanzhoudev (approved):
In the future it would be cool to see some overlay with the status of silencer/gammabright, only appearing when their respective features are enabled.

magicus:

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final Storage<Double> lastVolume = new Storage<>(1.0);
+
+    private boolean isSilencerEnabled = false;
+
+    @SubscribeEvent
+    public void onDisconnect(WynncraftConnectionEvent.Disconnected event) {
+        if (isSilencerEnabled) {
+            resetVolume();
+            isSilencerEnabled = false;
```
magicus, line 54: Setting of isSilencerEnabled should be done inside resetVolume and enableSilencer.

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onTitleScreenInit(TitleScreenInitEvent.Pre event) {
+        resetVolume();
```
magicus: ... otherwise you can easily get cases like this, where the volume is actually reset, but we do not track this so we still think the silencer is enabled.

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,77 @@
...
+                    Component.translatable("feature.wynntils.silencer.disabled").withStyle(ChatFormatting.RED));
+        }
+    }
+
+    private void resetVolume() {
+        McUtils.options().getSoundSourceOptionInstance(SoundSource.MASTER).set(lastVolume.get());
+    }
+
+    private void enableSilencer() {
+        lastVolume.store(McUtils.options()
```
magicus: This code's readability would benefit from extracting the vanilla volume into a local variable.

common/src/main/java/com/wynntils/features/SilencerFeature.java
magicus: Also, why is this needed?

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,77 @@
...
+@ConfigCategory(Category.UNCATEGORIZED)
+public class SilencerFeature extends Feature {
+    @RegisterConfig
+    public final Config<Double> silencerVolume = new Config<>(0.01);
+
+    @RegisterKeyBind
+    private final KeyBind silencerKeyBind =
+            new KeyBind("Toggle Silencer", GLFW.GLFW_KEY_UNKNOWN, true, this::toggleSilencer);
+
+    private final Storage<Double> lastVolume = new Storage<>(1.0);
```
magicus: This don't have to be a storage. You will never need to save volume between Minecraft runs.

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: I agree, will update this.

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: The decision to reset the volume is based on my design choice to enhance user-friendliness. This feature is intended as a "session-only" setting, typically used for grinding sessions or multitasking. I see two primary user scenarios:

1. Users finish their task, decide to do something different on the server, notice their sound is "silenced", and remember to disable the Silencer themselves.
2. Users log off and/or close the game by any means and return later, possibly on a different day.

In the second scenario, we reset the volume to the their value. It's unrealistic to expect users to remember the last Silencer state between different gaming sessions. To facilitate resetting on both disconnect and in subsequent Minecraft runs, we need `lastVolume` to be a storage.

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: Indeed, the `onDisconnect` function does handle a similar task, but as @kristofbolyai pointed out, we can't guarantee that it always successfully resets the volume, which might lead to unreliable behavior.
Hence, we reset the volume at the start of the game, specifically just before the `TitleScreen` loads.
Additionally we keep the `onDisconnect` to handle scenarios where the user disconnects from the server but doesn't return to the `TitleScreen`.

Regarding why we don't set `isSilencerEnabled` in `onTitleScreenInit`, the Boolean is defaulted to false on launch, but the volume is not reset. Furthermore, accessing the TitleScreen is only possible after disconnecting from the server, a prerequisite to set the Boolean to true.

Overall I think its not necessary to set `isSilencerEnabled` to false every time `resetVolume` is called but we certainly can do that anyway.

Let me know what you think!

common/src/main/java/com/wynntils/features/SilencerFeature.java
magicus: But the volume is a vanilla option setting that is persisted between runs, yes? I'm not sure I follow the logic here. If we start the silencer, the volume will go from lets say 100 to 1. This is stored in vanillas option, and we store the 100 in our lastVolume. Then the user quits. Or the game crashes, or whatever.

When they restart, the vanilla volume will be 1. But if they log on to Wynncraft, Wynntils will think the silencer is disabled. So if they enable the silencer, it will store 1 as the lastVolume, and set the volume to 0.01.

This seems broken.

If this is not the case, what am I missing?

common/src/main/java/com/wynntils/features/SilencerFeature.java
magicus: I had missed this conversation since it was marked resolved. I think it touches on the same worry I have. There is no point in storing the lastValue unless you also store whether it is enabled or not. Possibly you should do it like this: Have two storage options, "enabled" and "volume". Whenever we enable the silencer, you set "enabled" to true and "volume" to the current volume. 

If we start Wynntils and the silencer is enabled, then we somehow exited the game with the silencer enabled (this should not really happen, so it is a crash or bug). Then you do the normal disable dance, that is, set "enabled" to false and the vanilla volume to the stored "volume".

magicus: You can also consider merging the two stored variables into one, by setting "lastVolume" (or perhaps better `originalVolume`?) to -1 if the silencer is disabled. (Given that negative volumes are not allowed). This way you can never even get out of sync between the two variables.

common/src/main/java/com/wynntils/features/SilencerFeature.java
magicus: So is this an attempt to restore the volume at startup in case of a crash? I think you should better do that in the constructor. That will fix the volume for the entire minecraft game as soon as the mod loads. This event will be sent only when our event framework is up and running. And it will be resent (I think) whenever you get back to the title screen from a game.

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: Yes you understood correctly but you missed that we attempt to `resetVolume` during `onDisconnect` and additionally we also "force" this during "launch" `onTitleScreenInit`.

Therefore, instead of the volume transitioning from 100 to 1 and then to 0.01, the process looks more like this:

1. We toggle Silencer to 'on'
2. The volume goes from 100 to 1
3. We store 100 in `lastVolume`; Vanilla settings store 1
4. Upon a crash, exit, etc., we **attempt** to toggle Silencer to 'off' and we try to load Volume 100 from `lastVolume`
5. Upon reboot, we force Silencer to 'off' note that this is because of `private boolean isSilencerEnabled = false;` is the starting state
7. Vanilla settings load with Volume at 1
8. Vanilla settings attempt to load the Title Screen
9. We load Volume 100 from `lastVolume`
10. Vanilla settings finally load the Title Screen and we can proceed with `autoConnect` or nothing.

common/src/main/java/com/wynntils/features/SilencerFeature.java
re-vc: > And it will be reset (I think) whenever you get back to the title screen from a game.

Yes, you're correct. In the current design, the volume is reset to `lastVolume` every time you return to the Title Screen, which indeed is an oversight. To address this, I suggest introducing a new local variable `currentVolume`. We would then perform a check to see if `currentVolume` equals `silencerVolume`. Based on this check, we either proceed with resetting to `lastVolume` or do nothing. This check ensures that we only reset the volume when necessary, which should resolve the issue.

> So is this an attempt to restore the volume at startup in case of a crash?

No, not exactly. The intention behind this design isn't specifically to restore the volume at startup in case of a crash. Instead, it's primarily to reapply the user's vanilla volume value.

magicus: Ok, I think I understand this process a bit better now. There is a lot of cases, and I am not sure these are all correctly handled, or even what the correct way to handle them would be.

Imagine that you log onto Wynncraft and enable the silencer. Then you log out. I expect the normal vanilla volume to be restored. I believe this happens, and that the onConnect + onTitleScreen is basically a safeguard to make sure this really happens, regardless.

But what happens when you log back on to Wynncraft? Should the silencer still be on? If so, we need to change the volume upon connection. If not, we must make sure we do not believe the silencer is enabled. It is still not obvious to me that we have such a guarantee, but maybe I'm just being stupid. :-)

To help with making the class understandable, can you make things clearer by renaming `lastVolume` to `originalVolume`? And `resetVolume` to `restoreOriginalVolume`?

Also, I would like for you to write some documentation (like a block at the top of the class) of how this all comes together, summarizing what you have described here in the comments.

magicus:
I think not all comments have been addressed. Please let me know when you think they have, and I'll do a re-review.

re-vc:
> I think not all comments have been addressed. Please let me know when you think they have, and I'll do a re-review.

Hi, sorry I’ve been busy and couldn’t allocate any time to this PR as well as I did not want to take up your time since there were more pressing matters. I’ll update the code and incorporate the feedback today.

re-vc:
To prevent `restoreOriginalVolume` from being called every time `onTitleScreenInit` is called, I have added a boolean variable called `firstTitleScreenInit` that keeps track of whether `onTitleScreenInit` has been triggered at least once. Once it has, the code block inside `onTitleScreenInit` will no longer be executed.

I have also attempted to address all the comments.
Please let me know if I missed anything or if any new issues arose.

common/src/main/java/com/wynntils/features/SilencerFeature.java
```diff
@@ -0,0 +1,93 @@
...
+        if (isSilencerEnabled) {
+            restoreOriginalVolume();
+            isSilencerEnabled = false;
+        }
+    }
+
+    @SubscribeEvent
+    public void onTitleScreenInit(TitleScreenInitEvent.Pre event) {
+        if (!firstTitleScreenInit) return;
+        if (!masterVolume().get().equals(silencerVolume.get())) return;
```
magicus, line 61:
```suggestion
        if (!masterVolume().get().equals(silencerVolume.get())) return;
        // If the previous game crashed with the silencer enabled, restore normal volume now
```

magicus (commented):
Just add a comment to why you  need the title init, and you're good to go.

magicus:
@re-vc Thank you for your contribution, and for hanging in there! It was an extraordinary bumpy review process, partly due to lack of time from me and bolyai, and partly because of the bad timing with the 2.0.3 beta which stole most of our focus.

re-vc:
Thank you for working on this with me, yes I noticed my timing was very bad hahah

PR #1644, opened 2023-06-22
fix: Fix npc dialogues getting stuck after completing quests and quest started messages not parsing

kristofbolyai:


common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -316,7 +316,7 @@ private void handleFakeChatLine(Component message) {
      * message entirely.
      */
     private Component postChatLine(Component message, StyledText styledText, MessageType messageType) {
-        String plainText = message.getString();
+        String plainText = styledText.getStringWithoutFormatting();
         if (!plainText.isBlank()) {
             // We store the unformatted string version to be able to compare between
             // foreground and background versions
```
kristofbolyai, line 322: I am glad we had these comments here, otherwise I could not have recognized that we rely on `Component#getString()` to get an unformatted string, which broke in many cases.

PR #1645, opened 2023-06-25
feat: Discord rich presence

ryanzhoudev:


common/src/main/java/com/wynntils/mc/mixin/LocalPlayerMixin.java
```diff
@@ -49,4 +50,23 @@ private void playSoundPre(SoundEvent sound, float volume, float pitch, CallbackI
...
+
+    @Inject(
+            method = "Lnet/minecraft/client/player/LocalPlayer;sendPosition()V",
+            at = {
+                @At(
+                        value = "NEW",
+                        target = "(DDDFFZ)Lnet/minecraft/network/protocol/game/ServerboundMovePlayerPacket$PosRot;"),
+                @At(
+                        value = "NEW",
+                        target = "(DDDFF)Lnet/minecraft/network/protocol/game/ServerboundMovePlayerPacket$Pos;"),
```
kristofbolyai: I am wondering how you got this signature. I checked and the constructor doesn't even have that many variables.

 ```suggestion
                         target = "(DDDZ)Lnet/minecraft/network/protocol/game/ServerboundMovePlayerPacket$Pos;"),
```

Can you confirm which one is correct?

File: common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java [hunk 2]
```diff
@@ -0,0 +1,101 @@
...
+import com.wynntils.mc.event.SetXpEvent;
+import com.wynntils.models.character.event.CharacterUpdateEvent;
+import com.wynntils.models.territories.profile.TerritoryProfile;
+import com.wynntils.models.worlds.event.WorldStateEvent;
+import com.wynntils.models.worlds.type.WorldState;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class DiscordRichPresenceFeature extends Feature {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,101 @@
...
+    @SubscribeEvent
+    public void onPlayerTeleport(PlayerTeleportEvent event) {
+        handleLocationChange();
+    }
+
+    @SubscribeEvent
+    public void onCharacterUpdate(CharacterUpdateEvent event) {
+        if (displayCharacterInfo.get()) {
+            Managers.Discord.setClassType(Models.Character.getClassType());
```
kristofbolyai: IMO this is a little weird connection between managers and functions, perhaps @magicus can suggest a better way when he gets to review this.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,101 @@
...
+    @SubscribeEvent
+    public void onWorldChange(WorldStateEvent event) {
+        if (displayWorld.get()) {
+            if (event.getNewState() == WorldState.WORLD) {
+                Managers.Discord.setWorld(Models.WorldState.getCurrentWorldName());
+            } else if (event.getNewState() == WorldState.HUB || event.getNewState() == WorldState.CONNECTING) {
+                Managers.Discord.setLocation("");
+                Managers.Discord.setWorld("In Hub");
+                Managers.Discord.setWynncraftLogo();
+            }
```
kristofbolyai: Use a switch case please

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,101 @@
...
+            Managers.Discord.setWorld("");
+        }
+
+        if (event.getNewState() == WorldState.CHARACTER_SELECTION) {
+            if (displayLocation.get()) {
+                Managers.Discord.setLocation("Selecting a character");
+            }
+            Managers.Discord.setClassType(null);
+            Managers.Discord.setWynncraftLogo();
+        }
```
kristofbolyai: This should just fall under the display location config and into the switch

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,92 @@
...
+        core.activityManager().updateActivity(activity);
+    }
+
+    public void setWorld(String world) {
+        activity.setState(world);
+        core.activityManager().updateActivity(activity);
+    }
+
+    public void clearAll() {
+        core.activityManager().clearActivity();
```
kristofbolyai: Perhaps we should free the activity here and create a new one if needed. `activity.close()` and set it to null.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,92 @@
...
+        CreateParams params = new CreateParams();
+        try {
+            params.setClientID(387266678607577088L);
+            params.setFlags(CreateParams.getDefaultFlags());
+            core = new DiscordGameSDKCore(params);
+            activity = new Activity();
+            activity.timestamps().setStart(Instant.now());
+        } catch (RuntimeException e) {
+            WynntilsMod.error("Could not initialize Discord Game SDK", e);
+        }
```
kristofbolyai: I don't think this should run here. Let's add some dependency tree in here, basically like `ChatHandler`'s `dialogExtractionDependents`. I want to give the user the ability to completely bypass running Discord SDK. With this said, you need to move this to an enable method, and create a disable method where you need to clean up all your `AutoCloseable`s.

Lastly, I think you should widen the catch to `Throwable` however I don't think we can catch SDK loading errors.

kristofbolyai (commented):
Sorry for the many comments, I did not get to tell you how I imagined the manager to work yesterday. Overall it is really close and you just need to refactor it so it can be "turned off".

As for testing. Can you let me know what platforms you tested on? I want to test 2 things on a platform. Running the game with Discord open **and with Discord closed**. I heard some issues back in Wynntils with running the SDK with a closed Discord, let's make sure it doesn't happen here.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,92 @@
...
+import com.wynntils.mc.event.TickAlwaysEvent;
+import com.wynntils.models.character.type.ClassType;
+import com.wynntils.utils.mc.McUtils;
+import java.time.Instant;
+import java.util.List;
+import java.util.Locale;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class DiscordManager extends Manager {
+
```
kristofbolyai, line 23:
```suggestion
```

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,92 @@
...
+    private int level = 0;
+    private ClassType classType = null;
+
+    public DiscordManager() {
+        super(List.of());
+
+        DiscordGameSDKCore.loadLibrary();
+        CreateParams params = new CreateParams();
+        try {
+            params.setClientID(387266678607577088L);
```
kristofbolyai: Can you perhaps move this into a const?

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,92 @@
...
+    private Activity activity;
+
+    private int level = 0;
+    private ClassType classType = null;
+
+    public DiscordManager() {
+        super(List.of());
+
+        DiscordGameSDKCore.loadLibrary();
+        CreateParams params = new CreateParams();
```
kristofbolyai: Oh, I just noticed. You are leaking memory here with `params`. It is auto closable, but you mustn't close it until you close core. Which means you need to store this as a field too. I am thinking if you should have a holder record for all these so you don't forget to close all of them.

common/src/main/java/com/wynntils/mc/mixin/LocalPlayerMixin.java
ryanzhoudev: Probably copilot and I didn't notice since the first one gets tripped a lot anyway. DDDZ is correct.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private int level = 0;
+    private ClassType classType = null;
+
+    public DiscordManager() {
+        super(List.of());
+    }
+
+    public void load() {
+        if (!libraryLoaded) {
```
kristofbolyai: Comments mention that library can be reloaded so I am not sure if we should complicate this here.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus, line 1: I'd rather see this in `core/net/discord`, analogous to `core/net/athena` (and setting the pattern for any future net services).

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.discord;
+
+import com.wynntils.antiope.core.DiscordGameSDKCore;
```
magicus, line 7: Is this weird package name due to shadowing? How come we do not see that for other 3rd party libraries?

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,121 @@
...
+import com.wynntils.mc.event.TickEvent;
+import com.wynntils.models.character.type.ClassType;
+import com.wynntils.utils.mc.McUtils;
+import java.time.Instant;
+import java.util.List;
+import java.util.Locale;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class DiscordManager extends Manager {
+    private static final long DISCORD_APPLICATION_ID = 387266678607577088L;
```
magicus, line 21: What is this magic number? Is it like a API key, something we have registered with Discord?

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
kristofbolyai: Not sure what you mean?

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: Actually, this is borderline Manager stuff... It is part of the small but tricky-to-handle set of stuff that is neither "essential to Wynntils operations" (which characterize Managers) nor "a conceptual part of the Wynncraft world" (which characterize Models). It is more like an `ExternalModel` or something like that. 

Other stuff that fall into this category is Hades (which is considered a `Model`) and the Splashes (which is considered a `Manager`).

kristofbolyai: This is so much what I suggested, but you have not got around checking discord :). I even linked the original discussing about this, when I made PingModel.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
magicus: You mean between managers and features, I presume. 

Well, maybe yes. I think there are two reasonable splits to make here, and the current solutions blends those together.

Option 1 is to have a DiscordManager/DiscordExternalModel (or whatever we should call it) that is basically a thin wrapper around the discord API. In that case, it should not know about concepts like "Location" but rather "Large Text" (or whatever it was called).

Option 2 is to have the DiscordManager/DiscordExternalModel specify in detail how we are allowed to set rich presence info on discord, and deal in terms of "Location" and "Class" etc. But in that case, the model should itself listen to the events needed to keep it up to date, which renders the Feature almost moot. At this point, it would only be there to store config and have a way to turn this on/off; since we can't do that for Models yet.

I'm not sure which I prefer. But I think Option 1 have some kind of appeal. That would make the DiscordManager expose a more high level API with text and images (I guess?) to display, and the responsibility of setting those to sane Wynncraft values should fall on this feature. I think that would be neat.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
ghost: This is what we call a Client ID, unique to each application registered on Discord Developer Portal. You can read more about them here: https://discord.com/developers/docs/topics/rpc#connecting

common/src/main/java/com/wynntils/mc/mixin/LocalPlayerMixin.java
```diff
@@ -49,4 +50,23 @@ private void playSoundPre(SoundEvent sound, float volume, float pitch, CallbackI
...
+                @At(
+                        value = "NEW",
+                        target = "(DDDFFZ)Lnet/minecraft/network/protocol/game/ServerboundMovePlayerPacket$PosRot;"),
+                @At(
+                        value = "NEW",
+                        target = "(DDDZ)Lnet/minecraft/network/protocol/game/ServerboundMovePlayerPacket$Pos;"),
+            },
+            cancellable = true)
+    private void sendPositionPre(CallbackInfo ci) {
+        PlayerMoveEvent event = new PlayerMoveEvent();
```
magicus: This strikes me a bit as odd, that we have a event for moving that does not even include the new location we're moving to.

I wonder if we really need to do this. Afaict, this is used to update the Territory info in Discord. I think we might just as well check the current territory every second or so, and update Discord if it has changed. This looks like it add unnecessary overhead in a lag-sensitive part of the game.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: That an external library gets stuck under `com.wynntils`. But I see we have the same when shadowing the Hades lib, so I guess I'll just have to live with it. (If it is possible to make shadowing land in e.g. `com.wynntils.shadow.*` instead of just `com.wynntils`, I'd prefer that, but it is separate from this PR.

kristofbolyai: It is in com.wynntils since we made the package (from a fork)

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: In that case, I think it is a bit disturbing to hard-code it in the source. I wonder if there is a way that we can read it from a suitable json file instead. urls.json springs to mind, but that only contains URLs. Hm. I need to think a bit about this.

magicus:
Also, @DonkeyBlaster, thanks for implementing this! That was one of the remaining major functionality outages from Legacy in Artemis. Overall I think this is good, we just need to find the best way to fit it in the Artemis design.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: I think (no promises!) that I actually will have some time now to do a bit working on Artemis! :) I'll check up on Discord once I've finished handling the Github notifications.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
DevScyu: This Client ID should be fine to hard code, as long as it is our application that we have already setup, which it looks to be. This value should never change unless discord does something funky

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: Aha, I see. I missed that part. Then it all makes sense.

ryanzhoudev:
> As for testing. Can you let me know what platforms you tested on? I want to test 2 things on a platform. Running the game with Discord open **and with Discord closed**. I heard some issues back in Wynntils with running the SDK with a closed Discord, let's make sure it doesn't happen here.

On Windows, launching the game without Discord open just opens Discord. Same thing as Legacy. Should probably find a way to prevent this.

kristofbolyai:
Can I ask to not implement this while we have the beta and pre-alpha stuff going on? Once we sort out the Wynn update, we are going to come back to this and implement this, it has been missing for too long now.

kristofbolyai:
> > As for testing. Can you let me know what platforms you tested on? I want to test 2 things on a platform. Running the game with Discord open **and with Discord closed**. I heard some issues back in Wynntils with running the SDK with a closed Discord, let's make sure it doesn't happen here.
> 
> On Windows, launching the game without Discord open just opens Discord. Same thing as Legacy. Should probably find a way to prevent this.

Just to note this, it is easy to do this. You have to set CreateParams to not require Discord open, and if an exception happens in the callback loop, you know Discord is not opened. This case, we can just close the whole thing and implement a `/wynntils reannouncerichpresence` (or smth) command

ryanzhoudev:
Works fine on Windows now, from my testing. Someone please check macOS and Linux.

magicus:
I just tested this on my M1 mac, and it works fine!

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,115 @@
...
+            activity.timestamps().setStart(Instant.now());
+        } catch (GameSDKException e) {
+            if (e.getResult() == Result.INTERNAL_ERROR) {
+                // Occurs when player launches game without Discord open
+                return;
+            }
+            WynntilsMod.error("Could not initialize Discord Game SDK", e);
+        } catch (Throwable e) {
+            WynntilsMod.error("Could not initialize Discord Game SDK", e);
+        }
```
kristofbolyai: we could merge these and do an instanceof

magicus:
I tested that it worked fine in updating my info when I had discord started and then started Minecraft. I also checked that it did not start Discord by itself if it was not running.

However, if I started Discord after I loaded Minecraft, we did not discover this and start sending updates. I think we should. At the very least try to look for discord again each time we switch character or something like that.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,115 @@
...
+        activity.setState(state);
+        core.activityManager().updateActivity(activity);
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent event) {
+        if (isReady()) {
+            core.runCallbacks();
+        }
+    }
```
kristofbolyai: can we do smth like hades model does and only tick once like every 5 ticks?

magicus:
What happens if you don't even have Discord installed? Have you tested that?

Also, is there some way we can detect if that is the case, and if so, skip trying to re-detect Discord as I suggested above?

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
kristofbolyai, line 1: I think this is the singular file blocking this feature.. This feature does way too much. What blocks us from having a simple "updatePresenceInfo" method and that can get called by all the events, etc?

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
magicus: At first I thought you were borderline crazy to suggest using instanceof on exceptions, but now I see what you mean. The special case is if the exception is GameSDKException and the result is INTERNAL_ERROR, then we accept it as okay.

So yes, then I agree.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @RegisterConfig
+    public final Config<Boolean> displayWorld = new Config<>(true);
+
+    private boolean stopTerritoryCheck = false;
+    private TerritoryProfile lastTerritoryProfile = null;
+    private ClassType classType = null;
+    private int level = 0;
+
+    private void setCharacterDetails() {
```
magicus: I don't think it is a formal rule, but I prefer to put the `@SubscribeEvent` methods before private methods, and I think I've noticed @kristofbolyai does the same.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,154 @@
...
+            TerritoryProfile territoryProfile = Models.Territory.getTerritoryProfileForPosition(
+                    McUtils.player().position());
+            if (territoryProfile != null && territoryProfile != lastTerritoryProfile) {
+                lastTerritoryProfile = territoryProfile;
+                String location = territoryProfile.getName();
+                Managers.Discord.setDetails(location);
+            }
+        }
+
+        Managers.TickScheduler.scheduleLater(this::checkTerritory, 10);
```
magicus: Please make this value a constant

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,154 @@
...
+        Managers.Discord.setImageText(name + " - Level " + level + " " + classType.getName());
+        Managers.Discord.setImage(classType.getActualName(false).toLowerCase(Locale.ROOT));
+    }
+
+    @SubscribeEvent
+    public void onCharacterUpdate(CharacterUpdateEvent event) {
+        if (!Models.WorldState.onWorld()) return;
+
+        // classType needs to be set even when config is disabled so if the config is enabled later, it will not require
+        // a relog
```
magicus: Why is that? You are calling setCharacterDetails() from onConfigChange.

magicus (commented):
Overall this looks very good, apart from my remarks/questions.

Thanks for filling in this blank from the legacy partity!

ryanzhoudev:
> What happens if you don't even have Discord installed? Have you tested that?
> 
> Also, is there some way we can detect if that is the case, and if so, skip trying to re-detect Discord as I suggested above?

I believe not having Discord installed and not having Discord launched will result in the same state for our code.

magicus:
> I believe not having Discord installed and not having Discord launched will result in the same state for our code.

Sounds reasonable, now that I think about it. :) I was confused by the "launch Discord if not already running" part.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public void setState(String state) {
+        if (!isReady()) return;
+        activity.setState(state);
+        core.activityManager().updateActivity(activity);
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent event) {
+        if (!isReady() || ticksUntilUpdate > 0) return;
```
magicus: Maybe do ticksUntilUpdate-- somewhere here..? :-)

Also, please check ticksUntilUpdate first, since it is quickest. isReady() calls `.core.isOpen`. This sounds like it might be expensive; at the very least we should only check it once the TICKS_PER_UPDATE ticks have passed.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
magicus: You think so? I think it does exactly what it needs to, neither too much nor to little. Possibly I like it because @DonkeyBlaster is using the same kind of coding style that I am using :-) but I think it looks okay. What do you mean that the alternative would be? That every change, whatever it is, would trigger us to recalculate the entire discord info?

You would still need to do the WorldStateEvent thing, where you check if you are on the hub etc. And if you get a callback from the territory check every 10th tick or so, it makes no sense to be re-checking if you are on the hub or not every time that happens; instead it makes sense to just update the territory.

So I don't think it would be an improvement. 

Or I misunderstand you..?

ryanzhoudev: If we do the updatePresenceInfo thing, wouldn't we be delegating too much to the manager?

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
ryanzhoudev: setCharacterDetails() doesn't set the feature's classType field, it just tells the rich presence to display what is already stored. This is a misleading name now that you brought it up, I will change it.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
magicus: I'm assuming the updatePresenceInfo() method should be in the Feature.

common/src/main/java/com/wynntils/core/discord/DiscordManager.java
```diff
@@ -0,0 +1,122 @@
...
+    @SubscribeEvent
+    public void onTick(TickEvent event) {
+        if (ticksUntilUpdate > 0) {
+            ticksUntilUpdate--;
+            return;
+        }
+        if (!isReady()) return;
+
+        core.runCallbacks();
+        ticksUntilUpdate = TICKS_PER_UPDATE;
```
magicus: You need to move this up to before the isReady test, otherwise it will be tested every tick if it fails, since ticksUntilUpdate will remain at 0.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
ryanzhoudev: That is almost what displayCharacterDetails() does then. Currently it just sets it to the stored fields in the feature, presumably this proposed function will just get the details at the time and set it immediately. Currently the correct events set the stored fields.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
magicus: I was more wondering why you need a classType field at all, and not just read it from the model whenever you need to change the class info?

ryanzhoudev: Originally it was so that when the display function was called it wouldn't have to ask the model for the classType as much, but I suppose both are reading some field and that's it, so it would be the same for performance

magicus (approved):
I think this looks great now. If @kristofbolyai strongly disagrees, we'll have to continue discuss how this should be solved. But from my PoV, this is perfectly fine.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -0,0 +1,154 @@
...
+        if (displayCharacterInfo.get()) {
+            displayCharacterDetails();
+        }
+    }
+
+    @SubscribeEvent
+    public void onXpChange(SetXpEvent event) {
+        if (!Models.WorldState.onWorld()) return;
+
+        // same as above, level needs to be set even when config is disabled
```
magicus: This comment seems to be left behind, though...

kristofbolyai (approved):
Okay, I can accept this. Just make sure you merge it in at a time where you have time to deal with potential issues. I've had a long 3 days debugging container queries...

ryanzhoudev:
I have time to deal with issues today, tomorrow, and the afternoon of the day after, so I will merge now.

PR #1649, opened 2023-06-29
feat: Highlight search when reopening screen

Mcrtin:


magicus:
What problem is this supposed to solve?

(As a general rule, please include a description of what the PR is trying to accomplish when posting it. That is what the `Description` field is for.)

Mcrtin:
when reopening for example the settings screen you can just begin to type instead of first erasing the text and your last search will still be shown when reopening the screen.

magicus:
Ok, I see. It is a bit odd to put this code in onClose -- what you in effect are doing is to prepare for the next appearance of the screen when you close it. I'd rather see this done prior to showing the screen. But now I can't find a proper hook for that, so I can't recommend a concrete alternative. Maybe @kristofbolyai knows?

If we can't figure out a better place, at least the onClose needs to be commented to say something along the lines of your explanation here, that we prepare the search text to be selected in preparation of the next time the screen appears, to ease replacement of the text.

kristofbolyai:
> Ok, I see. It is a bit odd to put this code in onClose -- what you in effect are doing is to prepare for the next appearance of the screen when you close it. I'd rather see this done prior to showing the screen. But now I can't find a proper hook for that, so I can't recommend a concrete alternative. Maybe @kristofbolyai knows?

From a quick look I think `net.minecraft.client.gui.screens.Screen#added` should be the correct method.

kristofbolyai (approved):

PR #1651, opened 2023-06-30
feat: Add custom player list overlay

Mcrtin:
Ports the legacy player info overlay (https://github.com/Wynntils/Wynntils/blob/development/src/main/java/com/wynntils/modules/utilities/overlays/hud/PlayerInfoOverlay.java).
This includes the animation and an option to change the duration of it.


common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,252 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.ui;
+
+import static com.wynntils.utils.render.type.HorizontalAlignment.CENTER;
+import static com.wynntils.utils.render.type.HorizontalAlignment.LEFT;
+import static com.wynntils.utils.render.type.VerticalAlignment.MIDDLE;
```
ryanzhoudev: Please be consistent and remove the static imports, just use VerticalAlignment.x and HorizontalAlignment.x like you have already in some places

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,252 @@
...
+                    .map(styledText -> styledText.replace("§7", "§0"))
+                    .map(StyledText::fromString)
+                    .toList();
+
+            return lastPlayers;
+        }
+
+        private static String wrapText(String input, int maxLength) {
+            if (McUtils.mc().font.width(input) <= maxLength) return input;
+
```
ryanzhoudev: Since this method does not actually wrap text onto the next line, can you rename the method? Can you also rename maxLength to maxWidth, since length implies the number of characters and width is being checked here? Lastly, this would probably be a better fit in StringUtils instead of a feature.

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,252 @@
...
+    public static final int ROLL_WIDTH = 27;
+
+    public static final int HALF_WIDTH = 178;
+
+    public static final int WIDTH = HALF_WIDTH * 2;
+
+    public static final int TOTAL_WIDTH = WIDTH + ROLL_WIDTH * 2;
+
+    public static final int MAX_LENGTH = 73;
```
ryanzhoudev: Why do all of these need to be public with all the extra blank lines?

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,252 @@
...
+
+            lastPlayers = McUtils.player().connection.getListedOnlinePlayers().stream()
+                    .sorted(PLAYER_INFO_COMPARATOR)
+                    .limit(80)
+                    .map(defaultTabList::getNameForDisplay)
+                    .map(StyledText::fromComponent)
+                    .filter(styledText -> !styledText.contains("§l"))
+                    .map(StyledText::getString)
+                    .map(styledText -> wrapText(styledText, MAX_LENGTH))
+                    .map(styledText -> styledText.replace("§7", "§0"))
```
ryanzhoudev: Can you use the ChatFormatting enum instead of raw codes?

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,217 @@
...
+@ConfigCategory(Category.OVERLAYS)
+public class PlayerInfoFeature extends Feature {
+    private static final Comparator<PlayerInfo> PLAYER_INFO_COMPARATOR =
+            Comparator.comparing(playerInfo -> playerInfo.getProfile().getName(), String::compareToIgnoreCase);
+    private static final int DISTANCE_BETWEEN_CATEGORIES = 87;
+    private static final int ROLL_WIDTH = 27;
+    private static final int HALF_WIDTH = 178;
+    private static final int WIDTH = HALF_WIDTH * 2;
+    private static final int TOTAL_WIDTH = WIDTH + ROLL_WIDTH * 2;
+    private static final int MAX_WIDTH = 73;
```
kristofbolyai: Perhaps move this into the overlay class since it only used there.

common/src/main/java/com/wynntils/mc/mixin/PlayerTabOverlayMixin.java
```diff
@@ -0,0 +1,25 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(PlayerTabOverlay.class)
+public class PlayerTabOverlayMixin {
+
+    @Inject(method = "render", at = @At("HEAD"), cancellable = true)
+    private void render(PoseStack poseStack, int width, Scoreboard scoreboard, Objective objective, CallbackInfo ci) {
+        if (Managers.Feature.getFeatureInstance(PlayerInfoFeature.class).isEnabled()) ci.cancel();
```
kristofbolyai: This is not how we do this. Check out what `com.wynntils.mc.event.RenderEvent.ElementType` is and how it is used to replace GUI elements by other overlays.

common/src/main/resources/assets/wynntils/textures/tab_overlay.png
kristofbolyai, line 1: I know this is how it was added in legacy but in Artemis, I cut these textures down to the smallest size they can fit.

common/src/main/java/com/wynntils/utils/StringUtils.java
```diff
@@ -144,4 +146,17 @@ public static ByteBuffer decodeBase64(String base64) {
...
+        if (font.width(input) <= maxWidth) return input;
+
+        StringBuilder builder = new StringBuilder();
+        for (char c : input.toCharArray()) {
+            if (font.width(builder.toString() + c) > maxWidth) break;
+            builder.append(c);
+        }
+
+        return builder.toString();
+    }
```
kristofbolyai: com.wynntils.utils.mc.RenderedStringUtils

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    public final PlayerInfoOverlay playerInfoOverlay = new PlayerInfoOverlay();
+
+    private static class PlayerInfoOverlay extends Overlay {
```
kristofbolyai: I am not even sure if this is an actual overlay and if it should be one. What do you think @magicus?

Mcrtin:
it seems to be a bit shifted to the right
image

magicus:

magicus:
The name might be copied from legacy, but it is cr*p. I recommend `CustomPlayerList` instead.

common/src/main/java/com/wynntils/features/ui/PlayerInfoFeature.java
magicus: Maybe...? I agree that it is a bit borderline. I have not studied the code fully, but if we indeed do the entire rendering of the tab list, and not just draw a pretty background, then I think it can and should be an overlay. Since you can then move it around the screen, etc. That it is not showing anything most of the time is not really a problem, we have other overlays that behave like that, like the NPC dialogue overlay.

If we just decorate the existing vanilla tab list, so things will look messed up if we move this overlay around, then it should definitely not be an overlay.

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -287,6 +288,7 @@ public void init() {
         registerFeature(new CustomLoadingScreenFeature());
         registerFeature(new CustomSeaskipperScreenFeature());
         registerFeature(new LobbyUptimeFeature());
+        registerFeature(new CustomPlayerListFeature());
```
ShadowCat117: This should be moved up to be listed alphabetically

magicus:
I failed to formulate a question properly, so it never got answered. :) Let me try again:

Should this really be an overlay?

I have not studied the code fully, but if we indeed do the entire rendering of the tab list, and not just draw a pretty background, then I think it can and should be an overlay. Since you can then move it around the screen, etc. 

If we just decorate the existing vanilla tab list, so things will look messed up if we move this overlay around, then it should definitely not be an overlay.

@Mcrtin Can you please confirm that it is possible to move this overlay around.

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,224 @@
...
+
+@ConfigCategory(Category.OVERLAYS)
+public class CustomPlayerListFeature extends Feature {
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    public final CustomPlayerListOverlay customPlayerListOverlay = new CustomPlayerListOverlay();
+
+    @SubscribeEvent
+    public void onRender(RenderEvent.Pre event) {
+        if (event.getType() == RenderEvent.ElementType.GUI) event.setCanceled(true);
```
magicus: By the name of it, it sounds broader than just the tab list..?

magicus: Also, always use braces after if statements. The only exception is trivial returns, break/continue etc.

Mcrtin:
This Is a completely independent overlay. I disable the vanilla one with the onRender listener.

common/src/main/java/com/wynntils/mc/mixin/PlayerTabOverlayMixin.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Mixin(PlayerTabOverlay.class)
+public class PlayerTabOverlayMixin {
+
+    @Inject(method = "render", at = @At("HEAD"), cancellable = true)
+    private void render(PoseStack poseStack, int width, Scoreboard scoreboard, Objective objective, CallbackInfo ci) {
+        RenderEvent.Pre renderEvent =
+                new RenderEvent.Pre(poseStack, 0, McUtils.window(), RenderEvent.ElementType.PLAYER_TAB_LIST);
+        MixinHelper.post(renderEvent);
+        if (renderEvent.isCancelable()) ci.cancel();
```
magicus:
```suggestion
        if (renderEvent.isCancelable()) {
            ci.cancel();
        }
```

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,224 @@
...
+                    0,
+                    Texture.PLAYER_INFO_OVERLAY.width(),
+                    Texture.PLAYER_INFO_OVERLAY.height(),
+                    Texture.PLAYER_INFO_OVERLAY.width(),
+                    Texture.PLAYER_INFO_OVERLAY.height());
+        }
+
+        private double getAnimation() {
+            if (openingDuration.get() <= 0) return 1;
+            if (lastTime == -1) lastTime = System.currentTimeMillis() - 1;
```
magicus:
```suggestion
            if (lastTime == -1) {
                lastTime = System.currentTimeMillis() - 1;
            }
```

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,224 @@
...
+
+@ConfigCategory(Category.OVERLAYS)
+public class CustomPlayerListFeature extends Feature {
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    public final CustomPlayerListOverlay customPlayerListOverlay = new CustomPlayerListOverlay();
+
+    @SubscribeEvent
+    public void onRender(RenderEvent.Pre event) {
+        if (event.getType() == RenderEvent.ElementType.PLAYER_TAB_LIST) event.setCanceled(true);
```
magicus:
```suggestion
        if (event.getType() == RenderEvent.ElementType.PLAYER_TAB_LIST) {
            event.setCanceled(true);
        }
```

common/src/main/java/com/wynntils/core/features/FeatureManager.java
magicus:
```suggestion
        registerFeature(new CustomLoadingScreenFeature());
        registerFeature(new CustomPlayerListFeature());
        registerFeature(new CustomSeaskipperScreenFeature());
        registerFeature(new LobbyUptimeFeature());
```

magicus (commented):
Almost there. I helped you with suggestions for the remaining issues, then we can merge.

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    private static class CustomPlayerListOverlay extends Overlay {
```
kristofbolyai: From what I can read, this overlay can't render a preview. Shift can't be held while managing overlays without causing problems.. You need to have a method for rendering and call that in renderPreview. In normal render, you can check for the key, animations, etc.

Mcrtin:
Preview:
image

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,242 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features.ui;
```
magicus: I just noticed this. The feature should reside in the package corresponding to its `@Category`.

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,242 @@
...
+                            OverlayPosition.AnchorSection.TOP_MIDDLE),
+                    TOTAL_WIDTH,
+                    Texture.PLAYER_INFO_OVERLAY.height());
+        }
+
+        private List<StyledText> lastPlayers = new ArrayList<>();
+        private long nextExecution = 0;
+
+        private List<StyledText> getAvailablePlayers() {
+            if (System.currentTimeMillis() < nextExecution && !lastPlayers.isEmpty()) return lastPlayers;
```
magicus: You don't need the lastPlayers check. If it is less than 250 ms since last check, you can (and should!) always skip the lookup. 

I think it is unlikely that the player list ever is empty (it does include yourself, does it not?), but if it is, we should not spam the lookup nevertheless.

magicus: Otherwise this will put a heavy toll, since it is called from rendering, if we do this lookup for each frame.

common/src/main/java/com/wynntils/features/ui/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,242 @@
...
+                    0,
+                    0,
+                    ROLL_WIDTH,
+                    Texture.PLAYER_INFO_OVERLAY.height(),
+                    Texture.PLAYER_INFO_OVERLAY.width(),
+                    Texture.PLAYER_INFO_OVERLAY.height());
+        }
+
+        private void renderPlayerNames(PoseStack poseStack, float categoryStart) {
+            List<StyledText> players = getAvailablePlayers();
```
magicus: Since this is a potentially expensive operation, and not strictly just "rendering", can you please move it up one level to renderPlayerList, and pass the `players` as an argument to it instead?

Mcrtin:
Maybe we could also make an animation util class for the animation percentage?

kristofbolyai:
> Maybe we could also make an animation util class for the animation percentage?

This sounds exactly like something we can't really do since every overlay differs. Even if we sometime add this, let it be us, we know better what we are doing.

Mcrtin:
because i disabled scisoring when the player list is fully opened I noticed another bug:
image
the right roll is rendering twice
I think this is because of texture wrapping or what its called
well, Idk how to fix this @kristofbolyai can you help?

magicus:
Does this rendering bug mean that the PR is not ready for integration? Otherwise I was close to approving it, from a code quality point of view.

Mcrtin:
the problem basically is, that i have a `u` offset but I don't subtract that from the width so the texture rect is wider than the texture. I tried every possible combination of subtracting `ROLL_WIDTH` from the texture width in `renderBackground` but that messed up the aspect ratio of the texture.
DO NOT merge it until this is fixed and because I don't know so much about rendering it may take a while :(

MenacingPerson:
Um... I don't know how to say it...

It has no texture variation to it, so it looks bad

Maybe add some cracks or tears? I can try it later

kristofbolyai:
Sorry, I will take a look at this tomorrow. I've been busy with 2.0. 3 stuff.

magicus:
@MenacingPerson 

> Um... I don't know how to say it...
> 
> It has no texture variation to it, so it looks bad

The texture is the original legacy texture. It sure could be improved in the future, but for this PR it is more than enough to have Legacy parity.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,240 @@
...
+
+    @OverlayInfo(renderType = RenderEvent.ElementType.GUI)
+    public final CustomPlayerListOverlay customPlayerListOverlay = new CustomPlayerListOverlay();
+
+    @SubscribeEvent
+    public void onRender(RenderEvent.Pre event) {
+        if (event.getType() == RenderEvent.ElementType.PLAYER_TAB_LIST) {
+            event.setCanceled(true);
+        }
+    }
```
kristofbolyai, line 52: Replace this with `OverlayInfo` replace. Also, render on PLAYER_TAB_LIST, not GUI.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,240 @@
...
+                            styledText.replace(ChatFormatting.GRAY.toString(), ChatFormatting.BLACK.toString()))
+                    .map(StyledText::fromString)
+                    .toList();
+
+            return lastPlayers;
+        }
+
+        @Override
+        public void render(PoseStack poseStack, MultiBufferSource bufferSource, float partialTicks, Window window) {
+            if (!McUtils.options().keyPlayerList.isDown() && animationProgress <= 0.0) return;
```
kristofbolyai: If you render at the correct hook, then you don't need this keybind check.

kristofbolyai:
> because i disabled scisoring when the player list is fully opened I noticed another bug: image the right roll is rendering twice I think this is because of texture wrapping or what its called well, Idk how to fix this @kristofbolyai can you help?

Can you point out what code part this is? Otherwise, I will clone this tomorrow and check

Mcrtin:
have a look at `renderBackground`.
We can also, instead of fixing this, break the texture up into two parts so we don't have to worry about getting the right chunk.
fixed.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
Mcrtin: When I render at the correct hook the animation stops working because render doesn't get called after you have released the `tab` key. It would have worked if I had there `||` instead of `&&`

kristofbolyai:
image
image

This makes me question if this really should be an overlay..

common/src/main/java/com/wynntils/mc/mixin/PlayerTabOverlayMixin.java
```diff
@@ -0,0 +1,31 @@
...
+
+@Mixin(PlayerTabOverlay.class)
+public class PlayerTabOverlayMixin {
+
+    @Inject(method = "render", at = @At("HEAD"), cancellable = true)
+    private void render(PoseStack poseStack, int width, Scoreboard scoreboard, Objective objective, CallbackInfo ci) {
+        RenderEvent.Pre renderEvent =
+                new RenderEvent.Pre(poseStack, 0, McUtils.window(), RenderEvent.ElementType.PLAYER_TAB_LIST);
+        MixinHelper.post(renderEvent);
+        if (renderEvent.isCancelable()) {
```
kristofbolyai:
```suggestion
        if (renderEvent.isCanceled()) {
```

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+        private final TimelimitedExecutioner<List<StyledText>> availablePlayers =
+                new TimelimitedExecutioner<>(CustomPlayerListOverlay::getAvailablePlayers, Duration.ofMillis(250));
+
+        protected CustomPlayerListOverlay() {
+            super(
+                    new OverlayPosition(
+                            0,
+                            0,
+                            VerticalAlignment.TOP,
+                            HorizontalAlignment.LEFT,
```
kristofbolyai:
```suggestion
                            HorizontalAlignment.CENTER,
```
This should fix the positioning

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+            this.cooldown = cooldown;
+        }
+
+        public T get() {
+            Duration difference = Duration.between(lastExecutionTime, Instant.now());
+            if (difference.minus(cooldown).isNegative()) return current;
+            lastExecutionTime = Instant.now();
+            return current = method.get();
+        }
+    }
```
kristofbolyai: This is an util class that should be moved into `com/wynntils/utils/type`. This could also use a better name, @magicus do you have any suggestions?

kristofbolyai (commented):
Also, a small bug. When in the lobby, player names get rendered in the friends tab. IIRC the custom overlay didn't render in the lobby in legacy, perhaps it should the case here as well.

kristofbolyai:
If you could move the text a pixel up it would be perfect, I think. 
image

kristofbolyai:
> have a look at `renderBackground`. We can also, instead of fixing this, break the texture up into two parts so we don't have to worry about getting the right chunk. fixed.

I can't see what's wrong. It renders the middle part, and the two renderlScrolls render the correct scrolls.

common/src/main/java/com/wynntils/utils/mc/RenderedStringUtils.java
```diff
@@ -93,6 +93,19 @@ public static StyledText trySplitOptimally(StyledText line, float maxWidth) {
         }
     }
 
+    public static String cut(String input, int maxWidth) {
```
kristofbolyai:
```suggestion
    public static String substringMaxWidth(String input, int maxWidth) {
```

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+                    Texture.PLAYER_INFO_OVERLAY.height());
+        }
+
+        @Override
+        protected void onConfigUpdate(ConfigHolder configHolder) {
+            animationPercentage.setOpeningDuration(Duration.of(openingDuration.get(), ChronoUnit.MILLIS));
+        }
+    }
+
+    private static final class AnimationPercentage {
```
kristofbolyai: Perhaps even this could be moved `com/wynntils/utils/render/type` although it doesn't fully fit there either.

kristofbolyai (commented):
I think this basically looks good after you fix all the suggestions.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.gui.components.PlayerTabOverlay;
+import net.minecraft.client.multiplayer.PlayerInfo;
+import net.minecraft.client.renderer.MultiBufferSource;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.joml.Math;
+
+@ConfigCategory(Category.OVERLAYS)
+public class CustomPlayerListFeature extends Feature {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+            this.openingDuration = openingDuration;
+        }
+
+        public boolean finishedClosingAnimation() {
+            return openingProgress == 0;
+        }
+    }
+
+    private static final class TimelimitedExecutioner<T> {
+
```
kristofbolyai:
```suggestion
```

I really don't like extra empty lines after classes, I wish spotless would clean it up

common/src/main/java/com/wynntils/mc/mixin/PlayerTabOverlayMixin.java
```diff
@@ -0,0 +1,31 @@
...
+import net.minecraft.world.scores.Objective;
+import net.minecraft.world.scores.Scoreboard;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(PlayerTabOverlay.class)
+public class PlayerTabOverlayMixin {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
magicus: I agree. Finally got around to fixing it: #1779.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+    private static final class AnimationPercentage {
+        private final Supplier<Boolean> openingContitions;
+        private Duration openingDuration;
+        private Instant lastTime = Instant.EPOCH;
+
+        private double openingProgress;
+
+        private AnimationPercentage(Supplier<Boolean> openingContitions, Duration openingDuration) {
+            this.openingContitions = openingContitions;
```
magicus: openingConditions maybe?

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
magicus: `ThrottledSupplier`

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+        private TimelimitedExecutioner(Supplier<T> method, Duration cooldown) {
+            this.method = method;
+            this.cooldown = cooldown;
+        }
+
+        public T get() {
+            Duration difference = Duration.between(lastExecutionTime, Instant.now());
+            if (difference.minus(cooldown).isNegative()) return current;
+            lastExecutionTime = Instant.now();
+            return current = method.get();
```
magicus: There is no need to return an assignment. Split into two lines.

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -0,0 +1,289 @@
...
+        public void setOpeningDuration(Duration openingDuration) {
+            this.openingDuration = openingDuration;
+        }
+
+        public boolean finishedClosingAnimation() {
+            return openingProgress == 0;
+        }
+    }
+
+    private static final class TimelimitedExecutioner<T> {
```
magicus: ... implements Supplier<T> maybe?

magicus:
> This makes me question if this really should be an overlay..

Is there no way for us to make overlays not change with GUI scale?

kristofbolyai:
> > This makes me question if this really should be an overlay..
> 
> 
> 
> Is there no way for us to make overlays not change with GUI scale?

We do by default. Turns out this was just bad alignment by default, not sure how it wasn't caught

common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
magicus: @Mcrtin That means that if you merge in the latest main and run spotlessApply, this will be fixed automatically.

common/src/main/java/com/wynntils/utils/type/AnimationPercentage.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.type;
```
magicus: I am pretty certain bolyai said `utils.render.type`. That is where it belongs, anyway.

magicus (commented):
I think it is just the placement of AnimationPercentage, and the `@OverlayInfo` type left. Then I think this is good to go. Thanks for your hard work!

Mcrtin:
I can't render the custom player list instead of the default one because of the closing animation

magicus:
Hm. Seems the closing animation is playing tricks with our assumptions. I wonder if it is possible to separate that into some special handling? Maybe use the tick scheduler for drawing the closing animation..? (Just thinking out loud)

kristofbolyai:
> Hm. Seems the closing animation is playing tricks with our assumptions. I wonder if it is possible to separate that into some special handling? Maybe use the tick scheduler for drawing the closing animation..? (Just thinking out loud)

I think we can accept that in it's current state. It is a bit hacky, yes, but the system wasn't designed for this kind of use.

magicus (approved):

magicus:
@kristofbolyai So you are okay with the remaining hacks?

kristofbolyai (approved):
Yes, it is fine for now. We can revisit later to clean up, if needed.

PR #1652, opened 2023-06-30
feat: Add guild name and rank to book above player

ShadowCat117:
See https://github.com/Wynntils/Artemis/pull/1652#issuecomment-1614697107 for newer video demonstration


common/src/main/java/com/wynntils/screens/wynntilsmenu/WynntilsMenuScreen.java
```diff
@@ -218,6 +218,18 @@ private static void renderPlayerInfo(PoseStack poseStack, int mouseX, int mouseY
         InventoryScreen.renderEntityInInventoryFollowsMouse(
                 poseStack, posX, posY, 30, posX + 45 - mouseX, posY - 40 - mouseY, McUtils.player());
 
+        FontRenderer.getInstance()
```
ShadowCat117: Would it be better to only render this is `Models.Character.getGuild()` isn't blank or is this fine?

ShadowCat117:
Setting to draft quickly as I've found a different approach that I can also get guild rank from

ShadowCat117:
See https://github.com/Wynntils/Artemis/pull/1652#issuecomment-1615873480 for newer screenshot

https://github.com/Wynntils/Artemis/assets/8337467/fd6b41ad-8aee-436c-b700-a5d763ba426d

ShadowCat117:
Is it possible to access `GuildRankReplacementFeature` from `CharacterModel`? That way I can have it change the rank to whatever they have it set to in their config.

ryanzhoudev (dismissed):
Please create a new GuildModel and move all guild info parsing to it. CharacterModel is not meant for this type of thing.

You can leave it in models/players, and there are examples available in FriendsModel and PartyModel. Remember to register it in core/components/Models.java.

common/src/main/java/com/wynntils/screens/wynntilsmenu/WynntilsMenuScreen.java
ryanzhoudev: Yes, there should be a check for if the player is actually in a guild or not, and skip the render if they're not. No point rendering empty strings.

ryanzhoudev:
> Is it possible to access `GuildRankReplacementFeature` from `CharacterModel`? That way I can have it change the rank to whatever they have it set to in their config.

bolyai or magicus can correct me on this if I'm wrong, but I'm not sure you're supposed to access feature configs from other places. Maybe the easiest solution here is to render both the rank name and the number of stars by default.

ShadowCat117:
> > Is it possible to access `GuildRankReplacementFeature` from `CharacterModel`? That way I can have it change the rank to whatever they have it set to in their config.
> 
> bolyai or magicus can correct me on this if I'm wrong, but I'm not sure you're supposed to access feature configs from other places. Maybe the easiest solution here is to render both the rank name and the number of stars by default.

I thought that would be the case, I'll try something like "Rank (stars) of guildName)

ShadowCat117:
How does this look?

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final Pattern GUILD_NAME_MATCHER = Pattern.compile("§3(.*?)§b.*");
+    private static final Pattern GUILD_RANK_MATCHER = Pattern.compile("§7Rank: §f(.*)");
+
+    private String name = "";
+    private String rank = "";
+    private String rankStars = "";
+
+    public GuildModel() {
+        super(List.of());
+    }
```
kristofbolyai: This should depend on charactermodel as it parses it's data from there.

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -0,0 +1,66 @@
...
+            if (rankMatcher.matches()) {
+                rank = rankMatcher.group(1);
+
+                switch (rank.toLowerCase(Locale.ROOT)) {
+                    case "owner" -> rankStars = "★★★★★";
+                    case "chief" -> rankStars = "★★★★";
+                    case "strategist" -> rankStars = "★★★";
+                    case "captain" -> rankStars = "★★";
+                    case "recruiter" -> rankStars = "★";
+                    default -> rankStars = "";
```
kristofbolyai: Perhaps it's time to use `GuildRank` and extend it with star info.

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.regex.Pattern;
+import net.minecraft.world.item.ItemStack;
+
+public class GuildModel extends Model {
+    private static final Pattern GUILD_NAME_MATCHER = Pattern.compile("§3(.*?)§b.*");
+    private static final Pattern GUILD_RANK_MATCHER = Pattern.compile("§7Rank: §f(.*)");
+
+    private String name = "";
+    private String rank = "";
+    private String rankStars = "";
```
kristofbolyai: Use `GuildRank`

ryanzhoudev (commented):
This looks fine for now imo, but there is a small issue where leaving and joining guilds doesn't update the GuildModel with the correct information. I understand if it's difficult to test as you don't want to leave your existing guild, so I am ok with merging and I can fix it on my alt. Let me know.

ShadowCat117:
> This looks fine for now imo, but there is a small issue where leaving and joining guilds doesn't update the GuildModel with the correct information. I understand if it's difficult to test as you don't want to leave your existing guild, so I am ok with merging and I can fix it on my alt. Let me know.

Yes that would be great thanks.

Also I'm sure you thought of it, but reading the promotion/demotion chat message could also be used to update rank. I'm not sure if there is a leave/kicked message though.

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -13,12 +13,26 @@ public enum GuildRank {
...
 
     private final String name;
+    private final String stars;
 
     GuildRank(String name) {
         this.name = name;
+
+        switch (this.name) {
+            case "Owner" -> stars = "★★★★★";
+            case "Chief" -> stars = "★★★★";
```
ryanzhoudev: Actually, one last thing - can we remove this switch and just have it directly in the enum values?

ryanzhoudev (approved):

magicus:

magicus (approved):
Looks good! (But please hold it for now)

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void parseGuildInfoFromGuildMenu(ItemStack guildInfoItem) {
+        List<StyledText> lore = LoreUtils.getLore(guildInfoItem);
+
+        for (StyledText line : lore) {
+            Matcher guildNameMatcher = line.getMatcher(GUILD_NAME_MATCHER);
+            if (guildNameMatcher.matches()) {
+                name = guildNameMatcher.group(1);
+                continue;
```
magicus, line 35: Is name always before rank? Otherwise we will fail to get a rank. Perhaps move the rank block above the name block so they get in the "logical" order, and add a comment stating that this is so.

common/src/main/java/com/wynntils/screens/wynntilsmenu/WynntilsMenuScreen.java
```diff
@@ -218,6 +218,38 @@ private static void renderPlayerInfo(PoseStack poseStack, int mouseX, int mouseY
         InventoryScreen.renderEntityInInventoryFollowsMouse(
                 poseStack, posX, posY, 30, posX + 45 - mouseX, posY - 40 - mouseY, McUtils.player());
 
+        if (!Models.Guild.getGuildName().isEmpty()) {
+            String rank = Models.Guild.getGuildRank().getName();
+
+            if (!rank.equals("Recruit")) {
+                rank += " (" + Models.Guild.getGuildRank().getStars() + ")";
+            }
```
magicus: Extract this to a Models.Guild.getGuildDescription(), so you don't need to expose getStars. Also, you have an enum. Do not get a string from it and do a string comparison! O_o

magicus (changes requested):
My comments seems to have disappeared from github. :( Fix this two things and it's good to go.

common/src/main/java/com/wynntils/models/players/GuildModel.java
ShadowCat117: I think so, I'll double check later. If someone is able to check if they appear differently for different ranks that'd be great

ShadowCat117: Yeah name before rank
image

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -5,20 +5,30 @@
...
         this.name = name;
+        this.stars = stars;
     }
 
     public String getName() {
         return name;
     }
+
+    public String getGuildDescription() {
+        if (name.equals(RECRUIT.getName())) {
```
magicus:
```suggestion
        if (this == RECRUIT) {
```

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -5,20 +5,30 @@
...
+            return "";
```
magicus:
```suggestion
            return name();
```

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -5,20 +5,30 @@
...
+        } else {
+            return " (" + stars + ")";
```
magicus:
```suggestion
            return name() + " (" + stars + ")";
```

common/src/main/java/com/wynntils/screens/wynntilsmenu/WynntilsMenuScreen.java
```diff
@@ -218,6 +218,36 @@ private static void renderPlayerInfo(PoseStack poseStack, int mouseX, int mouseY
         InventoryScreen.renderEntityInInventoryFollowsMouse(
                 poseStack, posX, posY, 30, posX + 45 - mouseX, posY - 40 - mouseY, McUtils.player());
 
+        if (!Models.Guild.getGuildName().isEmpty()) {
+            String rank = Models.Guild.getGuildRank().getName();
+
+            rank += Models.Guild.getGuildRank().getGuildDescription();
```
magicus:
```suggestion
            String rank = Models.Guild.getGuildRank().getGuildDescription();
```

ShadowCat117:
Should I add getStars back and just not use it, in case some future feature wants it?

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -5,20 +5,30 @@
...
 
     public String getName() {
         return name;
     }
+
+    public String getGuildDescription() {
+        if (this == RECRUIT) {
+            return name();
+        } else {
+            return name() + " (" + stars + ")";
```
ShadowCat117: Should this just be `name`? With `name()` it returns the rank name in full caps, I think it looks cleaner with just the first letter being capitalised

magicus: Yeah, you are right! My mistake. `name()` is the intrinsic enum method. It should be `getName()` or `name`; chose whichever variant you prefer.

magicus (approved):

PR #1657, opened 2023-07-01
fix: fix version check

DevScyu:
Fixes bug in version stream check


kristofbolyai (approved):
Seems goos but I only checked this on phone.

PR #1662, opened 2023-07-01
fix: Reset button in settings menu highlights even if you undo the changes

Mcrtin:
what it should look like:
image

what it looks like when you double-press the enabled/disabled button:
image


kristofbolyai:
I don't even know what this fixes? I tested the reset button and the color behaves correctly for me.

Mcrtin:
Sry I should have clarified that I meant the enabled/disabled button which you double click

magicus:
I'm not sure this is a bug? If you start setting your own values, we consider the value "managed" by the user. Even if it is set to the same value as the default. This way you can make sure a value does not change even if the default changes in the future. The only way to go back to treating the value as "not managed" by the user is by pressing Reset.

This is similar to e.g. how about:config works in Firefox.

PR #1663, opened 2023-07-02
fix: Changing class does not clear status overlay

ryanzhoudev:


kristofbolyai (approved):
Thanks, this bothered me for a while now..

PR #1666, opened 2023-07-02
chore: Speed up build by caching (correctly)

kristofbolyai:


kristofbolyai:
Since the version changes in the build gradle, we can't rely on it's hash. Revert this if it somehow causes weird CI failures.

PR #1667, opened 2023-07-02
fix: Beta/pre-alpha usability fixes

magicus:
fix: Add option to connect to beta from Wynncraft button.
fix: Hard-code development version to behave as pre-alpha.
fix: On pre-alpha, warn if user connects to non-beta server.

### DIFF
--- a/common/src/main/java/com/wynntils/core/WynntilsMod.java
+++ b/common/src/main/java/com/wynntils/core/WynntilsMod.java
@@ -113,7 +113,8 @@ public final class WynntilsMod {
     }
 
     public static boolean isPreAlpha() {
-        return version.contains("pre-alpha");
+        // This is hard-coded in the development branch
+        return true;
     }
 
     public static boolean isDevelopmentBuild() {
--- a/common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
@@ -36,12 +36,17 @@ import org.apache.commons.lang3.Validate;
 
 @ConfigCategory(Category.UI)
 public class WynncraftButtonFeature extends Feature {
-    private static final String GAME_SERVER = "play.wynncraft.com";
-    private static final String LOBBY_SERVER = "lobby.wynncraft.com";
+    private static final String GAME_SERVER = "play";
+    private static final String BETA_SERVER = "beta";
+    private static final String LOBBY_SERVER = "lobby";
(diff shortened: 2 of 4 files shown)


common/src/main/java/com/wynntils/core/WynntilsMod.java
```diff
@@ -113,7 +113,8 @@ public static String getVersion() {
     }
 
     public static boolean isPreAlpha() {
-        return version.contains("pre-alpha");
+        // This is hard-coded in the development branch
+        return true;
```
kristofbolyai, line 117: We will have to not forget to revert this

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -36,13 +36,18 @@
 
 @ConfigCategory(Category.UI)
 public class WynncraftButtonFeature extends Feature {
-    private static final String GAME_SERVER = "play.wynncraft.com";
-    private static final String LOBBY_SERVER = "lobby.wynncraft.com";
+    private static final String GAME_SERVER = "play";
+    private static final String BETA_SERVER = "beta";
+    private static final String LOBBY_SERVER = "lobby";
```
kristofbolyai, line 41: Technically we also have "hub"

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
magicus: Is that different from lobby?

common/src/main/java/com/wynntils/core/WynntilsMod.java
magicus: Hopefully we will get a merge conflict here. If you want to be sure that we do, I can make a separate change to main, hardcoding it to `false`.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai: No, just a secondary name.

kristofbolyai: (For connecting it should not matter, I forgot this is not a model..)

common/src/main/java/com/wynntils/core/WynntilsMod.java
kristofbolyai: If you want to go ahead, but I won't require it.

PR #1670, opened 2023-07-02
feat: Custom bank quick jump pages

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/baa4a802-dc96-4122-a6dd-248ece3067af

Once this is reviewed and cleaned up I'll add it for the misc bucket as a seperate feature


common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+    private int pageDestination = 1;
+    private List<Integer> customJumpDestinations;
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent e) {
+        if (!(e.getScreen() instanceof AbstractContainerScreen<?> screen)) return;
+
+        StyledText title = StyledText.fromComponent(screen.getTitle());
+
+        Matcher pageMatcher = title.getMatcher(BANK_PAGE_MATCHER);
```
ryanzhoudev: Can you use ContainerModel#isBankScreen instead? If you need access to the page numbers, you should also implement a helper method there.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+                    buttonSixPage.get());
+
+            if (currentPage != pageDestination && quickJumping) {
+                goToPage();
+            } else if (quickJumping) {
+                quickJumping = false;
+            }
+        } else {
+            return;
+        }
```
ryanzhoudev, line 94: This entire else clause is not needed. Also, for consistency, when you switch to the ContainerModel method, you can just do an inverted check like you did at the top.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@ConfigCategory(Category.INVENTORY)
+public class CustomBankPagesFeature extends Feature {
+    // If possible, lock these to only allow between 1-21
```
ryanzhoudev: You could use the onConfigUpdate event to check if any of these are invalid, then reset it and tell the user to fix it.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+    public void onContainerClose(ContainerCloseEvent.Post e) {
+        currentPage = 1;
+        pageDestination = 1;
+    }
+
+    @SubscribeEvent
+    public void onSlotClicked(ContainerClickEvent e) {
+        if (e.getContainerMenu().containerId != containerId) {
+            return;
+        }
```
ryanzhoudev:
```suggestion
        if (e.getContainerMenu().containerId != containerId) return;
```

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+
+        buttonSlot.getItem().setHoverName(Component.literal("§7Jump to Page " + buttonDestination));
+    }
+
+    private void goToPage() {
+        quickJumping = true;
+
+        if (currentPage == pageDestination) {
+            return;
+        }
```
ryanzhoudev:
```suggestion
        if (currentPage == pageDestination) return;
```

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+
+        if (pageMatcher.matches()) {
+            if (BUTTON_SLOTS.contains(e.getSlot().index)) {
+                renderQuickJumpButton(e.getPoseStack(), e.getSlot());
+            }
+        }
+
+        // Prevent buying a new page
+        if (e.getSlot().index == NEXT_PAGE_SLOT
+                && e.getSlot().getItem().getHoverName().getString().contains("§c")
```
ryanzhoudev: See if you can use ContainerModel#isLastBankPage

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onRenderSlot(SlotRenderEvent.Pre e) {
+        if (!(e.getScreen() instanceof AbstractContainerScreen<?> screen)) return;
+
+        StyledText title = StyledText.fromComponent(screen.getTitle());
+
+        Matcher pageMatcher = title.getMatcher(BANK_PAGE_MATCHER);
```
ryanzhoudev: Running this regex on every render is expensive, maybe there is a way to check when the page updates instead?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,221 @@
...
+                Texture.COLUMN_ARROW_RIGHT.width(),
+                Texture.COLUMN_ARROW_RIGHT.height());
+
+        buttonSlot.set(new ItemStack(Items.SNOW));
+
+        int buttonIndex = BUTTON_SLOTS.indexOf(buttonSlot.index);
+        // When config limits 1-21, change to just customJumpDestinations.get(buttonIndex)
+        int buttonDestination = Math.max(1, Math.min(MAX_BANK_PAGES, customJumpDestinations.get(buttonIndex)));
+
+        buttonSlot.getItem().setHoverName(Component.literal("§7Jump to Page " + buttonDestination));
```
ryanzhoudev: Can you use the ChatFormatting enum instead of raw § codes?

ryanzhoudev:
Thanks for helping port this over!

magicus:

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: Or we should extend our config types to include a ranged integer...

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: I think that is fine. We do that elsewhere, I think. @kristofbolyai, what do you say?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: Yeah, the render method does way to much. I think we can solve this even without a render method. Can't you just replace the items in the corresponding slots with something that has the wanted icon as texture, and just use vanilla rendering?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: I'll use onConfigUpdate for now, but if/when ranged integer gets added it can be changed to that

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: Had a go at this, I've got it not running the regex each time but obviously it still calls `renderQuickJumpButton` each time.

But I've not been able to get it to replace the item, I'm not even sure what item uses the texture anyway, it might be NBT data.

If you're able to point me in the right direction I can try again, but no rush if no features aren't being merged until 2.0.3 is released

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
kristofbolyai: We use them in a few places yes, while it is usually recommended to use ChatFormatting, it is not strictly necessary.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
Mcrtin: try changing the item tag CustomModelData

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
kristofbolyai: Marked as resolved but you didnt change it?

ShadowCat117: I think you're viewing an old version

Fixed in latest https://github.com/Wynntils/Artemis/blob/8a79c17154af466804098f740b263269a326fd7a/common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,226 @@
...
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@ConfigCategory(Category.INVENTORY)
+public class CustomBankPagesFeature extends Feature {
+    // Change to ranged integer when implemented
+    @RegisterConfig
```
magicus, line 35: This duplication irks me, but I guess it is up to us not providing a proper array type for configs... It is the same with command hotkeys, iirc.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,226 @@
...
+                buttonOnePage.get(),
+                buttonTwoPage.get(),
+                buttonThreePage.get(),
+                buttonFourPage.get(),
+                buttonFivePage.get(),
+                buttonSixPage.get());
+
+        if (onLastPage) {
+            quickJumping = false;
+            pageDestination = currentPage;
```
magicus, line 89: What is pageDestination representing? I don't follow the flow here.

ShadowCat117: pageDestination is the page you want to jump to, so starts out as 1 so when you open the bank so that it doesn't auto start jumping somewhere.

When you click the next/previous buttons that will increment/decrement the value so that it doesn't auto jump back to the page you were on. This is also needed because the scroll feature can't just tell this feature that a scroll was performed.

Then once you click one of the jump buttons it will use the destination to find which of the default destinations is the closest and each time a new page is rendered, if you're on the last page you own, it will set this page to your destination to stop it from trying to go back to the destination if you hit the previous page button.

Since I don't own page 18, if I tried to jump to 18 it would take me to 17 and then if I clicked the previous button it would try and go back go 18 so destination has to get set to current page if on the last page

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: Yeah, I've been looking into the custom bank page names and that's going to have 21 configs...

magicus:
I would like to spend some more time on this PR. Please give me a day or two to fully understand if this is okay, or need more work.

ShadowCat117:
> I would like to spend some more time on this PR. Please give me a day or two to fully understand if this is okay, or need more work.

Yeah no worries :)

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: But if you were at the last page and wanted to quickjump to the first page? Wouldn't this cancel that behavior?

magicus:
I am still struggling to really make heads or tails of the flow in this feature. I think it is unnecessary convoluted. My gut feeling is that we have the wrong state stored in the instance, but I can't pin-point the problem, not give any more suggestions at this point.

However, I also think it will (fully, or at least mostly) work as intended, and it is a welcome addition. 

I realize I cannot give much more advice on the code as part of a review. Instead, I suggest that we merge this as it is now, and then I might give it a shot at trying to refactor this once it is part of the code base. Does that seem ok to you @JamieCallan117?

ShadowCat117:
> I am still struggling to really make heads or tails of the flow in this feature. I think it is unnecessary convoluted. My gut feeling is that we have the wrong state stored in the instance, but I can't pin-point the problem, not give any more suggestions at this point.
> 
> However, I also think it will (fully, or at least mostly) work as intended, and it is a welcome addition. 
> 
> I realize I cannot give much more advice on the code as part of a review. Instead, I suggest that we merge this as it is now, and then I might give it a shot at trying to refactor this once it is part of the code base. Does that seem ok to you @JamieCallan117?

I did have a thought last night on a better way to render the buttons so if you give me a few hours to try that out and I'll see if I can make some improvements elsewhere whilst I'm at it, will also add some comments to hopefully clear some things up. But after that, sure :)

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: Looking at it, yes I think it would.

Changing it to`if (onLastPage && pageDestination > currentPage)` should fix it I think

magicus:
> I did have a thought last night on a better way to render the buttons so if you give me a few hours to try that out and I'll see if I can make some improvements elsewhere whilst I'm at it, will also add some comments to hopefully clear some things up. But after that, sure :)

I still think the best way would be if we could actually replace the item for the jump slots. But that means we would need to find an item in the Wynn resource pack that seems to fit. I did look through some recently, but found no really good ideas. Best perhaps would be the teleport scroll icon, it looks kind of like you are going to a "page". If we did that. we could skip our individual rendering at all.

JamieCallan117:
> I still think the best way would be if we could actually replace the item for the jump slots. But that means we would need to find an item in the Wynn resource pack that seems to fit. I did look through some recently, but found no really good ideas. Best perhaps would be the teleport scroll icon, it looks kind of like you are going to a "page". If we did that. we could skip our individual rendering at all.

I think the only problem with replacing the item, for me personally, is that it will look like just a regular item without the different background style that the custom texture has.

Tried a prismarine shard as a quick test since it looks a bit like an arrow and to me it just looks like another item in my bank, a teleport scroll might be even more difficult to tell apart.

Maybe it's only me who thinks that way and if you think this or a teleport scroll will work then I can change it to that.

ShadowCat117:
> I still think the best way would be if we could actually replace the item for the jump slots. But that means we would need to find an item in the Wynn resource pack that seems to fit. I did look through some recently, but found no really good ideas. Best perhaps would be the teleport scroll icon, it looks kind of like you are going to a "page". If we did that. we could skip our individual rendering at all.

Ok, I've found 3 potential items, I think the scroll with the background would be my choice as it looks more like a button than an item in your bank. The ability tree arrow doesn't really fit the scene but it would be the most obvious as a button.

What do you think?

ShadowCat117:
An extra thought actually, could have a config option for any of those choices or no button. Just deciding which one to use as the default then

magicus:
Too many config choices is costly as well. I think the scroll with the background looks perfect for the purpose! You could also add the Legacy overlay with the target bank page number, by setting the item count to the target page. That will also give a bit of a clue what page this is taking you to.

ShadowCat117:
> Too many config choices is costly as well. I think the scroll with the background looks perfect for the purpose! You could also add the Legacy overlay with the target bank page number, by setting the item count to the target page. That will also give a bit of a clue what page this is taking you to.

Ooo yes that's a good idea.

I'm almost finished rewriting the flow of it and will push that soon

ShadowCat117:
I think it's a bit better but if you still want to improve it further then feel free too.

Also, setting the button items still happens in `onRenderSlot` I tried doing it in a few other events but it didn't work in the ones I tried, so I might just not know the best one for it.

Mcrtin:
did you try the set slot event?

ShadowCat117:
> did you try the set slot event?

I did, but there didn't seem to be a method to set the ItemStack, only get

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,215 @@
...
+
+        customJumpDestinations = List.of(
+                buttonOnePage.get(),
+                buttonTwoPage.get(),
+                buttonThreePage.get(),
+                buttonFourPage.get(),
+                buttonFivePage.get(),
+                buttonSixPage.get());
+
+        if (quickJumping && pageDestination > lastPage) {
```
magicus: Try converting this based on
`if (!quickJumping) return;`

and then simplify the following if statements

magicus:
Overall, this looks much better and closer to what I had in mind!

The "is on last page" detection is still somewhat spread out. I'm not sure if all checks for that are needed, perhaps it is a bit of safety guards as well?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,215 @@
...
+            e.getSlot().set(jumpButton);
+        }
+
+        if (e.getSlot().index == NEXT_PAGE_SLOT && Models.Container.isLastBankPage(screen)) {
+            lastPage = currentPage;
+        }
+    }
+
+    private void jumpToDestination() {
+        quickJumping = true;
```
magicus, line 152: There is a bit of structure here that seems to be longing to be exposed to the world. :-) Basically, this is a glorified switch case based on the value of `currentPage`. Can you help bring that structure forth?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,215 @@
...
+                int jumpDistance = Math.abs(jumpDestination - pageDestination);
+
+                if (jumpDistance < closestDistance) {
+                    closest = jumpDestination;
+                    closestDistance = jumpDistance;
+                }
+            }
+
+            if (closestDistance < currentDistance) {
+                clickJumpButton(QUICK_JUMP_DESTINATIONS.indexOf(closest));
```
magicus: Can you extract this into a tryUsingJumpButtons(), that returns true if successful?

magicus:
> I did, but there didn't seem to be a method to set the ItemStack, only get

You need to add this functionality. In `SlotMixin`, do something like this (untested code):
```
        SetSlotEvent.Pre event = new SetSlotEvent.Pre(container, slot, itemStack);
        MixinHelper.post(event);

        ItemStack oldItemStack = container.getItem(slot);
        original.call(container, slot, event.getItemStack());
```

Add a setItemStack() method to the Pre event. (You will need to remove the final designation on the itemStack field in the superclass)

Then you can listen to SetSlotEvent.Pre, and replace the itemstack when it is set.

ShadowCat117:
> > I did, but there didn't seem to be a method to set the ItemStack, only get
> 
> You need to add this functionality. In `SlotMixin`, do something like this (untested code):
> 
> ```
>         SetSlotEvent.Pre event = new SetSlotEvent.Pre(container, slot, itemStack);
>         MixinHelper.post(event);
> 
>         ItemStack oldItemStack = container.getItem(slot);
>         original.call(container, slot, event.getItemStack());
> ```
> 
> Add a setItemStack() method to the Pre event. (You will need to remove the final designation on the itemStack field in the superclass)
> 
> Then you can listen to SetSlotEvent.Pre, and replace the itemstack when it is set.

Not sure I've done this 100% correctly, it is rendering them but it's also rendering them in my inventory in the same slot indexes.

I think it's because `if (!isBankScreen) return;` isn't enough as it's setting the slots for both, but I can't get containerId from the pre event getContainer to check there.

I'll push what I've done so you can see if I've made a mistake.

Mcrtin:
try ContainerSetSlotEvent and compare the container id like this (in ClientPacketListener#handleContainerSetSlot):
```java
if (packet.getContainerId() == -1) {
			if (!(this.minecraft.screen instanceof CreativeModeInventoryScreen)) {
				player.containerMenu.setCarried(itemStack);
			}
		} else if (packet.getContainerId() == -2) {
			player.getInventory().setItem(i, itemStack);
		} else {
			boolean bl = false;
			Screen var7 = this.minecraft.screen;
			if (var7 instanceof CreativeModeInventoryScreen creativeModeInventoryScreen) {
				bl = !creativeModeInventoryScreen.isInventoryOpen();
			}

			if (packet.getContainerId() == 0 && InventoryMenu.isHotbarSlot(i)) {
				if (!itemStack.isEmpty()) {
					ItemStack itemStack2 = player.inventoryMenu.getSlot(i).getItem();
					if (itemStack2.isEmpty() || itemStack2.getCount() < itemStack.getCount()) {
						itemStack.setPopTime(5);
					}
				}

				player.inventoryMenu.setItem(i, packet.getStateId(), itemStack);
			} else if (packet.getContainerId() == player.containerMenu.containerId && (packet.getContainerId() != 0 || !bl)) {
				player.containerMenu.setItem(i, packet.getStateId(), itemStack);
			}
		}
```

ShadowCat117:
> try ContainerSetSlotEvent and compare the container id like this (in ClientPacketListener#handleContainerSetSlot):

Doesn't look like that's going to work either, can match the container ID's but the actual items don't render, added the same setItemStack method that I added to `SetSlotEvent` but still isn't showing the item

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,238 @@
...
+            pageDestination = customJumpDestinations.get(buttonIndex);
+            e.setCanceled(true);
+            jumpToDestination();
+        }
+    }
+
+    @SubscribeEvent
+    public void onSetSlot(SetSlotEvent.Pre e) {
+        if (!isBankScreen) return;
+
```
magicus, line 122: Try something like
```suggestion

if (e.getContainer() instanceof Inventory) return;
```

ShadowCat117:
Should all work now. If you're happy for this to be merged, shall I make another PR for the same thing for the misc bucket or add it into this one?

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onRenderSlot(SlotRenderEvent.Pre e) {
+        if (!(e.getScreen() instanceof AbstractContainerScreen<?> screen)) return;
+        if (!isBankScreen) return;
+
+        if (e.getSlot().index == NEXT_PAGE_SLOT && Models.Container.isLastBankPage(screen)) {
+            lastPage = currentPage;
```
magicus, line 147: This is all that is left of the rendering. I think we can move it to a more suitable location. onSetSlot should work fine, I think..?

magicus:
I found one more small thing, after all your other changes. :-)

But when that is done, I believe we can merge this.

I suggest we merge this, and then you start a new PR based on main with this PR included, that refactors this code so it supports the misc bucket as well. Better than to start mucking about more with this PR now.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: I did try that, but wasn't able to get a reference to the screen to pass in to isLastBankPage. Any idea on how to get it?

magicus: The ContainerModel is a bit wonky designed to work with screens... The best solution would probably be to look for e.g. MenuEvent.MenuOpenedEvent and check getTitle(). This requires rewriting the ContainerModel functionality a bit.

But since this is even more work, I am totally fine with keeping this for now. We can clean this up later as well, if/when we fix the ContainerModel.

ShadowCat117: Ok, I'll add a comment there as a reminder and then it should be good to go and I'll work on the misc bucket

magicus: I did realize though that a possibly better solution than the SetSlotEvent is to use ContainerSetContentEvent.Pre, where you get all the items at once, and can just replace the individual item stacks. You will not get a container name, but you do get a containerId which you can use to verify that it is the correct container. (But for this to work, you will need to save the containerId, perhaps from a MenuOpenedEvent.)

You have done enough of work on this, so I fully understand if you want to pass on doing this. Let me know if that is the case, and I'll approve it as it is. But if you strive for perfection (and a spot in the Wynntils dev team? :-)), please go ahead!

ShadowCat117: Did try using ContainerSetContentEvent.Pre yesterday after Mcrtin suggested it and the items didn't render but I'll give it another try quick

magicus: Have a look at how `ItemHandler.onContainerSetContent` works...

ShadowCat117: I'm definitely close, it renders them once I go to a different page but not the first time the container opens.

ShadowCat117: This is what I've got, I'm not sure why it doesn't work on the first open, I checked the hovername of the items after the loop by calling e.getItems() again and they did have the new names but when it goes to render it's still using the old items.

```
@SubscribeEvent
    public void onContainerSetContent(ContainerSetContentEvent.Pre e) {
        if (e.getContainerId() != containerId) return;

        List<ItemStack> items = e.getItems();

        for (int buttonSlot : BUTTON_SLOTS) {
            ItemStack jumpButton = new ItemStack(Items.DIAMOND_AXE);
            jumpButton.setDamageValue(92);

            CompoundTag jumpTag = jumpButton.getOrCreateTag();
            jumpTag.putInt("HideFlags", 6);
            jumpTag.putBoolean("Unbreakable", true);
            jumpButton.setTag(jumpTag);

            int buttonIndex = BUTTON_SLOTS.indexOf(buttonSlot);
            int buttonDestination = customJumpDestinations.get(buttonIndex);

            jumpButton.setCount(buttonDestination);

            jumpButton.setHoverName(Component.literal(ChatFormatting.GRAY + "Jump to Page " + buttonDestination));

            items.set(buttonSlot, jumpButton);
        }
    }
```

magicus: Hm, I can't say straight up what's wrong. Can you check by debugging if you even end up in the method for the first page? Is the containerId correct, so you actually go on to change the items, even on the first page?

Do you get SetSlotEvent afterwards (that is how it is usually sent; don't know why)? Can you check if you get that for the first page also? If you do, can you see if you have gotten your new itemstack or if the old one is still in place?

ShadowCat117: Did do a bit of debugging, I called e.getItems after the loop and checked the hovernames of items at those indexes and it was updated correctly. Then in SetSlot I think it had the old ItemStack so hadn't updated there, but then it does work for every other page which is strange

magicus: This is really strange. I noted recently that ContainerSetContentEvent.Pre is called twice every time you open a new container, and then only once. Maybe this is related somehow... Otherwise I don't know. In worst case, you'll have to drop this change and go back to the per-slot setting.

ShadowCat117: Yeah I did notice that when the bank first opens it's called once then every other page it's called twice. I wasn't sure if that was something on the servers end or was actually to do with the event. Well if someone figures it out in the future can always come back and retry it but for now guess it'll have to stay in SetSlot

magicus: In the code above, you seem to have a `containerId` field. Is it possible you can use that part of the code, to check if it is the last bank screen, instead of the render function? If we skip the ContainerSetContentEvent part.

ShadowCat117: I was getting that in onScreenInit, and I don't think the items have rendered by that point so it won't be able to check the name of the item. I can double check though.

Also, just as a quick fyi in case you were about to merge this, I've found a potential issue whilst making this for the misc bucket where if you try to jump to your final page which is 1 below one of the hard coded jump pages, it won't jump anywhere. So I'm just adding a bit more to tryUsingJumpButtons

ShadowCat117: Actually I'm not sure if the method I was thinking of will work. 

Aside from a render event constantly checking the title of the container to see if it's where we're supposed to be, is there a way to check if after we've tried to click a slot that it has actually moved page?

Also, calling isLastBankPage inside onScreenInit didn't work,

ShadowCat117: I've fixed this by adding the last pages to storage and in `tryUsingJumpButtons` when it calculates the closest jump, it won't allow it if the jump is greater than the last page.

I've added this method to `ContainerModel` which also eliminates the need for checking for last page in the render event, as this is called in `onScreenInit`.

```
public int updateFinalPage(Screen screen) {
        int currentPage = getCurrentBankPage(screen);

        if (isBankScreen(screen)) {
            if (currentPage > finalBankPage.get()) {
                finalBankPage.store(currentPage);
            }

            return finalBankPage.get();
        } else if (isBookshelfScreen(screen)) {
            if (currentPage > finalBookshelfPage.get()) {
                finalBookshelfPage.store(currentPage);
            }

            return finalBookshelfPage.get();
        } else if (isMiscBucketScreen(screen)) {
            if (currentPage > finalMiscBucketPage.get()) {
                finalMiscBucketPage.store(currentPage);
            }

            return finalMiscBucketPage.get();
        }

        return currentPage;
    }
```

The only problem with this, is that because the default values for final pages is 1, it won't let you quick jump anywhere until you start visiting more pages. I could fix this by having the default values be the maximum number of pages, but then I would need to call `isLastBankPage` again and that would still have to be in the render event.

magicus: Ok, I decided to go to the bottom of this. `isLastBankPage` is just a bit stupidly implemented for our case, since it tries to do too much.

The core thing it does is
```
StyledText.fromComponent(cs.getMenu().getSlot(8).getItem().getHoverName())
                        .endsWith(LAST_BANK_PAGE_STRING);
```
and this you could test from the full container update event.

So my suggestion is that you basically extract this part into a separate method in ContainerModel, like `isItemIndicatingLastBankPage`, and move the 8 to a public constant `LAST_BANK_PAGE_SLOT`. Then call this new function from the old isLastBankPage(), and also use it in your new code.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,237 @@
...
+                if (lastPage != currentPage) {
+                    clickNextPage();
+                }
+            }
+            case -1 -> clickPreviousPage();
+            default -> {
+                if (!tryUsingJumpButtons()) {
+                    if (pageDestination < currentPage && lastPage != currentPage) {
+                        clickPreviousPage();
+                    } else if (lastPage != currentPage) {
```
magicus: Could you please invert these checks, so you test "is currentPage ...?"

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: Should all be done now, I'll just make the changes on my misc bucket branch and check the jumping to last page works there and this should be good to go

ShadowCat117: Yup, seems to work there so once you visit your final page once it'll all work fine, hopefully most people will figure that out and we won't get too many complaints of people thinking the feature is broken

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -0,0 +1,237 @@
...
+    private void jumpToDestination() {
+        quickJumping = true;
+
+        if (currentPage == pageDestination) return;
+
+        int pageDifference = pageDestination - currentPage;
+
+        switch (pageDifference) {
+            case 1 -> {
+                if (lastPage != currentPage) {
```
magicus:
```suggestion
                if (currentPage != lastPage) {
```

magicus (approved):
Finally!!!!!!1!11!!! :-)

Thank you so much for hanging in there! It has been an extraordinary amount of back and forth for what seemed to be a simple feature. But I'm very happy you stayed with us, because this implementation is so much better than the first iteration.

ShadowCat117:
No worries! Learnt some more about the project and how to do certain things so future features can hopefully be easier :)

PR #1671, opened 2023-07-02
fix: Add fallback handler to scoreboard handler

magicus:
Otherwise new, unknown segments are just lost


magicus:
I updated to the PR so the temporary lootrun scoreboard part is removed. It was just a "specialized" placeholder, and the fallback placeholder will do the same job.

I also think we need to reconsider how we should integrate the new "lootrun" support in Wynntils. Up until now, we have used "lootrun" to mean specifically our own system of recording paths in the world. Now, there is a completely new and unrelated game mechanic known as "lootrun". We should be careful not to confuse the two. I think the best way is to rename our lootrun stuff to e.g. "lootrunpath" or maybe just "pathrecording" or "runmapping" or whatever. Even though we have had "lootrun" a long time, I think we should give priority to Wynncraft official names.

kristofbolyai (commented):
This was so easy to add...

Please remove every other dummy implementation, there are others other than LootrunScoreboardPart. After that, looks good.

magicus:
I only found one unused dummy implemementation, party finder, but I removed it.

kristofbolyai (approved):

Issue #1495, opened 2023-04-15
Quest tracking from vanilla quest book results in error after disabling Wynntils quest book and restarting


imIrony:
Attempting to track a quest from the vanilla quest book when the Wynntils quest book is disabled results in an error on a newly started instance. The only way around this is to enable and disable the Wynntils quest book again then quest tracking starts working with the vanilla quest book once again.

![2023-04-15_09 37 36](https://user-images.githubusercontent.com/12867627/232227420-13acc7c2-8f70-4b45-b139-bc24da1c4bb4.png)

kristofbolyai:
This is a rather complex issue, seems to be happening after @magicus patch for a much faster quest book. 

Multiple things can fix this:
1, Force quest book analysis on class join (not worth it imo)
2, Don't allow opening the regular quest book when the feature itself is enabled, so we can skip the work

magicus:
Regarding 1, I have been thinking that we should persist quest information per character. Then we could assume that our current persisted state is correct, and just update/verify it when doing quest stuff.

This could also allow us to re-track a tracked quest when logging in with a character, something that always has annoyed me, that tracked quests are forgotten when switching servers or relogging.

magicus:
I just wanted to get full player UUID + storage in models to work properly first, before implementing this.

kristofbolyai:
> I just wanted to get full player UUID + storage in models to work properly first, before implementing this.

Both work now, so it is time to do this :)

magicus:
Unless you've done things I've missed, both are not fully complete as to how I want to see it.

kristofbolyai:
> Unless you've done things I've missed, both are not fully complete as to how I want to see it.

How so?

You added storage to models (only configs don't work) and we parse the uuid from the compass?

magicus:
Right, storage to models work, not just configs. I was mixing this up. 

But, the UUID is still just partial. I want to match it to the full UUID from the Wynncraft API, before we start storing any substantial data for it.

I am also a bit unhappy that it is still guesswork involved in matching UUID and characters in the selection screen. :( But at least I'd like to implement that guesswork, so we can track the UUID of characters even before they are selected.

ShadowCat117:
I'm going to assume this is no longer an issue after the reworks of both the vanilla and Wynntils content book

PR #1672, opened 2023-07-02
feat: New Tracker Model, separated from QuestModel

magicus:
Let the QuestModel deal with the QuestBook only, and create a new `Tracker` model to deal with tracking various content tasks. Will hopefully help with https://github.com/Wynntils/Artemis/issues/1495.


common/src/main/java/com/wynntils/core/components/Models.java
```diff
@@ -81,6 +82,7 @@ public final class Models {
     public static final TerritoryModel Territory = new TerritoryModel();
     public static final TokenModel Token = new TokenModel();
     public static final TowerAuraTimerModel TowerAuraTimer = new TowerAuraTimerModel();
+    public static final TrackerModel Tracker = new TrackerModel();
```
kristofbolyai: Perhaps rename to `ContentTrackerModel`? It is not clear without context. Even with that added name, it is still so-so..

common/src/main/java/com/wynntils/features/overlays/TrackerOverlayFeature.java
```diff
@@ -40,35 +38,23 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @ConfigCategory(Category.OVERLAYS)
-public class QuestInfoOverlayFeature extends Feature {
+public class TrackerOverlayFeature extends Feature {
```
kristofbolyai: If you decide to rename the model, follow on here.

I am also not confident the feature separation is needed, but I can understand it.

Do you want to add a config upfixer so people's config don't get lost?

common/src/main/java/com/wynntils/models/quests/QuestModel.java
```diff
@@ -144,107 +131,36 @@ public void openQuestOnWiki(QuestInfo questInfo) {
...
-
     void setDialogueHistory(List<List<StyledText>> newDialogueHistory) {
         dialogueHistory = newDialogueHistory;
         WynntilsMod.postEvent(new QuestBookReloadedEvent.DialogueHistoryReloaded());
     }
 
     /** Shared between the container query classes */
     public String getQuestBookTitleRegex(int pageNum) {
-        return "^§0\\[Pg. " + pageNum + "\\] §8.*§0 (Mini-)?Quests$";
+        return "§f\uE000\uE072";
```
kristofbolyai, line 163: Weird!

common/src/main/java/com/wynntils/models/tracker/TrackerScoreboardPart.java
```diff
@@ -45,24 +46,31 @@ public void onSegmentChange(ScoreboardSegment newValue) {
             }
         }
 
+        String unformattedHeader = newValue.getHeader().getString(PartStyle.StyleType.NONE);
+        Matcher matcher = TRACKER_MATCHER.headerPattern().matcher(unformattedHeader);
+
+        assert matcher.matches();
```
kristofbolyai: Qodana is right? I am not sure how asserts work in PROD, but if they are completely ignored, then this matches call wouldn't happen, would it?

kristofbolyai (commented):
Overall this is great. I think the assert matcher needs fixing, but after that it can be merged.

common/src/main/java/com/wynntils/features/overlays/TrackerOverlayFeature.java
magicus: I'm a bit more gung-ho about dropping configuration as long as we call ourselves alpha. :-) If it is like a simple one-liner I might add an upfixer, otherwise I'm not sure I could be bothered to do this. :) If you want to contribute one, please go ahead.

common/src/main/java/com/wynntils/models/quests/QuestModel.java
magicus: Yeah, odd. This was not really part of this fix; I started out trying to get the Quest Book working and it just slipped in. But it will be needed anyhow.

common/src/main/java/com/wynntils/core/components/Models.java
magicus: Renamed. I hope I got all places. I have not renamed all internal variables, it got too unwieldy, and in those cases, I think it is clear from the context.

common/src/main/java/com/wynntils/models/tracker/TrackerScoreboardPart.java
magicus: I can't see Qodana's comment..? The assert will be ignored unless you run with asserts enabled. We do that in our IntelliJ profile but not in production. We know that the matcher is going to match since we only get a callback here if it matched in the caller. So this assert was more to show that I know this. Otherwise it might look suspicious to just check the group.

magicus: Now I saw it. I tried to read the docs for matches() and group(), and it seems that they say that I must call matches() before group(), so I guess this would have broken in prod.

common/src/main/java/com/wynntils/features/overlays/TrackerOverlayFeature.java
magicus: I added a generic upfixer for plain renamed configs, hoping it will be easier to add similar changes in the future.

kristofbolyai (approved):

PR #1675, opened 2023-07-03
refactor: Make party leader in PartyModel safer to use

magicus:
This is basically a fix of what was discussed in https://github.com/Wynntils/Artemis/pull/1586#discussion_r1220853312.

### DIFF
--- a/common/src/main/java/com/wynntils/utils/mc/McUtils.java
+++ b/common/src/main/java/com/wynntils/utils/mc/McUtils.java
@@ -31,6 +31,10 @@ public final class McUtils {
         return mc().player;
     }
 
+    public static String playerName() {
+        return player().getName().getString();
+    }
+
     public static Options options() {
         return mc().options;
     }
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/utils/mc/McUtils.java
```diff
@@ -31,6 +31,10 @@ public static LocalPlayer player() {
         return mc().player;
     }
 
+    public static String playerName() {
+        return player().getName().getString();
```
kristofbolyai, line 35: I know it worked before but should Component#getString be used? Maybe StyledText#getStringWithoutFormatting is more stable

kristofbolyai (approved):

common/src/main/java/com/wynntils/utils/mc/McUtils.java
magicus: I don't know... This is a pure refactoring, just shuffling code around. Can we keep it doing that, then we can experiment with player name formatting at some other time?

kristofbolyai: Yes

PR #1680, opened 2023-07-04
fix: Fix auto update not working on any stream

kristofbolyai:


common/src/main/java/com/wynntils/core/net/athena/UpdateManager.java
```diff
@@ -37,9 +37,10 @@ public UpdateManager(NetManager netManager) {
     public CompletableFuture<String> getLatestBuild() {
         CompletableFuture<String> future = new CompletableFuture<>();
 
-        WynntilsMod.info("Checking for update for stream " + getStream() + ".");
+        String stream = getStream();
+        WynntilsMod.info("Checking for update for stream " + stream + ".");
 
-        ApiResponse apiResponse = Managers.Net.callApi(UrlId.API_ATHENA_UPDATE_CHECK, Map.of("stream", getStream()));
+        ApiResponse apiResponse = Managers.Net.callApi(UrlId.API_ATHENA_UPDATE_CHECK, Map.of("stream", stream));
```
kristofbolyai, line 43: No wonder why this was so broken... We forgot to add the stream in every location...

PR #1685, opened 2023-07-05
feat: Add the disconnect and change world buttons to the character selection screen

sofiedotcafe:
Add the disconnect and change world buttons to the character selector screen 
Screenshot_20230630_112727


magicus:
I think the button icons are hard to understand. Maybe use text instead, `LEAVE` and `LOBBY`, akin to the `PLAY` button?

kristofbolyai:
> I think the button icons are hard to understand. Maybe use text instead, `LEAVE` and `LOBBY`, akin to the `PLAY` button?

I think if there is a hover tooltip then it is okay.

magicus:
Maaaaybe. But they do not at all feel very wynncraft-y nor wynntils-y. I agree the functionality is reasonable, but they need to fit better in, design-wise.

common/src/main/java/com/wynntils/screens/characterselector/widgets/DisconnectButton.java
```diff
@@ -0,0 +1,75 @@
...
+    private final CharacterSelectorScreen characterSelectorScreen;
+
+    public DisconnectButton(int x, int y, int width, int height, CharacterSelectorScreen characterSelectorScreen) {
+        super(x, y, width, height, Component.literal("Disconnect Button"));
+        this.characterSelectorScreen = characterSelectorScreen;
+    }
+
+    @Override
+    public void onPress() {
+        TitleScreen titleScreen = new TitleScreen();
```
magicus, line 37: This looks like we are doing a lot of vanilla housekeeping here. It sounds scary, like it could break at any upgrade. There is no one single method that we can call? What does the vanilla disconnect button do? Can't we call that method instead?

magicus (commented):
I am still not happy about the artwork, but we can fix that later.

My one remaining concern is about the disconnect methodology. I'd rather see we call a single vanilla method, even if it means having to expose it with an accesswidener.

Mcrtin:
this is how Minecraft does it in PauseScreen:
```java
private void onDisconnect() {
		boolean bl = this.minecraft.isLocalServer();
		boolean bl2 = this.minecraft.isConnectedToRealms();
		this.minecraft.level.disconnect();
		if (bl) {
			this.minecraft.clearLevel(new GenericDirtMessageScreen(SAVING_LEVEL));
		} else {
			this.minecraft.clearLevel();
		}

		TitleScreen titleScreen = new TitleScreen();
		if (bl) {
			this.minecraft.setScreen(titleScreen);
		} else if (bl2) {
			this.minecraft.setScreen(new RealmsMainScreen(titleScreen));
		} else {
			this.minecraft.setScreen(new JoinMultiplayerScreen(titleScreen));
		}
	}
```

magicus:
Hm, okay. Normally, I'd had preferred to call vanillas method, but if it is an instance method of a screen, we'll skip it.

magicus:
Can you document that the logic is based on `PauseScreen.onDisconnect` as well? Then I think this is ok.

sofiedotcafe:
Hey 👋!
Sorry if I haven't yet addressed the "requests" on this issue yet but I do them when I'm home from a vacation from London 🇬🇧
But thanks for the tips nevertheless!

Kind regards:
- jh-devv

common/src/main/java/com/wynntils/screens/characterselector/widgets/DisconnectButton.java
magicus, line 37:
```suggestion
        // This is replicating the behavior found in PauseScreen.onDisconnect()
        TitleScreen titleScreen = new TitleScreen();
```

magicus (approved):
I fixed the comment for you. Hope you don't mind.

PR #1687, opened 2023-07-05
fix: Do not register clicks outside displayed config button

Mcrtin:
this is a complete rework of the ConfigButton -now called ConfigTile. This aims to fix the missing bounding box of the ConfigOptionElements.


magicus:
What is the rationale for this change?

Mcrtin:
the config button (for example the enabled button) spans the whole config tile, only leaving room for the reset button. it should instead only register clicks when the actual button is clicked.

magicus:
Ok, I see. I suggest you rename this PR along those lines: `fix: Do not register clicks outside config button text`.

I'll review this after I'm finished with beta stuff.

magicus:
I started reviewing this. From the look of it, this will change how the config screen is rendered, not just how clicks are registered? The old code had a lot of calls to e.g. `RenderUtils.drawRoundedRectWithBorder`. I can't see that this is present in the new code. 

Does this mean the look of the config screen has changed? If so, can you post a screenshot? If not, can you explain how you keep the same look -- what am I missing?

Mcrtin:
there was a duplicate code. The `GeneralSettingsButton` does all the rendering; the same way the element classes did, but the settings screen looks a bit different because I changed the width and height a bit.
Before:
image
image
After:
image
image

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
```diff
@@ -0,0 +1,47 @@
...
+        poseStack.pushPose();
+        poseStack.translate(0f, (height - FontRenderer.getInstance().getFont().lineHeight + 8) / 2f - 5, 0f);
+        super.renderWidget(poseStack, mouseX, mouseY, partialTick);
+        poseStack.popPose();
+    }
+
+    protected void onUpdate(String text) {
+        Object parsedValue = configHolder.tryParseStringValue(text);
+        if (parsedValue != null) {
+            if (!Objects.equals(parsedValue, configHolder.getValue())) configHolder.setValue(parsedValue);
```
magicus: Please always use braces in if statements, except for trivial return/break/etc.

common/src/main/java/com/wynntils/screens/settings/widgets/ResetButton.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    protected CustomColor getBackgroundColor() {
+        return configHolder.valueChanged() ? super.getBackgroundColor() : BACKGROUND_COLOR;
+    }
+
+    @Override
+    public void playDownSound(SoundManager handler) {
+        if (configHolder.valueChanged()) super.playDownSound(handler);
```
magicus: Same here; use braces.

magicus (changes requested):
I think this looks okay, apart from the "if" changes. But I will need bolyai to look at it as well, this is his code originally.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
Mcrtin: he said it'd be ok if they're on one line, referring to this blog post: https://www.codecentric.de/wissens-hub/blog/curly-braces

kristofbolyai: No, this is not okay. I might not have been clear, "if it fits the line", but this borderline doesn't. So you just add the braces.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float partialTick) {
+        poseStack.pushPose();
+        poseStack.translate(0f, (height - FontRenderer.getInstance().getFont().lineHeight + 8) / 2f - 5, 0f);
+        super.renderWidget(poseStack, mouseX, mouseY, partialTick);
+        poseStack.popPose();
+    }
+
+    protected void onUpdate(String text) {
```
kristofbolyai, line 27: You don't have to invent this whole abstract/overriden method if you added this method as a consumer in the constuctor..

common/src/main/java/com/wynntils/screens/settings/widgets/ResetButton.java
kristofbolyai: Use it.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
Mcrtin: how can i do that? i tried `this::onUpdate` but I "Cannot reference 'this' before supertype constructor has been called"

common/src/main/java/com/wynntils/screens/settings/widgets/EnumSettingsButton.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public void onPress() {
+        // we use instead AbstractWidget#mouseClicked because we also want to have an action on the right mouse button
+    }
+
+    private static <E extends Enum<E>> int getWidth(Type type) {
+        Font font = FontRenderer.getInstance().getFont();
+        return EnumSet.allOf((Class<E>) type).stream()
```
magicus: Is this how Spotless format this statement? 😱 

I think you need to break it down. Can you extract the stream operation into a local variable `maxWidth` or such, and then return it + 8.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
magicus: What if you were to just pass it as a lambda, would that have worked? I'm not sure it's better, but maybe?

Or you could make this a private static helper function, and pass it the configHolder as well as argument, and call this as a lambda from the constructor. I'm pretty sure that is allowed.

(I agree with bolyai that it is better if we don't need this override when there is already a mechanism for setting a callback)

Mcrtin: I tried everything before i resorted to making an abstract method but to invoke `setRenderColor` I have to have the reference to the object.

magicus: I see. Ok, your explanation sounds reasonable to me. I'll leave it to bolyai to resolve this conversation if he's happy with the reply.

kristofbolyai (commented):
You might have realized this, but to be honest, I don't like this PR. We basically have an almost 500 line PR which touches an essential system, the config system GUI, a completely "rewritten" widget just to fix a small "bug". 

This is kind of a "if it is don't broken, don't fix it" approach from me, and I am also slightly paranoid it will accidentally break something.

Nevertheless, I will take a close look at this tomorrow, and try to integrate this (after the necessary reviews).

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
kristofbolyai: Okay, I see. Let's keep it.

kristofbolyai: You can resolve this after you add the `@Override` annotation for it.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
```diff
@@ -0,0 +1,39 @@
...
+package com.wynntils.screens.settings.widgets;
+
+import com.wynntils.core.config.ConfigHolder;
+import com.wynntils.screens.base.TextboxScreen;
+import com.wynntils.screens.base.widgets.TextInputBoxWidget;
+import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.render.FontRenderer;
+
+public class TextInputBoxSettingsWidget extends TextInputBoxWidget {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/screens/settings/widgets/ResetButton.java
```diff
@@ -0,0 +1,52 @@
...
+package com.wynntils.screens.settings.widgets;
+
+import com.wynntils.core.config.ConfigHolder;
+import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.colors.CustomColor;
+import java.util.List;
+import net.minecraft.client.sounds.SoundManager;
+import net.minecraft.network.chat.Component;
+
+final class ResetButton extends GeneralSettingsButton {
```
kristofbolyai:
```suggestion
private final class ResetButton extends GeneralSettingsButton {
```

common/src/main/java/com/wynntils/screens/settings/widgets/ResetButton.java
```diff
@@ -0,0 +1,52 @@
...
+                12,
+                Component.translatable("screens.wynntils.settingsScreen.reset.name"),
+                List.of(Component.translatable("screens.wynntils.settingsScreen.reset.description")));
+        this.configHolder = configHolder;
+        this.onClick = onClick;
+    }
+
+    @Override
+    protected CustomColor getTextColor() {
+        return !configHolder.valueChanged() ? CommonColors.GRAY : CommonColors.WHITE;
```
kristofbolyai: Can you flip this (to remove `!`)?

common/src/main/java/com/wynntils/screens/settings/widgets/EnumSettingsButton.java
```diff
@@ -0,0 +1,74 @@
...
+import com.wynntils.utils.render.FontRenderer;
+import java.lang.reflect.Type;
+import java.util.EnumSet;
+import java.util.List;
+import net.minecraft.client.gui.Font;
+import net.minecraft.network.chat.Component;
+import org.lwjgl.glfw.GLFW;
+
+public class EnumSettingsButton<E extends Enum<E>> extends GeneralSettingsButton {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/screens/settings/widgets/CloseButton.java
```diff
@@ -0,0 +1,33 @@
...
+package com.wynntils.screens.settings.widgets;
+
+import com.wynntils.screens.settings.WynntilsBookSettingsScreen;
+import com.wynntils.utils.render.Texture;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+
+public class CloseButton extends GeneralSettingsButton {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/screens/settings/widgets/BooleanSettingsButton.java
```diff
@@ -0,0 +1,50 @@
...
+import com.wynntils.core.config.ConfigHolder;
+import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.colors.CustomColor;
+import com.wynntils.utils.render.FontRenderer;
+import java.util.List;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.MutableComponent;
+
+public class BooleanSettingsButton extends GeneralSettingsButton {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/screens/settings/widgets/ApplyButton.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.wynntils.core.components.Managers;
+import com.wynntils.screens.settings.WynntilsBookSettingsScreen;
+import com.wynntils.utils.render.Texture;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+
+public class ApplyButton extends GeneralSettingsButton {
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -583,4 +583,9 @@ private void replaceHighlighted(String text) {
         this.setHighlightPosition(this.cursorPosition);
         this.onUpdateConsumer.accept(this.textBoxInput);
     }
+
+    /**
+     * if there is no on update consumer given in the constructor this method gets called instead.
```
kristofbolyai:
```suggestion
     * If there is no on update consumer given in the constructor, this method gets called instead.
```

Mcrtin:
The Tooltip of the categories is rendered under the config buttons:
image
how can I fix that? Is that an issue with the tooltips? It works with the tooltips of the config buttons:
image

kristofbolyai:
It is an unrelated issue to this PR, you didn't cause it. I will check if it requires a lot of stuff to be rewritten..

magicus (approved):

kristofbolyai (approved):
Okay, this looks sane. Thank you for holding on, this took a while.

PR #1688, opened 2023-07-05
feat: render lootrun path on map

HalcyonSuoh:


magicus:
Ooooh, interesting! O_o 

Can you post a screenshot of how this looks?

HalcyonSuoh:
> Ooooh, interesting! O_o
> 
> Can you post a screenshot of how this looks?

surely
image
image

common/src/main/java/com/wynntils/models/lootruns/LootrunCompiler.java
```diff
@@ -201,6 +209,80 @@ private static Long2ObjectMap<List<ColoredPath>> generatePointsByChunk(LootrunPa
...
+        List<Vector2d> simplified = new ArrayList<>();
+
+        // split list with the farthest point and simplify both sublist recursively if distance is greater than epsilon
+        if (dist > epsilon) {
+            List<Vector2d> left = simplify(points.subList(0, index + 1), epsilon);
+            List<Vector2d> right = simplify(points.subList(index, end + 1), epsilon);
+            simplified.addAll(left.subList(0, left.size() - 1));
+            simplified.addAll(right);
+
+            // return both end of the simplified line otherwise
```
kristofbolyai: this should be in the else block I assume

common/src/main/java/com/wynntils/models/lootruns/LootrunCompiler.java
```diff
@@ -201,6 +209,80 @@ private static Long2ObjectMap<List<ColoredPath>> generatePointsByChunk(LootrunPa
...
+        }
+
+        List<Vector2d> intermediate = simplify(points, tolerance);
+        List<Vector3f> result = new ArrayList<>();
+
+        for (Vector2d vec : intermediate) {
+            result.add(new Vector3f((float) vec.x, (float) 0, (float) vec.y));
+        }
+
+        return result;
```
kristofbolyai: Is this remapping necessary? Why not just store simplified points as Vector2f?

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -146,6 +150,265 @@ public static void renderCursor(
...
+
+        List<Vector3f> points = new ArrayList<>();
+
+        List<Vector3f> middlePoints = new ArrayList<>();
+
+        for (Vector3f point : lootrun.simplifiedPath()) {
+            points.add(new Vector3f(
+                    getRenderX((int) point.x(), mapTextureX, centerX, currentZoom),
+                    getRenderZ((int) point.z(), mapTextureZ, centerZ, currentZoom),
+                    0));
```
kristofbolyai: Shouldn't it be x, 0, z? (Again, does this strictly need to be a Vector3f if we have a filler element?)

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -146,6 +150,265 @@ public static void renderCursor(
...
+        }
+        Vector3f last = null;
```
kristofbolyai:
```suggestion

        Vector3f last = null;
```

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -146,6 +150,265 @@ public static void renderCursor(
...
+    private static void addVertex(BufferBuilder bufferBuilder, Vector3f pos, int color, PoseStack poseStack) {
+        bufferBuilder
+                .vertex(poseStack.last().pose(), pos.x(), pos.y(), pos.z())
+                .color(color)
+                .endVertex();
+    }
+
+    private static float signedArea(Vector3f p0, Vector3f p1, Vector3f p2) {
+        return (p1.x() - p0.x()) * (p2.y() - p0.y()) - (p2.x() - p0.x()) * (p1.y() - p0.y());
+    }
```
kristofbolyai: Move to MathUtils perhaps

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -146,6 +150,265 @@ public static void renderCursor(
...
+            return null;
+        } else {
+            var c0 = a0 * p0.x + b0 * p0.y;
+            var c1 = a1 * p2.x + b1 * p2.y;
+
+            var x = (b1 * c0 - b0 * c1) / det;
+            var y = (a0 * c1 - a1 * c0) / det;
+            return new Vector3f(x, y, 0);
+        }
+    }
```
kristofbolyai: Maybe this is worth creating a VectorUtils class

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
kristofbolyai: I am more worried about this allocation since this is render time, than the lootrun compiling one.

kristofbolyai (commented):
Looks good basically, once we discuss my comments.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
HalcyonSuoh: > Shouldn't it be x, 0, z?

I honestly don't know, getRenderZ in itself doesn't sounds very intuitive, as it's renderer in screenspace, and z is the forward vector there

magicus: I agree that this is doing scary much during rendering. All this processing, can't it be done outside rendering? Ideally, we should know exactly what we need to render when it is time to render, and just do that as fast as possible with as few allocations as possible.

HalcyonSuoh: as long as it still need screenspace transformation it will need to be in render thread, an object pool maybe?

kristofbolyai: > I agree that this is doing scary much during rendering. All this processing, can't it be done outside rendering? Ideally, we should know exactly what we need to render when it is time to render, and just do that as fast as possible with as few allocations as possible.

Perhaps I could profile this. I know it looks like a lot of code, but it is actually rather fast compared to what legacy does for example... Adding more vertexes (or rather not filtering them) is a LOT more slower than this whole process. I will come back here soon with some results.

Kepler-17c: Legacy didn't optimise at all and just always rendered the whole path, so that's like a worst case or hard limit comparison

kristofbolyai: Yes, I know. But having actual numbers will either make me and magicus relax, or we can find the problem. Win-Win situation.

Kepler-17c: True - it wasn't my intention to argue against profiling (rather the opposite), just to point out that Legacy isn't a good comparison

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -4,6 +4,9 @@
  */
 package com.wynntils.utils.render;
 
+import static com.wynntils.utils.MathUtils.signedArea;
+import static com.wynntils.utils.VectorUtils.lineIntersection;
```
kristofbolyai: I just noticed. I think we agreed on not using static imports in Artemis (like we don't use import *), but this was not blocked by spotless. Can you refactor?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -230,6 +233,31 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
...
+            MapRenderer.renderLootrunLine(
+                    currentLootrun,
+                    2f,
+                    poseStack,
+                    centerX,
+                    centerZ,
+                    mapCenterX,
+                    mapCenterZ,
+                    currentZoom,
+                    CommonColors.LIGHT_BLUE.asInt());
```
kristofbolyai: I would at least like to merge these calls, so we don't do most of the work twice. After that, I am most likely fine with performance.

If you don't have time, I can patch it up and merge it. Otherwise, thanks.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
kristofbolyai: Okay. Performance numbers. 97% of the time we spend is actual rendering, around 3% is the math. So I think this concludes the need for optimization.

EDIT: I can't math... the corrected number is: around 90% of time spent rendering. That leaves almost no room to optimize, considering around 9% of the processing time is spent on `Math#atan2`.

EDIT 2: Allocation is a bit more wild... I suggested a "fix" for this, https://github.com/Wynntils/Artemis/pull/1688/files#r1257100736. After that, I think this should be alright.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
HalcyonSuoh: intellij import it by default, but sure

HalcyonSuoh:
the continuity it's referring to
image

kristofbolyai:
> the continuity it's referring to image

Looks even better!

kristofbolyai (approved):
This looks great now! 

I know that the code could be simplified a little bit, but I think I (or anyone else) will go ahead and do that later when this is also rendered on the mini-map. While it is only on the main map, the bit higher allocation is acceptable.

kristofbolyai:
As with other features, please wait for 2.0.3 before merging. Thanks.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
magicus: @Kepler-17c I agree that legacy is not a good comparison. But our bar for Artemis is higher than replicating legacy.

@kristofbolyai Thanks for the measuring. I'm not sure I understand what fix you are pointing to, though...

common/src/main/java/com/wynntils/models/lootruns/LootrunCompiler.java
magicus: Why was this marked as resolved? I have the same question.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
kristofbolyai: @magicus https://github.com/Wynntils/Artemis/pull/1688#discussion_r1257100736

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -146,6 +152,250 @@ public static void renderCursor(
...
+        for (Vector3f point : lootrun.simplifiedPath()) {
+            points.add(new Vector3f(
+                    getRenderX((int) point.x(), mapTextureX, centerX, currentZoom),
+                    getRenderZ((int) point.z(), mapTextureZ, centerZ, currentZoom),
+                    0));
+        }
+
+        Vector3f last = null;
+        List<Vector3f> filteredPoints = new ArrayList<>();
+        for (Vector3f point : points) {
```
magicus: I'm trying to make sense of this. First you setup a list `points`, then you discard some of them? Can't you combine this, so you only add the points you really need to the `points` list in the first place?

common/src/main/java/com/wynntils/models/lootruns/LootrunCompiler.java
HalcyonSuoh: >I know that the code could be simplified a little bit, but I think I (or anyone else) will go ahead and do that later when this is also rendered on the mini-map. While it is only on the main map, the bit higher allocation is acceptable.

the reason it was marked as resolved, but I could do it myself I suppose

kristofbolyai (approved):
Looks better now. Do you think this is fine, @magicus?

magicus:
I would like to take a look, but I can't spend the time rn. Gimme a day or two.

magicus (approved):
Ship it!

PR #1689, opened 2023-07-05
fix: Container Queries did not get properly annotated with WynnItems

magicus:
This is a necessary prereq for parsing the content book.


kristofbolyai (approved):
Seems reasonable

PR #1694, opened 2023-07-07
feat: Refactor WynncraftButtonFeature connectTo* configs to an enum

kristofbolyai:


common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -232,4 +223,17 @@ private synchronized void loadServerIcon(ResourceLocation destination) {
             }
         }
     }
+
+    private enum ServerType {
+        LOBBY("lobby"),
+        GAME("game"),
```
magicus: I'm pretty sure the official address is `play`. I think we should use that.

magicus (changes requested):

magicus:
Also, my understanding is that `chore:` means "no real changes in the product, but fixing build system, upgrading dependencies, etc". I set it to `feat:` (as in "user visible changes that improves behavior/add new functionality"); it could arguably also be `fix:` (as in "fixes a bug, where behavior did not conform to expectations").

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai:
```suggestion
        GAME("play"),
```

magicus (approved):

PR #1696, opened 2023-07-07
fix: Proper fix for duplicated container content events

magicus:
Proper fix for duplicated container content events when using the container query.

This is a prerequisite for dynamic queries.

### DIFF
--- a/common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
+++ b/common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
@@ -60,6 +60,22 @@ public final class InventoryUtils {
         return accessories;
     }
 
+    public static boolean isItemListsEqual(List<ItemStack> firstItems, List<ItemStack> secondItems) {
+        if (firstItems.size() != secondItems.size()) return false;
+
+        for (int i = 0; i < firstItems.size(); i++) {
+            var newItem = firstItems.get(i);
+            var oldItem = secondItems.get(i);
+            if (!newItem.getItem().equals(oldItem.getItem())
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
```diff
@@ -60,6 +60,22 @@ public static List<ItemStack> getAccessories(Player player) {
...
+    public static boolean isItemListsEqual(List<ItemStack> firstItems, List<ItemStack> secondItems) {
+        if (firstItems.size() != secondItems.size()) return false;
+
+        for (int i = 0; i < firstItems.size(); i++) {
+            var newItem = firstItems.get(i);
+            var oldItem = secondItems.get(i);
+            if (!newItem.getItem().equals(oldItem.getItem())
+                    || newItem.getDamageValue() != oldItem.getDamageValue()
+                    || newItem.getCount() != oldItem.getCount()
+                    || !ItemStack.tagMatches(oldItem, newItem)) {
```
kristofbolyai, line 72: Doesn't ItemStack#equals do this which means firstItems.equals(secondItems) does this?

kristofbolyai: ItemStack doesn't implement equals... Okay.

kristofbolyai (approved):
Looks good!

common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
magicus: Yep, I noticed that. Hence this method. I hope that comparing item type, damage, count and lore that I capture everything that is relevant.

PR #1697, opened 2023-07-07
feat: Redesign Container Queries to allow dynamic queries, and support the new Content Book

magicus:
I have some kind of proof of concept of populating our quest book on beta, but with lots of limitations. The content book must have the correct filter selected. Clicking on a quest to track/untrack is not supported. Additional requirements (profession) is not supported.


kristofbolyai:
I don't fully get the concept of this PR. I thought we are building the dynamic query system first, but this PR is closer to shoehorning the new parsing into the old system. I would be glad to help out this PR, but first we need to have a chat what the next step is.

magicus:
Your reading is absolutely correct. I started with this part to be sure that the dynamic rewrite will work, once I start writing it, since it is tricky to debug unless you get a proper result out of the query to look at.

This is not intended to be merged, it is just a show-case that this part of the parsing works.

common/src/main/java/com/wynntils/handlers/container/scriptedquery/QueryBuilder.java
```diff
@@ -0,0 +1,63 @@
...
+ * 1) a startAction (to open the container)
+ * 2) a verification (to check the container that opened, if any)
+ * 3) a handleContent (to actually consume the content of the container)
+ *
+ * A QueryStep with no verification will assume no new container is opened.
+ * A QueryStep with no handleContent will just perform a no-op for consuming the content.
+ */
+public final class QueryBuilder {
+    private static final Consumer<String> DEFAULT_ERROR_HANDLER =
+            (errorMsg) -> WynntilsMod.warn("Error in ScriptedContainerQuery");
```
kristofbolyai:
```suggestion
            (errorMsg) -> WynntilsMod.warn("Error in ScriptedContainerQuery: " + errorMsg);
```

Why hide it?

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
```diff
@@ -48,59 +50,68 @@ public void updateParsedAbilityTree() {
...
+                        QueryStep.clickOnSlot(NEXT_PAGE_SLOT).processIncomingContainer(processor::processPage))
+
+                // Go back to initial page
+                .repeat(
+                        c -> {
+                            // Go back as many pages as the original count was from the end
+                            this.pageCount++;
+                            return this.pageCount != Models.AbilityTree.ABILITY_TREE_PAGES;
+                        },
+                        QueryStep.clickOnSlot(PREVIOUS_PAGE_SLOT))
```
kristofbolyai, line 94: This "resetting" is new. Is this needed / does this slow queries down? I assume this is more important in the quest book filters...

kristofbolyai (commented):
Just a quick status update. I started reviewing this, got through most of the system code. I started reviewing the new queries, but I am going to have to come back tomorrow to finish. 

Overall it is looking really great so far, can't wait to work with this :)

kristofbolyai:
> Your reading is absolutely correct. I started with this part to be sure that the dynamic rewrite will work, once I start writing it, since it is tricky to debug unless you get a proper result out of the query to look at.
> 
> This is not intended to be merged, it is just a show-case that this part of the parsing works.

I would assume this comment to be out of date by now.

common/src/main/java/com/wynntils/handlers/container/scriptedquery/QueryBuilder.java
magicus: No reason. I was just stupid.

common/src/main/java/com/wynntils/models/abilitytree/AbilityTreeContainerQueries.java
magicus: Yes, it is new. I wouldn't say it slows down things much. In the worst case, you had the tree on the first page, then you need to go back 6 pages. My rough estimate is that 20 pages (the entire content book) takes 3-4 seconds, so at most an additional second.

I think we have to include your ability tree stuff for real to know if this becomes a problem. But I think it is proper of us to restore everything to the way it was, so our inspections are "hidden" for the user. So unless it is a serious problem I'd rather keep it. My thinking is that the reason this was not implemented before was that it was virtually impossible to do with the old design.

magicus:
> > This is not intended to be merged, it is just a show-case that this part of the parsing works.
> 
> I would assume this comment to be out of date by now.

Yes, the branch meandered from a testing PoC to a fully fledged implementation, so I kept the same PR.

common/src/main/java/com/wynntils/models/content/ContentBookQueries.java
```diff
@@ -0,0 +1,206 @@
...
+                // Finally signal we're done
+                .execute(() -> processResult.accept(newContent, progress))
+                .build();
+
+        query.executeQuery();
+    }
+
+    private String getActiveFilter(ItemStack itemStack) {
+        StyledText itemName = InventoryUtils.getItemName(itemStack);
+        if (!itemName.equals(StyledText.fromString("§eFilter"))) return null;
```
kristofbolyai, line 133: Extract to a const?

common/src/main/java/com/wynntils/models/content/ContentBookQueries.java
```diff
@@ -0,0 +1,206 @@
...
+
+            ContentInfo contentInfo = contentItemOpt.get().getContentInfo();
+
+            newContent.add(contentInfo);
+        }
+    }
+
+    protected void toggleTracking(String name, ContentType contentType) {
+        // We do not want to change filtering when tracking, since we get
+        // no chance to reset it
```
kristofbolyai, line 160: I am not sure why this is the case?

common/src/main/java/com/wynntils/models/quests/QuestModel.java
```diff
@@ -106,11 +112,12 @@ public List<List<StyledText>> getDialogueHistory() {
...
 
     public void startTracking(QuestInfo questInfo) {
-        CONTAINER_QUERIES.toggleTracking(questInfo);
+        Models.Content.startTracking(
+                questInfo.getName(), questInfo.isMiniQuest() ? ContentType.MINI_QUEST : ContentType.QUEST);
     }
 
     public void stopTracking() {
-        McUtils.sendCommand("tracking");
+        Models.Content.stopTracking();
```
kristofbolyai, line 120: Did this break? I would assume the command to be faster...

common/src/main/java/com/wynntils/models/quests/QuestModel.java
```diff
@@ -143,26 +148,56 @@ private String stripPrefix(String name) {
...
+    private static QuestInfo getQuestInfoFromContent(ContentInfo content) {
+        // We should always have a length, but if not, better fake one than crashing
+
+        return new QuestInfo(
+                content.name(),
+                QuestStatus.fromContentStatus(content.status()),
+                QuestLength.fromContentLength(content.length()),
+                content.requirements().level().key(),
+                content.description().orElse(StyledText.EMPTY),
+                // FIXME! Additional requirements missing
```
kristofbolyai, line 191: Speaks for itself

kristofbolyai (approved):
Allow of my comments here are something I want to address later. Right now the focus is getting this out.

common/src/main/java/com/wynntils/models/content/ContentBookQueries.java
magicus: As soon as you click an item to track it, the container closes, but any changes we have made to the filter state will remain.

common/src/main/java/com/wynntils/models/quests/QuestModel.java
magicus: Yes, the command stopped working. :-(

common/src/main/java/com/wynntils/models/quests/QuestModel.java
magicus: Yeah, I forgot about that. But even if I hadn't, I wouldn't have had the time to fix it. Tracked in https://github.com/Wynntils/Artemis/issues/1730.

common/src/main/java/com/wynntils/models/content/ContentBookQueries.java
kristofbolyai: This I failed to observe, and added reset code to tracking... (although it is disabled atm)

common/src/main/java/com/wynntils/models/content/ContentBookQueries.java
magicus: Fixed in #1744.

PR #1698, opened 2023-07-07
fix: Parse loot chests correctly

kristofbolyai:


common/src/main/java/com/wynntils/features/map/MapFeature.java
```diff
@@ -138,12 +137,13 @@ private void openWaypointSetup() {
...
     @SubscribeEvent
-    public void onRightClick(PlayerInteractEvent.RightClickBlock event) {
+    public void onRightClick(PlayerInteractEvent.InteractAt event) {
         if (!autoWaypointChests.get()) return;
 
-        BlockEntity blockEntity = McUtils.mc().level.getBlockEntity(event.getPos());
-        if (blockEntity != null && blockEntity.getType() == BlockEntityType.CHEST) {
-            lastChestPos = event.getPos();
+        Entity entity = event.getEntityHitResult().getEntity();
+        if (entity != null && entity.getType() == EntityType.SLIME) {
```
magicus, line 144: Slime?! :-D

kristofbolyai: Most hidden entities where you need to "click" a text/block is slime because it's consistent shape.

magicus (approved):

common/src/main/java/com/wynntils/features/map/MapFeature.java
magicus: So the lootchest is no longer a normal minecraft chest that you interact with?

kristofbolyai: Yes, basically it is now a "fake" interaction with the slime

PR #1699, opened 2023-07-07
Merge from main

magicus:
I'm not fully sure how the flow of commits between main and development was intended, but here is a merge with features that were added to main that is not present in development. We should integrate them in development, to minimize differences, and to get testing of this functionality on beta.


kristofbolyai:
I think we can merge this, since development -> main PR would be squashed

kristofbolyai (approved):

magicus:
Now updated with the latest main.

kristofbolyai (approved):

PR #1700, opened 2023-07-08
feat: Add an option to override item favorite blocker

kristofbolyai:


common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
```diff
@@ -33,22 +35,49 @@ public class ItemFavoriteFeature extends Feature {
...
+
+        if (lootChestCloseOverride.get() == 0) {
+            McUtils.sendMessageToClient(Component.translatable("feature.wynntils.itemFavorite.closingBlocked")
+                    .withStyle(ChatFormatting.RED));
+            e.setCanceled(true);
+            return;
+        }
+
+        if (lootChestCloseOverride.get() < lootChestCloseOverrideCounter) return;
+        lootChestCloseOverrideCounter++;
```
magicus, line 69: These two lines seem to be in the wrong order..? Have you tried that this work?

kristofbolyai: Yes, it does. "If the current override counter is higher than the set minimum config, do not try to block closing".

magicus: Ah, right, the  comparison was between the config max to the left and the actual counter to the right.

For some reason I found that counter-intuitive. Do you mind fliipping the comparison to a > instead?

kristofbolyai: Did, no problem.

magicus (approved):

PR #1702, opened 2023-07-08
fix: TM filter and search did not work with scroll

ryanzhoudev:


common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -79,9 +82,17 @@ public boolean isGuildBankScreen(Screen screen) {
...
+        if (cs.getMenu().getRowCount() != 6) return false;
+
+        return StyledText.fromComponent(cs.getTitle()).equals(TRADE_MARKET_TITLE);
+    }
+
+    public boolean isSecondaryTradeMarketScreen(Screen screen) {
+        if (!(screen instanceof ContainerScreen cs)) return false;
+        if (cs.getMenu().getRowCount() != 6) return false;
+
+        return StyledText.fromComponent(cs.getTitle()).endsWith(TRADE_MARKET_FILTER_TITLE)
```
magicus: Please use fill regexp instead of endsWith

magicus (approved):

PR #1703, opened 2023-07-09
fix: Function names can contain '_'

kristofbolyai:


kristofbolyai:
I kind of forgot this when creating #1701...

PR #1721, opened 2023-07-12
feat: Add support for the new lootrun camp content type

magicus:


common/src/main/java/com/wynntils/models/content/type/ContentType.java
```diff
@@ -14,7 +14,8 @@ public enum ContentType {
     TERRITORIAL_DISCOVERY("Territorial Discovery", "f"),
     DUNGEON("Dungeon", "c"),
     RAID("Raid", "e"),
-    BOSS_ALTAR("Boss Altar", "d");
+    BOSS_ALTAR("Boss Altar", "d"),
+    LOOTRUN_CAMP("Lootrun Camp", "9");
```
kristofbolyai, line 18: I honestly thought we did the change the color code strings to ChatFormatting here.. But I guess that was only partially done?

kristofbolyai (approved):

common/src/main/java/com/wynntils/models/content/type/ContentType.java
magicus: Yeah, now that you say it... That's weird, I thought I did that. Hm. Did I miss to integrate something?

magicus: I'll fix that in a follow-up anyway.

PR #1724, opened 2023-07-13
feat: Replace the new content book

kristofbolyai:


common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
```diff
@@ -83,28 +83,28 @@ public class WynntilsQuestBookFeature extends Feature {
             () -> WynntilsMenuScreenBase.openBook(WynntilsGuidesListScreen.create()));
 
     @RegisterConfig
-    public final Config<Boolean> replaceWynncraftQuestBook = new Config<>(true);
+    public final Config<Boolean> replaceWynncraftContentBook = new Config<>(true);
```
magicus, line 86: No config upfixers?

kristofbolyai: This was made in a rush to support 2.0.3. Also perhaps it is better to reset the configs this case..

magicus: Okay

Issue #1728, opened 2023-07-13
[2.0.3] exception in container query


magicus:
```
[15:59:18] [Render thread/ERROR] (EventBus) Exception caught during firing event: Cannot invoke "com.wynntils.handlers.container.scriptedquery.QueryStep.getVerification()" because "this.currentStep" is null
    Index: 1
    Listeners:
        0: NORMAL
        1: ASM: com.wynntils.handlers.container.ContainerQueryHandler@6683ff7b onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
        2: ASM: com.wynntils.models.character.CharacterSelectionModel@6d2ecbad onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
        3: ASM: com.wynntils.models.worlds.WorldStateModel@36ef033c onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
        4: ASM: com.wynntils.models.containers.LootChestModel@2fca32a3 onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
        5: ASM: com.wynntils.models.seaskipper.SeaskipperModel@3adae4c1 onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
        6: ASM: com.wynntils.models.emeralds.EmeraldModel@4214711a onMenuOpened(Lcom/wynntils/mc/event/MenuEvent$MenuOpenedEvent;)V
java.lang.NullPointerException: Cannot invoke "com.wynntils.handlers.container.scriptedquery.QueryStep.getVerification()" because "this.currentStep" is null
    at com.wynntils.handlers.container.scriptedquery.ScriptedContainerQuery.verifyContainer(ScriptedContainerQuery.java:56)
    at com.wynntils.handlers.container.ContainerQueryHandler.onMenuOpened(ContainerQueryHandler.java:110)
    at com.wynntils.handlers.container.__ContainerQueryHandler_onMenuOpened_MenuOpenedEvent.invoke(.dynamic)
    at net.minecraftforge.eventbus.ASMEventHandler.invoke(ASMEventHandler.java:73)
    at net.minecraftforge.eventbus.EventBus.post(EventBus.java:315)
    at net.minecraftforge.eventbus.EventBus.post(EventBus.java:296)
[...]

PR #1733, opened 2023-07-13
fix: Fix fallback code for closing bad container queries

### LINKED ISSUE #1732: [2.0.3] Fail-back logic for closing containers in bad queries is wrong
Found an issue while reading code to try to fix #1728.

```java
if (currentStep == null && lastStep != null) {
            // We're in a possibly bad state. We have failed a previous call, but
            // we might still get the menu opened (perhaps after a lag spike).
            if (lastStep.verifyContainer(e.getTitle(), e.getMenuType())) {
                // This was the container we were supposed to be looking for
                WynntilsMod.warn("Closing container '" + e.getTitle().getString()
                        + "' due to previously aborted container query");
                e.setCanceled(true);
            } else {
                // This is some other container. Ignore it.
            }
            // Now say we're completely finished with the last query
            lastStep = null;
            return;
        }
 ```
 
This code does not do what it was meant to. `e.setCanceled(true);` will NOT close the container, rather, it will not display it, but make the server think we are in it.

kristofbolyai:
If I understand the code correctly, this should fix #1732. Please verify that this is how this code is meant to be executed.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/container/ContainerQueryHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/container/ContainerQueryHandler.java
@@ -112,6 +112,7 @@ public final class ContainerQueryHandler extends Handler {
                 WynntilsMod.warn("Closing container '" + e.getTitle().getString()
                         + "' due to previously aborted container query");
                 e.setCanceled(true);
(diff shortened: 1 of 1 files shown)


magicus:
Yes, this is probably correct. But see #1728, maybe this entire block of code can be removed.

magicus (approved):

Issue #1729, opened 2023-07-13
[2.0.3] content book parsing is too slow


magicus:
Consider removing the trailing filter reset to speed up things.

Additionally, we should save the last read content items in a Storage (per character), so we can present the Quest Book immediately, and then reload in the background.

kristofbolyai:
The cause is rather obvious, when taken a closer look. We have 5 use cases of content book parsing.

Quest book screen runs 2 separate full queries for mini and normal quests.
Discovery screen runs 3 separate full queries for it's discovery types.

We need to look at reducing this.

kristofbolyai:
Not worth to keep this issue, this most likely stays until we change the content book to be a wrapped screen.

PR #1736, opened 2023-07-13
fix: Speed up content book parsing by disabling filter resetting

kristofbolyai:


kristofbolyai:
This should help with #1729. However, I would not close that issue, yet.

magicus (approved):

PR #1737, opened 2023-07-14
fix: Skill points annotator did not parse properly

ryanzhoudev:
Changed in 2.0.3, fixes skill points not displaying properly

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/annotators/gui/SkillPointAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/gui/SkillPointAnnotator.java
@@ -15,8 +15,10 @@ import java.util.regex.Pattern;
 import net.minecraft.world.item.ItemStack;
 
 public final class SkillPointAnnotator implements ItemAnnotator {
+    // Test suite: https://regexr.com/7h0tl
     private static final Pattern SKILL_POINT_PATTERN = Pattern.compile("^§dUpgrade your §[2ebcf][✤✦❉✹❋] (.*)§d skill$");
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/models/items/annotators/gui/SkillPointAnnotator.java
```diff
@@ -26,7 +28,8 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
         String skillName = matcher.group(1);
         Skill skill = Skill.fromString(skillName);
 
-        Matcher m = LoreUtils.matchLoreLine(itemStack, 3, LORE_PATTERN);
+        // intelligence skill points has an extra % line so it will be on line 4
+        Matcher m = LoreUtils.matchLoreLine(itemStack, skill == Skill.INTELLIGENCE ? 4 : 3, LORE_PATTERN);
```
kristofbolyai: I honestly would rather loop.. Performance is trivial, but we can save ourselves from trouble.

kristofbolyai: But I won't require it here, up to you.

kristofbolyai (approved):

common/src/main/java/com/wynntils/models/items/annotators/gui/SkillPointAnnotator.java
magicus: You don't even need this extra code, matchLoreLine checks a handful of lines after the starting line as well, so you can just keep the `3`.

magicus (approved):

PR #1738, opened 2023-07-14
feat: Archetype/point counts in ability tree screen

ryanzhoudev:


common/src/main/java/com/wynntils/models/items/annotators/gui/AbilityTreeAnnotator.java
```diff
@@ -14,17 +14,32 @@
 import net.minecraft.world.item.ItemStack;
 
 public final class AbilityTreeAnnotator implements ItemAnnotator {
+    // Deals with the ability tree button in the compass menu
     private static final StyledText ABILITY_TREE_NAME = StyledText.fromString("§b§lAbility Tree");
     private static final Pattern ABILITY_POINTS_PATTERN = Pattern.compile("^§3✦ Unused Points: §f(\\d+)$");
 
+    // Deals with the reset button in the ability tree screen
+    private static final StyledText INTERNAL_ABILITY_TREE_NAME = StyledText.fromString("§3§lAbility Points");
```
kristofbolyai: I dont understand what is happening here..

common/src/main/java/com/wynntils/models/items/annotators/gui/ArchetypeAnnotator.java
```diff
@@ -0,0 +1,36 @@
...
+    // Test suite: https://regexr.com/7h133
+    private static final Pattern ARCHETYPE_PATTERN = Pattern.compile("^§a✔ §7Unlocked Abilities: §f(\\d+)§7/1[56]$");
+
+    @Override
+    public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
+        Matcher nameMatcher = name.getMatcher(ARCHETYPE_NAME);
+        if (!nameMatcher.matches()) return null;
+
+        // certain archetypes like Shadestepper have an extra line of description
+        int matchLine = LoreUtils.getLoreLine(itemStack, 5) == StyledText.EMPTY ? 6 : 5;
```
kristofbolyai: we can just loop, no need for this shady stuff that is going to break after a while

magicus: I don't think its shady. It is a decent performance/correctness optimization to just look about the correct line. But Donkey had missed the fact that you only need to be roughly precise with the number. So we can use 5 all the time here, no need to be that specific.

common/src/main/java/com/wynntils/models/items/annotators/gui/ArchetypeAnnotator.java
```diff
@@ -0,0 +1,36 @@
...
+import com.wynntils.utils.mc.LoreUtils;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.world.item.ItemStack;
+
+public final class ArchetypeAnnotator implements ItemAnnotator {
+    // Test suite: https://regexr.com/7h12h
+    private static final Pattern ARCHETYPE_NAME = Pattern.compile("^§([a-r0-9])§l[A-Za-z ]+ Archetype$");
+    // Test suite: https://regexr.com/7h133
+    private static final Pattern ARCHETYPE_PATTERN = Pattern.compile("^§a✔ §7Unlocked Abilities: §f(\\d+)§7/1[56]$");
```
magicus: Why not just specify `\\d` as the max? Or better yet, parse the max as well and return a CappedValue.

common/src/main/java/com/wynntils/models/items/annotators/gui/AbilityTreeAnnotator.java
magicus: There are two screens where the points are presented, the compass menu and the ability tree menu? They have different names on the two items that has the number of ability points in the lore. I don't think that is strange? 

Possibly the names could be made clearer: COMPASS_MENU_ABILITY_POINTS_NAME and ABILITY_TREE_ABILITY_POINTS_NAME or something like that.

common/src/main/java/com/wynntils/models/items/items/gui/ArchetypeItem.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright © Wynntils 2022.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.items.gui;
+
+import com.wynntils.models.items.properties.CountedItemProperty;
+import net.minecraft.ChatFormatting;
+
+public class ArchetypeItem extends GuiItem implements CountedItemProperty {
```
magicus: I think this should be named `ArchetypeAbilitiesItem` (and correspondingly for the annotator), unless you plan to extend this item to contain more information in the future. (I don't remember from the top of my head if there are more information to be gained)

common/src/main/java/com/wynntils/models/items/items/gui/ArchetypeAbilitiesItem.java
```diff
@@ -5,25 +5,30 @@
...
+        return count.current();
     }
 
     @Override
     public boolean hasCount() {
-        return count != 0;
+        return count.current() != 0;
+    }
+
+    public int getMax() {
```
magicus: Since this is for future expansion, I think it is better to have a getter for the CappedValue. Due to name conflicts it needs to be called like getAbilitiesCount instead; and to make the getter be sane, I recommend renaming the field to match.

magicus (approved):

PR #1741, opened 2023-07-14
fix: Fix multiple problems related to content book queries and tracking

kristofbolyai:


common/src/main/java/com/wynntils/handlers/container/ContainerQueryHandler.java
```diff
@@ -141,7 +122,11 @@ public void onMenuForcefullyClosed(MenuEvent.MenuClosedEvent e) {
 
         // Server closed our container window. This should not happen
         // but if it do, report failure
-        raiseError("Server closed container");
+        if (e.getContainerId() == containerId) {
+            raiseError("Server closed container");
+        } else {
+            WynntilsMod.warn("Server closed container " + e.getContainerId() + " but we are querying " + containerId);
+        }
```
kristofbolyai, line 129: This is purely experimental, but this allows even 10 queries to run after each other without problems, where as without this 2/3 repeated queries failed..

common/src/main/java/com/wynntils/handlers/container/ContainerQueryHandler.java
```diff
@@ -46,7 +45,7 @@ public void runQuery(ContainerQueryStep firstStep) {
         if (currentStep != null) {
             // Only add if it is not already enqueued
             if (queuedQueries.stream()
-                    .filter(query -> query.getName().equals(firstStepName))
+                    .filter(query -> query.getName().equals(firstStep.getName()))
```
kristofbolyai, line 48: What a typo! I am sure this is obvious after a second look...

kristofbolyai:
I want to request a post-merge review, @magicus.

common/src/main/java/com/wynntils/handlers/container/ContainerQueryHandler.java
magicus: Yeah, really! Maybe we should rename it, like storedNameOfFirstStep....

PR #1752, opened 2023-07-14
fix: Use hopefully improved method of opening e.g. content book for queries

magicus:


magicus:
This patch was created by @kristofbolyai. I am just integrating it.

PR #1754, opened 2023-07-14
feat: Custom jump pages for bookshelf and misc bucket

ShadowCat117:
I'm fairly sure the misc bucket has 10 pages and I've just assumed the same for the bookshelf, if anyone with champion and has all pages knows the max and it's not 10 then let me know.

Also, since by default Wynn still uses the same jump buttons as a bank whilst having less pages, I've set the defaults to some slightly better options.


magicus:
Woah, there's a lot of duplication here!

I understand the config screen will start to look messy if you add 18 different integer configs. But we can't let that force us to replicate the same code in three different places. We need to come up with some better solution here...

magicus:
What if we skip using integers as tbe config variable, and instead just use a string, that you will parse and check so it is six comma-separated numbers? Then you'd need just three config options, the page order for bank, bucket and bookshelf respectively. That'd be manageble.

And in the future, we could add some kind of "integer list" type to configs, and make the parsing go away.

ShadowCat117:
Yeah had a feeling that 3 features doing essentially the same thing was going to be a bad idea.

I'll try that string list tomorrow and as a quick thought, would an enum for current container type in the feature be a better idea than if bankScreen else if bookshelf else if miscbucket?

magicus:
There already is a SearchableContainerType enum. It should probably be extended to the bookshelf. (And the ContainerSearchFeature extended to search the bookshelf as well. If you'd like to fix that as well, please go ahead, I think it can make sense to put it in the same PR.)

ShadowCat117:
How's this? I think the area most likely in need of review is validating the config.

https://github.com/Wynntils/Artemis/assets/8337467/2bf44e23-ace0-4c3a-a788-7e0e5c74ac37

ShadowCat117:
I've also just found a bug where if you click the ingredient pouch whilst in a bank screen it doesn't trigger the close event. I've made a simple fix which just happens in onSlotClicked and if it was the ingredient pouch slot it will do what the close event does. The better solution would be to make that scenario actually trigger the close event but I'm not sure how to do that.

Should I add my fix to this PR or make a seperate one?

ryanzhoudev:
> I've also just found a bug where if you click the ingredient pouch whilst in a bank screen it doesn't trigger the close event. I've made a simple fix which just happens in onSlotClicked and if it was the ingredient pouch slot it will do what the close event does. The better solution would be to make that scenario actually trigger the close event but I'm not sure how to do that.
> 
> Should I add my fix to this PR or make a seperate one?

I did not see this. #1761 deals with it.

ShadowCat117:
> I did not see this. #1761 deals with it.

Ah nice thanks, I'll change that on this branch so it doesn't conflict

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -4,6 +4,10 @@
  */
 package com.wynntils.features.inventory;
 
+import static com.wynntils.models.containers.type.SearchableContainerType.BANK;
```
magicus: Please use fully qualified names (`SearchableContainerType.BANK`) instead of static imports.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -94,16 +95,35 @@ public void onScreenInit(ScreenInitEvent e) {
...
+            case MISC_BUCKET -> configDestinations = miscBucketDestinations.get();
+            default -> {
+                return;
+            }
+        }
+
+        customJumpDestinations = parseStringToDestinations(configDestinations);
+
+        if (customJumpDestinations == null) {
+            customJumpDestinations = QUICK_JUMP_DESTINATIONS;
```
magicus: This is the correct default only for bank pages, not housing island containers, right?

ShadowCat117: Yes, I didn't want to repeat another switch statement and in theory this should only fail if someone manually edited their config file as (hopefully) the onConfigUpdate should catch any non valid inputs

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -219,19 +244,61 @@ private void clickPreviousPage() {
...
+    private List<Integer> parseStringToDestinations(String destinationsStr) {
+        String[] destinationStrings = destinationsStr.split(",");
+
+        if (destinationStrings.length != 6) {
+            return null;
+        }
+
+        List<Integer> destinations = new ArrayList<>();
+
+        for (String destinationString : destinationStrings) {
```
magicus: I think this can be written in a more modern Java style using functional expressions and stream. Something along the lines of `destinationStrings.stream.map(Integer::parseInt).toList()`.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -184,19 +186,13 @@
...
-  "feature.wynntils.customBankPages.buttonOnePage.description": "Which page should button 1 jump to?",
-  "feature.wynntils.customBankPages.buttonOnePage.name": "Button 1 page",
-  "feature.wynntils.customBankPages.buttonSixPage.description": "Which page should button 6 jump to?",
-  "feature.wynntils.customBankPages.buttonSixPage.name": "Button 6 page",
-  "feature.wynntils.customBankPages.buttonThreePage.description": "Which page should button 3 jump to?",
-  "feature.wynntils.customBankPages.buttonThreePage.name": "Button 3 page",
-  "feature.wynntils.customBankPages.buttonTwoPage.description": "Which page should button 2 jump to?",
-  "feature.wynntils.customBankPages.buttonTwoPage.name": "Button 2 page",
-  "feature.wynntils.customBankPages.description": "Enable custom quick jump buttons in the bank?",
+  "feature.wynntils.customBankPages.bankDestinations.description": "Which pages would you like your bank quick jump buttons to jump to? Pages should be separated with an single comma.",
```
```suggestion
  "feature.wynntils.customBankPages.bankDestinations.description": "Which pages would you like your bank quick jump buttons to jump to? Enter a list of six numbers separated with a comma.",
```

(and similar for the others)

magicus:
Overall I think this looks pretty nice. I had a few comments, nothing major.

magicus:
As a general remark, our current implementation of ContainerModel is pretty dumb. I think we should put much more "intelligence" into it, which would mean moving some of the logic from this feature into the model. But that is definitely a future undertaking.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: Can change it to

```
switch (currentContainer) {
    case BANK -> customJumpDestinations = QUICK_JUMP_DESTINATIONS;
    case BOOKSHELF, MISC_BUCKET -> customJumpDestinations = HOUSING_DEFAULT_DESTINATIONS;
}
```

with `private static final List<Integer> HOUSING_DEFAULT_DESTINATIONS = List.of(1, 3, 4, 6, 8, 10);`?

magicus: Yes, please do!

ShadowCat117: Should be good to go now unless you have some more improvements to suggest :)

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -219,19 +245,55 @@ private void clickPreviousPage() {
...
+
+    private List<Integer> parseStringToDestinations(String destinationsStr) {
+        String[] destinationStrings = destinationsStr.split(",");
+
+        if (destinationStrings.length != 6) {
+            return null;
+        }
+
+        try {
+
```
kristofbolyai, line 295:
```suggestion
```

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -219,19 +245,55 @@ private void clickPreviousPage() {
...
+            return Arrays.stream(destinationStrings).map(Integer::parseInt).collect(Collectors.toList());
```
kristofbolyai: For the sake of sanity, shouldn't you limit this to whatever slots we have?

ShadowCat117: At that point I can just remove the check for it being 6 long right?

kristofbolyai: Oh, I didn't see that up there. But yes, limiting is a better way than not parsing it at all.

ShadowCat117: Actually, will it work if it's less than 6? Can try something like populating it with values from the default if less

kristofbolyai: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Stream.html#limit(long)

ShadowCat117: How are those changes I've just made. If you don't enter 6 values it will add the missing values from the default

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -246,35 +247,47 @@ private void clickPreviousPage() {
...
-        }
-
         try {
-            return Arrays.stream(destinationStrings).map(Integer::parseInt).collect(Collectors.toList());
+            List<Integer> destinations = Arrays.stream(destinationStrings)
+                    .limit(MAX_DESTINATIONS)
+                    .map(Integer::parseInt)
+                    .collect(Collectors.toCollection(ArrayList::new));
+
+            if (destinations.size() != MAX_DESTINATIONS) {
```
magicus:
```suggestion
            if (destinations.size() < MAX_DESTINATIONS) {
```

Logically equivalent but clearer.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: Your changes are fine. I'm brazen enough to speak for bolyai as well. :)

magicus (approved):

magicus:
I seem to have botched the merge. :( That's what you get for trying to use the Github web interface...

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -145,15 +167,20 @@ public void onContainerSetEvent(ContainerSetContentEvent.Post e) {
         if (!isBankScreen) return;
```
magicus:
```suggestion
        if (currentContainer == null) return;
```

ShadowCat117: Ah, I just updated my fork and fixed this but you got there first ;)

PR #1761, opened 2023-07-15
fix: Clicking on ingredient pouch does not reset bank status

ryanzhoudev:
https://discord.com/channels/394189072635133952/1023381446733352990/1129914781248274502


magicus:
We do not need to look at both events, just to be sure? Or is the screen close event good enough for all purposes?

magicus (approved):

ryanzhoudev:
Screen close event seems to work for normal closes as well. We listen for ScreenInit anyway, so theoretically, the opposite event is ScreenClose instead of ContainerClose.

PR #1762, opened 2023-07-15
fix: Force closing favourite item count goes negative

ryanzhoudev:
https://discord.com/channels/394189072635133952/1023381446733352990/1128300739987509360


common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
```diff
@@ -58,15 +58,15 @@ public void onChestCloseAttempt(ContainerCloseEvent.Pre e) {
...
 
-        if (lootChestCloseOverride.get() == 0) {
+        if (lootChestCloseOverride.get() == 0) { // never allow user to close
             McUtils.sendMessageToClient(Component.translatable("feature.wynntils.itemFavorite.closingBlocked")
                     .withStyle(ChatFormatting.RED));
             e.setCanceled(true);
             return;
         }
 
-        if (lootChestCloseOverrideCounter > lootChestCloseOverride.get()) return;
```
magicus, line 68: So, why does this resolve the bug? My gut reaction is that this looks more sane and the original code looks suspicious (and I think I tried to argue about it woth bolyai in his original PR recently), but when I try to reason about it, I seem to end at the conclusion that it should not matter. But apparently it does. What am I missing?

ryanzhoudev: lootChestCloseOverrideCounter starts at 0. When the user clicks the close button, this new code increments the count by 1, then checks if it has passed the override threshold. Previously, the check happened first. According to the description, "After how many ESC presses should the chest be closed? Set to 0 to disable.", we should increment this first because pressing it once would be after one press. 
So, the process for this new code (with threshold 3) would go like:
- User loots chest and presses close, leaving a favourite item in there
- We increment the counter by 1, it is now at 1, then run the check. Since 1<3, we tell the user that there are 2 (3-1) more presses before closing.
- User clicks again
- Counter is at 2. Run check, 2<3, there are 1 (3-2) more presses before closing (next press closes)
- User clicks again
- Counter is at 3. Run check, 3=3, close chest.
- So, 3 clicks were required, matching the setting of 3.

(continued)

ryanzhoudev: Previously, with the threshold at 3:
- User loots chest and presses close
- We run the check, 0 < 3, there are 3 (3-0) presses remaining. Increment counter by 1, it is at 1.
- User clicks
- Check, 1 < 3, there are 2 (3-1) presses remaining, counter is at 2.
- Click
- Check, 2 < 3, there are 1 (3-2) presses remaining, counter at 3.
- Click
- Check, 3 = 3, BUT, the previous check looked for (clicks > threshold), so again, there are 0 (3-3) presses remaining, counter at 4.
- Click
- Check, 4 > 3, finally, close chest
- So, 4 clicks were required, not matching 3.

magicus (approved):

PR #1763, opened 2023-07-15
fix: Search incorrectly stops after empty item

ryanzhoudev:
https://discord.com/channels/394189072635133952/424990456435310602/1129919273641455686
https://discord.com/channels/394189072635133952/424990456435310602/1129923253318721606

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
@@ -58,8 +58,7 @@ public class ContainerSearchFeature extends Feature {
     // If the guild bank has lots of custom (crafted) items, it can take multiple packets and a decent amount of time
     // for Wynn to send us the entire updated inventory. During this, the inventory will be in a weird state where
     // some items are updated and some are not. We will assume that after GUILD_BANK_SEARCH_DELAY milliseconds, the
-    // inventory
-    // is fully updated.
+    // inventory is fully updated.
     private static final int GUILD_BANK_SEARCH_DELAY = 500;
     private long guildBankLastSearch = 0;
 
--- a/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
@@ -28,11 +28,11 @@ import com.wynntils.utils.colors.CustomColor;
 import com.wynntils.utils.mc.McUtils;
 import com.wynntils.utils.render.RenderUtils;
 import com.wynntils.utils.wynn.ContainerUtils;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -201,11 +201,13 @@ private void addSearchWidget(AbstractContainerScreen<?> screen, int renderX, int
     private void matchItems(String searchStr, AbstractContainerScreen<?> screen) {
         String search = searchStr.toLowerCase(Locale.ROOT);
 
-        NonNullList<ItemStack> playerItems = McUtils.inventory().items;
-        for (ItemStack itemStack : screen.getMenu().getItems()) {
+        List<ItemStack> containerItems = screen.getMenu().getItems().stream()
```
magicus, line 204: So what we do want in the end is to iterate through all items that are in the container, but not the players inventory, right?

This probably works but does not look very elegant. I can say on the top of my head how to do this properly, but there ought to be a better way to do this. And if not, we should write one and add it to McUtils.

Mcrtin: I've been looking in some of the Minecraft menus (like `BrewingStandMenu`) and I've noticed that they always add the inventory slots after their own
```java

this.addSlot(new BrewingStandMenu.PotionSlot(container, 0, 56, 51));
this.addSlot(new BrewingStandMenu.PotionSlot(container, 1, 79, 58));
this.addSlot(new BrewingStandMenu.PotionSlot(container, 2, 102, 51));
this.ingredientSlot = this.addSlot(new BrewingStandMenu.IngredientsSlot(container, 3, 79, 17));
this.addSlot(new BrewingStandMenu.FuelSlot(container, 4, 17, 17));
this.addDataSlots(containerData);



for(int j = 0; j < 3; ++j) {
	for(int k = 0; k < 9; ++k) {
		this.addSlot(new Slot(inventory, k + j * 9 + 9, 8 + k * 18, 84 + j * 18));
	}
}

for(int j = 0; j < 9; ++j) {
	this.addSlot(new Slot(inventory, j, 8 + j * 18, 142));
}
```
So I've coded a bit of a sketchy solution which just drops the appended inventory slots from the stream (not tested) `screen.getMenu().getItems().stream().limit(screen.getMenu().slots.size() - McUtils.inventory().getContainerSize())`

kristofbolyai:
Oh.. I did just merge a fix for most likely the same problem. Why is all of that change needed, I think it was just the return?

ryanzhoudev:
Yeah we don't need this anymore. I just thought it was strange to try to parse to wynnItemOpt *before* checking if it was some item in the player's inventory.

PR #1766, opened 2023-07-16
fix: Container queries now work in various changing conditions

kristofbolyai:
This is the fix.. I have not been able to get this tested, but as far as I am aware, this should work. I will wait for a tester, than merge this.


kristofbolyai:
Aand this doesn't solve the issue for one tester. Perhaps @DonkeyBlaster could test it too, to see if it is completely bad...

ryanzhoudev:
This seems to work for me, on two of the settings that didn't yesterday.

kristofbolyai:
From discord:

Donkey's instance seem to be working without this...

Maybe this is a wynn thing, let's just keep this here for a while.

kristofbolyai:
This is not needed, for now.

PR #1768, opened 2023-07-16
fix: Linux crashes when changing rich presence settings

ryanzhoudev:


kristofbolyai:
If you happen to fix this while I am sleeping, and merge it, just remember that you need to free memory. You are leaking activity here.

PR #1769, opened 2023-07-17
fix: Linux instances crash when changing any setting with rich presence settings

ryanzhoudev:


common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -121,7 +123,11 @@ private void checkTerritory() {
     @Override
     protected void onConfigUpdate(ConfigHolder configHolder) {
         if (this.isEnabled()) {
-            Managers.Discord.load();
+            // This isReady() check is required for Linux to not crash on config change.
+            if (!Managers.Discord.isReady()) {
+                // Even though this is in the onConfigUpdate method, it is how the library is first loaded on launch
+                Managers.Discord.load();
+            }
```
kristofbolyai, line 130: So basically it can't reload it? This method get's called 3 times on load.. Only loading once makes it not crash when changing settings?

ryanzhoudev: Seems like it, can you also test on your instance?

kristofbolyai (approved):
Works fine.

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -88,7 +89,8 @@ public void onDisconnect(ConnectionEvent.DisconnectedEvent e) {
     }
 
     private void displayCharacterDetails() {
-        int level = Models.CombatXp.getCombatLevel().current();
+        CappedValue combatLevel = Models.CombatXp.getCombatLevel();
+        int level = combatLevel == null ? 0 : combatLevel.current();
```
magicus, line 93: Why did you think this was necessary? getCombatLevel cannot return null.

(In fact, I think we should verify so models *never* return null values)

ryanzhoudev: I know something related to this returned null before, as it called player.experienceLevel or something. I will check again in the morning

PR #1770, opened 2023-07-17
fix: Fix unreliable container queries by adding a tick delay

kristofbolyai:


magicus (approved):
This looks reasonable, and nicely implemented.

kristofbolyai:
We could not find issues on our characters, or any testers. But let's merge it, and see what comes back.

PR #1771, opened 2023-07-17
fix: Characters with modifiers show up as "This Character" in list

ryanzhoudev:


kristofbolyai (approved):
Looks reasonable, I thought this was fixed once already..

PR #1773, opened 2023-07-17
fix: Fix trying to set discord status before joining server

kristofbolyai:


kristofbolyai:
This is what you were looking for @DonkeyBlaster

PR #1774, opened 2023-07-17
feat: Add GatheringCooldownOverlayFeature and GatheringCooldownFunction

### LINKED ISSUE #1716: Gather Cooldowns When Switching Worlds
Legacy has a timer overlay that shows how long you have to wait before opening chests or being able to gather.

With chests this is a bit more complicated as different tiers have different cooldowns (which I don't know - please comment below if you do).

Gathering is always 60 seconds after world join without prof speed and 30 seconds with prof speed.

kristofbolyai:
Fixes #1716


magicus:
The original gathering cooldown in legacy only showed up once you had tried to gather, and got a "You need to wait X seconds before gathering" message (or whatever it said).

magicus:
Otherwise you'll just end up with a timer counting down from world join time. Sounds like it could be mostly annoying.

magicus:
> The timer seems to be off a bit...

Maybe you can adjust the timer based on the time given in the chat message? (That was the sole source of timer information in legacy)

kristofbolyai:
> > The timer seems to be off a bit...
> 
> Maybe you can adjust the timer based on the time given in the chat message? (That was the sole source of timer information in legacy)

Not really a great way either. Using 70/35 sec is stable (and is confirmed by profession people).

common/src/main/java/com/wynntils/functions/WorldFunctions.java
```diff
@@ -289,4 +290,19 @@ public FunctionArguments.Builder getArgumentsBuilder() {
                     List.of(new FunctionArguments.Argument<>("prefixOnly", Boolean.class, false)));
         }
     }
+
+    public static class GatheringCooldownFunction extends Function<Integer> {
+        @Override
+        public Integer getValue(FunctionArguments arguments) {
+            int cooldownLength = Models.Bomb.isBombActive(BombType.PROFESSION_SPEED)
```
magicus: Can we move this logic into ProfessionModel instead? I don't like it when functions become more than thin wrappers for the models. I understand this adds a dependency to the BombModel to ProfessionModel, but that is honest, since this is how things work.

kristofbolyai: Moved the relevant parts to the model.

magicus (approved):

PR #1775, opened 2023-07-17
feat: TM price match buttons

ryanzhoudev:
image
image

Also moved bulk sell to UI since it was not an overlay


common/src/main/java/com/wynntils/features/ui/TradeMarketPriceMatchFeature.java
```diff
@@ -0,0 +1,169 @@
...
+import net.minecraft.world.inventory.ChestMenu;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UI)
+public class TradeMarketPriceMatchFeature extends Feature {
+    private static final StyledText CLICK_TO_SET_PRICE = StyledText.fromString("§aClick to Set Price");
+    private static final StyledText SELL_DIALOGUE_TITLE = StyledText.fromString("What would you like to sell?");
+    private static final StyledText TYPE_SELL_PRICE =
+            StyledText.fromString("§6Type the price in emeralds or type 'cancel' to cancel:");
```
kristofbolyai: This is kind of duplicated knowledge across two features. In a perfect world, we would have a singular TradeMarketModel (and do most calculations there, expect displaying). I am not sure if I require that here though, or if magicus does.

magicus: At some point we should probably introduce a TradeMarketModel, yes. Not sure it is needed for this PR though.

common/src/main/java/com/wynntils/features/ui/TradeMarketPriceMatchFeature.java
```diff
@@ -0,0 +1,169 @@
...
+    @SubscribeEvent
+    public void onChatMessage(ChatMessageReceivedEvent e) {
+        if (!sendPriceMessage) return;
+        if (!e.getStyledText().equals(TYPE_SELL_PRICE)) return;
+
+        WynntilsMod.info("Trying to set trade market price to " + priceToSend);
+
+        McUtils.mc().getConnection().sendChat(String.valueOf(priceToSend));
+
+        sendPriceMessage = false;
```
magicus: Can you cancel e?

ryanzhoudev: Yes, but I don't really see a reason to. It is one small message that does not block anything, and it also helps the user understand why the trade market screen closes and opens quickly. Newer users may not know that we are interacting with the chat.

common/src/main/java/com/wynntils/core/features/FeatureManager.java
```diff
@@ -124,6 +123,8 @@
 import com.wynntils.features.ui.CustomSeaskipperScreenFeature;
 import com.wynntils.features.ui.LobbyUptimeFeature;
 import com.wynntils.features.ui.SoulPointTimerFeature;
+import com.wynntils.features.ui.TradeMarketBulkSellFeature;
```
magicus: With #1780 you can now put it in the trademarket category instead.

common/src/main/java/com/wynntils/features/ui/TradeMarketPriceMatchFeature.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private static final int PRICE_SET_ITEM_SLOT = 12;
+    private static final int PRICE_INFO_ITEM_SLOT = 17;
+
+    private boolean sendPriceMessage = false;
+    private long priceToSend = 0;
+
+    @SubscribeEvent
+    public void onSellDialogueUpdated(ContainerSetSlotEvent.Pre e) {
+        if (!(McUtils.mc().screen instanceof ContainerScreen containerScreen)) return;
```
magicus: You are doing this once per set slot event. Maybe use the set container event instead?

magicus (approved):
Overall looks okay

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -976,6 +976,14 @@
   "feature.wynntils.tradeMarketBulkSell.sellAll": "Sell All",
   "feature.wynntils.tradeMarketPriceConversion.description": "Extends entering prices from just a number to using math and having shorthand for Emeralds, Emerald Blocks, Liquid Emeralds, Liquid Emerald Stacks. Also adds the ability to include tax in the price.",
   "feature.wynntils.tradeMarketPriceConversion.name": "Market Price Conversion",
+  "feature.wynntils.tradeMarketPriceMatch.description": "Adds buttons to match lowest sell offer and highest buy offer in the trade market UI",
+  "feature.wynntils.tradeMarketPriceMatch.highestBuyOffer": "Highest Buy Offer",
+  "feature.wynntils.tradeMarketPriceMatch.highestBuyOfferTooltip": "Quick sells the item for the highest buy offer.",
+  "feature.wynntils.tradeMarketPriceMatch.lowestSellOffer": "Lowest Sell Offer",
+  "feature.wynntils.tradeMarketPriceMatch.lowestSellOfferTooltip": "Matches the price of the current lowest sell offer.",
```
magicus: The two tooltips have different wordings, but they do basically the same thing but picks highest respectively the lowest price, yes? Can you adjust them to match, if so. Otherwise, explain better how they differ.

magicus (approved):
The code looks fine. Just fix the i18n for consistency.

PR #1778, opened 2023-07-18
feat: Add sort button to discoveries, prepare quest book redesign

kristofbolyai:


common/src/main/java/com/wynntils/models/discoveries/DiscoveryModel.java
```diff
@@ -174,11 +176,29 @@ public List<Component> getSecretDiscoveriesTooltip() {
...
-        return Stream.concat(
-                        Stream.concat(territoryDiscoveries.stream(), worldDiscoveries.stream()),
-                        secretDiscoveries.stream())
-                .filter(DiscoveryInfo::isDiscovered);
+    public Stream<DiscoveryInfo> getAllDiscoveries(ContentSortOrder sortOrder) {
+        if (sortOrder == ContentSortOrder.DISTANCE) {
+            throw new IllegalArgumentException("Cannot sort discoveries by distance");
+        }
+
+        // All quests are always sorted by status (available then unavailable), and then
```
magicus:
```suggestion
        // All discoveries are always sorted by status (available then unavailable), and then
```

common/src/main/java/com/wynntils/screens/base/widgets/SortableContentScreen.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.screens.base.widgets;
+
+import com.wynntils.models.content.type.ContentSortOrder;
+
+public interface SortableContentScreen {
+    ContentSortOrder getQuestSortOrder();
```
magicus: Maybe getContentSortOrder()?

kristofbolyai: Fixed

common/src/main/java/com/wynntils/screens/discoveries/WynntilsDiscoveriesScreen.java
```diff
@@ -255,9 +267,26 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
     }
 
     private void renderTooltip(PoseStack poseStack, int mouseX, int mouseY) {
-        if (!(this.hovered instanceof TooltipProvider tooltipWidget)) return;
+        List<Component> tooltipLines = new ArrayList<>();
+
+        if (this.hovered instanceof SortOrderWidget) {
+            switch (contentSortOrder) {
+                case LEVEL -> tooltipLines = List.of(
```
magicus: Several comments. 

The i18n name `screens.wynntils.wynntilsQuestBook` is not really correct here. If it is shared between screens, maybe it need a more generic name?

This is borderline close to begging to be generalized. I'm not sure it should be the role of ContentSortOrder to know how it should be displayed, but ... maybe ..? At least the i18n names seem to be derivable from the sort order.

kristofbolyai: Fixed.

common/src/main/java/com/wynntils/screens/base/widgets/SortOrderWidget.java
```diff
@@ -73,4 +75,19 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
                     sortTexture.height());
         }
     }
+
+    @Override
+    public List<Component> getTooltipLines() {
```
magicus, line 81: I did not even realize this was duplicated across the PR...! :-)

So fixing this was certainly nice.

But what I meant was should we have something like
  `Component.translatable("screens.wynntils.wynntilsContentBook.sort." + sortableContentScreen.getContentSortOrder().name.toLowerCase() +".name")` instead... :)

I am not 100% sure, but I don't like the repetition. 3 copies is just about my personal rule of thumb when you are allowed to copy instead of generalize, but it always leaves a bad taste to let such instances pass.

You don't have to do this, but at least I'd like to hear an argument of why not. :)

kristofbolyai: Fixed it.

magicus (approved):

PR #1784, opened 2023-07-19
feat: Redesign quest book page to look similarly to discovery page

kristofbolyai:
Merge https://github.com/Wynntils/Artemis/pull/1778 first.

image


magicus:
The screenshot indicates that the quest/miniquest selection button in the top is still there. That looks a bit odd, should it not be removed?

common/src/main/java/com/wynntils/screens/questbook/WynntilsQuestBookScreen.java
```diff
@@ -374,7 +359,9 @@ protected void reloadElementsList(String searchText) {
     }
 
     private List<QuestInfo> getSortedQuests() {
-        return miniQuestMode ? Models.Quest.getMiniQuests(contentSortOrder) : Models.Quest.getQuests(contentSortOrder);
+        List<QuestInfo> quests = showQuests ? Models.Quest.getQuests(contentSortOrder) : List.of();
```
magicus: This will always sort quests before mini-quests. Is that really what you want? If you want to sort by distance, and has both quest and mini-quest toggle enabled, then I'd assume you wanted to see the closest content, regardless of type.

common/src/main/java/com/wynntils/screens/questbook/WynntilsQuestBookScreen.java
```diff
@@ -432,4 +406,59 @@ public void setContentSortOrder(ContentSortOrder newSortOrder) {
         setQuests(getSortedQuests());
         this.setCurrentPage(0);
     }
+
+    private void addQuestProgressTooltipLines(List<Component> tooltipLines, boolean miniQuestMode) {
```
magicus, line 407: Something very similar to this information is now available as tooltips from Wynncraft. See e.g. the call chain around DiscoveryModel.updateWorldDiscoveriesFromQuery() on how go extract it.

I'm not sure if it is a suitable replacement, but maybe. (Otoh, changing this can be the fodder of another PR)

common/src/main/java/com/wynntils/screens/questbook/widgets/QuestInfoButton.java
```diff
@@ -6,17 +6,13 @@
 
 import com.mojang.blaze3d.vertex.PoseStack;
 import com.wynntils.screens.base.widgets.WynntilsButton;
-import com.wynntils.screens.questbook.WynntilsQuestBookScreen;
 import com.wynntils.utils.render.RenderUtils;
 import com.wynntils.utils.render.Texture;
 import net.minecraft.network.chat.Component;
 
 public class QuestInfoButton extends WynntilsButton {
```
magicus, line 13: I think you should remove this entire class. With the filtering, it serves no purpose, and is just confusing.

kristofbolyai:
> The screenshot indicates that the quest/miniquest selection button in the top is still there. That looks a bit odd, should it not be removed?

It shows the progress tooltip you mentioned later.

common/src/main/java/com/wynntils/screens/questbook/WynntilsQuestBookScreen.java
kristofbolyai: If you don't mind, let's keep it like this now. (I also believe we could add some custom progress info with this code..)

common/src/main/java/com/wynntils/screens/questbook/widgets/QuestInfoButton.java
kristofbolyai: It still shows the progress tooltip, and we can't really render it any other way atm.

kristofbolyai: image

common/src/main/java/com/wynntils/screens/questbook/WynntilsQuestBookScreen.java
kristofbolyai: I wondered why it looked weird.. Fixed.

common/src/main/java/com/wynntils/screens/questbook/WynntilsQuestBookScreen.java
magicus: Yes, it was more a case of "oh, this reminds me of...".

common/src/main/java/com/wynntils/screens/questbook/widgets/QuestInfoButton.java
magicus: Ah, I see.

magicus (approved):

PR #1785, opened 2023-07-19
feat: Refactor the main Wynntils menu page

kristofbolyai:


kristofbolyai:
How it looks:

image

First row will soon have the other content button.
Second row could use an overlay group manager.
Third row could use a POI Provider button and a POI Management button.
Forth row is currently WIP and I am not fully sure how it will be laid out.

magicus:
The icons is really, really non-transparent. I actually had to go and check to understand what two of the icons meant (lootrun and quests, it turned out). And that the trophy icon means "discoveries" is just something I've learned after a long time.

But yeah, your grouping makes sense. As for the fourth row, maybe we can soon add an archetype/abilities gui button? (hint hint, nudge nudge ;-))

The one thing I thinking we might change is to put the second row at the bottom instead. That way we'd have like all "game content" at the first three rows, and then the "meta content" with the settings at the bottom.

magicus (approved):

PR #1788, opened 2023-07-20
 feat: Custom jump pages for block bank

ShadowCat117:
A few problems with this that I wanted to get your opinion on before I attempted them. But this does currently work ok.

Block bank has personal and island storage which all have a seperate page count, so you can have 12 pages of personal storage and then (presumably) 12 for each of your up to 5 islands. Currently, this just uses 1 set of custom destinations for all of them, however the problem then occurs that we only have 1 final page in storage, so the quick jumping on same pages that don't have as many may not work fully.

The solution would be to store 6 different final block bank page values and then also store each of the up to 5 island names, so we know which one's final page to use. This could then also lead into having a set of custom destinations for each island but I'm not entirely sure if that's a good idea so I didn't want to add that if you felt it was unnecessary and we can just use 1 set of custom destinations.

Let me know what would be the best option to go for and I'll have a crack at it.

This also enables searching in the block bank. (:


magicus:
It might be interesting to track which island the player is at (my thinking is to use the coordinates as a fast and invariant way to do this), but it will also require some work to start going with that.

I'm honestly not sure it is worth it. But then again, I am honestly not sure it is worth being able to set your own page destinations for the block bank at all.

In legacy, we could set personalized titles for bank pages. That is a piece of functionality that I'm sorely missing, at least for the normal bank. I can possibly imagine that if you sort your block bank and name the pages, then you might want to go be able to set destinations. 

So, I'm guess I'm saying:
1) I think this approach is fine for now, to treat all block banks the same

2) If you would like to have a stab at saving pages names for "bank" pages, I'd very much appreciate it!

3) To improve this, we should probably make a IslandModel, that can figure out what island you are on based on the coordinates. I assume there is a method in the madness, so we can derive a way to give islands an ID based on coordinates. I think we can get the name from the middle section of the info bar, iirc. If we can figure out if we are visiting someone else's island etc, that'd be nice too. We could e.g. publish that in the discord rich presence.

Both 2 and 3 are rather large undertakings.

JamieCallan117:
Yeah that sounds good. I was working on page names but I think someone else was too so I went back to cleaning up the original jump pages feature.

After I finish the translation feature I take a look into the IslandModel and try and get that working

ShadowCat117:
Yeah that sounds good. I didn't even know this had jump buttons until someone on the Discord asked for it. I was working on page names after I did the first PR for custom jump destinations but bolyai mentioned that someone else was working on it so I thought I'd let them do it and I'd work on something else.

After I finish the translation feature I'll take a look into the IslandModel and try and get that working

magicus:
I have not heard that anyone else is working on page names. If someone started, I guess the work has stalled. Given that you have been working with the area recently, I'd recommend that you start with that, rather than jumping into an IslandModel (unless you think that sounds more interersting, of course :-)).

magicus:
@kristofbolyai Do you know more details about this?

kristofbolyai:
I think @Mcrtin is doing work on it.

magicus (approved):
Can you just verify that there is no problem in that the mod tries to buy new pages for you, if you go to a different island where the final page is lower than the one you have recorded?

It's okay if it is a bit wonky, but not that it tries to click on the buy button.

If so, we can merge this PR.

kristofbolyai:
> Can you just verify that there is no problem in that the mod tries to buy new pages for you, if you go to a different island where the final page is lower than the one you have recorded?
> 
> It's okay if it is a bit wonky, but not that it tries to click on the buy button.
> 
> If so, we can merge this PR.

There is a hidden delay of a few seconds after going to the last page and the buy button enabling, from Wynn's side.

ShadowCat117:
Seems to work fine. I bought all 12 pages for personal storage and then quick jumping to page 2 on island storage doesn't try and buy it

PR #1794, opened 2023-07-21
fix: Parse discoveries faster with a "hack"

kristofbolyai:
I know this is a bit.. weird.. to say the least. But I think we can all agree this is worth the performance gain (we save looping around all filters 2 times)


magicus (approved):
+1

PR #1796, opened 2023-07-21
fix: Fix player list not rendering when overlay is disabled

kristofbolyai:


common/src/main/java/com/wynntils/features/overlays/CustomPlayerListFeature.java
```diff
@@ -46,7 +47,8 @@ public class CustomPlayerListFeature extends Feature {
 
     @SubscribeEvent
     public void onRender(RenderEvent.Pre event) {
-        if (event.getType() == RenderEvent.ElementType.PLAYER_TAB_LIST) {
+        if (event.getType() == RenderEvent.ElementType.PLAYER_TAB_LIST
```
magicus, line 50: That's interesting. Why do disabled features get events? 🤔 

That sounds like a bug to me. I think the feature itself should be able to rely on that it is enabled when receiving an event.

kristofbolyai: If you only disable the overlay...

magicus: Ah, I see... Ok, then this is fine.

magicus: (But maybe we should do something about the way we end up with "overlay-wrapping features"; somehow promote the overlay themselves to proper features?)

PR #1798, opened 2023-07-21
feat: Add cave support to Wynntils book

kristofbolyai:
This PR ended up as a mixed bag...:
- Refactored the content book pages to reduce duplication, and moved tooltip rendering in some places to a more sensible place (the widgets should contain their tooltips). There are many other places where this needs changing, but this is a start.
- Tracked content will always show up first in the content lists.

Overall, this ended up as a preparation for the last content book screen...


kristofbolyai:
image

kristofbolyai:
image
One could argue that the tooltip is a little long, but I would rather show info than hide it.. I propose to even do this in the quest book..

common/src/main/java/com/wynntils/screens/content/WynntilsCaveScreen.java
```diff
@@ -0,0 +1,216 @@
...
+    @SubscribeEvent
+    public void onTrackedContentUpdate(ContentTrackerUpdatedEvent event) {
+        // Reload so we have the proper order
+        if (McUtils.mc().screen == this) {
+            this.reloadElements();
+        }
+    }
+
+    // FIXME: We only need this hack to stop the screen from closing when tracking Caves.
+    //        Adding a proper way to add quests with scripted container queries would mean this can get removed.
```
magicus, line 67: Not sure I understand this..?

kristofbolyai: This is copied from the quest book, not a new invention. When you track a quest/cave, it will close the screen (because wynn closes the book after tracking).

magicus: But the comment? "Adding a proper way to add quests with scripted container queries would mean this can get removed." -- what is missing from the scripted container queries?

kristofbolyai: We don't care if the screen get's closed by the server, we just end our query, send a packet to close the screen and go on. We need to add a query step, something like "expectContainerClose".

common/src/main/java/com/wynntils/models/content/CaveModel.java
```diff
@@ -0,0 +1,97 @@
...
+    public CaveModel(ContentModel contentModel) {
+        super(List.of(contentModel));
+    }
+
+    public void reloadCaves() {
+        WynntilsMod.info("Requesting rescan of caves in Content Book");
+        Models.Content.scanContentBook(ContentType.CAVE, this::updateCavesFromQuery);
+    }
+
+    private void updateCavesFromQuery(List<ContentInfo> newContent, List<StyledText> progress) {
```
magicus, line 34: ahould we have a way to show the progress? we do for quests and discoveries...

kristofbolyai: If there is a progress item in the content book, I will add that tomorrow.

magicus: There is. All filters has a progress item. And the "All content" filter has a general overview of your character's progress. (We should probably parse this and display it nicely in the character view when we open the Wynntils book, i just realized)

common/src/main/java/com/wynntils/screens/content/WynntilsCaveScreen.java
magicus: Ah, now I see. When the server closes the container after selecting a content item for tracking, we still send a request to close the screen. https://github.com/Wynntils/Artemis/issues/1808

common/src/main/java/com/wynntils/models/content/CaveModel.java
magicus: You can probably add a cave-"button" on the top left page, similar to the quest "button" we discussed earlier. It's probably not the best interface (not very intuitive!) but at least it is consistent with what we have.

kristofbolyai:
image

magicus (approved):

PR #1800, opened 2023-07-22
feat: Add settings to make spell casting faster and more responsive

Mcrtin:
Add a setting to ignore the click cooldown when left-clicking in a spell. Make spells chain when holding. Complete spells with spells from the queue. Move spellSpellInProgress to spell model.
I mostly tested this on mage. When holding a spell hotkey and pressing another on the held key gets released. That's a feature from Minecraft (as far as I can see).


common/src/main/java/com/wynntils/models/spells/SpellModel.java
```diff
@@ -60,14 +64,21 @@ public void onItemRenamed(ItemRenamedEvent event) {
...
     @SubscribeEvent
     public void onSpellSegmentUpdate(SpellSegmentUpdateEvent e) {
         Matcher matcher = e.getMatcher();
         if (!matcher.matches()) return;
 
         SpellDirection[] spell = getSpellFromMatcher(e.getMatcher());
-        if (Arrays.equals(spell, lastSpell)) return; // Wynn sometimes sends duplicate packets, skip those
-        lastSpell = spell;
+        if (Duration.between(lastSpellUpdate, Instant.now()).toSeconds() < 3 && Arrays.equals(spell, lastSpell))
+            return; // Wynn sometimes sends duplicate packets, skip those
```
kristofbolyai: Comment is out-of-date and there is a multiline if without braces..

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -124,47 +114,61 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
+        spells.remove(spell);
+        spell = null;
+        if (--spellCountdown > 0) return false;
+        if (spells.isEmpty()) return false;
+        SpellDirection[] current = Models.Spell.getLastSpell().length != 3
+                        && Duration.between(Models.Spell.getLastSpellUpdate(), Instant.now())
+                                        .toSeconds()
+                                < 3
+                ? Models.Spell.getLastSpell()
+                : SpellDirection.NO_SPELL;
```
kristofbolyai: I can't understand what this is doing, and I think it is wrong.

Here is my understanding: `if (incorrectSpell) AND (recastTimeout) then getLastSpell else NO_SPELL`

This seems like that it was meant to be an `OR`? Also, perhaps extracting the two boolean statements like I did would clean up this code.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -124,47 +114,61 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
         if (slotChanged) McUtils.sendPacket(new ServerboundSetCarriedItemPacket(currSelectedSlot));
 
         // Waiting a few ticks is useful for avoiding lag related input-overlaps
-        if (!SPELL_PACKET_QUEUE.isEmpty()) packetCountdown = 3;
+        packetCountdown = Math.max(1, rightClickCooldown.get());
+    }
+
+    private boolean pollSpell() {
+        spells.remove(spell);
+        spell = null;
```
kristofbolyai: Overall this whole method seems really weird. You return early in multiple cases, but still do this before returning, every time..? It could also use a bit of spacing in logical places.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -49,55 +52,44 @@ public class QuickCastFeature extends Feature {
...
+        tryCastSpell(SpellUnit.PRIMARY, SpellUnit.SECONDARY);
     }
 
-    private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
-        if (!SPELL_PACKET_QUEUE.isEmpty()) {
-            sendCancelReason(Component.translatable("feature.wynntils.quickCast.anotherInProgress"));
-            return;
-        }
-
+    private void tryCastSpell(SpellUnit a, SpellUnit b) {
```
kristofbolyai: I don't appreciate that you removed the first part of the spells. This is really unclear to anyone reading this not knowing it's history. I think this method should take in the full spell combo, even if it is unused. I think @magicus agrees?

kristofbolyai (commented):
Before doing any of my reviews, can you tell me why this whole change is needed? As far as I am concerned, disabling safeCasting is a matter of wrapping the incompatibility check in a condition and changing the spell packet cooldown to 1/0.

Mcrtin:
the spell completion didn't work correctly that's why I had to do so many changes. I also initially thought this was an easy feature which is why I started this in the first place. I wanted to make it easier to spam spells (ie. mage tp) and reset spell cost with for example heal. Which didn't quite work (I don't remember exactly why it didn't).

common/src/main/java/com/wynntils/models/spells/SpellModel.java
Mcrtin: so it doesn't send duplicate packets anymore?

kristofbolyai:
> the spell completion didn't work correctly that's why I had to do so many changes. I also initially thought this was an easy feature which is why I started this in the first place. I wanted to make it easier to spam spells (ie. mage tp) and reset spell cost with for example heal. Which didn't quite work (I don't remember exactly why it didn't).

How did it not work correctly? Did it fail to realize a spell was in progress? Did it fail to ignore the already cast clicks? Even if it was bugged, I am sure it did not need a systematical change, like this PR.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
Mcrtin: `spell` is the currently active spell. `pollSpell` gets called once the spell is finished casting. than and only than can I remove the spell from the queue because I don't what It to get In the queue while its still active because when the user releases the appropriate spell button it would feel unresponsive when the spell casts again.

kristofbolyai: The problem is that you have to explain it to me. Now I see, but anyone reading this code later won't. Neither `spell` or `pollSpell` is something I remotely understand when reading, which is a problem, they need better names. As for the method, again, it is not something that is easy to understand when looking at/reading, which should be the goal of every method.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
Mcrtin: `if (isInProgress) AND (!hasWnnycraftDiscardedTheProgress) then getProgress else noProgress`

Mcrtin:
the issues with the old system:

1. slow
2. the spell completion didn't work
3. you couldn't schedule something for example heal when casting another spell/when holing another spell key

magicus:
I have not reviewed in detail, but I share bolyai's concern about this kind of sweeping changes in a very delicate system, that has required a lot of testing (and has had a lot of real-world testing) to be robust.

I recommend you put this particular PR on hold, and instead open a separate PR where you explore the road pointed to by bolyai, with much less intrusive changes. 

The approach must be to start there, and if that does not work, look at each separate problem that creeps up, how to best resolve it.

Mcrtin:
I haven't really changed any logic here. I mostly moved everything around to add another queue for spells. I actually did this like you said. I originally just wanted to make the casting system faster and then fixed every bug that came with the now possible applications.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
Mcrtin: is it fine now?

ryanzhoudev:
> 3. you couldn't schedule something for example heal when casting another spell/when holing another spell key

This is straight up illegal, both the scheduling and the key holding
https://discord.com/channels/394189072635133952/1037470382233501707/1132543945428193380

Mcrtin:
> I have not reviewed in detail, but I share bolyai's concern about this kind of sweeping changes in a very delicate system, that has required a lot of testing (and has had a lot of real-world testing) to be robust.
> 
> I recommend you put this particular PR on hold, and instead open a separate PR where you explore the road pointed to by bolyai, with much less intrusive changes.
> 
> The approach must be to start there, and if that does not work, look at each separate problem that creeps up, how to best resolve it.

now I don't know anymore if adding a buffer to the spells makes a difference or is even worth it, but I also don't want to start all over again.

magicus:
First of all, we need to figure out the legality of the feature request here. Is it even okay to make spell casting faster, in any way? And if it is, what is the limits?

The next step is discussing the actual implementation.

Mcrtin:
there is no limitation on how fast the spells may be cast. We just can't queue the user input to cast after another.
this: https://forums.wynncraft.com/threads/game-forum-rules.111874/
is the only official post I found where it says something about macros but Shadow has another post where it goes more into the specifics. That one stats that "Macros are only allowed to be used to input a single 3-click spell command at a time. You are not allowed to queue spells so they automatically cast after each other."
I fixed this pull request so it doesn't keep the user inputs when a spell gets cast.

ryanzhoudev:
> Is it even okay to make spell casting faster, in any way? And if it is, what is the limits?

> there is no limitation on how fast the spells may be cast.

There is a limitation. This is a bit of a common sense thing. The rules state that "Any modified client that gives you a gameplay advantage is hacking." and I guarantee you that sending clicks every tick (effectively 20 CPS) will be considered an advantage. At best, with no attention to accuracy at all, I can hit 12 CPS using both LMB and RMB.

Mcrtin:
To avoid that we can set a lower bound to the spell cooldown (something like 5 ticks maybe). Additionally, according to my tests right clicks only register every 2 ticks.
Also, if macros have a limitation on how fast they can cast, why is that not in the macros part of the rules?

Mcrtin:
20 cps is the maximum according to Shadow. because fast cast clicks multiple times per second I have to add a counter when fast cast is enabled

https://discord.com/channels/394189072635133952/1037470382233501707/1133276595092721697

ryanzhoudev:
That is so strange imo. Average and WR CPS for humans is ~7 and ~15 respectively, and most people can remember two spells in a row to cast. Yet the allowed feature is 20 CPS and queueing one spell is not.

Mcrtin:
I don't make the rules ^^ (but you can get 40+ cps with some clicking methods)

Mcrtin:
implemented in #1942

PR #1803, opened 2023-07-22
fix: Challenge rewards does not block mythics or favourites

ryanzhoudev:


kristofbolyai (approved):

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -217,7 +217,14 @@ public boolean isLootOrRewardChest(Screen screen) {
         if (!(screen instanceof AbstractContainerScreen<?>)) return false;
 
         String title = screen.getTitle().getString();
-        return isLootChest(title) || title.startsWith("Daily Rewards") || title.contains("Objective Rewards");
+        return isLootOrRewardChest(title);
+    }
+
+    public boolean isLootOrRewardChest(String title) {
```
magicus, line 229: Maybe make a `isRewardChest` instead, and keep "is loot || is reward"?

PR #1804, opened 2023-07-22
fix: Annotate lootrun tomes correctly

kristofbolyai:


common/src/main/java/com/wynntils/models/items/annotators/game/TomeAnnotator.java
```diff
@@ -17,7 +17,7 @@
 
 public final class TomeAnnotator implements ItemAnnotator {
     private static final Pattern TOME_PATTERN = Pattern.compile(
-            "^§[5abcdef]((?<Variant>[\\w']+)? ?Tome of (?<Type>\\w+)" + "(?:| Mastery (?<Tier>[IVX]{1,4})))$");
+            "§[5abcdef]((?<Variant>[\\w']+)? ?Tome of (?<Type>\\w+))(?:( Mastery( (?<Tier>[IVX]{1,4}))?))?");
```
magicus: Why did you remove ^ and $?

kristofbolyai: If we use .matches, I don't think there is a point for it?

magicus: Hm. Is this really guaranteed to be equivalent? Nevertheless, I have tried using ^...$ consistently, if not for anything else to show that this has been considered. I think it is a good practice. When testing on the online site, it is needed to get proper results. (In fact, I think we should setup a test suite for this expression, but you don't have to do it now if you don't want to.)

magicus (approved):

magicus:
Thanks!

PR #1807, opened 2023-07-22
fix: Fix chat time stamps making ranks appear gray

kristofbolyai:


ryanzhoudev (approved):
Works

kristofbolyai:
Actually, I might have typod `)`...

kristofbolyai:
It doesn't actually matter, but here is the most logical component layout..

PR #1811, opened 2023-07-23
refactor: Rename "Content" to "Activity"

magicus:
Have a look. What do you say? I think this is a definite improvement.


common/src/main/java/com/wynntils/models/activities/ContentBookQueries.java
```diff
@@ -272,15 +272,15 @@ protected void toggleTracking(String name, ContentType contentType) {
         query.executeQuery();
     }
 
-    private int findTrackedContent(ContainerContent container, String name, ContentType contentType) {
+    private int findTrackedContent(ContainerContent container, String name, ActivityType activityType) {
```
kristofbolyai: Content is actually rooted a bit deeper than just types.

common/src/main/java/com/wynntils/models/activities/ActivityModel.java
```diff
@@ -40,7 +40,7 @@
 import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
-public final class ContentModel extends Model {
+public final class ActivityModel extends Model {
```
kristofbolyai, line 50: Add a comment saying this is wynn's content book basically?

kristofbolyai (commented):
I only fear that this would be harder to find for new developers, who don't know this naming.

common/src/main/java/com/wynntils/models/activities/ContentBookQueries.java
magicus: This is sooo weird! I searched for "content" in IntelliJ, and it does not show this! I have checked again and again now, it shows this if you open in the Find window, but not in the normal list. I have not selected any case or complete word filters. It just ignores the entire ContentBookQueries.java file.

common/src/main/java/com/wynntils/models/activities/ActivityModel.java
magicus: Good point.

magicus:
> I only fear that this would be harder to find for new developers, who don't know this naming.

I agree that this is a potential problem. But then again, we have other very specific terms that people will need to learn, like "model" and "manager", so it is one of many.

magicus:
Note that I have not renamed features at the time. I am not sure if we should do that, but if so, we need to make config upfixers.

kristofbolyai (approved):

PR #1814, opened 2023-07-23
WIP: Create new map data provider architecture

magicus:


magicus:
I think the best way to understand this is to start by looking at the json file format. Here is an example:
https://github.com/magicus/Artemis/blob/map-data-rewrite/common/src/main/resources/assets/wynntils/mapdata.json

magicus:
This has become too messy to rebase. I will restart from fresh with the same code.

PR #1815, opened 2023-07-23
feat: Show 2.0.3 caves on the map

kristofbolyai:
image


common/src/main/resources/assets/wynntils/urls.json
```diff
@@ -92,6 +92,11 @@
     "md5": "f8d2b16d8cd33181b6a3857fa5e390a9",
     "url": "https://raw.githubusercontent.com/Wynntils/WynntilsWebsite-API/master/combatlocations.json"
   },
+  {
+    "id": "dataStaticCaveInfo",
+    "md5": "ac509a4e620ac65718a6b0816e848db5",
+    "url": "https://raw.githubusercontent.com/Wynntils/Reference/cave_data/content/cave.json"
```
kristofbolyai: Don't forget to use main here.

PR: https://github.com/Wynntils/Reference/pull/26

common/src/main/java/com/wynntils/models/map/PoiModel.java
```diff
@@ -78,11 +82,26 @@ private void loadData() {
...
+            Type type = new TypeToken<List<CaveProfile>>() {}.getType();
+
+            List<CaveProfile> profiles = WynntilsMod.GSON.fromJson(reader, type);
+
+            cavePois.addAll(profiles.stream()
+                    .map(profile ->
+                            new CombatPoi(PoiLocation.fromLocation(profile.location), profile.name, CombatKind.CAVES))
+                    .collect(Collectors.toUnmodifiableSet()));
+        });
+    }
```
kristofbolyai, line 103: Ended up loading this here since otherwise it would be really messy..

magicus (approved):

common/src/main/java/com/wynntils/models/map/PoiModel.java
magicus: I think that seems fine. Not sure what you should have done otherwise...

common/src/main/resources/assets/wynntils/urls.json
magicus: It is merged so you can update now.

magicus (approved):

PR #1817, opened 2023-07-23
Adding some parts from my old LootrunStatistics mod

McPlayHD:
In this first pull request I only added dry emeralds and item tiers. So for example how many Fabled items since a last mythic was found. I also add this data to the mythic find information.


McPlayHD:
wrong commits. sorry.

PR #1818, opened 2023-07-23
feat: Collect item type/emerald dry count

McPlayHD:
In this pull request, I only added counting emeralds and item types since the last mythic was found. For example how many Fabled you have found since the last mythic. This data also gets attached to the mythic find now.


common/src/main/java/com/wynntils/models/containers/LootChestModel.java
```diff
@@ -102,18 +109,42 @@ public void onSetSlot(ContainerSetSlotEvent.Post event) {
...
                                 new Location(lastChestPos)));
                 mythicFinds.touched();
 
                 dryBoxes.store(0);
                 dryCount.store(0);
+                dryEmeralds.store(0);
+                dryItemTiers.store(new EnumMap<>(GearTier.class));
             }
         } else {
             dryBoxes.store(dryBoxes.get() + 1);
```
magicus: If you moved the call to     progressItemFound() to here, you would not need to filter out mythics from there.

magicus: Ah, right, I see now that it is not that simple...

kristofbolyai:
I will review this tomorrow.

magicus (approved):

common/src/main/java/com/wynntils/models/containers/LootChestModel.java
```diff
@@ -36,6 +39,8 @@ public final class LootChestModel extends Model {
     private final Storage<Integer> openedChestCount = new Storage<>(0);
     private final Storage<Integer> dryCount = new Storage<>(0);
     private final Storage<Integer> dryBoxes = new Storage<>(0);
+    private Storage<Integer> dryEmeralds = new Storage<>(0);
+    private Storage<EnumMap<GearTier, Integer>> dryItemTiers = new Storage<>(new EnumMap<>(GearTier.class));
```
kristofbolyai: Without reviewing any further, this is wrong. EnumMaps are unsupported by storage, we only support HashMap, right @magicus? Also, can we just use the Map interface here?

kristofbolyai: (also this should be made more clear)

magicus: Yes, we should use `Map` in the left hand side. I am actually unsure what happens if we use an EnumMap... I think that depends on how Gson handles it. But HashMap sounds like a safer, and better, bet.

McPlayHD: Well it does seem to work for me:
```json
      "dryItemTiers": {
        "Normal": 5,
        "Unique": 9,
        "Rare": 4,
        "Legendary": 2
      },
```

McPlayHD: I looked at the source code of EnumMap and it looks like it's just an optimized variant of Map specifically for Enums. Basically, it does not calculate any hashes but uses the `Enum.ordinal()` integer number as the key.

magicus: Now that I have slept on it, this will do just fine. 

I remember what the issue was.

When reading from json, GSON will use the type declared (in this case EnumMap). So we want to use a suitable type.

However, when writing to json, Gson will use the type's iterator order. And that is just random for HashMap, so we get large arbitrarily discrepancys each time we save the json file.

That was why I went through the source code and replaced all HashMap and HashSet with TreeMap and TreeSet in configs.

I just checked EnumMap, as can be expected, it is maintained and iterated in enum ordinal order, so this is fine.

However, I think there have crept in other HashMaps in Storage and Config. We should probably add some check for that, if possible...

common/src/main/java/com/wynntils/models/containers/type/MythicFind.java
```diff
@@ -4,7 +4,16 @@
...
+import java.util.Map;
 
 public record MythicFind(
-        String itemName, int chestCount, int dryCount, int dryBoxes, long timestamp, Location chestCoordinate) {}
+        String itemName,
+        int chestCount,
+        int dryCount,
+        int dryBoxes,
+        int dryEmeralds, // Added in later patch
+        Map<GearTier, Integer> dryItemTiers, // Added in later patch
```
kristofbolyai, line 17: We need to keep track of these so we don't crash because of old data in the screen..

magicus (approved):

common/src/main/java/com/wynntils/models/containers/type/MythicFind.java
magicus: How do you mean? As a reminder to the future us?

kristofbolyai: Yes.

PR #1819, opened 2023-07-23
feat: Initial page setting for custom book

Rafii2198:
Replaces contentBookShouldOpenWynntilsMenu setting with InitialPage enum, players can now choose which activity they want to open, by default it opens the main page as opening quests was confusing for new players (they didn't know there was more to the book) as well as Content Book having more activities than quests made it a bit tedious if you didn't want to focus on quests


magicus (approved):
This will make users lose their current config, but I think this sounds ok in this case.

magicus:
Also, it would be great if we could have a keybind by default to open the Wynntils main menu (profile).

Rafii2198:
On Discord we had multiple conversations about setting it false, this will essentially do that and give more possibilities

Rafii2198:
We could change the K button which now opens Quests but as now there are more incentives to do other activities, changing it to main page would also make sense as it has most things that Content Book has, not only quests

magicus:
No, please don't touch K! I still think quests is the most valuable part of the content book.

What about assigning J to open the profile page? It's next to K, and afaik it is not currently bound to anything.

Rafii2198:
J is the guild map :( Maybe it could be . then? Or maybe Home button?

magicus:
Maybe "i"?

Rafii2198:
"i" could work. Should I change it in this pr or make a new one? Although I kinda like Home button as you know it is the home page of Wynntils, but it is your call.

magicus:
I'm on a mac laptop. There is no home button. ;-) I think "i" as in "information" sounds like a good mnemonic. But yeah, it's probably better to make it in a separate PR, even if it is just a single line.

magicus:
I'll wait for @kristofbolyai to approve that we skip the upfixer as well...

ryanzhoudev:
> J is the guild map :( Maybe it could be . then? Or maybe Home button?

Unrelated, but why is the guild map not L? The default advancements map is L at the moment, but we have hybrid data parsing, what would someone use the vanilla map for?

common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
```diff
@@ -88,7 +90,7 @@ public class WynntilsContentBookFeature extends Feature {
     public final Config<Boolean> replaceWynncraftContentBook = new Config<>(true);
 
     @RegisterConfig
-    public final Config<Boolean> contentBookShouldOpenWynntilsMenu = new Config<>(false);
+    public final Config<InitialPage> initialPage = new Config<>(InitialPage.USER_PROFILE);
```
kristofbolyai, line 93: But why did you change the default? By default, we used to open the quest book, now we open the user profile.

I don't mind not having the upfixer, but let's open the quest book by default.

kristofbolyai: Okay, I see, it was confusing. Well, it is confusing either way..

magicus:
I agree that it makes sense to replace the vanilla functionality, but I don't think we can change the vanilla keybinding, and the result would be that we have a keybinding conflict.

PR #1821, opened 2023-07-24
feat: LE stx in inventory has nicer symbol

ryanzhoudev:
image

Instead of looking identical to LE


common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -183,6 +184,23 @@ private void renderTexturedCount(PoseStack poseStack, int x, int y, int emeralds
                     .getItemRenderer()
                     .renderGuiItem(poseStack, EmeraldUnits.values()[i].getItemStack(), renderX + 6, renderY + 6);
 
+            if (EmeraldUnits.values()[i].getSymbol().equals("stx")) { // Make stx not look like normal LE
+                McUtils.mc()
```
magicus, line 188: Oh, I were assuming you created a new texture...

Well, this works, but I think a new texture would make more sense. It makes it available to future changes by an artist (whenever we get one to the team ;-)). And it keeps the code simpler.

magicus:
Just to be clear: I think the new symbol looks great! Just screenshot it and turn it into a png. :)

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -179,9 +180,15 @@ private void renderTexturedCount(PoseStack poseStack, int x, int y, int emeralds
                     Texture.EMERALD_COUNT_BACKGROUND.width(),
                     Texture.EMERALD_COUNT_BACKGROUND.height());
 
-            McUtils.mc()
-                    .getItemRenderer()
-                    .renderGuiItem(poseStack, EmeraldUnits.values()[i].getItemStack(), renderX + 6, renderY + 6);
+            if (EmeraldUnits.values()[i].getSymbol().equals("stx")) { // Make stx not look like normal LE
+                RenderUtils.drawTexturedRect(
```
magicus: Oh. I am kind of stupid here. We did not have textures for the other values, we just drew the vanilla items. 🤦 

I thought we would be able to unify stx with the other emerald denominations.

Eh... I am terribly sorry for giving you extra work, but can you please revert back to your original implementation? That was much better; I just did not read it properly. :-(

magicus:
I think you can do `git reset a0afa50` and then `git push --force` to achieve this.

magicus (approved):

PR #1822, opened 2023-07-24
feat: Introduce statistics system

magicus:
This is the start of a new system for tracking Wynncraft-related statistics. Currently, it only tracks two simple statistics (spells cast and total damage dealt), for showcasing.

Limitations:
* No UI is present
* No integration with Vanilla statistics (not sure if this is really something we want)

Also, this should not really be a manager, but it really highlights the need of a 3rd kind of components apart from models and managers, the "external models" (or whatever).


magicus:
This patch grew a bit since I opened the PR. Sorry for that. Now I am quite done, thank you very much.

The next steps is to create a GUI, and to add suitable statistics. That will have to be done in separate PRs.

common/src/main/java/com/wynntils/commands/StatisticsCommand.java
```diff
@@ -0,0 +1,146 @@
...
+    public LiteralArgumentBuilder<CommandSourceStack> getCommandBuilder(
+            LiteralArgumentBuilder<CommandSourceStack> base) {
+        return base.then(Commands.literal("show").executes(this::showStatistics))
+                .then(Commands.literal("list").executes(this::listStatistics))
+                .then(Commands.literal("get")
+                        .then(Commands.argument("statistic", StringArgumentType.greedyString())
+                                .suggests(STATISTIC_SUGGESTION_PROVIDER)
+                                .executes(this::getStatistic)))
+                .then(Commands.literal("reset")
+                        .then(Commands.literal("confirmed").executes(this::doResetStatistics))
```
kristofbolyai: maybe "confirm"?

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
```diff
@@ -0,0 +1,89 @@
...
+import com.wynntils.models.stats.type.DamageType;
+import com.wynntils.models.worlds.event.WorldStateEvent;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+// This should really be an "ExternalModel"...
+public final class StatisticsManager extends Manager {
```
kristofbolyai: The other "this really should be an external model" classes are all modes, what's different here?

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @SubscribeEvent
+    public void onWorldChange(WorldStateEvent event) {
+        // If we do not have a proper character, set up a fake statistics map so we always
+        // have a valid map.
+        if (!Models.Character.hasCharacter()) {
+            currentStatistics = new EnumMap<>(StatisticKind.class);
+            return;
+        }
```
kristofbolyai: Is this just a crash safety measure?

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    public void increaseStatistics(StatisticKind kind) {
+        addToStatistics(kind, 1);
+    }
+
+    public void addToStatistics(StatisticKind kind, int amount) {
+        currentStatistics.put(kind, currentStatistics.getOrDefault(kind, 0) + amount);
+        statistics.touched();
+    }
```
kristofbolyai: Before doing much, we might have to introduce types to statistics. We currently only have "counters", but what about "averages", or any other special type?

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
```diff
@@ -0,0 +1,89 @@
...
+        int neutralDamage = event.getDamages().getOrDefault(DamageType.ALL, 0);
+        addToStatistics(StatisticKind.DAMAGE_DEALT, neutralDamage);
+    }
+
+    @SubscribeEvent
+    public void onSpellEvent(SpellEvent.Completed event) {
+        increaseStatistics(StatisticKind.SPELLS_CAST);
+    }
+
+    // endregion
```
kristofbolyai: These should be in their own files, like functions are, I think.

kristofbolyai (commented):
My reviews are more of questions/improvement ideas, but I think we should deal with this before it becomes a pain to do later.

kristofbolyai (commented):
My reviews are more of questions/improvement ideas, but I think we should deal with this before it becomes a pain to do later.

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
magicus: Yep, I tried to explain that in the comment but I can try to be more clear about it

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
magicus: So, for better or for worse, I modelled this after the Vanilla statistics. I have some vague idea of exporting these statistics to Vanilla as well (though I am not sure of the benefits, but perhaps there are other mods that monitor statistics and can make, well, statistics out of them, showing changes over time etc, which was requested in Discord).

Also, I think Vanilla has had this system for quite a long time, and I think we can learn from them.

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
magicus: Yes, probably. But that means setting up them as event listeners, and I could not be bothered to do that right now. I hope it is okay. After all, this PR started out as a very minimalistic implementation.

I am also a bit on the fence whether we should require all statistics to go through events, or if we should be allowed to call `Managers.Statistics.increaseStatistic()` & co directly from the code.

common/src/main/java/com/wynntils/core/statistics/StatisticsManager.java
magicus: So, my initial idea was that we should sprinkle `Managers.Statistics.increaseStatistic()` all around the code base, and I did not want models to be dependent on another model, but a manager would be fine.

But then when I implemented the two initial statistics, it turned out that they were just listening to events. And maybe we should require that this is the only way statistics is allowed to be used, that if we want to have statistics, we need to send an event, which can then be listened to by the statistics manager/model.

I am worried that this will raise the bar too much and make it cumbersome to add new statistics, instead of just a single line of code. But otoh, it might also increase code quality, since there might be other potential listeners for that event. And having a "listener only" interface into the statistics would make it more self-sustaining, and allow it to be a model.

I could go either way here, so I'm interested in your opinion.

kristofbolyai:
To answer the 2 most relevant issues:

1, Should we use an event / inline statistics calls? 
I think we need should use events, they make the code so much cleaner. After a few of these, we might be able to come up with an alternative solution that has the cleanness benefit of events, but is easier to work with. (So let's move to a model if you agree.)

2, I am still very keen on having different types of statistics like counters, averages, max, min values. I can think of a "highest/max dry streak" stat, an "average time spent in a lootrun", etc stats. I think the vanilla system is very limited and obsolete in some sense. We can do better. I would be happy to work on this part of the statistics system, if you agree, and you don't want to spend time on (like you would rather focus on the map rewrite, or anything else).

kristofbolyai:
To extend 2.

After all, maybe we could even hook statistics into the function system. Having a function be able to get statistics data is just a new and improved way of having a function for everything. However this needs to be designed carefully.

3. I also realize that there are actually stats that could be both a min, max, or an average statistics, and we could show all of this. 

4. We might need some kind of "external statistics provider". For example, lootrun model already has "statistics" for a few stuff, do we migrate it? If so, how? Do you keep it in the lootrun model, and provide it to the statistics model? If so, we kind of break the "we only set stats by events" abstraction..

magicus:
Alright, let's go with the "require event" approach. I made it a Model in response.

It sounds like you have a lot of ideas on extending this to not only monotonically increasing integers. I am slightly skeptical and am worried we might overengineer a simple but robust system, but that has never stopped us before, has it? ;-)

But I'm leaving that to you. I think this PR is enough to get the ball rolling.

common/src/main/java/com/wynntils/models/statistics/StatisticKind.java
```diff
@@ -0,0 +1,44 @@
...
+
+        return null;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public String getName() {
+        return I18n.get("statistics.wynntils." + id);
```
kristofbolyai: Just a small nit. Can we save ourselves from the trouble of renaming all of these when we decide to add descriptions?

```suggestion
        return I18n.get("statistics.wynntils." + id + ".name");
```

kristofbolyai: If you are done changing this (and the keys), you can merge this.

kristofbolyai (approved):

kristofbolyai (approved):

kristofbolyai:
Thanks!

PR #1824, opened 2023-07-24
fix: Damage detection was broken, and missed poison

magicus:
This patch solves two problems:

1) Sometimes Wynn inserts a `§l` after the color code in damage labels

2) We missed poison damage completely. I added this as a damage type, but made sure it is not mixed in with normal stats.

### DIFF
--- a/common/src/main/java/com/wynntils/models/stats/type/DamageType.java
+++ b/common/src/main/java/com/wynntils/models/stats/type/DamageType.java
@@ -5,6 +5,8 @@
 package com.wynntils.models.stats.type;
 
 import com.wynntils.models.elements.type.Element;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/models/stats/type/DamageType.java
```diff
@@ -52,6 +55,10 @@ public enum DamageType {
         this.colorCode = element.getColorCode();
     }
 
+    public static List<DamageType> statValues() {
+        return Arrays.stream(values()).filter(d -> d == POISON).toList();
```
kristofbolyai: d != Poison?

magicus: OOOOPPSIE!!! 😊 

Goodness me. Thanks for checking on me! :)

kristofbolyai (approved):

PR #1191, opened 2023-02-22
New Crowdin updates

WynntilsBot:

PR #1828, opened 2023-07-24
chore: Automatically delete empty language files

magicus:
These are produced by CrowdIn. With this patch, we will automatically remove such files, so we can start accepting CrowdIn PRs.

Testing new build.yml changes is always tricky, but I have added this to the main branch of my personal fork and verified that it works.


magicus:
This is a prerequisite for merging #1191

.github/workflows/build.yml
```diff
@@ -36,6 +36,9 @@ jobs:
         with:
           arguments: spotlessApply
 
+      - name: Delete empty files
+        run: find $(git ls-files -m) -size 0 -delete
```
kristofbolyai, line 40: Shouldnt this be run before spotless?

DevScyu:
Files are not empty, they contain `{}`, will the check work for this?

.github/workflows/build.yml
magicus: No. The files are present but contains only `{}`. The spotless rule will delete this, but cannot remove the file. This will find any empty files present after the spotless run, and delete those.

Not the most elegant solution, but it works.

magicus:
> Files are not empty, they contain `{}`, will the check work for this?

The spotless rule will remove the `{}`.

I have verified that this fix does indeed work.

kristofbolyai (approved):

magicus:
@DevScyu Seems that you need to approve this as well...

DevScyu (approved):

PR #1829, opened 2023-07-24
refactor: Move away from old quest enums

kristofbolyai:


magicus (approved):
Thanks for the cleanup!

Issue #1809, opened 2023-07-23
Parse overall progress and show in character overview screen


magicus:
We should parse the general overview of your character's progress from the progress tooltip from the "All content" filter in the Content book, and display it nicely in the character view when we open the Wynntils book.

magicus:
Actually, #1833 only provides a partial implementation. We should also add the tooltip; which hopefully is easier now with the new tooltip code in 1.20.

PR #1833, opened 2023-07-24
feat: Show overall content activity progress on profile page

magicus:
Solves #1809, at least partially. My initial thought was to show the full tooltip when hovering over the progress text, but unfortunately hooking in tooltips to rendered text was not trivial, so I gave up on that for the time being.

Here is how it looks:
Screenshot 2023-07-25 at 00 54 44


magicus:
(Also, we need to wrap long splashes!)

common/src/main/java/com/wynntils/models/activities/ContentBookQueries.java
```diff
@@ -57,7 +57,8 @@ public class ContentBookQueries {
     protected void queryContentBook(
             ActivityType activityType,
             BiConsumer<List<ActivityInfo>, List<StyledText>> processResult,
-            boolean showUpdates) {
+            boolean showUpdates,
+            boolean includeAllPages) {
```
kristofbolyai: firstPageOnly is clearer, not all pages can still be multiple pages..

kristofbolyai (approved):

PR #1838, opened 2023-07-25
feat: Right-click text field to clear it

Mcrtin:
Add a feature so you can right-click a `TextInputBoxWidget` to clear its contents.


magicus:
I see the need to make it easy to clear a text field, but this does not seem very discoverable, nor intuitive. 

I think a better and more traditional approach would be to render a "clear icon" (a cross in a circle) at the very end of the input field.

Mcrtin:
I think JEI does it like this which is why I tried right-clicking the search field and discovered that we don't have this feature.

magicus (approved):
After some discussion with other Wynntils developers, I'm prepared to accept this. If/when we implement a clear button, we'll reconsider this.

PR #1839, opened 2023-07-25
refactor: Turn WynnPlayerUtils into PlayerRank enum

magicus:


kristofbolyai (approved):
Nice refactor. Did you start fixing the regex for this..?

magicus:
No..? I just looked through utils to see if there were anything that perhaps should be moved to services.

magicus:
I have been trying to keep a strict regime to keep "model" stuff out of utils (I have a thorn in the side for InventoryUtils and WynnItemMatchers, and some day I'm gonna fix them), but services has had no real place to go, so I thought there would be such. I did not find any, but I found this.

kristofbolyai:
> No..? I just looked through utils to see if there were anything that perhaps should be moved to services.

Oh.. This feature is broken atm pending a regex fix for 2.0.3, I thought you found this by fixing it...

PR #1841, opened 2023-07-25
fix: Fix ContainerSearchFeature searching in the sidebar

### LINKED ISSUE #1691: bank search also searches in the right most column of the bank
image

Mcrtin:
When typing something like "page 2" in the search bar of your bank it also highlights the page 2 button as mentioned in #1691.


common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -215,22 +215,24 @@ private void addSearchWidget(AbstractContainerScreen<?> screen, int renderX, int
...
-                    .getStringWithoutFormatting()
-                    .toLowerCase(Locale.ROOT);
-
-            boolean filtered = !search.isEmpty() && name.contains(search) && itemStack.getItem() != Items.AIR;
-            wynnItemOpt.get().getCache().store(WynnItemCache.SEARCHED_KEY, filtered);
-            if (filtered) {
-                autoSearching = false;
-            }
-        }
+        System.out.println(screen.getMenu().slots);
```
kristofbolyai: debug leftover

magicus: debug code

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -215,22 +215,24 @@ private void addSearchWidget(AbstractContainerScreen<?> screen, int renderX, int
...
+        screen.getMenu().slots.stream()
+                .filter(slot -> slot.x <= 6 * 18 + 8 && slot.y <= 6 * 18)
```
magicus: have you verified that this is correct for misc bucket, block bank etc?

magicus: I think it could be generally useful to identify the "content area" of a container. We are doing something similar in ContentBookQueries.processContentBookPage.

If you are up to the challenge, I think it would be good to try and generalize this into ContainerModel, to have some way of defining which slots are "game" slots, and which are "ui" slots, depending on which kind of container we have.

Mcrtin: I think it would be better if we do something like the ItemAnnotator for containers. So have for every container (ie. bank, misc. bucket etc.) a class

Mcrtin: > have you verified that this is correct for misc bucket, block bank etc?

I forgot that it not only searches in your bank -.-
Im not in a guild so I can't verify that it works in the guild bank/member list even if I implement a solution

magicus: But you can do the rest of the searchable containers, and then someone with guild access can test the guild stuff

magicus: Also, I agree that we need a more comprehensive container model. It was an interesting idea to use the ItemAnnotator approach, I'll need to ponder that one for a while.

Nevertheless, that is going to require some effort to transform the current state of the code to a better model. But we will help that future change by not having magic numbers or special dances like this spread all over the place. So at the very least, create a function that e.g. streams the valid slot numbers for your bank account in ContainerModel.

Mcrtin: I can add a function parameter to the `SearchableContainerType` enum

Mcrtin: the search in the guild bank should now also work but someone should text that

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerType.java
```diff
@@ -51,4 +69,11 @@ public static SearchableContainerType getContainerType(StyledText title) {
 
         return null;
     }
+
+    public Stream<ItemStack> getSearchableItems(AbstractContainerScreen<?> screen) {
+        return screen.getMenu().slots.stream()
+                // see ChestMenu
+                .filter(slot -> isSearchableSlot.test((slot.x - 8) / 18, (slot.y - 18) / 18))
```
magicus: This looks a bit odd, but maybe the slot id was not a good way of determining if it was searchable or not?

magicus (approved):
I think it looks good. Perhaps a comment about the choice of using x/y coordinates.

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerType.java
kristofbolyai: I honestly have no idea what the -8 does, as well as 18 (2*9, 2 rows)?

Mcrtin: The x and y components of Slot are in screen coordinates relative to the upper left corner of the container screen. You can see where these numbers come from in ChestMenu

kristofbolyai: And why do we compare to rendered coordinates in logic that is not about rendering? Refactor this class to use `Slot.index`, otherwise it is really sketch.

kristofbolyai: (Otherwise this will break really easy if a mod, or a MC update moves slots by 1 pixel..)

magicus: I agree with bolyai that using screen coordinates is not the best way to do this. We really want to use the slot index id.

However, in my experience, it is not obvious how to express "3rd row, 2nd column" as a slot id, or contrary, what position in the container slot id 13 means.

But I think the solution to this is instead to create a function like `getSlotId(int row, int column)` in `InventoryUtils`, and use it here.

magicus:
There is something nagging me about this solution. I cannot pinpoint any obvious errors, but please hold on integrating this until I can think about it.

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerType.java
```diff
@@ -5,33 +5,59 @@
...
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import net.minecraft.world.inventory.ChestMenu;
+import net.minecraft.world.item.ItemStack;
 
 public enum SearchableContainerType {
-    BANK(Pattern.compile("§0\\[Pg. (\\d+)\\] §8(.*)'s?§0 Bank"), Pattern.compile("§f§lPage \\d+§a >§2>§a>§2>§a>"), 8),
+    BANK(
+            java.util.regex.Pattern.compile("§0\\[Pg. (\\d+)\\] §8(.*)'s?§0 Bank"),
```
ShadowCat117:
```suggestion
            Pattern.compile("§0\\[Pg. (\\d+)\\] §8(.*)'s?§0 Bank"),
```
Same for the others :)

kristofbolyai: Good catch!

Mcrtin: i don't even know how that got there xD

magicus (commented):
Withdrawing approval for now

magicus (changes requested):
So how should you really withdraw your approval in github..? I don't understand...

ryanzhoudev:
> So how should you really withdraw your approval in github..? I don't understand...

image

This button

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerType.java
```diff
@@ -5,33 +5,59 @@
...
+import net.minecraft.world.inventory.ChestMenu;
+import net.minecraft.world.item.ItemStack;
 
 public enum SearchableContainerType {
-    BANK(Pattern.compile("§0\\[Pg. (\\d+)\\] §8(.*)'s?§0 Bank"), Pattern.compile("§f§lPage \\d+§a >§2>§a>§2>§a>"), 8),
+    BANK(
+            Pattern.compile("§0\\[Pg. (\\d+)\\] §8(.*)'s?§0 Bank"),
+            Pattern.compile("§f§lPage \\d+§a >§2>§a>§2>§a>"),
+            8,
+            slotPos -> slotPos.x() < 7 && slotPos.y() < 6),
```
magicus: It looks like the test methods are very similar. You can extract them to two helper methods, like `testBankStyleSlot` and `testGuildStyleSlot`.

common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
```diff
@@ -69,4 +72,16 @@ public static void clickOnSlot(int clickedSlot, int containerId, int mouseButton
     public static void closeContainer(int containerId) {
         McUtils.sendPacket(new ServerboundContainerClosePacket(containerId));
     }
+
+    public static Stream<SlotPos> getSlots(ChestMenu chestMenu) {
+        return chestMenu.slots.stream().map(SlotPos::of);
+    }
+
+    public record SlotPos(ItemStack itemStack, int x, int y) {
```
magicus: It just feels odd to have an ItemStack here. Look at how Vanilla implements `Slot.getItem()`:
`return this.container.getItem(this.slot);`

You should do the same. Construct this with a container instead. I would also recommend renaming x and y to "row" and "column" instead, since it is not about graphic coordinates.

magicus: Actually, the entire type seems wrong. All you really need is to convert a slot.index to row and column values. You can send in a Slot to the filter predicate, and have ContainerUtils methods "getSlotRow" and "getSlotColumn" to extract these values using % and / 9.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -212,14 +218,21 @@ private void addSearchWidget(AbstractContainerScreen<?> screen, int renderX, int
         screen.addRenderableWidget(lastSearchWidget);
     }
 
-    private void matchItems(String searchStr, AbstractContainerScreen<?> screen) {
```
magicus, line 215: Do you really have to change the signature of this method? You could just as well just do `screen.getMenu()` inside it, right?

Mcrtin: because im doing `%9` it makes more sense to "guarantee" a ChestMenu although the change is not necessary. Do you think it's better to revert it?

magicus: I think it is in more or less always preferable to change as little as possible when fixing a bug, unless that leads to code of too low quality.

Minimal impact of changes is defintely a quality of its own, worthy to pursue in PRs.

So, yes.

Mcrtin: I'm using ChestMenu#getContainer to only get the items of the actual container which an exclusive method of ChestMenu so all these changes are necessary.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -75,7 +76,7 @@ public class ContainerSearchFeature extends Feature {
     @SubscribeEvent
     public void onScreenInit(ScreenInitEvent event) {
         if (!(event.getScreen() instanceof AbstractContainerScreen<?> screen)) return;
-
+        if (!(screen.getMenu() instanceof ChestMenu chestMenu)) return;
```
magicus, line 79: Please keep the empty separator line between the "if this and that then return early" and the real code.

Mcrtin: can you add that to Spotless?

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -87,7 +89,7 @@ public void onScreenInit(ScreenInitEvent event) {
 
         currentSearchableContainerType = searchableContainerType;
 
-        addSearchWidget(screen, renderX, renderY);
+        addSearchWidget(((AbstractContainerScreen<ChestMenu>) screen), renderX, renderY);
```
magicus, line 92: Can you replace ? with ChestMenu in line 78, so you can skip this cast? (Maybe that is impossible, the type erasure is hard to understand, but I think it should work)

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -212,14 +219,21 @@ private void addSearchWidget(AbstractContainerScreen<?> screen, int renderX, int
...
 
-    private void matchItems(String searchStr, AbstractContainerScreen<?> screen) {
+    private void matchItems(String searchStr, ChestMenu chestMenu) {
         String search = searchStr.toLowerCase(Locale.ROOT);
 
-        NonNullList<ItemStack> playerItems = McUtils.inventory().items;
-        for (ItemStack itemStack : screen.getMenu().getItems()) {
+        Container container = chestMenu.getContainer();
+        for (int i = 0; i < container.getContainerSize(); i++) {
+            if ((currentSearchableContainerType == SearchableContainerType.GUILD_BANK
```
magicus, line 227: Somewhere along the line you lost the code that had the "content" part of the container encoded as part of each individual SearchableContainerType enum value. I think that was a good design idea. The implementation you had at that time was lacking somewhat (it was overly complex with a new type which was not needed) but the core of the idea was sound, I think, to let the enums know if an index was inside the "content" area or not.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
Mcrtin: 'Screen' cannot be safely cast to 'AbstractContainerScreen<ChestMenu>'

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
Mcrtin: I don't think that was necessary, with only two values and I also should keep the change to a minimum, but I hope we can someday change it when we have a better system to handle Wynncrafts custom UIs

magicus (approved):

magicus (approved):
Okay. I don't think we're getting any further here now.

PR #1845, opened 2023-07-25
feat: Group overlay settings with the actual overlays, and restructure overlays

magicus:
Title gives the user visible changes. But the driving force behind this PR is a heavy refactoring, where I have moved all event listeners (and, where applicable, variables, constants and inner types) into the scope of Overlays.


magicus:
I have verified all overlays that I could (guilds stuff and some archetype things were omitted).

kristofbolyai:
I will review this tomorrow. Is there a need for upfixers, if so, are they working?

magicus:
No, there is no need for upfixers. I have deliberately not touched configs.

kristofbolyai:
> No, there is no need for upfixers. I have deliberately not touched configs.

You renamed features holding overlays, how do you not need upfixers?

magicus:
Oh, right, the overlays have inherent configs. I forgot about those. :-(

Oh well. If I do need an upfixer, then I might just as well be a bit more aggressive with the renaming. The resulting features did not have as well suited names anymore, but I did not want to rename so I could skip doing upfixers. But if I got to do it anyway, then whatever...

magicus:
So, a summary of what I have done:

* Every Overlay now contain all config variables that pertain to it
* Most overlays are the sole content of the corresponding wrapping feature ( a few exceptions are when there are multiple overlays in a single feature, and the StopwatchFeature which is kind of ... different).
* In these cases, the feature name is consistently overlay name + `Feature`
* Where a feature previously had both an overlay and other functionality, I've split this in two, and given the non-overlay part better names, and placed them in "combat" (i.e gameplay)
* I believe I have made proper adjustment to i18n, and that I have provided upfixers for all changes. I will try to verify the latter.
* This has almost drove me nuts, so I'd appreciate if we can merge this before any conflicts arise. ;-)

magicus:
Verified that upfixers work correctly.

kristofbolyai (approved):
The refactors look good, but I have not verified every change you made (it is impossible mostly. I see you did some minor refactors, but they looked sane to me).

Testing this is almost impossible. If you think you covered everything, let's merge it. But we need to keep be around for a few hours to fix any arising problems. Otherwise, it needs to be delayed..

PR #1846, opened 2023-07-25
refactor: Move all subclasses of Overlay into top-level classes

magicus:
This depends on #1845.

Screenshot 2023-07-25 at 23 11 20

The overlays are still instantiated and registered in their respective holding feature. As you might guess, my intention is to be able to remove this middle step and make Overlays fully-fledged first class citizens. That will require a bit of rework with the feature/overlay framework, adaptations to the config screen, and some careful upfixers, so we do not lose settings. But this will at least take us a step in this direction.


kristofbolyai (approved):
Not much to review here, hopefully we don't miss some reflection details that render these unusable... Have you verified this doesn't cause any issues, config and usage wise? (TLDR just boot up the game..)

PR #1849, opened 2023-07-26
feat: Store when we started collecting a statistic entry

kristofbolyai:


kristofbolyai:
Old values just turn into 0, I'll handle it in the screens..

magicus:
Or we really, really need to have storage upfixers as well, so we could replace 0 with the current time at least.

Also, maybe we should store the last time the stat was updated as well?

kristofbolyai:
> Or we really, really need to have storage upfixers as well, so we could replace 0 with the current time at least.
> 
> 
> 
> Also, maybe we should store the last time the stat was updated as well?

We could also upfix mythic info then...
I am happy to add last modified timestamp field.

common/src/main/java/com/wynntils/services/statistics/type/StatisticEntry.java
```diff
@@ -4,11 +4,17 @@
  */
 package com.wynntils.services.statistics.type;
 
-public record StatisticEntry(int total, int count, int min, int max) {
-    public static final StatisticEntry EMPTY = new StatisticEntry(0, 0, 0, 0);
+public record StatisticEntry(int total, int count, int min, int max, long lastModified, long lastReset) {
```
magicus:
```suggestion
public record StatisticEntry(int total, int count, int min, int max, long lastModified, long firstModified) {
```

That the first modified field is reset when we reset statistics is kind of impied...

magicus: And possibly it might make more sense to switch the order around for those two

magicus (approved):

magicus (approved):

PR #1850, opened 2023-07-26
feat: Fix some bugs related to scoreboard parsing correctly by simplifying the handler

kristofbolyai:
This PR in theory does not change much related to logic, except that instead of relying on our system for constructing a scoreboard, we let MC handle that.

The part which I can see fixing the weird issues where "half-half" quest info is shown is that now we parse both the primary (SB) and backup (BF) scoreboard. These are both rendered by Wynn and are set to display multiple times during a scoreboard reconstruction to reduce visual glitches. Previously, we only cared about SB because it seemed like they are always in sync, after a while. Maybe they are not..


common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -56,41 +55,35 @@ public void addPart(ScoreboardPart scoreboardPart) {
...
-        String baseName = "sb" + McUtils.player().getScoreboardName();
+    private boolean isValidScoreboardName(String scoreboardName) {
+        // If the name is longer than 14 characters, we need to trim it (16 chars max, 2 reversed for sb/bf)
+        String name = McUtils.player().getScoreboardName();
+        name = name.length() > 14 ? name.substring(0, 14) : name;
 
-        // If the baseName is longer than 16 characters, we need to trim it
-        scoreboardNameCache = baseName.length() > 16 ? baseName.substring(0, 16) : baseName;
-        return scoreboardNameCache;
+        return (scoreboardName.startsWith("sb") || scoreboardName.startsWith("bf")) && scoreboardName.endsWith(name);
```
magicus, line 63: Is it okay that we only have a SCOREBOARD_KEY for SB? (Still not through understanding the logic, just noting the inconsistency)

kristofbolyai: Yes, scoreboard key is just a third scoreboard the mod creates to display. We don't do any switcheroos, so a static key is fine.

kristofbolyai: (Our "switcheroos" logic is handled by wynn, we just copy what they would display.)

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -100,80 +93,39 @@ public void onWorldStateChange(WorldStateEvent event) {
...
-                .filter(line -> line.line().equals(owner))
-                .findFirst()
-                .ifPresent(reconstructedScoreboard::remove);
+        return lines.stream()
+                .map(s -> new ScoreboardLine(StyledText.fromString(s.getOwner()), s.getScore()))
+                .toList();
     }
 
-    private boolean isScoreboardValid() {
+    private boolean isScoreboardValid(List<ScoreboardLine> reconstructedScoreboard) {
```
magicus, line 128: I can't comment on the right lines (thanks, github :-/) but the description below has some typos and strange grammar.

magicus (approved):
I think it is okay, but honestly the scoreboard handler has always been a bit too complex for me to fully get a grip on.

kristofbolyai:
@magicus can you try to test it like you did with the old one? I couldn't find any issues with this or the old one...

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -131,10 +131,11 @@ private boolean isScoreboardValid(List<ScoreboardLine> reconstructedScoreboard)
         // 2. There are no gaps in the scores, and they are decreasing (there are no duplicate scores)
 
         // We can also check for validness by checking scoreboard parts:
-        // 3. A valid scoreboard always starts with a new line (À)
-        // 4. A valid scoreboard if it consists of valid segments:
-        //    - A valid segment is a part that stats with a header, then one or more lines, then a footer which is (À+).
-        //    - The footer is not present if the segment is the last one.
+        // 3. A valid scoreboard always starts with a newline (À)
+        // 4. A scoreboard is valid if it consists of valid segments:
+        //    - A valid segment is a part that stats with a header, then one or more lines, then a footer which is a
```
magicus:
```suggestion
        //    - A valid segment is a part that starts with a header, then one or more lines, then a footer which is a
```

magicus:
I'll test this in a while, just need to get closure on my overlay fixes first.

magicus:
I'll have to test this tomorrow; it got too late.

magicus (approved):
I have thrown everything against this that I could think of, and none of it could bring this down. So I'm happy, in terms of user experience.

(I still want to fix so we only send a single event when the scoreboard segment is actually done, though)

PR #1852, opened 2023-07-26
feat: Collect lootrun statistics

kristofbolyai:
> But otoh, I think it is okay to add other stats first, so we have more than the measly 2 I created when we start to show anything. - @magicus 

I think this should do it :)


kristofbolyai:
We don't error for missing i18ns...

magicus:
Should we have a i18n assertion for statistics as well?

kristofbolyai:
> Should we have a i18n assertion for statistics as well?

Most likely yes, I missed them, others will as well.

magicus:
> Most likely yes, I missed them, others will as well.

Fixed in #1863

common/src/main/java/com/wynntils/screens/lootrun/WynntilsLootrunsScreen.java
magicus, line 1: This should be WynntilsLootrunPathsScreen, and the package should be lootrunpaths.

kristofbolyai: Fixed them

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -0,0 +1,147 @@
...
+    private LootrunFinishedEvent.Completed.Builder lootrunCompletedBuilder;
+    private LootrunFinishedEvent.Failed.Builder lootrunFailedBuilder;
+
+    public LootrunModel() {
+        super(List.of());
+    }
+
+    @SubscribeEvent
+    public void onChatMessage(ChatMessageReceivedEvent event) {
+        StyledText styledText = event.getOriginalStyledText();
```
magicus, line 62: You should check the recipientType so it is info.

common/src/main/java/com/wynntils/models/lootrun/event/LootrunFinishedEvent.java
```diff
@@ -0,0 +1,131 @@
...
+
+        public int getExperienceGained() {
+            return experienceGained;
+        }
+
+        public int getMobsKilled() {
+            return mobsKilled;
+        }
+
+        public static class Builder {
```
magicus: The builder idea was nice, but I think you should move them to the model. We have always let the event classes be very lean and not contain anything apart from final fields and getters. It is the responsibility of the model to collect all data needed to be able to construct and send out an event.

magicus (commented):
Looks good apart from the builder placement.

common/src/main/java/com/wynntils/models/lootrun/event/LootrunFinishedEvent.java
kristofbolyai: I moved it to a new file, I don't like it being in the model.

magicus (approved):

PR #1857, opened 2023-07-27
fix: Don't overwrite default bank jump behavior

P0keDev:
This fixes a bug where bank page jump buttons might not function properly despite being the vanilla value

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
@@ -121,12 +121,17 @@ public class CustomBankPagesFeature extends Feature {
         customJumpDestinations = parseStringToDestinations(configDestinations, currentContainer);
 
         if (customJumpDestinations == null) {
-            switch (currentContainer) {
-                case BANK -> customJumpDestinations = QUICK_JUMP_DESTINATIONS;
-                case BLOCK_BANK -> customJumpDestinations = BLOCK_BANK_DESTINATIONS;
-                case BOOKSHELF, MISC_BUCKET -> customJumpDestinations = HOUSING_DEFAULT_DESTINATIONS;
-            }
+            customJumpDestinations = getDefaultJumpDestinations();
         }
+    }
+
+    private List<Integer> getDefaultJumpDestinations() {
+        switch (currentContainer) {
+            case BANK -> return QUICK_JUMP_DESTINATIONS;
+            case BLOCK_BANK -> return BLOCK_BANK_DESTINATIONS;
+            case BOOKSHELF, MISC_BUCKET -> return HOUSING_DEFAULT_DESTINATIONS;
+        }
+        return HOUSING_DEFAULT_DESTINATIONS; // use as default for safety
     }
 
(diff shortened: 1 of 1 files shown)


magicus:
I can accept this, since it looks safe and conservative, but I am a bit curious as to why this does not work properly.

My understanding is that the system should decide which page you try to go to (which should be the vanilla default in your use case), and then it should figure out the shortest way to get there, by starting at looking if a vanilla default page is nearby the target page. In this case it is, so it should just press the vanilla "go to page" button, and you should have the same behavior.

magicus:
So maybe there is something broken in the jumping logic.

kristofbolyai:
> So maybe there is something broken in the jumping logic.

(I have the same understanding of this.)

P0keDev:
The issue is that the jumping logic checks the lastPage variable and avoids clicking on any vanilla buttons leading to "unconfirmed" pages.

I must admit that I was (and am) quite sleep-deprived when writing this fix and didn't realize that the proper solution is to make the lastPage tracking persistent, instead of resetting it every session. That behavior would match the logic I designed in legacy. I can try rewriting this tomorrow for a more complete fix

ShadowCat117:
> The issue is that the jumping logic checks the lastPage variable and avoids clicking on any vanilla buttons leading to "unconfirmed" pages.
> 
> I must admit that I was (and am) quite sleep-deprived when writing this fix and didn't realize that the proper solution is to make the lastPage tracking persistent, instead of resetting it every session. That behavior would match the logic I designed in legacy. I can try rewriting this tomorrow for a more complete fix

I made it so that the default lastPage for each container is the highest number of pages you can have in that container type, then when you access your final page, the storage value then gets updated to that so it shouldn't be reset every session?

ShadowCat117:
There might be something wrong in `tryUsingJumpButtons` but from my testing, it did seem to jump to vanilla destinations fine. I've not been able to test if it jumps to the final page as I don't own it so maybe there'd be an issue there

P0keDev:
Ah, my mistake, I think I skimmed the code relating to lastPage too quickly

The issue I've encountered multiple times involves the page  13/17/21 buttons - maybe the code is checking the wrong container's page limits somehow?

ShadowCat117:
Hmm that's strange. Are you able to check your storage value for final bank page to make sure it's not gotten lower than it should by accident, if that's the case then might need to have a greater than check when updating it

ShadowCat117:
```
int defaultDestination = getDefaultJumpDestinations().get(buttonIndex);
if (pageDestination != defaultDestination) {
    e.setCanceled(true);
    jumpToDestination();
}
```

I think this is worth adding anyway as it saves doing all the math for finding closest destination

magicus:
I will merge this as-is, since it is a safe addition in any case.

@JamieCallan117 Will you continue working with getting to the bottom of the problem? You can open an issue to track this, if you want.

magicus (approved):

ShadowCat117:
> @JamieCallan117 Will you continue working with getting to the bottom of the problem? You can open an issue to track this, if you want.

Sure, I was planning on revisting my additions to do some improvements next anyway so I'll also look at this

PR #1858, opened 2023-07-27
feat: Add WynntilsStatisticsScreen

kristofbolyai:
Depends on #1852

image
image
image
image


magicus:
Some comments from just looking at the screenshots:

1) Maybe add the total to the tooltip? That is kind of a compromise of what I wanted and what is possible to do with the format. And you still would not have to click on every statistic to see the value.

2) We have `StringUtils.formatAmount()` which would make like `6.1m` of `6129321`. Maybe use that instead, and only display the complete number if you press down shift, or select a toggle.

kristofbolyai:
> We have StringUtils.formatAmount() which would make like 6.1m of 6129321. Maybe use that instead, and only display the complete number if you press down shift, or select a toggle.

In the tooltip?

magicus:
> In the tooltip?

No, not only, also at the side page. At some point it is just hard for humans to read and understand large numbers. Compare:

Count: 78.2 k

with

Count: 78177

kristofbolyai:
> > In the tooltip?
> 
> 
> 
> No, not only, also at the side page. At some point it is just hard for humans to read and understand large numbers. Compare:
> 
> 
> 
> Count: 78.2 k
> 
> 
> 
> with
> 
> 
> 
> Count: 78177
> 
> 

But the point of statistics is to get accurate info?

kristofbolyai:
image

common/src/main/java/com/wynntils/services/statistics/CustomStatFormatters.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.statistics;
+
+import net.minecraft.stats.StatFormatter;
+
+public final class CustomStatFormatters {
```
magicus, line 9:
```suggestion
public final class CustomStatFormatters implements StatFormatter {
```

?

magicus: No, sorry, that's not how any of this work! 🤦

kristofbolyai: I am not sure how this helps? We don't create our own interface/abstract class, we just needed a class to hold our custom formatters.

kristofbolyai:
I did not implement `StringUtils.formatAmount()` because it needs a new formatter to work properly. We should also make a way to hide certain stat fields where it doesnt matter (like spells cast). That needs to be a follow up.

magicus:
> But the point of statistics is to get accurate info?

Yeah, but if you have a value in the millions, small differences is not relevant. I mean, you display the total lootrun time as "3.25 h", even if you could have said 12300 seconds. Or maybe 12344 seconds; we don't know since you lose accuracy. But that would not be very helpful.

Once again, I am all in favor of making it possible to see the exact details (and in that case, maybe also seeing every second for time based entries), but this should not be the default, but only if you e.g. hold down shift. Or, we should have a simple toggle button for "compact" or "full" representation.

Basically, I am arguing that by default we should pass the value through `StringUtils.integerToShortString()`.

kristofbolyai:
Is it acceptable like this, and I will work on filtering/formatting improvements today, separately?

magicus:
Okay.

magicus (approved):

PR #1859, opened 2023-07-27
fix: Fix item guide crashing on last page

kristofbolyai:


common/src/main/java/com/wynntils/models/gear/tooltip/GearTooltipFooter.java
```diff
@@ -74,7 +74,8 @@ public static List<Component> buildTooltip(GearInfo gearInfo, GearInstance gearI
         footer.add(tier);
 
         // restrictions (untradable, quest item)
-        if (gearInfo.metaInfo().restrictions() != GearRestrictions.NONE) {
+        // Note: "Zeer ID Test" item has null as restrictions...
```
magicus, line 77: I think we should fix this when reading the API json instead.

magicus: Assume a null restriction means "NONE".

In general, I dislike having enums that can be null. It is easy to forget about that since you think that an enum must have one of the enum values. I think we should avoid it wherever at all possible.

kristofbolyai: Can you look into that? I will keep this open as a reminder..

magicus: Okay

magicus: Actually, that was good luck I did! It did not have a null restriction. It had "Soulbound", which we could not parse and thus replaced with null. Presumably more items will have that in the future.

magicus:
Properly fixed in #1862

Issue #1648, opened 2023-06-29
Bank/Inventory/Pouch Search


Kepler-17c:
In Legacy the bank and inventory have a button that cycles through all crafting professions and highlights ingredients that can be used for it.

Having a search function with highlight for bank, inventory and pouch would be a great help:
- The search box from the bank could be applied to the inventory and pouch as well. (At least the design - I didn't check the implementation details)
- Additional search parameters to cover are:
  - profession type
  - level range
  - item tier
- Search features should be accessible via
  - either a comprehensive search string syntax
  - or additional buttons and text-fields to enter the different parameters.

magicus:
Do you intend to implement this, or is it a general feature request?

Kepler-17c:
This ticket is for:
- feature parity
- idea(s) how to improve the feature
- idea(s) how to merge with existing features

(so more like a feature request - I currently don't find much time to work on Wynntils or my personal coding projects)

Etrenak:
Hello,
For information, I am currently working on adding this feature and will be opening a merge request in a few days when it will be finished.

kristofbolyai:
Already implemented :)

PR #1860, opened 2023-07-27
feat: Add advanced search filters to container search and item guides

Etrenak:
Please note that I started using Wynntils a few weeks ago so I didn't know this feature existed in the legacy. That's why my implementation of this feature might be very different from the legacy one. I still hope it's good enough to be used.

I didn't add many filter options at the moment because I wanted to be sure the design was okay for you.

This PR also refers to #1648


kristofbolyai:
Oh, I opened this notification in the morning on my phone, then went ahead and completely forgot about it... Sorry about it, I will review this tomorrow.

In the meantime can I see some ingame screenshots of how this is supposed to work and look? (if it is really different from the legacy one, we might have to have both this implementation, and that one..)

Rafii2198:
I was curious myself, so I checked it out. I personally don't recall it being a Legacy thing, but then again I didn't use it for general play. I must say tho, the search doesn't work, the advanced search is fine but searching the name of the item is not possible at all as seen on second screenshot.
image
image
What legacy had (what the linked issue mentions) was a button to the left of bank and pouch, pressing it would cycle through all professions, highlighting related materials, basically what this `prof` parameter does.
image
As you can see, Armouring materials are highlight in different color (2 in bank, 1 in inventory) but this is all it did. It basically showed everywhere but the player inventory, chests, pouches, bank etc. had that button.

Rafii2198:
I forgot to add, so a follow-up
This is closer to Advanced Item Search Mode for Item Guide that was in Legacy. It is basically the same idea this is doing but for Item Guide
image
There were multiple parameters you could put in the search, similarly how this container search is doing, of course it had more parameters related to items and their ids as it was for the guide.

Etrenak:
Thank you for your comments. A last minute change that I did not properiy test is causing the search query not to match anything if no filters are used, thank you for pointing this out. It will be a quick easy fix but i will only be able to push it in ~12hours. I will try to also share some screenshots, in adddition to those from Rafii.

magicus:
I have not looked at the patch, just the screenshots and the comments. An item query language would indeed be an interesting addition. But it should be shared between the guide and inventories, I think. So if this PR adds a complex addition we will probably need to work a bit with it to get in the right place and form.

common/src/main/java/com/wynntils/features/inventory/search/ProfessionSearchFilter.java
```diff
@@ -0,0 +1,47 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean matches(WynnItem wynnItem) throws IllegalStateException {
+        if (profession == null) throw new IllegalStateException("ProfessionSearchFilter must be prepared before use");
+        else if (wynnItem instanceof IngredientItem ingredient)
+            return ingredient.getIngredientInfo().professions().contains(profession);
+        else if (wynnItem instanceof MaterialItem materialItem)
+            return materialItem
```
magicus: This is just highlighting how it was gathered. Would it not be interesting (perhaps more so?) to see materials that is used by any of the crafting professions? (I'm not sure we have that mapping present in Wynntils today but it is easy to add)

common/src/main/java/com/wynntils/features/inventory/search/WynnItemFilterManager.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.wynntils.features.inventory.search;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+import net.minecraft.client.resources.language.I18n;
+
+public class WynnItemFilterManager implements WynnItemFilterFactory {
```
magicus: This, and the entire search package should move to `services/itemfilter`. Turn this class into a `Service` and register it (check other services to see how). Then you can access the method to create filters like this: `Services.ItemFilter.createFilter()`.

magicus (commented):
I added some comments. This is not a complete review, I'll have to return later to do that.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -71,6 +79,13 @@ public class ContainerSearchFeature extends Feature {
     private SearchWidget lastSearchWidget;
     private SearchableContainerType currentSearchableContainerType;
     private boolean autoSearching = false;
+    private SearchQuery lastSearchQuery;
+
+    private final WynnItemFilterManager wynnItemFilterManager;
+
+    public ContainerSearchFeature() {
+        this.wynnItemFilterManager = new WynnItemFilterManager();
+    }
```
kristofbolyai: Features don't use constructors like this, you can just declare this in-line?

kristofbolyai: Oh god, this is a manager... Yeah, this is extra bad. Do the service refactor magicus mentioned, add this to Services.java and the mod will register it.

common/src/main/java/com/wynntils/features/inventory/search/LevelSearchFilter.java
```diff
@@ -0,0 +1,56 @@
...
+        if (!matcher.find())
+            throw new InvalidSyntaxException(
+                    "feature.wynntils.containerSearch.filter.level.invalid_range", searchString);
+
+        try {
+            minLevel = parseInt(matcher.group(1));
+            maxLevel = matcher.group(3) == null ? minLevel : parseInt(matcher.group(3));
+        } catch (NumberFormatException ignore) {
+            throw new InvalidSyntaxException(
+                    "feature.wynntils.containerSearch.filter.level.invalid_range", searchString);
```
kristofbolyai: Lot of i18n floating around the code. Perhaps you can make a custom exception, that accepts an enum you created with all the exceptions/keys.

kristofbolyai (commented):
1. Please self-review your code and fix every case of not using braces with if(-else) conditions. If it is a control flow statement (return, continue, break) and it fits in a singular line with the condition, it's fine, otherwise add brackets.

2. I think we might need to move this to the item guide. This feature looks promising, but nothing what it was. Or we make it a config to appear in some screens. But definitely not all screens need this, where as the profession type filter was useful everywhere. We will need to discuss that after we've seen what this can do.

Overall, this looks to be a great feature, but we will need to do a lot of rounds of reviews to adapt the code style to Artemis standards.

common/src/main/java/com/wynntils/features/inventory/search/ProfessionSearchFilter.java
Etrenak: I agree. I was not able to find such a mapping but I'll look into it. If there's none I'll see how to properly add it.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
Etrenak: I'll do this

common/src/main/java/com/wynntils/features/inventory/search/LevelSearchFilter.java
Etrenak: To be honest, I'm not quite sure to understand what you mean.
Would it be something like that ?

```
public class InvalidSyntaxException extends Exception {
    public InvalidSyntaxException(InvalidSyntaxCause cause, String... parameters) {
        super(I18n.get(cause.getTranslateKey(), parameters));
    }
    public enum InvalidSyntaxCause {
        LVL_INVALID_RANGE("feature.wynntils..."),
        PROF_INVALID_PROFESSION("feature.wynntils...");
        ...
    }
}
```

Etrenak:
Thank you for the reviews. There was a bit of confusion from me between the item guide search and the container search features of the legacy. I'll make all the adjustments, and I definitely think we will be able to share the filter/query system for the Item guide, assuming someone does the integration for it.

Etrenak:
Here are two additional screenshots.

##


However, I just found out there was a bug with the tooltips overlapping. I'm not sure how to handle this, given that `AbstractWidget` computes `isHovered` based only on the size of the widget and then use it right away to decide whether to show the tooltip or not. We could move the info icon outside of the input box but it looks really off.


Also, I'm not used at all to code reviews, should I resolve the conversations myself or let the reviewer do it ?

kristofbolyai:
> Also, I'm not used at all to code reviews, should I resolve the conversations myself or let the reviewer do it ?

If you think you completed the asked request, then resolve it.

kristofbolyai:
I am not quite sure what to do with the tooltip at the moment. But for the time being, I am kind of leaning towards adding this filter to the item guide, and removing it from general screens. Would you be able to do that?

What do you think @magicus?

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -115,7 +115,7 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
 
     protected void doRenderWidget(
             PoseStack poseStack,
-            String renderedText,
+            Pair<String, Integer> renderedTextDetails,
```
magicus: I have not followed the complete flow, but does this really need to be passed as a pair? As opposed to two individual arguments, that is.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
```diff
@@ -0,0 +1,29 @@
...
+    protected String searchString;
+
+    protected ItemFilter(String searchString) {
+        this.searchString = searchString;
+    }
+
+    /**
+     * Prepares the operator with the given search string. Similar to compiling for a regex
+     * @return true if the operator was prepared successfully
+     * @throws InvalidSyntaxException if the search string is not invalid
```
magicus: What does returning false mean? Or put differently, why have an exception instead of returning false? Or, if having an exception is what you want, why have a return value? (I personally prefer return value over exception in this case)

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public ItemFilterService() {
+        super(List.of());
+        registerFilter("lvl", "level", LevelSearchFilter::new);
+        registerFilter("prof", "profession", ProfessionSearchFilter::new);
+    }
+
+    private void registerFilter(String keyword, String translateKey, Function<String, ? extends ItemFilter> supplier) {
+        filterSuppliers.put(keyword, supplier);
+        filterUsages.put(keyword, "feature.wynntils.itemFilters." + translateKey + ".usage");
```
magicus: You need to turn this into "service.wynntils.itemFilter."

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Creates a new ItemFilter based on the given keyword an initates it with the given search string
```
magicus:
```suggestion
     * Creates a new ItemFilter based on the given keyword and initiates it with the given search string
```

common/src/main/java/com/wynntils/services/itemfilter/LevelSearchFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static java.lang.Integer.parseInt;
+
+import com.wynntils.models.items.WynnItem;
+import com.wynntils.models.items.properties.LeveledItemProperty;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class LevelSearchFilter extends ItemFilter {
+    private static final Pattern LEVEL_RANGE_PATTERN = Pattern.compile("^(\\d+)(-(\\d+))?$");
```
magicus: If you make the optional part a non-matching group with `(?:` there will be no confusing group 2, but just group 1 and 2 (if present).
```
    private static final Pattern LEVEL_RANGE_PATTERN = Pattern.compile("^(\\d+)(?:-(\\d+))?$");
```

common/src/main/java/com/wynntils/services/itemfilter/LevelSearchFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public LevelSearchFilter(String searchString) {
+        super(searchString);
+    }
+
+    @Override
+    public boolean prepare() throws InvalidSyntaxException {
+        Matcher matcher = LEVEL_RANGE_PATTERN.matcher(searchString);
+        if (!matcher.find()) {
+            throw new InvalidSyntaxException("feature.wynntils.itemFilters.level.invalid_range", searchString);
```
magicus: Please use lowerCamel for i18n keys

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
Etrenak: It would work perfectly as two individual arguments. I just thought it was more consistent since it is the exact value we receive from `getRenderedHighlighedInterval()`. Please tell me if you think it's better to change

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
Etrenak: It should throw an exception in order to give information on the cause of the failure to prepare the filter. I don't really see why it should ever return false. I will turn this into a void method.

Etrenak: Oh i didn't read this right. I don't really see how a return value is better than throwing an exception since we want to explicitely give the cause of the error ?

Etrenak: Well I removed the return value in [0c64aeb](https://github.com/Wynntils/Artemis/pull/1860/commits/0c64aeb7b3d2547fc709bd0f4d43d11fbd057936). If you don't think it was the right thing to do, please tell me.

Etrenak:
I will certainly add this to the item guide, but I can't guarantee that it will be done in a few days.

However, I'd like to mention that as a player myself, I wanted to be able to filter efficiently ingredients in my bank, and that's the real purpose of this PR.

If you think it's a bit overwhelming to have in every container, why not add an option in the config to switch from the current simple search system to the advanced one ? It can be done in addition to what you said about filtering screens which should benefit from the advanced search.

kristofbolyai:
If you add a config to disable this filter and just have the normal search, it will be fine.

Etrenak:
So it's still a WIP, but I think I shouldn't have to refactor more than that to properly implement the feature. At the moment it looks like this:


I'll have to think about how to make a proper explanation message for the end users and begin to add every useful filter I can think of.

@magicus and @kristofbolyai I'd love to hear back from you on the questions left for the opened reviews!

kristofbolyai:
Hey! I can't review this code wise yet, but from visuals it looks good. Did you add the config yet for normal containers?

Etrenak:
> Hey! I can't review this code wise yet, but from visuals it looks good. Did you add the config yet for normal containers?

Not yet, I'll try to do this this evening.

magicus:
I'll have a look at this as soon as I can (most likely at the latest tomorrow night).

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -199,15 +200,13 @@ private Pair<String, Integer> getRenderedText(float maxTextWidth) {
...
-            builder.append(textBoxInput.charAt(stringPosition));
-
+        int stringPosition = cursorPosition;
+        while (font.width(builder.toString()) < maxTextWidth && stringPosition > 0) {
             stringPosition--;
+            builder.append(textBoxInput.charAt(stringPosition));
         }
 
-        final int startingAt = Math.max(stringPosition, 0); // If we went too far, start at the beginning
+        final int startingAt = stringPosition;
```
kristofbolyai, line 211: Does this fix anything or did you just refactor this?

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -14,17 +14,19 @@
...
 import java.util.Objects;
 import java.util.function.Consumer;
 import net.minecraft.client.gui.Font;
 import net.minecraft.network.chat.Component;
 import net.minecraft.sounds.SoundEvents;
 import org.lwjgl.glfw.GLFW;
 
-public class SearchWidget extends TextInputBoxWidget {
-    private static final Component DEFAULT_TEXT =
+public abstract class SearchWidget<S> extends TextInputBoxWidget {
```
kristofbolyai: Do you think this refactor of this class is strictly needed? The way I see it is that you can just extend SearchWidget in `ItemSearchWidget`. I am not strictly opposed to this, but it does feel like loads of unnecessary change.

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
```diff
@@ -20,23 +20,59 @@ public class BasicTexturedButton extends WynntilsButton {
     private final Consumer<Integer> onClick;
     private List<Component> tooltip;
 
+    private int renderTooltip;
```
kristofbolyai: Why not just a bool? Or an enum?

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
```diff
@@ -20,23 +20,59 @@ public class BasicTexturedButton extends WynntilsButton {
...
+                    poseStack,
+                    texture.resource(),
+                    this.getX(),
+                    this.getY(),
+                    0,
+                    getWidth(),
+                    getHeight(),
+                    texture.width(),
+                    texture.height());
+        else RenderUtils.drawTexturedRect(poseStack, texture, this.getX(), this.getY());
```
kristofbolyai: I am not going to comment on every case I find, but please self-review yourself and use braces for all if(-else) statements unless they fit in 1 line and are control flow operations (like return, continue). Thanks. 

(As this is one of our not so well documented code style requirements, I will make sure to document this).

common/src/main/java/com/wynntils/services/itemfilter/InvalidSyntaxException.java
kristofbolyai, line 1: I am not sure how familiar you are with the function system in Artemis. But, the basic concept is the same, parsing user input. Instead of using exceptions, I added `ErrorOr`. I originally saw the class used in an unrelated project, but I think it is a really great way of handling situations like this (where you want to propagate errors up multiple classes to display to the user in a safer way).

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilter.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.wynntils.models.items.WynnItem;
+import com.wynntils.models.items.properties.LeveledItemProperty;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class LevelItemFilter extends ItemFilter {
+    private static final Pattern LEVEL_RANGE_PATTERN = Pattern.compile("^(\\d+)(?:-(\\d+))?$");
+
+    private int minLevel, maxLevel;
```
kristofbolyai: Second kind of unwritten code style rule is that fields must be declared in separate lines. Thanks.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
```diff
@@ -0,0 +1,28 @@
...
+
+    protected ItemFilter(String filterValue) {
+        this.filterValue = filterValue;
+    }
+
+    /**
+     * Prepares the filter with the provided value. Similar to compiling for a regex
+     * @throws InvalidSyntaxException if the provided value is not valid
+     */
+    public abstract void prepare() throws InvalidSyntaxException;
```
kristofbolyai: General comment about this class, why can't you just call prepare in the constructor? That way you can remove all the "isPrepared" checks from subclasses.

common/src/main/java/com/wynntils/screens/guides/WynnItemGuideScreen.java
```diff
@@ -0,0 +1,54 @@
...
+import com.wynntils.screens.base.WynntilsListScreen;
+import com.wynntils.screens.base.widgets.BackButton;
+import com.wynntils.screens.base.widgets.ItemSearchWidget;
+import com.wynntils.screens.base.widgets.PageSelectorButton;
+import com.wynntils.screens.base.widgets.WynntilsButton;
+import com.wynntils.services.itemfilter.SearchQuery;
+import com.wynntils.utils.render.Texture;
+import net.minecraft.network.chat.Component;
+
+public abstract class WynnItemGuideScreen<E, B extends WynntilsButton> extends WynntilsListScreen<E, B, SearchQuery>
```
kristofbolyai:
```suggestion
public abstract class WynntilsGuideScreen<E, B extends WynntilsButton> extends WynntilsListScreen<E, B, SearchQuery>
```

Perhaps this is a better name. Alternatively, `WynntilsGuideScreenBase`.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
kristofbolyai: Seeing the changes in WynntilsListScreen, it made my feeling stronger about not keeping this change, if it is not needed. (I don't currently see why it would be, at the moment.)

kristofbolyai (commented):
Hi there!

I could not manage to review this in full, but I think I given you enough direction to work on when you are free. I hope I have not overwhelmed you. As always, feel free to ask me any questions, if needed.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
Etrenak: This is a fix. Assuming the text was larger than the maximum text width by 2 chars, the behavior was:
- Decrement `stringPosition`
- while `stringPosition`>= 0
  - Add char at `stringPosition`
  - Decrement `stringPosition`

It means that for the latest iteration, we get `stringPosition` offseted by `-1` instead of the actual index of the last char added. That's why I simply switched the two instructions in the loop and removed the initial decrementation

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
Etrenak: The thing is, `WynntilsListScreen` has a `searchWidget` attribute, and uses `reloadElementsList(String searchTerm)` in its `reloadElements(String searchterm)` **private** method.

The way I see it is that `WynntilsItemGuideScreen` and `WynntilsIngredientGuideScreen` cannot inherit from a class where `reloadElementsList(String searchTerm)` is implemented, because it makes no sense.

In the mean time, I wanted to keep the logic in the overrides of `charTyped()` and `keyPressed()` for example.

Having a generic type for the search query was the simplest and lightest way I found to refactor the less I could. The other ways I could think of were either heavier or seemed wrong to me.

I'll look into it again but I really can't guarantee I'll figure out a better and lighter way to do it.

common/src/main/java/com/wynntils/services/itemfilter/InvalidSyntaxException.java
Etrenak: I'm a casual Wynncraft player so I must say I never heard of it. Thanks though, I'll have a look since using Exceptions doesn't seem like the right way to do.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
kristofbolyai: Sounds good. Okay.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
kristofbolyai: I can look into this more deeply tomorrow, so I can find something I can accept (or keep this).

kristofbolyai: What do you think of having `WynntilsGuideScreen` extending `WynntilsListScreen`  and adding an empty implementation for `reloadElementsList(String searchTerm)`. You can create your own reload method with the custom filters, however you want. And you just override the method calling `reloadElementsList` since it is only called in 1 location. That should bring down the amount of changes drastically. I know it will be a bit weird inheritance wise, but I think it is weird either way.. and this way is much less code.

Etrenak: That's the way I did it originally, but then I thought it was not right inheritance wise as you mentioned it.
If this is fine for you, I also think this is the best option to reduce the amount of "unnecessary" change.

kristofbolyai: Yeah, it is okay that way. Thanks.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
magicus: Yes, I think it is better to not have a pair. They are mostly useful when you need to return two values at the same time, or possibly when you need to store it in a container of some kind.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.itemfilter;
+
+import com.wynntils.models.items.WynnItem;
+
+public abstract class ItemFilter {
```
magicus: This smells like something that wants to be an interface instead of an abstract class.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.wynntils.core.components.Service;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+public class ItemFilterService extends Service {
+    private final Map<String, Function<String, ? extends ItemFilter>> filterSuppliers = new HashMap<>();
```
magicus: I think the code can be made clearer if you create a ItemFilterFactory class instead.

magicus: Also, the prepare() method seem to be an artifact of this design choice, rather than something inherent in the problem domain.

So if you move the validation and parsing of the string into a ItemFilterFactory, I think you can get a simpler design. You can have something along these lines:
```
interface ItemFilterFactory {
  ItemFilter create(String inputString);

  String getKeyword();

  String getI18nKey();
}

interface ItemFilter {
   boolean matches(WynnItem wynnItem);
}
```

and you can basically turn the current level and profession item filters into item filter factories, and just have a trivial nested class that implements their respective ItemFilter implementations.

And then you just do something like:
```
       registerFilterFactory(new  LevelItemFilterFactory());
       registerFilterFactory(new  ProfessionIItemFilterFactory());
```
in the Service.

Etrenak:
Hi, I did not have much time last week, sorry for the late update on this. I thought the suggestion of magicus was great so I refactored to use ItemFilter factories, and also used `ErrorOr` instead of throwing `Exceptions`.

Still, I think there are improvements to make on returning error messages inside factories.
I'm not satisfied with using `(getI18n() + ".XX")`, but I really don't know what's the proper way to implement it :

- Something like `static final INVALID_RANGE_ERROR = "features.wynntils....";` in every `XItemFilterFactory` class ?
- A `getI18nSubkey(String subkey)` function to extract the string concatenation?

Note: I reverted then re-added the feature, so this batch of commits is only refactoring with zero changes to the above screenshots.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
```diff
@@ -103,7 +103,8 @@ public void renderWidget(PoseStack poseStack, int mouseX, int mouseY, float part
 
         doRenderWidget(
                 poseStack,
-                renderedText,
+                renderedTextDetails.a(),
+                renderedTextDetails.b(),
```
kristofbolyai: Why not use the proper variables? Extract pair.b() and use the named variables.

common/src/main/java/com/wynntils/screens/guides/WynntilsGuideScreen.java
```diff
@@ -0,0 +1,68 @@
...
+        super(component);
+        this.searchWidget =
+                new ItemSearchWidget(0, -22, Texture.QUEST_BOOK_BACKGROUND.width(), 20, q -> reloadElements(), this);
+    }
+
+    @Override
+    protected void doInit() {
+        super.doInit();
+
+        // TODO Add the explanation text
```
kristofbolyai, line 28: What kind of explanation?

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+     * in the search query are ignored. If there are no plain text tokens, this method always returns true.
+     *
+     * @param itemName the name to check
+     * @return true if the name contains the concatenated plain text tokens, false otherwise
+     */
+    public boolean itemNameMatches(String itemName) {
+        return plainTextTokens.isEmpty()
+                || itemName.toLowerCase(Locale.ROOT)
+                        .contains(String.join(" ", plainTextTokens).toLowerCase(Locale.ROOT));
+    }
```
kristofbolyai: Why not have a generic matches() function and these two as private helpers?

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private final List<ItemFilter> itemFilters;
+
+    private final List<String> plainTextTokens;
+
+    private final List<Integer> ignoredCharIndices;
+
+    private final List<Integer> validFilterCharIndices;
+
+    private final List<String> errors;
```
kristofbolyai: You can remove that spacing, we only put them in case there is a reasonable logic separation. This is all the same logic.

common/src/main/java/com/wynntils/screens/base/WynntilsListScreen.java
```diff
@@ -34,7 +35,7 @@ public abstract class WynntilsListScreen<E, B extends WynntilsButton> extends Wy
     protected List<E> elements = new ArrayList<>();
 
     private final List<B> elementButtons = new ArrayList<>();
-    protected final QuestBookSearchWidget searchWidget;
```
kristofbolyai, line 37: Is this still needed? I thought WynntilsGuideScreen can handle this.

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
```diff
@@ -20,23 +20,58 @@ public class BasicTexturedButton extends WynntilsButton {
     private final Consumer<Integer> onClick;
     private List<Component> tooltip;
 
+    private boolean renderTooltipAbove;
```
kristofbolyai: `renderTooltipAbove`, above what, the mouse? I suggest you rename this to `renderTooltipAboveMouse` then

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
```diff
@@ -0,0 +1,165 @@
...
+            int renderedTextStart,
+            String firstPortion,
+            String highlightedPortion,
+            String lastPortion,
+            Font font,
+            int firstWidth,
+            int highlightedWidth,
+            int lastWidth,
+            boolean defaultText) {
+        if (defaultText || !getTextBoxInput().equals(searchQuery.getQueryString())) {
```
kristofbolyai: `!getTextBoxInput().equals(searchQuery.getQueryString())` If our parsed query is not the same as the input, do we just render the default text? Why? Can this happen?

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
```diff
@@ -0,0 +1,165 @@
...
+        } else if (searchQuery.getValidFilterCharIndices().contains(charIndex)) {
+            return StyledText.fromComponent(Component.literal(String.valueOf(c)).withStyle(ChatFormatting.YELLOW));
+        } else {
+            return StyledText.fromString(String.valueOf(c));
+        }
+    }
+
+    @Override
+    protected int getMaxTextWidth() {
+        return super.getMaxTextWidth() - 4;
```
kristofbolyai: Might be cleaner to just implement this without super.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterFactory.java
```diff
@@ -0,0 +1,27 @@
...
+     *
+     * @param inputString the input string
+     * @return the created item filter, or a translated error string if the input string is invalid
+     */
+    ErrorOr<? extends ItemFilter> create(String inputString);
+
+    /**
+     * @return the keyword used to identify this filter
+     */
+    String getKeyword();
```
kristofbolyai: keyword. Isn't it just the name of the filter? Perhaps that is better. Maybe even this could be an list. Like "lvl" and "level" are both fully acceptable.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+import net.minecraft.client.resources.language.I18n;
+
+public class ItemFilterService extends Service {
+    private final Map<String, ItemFilterFactory> filterFactories = new HashMap<>();
+
+    public ItemFilterService() {
+        super(List.of());
+        registerFilter(new LevelItemFilterFactory());
+        registerFilter(new ProfessionItemFilterFactory());
```
kristofbolyai: Just so we don't do this in the constructor, can you extract this to a method? Kind of like `com.wynntils.core.consumers.functions.FunctionManager#registerAllFunctions`

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private void registerFilter(ItemFilterFactory factory) {
+        filterFactories.put(factory.getKeyword(), factory);
+    }
```
kristofbolyai: Please put private methods after public ones. Thanks

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public String getKeyword() {
+        return "lvl";
+    }
+
+    @Override
+    public String getI18nKey() {
+        return "feature.wynntils.itemFilter.level";
```
kristofbolyai: You can use a little bit of reflection magic to have this automatically. Check out `com.wynntils.core.consumers.functions.Function#getTranslatedName` and try to copy what it does.

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
```diff
@@ -0,0 +1,40 @@
...
+import com.wynntils.models.items.items.game.GatheringToolItem;
+import com.wynntils.models.items.items.game.IngredientItem;
+import com.wynntils.models.items.items.game.MaterialItem;
+import com.wynntils.models.profession.type.ProfessionType;
+
+public class ProfessionItemFilter implements ItemFilter {
+    private ProfessionType profession;
+
+    public ProfessionItemFilter(ProfessionType profession) {
+        // TODO FILTER NULL
```
kristofbolyai: What does this mean :)?

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
```diff
@@ -0,0 +1,40 @@
...
+        } else if (wynnItem instanceof MaterialItem materialItem) {
+            return materialItem
+                            .getMaterialProfile()
+                            .getResourceType()
+                            .getMaterialType()
+                            .getProfessionType()
+                    == profession;
+        } else if (wynnItem instanceof GatheringToolItem gatheringToolItem) {
+            return gatheringToolItem.getToolProfile().toolType().getProfessionType() == profession;
+        }
```
kristofbolyai: We kind of need to add `ProfessionItemProperty`. If you want to, go ahead. Otherwise, this is fine and we will add it later.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1754,6 +1761,7 @@
   "screens.wynntils.wynntilsGuides.importedFavorites": "Imported %s favorites!",
   "screens.wynntils.wynntilsGuides.ingredientGuide.name": "Ingredient Guide",
   "screens.wynntils.wynntilsGuides.invalidClipboard": "Failed to import favorites! Clipboard does not contain exported list of favorite items.",
+  "screens.wynntils.wynntilsGuides.itemGuide.advancedSearch.help": "Available filters :",
```
```suggestion
  "screens.wynntils.wynntilsGuides.itemGuide.advancedSearch.help": "Available filters: ",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -161,6 +161,7 @@
   "feature.wynntils.containerSearch.highlightColor.description": "What color should the highlight circles be?",
   "feature.wynntils.containerSearch.highlightColor.name": "Highlight color",
   "feature.wynntils.containerSearch.name": "Container Search Highlight",
+  "feature.wynntils.containerSearch.tooltip": "Available filters : ",
```
```suggestion
  "feature.wynntils.containerSearch.tooltip": "Available filters: ",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -438,6 +439,12 @@
   "feature.wynntils.itemFavorite.lootChestCloseOverride.description": "After how many ESC presses should the chest be closed? Set to 0 to disable.",
   "feature.wynntils.itemFavorite.lootChestCloseOverride.name": "Loot Chest Close Override",
   "feature.wynntils.itemFavorite.name": "Favorite Items",
+  "feature.wynntils.itemFilter.level.invalidRange": "%s: should be an integer or two integers seperated by a dash",
```
```suggestion
  "feature.wynntils.itemFilter.level.invalidRange": "%s: Should be an integer or a range, seperated by a dash",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -438,6 +439,12 @@
...
+  "feature.wynntils.itemFilter.level.maxLessThanMin": "%s: min must be less than max",
```
kristofbolyai: Can you capitalize all errors?

```suggestion
  "feature.wynntils.itemFilter.level.maxLessThanMin": "%s: Min must be less than max",
```

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+                validFilterCharIndices.addAll(
+                        IntStream.rangeClosed(tokenStartIndex, tokenStartIndex + filterString.length())
+                                .boxed()
+                                .toList());
+
+                // ...and try to create the filter only if the filter value is not empty, because a filter without a
+                // value is pointless and we don't want to show the error the factory might return because of that.
+                // We still want to highlight the filter keyword though, that's why we handle the empty value here.
+                if (valueString.isEmpty()) continue;
+                ErrorOr<? extends ItemFilter> filterOrError =
```
kristofbolyai:
```suggestion

                ErrorOr<? extends ItemFilter> filterOrError =
```

kristofbolyai (commented):
I like the new factory/filter/service implementation. I still have some questions around the screens, widgets, and some implementation details. But it is close to approve-able now.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.itemfilter;
```
kristofbolyai: Can you move the interfaces to a new, `type` sub package? And the implementations in `filters`.

kristofbolyai:
> Still, I think there are improvements to make on returning error messages inside factories. I'm not satisfied with using `(getI18n() + ".XX")`, but I really don't know what's the proper way to implement it :
> 
> * Something like `static final INVALID_RANGE_ERROR = "features.wynntils....";` in every `XItemFilterFactory` class ?
> * A `getI18nSubkey(String subkey)` function to extract the string concatenation?

I don't think it is that bad, but perhaps a method to get the error? `String getError(String errorName, Object.. formatValues)`

common/src/main/java/com/wynntils/screens/base/WynntilsListScreen.java
Etrenak: I don't really see what you mean to be honest. This allows to keep all the pre existing logic in WynntilsListScreen for the search bar, and properly overwrite the default search widget.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
Etrenak: If `defaultText`, then we let the superclass handle it. (It should display the default text or whatever, this does not relate to our specific implementation)

The parsed query should never be the same as the input, since it is updated in `onUpdate()`. If for a mysterious reason this happens, I thought it was better to let the default SearchWidget renering behavious (i.e. rendering the text in plain white), rather than messed up colouring not matching the string displayed.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
Etrenak: Done, but I would be happy to know why this is better like this. Is this a convention? I never heard it was better to extract named variables before passing it so I'm really curious.

common/src/main/java/com/wynntils/screens/guides/WynntilsGuideScreen.java
Etrenak: Something listing the available filters, but there is not much space so we must think of something. It might be in the same format as legacy for example : https://user-images.githubusercontent.com/3767283/131205683-05a060ba-23db-4ae4-8004-f1d1500034fe.png

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterFactory.java
Etrenak: Since we would have multiple names, what do you think about `getAliases()` ?

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
magicus: I think the intention is if the tooltip should have its top position at the mouse position, or its bottom (which is used if the mouse is too close to the bottom of the screen). I think we have used similar names in other places were we handle this, but we can probably think of something better. `tooltipOverlaidUpwards` perhaps?

magicus: (and we should probably change the other places as well, then; but that need not be done as part of this PR)

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilterFactory.java
Etrenak: I agree, but I would have to copy the code to every `ItemFilterFactory` implementation. Maybe I should create an abstract superclass implementing `ItemFilterFactory` which would handle the translations & names/keywords/aliases?

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
Etrenak: Oops my apologies, I forgot to remove this.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
magicus: This is just a problem with the Pair class. Ideally, we could have created simple "throw away" classes in Java, were we have a need to group two objects (e.g. to be able to return it from a function, or store it in a map). Now we can't due to language restrictions. 

So we have created the Pair class to help with this. But it is really non-descriptive. a() and b() says nothing about the kind of data that is stored in them. So we want to use the Pair for as little as absolutely needed, and when we extract the values, it helps by extracting them to local variables and give them suitable names.

This is of course a matter of subjectivity, but I think that @kristofbolyai's suggestion is reasonable here.

magicus: (I see now that you already have extracted `renderedTextDetails.a()` to `renderedText`...)

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
Etrenak: I think it is better to wait the end of this infinitely long PR because of all my mistakes, before focusing on adding more filters and improving this one.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
magicus: Your comment contradicts the code. You mean that the parsed query should always be the same as the input? Otherwise, the test in the code is wrong.

common/src/main/java/com/wynntils/screens/guides/WynntilsGuideScreen.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    protected final void reloadElementsList(String ignored) {
+        /*
+        At the time this code is written, this method will always be called with this.searchWidget().getTextInput() as
+        argument. Therefore it makes sense to use this hack to migrate from simple text based search to advanced filters
+        search.
+        This hack is preferable to adding even more complexity in the superclass.
+        See https://github.com/Wynntils/Artemis/pull/1860#discussion_r1279180536 for more details.
+        */
+        if (searchWidget instanceof ItemSearchWidget) { // Should always be true
```
magicus: Maybe invert this check an add an assertion then? Like:

if (!searchWidget instanceof ItemSearchWidget) {
  assert false;
  return;
}

That way we can see any breakage of this assumption in the dev environment.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilter.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.wynntils.models.items.WynnItem;
+
+public interface ItemFilter {
```
magicus:
```suggestion
@FunctionalInterface
public interface ItemFilter {
```

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterFactory.java
magicus: I'm not sure where this method is used, but in general I think we should not *only* have a list of aliases to identify something. You need a primary name, and then you can have additional aliases (or an empty list). This helps a lot with making good design in future use cases.

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
Etrenak: A `WynnItem` does not have a `name` property, so we can only match the name for an `ItemStack`, whereas the filter matching is done on a `WynnItem`. 
So a `matches()` function should take an `ItemStack` as argument because AFAIK we can't get back a name from a `WynnItem`. But then if the `ItemStack` does not resolve to a `WynnItem`, what do we do ?

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
magicus: I agree, that sounds like a reasonable extension. I can help do it in a separate PR as well, push it to main, and then you can merge it and use it. Ping me if you want that to happen. (I only check this PR once in a while otherwise).

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilter.java
magicus, line 1: I'll need @kristofbolyai to weigh in on this also, but I think that the ItemFilter classes are so trivial, and closely related to their respective factories, that I think it would be much better to have them as nested inner classes inside the factories. 

They are in fact close to not having to exist as classes at all, but just being a lambda returned from the factory... At least with the convention I'm suggesting, the choice of wether to just return a lambda, or make an actual inner class, can be left to the designer of each respective new itemfilterfactory.

common/src/main/java/com/wynntils/screens/base/widgets/TextInputBoxWidget.java
Etrenak: Thanks for the explanation ! (renderedText was already extracted not by me)

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+            List<String> plainTextTokens) {
+        this.queryString = queryString;
+        this.itemFilters = itemFilters;
+        this.ignoredCharIndices = ignoredCharIndices;
+        this.validFilterCharIndices = validFilterCharIndices;
+        this.errors = errors;
+        this.plainTextTokens = plainTextTokens;
+    }
+
+    public static SearchQuery fromQueryString(String queryString) {
```
magicus: I would recommend moving this functionality to the ItemFilterService. If you do, it will be clear that this is a very thin data carrying class. In fact, it could probably be turned into a record instead of a class if you do that.

The matcher functions do also seem a bit out of place, as bolyai suggests. I'm not sure where they are used, but they could perhaps also move to the service.

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+    public boolean isEmpty() {
+        return itemFilters.isEmpty() && plainTextTokens.isEmpty();
+    }
+
+    public String getQueryString() {
+        return queryString;
+    }
+
+    public List<Integer> getIgnoredCharIndices() {
+        return Collections.unmodifiableList(ignoredCharIndices);
```
magicus: I know wrapping lists to be returned in a read-only container is a recommended best practice, but we don't do it elsewhere in the code base. Our assumption has been (at least that's how I've seen it) that we trust the users of the APIs we provide to know that they should not modify collections, or make a copy if they need to. This makes the code slightly more fragile, but lets us get rid of a lot of boilerplate, and be able to treat returns of collections like returns of any other variable.

@kristofbolyai Should we start doing this? I have been pondering it every time I see that yellow marker from IntelliJ. I see the point of getting the extra safety, but I really dislike having all this extra boilerplate. On the other hand, IntelliJ can probably help with writing it, so it is mostly a matter of it being in the way when reading the code.

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
```diff
@@ -0,0 +1,162 @@
...
+import com.wynntils.core.components.Services;
+import com.wynntils.models.items.WynnItem;
+import com.wynntils.utils.type.ErrorOr;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.IntStream;
+
+public class SearchQuery {
```
magicus: So maybe you have already been discussing this, and I'm coming in late to the party. But I think this name is too generic, and would rather see it named `ItemSearchQuery`.

Etrenak: I received no comments about this but it is a better name indeed.

magicus:
I know I was the one to suggest the ItemFilterFactory name, but when I see it now I don't like it. The Factory is what I think of when I hear the term "ItemFilter". So I think we should flip this around. The ItemFilterFactories should be the ones that are called ItemFilters. That means that the class currently called "ItemFilter" needs to be renamed. I think it should be clear that it is an instantiated item filter that is bound to a particular string. 

So something like `ItemFilterInstance` perhaps?

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
magicus: Also, if you make it a record, please move it to the `type` subpackage. (We use it for interfaces, enums and records)

magicus:
I have not looked at this for a while, but I must say I am very pleased at how this is turning out! Some more polishing remains, but this is going to be a nice and solid extension to Wynntils when it is done. ❤️

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
kristofbolyai: > But then if the ItemStack does not resolve to a WynnItem, what do we do?

Return false?

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilter.java
kristofbolyai: I think the lamba method should be fine.

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
kristofbolyai: I did it in some places where I saw it fit. But not everywhere, which is not good. To be honest, if we do this, we should also make models/services/handlers/managers return optionals..

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
Etrenak: Exactly as @magicus said. However, I could not find similar names in other places.
If I'm not mistaken, the default behavior was to display the tooltip upward because `BasicTexturedButton` was only used for the icon at the bottom of the map :

image

Now, I want to use it for an icon that is on top of container screens. Because English is clearly not my native language, I'll let you decide what you prefer as a name for this attribute.

I think `renderTooltipAboveMouse` is fine `tooltipOverlaidUpwards` might be better. Either way, it's pretty straightforward.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
Etrenak: @magicus My apologies, the parsed query should indeed always be the same as the input.

common/src/main/java/com/wynntils/screens/base/widgets/BasicTexturedButton.java
kristofbolyai: Well, okay. This is changing in 1.20, I will fix it there...

common/src/main/java/com/wynntils/services/itemfilter/LevelItemFilterFactory.java
Etrenak: @magicus I would love to have your opinion on that. The way I see it is it's always better not to have duplicated code, so I should have an abstract superclass implementing `ItemFilterFactory` handling the `getTranslationKey()` and having a constructor prototype like  `AsbtractItemFilterFactory(String name, List aliases)`.

That's the way I would have done it but now it seems kind of wrong to me with all the suggestion you made in this PR.

kristofbolyai: I think you mentioned moving to a new name, maybe it is time to move away from being an interface as well.. I know we suggested it first..

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
magicus: > if we do this, we should also make models/services/handlers/managers return optionals..

That is a much more and serious problem we have, since that has caused a lot of real-world problems with NPEs. The modifying of returned collections is more of a theoretic problem.

common/src/main/java/com/wynntils/services/itemfilter/SearchQuery.java
Etrenak: Done

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
Etrenak: @magicus It would be great. It will probably take you far less time than me to properly implement it 😄. Thank you!

Etrenak:
Based on your suggestions and my concerns:

- `SearchQuery` is now a record, all the logic has been transferred to the `ItemFilterService`
- `ItemFilters` are now `ItemFilterMatchers`
- `ItemFiltersFactories` are now `ItemFilters`
- `ItemFilters` (former factories) is now an abstract class instead of an interface to handle translation & name/aliases
- `ItemFilterMatcher` is a now functional interface to reduce verbosity for simple matchers (such as `LevelItemFilter`'s matcher)

The term "matcher" might not be the most relevant. As always I'll be happy to change it if you come with a better idea.

The handling of translations and name/aliases is the same as in the `Function` class, as suggested by @kristofbolyai

common/src/main/java/com/wynntils/services/itemfilter/type/ItemFilter.java
```diff
@@ -0,0 +1,62 @@
...
+    public List<String> getAliases() {
+        return Collections.unmodifiableList(aliases);
+    }
+
+    protected String getTranslationName() {
+        return translationName;
+    }
+
+    protected String getTranslation(String keySuffix, Object... formatValues) {
+        return I18n.get("feature.wynntils.itemFilter." + getTranslationName() + "." + keySuffix, formatValues);
```
magicus:
```suggestion
        return I18n.get("service.wynntils.itemFilter." + getTranslationName() + "." + keySuffix, formatValues);
```
(and also update en_us.json)

common/src/main/java/com/wynntils/services/itemfilter/filters/ProfessionItemFilterMatcher.java
```diff
@@ -0,0 +1,40 @@
...
+package com.wynntils.services.itemfilter.filters;
+
+import com.wynntils.models.items.WynnItem;
+import com.wynntils.models.items.items.game.GatheringToolItem;
+import com.wynntils.models.items.items.game.IngredientItem;
+import com.wynntils.models.items.items.game.MaterialItem;
+import com.wynntils.models.profession.type.ProfessionType;
+import com.wynntils.services.itemfilter.type.ItemFilterMatcher;
+
+public class ProfessionItemFilterMatcher implements ItemFilterMatcher {
```
magicus: I still think this should be an inner class in ProfessionItemFilter.

magicus:
This looks so much better! I'm just pondering if "ItemFilterInstance" would not be more descriptive than "ItemFilterMatcher", after all... I think the gist of it is that it is an ItemFilter that is instantiated with a specific search string.

common/src/main/java/com/wynntils/services/itemfilter/type/ItemFilter.java
Etrenak: I also think it would make more sense. However other services using translations use "feature." (`UpdateService` & `LootrunService`). Do you still want to make the change ?

common/src/main/java/com/wynntils/services/itemfilter/filters/LevelItemFilter.java
```diff
@@ -0,0 +1,48 @@
...
+        int minLevel;
+        int maxLevel;
+        Matcher matcher = LEVEL_RANGE_PATTERN.matcher(inputString);
+        if (!matcher.find()) {
+            return ErrorOr.error(getTranslation("invalidRange", inputString));
+        }
+
+        try {
+            minLevel = parseInt(matcher.group(1));
+            maxLevel = matcher.group(2) == null ? minLevel : parseInt(matcher.group(2));
```
magicus, line 35: I think that `lvl:75` should be interpreted as `lvl:75-♾️`, not as `lvl:75-75`. I would assume that the basic use here would be "find anything above level X", so this should be easy to do. If we really want to find *exactly* level 75, then I can do `lvl:75-75`.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -438,6 +439,12 @@
   "feature.wynntils.itemFavorite.lootChestCloseOverride.description": "After how many ESC presses should the chest be closed? Set to 0 to disable.",
   "feature.wynntils.itemFavorite.lootChestCloseOverride.name": "Loot Chest Close Override",
   "feature.wynntils.itemFavorite.name": "Favorite Items",
+  "feature.wynntils.itemFilter.level.invalidRange": "%s: Should be an integer or a range, separated by a dash",
+  "feature.wynntils.itemFilter.level.maxLessThanMin": "%s: Min must be less than max",
+  "feature.wynntils.itemFilter.level.usage": "<level> or <min>-<max>",
+  "feature.wynntils.itemFilter.profession.invalidProfession": "%s: invalid profession",
```
magicus: I think you'd like a `name` and `description` for each filter, as well...

common/src/main/java/com/wynntils/services/itemfilter/filters/LevelItemFilter.java
kristofbolyai: I disagree with this, i think level:x should be a constant level. If I want a range, then I put a range. Perhaps allow >60 or <70

common/src/main/java/com/wynntils/services/itemfilter/type/ItemFilter.java
magicus: Yes. We'll fix the other services separately. (Good catch though, I had not noticed we did this.)

common/src/main/java/com/wynntils/services/itemfilter/filters/LevelItemFilter.java
magicus: Ok, fine. 

@Etrenak If we go this route, then we can save the extension to support `lvl:>75` to a follow up PR, if you want.

Etrenak:
I don't know if you prefer a temporary solution or no description at all in the guide screen for the moment. I have a commit at hand to temporarily add the "info" widget from `ContainerSearchFeature` to the search bar in the guides, but it's a bit dirty.

common/src/main/java/com/wynntils/services/itemfilter/filters/LevelItemFilter.java
Etrenak: I agree with this.

Etrenak:
Just as a reminder, here is the "info" widget I'm talking about : 
image

common/src/main/java/com/wynntils/screens/base/WynntilsListScreen.java
magicus: @kristofbolyai I am also not sure what you mean here...

common/src/main/java/com/wynntils/screens/guides/WynntilsGuideScreen.java
magicus: I think it might be acceptable to skip the explanation for now. This PR has been open long enough; it is now of sufficiently high quality to be merged imho, so then it's better to merge it and fix additions as follow up PRs.

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * Returns a filter for the given alias, or an error string if the alias does not match any filter.
+     * @param alias an alias of the filter
+     * @return the filter, or an error string if the alias does not match any filter
+     */
+    public ErrorOr<? extends ItemFilter> getFilter(String alias) {
+        Optional<ItemFilter> filterOpt = filters.stream()
+                .filter(filter ->
+                        filter.getAliases().contains(alias) || filter.getName().equals(alias))
```
magicus: Flipping this or would feel a bit more natural to me, and assuming that the name is the most commonly used way to refer to the filter, it will also be slightly more efficient. (Not that it matters really)

common/src/main/java/com/wynntils/services/itemfilter/ItemFilterService.java
```diff
@@ -0,0 +1,186 @@
...
+
+        return new ItemSearchQuery(
+                queryString, itemFilters, ignoredCharIndices, validFilterCharIndices, errors, plainTextTokens);
+    }
+
+    /**
+     * Checks if the given item matches the search query. The item must match all filters and contain all plain text
+     * tokens. Therefore, if there are no plain text tokens, and no filters, this would be considered a match.
+     * <br>
+     * If the item is not a WynnItem, this method always returns false.
```
magicus, line 132: Commendable with the javadoc! We have been quite sloppy at this. For most methods it is better to skip it, but the public methods of services, models and other components, we should really have them.

common/src/main/java/com/wynntils/services/itemfilter/ProfessionItemFilter.java
magicus: I totally forgot about this! Sorry! 😊 Here is the PR: https://github.com/Wynntils/Artemis/pull/2015

magicus (approved):
Except for the open discussions here on Github, this looks fine to merge. Thank you for your hard work!

magicus:
To be clear:

> I don't know if you prefer a temporary solution or no description at all in the guide screen for the moment. I have a commit at hand to temporarily add the "info" widget from ContainerSearchFeature to the search bar in the guides, but it's a bit dirty.

In this case, I prefer no description. We can help you work on getting a description a non-dirty way in a separate PR.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchWidget.java
Etrenak: @kristofbolyai Does it seem good to you ? I understand that it may be wrong to delegate to the superclass in case of an unexpected state (i.e. query != input), but I could not figure out a better way. It's either rendering it with wrong syntax colouring or no colouring at all.

Etrenak:
@magicus, could you please let me know your final word about methods returning read-only collections even though it's not done elsewhere ?

Just as a personnal reminder, I forgot about adding the setting to disable advanced search in containers. Either way I should be able to finish the final additions in the next few days.

magicus:
> could you please let me know your final word about methods returning read-only collections even though it's not done elsewhere ?

If I have to chose, I say: no, remove that. I looked at the actual code now, from what I can tell this is done in two places. One is the aliases for filters. They will in practice be initialized using List.of() which is immutable anyway. 

The other one is the list of filters. If you really think it is important that it is immutable, you could make it such at filter registration time. But I think it is looking for the wrong kind of technical protection against a very unlikely problem. 

Had we designed a library that were used by random applications, my stance would be quite different, but here I think we can assume that callers know how to responsibly handle a returned list.

kristofbolyai:
If you fix the merge and the discussions we can merge this.

Etrenak:
> If you fix the merge and the discussions we can merge this.

Great! I currently don't have my computer with me but it should be done before Tuesday.

kristofbolyai:
Actually, would you mind if I took over and finished this tomorrow? I don't want to delay this (and I kind of depend on this for something). @Etrenak

Etrenak:
> Actually, would you mind if I took over and finished this tomorrow? I don't want to delay this (and I kind of depend on this for something). @Etrenak

Please just let me push the work I've done so far. It should resolve all conversations, and I rebased on top of main

kristofbolyai:
Okay. I will wait then. No problem.

Etrenak:
Well what a terrible idea that was. Not quite sure how to handle this properly. I guess I should have merged instead of rebasing

kristofbolyai:
It's alright. I will review this tomorrow, and merge it. (I only wanted to do the rebase as a fixup, then merge it). Thank you for working on this :)

Etrenak:
The only thing missing is the config to toggle advanced search. Appart for the rebase (oops), I only did minor changes resolving the last conversations. 

Thank you very much for your patience and your reviews during the whole process!

magicus:
A good rule of thumb is to rebase until you publish a PR, and from then on, just merge.

kristofbolyai:
Merged in #2035. I had to make a new PR, I couldn't push to this one.

PR #1866, opened 2023-07-27
feat: Guild model chat detection

ryanzhoudev:
I said I was going to do this a while ago, but forgot.


common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -4,54 +4,108 @@
...
+    // Test suite: https://regexr.com/7hob9
+    private static final Pattern GUILD_NAME_MATCHER = Pattern.compile("§3([a-zA-Z ]*?)§b \\[[a-zA-Z]{3,4}]");
 
-    private String name = "";
-    private GuildRank rank;
+    // Test suite: https://regexr.com/7hobf
+    // Recruiter is intentionally first, otherwise Recruit will incorrectly match first
+    // This applies to all other patterns in this class involving guild ranks
+    private static final Pattern GUILD_RANK_MATCHER =
+            Pattern.compile("§7Rank: §f(Recruiter|Recruit|Captain|Strategist|Chief|Owner)");
```
magicus: I think if you use the start/end of line markers like this, it will work in the normal rank order:

```
            Pattern.compile("^§7Rank: §f(Recruit|Recruiter|Captain|Strategist|Chief|Owner)$");
```

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -4,54 +4,108 @@
...
+
+        Matcher joinedGuildMatcher = message.getMatcher(MSG_JOINED_GUILD);
+        if (joinedGuildMatcher.matches()) {
+            guildName = joinedGuildMatcher.group(1);
+            guildRank = GuildRank.RECRUIT;
+            WynntilsMod.info("User joined guild " + guildName + " as a " + guildRank);
+            return;
+        }
+
+        if (message.startsWith("§3[INFO]§b ") && message.contains(McUtils.playerName() + "'s guild rank from")) {
```
magicus: What about user names that end on `s`? Sometime Wynn does like `mag_icus's` and sometimes `mag_icus'`.  I can probably help you test, but I'd need bolyai to help with changing my rank.

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -4,54 +4,108 @@
...
+            // Username is checked first with simple string comparison
+            // Both versions of the 's string are checked as they will not include the color code
+            // Then determine rank with regex
+            // Username cannot be included in regex as at compile time it is unknown
```
magicus: No, but you could add a `(.+)` (or more specific `([a-zA-Z0-9_]+)` or whatever are valid user name characters) and then check the group that it is equal to the player name.

That would be preferable, this is hard to read and understand what we are matching.

kristofbolyai (commented):
What is this used for? Preparation for something?

ShadowCat117:
> What is this used for? Preparation for something?

I believe it's for updating guild info in the book as they said they would implement this part of it since I was unable to leave/join my guild to test this

magicus:
@kristofbolyai It will keep the guild model up to date when your guild status changes.

common/src/main/java/com/wynntils/models/players/GuildModel.java
ryanzhoudev: Yeah it does, but this does not work for the MSG_RANK_CHANGED regex. Do you want to use ^$ or just keep it consistent and do recruiter first?

common/src/main/java/com/wynntils/models/players/GuildModel.java
ryanzhoudev: I am not sure. If you two could try then let me know, I would highly appreciate it

kristofbolyai: I can help, if we organize something today or tomorrow

kristofbolyai: image

common/src/main/java/com/wynntils/models/players/GuildModel.java
magicus: Why does it not? It looks like it should work just as well there.

magicus: If you have no end of line marker, then the last Recruit will match before Recruiter, but if you end with `$` it should be all fine.

magicus: (As a side note: @kristofbolyai This is one of the reasons why I prefer to always use the explicit `^...$` construct)

ryanzhoudev: Oh yeah the first ncg doesn't matter. Okay I will change it.

magicus (approved):

PR #1869, opened 2023-07-28
fix: Typing occurs in unfocused input boxes

### LINKED ISSUE #1692: you can type in the bank search widget even if its not selected
when you select the search widget and then click the background the caret isn't shown anymore but you can still type in the text field.

ryanzhoudev:
Fixes #1692


magicus (approved):

Mcrtin:
This solution is a bit sketchy and is also related to #ticket-981 on Discord. You can also still use backspace. The root cause is basically, that we depend on the focused state of the TextboxScreen, disregarding the focused widget of the screen.

magicus:
@Mcrtin You can reopen #1692 and publish a PR with a better solution, if you want. I agree that we have probably not been handling focus really correct; I have never ever tried to use tab to navigate our screens. If no one has tested that, it is likely to be broken.

PR #1871, opened 2023-07-28
chore: Start using spotless on gradle files

magicus:


kristofbolyai (approved):
Good PR, also we can at least we can try to see if Github releases a build correctly now..

PR #1873, opened 2023-07-28
feat: Add option to stack duplicate effects

atiedebee:
Quick preview of what this implementation looks like:
stacked_effects

I'm not sure what colour the prefix should have, so I kept it as is for now. Shouldn't be too hard to change.

This pr also adds an option to sort the status overlay list, making it a lot easier to navigate.


common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -32,6 +36,12 @@ public class StatusEffectsOverlay extends Overlay {
     @RegisterConfig
     public final Config<Float> fontScale = new Config<>(1.0f);
 
+    @RegisterConfig
+    public final Config<Boolean> doStackEffects = new Config<>(true);
+
+    @RegisterConfig
+    public final Config<Boolean> doSortEffects = new Config<>(true);
+
```
kristofbolyai: You can name these "*Effects", no need for do.

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
```diff
@@ -87,7 +90,18 @@ public void onTabListCustomization(PlayerInfoFooterChangedEvent event) {
...
+            StyledText modifier;
+
+            // Split the modifier and name, which are separated by a space
+            StyledText[] modifierStr = name.split("\\s{1}");
+            if (modifierStr[0].matches(MODIFIER_REGEX)) {
+                modifier = modifierStr[0];
+                name = name.substring(modifier.length()).trim(); // Get all but first part of the string
+            } else {
+                modifier = StyledText.EMPTY;
+            }
```
kristofbolyai: I would like to see you create a new group in the original STATUS_EFFECT_PATTERN and just get it, without doing this whole modification. 

Is there a reason why you didn't do that? I can help you with it if needed.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -9,15 +9,18 @@
...
+    public StatusEffect(StyledText name, StyledText modifier, StyledText displayedTime, StyledText prefix) {
         this.name = name;
         this.displayedTime = displayedTime;
         this.prefix = prefix;
-        this.fullName =
-                StyledText.concat(prefix, StyledText.fromString(" "), name, StyledText.fromString(" "), displayedTime);
+        this.modifier = modifier;
+
+        final StyledText space = StyledText.fromString(" ");
+        this.fullName = StyledText.concat(prefix, space, modifier, space, name, space, displayedTime);
```
kristofbolyai: The newline and extracting into "space" is arguably messier, so please revert that. Thanks.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +133,38 @@ private void updateTextRenderSetting() {
     protected TextRenderSetting getTextRenderSetting() {
         return textRenderSetting;
     }
+
+    private Stream<StatusEffectWithProperties> stackEffects(List<StatusEffect> effects) {
+        LinkedHashMap<String, StatusEffectWithProperties> effectsToRender = new LinkedHashMap<>();
```
kristofbolyai:
```suggestion
        Map<String, StatusEffectWithProperties> effectsToRender = new LinkedHashMap<>();
```

Should be good to lower this type?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +133,38 @@ private void updateTextRenderSetting() {
...
+
+        private StatusEffectWithProperties(
+                StyledText name, StyledText modifier, StyledText displayedTime, StyledText prefix) {
+            super(name, modifier, displayedTime, prefix);
+        }
+
+        private StatusEffectWithProperties(StatusEffect effect) {
+            this(effect.getName(), effect.getModifier(), effect.getDisplayedTime(), effect.getPrefix());
+        }
+    }
```
kristofbolyai: Personally, I think this class has little use and you could just use `Pair<Integer, StatusEffect>` (since it is not used in many places).

You don't necessary have to do this, but I think it would be a bit clearer.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +133,38 @@ private void updateTextRenderSetting() {
     protected TextRenderSetting getTextRenderSetting() {
         return textRenderSetting;
     }
+
+    private Stream<StatusEffectWithProperties> stackEffects(List<StatusEffect> effects) {
+        LinkedHashMap<String, StatusEffectWithProperties> effectsToRender = new LinkedHashMap<>();
+
+        for (var effect : effects) {
```
kristofbolyai: Don't use var, please. Thank you.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +133,38 @@ private void updateTextRenderSetting() {
     protected TextRenderSetting getTextRenderSetting() {
         return textRenderSetting;
     }
+
+    private Stream<StatusEffectWithProperties> stackEffects(List<StatusEffect> effects) {
```
kristofbolyai: I don't think this method is fully correct. I might have the same effect, with different timers. Or, I might have the same effect, with different modifiers, and this doesn't take it into account, does it?

kristofbolyai (commented):
I wanted to add this for a while, great to see it happening. Code needs a little adaptation, both in style and logic. Should be easy to handle. Let me know if anything needs further explaining.

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
atiedebee: I tried it with groups, but all strings that were accessed using the `.group()` methods lost all their Styles/colors

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
atiedebee: Done, the newline was put there by the formatter I think

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: I have some reasons for not using the `Pair` class:

- This way it is easier to add more options later, like stacking effects with the same name but a different modifier
-  The `Pair` class contains immutable data, which doesn't fit this use case because the integer needs to be incremented. It would require either creating a new Pair on each iteration or creating an `ImmutablePair` type.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Noted

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: `.asString()` returns the fullName from the StyledText class, which includes the modifier and timer.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Doesn't impact the usage, done

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: When creating these I had some local variables that cached them as to not call it in a loop. I've changed the names now

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
atiedebee: I have given it another try and can confirm that `.group()` messes up the styles. The `name` part and the `modifier` part share the same token for giving the style, which would have to be included in the capture group in the regex.

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
```diff
@@ -87,7 +90,18 @@ public void onTabListCustomization(PlayerInfoFooterChangedEvent event) {
             StyledText prefix = parts.get(0);
             StyledText name = parts.get(1);
             StyledText displayedTime = parts.get(2);
-            newStatusEffects.add(new StatusEffect(name, displayedTime, prefix));
+            StyledText modifier;
+
+            // Split the modifier and name, which are separated by a space
+            StyledText[] modifierStr = name.split("\\s{1}");
```
kristofbolyai:
```suggestion
            StyledText[] modifierStr = name.split("\\s");
```

Isn't this the same?

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
kristofbolyai: Okay, then, let's keep this open for now and I will think about this.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
kristofbolyai: Did you not push yet? I can't see this being fixed.

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
atiedebee: Oh right, I was just blindly copying what was above

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
atiedebee: Haven't pushed yet. It also appears that the formatter puts each argument on a new line now:

```java
        this.fullName = StyledText.concat(
                prefix,
                StyledText.fromString(" "),
                modifier,
                StyledText.fromString(" "),
                name,
                StyledText.fromString(" "),
                displayedTime);
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
kristofbolyai: Okay then. Let's rename this class to something easier to read. Perhaps RenderedStatusEffect.

Secondly, what do you think of not extending status effect, but rather having it as an immutable field? That is closer to what I want / what we have done before.

Lastly, please make the count field private. Fields are private. If needed, use a getter.

kristofbolyai: Also, `getTextToRender` could easily be moved here as a member method.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
kristofbolyai: Yeah, thats okay.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
kristofbolyai: And as for the `getTextToRender` format. Can you make it so it's something like "+2x40% Attack Speed"?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -54,8 +64,23 @@ public void onStatusChange(StatusEffectsChangedEvent event) {
...
+        if (doStackEffects.get()) {
+            effectWithProperties = stackEffects(effects);
+        } else {
+            effectWithProperties = effects.stream().map(StatusEffectWithProperties::new);
+        }
+
+        if (doSortEffects.get()) {
+            // Sort effects based on their prefix and their name
+            effectWithProperties = effectWithProperties.sorted(Comparator.comparing(
+                    e -> e.getPrefix().getString() + e.getName().getString()));
```
kristofbolyai: Instead of doing this in-line, please make `StatusEffect` implement `Comparator<StatusEffect>`. I advise you to use `ComparisonChain` in your implementation. Also, perhaps we could also sort by modifier?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: I'll do my best, the proposed changes seem reasonable. Should I put the class in a separate file or keep it here?

kristofbolyai: Keep it here for now.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Sorting by modifier is a nice idea. I'd think the sorting priority could be as follows:

1. prefix (the icon)
2. name (to group similar effects together)
3. modifier

atiedebee: I have applied most of the changes, though I am not familiar with the `Comparator` interface. So I am assuming my current usage is incorrect.

I implemented the interface like this:

```java
        @Override
        public int compare(StatusEffect effect, StatusEffect t1) {
            return ComparisonChain.start()
                    .compare(effect.getPrefix().getString(), t1.getPrefix().getString())
                    .compare(effect.getName().getString(), t1.getName().getString())
                    .compare(effect.getModifier().getString(), t1.getModifier().getString())
                    .result();
}
```

Which required me to use it like this:

```java
            effectWithProperties = effectWithProperties.sorted(((t1, t2) -> t1.compare(t1.getEffect(), t2.getEffect())));
```

So I'm guessing I don't really understand it

kristofbolyai: Sorry for misleading you. The correct interface is `Comparable`, which will allow you to use `.sorted(Comparator.comparing(...))`.

The chain implementation itself looks good, although I am not sure how well string-sorting works for the modifier, which is a number by essence.

atiedebee: We can extract the first number in the modifier like this:
```java
        private static final Pattern DOUBLE_PATTERN = Pattern.compile("\\d+(\\.\\d+)?");

        public static Double getModifierNumber(StyledText modifier){
            Matcher m = DOUBLE_PATTERN.matcher(modifier.getStringWithoutFormatting());
            if( !m.find() ){
                return 0.0;
            }
            String numberStr = DOUBLE_PATTERN.matcher(modifier.getStringWithoutFormatting()).group();
            return Double.parseDouble(numberStr);
        }
```
and use it like this:

```java
        @Override
        public int compareTo(@NotNull StatusEffect effect) {
            return ComparisonChain.start()
                    .compare(this.effect.getPrefix().getString(), effect.getPrefix().getString())
                    .compare(this.effect.getName().getString(), effect.getName().getString())
                    .compare(getModifierNumber(this.effect.getModifier()), getModifierNumber(effect.getModifier()))
                    .result();
        }
```

Which would use the first number in the modifier string and compare it

kristofbolyai: No, we shouldn't do that. It's messy, adds a new regex, etc. Just sort the string for now, and if it becomes problematic, we will see and fix it.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +135,59 @@ private void updateTextRenderSetting() {
...
+        }
+
+        @Override
+        public int compareTo(@NotNull RenderedStatusEffect effect) {
+            return ComparisonChain.start()
+                    .compare(this.effect.getPrefix().getString(), effect.effect.getPrefix().getString())
+                    .compare(this.effect.getName().getString(), effect.effect.getName().getString())
+                    .compare(this.effect.getModifier().getString(), effect.effect.getModifier().getString())
+                    .result();
+        }
```
kristofbolyai: I might not have been clear the first time around, but please implement this in the real `StatusEffect` class.

Also, we disallow using `@NotNull`, so please remove it.

atiedebee: Alright

The `@NotNull` was inserted there by IntellIJ, I'll move it.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +135,59 @@ private void updateTextRenderSetting() {
...
+
+        private RenderedStatusEffect(StyledText name, StyledText modifier, StyledText displayedTime, StyledText prefix) {
+            this.effect = new StatusEffect(name, modifier, displayedTime, prefix);
+        }
+
+        private RenderedStatusEffect(StatusEffect effect) {
+            this.effect = effect;
+        }
+
+        private StyledText getTextToRender() {
```
kristofbolyai: `getRenderedText` sounds a bit more natural.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +135,59 @@ private void updateTextRenderSetting() {
...
+        return effectsToRender.values().stream();
+    }
+
+    private static final class RenderedStatusEffect implements Comparable<RenderedStatusEffect> {
+        private int count = 0;
+        private final StatusEffect effect;
+
+        private RenderedStatusEffect(StyledText name, StyledText modifier, StyledText displayedTime, StyledText prefix) {
+            this.effect = new StatusEffect(name, modifier, displayedTime, prefix);
+        }
```
kristofbolyai: Why does this exist? The second constructor should suffice?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +135,59 @@ private void updateTextRenderSetting() {
...
+
+        private RenderedStatusEffect(StatusEffect effect) {
+            this.effect = effect;
+        }
+
+        private StyledText getTextToRender() {
+            return this.count > 1
+                    ? StyledText.fromString(this.count + "x ").append(this.effect.asString())
```
kristofbolyai: Before it gets lost, this is yet to be changed. ("2x30% Attack Speed")

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +134,51 @@ private void updateTextRenderSetting() {
...
+        }
+
+        public void setCount(int c) {
+            this.count = c;
+        }
+
+        public StatusEffect getEffect() {
+            return this.effect;
+        }
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +134,51 @@ private void updateTextRenderSetting() {
...
+        public int getCount() {
+            return this.count;
+        }
+
+        public void setCount(int c) {
+            this.count = c;
+        }
+
```
kristofbolyai, line 207: You added this but you still use
```
            entry.count += 1;
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Right, it was used before, but its not needed anymore

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: I can put the "2x" before the modifier (making it something like "<icon> 2x <modifier> <etc>")

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Ah, forgot about that one. I thought it would give me errors where it changed but I guess it has access to the inner parts of the class.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: I'm guessing the "2x" part still needs the same colour as the rest of the message? I tried to do something like that but didn't find how to.

kristofbolyai:
```suggestion
                    ? StyledText.fromString(ChatFormatting.GRAY + this.count + "x ").append(this.effect.asString())
```

(or DARK_GRAY, I don't remember)

kristofbolyai: (also, please don't resolve the conversation if the change wasn't done)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -941,6 +941,10 @@
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.fontScale.description": "How large should the status text be?",
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.fontScale.name": "Font Scale",
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.name": "Status Overlay",
+  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.sortEffects.description": "Should effects be sorted based on their type?",
```
```suggestion
  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.sortEffects.description": "Should effects be sorted based on their type, name and modifier?",
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Alright, I think the conversation it was in had some other things that were added so I resolved it. Will be more careful next time

atiedebee: I changed the formatting to grey, I do think the visibility suffers a bit from it.
stacked_effects2

kristofbolyai: But it still doesn't use the format I requested. `"+2x30% Attack Speed"` is arguably the cleanest way to show this. Can you implement this?

atiedebee: It has been done.
stacked_effects3

```java
private StyledText getRenderedText() {
            byte[] modifierBytes = this.effect.getModifier().getString().getBytes(StandardCharsets.UTF_8);
            StyledText modifierText;

            if( modifierBytes.length > 0 && this.count > 1) { // Modifier and stacks
                int start;
                for (start = 0; !Character.isDigit(modifierBytes[start]) && start < modifierBytes.length; start++) {
                }
                start -= 1;
                StyledText modifierStart = this.effect.getModifier().substring(0, start);
                StyledText modifierEnd = this.effect.getModifier().substring(start);
                modifierText = modifierStart
                                .append(StyledText.fromString(ChatFormatting.GRAY + (this.count + "x")))
                                .append(modifierEnd);

            } else if( this.count > 1 ){ // No modifier, does stack
                modifierText = StyledText.fromString(ChatFormatting.GRAY + (this.count + "x"));
            } else { // No stacking
                modifierText = this.effect.getModifier();
            }

            return this.count > 1
                    ? this.effect
                            .getPrefix()
                            .append(StyledText.fromString(" "))
                            .append(modifierText)
                            .append(this.effect.getName())
                            .append(StyledText.fromString(" "))
                            .append(this.effect.getDisplayedTime())
                    : this.effect.
}
```
Not the fanciest java code I've written

kristofbolyai: Oh, god. Wouldn't a simpler way do it?

```
let s = this.effect.getModifier().getString()
let index = 0
if s startsWith - or + then index = 1
s.insert(index, this.count + "x")
```

atiedebee: That depends, is there any other prefixes than `+` and `-`?

kristofbolyai: Have you seen any? I have not.

kristofbolyai: Or nothing is a possible prefix.

atiedebee: I simplified it a bit, knowing there'll only be a +/-/nothing
```java

        private StyledText getRenderedText() {
            String modifierString = this.effect.getModifier().getString();
            StyledText modifierText;

            int minusIndex = modifierString.indexOf('-');
            int plusIndex = modifierString.indexOf('+');
            int index = max(minusIndex, plusIndex); // look for either a - or a +
            if( index < 0 ){// We can simply put the count string at the start
                modifierText = (this.count > 1 ? StyledText.fromString(ChatFormatting.GRAY + (this.count + "x")) : StyledText.EMPTY)
                        .append(this.effect.getModifier());
            } else { // The count string is inserted between the +/- and the number
                index += 1;
                modifierText = StyledText.fromString(modifierString.substring(0, index))
                        .append(this.count > 1 ? StyledText.fromString(ChatFormatting.GRAY + (this.count + "x")) : StyledText.EMPTY)
                        .append(StyledText.fromString(ChatFormatting.GRAY + modifierString.substring(index)));
            }

            return this.count > 1
                    ? this.effect
                            .getPrefix()
                            .append(StyledText.fromString(" "))
                            .append(modifierText)
                            .append(this.effect.getName())
                            .append(StyledText.fromString(" "))
                            .append(this.effect.getDisplayedTime())
                    : this.effect.asString();
        }
```

kristofbolyai: Can you push this so I can comment on it properly?

1, You should return early if the count is 1.
2, index < 0 should be index == -1
3, move the comments to be above the lines, not "next" to them
4, you do a lot of StyledText.fromString and append calls. Use string concatenation and call StyledText.fromString once. It will make the code cleaner (talking about the if-else calls).

atiedebee: It was pushed when I wrote the message.

I've processed all of these points in the newly pushed code now.

kristofbolyai:
Can you run `./gradlew spotlessApply` and commit it?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +136,74 @@ private void updateTextRenderSetting() {
...
+        for (StatusEffect effect : effects) {
+            RenderedStatusEffect entry = effectsToRender.get(effect.asString().getString());
+            if (entry == null) {
+                entry = new RenderedStatusEffect(effect);
+                effectsToRender.put(effect.asString().getString(), entry);
+            }
+            entry.setCount(entry.getCount() + 1);
+        }
+        return effectsToRender.values().stream();
```
kristofbolyai:
```suggestion
    private Stream<RenderedStatusEffect> stackEffects(List<StatusEffect> effects) {
        Map<String, RenderedStatusEffect> effectsToRender = new LinkedHashMap<>();

        for (StatusEffect effect : effects) {
            RenderedStatusEffect entry = effectsToRender.get(effect.asString().getString());
            
            if (entry == null) {
                entry = new RenderedStatusEffect(effect);
                effectsToRender.put(effect.asString().getString(), entry);
            }
            
            entry.setCount(entry.getCount() + 1);
        }
        
        return effectsToRender.values().stream();
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +136,74 @@ private void updateTextRenderSetting() {
...
+    }
+
+    private static final class RenderedStatusEffect {
+        private int count = 0;
+        private final StatusEffect effect;
```
kristofbolyai:
```suggestion
        private final StatusEffect effect;

        private int count = 0;
```

finals and not finals are usually ordered and are separate

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +136,74 @@ private void updateTextRenderSetting() {
...
+
+        private RenderedStatusEffect(StatusEffect effect) {
+            this.effect = effect;
+        }
+
+        private StyledText getRenderedText() {
+            if (this.count == 1) {
+                return this.effect.asString();
+            }
+            String modifierString = this.effect.getModifier().getString();
```
kristofbolyai, line 169:
```suggestion

            String modifierString = this.effect.getModifier().getString();
```

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -108,4 +136,74 @@ private void updateTextRenderSetting() {
...
+
+            // look for either a - or a +
+            int minusIndex = modifierString.indexOf('-');
+            int plusIndex = modifierString.indexOf('+');
+            int index = max(minusIndex, plusIndex);
+
+            // We can simply put the count string at the start
+            if (index == -1) {
+                modifierText = StyledText.fromString(ChatFormatting.GRAY + (this.count + "x"))
+                        .append(this.effect.getModifier());
```
kristofbolyai:
```suggestion
            if (index == -1) {
                 // We can simply put the count string at the start
                modifierText = StyledText.fromString(ChatFormatting.GRAY + (this.count + "x"))
                        .append(this.effect.getModifier());
```

atiedebee: Should I just commit all the changes you proposed?

kristofbolyai: yeah

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -4,6 +4,8 @@
  */
 package com.wynntils.overlays;
 
+import static java.lang.Math.max;
```
kristofbolyai: Oh, sorry I did not notice this before. We don't allow using static imports.

atiedebee: Ah I'll fix that quickly. I should get my IDE configured to follow some of these rules

kristofbolyai (approved):
And this looks great now! Thank you for going through the review process, it was a lot of back and forth. Kind of happens with new people.. Overall, it was pleasant working with you, considering you are new here.

kristofbolyai:
If @magicus wants to take a look, go ahead. If you are fine with my review, you can enable merging.

atiedebee:
> And this looks great now! Thank you for going through the review process, it was a lot of back and forth. Kind of happens with new people.. Overall, it was pleasant working with you, considering you are new here.

Thanks a lot for guiding me through the process. I hope to be of use in the future as well. It has been a pleasure so far.

PR #1876, opened 2023-07-28
feat: Add the ability to close the Wynntils menu with 'e' (or your corresponding hotkey)

Mcrtin:


magicus:
Do we really want this? I am skeptical. The wynn menu is not an inventory. You cannot close vanilla screens with the inventory key.

Mcrtin:
It is as big as the inventory and also frequently used and you can close the vanilla content book with that hotkey.

Mcrtin:
It is as big as the inventory and also frequently used and you can close the vanilla content book with that hotkey. Alternatively if that's not already a thing we can change the key to the one you are opening the screen with.

magicus:
I'd like to hear input from more developers on this.

kristofbolyai:
Since we have the keyboard as a search functionality in many screens, I don't think this is needed. However, I close inventories with ESC..

Mcrtin:
what about changing the key to the one you set for opening the screen?

magicus:
What is the problem of just using esc?

ryanzhoudev:
I am also not a fan of this change. It feels like a half measure and inconsistent. As bolyai mentioned, screens with search do not close, and additionally, secondary screens like the waypoint management and overlay list also do not close with e.

magicus:
@Mcrtin Sorry, but with three developers against, I am going to close this.

If you want to contribute more fixes, it might not be a bad idea to first open an issue describing what you want to achieve.

PR #1881, opened 2023-07-28
fix: Change À to Á in translation replacement

Rafii2198:
It appears the Á symbol was the issue, not Á according to the logs


MenacingPerson (commented):
I think you forgot to change the comment above

kristofbolyai:
So we used the bad character previously..?

Rafii2198:
Yes, it was because of my mistake, I looked at my logs and saw À and thought it was used during a cutscene to make my screen go black, but now I got a log with actual translations, and it had Á. The message is `Área descoberta: North Nivla Woods (25 XP)`.
That was my fault, I apologise.

kristofbolyai (approved):

PR #1882, opened 2023-07-28
feat: improve stat screen formatting

kristofbolyai:
Count and everything else is basically two different statistic types. Maybe this could be reflected in the stored stuff itself, but formatting it is good enough too, IMO. Especially it is a lot easier on us if we want to add new types.

I also added the custom formatter for the readable number formats.

image
image


magicus:
I think COUNT is always relevant, is it not? I mean, you calculate the average by using the count, so it is indirectly present -- why not show it directly to the user as well?

Sure, for some statistics it is not too important, but then again, maybe "min" value is not too important for e.g. "damage dealt" as well, etc. So do not hide it.

magicus (approved):

PR #1883, opened 2023-07-28
fix: Fix text input box not properly focusing

Mcrtin:


kristofbolyai:
@Mcrtin why was the solution sketchy?

Mcrtin:
Because there are because of TextInputScreen two different focused elements (Minecraft also keeps track of the focused element) which can and are causing many bugs

kristofbolyai:
> Because there are because of TextInputScreen two different focused elements?

I have no idea what this means.

What bugs? Which screens, what behavior? I really dislike the solution of fixing simple bugs with these so large and unreviewable PRs, I already kind of ranted about this before, in the config screen widget PR.

Mcrtin:
Minecraft keeps track of the focused element and we made our own way of keeping track of it. That creates unexpected behavior and incompatibility with vanilla and other mods (for example we made our own tab complete instead of using the vanilla one)

kristofbolyai:
I still think this can be solved much easier. Can you let me know actual bugs and how to reproduce them? Which screen, etc?

Mcrtin:
what I can see after some quick testing is that tab doesn't work

Mcrtin:
when selecting the timestamp format box and then clicking a button you can write in it without it showing what you write

Mcrtin:
when searching for "time" in the settings screen and scrolling down the tooltips of the last three configurables don't appear

Mcrtin:
EditBox and TextInputWidget are incompatible and can't appear on the same screen

kristofbolyai:
> what I can see after some quick testing is that tab doesn't work

What is the expected behavior? What screen?

> EditBox and TextInputWidget are incompatible and can't appear on the same screen

Where is this a problem?

> when searching for "time" in the settings screen and scrolling down the tooltips of the last three configurables don't appear

How is this even connected to the issue we are talking about? Overlays never had tooltips in the first place.

Mcrtin:
> > what I can see after some quick testing is that tab doesn't work
> 
> What is the expected behavior? What screen?

It works nowhere correctly. You can see the expected behavior in the Minecraft config screen (the tab system Minecraft has is quite complex).


> > EditBox and TextInputWidget are incompatible and can't appear on the same screen
> 
> Where is this a problem?

it is currently not a problem I just wanted to bring that up because it could potentially become a problem in the future.

> > when searching for "time" in the settings screen and scrolling down the tooltips of the last three configurables don't appear
> 
> How is this even connected to the issue we are talking about? Overlays never had tooltips in the first place.

I didn't know that. I just noticed it.

Mcrtin:
i've noticed that i did some unnecessary changes that i can move to another pr (or drop)

kristofbolyai:
> i've noticed that i did some unnecessary changes that i can move to another pr (or drop)

It would be really helpful, otherwise we can't review/accept it. Generally try to limit the changes you do so they are all needed to achieve your goal (PR title).

Mcrtin:
i had to make many changes in `SettingsScreen` and `ConfigTile` because Minecraft focusing system doesn't allow nested widgets. The search widget is also not always focused anymore. if that's a problem I could add a ctrl f listener to focus it.

kristofbolyai:
Most likely I will need to close this and revisit it after/during the 1.20 porting. 1.20 changes a lot of screen related stuff, and I would rather just do this refactor myself, if needed.

Mcrtin:
i'll review then like i did in the quick cast pr

PR #1885, opened 2023-07-28
chore: Add @RegisterStorage annotation to all Storage fields

magicus:
It does not do anything as such, but clearly marks that the field is persisted, and it aligns with @RegisterConfig


kristofbolyai (approved):
I assume you tested we have all annotations correct.

kristofbolyai:
> I assume you tested we have all annotations correct.

I should not have merged this without confirming..
```
 java.lang.RuntimeException: A Storage is missing @RegisterStorage annotation:public final com.wynntils.core.storage.Storage com.wynntils.features.wynntils.ChangelogFeature.lastShownVersion
 ```

PR #1886, opened 2023-07-29
fix: Fix bank jump buttons in housing storage items

P0keDev:
Turns out I over-complicated the previous fix and made it break for non-bank storage menus--despite having fewer pages, all menus have the same quick jump buttons in vanilla Wynncraft, so the `getDefaultJumpDestinations` method wasn't necessary.


common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -121,18 +121,14 @@ private void getCustomJumpDestinations() {
...
-            customJumpDestinations = getDefaultJumpDestinations();
+            switch (currentContainer) {
+                case BANK -> customJumpDestinations = QUICK_JUMP_DESTINATIONS;
+                case BLOCK_BANK -> customJumpDestinations = BLOCK_BANK_DESTINATIONS;
+                default -> customJumpDestinations = HOUSING_DEFAULT_DESTINATIONS;
+            }
         }
     }
 
-    private List<Integer> getDefaultJumpDestinations() {
```
magicus: I still think this refactor was nice, even if it is called from just one place.

kristofbolyai:
I'll let @JamieCallan117 review this, I have not much idea what the correct behavior is.

ShadowCat117:
Looks good, I don't have the issue with block bank anymore. Up to you if you want to keep the `getDefaultJumpDestinations` method, it does make it more obvious what it's doing so I'd say keep that

kristofbolyai (approved):

magicus (approved):

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
magicus: I fixed it for ya :)

PR #1887, opened 2023-07-29
feat: Dump functions to CSV files

ryanzhoudev:
Yes I committed a plaintext password. I already reset it.

Automatically generates the db for https://wynntils-functions.donkeyblaster.tk/docs
I am happy with how the Java side of this works, so I will mark this for review. The website itself, and if we want to transfer it to the org can be discussed elsewhere.


magicus:
I am not very happy to include a postgres dependency. Can't you just dump the SQL to a text file and then import it into your database? That way you also don't need to worry about credentials.

Or possibly even better, write it to a json file. I am confident that there are tons of tools out there that can convert a simple json file into SQL insert statements.

ryanzhoudev:
I will see what I can do about json instead, but if it does not work out I am also okay with keeping this on a separate branch or something like that

magicus:
That seems much better. But it's not a CSV file, is it? Looks like pure SQL.

ryanzhoudev:
There are two generated CSV files. dumpFunctionsToCSV() and dumpArgumentsToCSV() both call writetoCSV(), which writes to arguments.csv and functions.csv. The column names are stored in two LinkedHashMaps at the top, and the related SQL for generating the tables is stored with them. The last function creates one massive SQL query that clears the db, generates the types enum from the types in Artemis, creates the tables according to the columns saved in the LinkedHashMaps, and copies to clipboard.

magicus (approved):

PR #1892, opened 2023-07-30
fix: update map hash and fix reloading caches

### LINKED ISSUE #1674: Map tiles are removed from cache even though we cannot download new
China blocked the github cdns but I used vpn to make my wynntils successfully download the cache files. item guide and map both are fine if I turn vpn on, but map would be black if I turn it off
And I cannot use vpn to play game because vpn has lantency issue

in-game: https://photos.app.goo.gl/tFVbcb4fX7qFs6xF7
file: https://photos.app.goo.gl/wZ11iz6CmNus98E57

**logs**
with vpn: https://hst.sh/uwuxobukej
without vpn: https://hst.sh/uquhurisog

kristofbolyai:
I've investigated https://github.com/Wynntils/Artemis/issues/1674 and I can't find anything else that would cause loading to fail. I've tried to load without internet, and it seems to just work fine after the new hash. I already committed it to the API repo.

Fixes https://github.com/Wynntils/Artemis/issues/1674

### DIFF
--- a/common/src/main/java/com/wynntils/core/WynntilsMod.java
+++ b/common/src/main/java/com/wynntils/core/WynntilsMod.java
@@ -79,6 +79,7 @@ public final class WynntilsMod {
     public static void reloadAllComponentData() {
         componentMap.get(Manager.class).forEach(c -> ((Manager) c).reloadData());
         componentMap.get(Model.class).forEach(c -> ((Model) c).reloadData());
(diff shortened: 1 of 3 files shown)


magicus (approved):

magicus:
Let's try this. We'll need to get confirmation from the bug reporters if this actually solves their problem.

PR #1894, opened 2023-07-30
feat: add a small-map in the caves screen

kristofbolyai:
I am not sure if this should be kept, but I just added this small detail to fill the space.

image


magicus:
Oh, looks very nice! Maybe also draw the cave icon at the actual cave location?

magicus:
When we fix raids, boss altars, dungeons we can add the map to them as well.

magicus (approved):

kristofbolyai:
> Oh, looks very nice! Maybe also draw the cave icon at the actual cave location?

I tried it but it looked weird with the icon.

magicus:
Ok

PR #1898, opened 2023-07-30
feat: Add functions for lootrun beacon counts (and lootrun state)

kristofbolyai:
This is not a small PR. It adds a lot of stuff that can be used by us in the future:
- We can now track Wynn beacons
- We can track lootrun states
- We know how many of each beacon color we collected in the current lootrun

The only thing I did not do is persisting this in a storage, I will do that as a follow up.


common/src/main/java/com/wynntils/mc/mixin/ClientPacketListenerMixin.java
```diff
@@ -555,11 +559,26 @@ private void handleRemoveMobEffectPost(ClientboundRemoveMobEffectPacket packet,
 
     @Inject(
             method = "handleAddEntity(Lnet/minecraft/network/protocol/game/ClientboundAddEntityPacket;)V",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "Lnet/minecraft/client/multiplayer/ClientPacketListener;postAddEntitySoundInstance(Lnet/minecraft/world/entity/Entity;)V",
+                            shift = At.Shift.AFTER))
+    private void handleAddEntity(ClientboundAddEntityPacket packet, CallbackInfo ci, @Local Entity entity) {
```
kristofbolyai, line 568: For those wondering (and so you dont have to open your IDEA while reviewing), this is the same mixin, but we are not calling it at RETURN, but at the last instruction. Original code is like

```
if (entity != null) ...
else log error
```

we need the entity local to still exist, and also this fixes null entities posting this event. win-win I guess.

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
```diff
@@ -0,0 +1,62 @@
...
+    BLUE((int) (0.125 * 32)),
+    PURPLE((int) (0.15625 * 32)),
+    GRAY((int) (0.1875 * 32)),
+    ORANGE((int) (0.21875 * 32)),
+    RED((int) (0.25 * 32)),
+    DARK_GRAY((int) (0.28125 * 32)),
+    WHITE((int) (0.3125 * 32)),
+    CYAN((int) (0.34375 * 32)),
+    RAINBOW((int) (0.375 * 32));
+
```
kristofbolyai: Obviously I got this from the resource pack, but I wanted to keep the format (damagePredicate * itemDurability)

ryanzhoudev:
Lootrun ending from skill issue (dying too much, running out of time) does not reset the state

kristofbolyai:
> Lootrun ending from skill issue (dying too much, running out of time) does not reset the state

That's weird.. Btw, Do you know if we should decrement the count if the user dies mid challenge?

ryanzhoudev:
> > Lootrun ending from skill issue (dying too much, running out of time) does not reset the state
> 
> That's weird.. Btw, Do you know if we should decrement the count if the user dies mid challenge?

I am not sure if it counts towards the number of beacons selected by the user. I will try to find out

ryanzhoudev:
We should not remove them, they still count towards total beacon numbers

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
```diff
@@ -0,0 +1,62 @@
...
+public enum BeaconColor {
+    YELLOW((int) (0.09375 * 32)),
+    BLUE((int) (0.125 * 32)),
+    PURPLE((int) (0.15625 * 32)),
+    GRAY((int) (0.1875 * 32)),
+    ORANGE((int) (0.21875 * 32)),
+    RED((int) (0.25 * 32)),
+    DARK_GRAY((int) (0.28125 * 32)),
+    WHITE((int) (0.3125 * 32)),
+    CYAN((int) (0.34375 * 32)),
```
ryanzhoudev: Technically this should be aqua

ryanzhoudev: Also we are missing green?

kristofbolyai: Yes, you are right. I missed green because it is a different item, as it is also used for content tracking. Now this class is complete.

kristofbolyai:
> Lootrun ending from skill issue (dying too much, running out of time) does not reset the state

I am not sure how you got this to happen. I checked timing out and it does reset. I assume dying does too, since we check for scoreboard removal.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -83,6 +103,67 @@ public void onChatMessage(ChatMessageReceivedEvent event) {
     public void onWorldStateChanged(WorldStateEvent event) {
         lootrunCompletedBuilder = null;
         lootrunFailedBuilder = null;
+
+        // FIXME: Persist this in a later PR.
```
kristofbolyai, line 107: I tried persisting but it is a bit harder than it looks.. So I will hold off on doing that for now. I will come back to it soon.

ryanzhoudev (approved):

kristofbolyai (commented):
I self-reviewed this PR before merging. Obviously, it is not as good as a review from someone else, but DonkeyBlaster's and my review is enough for me to merge this. I am happy to do any post-merging cleanups, if needed.

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.List;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import org.apache.commons.compress.utils.Lists;
+
+public enum BeaconColor {
+    GREEN((int) (0.78125 * 32), Items.GOLDEN_SHOVEL, ContentType.BOTH),
```
magicus, line 15: Why these 0.xxx * 32 instead of normal integers?

kristofbolyai: They are damage values taken from wynns resource pack

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import net.minecraft.core.Position;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EntityType;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class BeaconModel extends Model {
+    // Amount of armor stands above each other to consider this a beacon
+    private static final int VERIFICATION_ENTITY_COUNT = 6;
```
magicus, line 29: Is this a good margin? How many are there typically? 7 or 27?

kristofbolyai: There are like 128/7.5 of them. This is more than enough.

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -0,0 +1,123 @@
...
+        for (VerifiedBeacon removedBeacon : removedBeacons) {
+            verifiedBeacons.remove(removedBeacon);
+            WynntilsMod.postEvent(new BeaconEvent.Removed(removedBeacon));
+        }
+    }
+
+    private boolean isDuplicateBeacon(Position position) {
+        return verifiedBeacons.stream().anyMatch(verifiedBeacon -> {
+            Position beaconPosition = verifiedBeacon.getPosition();
+            return beaconPosition.x() == position.x() && beaconPosition.z() == position.z();
```
magicus, line 110: Maybe VectorUtils.equalsDiscardingY?

kristofbolyai: I guess yes, that can be added.

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
magicus: And there they are represented as floats times 32? 

Even if they are, can't we just replace them with the normal integer damage value? To me, a damage value of `3` makes sense, while `0.09375 * 32` does not.

kristofbolyai: I wanted to preserve how the resource pack has it which is 0-1 damage * itemDurability. But if you want to, we can make these integers

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -0,0 +1,123 @@
...
+
+        boolean correctPosition = unverifiedBeacon.addEntity(entity);
+
+        if (!correctPosition) {
+            unverifiedBeacons.remove(unverifiedBeacon);
+            return;
+        }
+
+        if (unverifiedBeacon.getEntities().size() == VERIFICATION_ENTITY_COUNT) {
+            BeaconColor beaconColor = BeaconColor.fromUnverifiedBeacon(unverifiedBeacon);
```
magicus, line 62: It could be that you are just a bit lucky that this works. Donkey introduced a tick delay in ShamanTotemModel:

```
                    // These must be ran with a delay, as inventory contents are set a couple ticks after the totem
                    // actually spawns
```

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
```diff
@@ -0,0 +1,83 @@
...
+                return color;
+            }
+        }
+
+        return null;
+    }
+
+    public enum ContentType {
+        LOOTRUN,
+        CONTENT,
```
magicus, line 76: This actually means QUEST, right?

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
kristofbolyai: I don't think so. They are always built up in order, and we also allow a delay. 1000 ms. See timed set.

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
kristofbolyai: well caves do have beacons as well

PR #1900, opened 2023-07-31
fix: Wrap config button descriptions

ShadowCat117:
Before and after


kristofbolyai (commented):
Good catch! But please use `com.wynntils.utils.mc.ComponentUtils#wrapTooltips` to make this much simpler. After that, this is good!

ShadowCat117:
Changed in ConfigurableButton as well, since that was using `RenderedStringUtils.wrapTextBySize` too

kristofbolyai (approved):

PR #1901, opened 2023-07-31
feat: Transcribe into and from Gavellian and Wynnic

ShadowCat117:
Feel free to test it as I've tried all I can think of besides an NPC speaking Gavellian as I think only Lari does and I don't have a class close to one of those quests.
GavellianAndWynnicTranslation

Also, I can look into capitalising the first letter of a new sentence and i's by themselves if it seems worth it


common/src/main/java/com/wynntils/core/components/Services.java
```diff
@@ -38,4 +39,5 @@ public final class Services {
     public static final TranslationService Translation = new TranslationService();
     public static final UpdateService Update = new UpdateService();
     public static final WynntilsAccountService WynntilsAccount = new WynntilsAccountService();
+    public static final WynnLanguageService WynnLanguageSerivce = new WynnLanguageService();
```
kristofbolyai:
```suggestion
    public static final WynnLanguageService WynnLanguage = new WynnLanguageService();
```

common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java
```diff
@@ -0,0 +1,162 @@
...
+        Optional<DiscoveryInfo> foundDiscoveryInfo = discoveryInfoStream
+                .filter(discoveryInfo -> discoveryInfo.getName().equals(nameToFind))
+                .findFirst();
+
+        return foundDiscoveryInfo.map(DiscoveryInfo::isDiscovered).orElse(false);
+    }
+
+    public void setSelectedLanguage(WynnLanguage selectedLanguage) {
+        this.selectedLanguage = selectedLanguage;
+    }
```
kristofbolyai: Perhaps we should reset this after each message sent, so you don't need to set that many defaults..

kristofbolyai (commented):
I will need to review this later separately.

common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java
ShadowCat117: I did think that as opening the chat screen for commands via / should reset it but I couldn't figure out how to do that. This is just how legacy behaved

kristofbolyai: Oh. Okay then, but It kind of feels messy. Maybe magicus has a better idea later.

common/src/main/java/com/wynntils/mc/event/EditBoxInsertEvent.java
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+@Cancelable
+public class EditBoxInsertEvent extends Event {
```
Mcrtin, line 11: Where is this used?

ShadowCat117: It's for chat input

common/src/main/java/com/wynntils/screens/translation/widgets/WynnLanguageButton.java
```diff
@@ -0,0 +1,80 @@
...
+        if (Services.WynnLanguage.getSelectedLanguage() == wynnLanguage) {
+            return CommonColors.GREEN;
+        } else {
+            return CommonColors.WHITE;
+        }
+    }
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (!isMouseOver(mouseX, mouseY)) return false;
```
Mcrtin: this shouldn't be needed since it should only be called by ContainerEventHandler and that does the check

common/src/main/java/com/wynntils/screens/translation/widgets/WynnLanguageButton.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Services.WynnLanguage.setSelectedLanguage(wynnLanguage);
+
+        return false;
```
Mcrtin: return super#mouseClicked and is there a reason why you are not using onPress()?

ShadowCat117: We don't want it to be focused so it returns false, same as ChatTabs

common/src/main/java/com/wynntils/screens/translation/widgets/WynnLanguageButton.java
ShadowCat117: It is required, otherwise wherever you click it always selects the Gavellian button

common/src/main/java/com/wynntils/screens/translation/widgets/WynnLanguageButton.java
ShadowCat117: As for the onPress, it doesn't trigger when you click it

Mcrtin: That's strange. Maybe onClick works. Why does that method even exist when it doesn't trigger

Mcrtin: oh... you always return false and don't call super so it obviously doesn't get called

ShadowCat117: Oh yeah that would do it. Chat Tabs do the same which is where I copied most of the code for this button from, don't reinvent the wheel as they say

magicus:
Maybe I'm picky here, but I think we should try to use the technically correct term "transliteration" instead of "translation".

common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java
```diff
@@ -0,0 +1,162 @@
...
+import com.wynntils.utils.mc.McUtils;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.item.ItemStack;
+
+public class WynnLanguageService extends Service {
```
magicus: Since this is about the Wynn world, it should be a Model, not a service.

The respective base classes give these descriptions to help tell them apart:

```
 * a Model.

 * Services provide support to consumers for anything that is not specifically
 * related to the Wynncraft model, either since it is an external service
 * (3rd party library, or system/environment interaction), or since it is something
 * that we augment the Wynncraft world with, but the Wynncraft servers has no knowledge
 * about. 
 ```

**Comment by magicus** [comment]
I started looking through this, but it mostly became skimming. Will have to return later.

One immediate reflection is that the feature is very heavy. Maybe see if there are some more functionality that can be extracted to the model? That is, if it makes sense; we should not force stuff into the model for the sake of it.

File: common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java (hunk 6 above)
Comment (ShadowCat117) [reply common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java]: Ah didn't know they had a description, should probably look at those base classes

Comment (Mcrtin) [reply common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java]: can you add that to for example the contribution guide line

**Comment by ShadowCat117** [comment]
> I started looking through this, but it mostly became skimming. Will have to return later.
> 
> One immediate reflection is that the feature is very heavy. Maybe see if there are some more functionality that can be extracted to the model? That is, if it makes sense; we should not force stuff into the model for the sake of it.

Yeah I did originally have most of it in the feature before I moved a lot into the Service (Model), I'll see if anything else can be moved there

File: common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java (hunk 6 above)
Comment (magicus) [reply common/src/main/java/com/wynntils/services/wynnlanguage/WynnLanguageService.java]: You can also check out the overall architecture document: https://github.com/Wynntils/Artemis/wiki/Class-and-package-structure

It's not completely up to date (it does not list Services, for instance) but it is a good start.

**Comment by ShadowCat117** [comment]
Ok, I'm not sure if anything else is worth moving to the model without making the methods take in even more parameters but let me know if you think there's any that can be moved/don't belong

File: common/src/main/java/com/wynntils/features/commands/CustomCommandKeybindsFeature.java [hunk 7]
```diff
@@ -96,6 +98,8 @@ public class CustomCommandKeybindsFeature extends Feature {
             () -> this.executeKeybind(keybindCommand6.get(), commandType6.get()));
 
     private void executeKeybind(String keybindCommand, CommandType commandType) {
+        Models.WynnLanguage.setSelectedLanguage(WynnLanguage.DEFAULT);
```
magicus: This bugs me. We got to find a solution to not do this. 

I need to dig deeper into the patch to fully understand why this is here and how to properly resolve it. But it got to go.

common/src/main/java/com/wynntils/models/wynnlanguage/WynnLanguage.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.wynnlanguage;
+
+public enum WynnLanguage {
```
magicus: Just as this is a transcription, not a translation, those are not languages, but scripts (or alphabets).

While "script" is the more technically correct term, I think "alphabet" is a more approachable term (and "script" tends to mean like short simple programs in computer contexts).

So I think we should call this enum Alphabet (or WynnAlphabet, if we really need that precision), and the model (and package name) should be AlphabetModel (or WynnAlphabetModel).

Let's save "Wynn Language" until the day the quest dialogues starts appearing in Quenya. ;-)

common/src/main/java/com/wynntils/models/wynnlanguage/WynnLanguageModel.java
```diff
@@ -0,0 +1,320 @@
...
+    private static final List<Character> wynnicNumbers =
+            List.of('⑴', '⑵', '⑶', '⑷', '⑸', '⑹', '⑺', '⑻', '⑼', '⑽', '⑾', '⑿');
+    private static final Map<Character, Character> englishToGavellianMap = new HashMap<>();
+    private static final Map<Character, Character> englishToWynnicMap = new HashMap<>();
+    private static final Map<Character, Character> gavellianToEnglishMap = new HashMap<>();
+    private static final Map<Character, Character> wynnicToEnglishMap = new HashMap<>();
+    private static final Pattern NUMBER_PATTERN = Pattern.compile("\\d+");
+    private static final String GAVELLIAN_TRANSCRIBER_DISCOVERY = "Ne du Valeos du Ellach";
+    private static final String WYNNIC_TRANSCRIBER_DISCOVERY = "Wynn Plains Monument";
+    private static final StyledText GAVELLIAN_TRANSCRIBER = StyledText.fromString("§rHigh Gavellian Transcriber");
```
magicus: Oh, right, Wynn already uses the term "transcription" (which is a synonym to "transliteration"). We should stick to that instead of "transliteration".

common/src/main/java/com/wynntils/features/commands/CustomCommandKeybindsFeature.java
ShadowCat117: Yeah bolyai pointed it out too and I don't like it either but it was just a slight improvement over legacy behaviour.
In legacy, it preserves your selected language after sending a message, so this just resets it when the chat screen is opened from a feature that won't want the chat to be translated.

Easy fix is to simply not preserve the language but then if you wanted to send repeated messages in one of the languages, you'd have to click the button each time unless you used the brackets system.

common/src/main/java/com/wynntils/features/utilities/GavellianAndWynnicTransliterationFeature.java
```diff
@@ -0,0 +1,441 @@
...
+import net.minecraft.client.gui.screens.ChatScreen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.Style;
+import net.minecraftforge.eventbus.api.Event;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import org.lwjgl.glfw.GLFW;
+
+@ConfigCategory(Category.UTILITIES)
+public class GavellianAndWynnicTransliterationFeature extends Feature {
```
magicus: This is quite a mouthful. Maybe just `WynnTranscriptionFeature` or `AutoTranscribeFeature`?

common/src/main/java/com/wynntils/models/wynnlanguage/WynnLanguageModel.java
magicus: (Sorry for requesting all these renames, but I think a good design is about 80% having a proper and well-functioning terminology)

magicus:
I looked at the feature and no, not much else can be moved to the model. However, I wonder if this maybe is two different features in disguise. One feature converts incoming text, in chat or npc dialogues. And one feature allows you to type in one of the wynn alphabets.

I think it might make sense to split them up. It looks like they are fairly separated inside the feature already, and I think it is reasonable to assume that you might want to view transcriptions without being able to type yourself or clutter the chat bar with buttons.

If this helps bring down the complexity I think it is a good thing. But if it just makes the code worse with lot of interdependencies, or stuff that is shared between the features but still does not seem to fit in the model, then it is probably a bad idea. But maybe you can give it a shot?

common/src/main/java/com/wynntils/features/commands/CustomCommandKeybindsFeature.java
magicus: So what you are saying is that you really want to catch when the chat edit line appears, and restore the previous language then, but you cannot do that, so you always restore the previous language, and then cancel that for known places where this does not make sense? Did I get it right?

common/src/main/java/com/wynntils/models/wynnlanguage/WynnLanguageModel.java
ShadowCat117: Not a problem, definitely needs good terminology in case someone else comes along to change things. And at least they're easy changes :)

ShadowCat117:
> I looked at the feature and no, not much else can be moved to the model. However, I wonder if this maybe is two different features in disguise. One feature converts incoming text, in chat or npc dialogues. And one feature allows you to type in one of the wynn alphabets.
> 
> I think it might make sense to split them up. It looks like they are fairly separated inside the feature already, and I think it is reasonable to assume that you might want to view transcriptions without being able to type yourself or clutter the chat bar with buttons.
> 
> If this helps bring down the complexity I think it is a good thing. But if it just makes the code worse with lot of interdependencies, or stuff that is shared between the features but still does not seem to fit in the model, then it is probably a bad idea. But maybe you can give it a shot?

I'll give it a shot, there'll definitely be some things shared but since a lot of the transcribing is in the model I think it shouldn't be too bad

common/src/main/java/com/wynntils/features/utilities/GavellianAndWynnicTransliterationFeature.java
ShadowCat117: If it's going to be 2 features, I suppose I'll have to think of 2 different names. Any suggestions? :)

common/src/main/java/com/wynntils/features/commands/CustomCommandKeybindsFeature.java
ShadowCat117: So you send a message and once you hit enter, whatever alphabet you last had selected is what the Model remembers so that next time the chat screen is opened it will automatically select that alphabet so that you can continue typing in the alphabet you last were. That's just how legacy behaved. 

But in certain circumstances, such as executing a custom command, or the trade market auto opening chat, you likely don't want that to be typing in anything other than the default alphabet, so it resets to default. I also wanted it to reset to default when you hit / to open the chat screen with / already entered but couldn't figure out how.

I only had it remember as legacy had it and this is a port so wanted to make it function the same but with improvements, but I'm not sure if it's actually worth having as even if you meant to send in a different alphabet, people can still see your message.

common/src/main/java/com/wynntils/features/utilities/GavellianAndWynnicTransliterationFeature.java
magicus: InputTranscriptionFeature and TranscribeMessagesFeature ... maybe? 😰

ShadowCat117: Better than what I can come up with :D

common/src/main/java/com/wynntils/core/components/Models.java
```diff
@@ -79,7 +79,7 @@ public final class Models {
     public static final WarModel War = new WarModel();
     public static final WorldStateModel WorldState = new WorldStateModel();
     public static final WynnItemModel WynnItem = new WynnItemModel();
-    public static final WynnLanguageModel WynnLanguage = new WynnLanguageModel();
+    public static final WynnAlphabetModel WynnLanguage = new WynnAlphabetModel();
```
magicus: Also rename the variable

common/src/main/java/com/wynntils/features/utilities/GavellianAndWynnicTranscriptionFeature.java
```diff
@@ -44,22 +44,22 @@
 import org.lwjgl.glfw.GLFW;
 
 @ConfigCategory(Category.UTILITIES)
-public class GavellianAndWynnicTransliterationFeature extends Feature {
+public class GavellianAndWynnicTranscriptionFeature extends Feature {
     @RegisterConfig
-    public final Config<Boolean> transliterateChat = new Config<>(true);
+    public final Config<Boolean> transcriptChat = new Config<>(true);
```
magicus: The verb is "to transcribe", so "transcribeChat"...

ShadowCat117:
I've gotten rid of the behaviour where it remembers your previous selected alphabet so no more resetting to default in other features.

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
```diff
@@ -0,0 +1,287 @@
...
+    private static final int MAX_CHAT_LENGTH = 256;
+    // Numbers higher than this will be replaced with "∞"
+    private static final int MAX_TRANSCRIPTABLE_NUMBER = 5000;
+    private static final Pattern NUMBER_PATTERN = Pattern.compile("\\d+");
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent event) {
+        if (useBrackets.get()) return;
+
+        if (event.getScreen() instanceof ChatScreen chatScreen) {
```
magicus: Maybe make a config for the buttons, in case you only want to use brackets?

ShadowCat117: That's how it currently works. by default brackets will do nothing and you have to use the buttons, toggling the brackets on will remove the buttons and transcribe what you put in brackets

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
```diff
@@ -0,0 +1,287 @@
...
+            Models.WynnAlphabet.setSelectedAlphabet(WynnAlphabet.DEFAULT);
+        }
+    }
+
+    private void addAlphabetButton(ChatScreen chatScreen, int xOffset, WynnAlphabet alphabet) {
+        chatScreen.addRenderableWidget(new WynnAlphabetButton(xOffset, chatScreen.height - 14, 12, 12, alphabet));
+    }
+
+    private boolean containsBrackets(String message) {
+        return (message.contains("{") && message.contains("}")) || (message.contains("<") && message.contains(">"));
```
magicus: I am worried this is too specific and might be surprising. What about this syntax instead?
`{{wynnic text}} and <<gavellian>>`

It is not much harder to type, but much less likely to be typed by mistake.

Also, for `{}`, we should think about how that plays with the function parsing. It would be nice if we could insert wynnic text in info boxes (for the real hardcore players! :-)) without breaking the function parsing. 

So possibly use `[[ wynnic ]]` instead, to avoid that completely. Or check that the function string parser does not choke on `{{` but can pass it on unaltered.

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
ShadowCat117: ~~Unless you think always transcribing what's in brackets regardless of if the buttons are enabled or not is a good idea~~

Seen suggestion in other comment

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    private String transcriptSentMessage(String message, Event event) {
```
magicus: I think this, with some minor modifications, can be a candidate to move to the model. That basically moves the bracket thing from a feature-specific functionality to a system-wide functionality. I think that would be a good thing. 

(Some collateral damage will result but I'm sure you can figure that out)

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
ShadowCat117: I've not actually looked into functions so might take me a bit to figure out, but I'll certainly give it a try

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
ShadowCat117: Works quite nicely if you nest them too.
image

common/src/main/java/com/wynntils/features/utilities/InputTranscriptionFeature.java
ShadowCat117: I've put it in here, is this the best place for it?

```
public StyledText[] doFormatLines(String templateString) {
        StringBuilder resultBuilder = new StringBuilder();

        if (Models.WynnAlphabet.containsBrackets(templateString)) {
            templateString = Models.WynnAlphabet.transcribeBracketedText(templateString);
        }
```

image

kristofbolyai: Can you just add a new generic function for this? I would rather not have this that deeply implemented in functions

ShadowCat117: That is a much better idea (:

common/src/main/java/com/wynntils/features/chat/InputTranscriptionFeature.java
```diff
@@ -0,0 +1,242 @@
...
+        String message = event.getMessage();
+
+        if (Models.WynnAlphabet.containsBrackets(message)) {
+            String updatedMessage = Models.WynnAlphabet.transcribeBracketedText(message);
+
+            updatedMessage = updatedMessage.substring(0, Math.min(updatedMessage.length(), MAX_CHAT_LENGTH));
+
+            if (!updatedMessage.equals(message)) {
+                event.setCanceled(true);
+                McUtils.mc().getConnection().sendChat(updatedMessage);
```
magicus: I just pushed McUtils.sendChat() to main. The asymmetry with commands here killed me :)

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -0,0 +1,173 @@
...
+                    .map(s -> ((Component) s.getComponent()))
+                    .toList();
+
+            Managers.TickScheduler.scheduleNextTick(() -> {
+                NpcDialogEvent transcriptedEvent = new WynnTranscriptedNpcDialogEvent(
+                        transcriptedComponents, event.getType(), event.isProtected());
+                WynntilsMod.postEvent(transcriptedEvent);
+            });
+        } else {
+            NpcDialogEvent transcriptedEvent =
```
magicus, line 97: Can this ever happen? If there is no message, then Models.WynnAlphabet.hasWynnicOrGavellian can hardly return true, can it?

ShadowCat117: Shouldn't happen, originally got this from translation feature and didn't really think about it

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -0,0 +1,173 @@
...
+
+        event.setMessage(modified.getComponent());
+    }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onNpcDialogue(NpcDialogEvent event) {
+        if (!transcribeNpcs.get()) return;
+        if (!Models.WynnAlphabet.hasWynnicOrGavellian(event.getChatMessage().toString())) return;
+
+        boolean transcribeWynnic = Models.WynnAlphabet.shouldTranscribe(transcribeCondition.get(), WynnAlphabet.WYNNIC);
```
magicus, line 82: Since shouldTranscribe() is in the model, the condition enum needs to move there as well. Please put it as a separate file under the "type" subpackage.

common/src/main/java/com/wynntils/models/wynnalphabet/WynnAlphabetModel.java
```diff
@@ -0,0 +1,368 @@
...
+            if (gavellianCharacters.contains(c) || wynnicCharacters.contains(c) || wynnicNumbers.contains(c)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public boolean containsBrackets(String message) {
+        return (message.contains("[[") && message.contains("]]")) || (message.contains("<<") && message.contains(">>"));
```
magicus: To add a bit more layer of certainty that this matches correctly (and possibly clarity to the code), you can replace this with an regex, like `(\[\[.*\]\])|(<<.*>>)` or something like that.

magicus (commented):
This is like almost done now. Just a few minor fixes left.

It is looking really nice! You have provided a solid model, and two complex features that go with it. Great work!

common/src/main/java/com/wynntils/models/wynnalphabet/WynnAlphabetModel.java
```diff
@@ -263,7 +264,9 @@ public boolean hasWynnicOrGavellian(String message) {
     }
 
     public boolean containsBrackets(String message) {
-        return (message.contains("[[") && message.contains("]]")) || (message.contains("<<") && message.contains(">>"));
+        Pattern bracketPattern = Pattern.compile("(\\[\\[.*\\]\\])|(<<.*>>)");
+
+        return bracketPattern.matcher(message).matches();
```
magicus: I think you need to use find(), or you need to add some extra .* around the sides, since afaik matches() will only accept a complete match. Also please test it to verify that I did not inadvertently trick you into writing something that do not work!

Also, extract the Pattern to a constant to save some time compiling it (micro-optimization, but we do it as practice).

ShadowCat117: Ah did a quick test with just the brackets but not with stuff outside them, will double check

ShadowCat117: Works now. Unless you or anyone else wants to do some more tests it should be good to go

magicus: Your testing is fine. Thanks!

magicus (approved):

PR #1902, opened 2023-07-31
feat: Add an (opt-in) feature to collect lootrun task location data

kristofbolyai:
Depends on #1898.

This PR adds an feature so the community can collect lootrun data, and we can display it on the map.

### DIFF
--- a/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
+++ b/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
@@ -112,7 +112,7 @@ public class LootrunModel extends Model {
 
     // When we get close to a beacon, it get's removed.
     // This is our signal to know that this can be the current beacon,
-    // but we don't know for sure until scoreboard to confirms it.
+    // but we don't know for sure until the scoreboard confirms it.
     @SubscribeEvent
     public void onBeaconRemove(BeaconEvent.Removed event) {
         VerifiedBeacon beacon = event.getBeacon();
--- a/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
+++ b/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
@@ -111,7 +111,8 @@ public class LootrunModel extends Model {
     }
 
     // When we get close to a beacon, it get's removed.
-    // This is our signal to know that this is the current beacon, but we use the scorebaord to confirm it.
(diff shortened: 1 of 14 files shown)


common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -83,6 +109,86 @@ public void onChatMessage(ChatMessageReceivedEvent event) {
...
+        lootrunningState = LootrunningState.NOT_RUNNING;
+        currentLocation = null;
+        currentTaskType = null;
+        currentLootrunBeacons = new HashMap<>();
+        currentBeacon = null;
+    }
+
+    // When we get close to a beacon, it get's removed.
+    // This is our signal to know that this can be the current beacon,
+    // but we don't know for sure until scoreboard to confirms it.
```
Mcrtin:
```suggestion
    // but we don't know for sure until scoreboard confirms it.
```

kristofbolyai: 2 am moment, I can't speak

common/src/main/java/com/wynntils/models/lootrun/type/LootrunLocation.java
```diff
@@ -0,0 +1,28 @@
...
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.lootrun.type;
+
+import com.wynntils.utils.type.BoundingBox;
+import net.minecraft.core.Position;
+
+public enum LootrunLocation {
+    SILENT_EXPANSE(400, -1150, 1600, -180),
+    CORKUS(-2150, -3500, -1200, -2100);
```
kristofbolyai, line 12: Not the prettiest way to get this location, but this is the best we can do to get this reliably.

ryanzhoudev: Do we need these massive locations? The starting NPC only has a range of like 4-5 blocks radius

kristofbolyai: Yes, since you can class and just rejoin anywhere. These coords include the full regions.

ryanzhoudev (approved):

kristofbolyai (commented):
Just like the PR before this, I self reviewed this also, and I am happy to change anything if needed later.

PR #1903, opened 2023-07-31
feat: Add export waypoints button

ShadowCat117:
Importing now also only imports new POIs

https://github.com/Wynntils/Artemis/assets/8337467/606f0cd2-be58-4875-a09a-fac68dd464aa


common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -134,27 +145,48 @@ private void importFromClipboard() {
...
                 .customPois
-                .get()
-                .addAll(List.of(customPois));
+                .get();
+
+        List<CustomPoi> poisToAdd = Stream.of(customPois)
+                .filter(newPoi -> !existingPois.contains(newPoi))
+                .toList();
+
+        existingPois.addAll(poisToAdd);
```
kristofbolyai, line 165: I thought POIs were in a set so we didn't need this. I guess I was wrong.

kristofbolyai (approved):
Looks good

PR #1908, opened 2023-08-01
fix: Properly map beacon colors to content book activities

magicus:
Depends on #1907.

PR #1909, opened 2023-08-01
refactor: Beacon model refinements

magicus:
Here are some additional changes I'd like to see to the Beacon model. In summary:
* Try to make the gory logic stay inside the model
* Use Location instead of Position (we should strive to use Location whenever we are not interested in sub-block positioning)
* Simplify VerifiedBeacon into a record Beacon

This depends on #1908


common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -78,46 +81,75 @@ public void onEntityAdd(AddEntityEvent event) {
...
+        if (armorSlots.size() != 4) return null;
+
+        ItemStack bootsItem = armorSlots.get(3);
+        return BeaconColor.fromItemStack(bootsItem);
+    }
+
+    private static final class UnverifiedBeacon {
+        private static final float POSITION_OFFSET_Y = 7.5f;
+
+        private final Location location;
```
kristofbolyai, line 130: How can you change this to location when sub-block position matters in the case of Y..?

kristofbolyai:
@magicus can you do the merge?

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
magicus: We only do sub-block checking between the entities, which are stored as a `List<Entity>`, and they each have a `Position`.

The "location" of the unverified beacon is only used to get the correct beacon by matching x and z in getUnverifiedBeaconAt(location).

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -80,46 +82,77 @@ public void onEntityAdd(AddEntityEvent event) {
 
     @SubscribeEvent
     public void onEntityTeleport(TeleportEntityEvent event) {
-        Optional<VerifiedBeacon> verifiedBeaconOpt = verifiedBeacons.stream()
-                .filter(verifiedBeacon -> verifiedBeacon.getBaseEntity().equals(event.getEntity()))
-                .findFirst();
+        if (event.getEntity() == null) return;
```
kristofbolyai: I think we should use mixins so this never happens.

kristofbolyai: Oh, sorry. This is an edge case, you can't do the invoke-shift trick I did before. Maybe only post the event if entity is not null?

kristofbolyai (approved):

PR #1911, opened 2023-08-01
refactor: Generify ConfigHolder

magicus:


common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
```diff
@@ -291,17 +291,18 @@ private boolean isInBound(Position position, Window window) {
 
     // limit the bounding distance to prevent divided by zero in getBoundingIntersectPoint
     @Override
-    protected void onConfigUpdate(ConfigHolder configHolder) {
+    protected void onConfigUpdate(ConfigHolder<?> unknownConfigHolder) {
+        ConfigHolder<Float> configHolder = (ConfigHolder<Float>) unknownConfigHolder;
```
kristofbolyai: This will surely crash if we change any setting other than these. Like enabled.

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -268,8 +268,10 @@ private void clickPreviousPage() {
     }
 
     @Override
-    protected void onConfigUpdate(ConfigHolder configHolder) {
-        String valueString = (String) configHolder.getValue();
+    protected void onConfigUpdate(ConfigHolder<?> unknownConfigHolder) {
+        // All our configs are strings
+        ConfigHolder<String> configHolder = (ConfigHolder<String>) unknownConfigHolder;
```
kristofbolyai: Again, enabled is a config too, which is a boolean.

common/src/main/java/com/wynntils/core/config/ConfigHolder.java
```diff
@@ -143,6 +143,10 @@ public void setValue(Object value) {
         userEdited = true;
     }
 
+    void restoreValue(Object value) {
+        setValue((T) value);
+    }
```
kristofbolyai, line 148: Is this for some future additions?

magicus: Yes, and no. It is needed in this patch, since I could not find any other way to cast the argument to T when reading the value back from a file.

But in my working branch I have unified this behavior with Storage under the same name, as well.

common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
```diff
@@ -28,7 +28,7 @@ protected void onUpdate(String text) {
         Object parsedValue = configHolder.tryParseStringValue(text);
         if (parsedValue != null) {
             if (!parsedValue.equals(configHolder.getValue())) {
-                configHolder.setValue(parsedValue);
+                ((ConfigHolder<Object>) configHolder).setValue(parsedValue);
             }
```
kristofbolyai: We changed `tryParseStringValue` to return T, so I think this can be ported as well.

magicus: The problem here is that we have subclassed TextInputBoxSettingsWidget with CustomColorSettingsButton, which has another actual type in the ConfigHolder (CustomColor instead of String). Due to this, I cannot type the genetic paramenter to String, and that complicates this code.

magicus: Actually, I should probably just generify TextInputBoxSettingsWidget...

kristofbolyai: Perhaps a new PR?

kristofbolyai (approved):
Reviewed, and tested this version. Looks good. Before merging, please test the game by running and try to set a few configs.

magicus:
I tested a random collection of configs, and verified that changing it had the desired effect and were persisted and read back properly.

PR #1912, opened 2023-08-01
feat: Allow an infinite amount of waypoints; content waypoints are different from user waypoints; colored POI waypoints keep their color

kristofbolyai:
image


magicus:
Suggestion: Call your new model "MarkerModel", and let's move away from using "compass" as a term for locations. (It's just odd) 

I think "marker" serves as a good name for those ephemeral POIs.

magicus:
(this way you can skip renaming Compass to OldCompass, and we can just slowly dismantle it.

kristofbolyai:
I will rename the model. Is there any comment on how the base concept is?

magicus:
I think it is fine. It partly aligns with the changes I have made in mapdata, but also partly cuts the same problem space another way. I think you can go ahead with this, and I'll have to worry about the best way to integrate this into the mapdata rewrite when that day comes.

common/src/main/java/com/wynntils/utils/colors/CommonColors.java
```diff
@@ -1,10 +1,13 @@
 /*
- * Copyright © Wynntils 2022.
+ * Copyright © Wynntils 2022-2023.
  * This file is released under AGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.utils.colors;
 
 public final class CommonColors {
+    // Note: Rendering this is not supported, except in BeaconBeamFeature
+    public static final CustomColor RAINBOW = CustomColor.fromInt(0xf0f0f0).withAlpha(0);
```
magicus, line 9: ... a tad hacky, yes? 🤔

common/src/main/java/com/wynntils/models/marker/type/MarkerProvider.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.marker.type;
+
+import java.util.stream.Stream;
+
+public interface MarkerProvider {
+    Stream<MarkerInfo> getCompassInfos();
```
magicus:
```suggestion
    Stream<MarkerInfo> getMarkerInfos();
```

common/src/main/java/com/wynntils/utils/colors/CommonColors.java
kristofbolyai: This is actually rather easy to unhack. We change this to a supplier, and just use the rainbow generation method from beacon beam feature. 1 problem is that there is no place for that event.. otherwise, voila, we have a generic rainbow.

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
```diff
@@ -0,0 +1,77 @@
...
+                        spawnLocation,
+                        new MarkerInfo(
+                                new StaticLocationSupplier(spawnLocation),
+                                Texture.QUESTS_BUTTON,
+                                CommonColors.WHITE,
+                                CommonColors.WHITE));
+    }
+
+    public Location getSpawnInfo() {
+        return spawnInfo.a();
```
magicus, line 36: Either the name needs changing or this returns the wrong thing..?

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
```diff
@@ -5,34 +5,43 @@
...
+    GREEN(25, Items.GOLDEN_SHOVEL, true, ActivityType.STORYLINE_QUEST, CommonColors.GREEN),
+    PINK(24, Items.GOLDEN_SHOVEL, false, ActivityType.BOSS_ALTAR, CommonColors.PINK),
+    YELLOW(3, Items.GOLDEN_PICKAXE, true, ActivityType.RAID, CommonColors.YELLOW),
+    BLUE(4, Items.GOLDEN_PICKAXE, true, ActivityType.LOOTRUN_CAMP, CommonColors.BLUE),
+    PURPLE(5, Items.GOLDEN_PICKAXE, true, ActivityType.MINI_QUEST, CommonColors.PURPLE),
+    GRAY(6, Items.GOLDEN_PICKAXE, true, null, CommonColors.GRAY),
+    ORANGE(7, Items.GOLDEN_PICKAXE, true, ActivityType.CAVE, CommonColors.ORANGE),
+    RED(8, Items.GOLDEN_PICKAXE, true, ActivityType.DUNGEON, CommonColors.RED),
+    DARK_GRAY(9, Items.GOLDEN_PICKAXE, true, null, CommonColors.DARK_GRAY),
+    WHITE(
```
magicus, line 25: What about those long lines..? ;-) (maybe move the // to a separate line)

kristofbolyai: The downside of auto formatters..

kristofbolyai: Actually.. This comment can be removed. Because the code needs to adapt. It is an edge case but we use this to get the color for activity beacons..

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -164,8 +168,14 @@ protected void doInit() {
                         return;
                     }
 
-                    if (Models.Compass.getCompassLocation().isPresent()) {
-                        Location location = Models.Compass.getCompassLocation().get();
+                    List<MarkerInfo> markers = Models.Marker.USER_WAYPOINTS_PROVIDER
+                            .getCompassInfos()
+                            .toList();
+                    if (!markers.isEmpty()) {
+                        int index = markers.indexOf(focusedMarker);
```
magicus, line 179: what if indexOf fails? Maybe have a safety check for -1, even if that "is not supposed to happen".

kristofbolyai: -1 + 1 = 0 = reset the counter. Perfect behavior. I can add a comment for it..

magicus: Actually, maybe save a `int focusedMarkerIndex` instead, and if it is 0 it is okay. And just verify that it has not suddenly gone out of bounds.

common/src/main/java/com/wynntils/models/beacons/type/BeaconColor.java
kristofbolyai: It is sad bit it might have to stay like this.

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
kristofbolyai: Yes, fixed

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -336,17 +346,22 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
...
-            if (McUtils.player().isShiftKeyDown()
-                    && Models.Compass.getCompassLocation().isPresent()) {
-                Location location = Models.Compass.getCompassLocation().get();
+            List<MarkerInfo> markers =
+                    Models.Marker.USER_WAYPOINTS_PROVIDER.getCompassInfos().toList();
+            if (McUtils.player().isShiftKeyDown() && !markers.isEmpty()) {
+                int index = markers.indexOf(focusedMarker);
+                MarkerInfo markerInfo = markers.get((index + 1) % markers.size());
+                focusedMarker = markerInfo;
+                Location location = markerInfo.location();
```
magicus, line 360: Is this the right behavior, to both center and switch to the next marker in the same operation? Maybe it is, I dunno. Does it feel natural when you do this?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: But then we have to check if it is in bounds. I think this is a bit cleaner. But this whole screen needs adaptation, this is just to get it working somewhat reasonably.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -355,18 +370,22 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
                 if (hovered.hasStaticLocation()) {
                     if (hovered instanceof IconPoi iconPoi) {
                         if (iconPoi instanceof CustomPoi customPoi) {
-                            Models.Compass.setCompassLocation(
-                                    new Location(hovered.getLocation()), iconPoi.getIcon(), customPoi.getColor());
+                            Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(
```
magicus, line 384: Are we okay with not clearing the old waypoint in these cases? It is a pretty steep change compared to the old behavior; that means you will have to actively clear the old waypoint, or you'll end up with multiple ones.

I am not sure this is okay. At this point we should probably keep the old behavior and clear all old markers before adding a new. Later on, we can rework the UI so you can select if you want to add a new waypoint, or replace the existing.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -355,18 +370,22 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
...
                         }
                     } else {
-                        Models.Compass.setCompassLocation(new Location(hovered.getLocation()));
+                        Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(new Location(hovered.getLocation()));
                     }
                 } else {
                     final Poi finalHovered = hovered;
-                    Models.Compass.setDynamicCompassLocation(
-                            () -> finalHovered.getLocation().asLocation());
+                    Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(new DynamicLocationSupplier(
```
magicus, line 398: I'm losing my way among all those else clauses, but I guess the same applies to these.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: This feels okay imo, we focus on the next one with a click.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: I guess that could work. But I want this to at least be possible. And /compass clear can just fix any state.

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
magicus, line 1: This was tricky to review. I assumed you have only put everything inside a loop over all markers, but that does not seem to be the case..? Can you shortly describe what you have done?

kristofbolyai: I basically did that but I had to do adaptations to the code. We had many fields that assumed only 1 marker. These were set in an event, then rendered in a second one. Now, we calculate and store these per marker, and render accordingly.

common/src/main/java/com/wynntils/features/map/BeaconBeamFeature.java
```diff
@@ -29,55 +32,96 @@ public class BeaconBeamFeature extends Feature {
...
+    private static final int RAINBOW_CHANGE_RATE = 10;
+    private CustomColor currentRainbowColor = CommonColors.RED;
+
+    @SubscribeEvent
+    public void onTick(TickEvent event) {
+        int r = currentRainbowColor.r;
+        int g = currentRainbowColor.g;
+        int b = currentRainbowColor.b;
+
+        if (r > 0 && b == 0) {
```
magicus, line 44: I get it that this will change the color, but will it be a rainbow? Why not use a HSV model instead?

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
kristofbolyai: `DUMMY_WAYPOINT` is just a hack for us to render the icon correctly without hassle..

common/src/main/java/com/wynntils/utils/colors/CommonColors.java
magicus: You mean like a CustomColorSupplier..?

kristofbolyai: Just a `Supplier<CustomColor`. For this only case. We would get a new value every render call.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: If you check this class now, and the i18n, I added support for the new behavior, and restored the old one.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
magicus: Oh. Very magic. Not sure I like it. But sure, as long as you write a comment.

common/src/main/java/com/wynntils/features/map/BeaconBeamFeature.java
kristofbolyai: Looked pretty rainbow-y to me. I have no experience with HSV, so I am not sure how I would implement that.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
kristofbolyai: Done

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
magicus: I think we will need to add something along the lines of "if you press shift while you click, you create an additional marker, otherwise you'll replace the old".

The problem with that is that I think we have already "used up" many such combos for other purposes, so there are not many left. But maybe like Alt + click for adding instead of replacing?

magicus: That was why I recommended that we keep the old "replace" behavior for now and then we can add a way to set additional markers as a follow up. But if you think you can find some free combo, that is at least somewhat reasonable, go ahead.

kristofbolyai: Shift + old keys should be fine.

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
magicus: Does that not mean that all markers get the same generic icon? I'm trying to understand the code. It seems like you will draw the correct icon when the marker is in view, but otherwise you do `drawScalingTexturedRect` which I do not understand what it does. It takes in a texture, but if it draws the icon, then what does the DUMMY_WAYPOINT rendering do?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1637,14 +1637,16 @@
...
-  "screens.wynntils.map.help.description7": "Left click on the compass marker to remove it.",
-  "screens.wynntils.map.help.description8": "Right click to center on player.",
-  "screens.wynntils.map.help.description9": "Shift-Right click to center on the compass marker.",
+  "screens.wynntils.map.help.description10": "Right click to center on player.",
+  "screens.wynntils.map.help.description11": "Shift-Right click to center on the compass marker.",
+  "screens.wynntils.map.help.description2": "Left click on waypoint to replace the compass marker.",
+  "screens.wynntils.map.help.description3": "Shift + Left click on waypoint to place an additional compass marker.",
+  "screens.wynntils.map.help.description4": "Middle click to replace the compass marker on the map.",
+  "screens.wynntils.map.help.description5": "Shift + Middle click to place an additional the compass marker on the map.",
+  "screens.wynntils.map.help.description6": "Shift-Middle click on the map to create a waypoint.",
```
magicus: These seems to be the same, or conflicting..?

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
kristofbolyai: image
DUMMY_WAYPOINT is used for rendering that small pointer when the marker off screen.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1637,14 +1637,16 @@
...
-  "screens.wynntils.map.help.description6": "Alt-Middle click on a waypoint to remove it.",
-  "screens.wynntils.map.help.description7": "Left click on the compass marker to remove it.",
-  "screens.wynntils.map.help.description8": "Right click to center on player.",
-  "screens.wynntils.map.help.description9": "Shift-Right click to center on the compass marker.",
+  "screens.wynntils.map.help.description10": "Right click to center on player.",
+  "screens.wynntils.map.help.description11": "Shift-Right click to center on the compass marker.",
+  "screens.wynntils.map.help.description2": "Left click on waypoint to replace the compass marker.",
+  "screens.wynntils.map.help.description3": "Shift + Left click on waypoint to place an additional compass marker.",
+  "screens.wynntils.map.help.description4": "Middle click to replace the compass marker on the map.",
+  "screens.wynntils.map.help.description5": "Shift + Middle click to place an additional the compass marker on the map.",
```
magicus: grammar: " an additional the compass "

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: I just removed that part. We can still add multiple waypoints on POIs but not the map. for now.

magicus: The reordering confused me. But it seems that shift+middle click on the map will at the same time both create a new (persisted) waypoint (the old behavior), and add a new (temporary) marker (the new behavior).

But our entire set of operations is just messed up. :-( There is no logic or consistency here... This is why I was worried this will just be messy getting into place in this PR.

common/src/main/java/com/wynntils/utils/colors/CommonColors.java
magicus: I'm not sure I follow you completely where and how this supplier should be used, but if you think you can easily un-hack this then I'm okay with this for now.

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
magicus: Ah, I see! Thanks!

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: Yeah. I am just not sure if we can do anything when people are used to this already.  Maybe the right-click actions list..

common/src/main/java/com/wynntils/features/map/BeaconBeamFeature.java
magicus: Ok, fair enough.

magicus: The thing with HSV is that you can take the Hue (the H in the model name) from 0 to 255 (or 0-1.0 or whatever), and it goes through the entire spectrum of hues, from red to violet. In a way that well matches the human perception of a rainbow. And then it is easy to convert to RGB.

magicus: We use this method in DurabilityArcFeature for example.

kristofbolyai: Perhaps I can fix that when I unhack the rainbow field?

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: We are probably a bit stuck with this for now, yes. At some point we can probably make a more meaningful alternative, and have it as a config to keep the old and weird way. But that is a bigger change.

Did I understand correctly that now the old behavior with replacing is kept for the old "gestures", but there is a new one, and that is shift + left click on an icon to add a marker to it? (but no way to have multiple markers on random places on the map, for now)

common/src/main/java/com/wynntils/features/map/BeaconBeamFeature.java
magicus: Sounds good!

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: Yes, that is how it is.

magicus (approved):
*phew*

I assume you have tested all relevant parts? If so, ship it!

kristofbolyai:
I checked the map, most commands, the rendering. I tracked quests too to check it.

PR #1913, opened 2023-08-01
chore: Restore proper zh_cn translation

magicus:


magicus:
Hopefully this will make its way to Crowdin and replace its bad translations...

PR #1916, opened 2023-08-01
refactor: Fix remaining raw use of generic classes

magicus:


kristofbolyai (commented):
Code looks fine. Let me test it.

kristofbolyai (approved):
Can't see anything broken from launching and checking some settings. Can you test too?

magicus:
I did some in-depth testing of enum json reading, and a generic check that it was not DOA. I think that is enough.

kristofbolyai:
Sounds good.

PR #1920, opened 2023-08-01
feat: Persist lootruns between restarts and class switches (and keep track of every class separately)

kristofbolyai:
2 AM me could not figure this out.. Here we are now. Thanks for the great idea of persisting per class, magicus (in StatisticsService).

### DIFF
--- a/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
+++ b/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
@@ -7,12 +7,16 @@ package com.wynntils.models.lootrun;
 import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Handlers;
 import com.wynntils.core.components.Model;
+import com.wynntils.core.components.Models;
+import com.wynntils.core.storage.RegisterStorage;
+import com.wynntils.core.storage.Storage;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.handlers.chat.event.ChatMessageReceivedEvent;
 import com.wynntils.handlers.chat.type.RecipientType;
 import com.wynntils.models.beacons.event.BeaconEvent;
 import com.wynntils.models.beacons.type.Beacon;
 import com.wynntils.models.beacons.type.BeaconColor;
+import com.wynntils.models.character.event.CharacterUpdateEvent;
 import com.wynntils.models.lootrun.event.LootrunBeaconSelectedEvent;
 import com.wynntils.models.lootrun.scoreboard.LootrunScoreboardPart;
 import com.wynntils.models.lootrun.type.LootrunLocation;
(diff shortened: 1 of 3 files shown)


magicus:
I have been toying with the idea of making this a core functionality of Storage/Config, to offer the ability to be per character instead of per account...

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -70,18 +74,40 @@ public class LootrunModel extends Model {
...
+    // Data that can live in memory, when joining a class we will parse these
     private LootrunningState lootrunningState = LootrunningState.NOT_RUNNING;
     private LootrunLocation currentLocation;
     private LootrunTaskType currentTaskType;
-    private Map<BeaconColor, Integer> currentLootrunBeacons = new HashMap<>();
-    private Beacon currentBeacon;
+
+    // Data to be persisted
+    @RegisterStorage
+    private Storage<Map<String, Map<BeaconColor, Integer>>> currentLootrunSelectedBeaconStorage =
```
magicus: There is a lot of `current` prefixes in the lootrun variables. Maybe this can be implicitly understood?

kristofbolyai: I guess yes.

kristofbolyai: But then maybe take away lootrun as well?

magicus (approved):
I guess this looks like it could work. But perhaps drop the current- ..? (And perhaps also for the pre-existing fields as well)

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
magicus: Yes, you can drop that as well.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -70,18 +74,38 @@ public class LootrunModel extends Model {
...
-    private LootrunLocation currentLocation;
-    private LootrunTaskType currentTaskType;
-    private Map<BeaconColor, Integer> currentLootrunBeacons = new HashMap<>();
-    private Beacon currentBeacon;
+    private LootrunLocation location;
+    private LootrunTaskType taskType;
+
+    // Data to be persisted
+    @RegisterStorage
+    private Storage<Map<String, Map<BeaconColor, Integer>>> selectedBeaconStorage = new Storage<>(new TreeMap<>());
```
magicus: For consistency:

```suggestion
    private Storage<Map<String, Map<BeaconColor, Integer>>> selectedBeaconsStorage = new Storage<>(new TreeMap<>());
```

magicus (approved):

PR #1921, opened 2023-08-01
refactor: Introduce Config.touched()

magicus:
This will help bring Config and Storage closer together.


kristofbolyai (approved):
If you think this is the way to go, okay. But this differs from storage touched, as this is a direct save call, where as storage is "we can't detect changes in objects, only primitives, mark this storage as changed".

magicus:
Yes, this is just a partial solution. I will work on getting them more aligned.

Issue #1917, opened 2023-08-01
DiscordRichPresence will crash when no valid version of GLIBC is found


atiedebee:
```
[19:19:42] [Render thread/ERROR]: Failed to initialize Wynntils features
java.lang.UnsatisfiedLinkError: /tmp/java-discord-game-sdk-32228091688074/libdiscord_game_sdk_jni.so: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/java-discord-game-sdk-32228091688074/libdiscord_game_sdk_jni.so)
```

Simply, the `discord_game_sdk_jni.so` requires version 2.34 of glibc, which is apparently not present on my system. This will crash every time the config is loaded, which includes saving the content book causing other changes not to be saved.

I got a quick workaround working by try-catching the error and disabling the feature, though this is very dirty and I assume a more solution should exist.

The quick and dirty try-catch ([DiscordRichPresenceFeature.java:127](https://github.com/Wynntils/Artemis/blob/3052ca7ab7c9a7c9833f98f53fa11fe5f0e53b06/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java#L127C3-L127C3))
```java
protected void onConfigUpdate(ConfigHolder configHolder) {                                                                                                                                            
         if (this.isEnabled()) {                                                                                                                                                                           
             // This isReady() check is required for Linux to not crash on config change.                                                                                                                  
             if (!Services.Discord.isReady()) {                                                                                                                                                            
                try{                                                                                                                                                                                       
                     // Even though this is in the onConfigUpdate method, it is how the library is first loaded on launch                                                                                  
                     Services.Discord.load(); 
                 }catch(UnsatisfiedLinkError e){                                                                                                                                                           
                     WynntilsMod.getLogger().warn("Failed setting up discord rich presence: {}", e.getMessage());                                                                                          
                     Managers.Feature.disableFeature(this);
                     return; 
                 }                                                                                                                                                                                       
             }                                                                                                                                                                                           

```

The full backtrace:

```
[19:19:42] [Render thread/ERROR]: Failed to initialize Wynntils features
java.lang.UnsatisfiedLinkError: /tmp/java-discord-game-sdk-32228091688074/libdiscord_game_sdk_jni.so: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/java-discord-game-sdk-32228091688074/libdiscord_game_sdk_jni.so)
[...]

PR #1924, opened 2023-08-02
feat: Add crash protection to config updates in features and overlays

magicus:
As reported in https://github.com/Wynntils/Artemis/issues/1917, crashing during a config update can bring down the entire config system, with potentially catastrophic results.


magicus:
I have tested this by hacking in a fake crash in a feature and an overlay.

kristofbolyai (approved):
Good, thanks. This doesn't fix the whole "config system crashes while init" issue, but I think that will stay until the refactor is done.

magicus:
I realized that all abstract methods in our consumers must be crash protected, just as well as all events we send to them. I will check if we have any more of those.

PR #1925, opened 2023-08-02
fix: Fix beacon counter incrementing by two rarely and persisting issues

kristofbolyai:


kristofbolyai:
In hindsight, the first persisting PR had some issues.. This should be the correct implementation now.

PR #1926, opened 2023-08-02
refactor: Bring Config and Storage under a single Persisted roof

magicus:
This is the first real step towards actually unifying them.


common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public List<Pair<Field, Persisted>> getPersisted(PersistedOwner owner, Class<? extends PersistedValue> clazzType) {
+        // Get pairs of field and annotation for all persisted values of the requested type
+        return Arrays.stream(FieldUtils.getFieldsWithAnnotation(owner.getClass(), Persisted.class))
+                .filter(field -> clazzType.isAssignableFrom(field.getType()))
+                .map(field -> Pair.of(field, field.getAnnotation(Persisted.class)))
+                .toList();
+    }
+
+    public void verifyAnnotations(PersistedOwner owner) {
```
kristofbolyai, line 27: Do we catch these now..?

common/src/main/java/com/wynntils/core/storage/StorageManager.java
```diff
@@ -65,46 +65,25 @@ public void initFeatures() {
...
+            String baseName = owner.getStorageJsonName();
+            Storage<?> storage = (Storage<?>) FieldUtils.readField(field, owner, true);
+            String jsonName = baseName + "." + field.getName();
+            storages.put(jsonName, storage);
+
+            Type valueType = Managers.Json.getJsonValueType(field);
+            storageTypes.put(storage, valueType);
+            storageOwner.put(storage, owner);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
```
kristofbolyai, line 86: Is this the time to start catching these somewhere? Currently we would crash the system just like we used to..

common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
magicus: This is a pure refactoring; we do the same checks currently as we do in this patch.

common/src/main/java/com/wynntils/core/storage/StorageManager.java
magicus: Perhaps. The illegal access would only happen if we were not allowed to read the field despite the `force` flag, I think. Which could only happen if we have a securit manager, I think. So it is more of a "won't happen but we need to have this by formality".

Anyway, this is a pure refactoring. Adding extra stability will have to come later.

magicus:
I have verified with some random storage and config variables that this still work properly.

kristofbolyai (approved):
Code looks good. I also tested this.

PR #1928, opened 2023-08-02
refactor: Move config and storage packages to persisted

magicus:
Well... almost done with the sweeping changes, anyway.


kristofbolyai (approved):
Simple refactor, but can you test this? Just to make sure we don't have some bad reflection lying around which IDEA's refactor can't handle.

magicus:
Tested, it seems to work. 

I got the "telemetry?" prompt first time I ran with this PR checked out. However, my config was in a messed-up state before testing (due to my other development) so I assume that was because of my earlier development. 

If you want to be like 105% sure then maybe you should test also.

kristofbolyai:
LGTM from testing

Issue #1758, opened 2023-07-15
archivesBaseName is depracated, will be removed in Gradle 9


magicus:
We must figure out a replacement. I tried following the docs but my gradle/groovy knowledge was too thin for me to understand what I should do, since my trivial attempts failed.

xCykrix:
@magicus Closed by the above merged, just FYI. Didn't link correctly it seems haha. Happy to help though, random knowledge I just happened to already have a problem with recently.

PR #1930, opened 2023-08-02
fix: Deprecation warning for archivesBaseName in Gradle 9.0

xCykrix:
Resolves deprecation alert for https://github.com/Wynntils/Artemis/issues/1758


magicus (approved):
The deprecation warning is gone, and it seems sane.

Thank you @xCykrix for the assistance!

PR #1931, opened 2023-08-03
fix: Config page does not refresh when feature is disabled

ryanzhoudev:
@magicus 
Based on
>Also, this looks fragile, assuming that you will always toggle, and faking a click. It is better to propagate the request to make sure the tile refreshes its state.

It seems like you want the config book to reobtain the boolean value of that config, and essentially reinit that page? This goes back to what I mentioned in #artemis-dev where configs aren't saved until apply is clicked. Here, if you toggle a config and it raises the link error, the feature will be disabled without the config being saved. Actually, when anything calls .disableFeature, it will not save this to the user's config. So I tried to fix that here, but that might not be what was originally intended.

### DIFF
--- a/common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
+++ b/common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
@@ -468,6 +468,7 @@ public final class FunctionManager extends Manager {
         registerFunction(new LootrunFunctions.LastDryStreakFunction());
         registerFunction(new LootrunFunctions.LastMythicFunction());
         registerFunction(new LootrunFunctions.LootrunBeaconCountFunction());
+        registerFunction(new LootrunFunctions.LootrunLastSelectedBeaconColorFunction());
         registerFunction(new LootrunFunctions.LootrunStateFunction());
         registerFunction(new LootrunFunctions.LootrunTaskNameFunction());
         registerFunction(new LootrunFunctions.LootrunTaskTypeFunction());
--- a/common/src/main/java/com/wynntils/functions/LootrunFunctions.java
+++ b/common/src/main/java/com/wynntils/functions/LootrunFunctions.java
@@ -149,4 +149,14 @@ public class LootrunFunctions {
                     List.of(new FunctionArguments.Argument<>("color", String.class, null)));
         }
     }
+
+    public static class LootrunLastSelectedBeaconColorFunction extends Function<String> {
+        @Override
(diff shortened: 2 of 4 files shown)


ryanzhoudev:
I am not sure how .crashFeature handles things like this. But without the overwriteConfig boolean, we get stack overflows. Maybe it is easier just to crash the rich presence feature and forget about this... But then this issue will still exist.

magicus:
It might be that crashing the rich presence feature is the best, yes. We have no other case really for how features suddenly turn themselves off. But I actually don't know how the setting screens will react if a feature crashes while it is open and showing that feature...

Overall I think the feature enable/disable functionality is way too convoluted. I plan to remake this as part of the upcoming config/storage rewrite.

common/src/main/java/com/wynntils/core/consumers/features/FeatureManager.java
```diff
@@ -449,29 +450,37 @@ public void disableFeature(Feature feature) {
 
         feature.onDisable();
 
+        if (overwriteConfig) {
+            feature.setUserEnabled(false);
```
magicus: this will end up calling disableFeature() recursively. That seems like a bad thing. Maybe @kristofbolyai has any  suggestions? Or maybe we should just drop this, and hope this resolves itself when I rewrite this.

ryanzhoudev: I don't think it calls infinitely, just twice at most. What is your plan on the rewrite? Before 1.20.x or after?

magicus: The rewrite has already started; I'm pushing it incrementally. But I can't say when I will be done. Otoh, we don't have a plan for 1.20 either, right? (Not that I know about anyway...) So the answer is: possibly, maybe likely, or perhaps not :).

kristofbolyai:
> Actually, when anything calls .disableFeature, it will not save this to the user's config. So I tried to fix that here, but that might not be what was originally intended.

I don't think disabling a feature is supposed to change a config. For example, on crash, we disable the feature, but don't want to change it's enabled config. So I don't think this is the correct solution.

ryanzhoudev:
We can just crash the rpc feature then? I am okay with that

kristofbolyai:
> We can just crash the rpc feature then? I am okay with that

`com.wynntils.core.consumers.features.FeatureManager#crashFeature`

PR #1932, opened 2023-08-03
feat: Ticks function

ryanzhoudev:
Returns ticks since world start


common/src/main/java/com/wynntils/services/counter/CounterService.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public Integer getCounterValue(String id) {
+        return counterMap.get(id).getCurrent();
+    }
+
+    public boolean counterExists(String id) {
+        return counterMap.containsKey(id);
+    }
+
+    public void createCounter(Integer max, Integer maxTicks, String id) {
```
magicus: The names are confusing. `maxTicks` means `ticksPerIncrement`. right? And `max` means `wrapAfter` (or more technically, `modulus`)?

ryanzhoudev: Yeah. I will refactor to something like https://discord.com/channels/394189072635133952/946471329857675307/1136544889551732747 when I have time. Maybe also get rid of id and replace with hash code or something like that

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
```diff
@@ -236,10 +236,10 @@ public StyledText[] doFormatLines(String templateString) {
             resultBuilder.append(c);
         }
 
-        // Parse color codes before calculating the templates
-        String escapedTemplate = parseColorCodes(resultBuilder.toString());
+        // Calculate templates first to allow for color codes in templates
```
magicus: @kristofbolyai Are there any unintended side-effects of changing this order?

common/src/main/java/com/wynntils/services/iterator/IteratorService.java
```diff
@@ -0,0 +1,116 @@
...
+                } else {
+                    iterator.setCurrentIndex(iterator.getCurrentIndex() + 1);
+                }
+                iterator.setCurrentTicks(0);
+            } else {
+                iterator.setCurrentTicks(iterator.getCurrentTicks() + 1);
+            }
+        });
+        toRemove.forEach(iteratorMap::remove);
+        toRemove.clear();
```
magicus: toRemove is created in the method start and do not need to be cleared.

common/src/main/java/com/wynntils/services/iterator/IteratorService.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public String getIteratorValue(String characters, Integer ticksPerIncrement) {
+        int hashCode = getHashCode(characters, ticksPerIncrement);
+        if (!iteratorMap.containsKey(hashCode)) {
+            createCounter(characters, ticksPerIncrement);
+        }
+        return iteratorMap.get(hashCode).getCurrentString();
+    }
+
+    private void createCounter(String characters, Integer ticksPerIncrement) {
```
magicus: This has still the old name

common/src/main/java/com/wynntils/functions/generic/StringFunctions.java
```diff
@@ -282,4 +283,20 @@ public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
                     new FunctionArguments.Argument<>("replacement", String.class, null)));
         }
     }
+
+    public static class IteratorFunction extends GenericFunction<String> {
```
magicus: This is a very generic name for a very specific function. Maybe you should bite the bullet about it's specialized use, call it "ColorIterator", and let it return a full color code including the `§`?

Otherwise I think it should be at least renamed something more specific, like CharIterator or CharacterIterator or StringIterator.

Actually, even "iterator" is not ideal. It sounds like a programming construct.

Maybe `ColorProgression`?

common/src/main/java/com/wynntils/services/iterator/IteratorService.java
```diff
@@ -0,0 +1,116 @@
...
+import com.wynntils.core.components.Service;
+import com.wynntils.mc.event.TickEvent;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class IteratorService extends Service {
```
magicus: It's kind of weird to have a complete service for this... otoh, I have no better suggestion at the moment. We'll have to be on the lookout for some suitable functionality to group it with.

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
kristofbolyai: Although I really don't like changing working systems like this, I can reason nothing against this. 

What I don't see is that how or when this is needed? We can have color codes in templates even now?

common/src/main/java/com/wynntils/services/iterator/IteratorService.java
kristofbolyai: Just like the function this name is too related to programming.

kristofbolyai: All this code in this service.. I feel like this is a ticking bomb. If the only use case of an iterator function is coloring, can we just simplify this class?

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
magicus: That is the whole purpose of this fix. 

But see my suggestion below to make this clearer, by turning this into an explicit `Color` function.

common/src/main/java/com/wynntils/services/iterator/IteratorService.java
magicus: If it were not for the onTick, I would have suggested putting all of this in the function itself. It is a very special thing that is only needed for showing rainbow colors in info boxes. I fear that Donkey might have made this too generic which only confuses things.

magicus: Actually, the function can probably use the TickManager instead. A bit more overhead, but hey, we can live with that.

ryanzhoudev: I actually wanted it to be generic. Since you can pass in strings and concat them with `[` `]`, you can get one area in your screen to change the actual text it's displaying. I don't have an exact idea or use case for this in mind, but generally the community is pretty creative and might find it useful. After all I did not expect them to use soul point timers or mempct to change colors of text.

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
kristofbolyai: image
So why is this needed?

kristofbolyai: (ignore us not reseting the color for that command..)

ryanzhoudev: image

ryanzhoudev: You could force them to include & in every iterator set with [&a] or whatever but if there is no side effect of reversing the order, why not?

kristofbolyai: well yeah, i see. I am just always afraid of changing stuff like this..

kristofbolyai:
Okay so I kind of get the concept of this PR. But what if I propose a 100 times easier solution. `ticks` function. You can just have a counter in tick manager count up (and reset at int limit). And this function could return it. Then people can do anything you wanted to do here with doing a if-modulo combo, and the code is much much less. What do you think?

kristofbolyai:
Do you see something this function could do that ticks wouldnt be able to?

ryanzhoudev:
I will implement ticks and try it when I get time.

common/src/main/java/com/wynntils/core/mod/TickSchedulerManager.java
```diff
@@ -43,4 +45,17 @@ public void onTick(TickAlwaysEvent e) {
...
+    @SubscribeEvent
+    public void onTick(TickEvent e) {
+        if (ticks == Integer.MAX_VALUE) {
+            ticks = 0;
+        } else {
+            ticks++;
+        }
+    }
+
+    public int getTicks() {
```
magicus: I think there is a vanilla function returning the number of ticks since the world start, or whatever. Maybe you can just use that?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1544,6 +1544,8 @@
   "function.wynntils.stopwatchSeconds.name": "Stopwatch Seconds",
   "function.wynntils.stopwatchZero.description": "Whether the stopwatch is currently at zero",
   "function.wynntils.stopwatchZero.name": "Stopwatch Zero",
+  "function.wynntils.ticks.description": "The number of ticks since joining a world. Resets to 0 at 32-bit int limit.",
```
magicus: This description is incorrect. The tick count start at 0 when the manager starts, basically at game startup.

To have it reset when joining a world would require you to reset it when joining, and that in turn would make the manager depends on models, which we do not want to do.

ryanzhoudev: What does the tickAlways event do compared to onTick then? I meant that the tick counter initially starts on your first world join

ryanzhoudev:
The int cast is fine because even the oldest worlds (12 hours) will not go even close to the int limit

common/src/main/java/com/wynntils/core/mod/TickSchedulerManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022.
+ * Copyright © Wynntils 2022-2023.
```
kristofbolyai:
```suggestion
 * Copyright © Wynntils 2022.
```

common/src/main/java/com/wynntils/functions/MinecraftFunctions.java
```diff
@@ -37,4 +37,11 @@ public Integer getValue(FunctionArguments arguments) {
             return MinecraftAccessor.getFps();
         }
     }
+
+    public static class TicksFunction extends Function<Integer> {
+        @Override
+        public Integer getValue(FunctionArguments arguments) {
+            return (int) McUtils.mc().level.getGameTime();
```
kristofbolyai, line 44: So this is ticks since world join?  Okay

kristofbolyai (approved):

magicus (approved):

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: Yes, the tick event is only sent once you are connected, but the value is reset to 0 only when the mod starts, not each time you join a new world. So maybe it was just the way you described the behavior that made me react, not the actual implementation.

PR #1937, opened 2023-08-03
refactor: Profession percent now returns a float

ShadowCat117:
Before & after
Double
Float

Unless there's a reason why it's stored as a float yet returns a double?

### DIFF
--- a/common/src/main/java/com/wynntils/functions/ProfessionFunctions.java
+++ b/common/src/main/java/com/wynntils/functions/ProfessionFunctions.java
@@ -36,12 +36,12 @@ public class ProfessionFunctions {
         }
     }
 
-    public static class ProfessionPercentageFunction extends Function<Double> {
+    public static class ProfessionPercentageFunction extends Function<Float> {
         @Override
-        public Double getValue(FunctionArguments arguments) {
+        public Float getValue(FunctionArguments arguments) {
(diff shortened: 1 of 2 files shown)


kristofbolyai (commented):
I think all functions
"have to be" doubles instead of floats. Just so we only use that type.

ShadowCat117:
> I think all functions "have to be" doubles instead of floats. Just so we only use that type.

Ah I see. Probably a better way to do this then, shall I close this pr?

kristofbolyai:
you can just add :2 and it will round the function. That is how it works.

ShadowCat117:
Ah no, prof_pct itself is fine, it's when you wrap it in string() this becomes a problem and I don't think you can do :2 on that unless I'm mistaken

kristofbolyai:
ROUND? Can you try :2 string? I have no idea how that behaves.

ShadowCat117:
No those aren't working either, it's just that float -> double -> string that causes problems, I'll just stick to showing % without the decimal

kristofbolyai:
We can do adjustments to the function then, but not by changing the type

ShadowCat117:
~~Ok, I'll look into that then :)~~

Scratch that I'm an idiot and put the round in the wrong place (:

PR #1938, opened 2023-08-04
feat: option to sum effect modifiers

atiedebee:
I made the default false for now, because it may confuse players when they suddenly see `+45% Enemy Resistance` thinking it may be a "new" or buffed effect.

I made retrieving the hashmap key a separate function because it became more complicated.

One "parity" thing that this option does change is that it automatically puts a + before all positive effects. This could be changed with relative ease though if needed.

Some things that probably still need changing:

- `indexAfterDigits` is a bad name.
- There's a lot more comments than is probably needed.


kristofbolyai:
How does this work with `stackEffects`?

atiedebee:
> How does this work with `stackEffects`?

It requires stackEffects to be enabled, and overrides it's behaviour.

It may be beneficial to have the options as an enum, switching between no stacking, stacking with the multiplier and summing all modifiers

kristofbolyai:
> > How does this work with `stackEffects`?
> 
> It requires stackEffects to be enabled, and overrides it's behaviour.
> 
> It may be beneficial to have the options as an enum, switching between no stacking, stacking with the multiplier and summing all modifiers

yeah, please do that otherwise it is really weird

atiedebee:
Alright, I also just found out that disabling `stackEffects` somehow makes all effects appear with a 0x behind them. Fixing that as well

magicus:
Let me just start with a game technical question: Is this indeed how multiple effects of the same kind is interpreted by the Wynncraft server? So if you have 20% Enemy Resistance and 25% Enemy Resistence, the effect would be the same as if you got 45% Enemy Resistence? (There are other possible but reasonable ways to solve this, gamewise).

atiedebee:
> Let me just start with a game technical question: Is this indeed how multiple effects of the same kind is interpreted by the Wynncraft server? So if you have 20% Enemy Resistance and 25% Enemy Resistence, the effect would be the same as if you got 45% Enemy Resistence? (There are other possible but reasonable ways to solve this, gamewise).

I'm gonna test out different stats and come back in a minute

atiedebee:
I quickly did some calculations using my own stats and so far I can confirm that the following % stats are added up to eachother:
- Walk speed (including psithurism and the lootrun walk speed bonus)
- Spell damage %
- Main attack damage %

We can't really test the curses etc. but I think assuming these effects are all additive is quite a safe bet.

atiedebee:
Found a bug in the current commit. Something is going wrong parsing. I'll have it fixed soon

magicus:
@atiedebee Thanks for confirming this!

atiedebee:
The problem was me parsing strings as if they are C strings, expecting the final character not to match `Character.isDigit()`

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -138,52 +139,93 @@ private Stream<RenderedStatusEffect> stackEffects(List<StatusEffect> effects) {
...
+                    if (this.modifierList.size() > 0) {
+                        // SUM modifiers
+                        double modifierValue = 0.0;
+                        String baseModifier = this.modifierList.get(0);
+                        for (String modifier : modifierList) {
+                            modifierValue += extractDoubleFromString(modifier);
+                        }
+
+                        // Eliminate .0 when the modifier needs trailing decimals. This is the case for powder specials
+                        // on
```
magicus: Fix the linebreak

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -207,5 +249,44 @@ public void setCount(int c) {
...
+            while ( i < len && (Character.isDigit(s[i]) || s[i] == '.')) {
+                i += 1;
+            }
+            return i;
+        }
+    }
+
+    private enum StackingBehaviour {
+        NONE,
+        MULTIPLIER,
```
magicus: Using "multiplier" next to "sum" sounds like you are multiplying effects rather than adding them, which would be just weird. 

What do you think about renaming MULTIPLIER to GROUP?

atiedebee: I agree that `GROUP` is more descriptive than `MULTIPLIER`, changing it

magicus (approved):
I think this looks okay. I'd like for @kristofbolyai to have a look as well first though.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -138,52 +139,92 @@ private Stream<RenderedStatusEffect> stackEffects(List<StatusEffect> effects) {
...
+                        String baseModifier = this.modifierList.get(0);
+                        for (String modifier : modifierList) {
+                            modifierValue += extractDoubleFromString(modifier);
+                        }
+
+                        // Eliminate .0 when the modifier needs trailing decimals. This is the case for powder specials
+                        // on armor.
+                        String numberString = (Math.round(modifierValue) == modifierValue)
+                                ? String.format("%+d", (long) modifierValue)
+                                : String.format("%+.1f", modifierValue);
```
kristofbolyai, line 200: This format assumes positive effects only?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -207,5 +248,44 @@ public void setCount(int c) {
...
+            int len = string.length();
+            int i = 0;
+            while (!Character.isDigit(s[i])) {
+                i += 1;
+            }
+            while ( i < len && (Character.isDigit(s[i]) || s[i] == '.')) {
+                i += 1;
+            }
+            return i;
+        }
```
kristofbolyai: I will most likely never accept anything with byte arrays.. We have so many tools available to us, there is no need to do this.

We might have had this conversation already but what's stopping you from adding a new capture group the the regex and parsing the number there?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Nope, the + in the formatting will add a + to the numbers when they are positive.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
kristofbolyai: Also, I am just checking the previous addition, and I think even `MODIFIER_REGEX` is unnecessary. I need to login to get a grip on how this regex works though..

atiedebee: The regex is in the `StatusEffectModel` class. So if we were to extract the number from there we could add a field to the `StatusEffect` class to hold the numerical value of the modifier.

We should also get the suffix of the modifier then, this can probably be done in the `RenderedStatusEffect` class by searching for either a '%' or a '/' (in /3s modifiers).

Just thinking out loud. What do you think of that approach?

kristofbolyai: Why not put the suffix part in StatusEffect as well?

atiedebee: That works too. I thought RenderedStatusEffect would be a better place since its only used in 1 of the 3 options

atiedebee:
Wait, it's still spamming errors, I missed that

magicus:
@atiedebee Please move it back from draft state when you believe it is working properly.

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
```diff
@@ -35,9 +36,9 @@ public final class StatusEffectModel extends Model {
      * <p>Originally taken from: <a href="https://github.com/Wynntils/Wynntils/pull/615">Legacy</a>
      */
     private static final Pattern STATUS_EFFECT_PATTERN = Pattern.compile(
-            "(?<prefix>.+?)(?<modifier>§7 ?([%\\-+\\.\\/\\d]+s?)?) *(?<name>[a-zA-Z\\s]+?) (?<timer>§[84a]\\((.+?)\\)).*");
+"(?<prefix>.+?)§7(?<modifier>\\s?([%\\-+\\.\\/\\d]+s?)?)\\s*(?<name>[a-zA-Z\\/\\s]+?)\\s(?<timer>§[84a]\\((.+?)\\))");
 
-    private static final Pattern MODIFIER_REGEX = Pattern.compile("§7 ??([%\\-+\\.\\/\\d]+s?)");
+    //private static final Pattern MODIFIER_REGEX = Pattern.compile("§7 ??([%\\-+\\.\\/\\d]+s?)");
```
kristofbolyai: I guess  this is not needed now

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
```diff
@@ -78,26 +79,16 @@ public void onTabListCustomization(PlayerInfoFooterChangedEvent event) {
...
-                name = name.substring(modifier.length()).trim(); // Get all but first part of the string
-            } else {
-                modifier = StyledText.EMPTY;
-            }
+            String color = ChatFormatting.GRAY.toString();
+            StyledText prefix = StyledText.fromString(m.group("prefix").trim());
+            StyledText name = StyledText.fromString(color + m.group("name").trim());
+            StyledText displayedTime = StyledText.fromString(m.group("timer").trim());
+            String modifierGroup = m.group("modifier");
+            StyledText modifier = modifierGroup == null ? StyledText.EMPTY : StyledText.fromString(color + modifierGroup.trim());
```
kristofbolyai:
```suggestion
            String color = ChatFormatting.GRAY.toString();
            
            StyledText prefix = StyledText.fromString(m.group("prefix").trim());
            StyledText name = StyledText.fromString(color + m.group("name").trim());
            StyledText displayedTime = StyledText.fromString(m.group("timer").trim());
            
            String modifierGroup = m.group("modifier");
            StyledText modifier = modifierGroup == null ? StyledText.EMPTY : StyledText.fromString(color + modifierGroup.trim());
```

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -28,8 +33,49 @@ public StatusEffect(StyledText name, StyledText modifier, StyledText displayedTi
                 name,
                 StyledText.fromString(" "),
                 displayedTime);
+
+        if( modifier == StyledText.EMPTY ){
```
kristofbolyai: I am not sure if spotless catches this or not, but try to run it.. Our code base does not use this style of conditions, please remove the spaces from before and after the conditions.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -28,8 +33,49 @@ public StatusEffect(StyledText name, StyledText modifier, StyledText displayedTi
...
+        int end = Math.max(modifierStr.indexOf('%'), modifierStr.indexOf('/'));
+        if( end == -1 ) end = modifierStr.length(); // no suffix present
+        
+        if( end == start || end == 0 ){
+            return Optional.empty();
+        }
+
+        double val = Double.parseDouble(modifierStr.substring(start, end));
+        return Optional.of(val);
     }
```
kristofbolyai: Private methods should be after public ones

kristofbolyai: Also it seems like this method does the same thing. Just return a pair, the value and suffix.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -138,52 +140,93 @@ private Stream<RenderedStatusEffect> stackEffects(List<StatusEffect> effects) {
...
         return effectsToRender.values().stream();
     }
 
-    private static final class RenderedStatusEffect {
+    private String getEffectsKey(StatusEffect effect) {
+        return switch (effectsStackBehaviour.get()) {
+            default -> effect.asString().getString();
+            case SUM -> effect.getPrefix().getString()
+                    + effect.getName().getString()
+                    + (effect.getModifier().getString().indexOf('%') == -1)
```
kristofbolyai: You want to add a true/false to the string? I guess it would work but this is weird.

atiedebee: It's simply a check for if the effect is % based or not. Do you know of a different method that would work?

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
atiedebee: I ran spotlessCheck and it passed, so I thought it'd be ok

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Is this something that should be changed or is it fine as is?

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -30,6 +31,7 @@
 import net.minecraft.client.renderer.MultiBufferSource;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
+
```
magicus: Spotless ought to have caught this?

atiedebee: image
It didn't.
Should it be 1 or 2 spaces?

magicus: One line. And it complains to me in my clone if I add an additional newline here.. weird.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
```diff
@@ -38,7 +40,7 @@ public class StatusEffectsOverlay extends Overlay {
     public final Config<Float> fontScale = new Config<>(1.0f);
 
     @Persisted
-    public final Config<Boolean> stackEffects = new Config<>(true);
+    public final Config<StackingBehaviour> effectsStackBehaviour = new Config<>(StackingBehaviour.GROUP);
```
magicus: Since this is in the StatusEffect overlay, I think it is better to just name this config `stackingBehavior`.

magicus: Also, you might need a upfixer for this. I'll let @kristofbolyai decide if he thinks we can skip it in this case (since the original config were recently added).

kristofbolyai: We can skip this. I don't think many people disabled it.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -76,6 +85,18 @@ public StyledText asString() {
...
+    public StyledText getModifierSuffix() {
+        return this.modifierSuffix;
+    }
+
+    public boolean hasModifierValue() {
+        return this.modifierValue.isPresent();
+    }
+
+    public double getModifierValue() {
+        return this.modifierValue.get();
```
kristofbolyai: This is so backwards.. The field should be a double and this method should return optionals. Optional fields should not exist, and you just call .get without checking if it exists here.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -84,4 +105,23 @@ public int compareTo(StatusEffect effect) {
                 .compare(this.getModifier().getString(), effect.getModifier().getString())
                 .result();
     }
+
+    private Pair<StyledText, Optional<Double>> readModifierValueAndSuffix(StyledText modifier) {
+        String modifierStr = modifier.getStringWithoutFormatting();
+        if (modifierStr.isEmpty()) {
+            return new Pair<>(StyledText.EMPTY, Optional.empty());
+        }
+        int start = Math.max(modifierStr.indexOf('+') + 1, 0);
```
kristofbolyai: Technically if there is no +, index is -1 + 1 is always 0. You can remove the max and comment this.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -84,4 +105,23 @@ public int compareTo(StatusEffect effect) {
...
+
+        int end = Math.max(modifierStr.indexOf('%'), modifierStr.indexOf('/'));
```
kristofbolyai: I would rather have a static final fields with smth like "modifierSuffixes" and just check all of them here. That way if this breaks and a new one is added it is much faster to fix it.

kristofbolyai: You know what. Scrap that. Can you add a regex for this whole logic? This is too much string modification that relies on string format, easy to crash if wynn changes something. Regexes are just more stable by nature, and cleaner.

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
kristofbolyai: Why not just append the modifier?

kristofbolyai: Why not just append the modifier?

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
atiedebee: Would changing it to use `Double.NaN` for a "null" value be alright?

kristofbolyai: What? I dont know how you want to do it, just do as i said.. optional fields shouldnt exist

common/src/main/java/com/wynntils/overlays/StatusEffectsOverlay.java
atiedebee: Because that would put `+80% Damage` and `+40% Damage` for example into different buckets, which we don't want when summing the modifiers.

With the current code I could append the modifier suffix though.

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
atiedebee: It would get rid of the optional field and not require storing a boolean for if there is no modifier for the effect

kristofbolyai: You can just use the boxes Double class to allow null

atiedebee: Do you mean I would use `Double` instead of `double`?

kristofbolyai: yes

common/src/main/java/com/wynntils/models/statuseffects/StatusEffectModel.java
```diff
@@ -34,10 +34,18 @@ public final class StatusEffectModel extends Model {
...
-    private static final Pattern STATUS_EFFECT_PATTERN = Pattern.compile(
-            "(?<prefix>.+?)(?<modifier>§7 ?([%\\-+\\.\\/\\d]+s?)?) *(?<name>[a-zA-Z\\s]+?) (?<timer>§[84a]\\((.+?)\\)).*");
+    /*
+     * current regex:
+     * <prefix>         captures any characters before the §7 colour-indicator.
+     * <modifier>       optionally captures effects' modifiers' values (+100, -20, 5)
+     * <modifierSuffix> optionally captures the suffixes of the modifiers (/5s, /3s, %)
+     * <name>           captures the effects name (Strength, and even +Lightweight)
+     * <timer>          captures the effects duration ( (xx:xx) or (**:**) ).
+     *
```
magicus, line 44: 👍 

We should really have descriptions like this on more regexps. (They are powerful, but tend to be write-only...)

common/src/main/java/com/wynntils/models/statuseffects/type/StatusEffect.java
```diff
@@ -6,28 +6,42 @@
 
 import com.google.common.collect.ComparisonChain;
 import com.wynntils.core.text.StyledText;
+import java.util.List;
 
 public class StatusEffect implements Comparable<StatusEffect> {
+    private final List<String> modifierSuffixes = List.of("%", "/3s", "/5s");
```
magicus: Have a look at `StatUnit`...

atiedebee:
I filtered out `StatUnit.RAW` because it looks like it may bring problems later on.

I'm also not sure whether or not to remove the old regex comment, since it does contain some useful links.

magicus:
I'm not sure which comment you are referring to?

Also, I tried to check the code -- where are you using `modifierSuffixes`?

atiedebee:
> I'm not sure which comment you are referring to?
> 
> Also, I tried to check the code -- where are you using `modifierSuffixes`?

Ah, I forgot about it after I integrated obtaining the modifier suffix in the regex in `StatusEffectModel`. I'll remove it

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -967,8 +967,8 @@
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.name": "Status Overlay",
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.sortEffects.description": "Should effects be sorted based on their type, name and modifier?",
   "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.sortEffects.name": "Sort Effects",
-  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.stackEffects.description": "Should similar effects be stacked?",
-  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.stackEffects.name": "Stack Effects",
+  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.stackingBehaviour.description": "How should effects be stacked?",
+  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.stackingBehaviour.name": "Stack Effects",
```
```suggestion
  "feature.wynntils.statusEffectsOverlay.overlay.statusEffects.stackingBehaviour.name": "Stacking Behaviour",
```

kristofbolyai (approved):
If you think back to your original solution, you must really appreciate this one. Look so much simpler, yet more reliable, easier to read this became. Thank you for working with us.

atiedebee:
> If you think back to your original solution, you must really appreciate this one. Look so much simpler, yet more reliable, easier to read this became. Thank you for working with us.

Yup, I do agree the regex solution is a lot nicer

magicus (approved):

magicus:
Thank you for your contribution @atiedebee!

PR #1940, opened 2023-08-05
feat: render markers on map

kristofbolyai:
image

I also realize that MarkerModel should be a service.., should I PR a change?

### DIFF
--- a/common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
+++ b/common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
@@ -10,48 +10,61 @@ import com.wynntils.models.beacons.type.BeaconColor;
 import com.wynntils.models.marker.type.MarkerInfo;
 import com.wynntils.models.marker.type.MarkerProvider;
 import com.wynntils.models.marker.type.StaticLocationSupplier;
+import com.wynntils.services.map.pois.MarkerPoi;
 import com.wynntils.utils.colors.CommonColors;
 import com.wynntils.utils.mc.type.Location;
+import com.wynntils.utils.mc.type.PoiLocation;
 import com.wynntils.utils.render.Texture;
-import com.wynntils.utils.type.Pair;
 import java.util.stream.Stream;
 
-public class ActivityMarkerProvider implements MarkerProvider {
-    private Pair<Location, MarkerInfo> spawnInfo;
-    private Pair<Location, MarkerInfo> trackedActivityInfo;
+public class ActivityMarkerProvider implements MarkerProvider<MarkerPoi> {
+    private static final String ACTIVITY_LOCATION_NAME = "Activity Location";
+
+    private ActivityMarkerInfo spawnInfo;
+    private ActivityMarkerInfo trackedActivityInfo;
 
     public void setSpawnLocation(Location spawnLocation) {
         this.spawnInfo = spawnLocation == null
                 ? null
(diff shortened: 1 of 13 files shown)


magicus:
> I also realize that MarkerModel should be a service.., should I PR a change?

Kind of borderline, since a typical source of markers are the models, like lootrun or activities... Let it be a model for now.

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
```diff
@@ -10,59 +10,86 @@
...
                                 new StaticLocationSupplier(spawnLocation),
                                 Texture.QUESTS_BUTTON,
                                 CommonColors.WHITE,
                                 CommonColors.WHITE,
-                                CommonColors.WHITE));
+                                CommonColors.WHITE),
+                        new MarkerPoi(
+                                PoiLocation.fromLocation(spawnLocation),
+                                ACTIVITY_LOCATION_NAME,
+                                Texture.QUESTS_BUTTON));
```
magicus, line 41: I think we should either keep the "pin" icon, or make an item that fits the current content type (e.g. a cave icon for caves, etc).

magicus: Which is kind of feature creep for this PR. So basically I am suggesting that we stay with the current pin icon.

kristofbolyai: But I didn't change the texture?

common/src/main/java/com/wynntils/models/marker/type/MarkerProvider.java
```diff
@@ -4,10 +4,13 @@
...
 package com.wynntils.models.marker.type;
 
+import com.wynntils.services.map.pois.Poi;
 import java.util.stream.Stream;
 
-public interface MarkerProvider {
+public interface MarkerProvider<T extends Poi> {
     Stream<MarkerInfo> getMarkerInfos();
 
+    Stream<T> getPois();
```
magicus, line 13: This partially overlaps getMarkerInfos().map(m -> m.locationSupplier.get()) ... But I think we can keep doing it this way and hope that I get to finish the mapdata rewrite soon and then we'll have the tools do to this better.

kristofbolyai: Yes, the poi system wasn't really ready for this. But we need to show quest/lootrun waypoints on the map..

magicus (approved):

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
magicus: You didn't? The activity tracker will show up as a Quest icon in current main? I missed that part when reviewing, and I have not been playing with it...

kristofbolyai: Yes. It is like that, it looks okay, although it could use a white border.

kristofbolyai: So, I assume this is fine? @magicus

magicus: Yeah, I guess :)

Issue #1934, opened 2023-08-03
ItemHandler needs to support shiny stats


magicus:
They can appear like this in the tooltip:
```
§c✖§7 Class Req: Mage/Dark Wizard§r
§c✖§7 Combat Lv. Min: 97§r
§c✖§7 Strength Min: 105§r
§f⬡ §7Mobs Killed: §f0§r
§a+25 §7Strength§r
§a+8% §7Thorns§r
```

We need to look for `§f⬡ §7.*: §f.*§r` in the item matcher; ignoring it when comparing two items to see if they are the same, but if it has changed, we need to send an event saying it has changed. 

This will need some design changes, since we have not had items which change their tooltip during use like this.

geekazodium:
#1941 I wrote a tiny bit of code that reads the shiny stats.
(only did a few tests, regex is here https://regexr.com/7i5h5 )

magicus:
This is fixed in #2000.

Issue #1676, opened 2023-07-03
[2.0.3] Shiny item support


kristofbolyai:
We are currently lacking any support whatsoever for shiny items. I've created this issue to keep track of what needs fixing / adding:
- [x] Parse it as an actual item
- [x] Fix highlight color
- [ ] Parse shiny stat
- [ ] Make it display in chat item encoding (needs encoding v2)

Example item:
![image](https://github.com/Wynntils/Artemis/assets/49001742/5d29e71d-cd42-4788-a0d2-76a7a6323c56)

kristofbolyai:
Progress made in #1677

geekazodium:
https://regexr.com/7i5h5 I think this regex should be able to capture the shiny mythic stat.

geekazodium:
progress made in #1941

ShadowCat117:
Believe this is all done across various prs

PR #1941, opened 2023-08-05
feat: reads shiny item stats to optional

geekazodium:
there's a 99% chance the commit title is not using the correct prefix, pls correct me if I'm titling this wrong.


kristofbolyai:
@magicus should review this.

This is good, but we can't use this for chat items nor to fix item handler atm.

geekazodium:
> @magicus should review this.
> 
> This is good, but we can't use this for chat items nor to fix item handler atm.

I'm going to check the issues for this, hopefully I can help, if you could pls reference the issue with info about the item handler in issue tracker so I don't do something dumb like read the wrong issue and get massively confused about what's going on.

kristofbolyai:
https://github.com/Wynntils/Artemis/issues/1934 is the issue, but it is not an easy fix. If it was easy, we wouldn't have made an issue, but fixed it.

geekazodium:
alright I'm going to clothesline into the wall I just realized a detail I missed with the itemhander issue.

geekazodium:
> #1934 is the issue, but it is not an easy fix. If it was easy, we wouldn't have made an issue, but fixed it.

part 1/2 complete, mod should now ignore when checking if items are the same.

geekazodium:
P.S. the way I obtained the line where the shiny mythic statistic is might need 1. testing, 2. an entire rewrite because of issues when writing that code (should be in comments)

geekazodium:
also I think the spotless check is failing but I can't find where in the code is causing it.

geekazodium:
Nevermind I fixed the * import issue.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -55,6 +58,10 @@ public final class WynnItemParser {
 
     public static final Pattern SET_BONUS_PATTEN = Pattern.compile("^§aSet Bonus:$");
 
+    // Test suite: https://regexr.com/7i5h5
+    public static final Pattern SHINY_STAT_PATTERN = Pattern.compile("^§f⬡ §7([a-zA-Z ]*?): §f(\\d+)$");
+    public static final Pattern SHINY_STAT_NO_CAPTURE_PATTERN = Pattern.compile("^§f⬡ §7[a-zA-Z ]*?: §f\\d+$");
```
kristofbolyai: What's the point of having the no capture pattern?

common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
```diff
@@ -1,9 +1,11 @@
 /*
- * Copyright © Wynntils 2021.
+ * Copyright © Wynntils 2021-2023.
  * This file is released under AGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.utils.mc;
 
+import static com.wynntils.models.wynnitem.parsing.WynnItemParser.SHINY_STAT_NO_CAPTURE_PATTERN;
```
kristofbolyai: We disallow static imports, please remove them from the code.

kristofbolyai (commented):
Please run `spotlessCheck` and `spotlessApply`, your code needs some formatting.

common/src/main/java/com/wynntils/utils/mc/LoreUtils.java
geekazodium: okay, got it

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
geekazodium: there are cases where the capture return values are discarded, should I still use the capture pattern anyways? I am new to using regex.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -104,13 +108,35 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
...
+            StyledText newLine = newItemLore.get(index).getNormalized();
+
+            Matcher existingLineMatcher = existingLine.getMatcher(SHINY_STAT_PATTERN);
+            String statisticName = existingLineMatcher.group(1);
+            int existingStatistic = Integer.parseInt(existingLineMatcher.group(2));
+
+            Matcher newLineMatcher = newLine.getMatcher(SHINY_STAT_PATTERN);
+            int newStatistic = Integer.parseInt(newLineMatcher.group(2));
+            WynntilsMod.postEvent(
+                    new ShinyStatisticUpdateEvent(newItem, statisticName, existingStatistic, newStatistic));
```
kristofbolyai: Can you hold off on adding the event until it is actually used?

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -104,13 +108,35 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
             return;
         }
 
+        // I don't actually know a clean way to get 2 returns from one method when one is a boolean for an if condition
+        // or somehow not repeat a lot of code and seperate it to 2 methods.
+        int[] shinyStatLineIndex = {-1};
```
kristofbolyai: You can just use pairs, although if you remove the event, this is not needed atm.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -193,10 +201,29 @@ public static WynnItemParseResult parseItemStack(ItemStack itemStack, GearInfo g
...
+            // Look for shiny stat
+            Matcher shinyStatMatcher = normalizedCoded.getMatcher(SHINY_STAT_PATTERN);
+            if (shinyStatMatcher.matches() && shinyStat.isEmpty()) {
+                String statName = shinyStatMatcher.group(1);
+                int stat = Integer.parseInt(shinyStatMatcher.group(2));
+                shinyStat = Optional.of(Pair.of(statName, stat));
+            }
         }
 
+        McUtils.sendMessageToClient(Component.literal("parsed: ").append(itemStack.getDisplayName()));
```
kristofbolyai: debug

geekazodium: *why did I forget to remove this.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
kristofbolyai: capture groups are only relevant if you actually get them. you can just ignore them existing, if you dont need them.

geekazodium: okay, thx for letting me know.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
geekazodium: ok :thumbs_up:

geekazodium:
I think that's all of the immediate issues fixed, should probably do a bit of more testing on shiny mythics to verify this currently works, it's 1 AM right now though so I must sleep.

kristofbolyai:
@P0keDev What do you think of this? Do you want to implement it in a different way?

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -55,6 +57,9 @@ public final class WynnItemParser {
 
     public static final Pattern SET_BONUS_PATTEN = Pattern.compile("^§aSet Bonus:$");
 
+    // Test suite: https://regexr.com/7i5h5
+    public static final Pattern SHINY_STAT_PATTERN = Pattern.compile("^§f⬡ §7([a-zA-Z ]*?): §f(\\d+)$");
```
magicus: What does the `?` do after the `*`? Also, I'd suggest using `+` instead of `*`; we'd want at least one character here.

common/src/main/java/com/wynntils/models/gear/GearChatEncoding.java
```diff
@@ -201,7 +201,9 @@ public GearItem fromEncodedString(String encoded) {
         }
 
         // create chat gear stack
-        GearInstance gearInstance = GearInstance.create(gearInfo, identifications, powderList, rerolls);
+        // TODO: 8/7/2023 replace Optional.empty() with actual shiny mythic stat when it exists
```
magicus, line 204: That is never going to happen. We cannot change the "chat transfer" protocol to support new functions, due to the lack of version handling in the protocol. :-(

magicus: Or rather, it basically requires us to design a new protocol. And figure out how to handle old and incompatible clients.

magicus (approved):
I think this is basically looking good. One open question is if we really should store shinyStat with the GearInstance. I was thinking that the other data there are unchanging, but it is not strictly true I realize now -- rerolls can cause changes. So I guess we'll accept that shiny stats can change as well.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
geekazodium: okay the unneeded char in the regex has been removed 👍

common/src/main/java/com/wynntils/models/gear/GearChatEncoding.java
geekazodium: I mean my first instinct is to suggest attaching the shiny stat to the the end of the existing item string with a spacebar in-between so older clients will just parse the item the same but updated clients can see the stat by checking if there's more unicode encoded item info after the space char.

(but would that be a janky solution idk.)

magicus: The real solution is that we need to create a new "protocol" that has some kind of version information, or is extendable in some other way, so we can continue adapting for future Wynn changes. But that is a quest for another day...

geekazodium: following through that idea, maybe encoding the statistic name with a single byte should be fine I doubt I'll see the day in which wynncraft has 255 or something statistics on shiny mythics, but who knows, if we suppose that wynncraft will eventually have the 128th unique shiny mythic statistic, we can reserve the first bit in that as a flag value just in case this happens and if that bit is true we can check the next byte as well or something. (can't wait for the day when wynncraft releases the 128th unique shiny mythic stat, and we don't have to worry about backwards compatibility in that very specific scenario.

(but haven't I made a janky enough of an idea maybe shouldn't make it jankier.

magicus: Please just remove this comment, and then you're good to go.

geekazodium: tldr: <itemname> + ' ' + new stuff here

magicus: Problem is, there are already other information we can't properly pass using this chat protocol. So we need to fix that as well.

geekazodium: oh, that makes sense.

geekazodium:
update: this may not be working as intended because it appears the this shiny divzer's mobs killed stats are not being read properly.

geekazodium:
*correction, it works but it doesn't update it when a mob is killed.

geekazodium:
I think I know the issue... I was planning to make it so that the shiny mythic stat update event will trigger that stat to be parsed, and just... forgot.

geekazodium:
can't merge until that issue has been fixed

geekazodium:
> I think this is basically looking good. One open question is if we really should store shinyStat with the GearInstance. I was thinking that the other data there are unchanging, but it is not strictly true I realize now -- rerolls can cause changes. So I guess we'll accept that shiny stats can change as well.

... so turns out you can't modify values on records* so...

* -> I can always just make a something similiar to an optional but non final to work around that, or should I be storing this somewhere else?

geekazodium:
*doing a bit of testing and it works with my shiny divzer I bought from trade market for 2stx

geekazodium:
> *doing a bit of testing and it works with my shiny divzer I bought from trade market for 2stx

tldr: not worth but I have it sooooo............

geekazodium:
I don't know if my implementation might be messy tho..

P0keDev (changes requested):
This approach in fixing `ItemHandler` is the wrong one, IMO. See #1957

geekazodium:
> This approach in fixing `ItemHandler` is the wrong one, IMO. See #1957

I see, I was suggesting to have the shiny stat saved to an accessible value so it can also be sent and read with chat encoding #1676, although I can see that yeah your fix with #1957 is probably much better at preventing shiny statistics from messing up the `ItemHandler` (If I'm understanding this correctly).

magicus:
The most critical problem with shiny stats is that it breaks the ItemHandler. This is addressed in #1977 (as an alternative to #1957). Once that is resolved , I can start looking at how to integrate the extra value shiny stats give into the Wynntils models. This is a start in that direction, but we need to do more. I'll come back with a more precise review of this PR when we are at that point.

magicus:
This will be tricky to get right, both in terms of adapting this PR to the changes made in the Itemhandler recently, and to getting ShinyStats to work with the system we have. I have picked out the parts from this PR that I think is a relevant contribution, and will continue massaging them into something we can merge in #2006. I will to make sure you get proper credit for your part of that work, @geekazodium, and thanks for your contribution!

kristofbolyai:
Supersed by #2006

kristofbolyai:
Thank you for contributing, your code will be partially merged by that PR.

PR #1942, opened 2023-08-05
feat: Make spell casts faster, add an option to disable safety checks

kristofbolyai:


common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,58 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
+        } while (!SPELL_PACKET_QUEUE.isEmpty()
+                && (!didRightClick || SPELL_PACKET_QUEUE.peek() != SpellDirection.RIGHT));
+
+        if (slotChanged) {
+            McUtils.sendPacket(new ServerboundSetCarriedItemPacket(currSelectedSlot));
+        }
 
         // Waiting a few ticks is useful for avoiding lag related input-overlaps
-        if (!SPELL_PACKET_QUEUE.isEmpty()) packetCountdown = 3;
+        if (!SPELL_PACKET_QUEUE.isEmpty()) packetCountdown = rightClickTickDelay.get();
```
Mcrtin:
```suggestion
        if (!SPELL_PACKET_QUEUE.isEmpty()) {
            packetCountdown = rightClickTickDelay.get();
        }
```

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,58 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
         boolean slotChanged = currSelectedSlot != lastSelectedSlot;
 
-        if (slotChanged) McUtils.sendPacket(new ServerboundSetCarriedItemPacket(lastSelectedSlot));
-        SPELL_PACKET_QUEUE.poll().run();
-        if (slotChanged) McUtils.sendPacket(new ServerboundSetCarriedItemPacket(currSelectedSlot));
+        if (slotChanged) {
+            McUtils.sendPacket(new ServerboundSetCarriedItemPacket(lastSelectedSlot));
+        }
+
+        // Right clicks need a 1 tick delay between them, left clicks don't
```
Mcrtin: i think the delay is actually 2 ticks

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,58 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
+        boolean didRightClick = false;
+        do {
+            SpellDirection spellDirection = SPELL_PACKET_QUEUE.poll();
+            spellDirection.getSendPacketRunnable().run();
+
+            if (spellDirection == SpellDirection.RIGHT) {
+                didRightClick = true;
+            }
+        } while (!SPELL_PACKET_QUEUE.isEmpty()
```
Mcrtin, line 182: because the mod said that 20 cps is the cap and there are some spells (like archer movement spell) that require only a minimum of right clicks we may need a cps counter

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,58 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
+                && (!didRightClick || SPELL_PACKET_QUEUE.peek() != SpellDirection.RIGHT));
```
Mcrtin, line 183:
```suggestion
                && !(didRightClick && SPELL_PACKET_QUEUE.peek() == SpellDirection.RIGHT));
```

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
kristofbolyai: I think that rule doesn't apply here, others do this as well

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
kristofbolyai: This suggestion doesn't improve anything, but hurts logic readability

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -890,6 +890,10 @@
   "feature.wynntils.quickCast.levelRequirementNotReached": "You need to reach %s %s before using this weapon.",
   "feature.wynntils.quickCast.name": "Quick Spell Cast",
   "feature.wynntils.quickCast.notAWeapon": "You can only quick-cast with a weapon!",
+  "feature.wynntils.quickCast.rightClickTickDelay.description": "How many ticks delay should there be before a new right-click is sent?",
```
Mcrtin: You should add here something like "clicks may not register sever-side when setting this lower than 2" because people probably don't know that there is a delay. Also, I think you should call it cooldown because you only delay it if there was a right-click before.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,60 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
         if (!Models.WorldState.onWorld()) return;
 
         // Clear spell after the 40 tick timeout period
-        if (spellCountdown > 0 && --spellCountdown <= 0) spellInProgress = SpellDirection.NO_SPELL;
+        if (spellCountdown > 0 && --spellCountdown <= 0) {
+            spellInProgress = SpellDirection.NO_SPELL;
+        }
 
         if (SPELL_PACKET_QUEUE.isEmpty()) return;
         if (--packetCountdown > 0) return;
```
Mcrtin, line 164: the packet cooldown should also tick down if the queue is empty

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,60 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
-        remainder.stream().map(SpellDirection::getSendPacketRunnable).forEach(SPELL_PACKET_QUEUE::add);
+        SPELL_PACKET_QUEUE.addAll(remainder);
     }
 
     @SubscribeEvent
     public void onTick(TickEvent e) {
         if (!Models.WorldState.onWorld()) return;
 
         // Clear spell after the 40 tick timeout period
-        if (spellCountdown > 0 && --spellCountdown <= 0) spellInProgress = SpellDirection.NO_SPELL;
```
magicus, line 148: Thanks! I really wish there were a Spotless check for this...

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
kristofbolyai: Confirmed this.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
kristofbolyai: It doesn't get set if the queue is empty.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -128,37 +136,60 @@ private void tryCastSpell(SpellUnit a, SpellUnit b, SpellUnit c) {
...
+                didRightClick = true;
+            }
+        } while (!SPELL_PACKET_QUEUE.isEmpty()
+                && (!didRightClick || SPELL_PACKET_QUEUE.peek() != SpellDirection.RIGHT));
+
+        if (slotChanged) {
+            McUtils.sendPacket(new ServerboundSetCarriedItemPacket(currSelectedSlot));
+        }
 
         // Waiting a few ticks is useful for avoiding lag related input-overlaps
```
magicus: This packet tick pause is only used between two consecutive right clicks, or after the last click of the spell. Am I reading this right? If so, the comment could perhaps be updated to reflect that.

kristofbolyai: Yes. Okay. I think this can be even faster, we only need this for right-clicks.

magicus: So if you have like a "LRL, LRL" sequence, you would not need a single pause at all?

kristofbolyai: Well, you can't do the two Rs in one tick. I'll just update the comment and keep this logic..

magicus: Can we replace all this with some kind of "click sender" machinery, that has a simple state machine that knows if the last click sent was a right click, and if so and the new click to send is a right click, it adds a pause?

magicus: Like `ClickSendHandler`?

kristofbolyai: We probably can, but this PR was just meant to be a quick fix. That new system would need work + moderation checks.

magicus: It could also listen to user-initiated clicks, and make sure we don't start a spell cast in less than X ticks after that.

Mcrtin: We could also make (better) predictions on the spell progress when we are not only relying on the sever send packet but also the user input.

kristofbolyai: Added the comment.

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: okay, updated.

magicus (approved):

PR #1944, opened 2023-08-05
feat: add named lootrun tasks, overlays for lootrun functions

kristofbolyai:
Added function for task name and type, added a default overlay:
image

Added a beacon counter default overlay:
image


kristofbolyai:
Actually, let me add 1 last thing.

common/src/main/java/com/wynntils/overlays/lootrun/LootrunBeaconCountOverlay.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.wynntils.core.consumers.overlays.OverlayPosition;
+import com.wynntils.core.consumers.overlays.TextOverlay;
+import com.wynntils.utils.render.type.HorizontalAlignment;
+import com.wynntils.utils.render.type.VerticalAlignment;
+
+public class LootrunBeaconCountOverlay extends TextOverlay {
+    private static final String TEMPLATE =
+            "{if_string(string_equals(lootrun_state; \"NOT_RUNNING\"); \"\"; \\nCONCAT(\"§eYellow: \"; STRING(lootrun_beacon_count(\"YELLOW\"));\"\\n§9Blue: \"; STRING(lootrun_beacon_count(\"BLUE\"));\"\\n§5Purple: \"; STRING(lootrun_beacon_count(\"PURPLE\"));\"\\n§7Gray: \"; STRING(lootrun_beacon_count(\"GRAY\"));\"/1\";\"\\n§6Orange: \"; STRING(lootrun_beacon_count(\"ORANGE\"));\"/2\";\"\\n§bAqua: \"; STRING(lootrun_beacon_count(\"AQUA\"));\"/10\";\"\\n§8Dark Gray: \"; STRING(lootrun_beacon_count(\"DARK_GRAY\"));\"/1\";\"\\n§aGreen: \"; STRING(lootrun_beacon_count(\"GREEN\"));\"/10\";\"\\n§cRed: \"; STRING(lootrun_beacon_count(\"RED\"));\"/10\";\"\\n§fWhite: \"; STRING(lootrun_beacon_count(\"WHITE\"));\"/1\";\"\\n§4R§ca§6i§en§ab§2o§bw§9: §f\"; STRING(lootrun_beacon_count(\"RAINBOW\"));\"/1\"))}\\n";
```
magicus, line 14: The max counds per beacon is always fixed depending on color?

It sounds to me like this should really be part of our model, so that the beaconCount should return a capped value (and we need to handle like ♾️ in our CappedValue type).

Not saying you should necessarily fix that in this PR, but it really stroke me as the wrong place to hardcode the max beacon count in the string... Having that in the model allows us to e.g. represent the number of beacons as a percentage, or easily change this in once place if Wynn ever decides to change the max number for any color.

kristofbolyai: Yeah, that could be a way. The community just added these statically for the time being.

magicus (approved):

PR #1945, opened 2023-08-05
feat: Attack while key is held down

Mcrtin:


Mcrtin:
It is mostly working; just a small bug where it doesn't register spells casting and waits for the spell reset to happen (when the server resets your spell progress)

kristofbolyai:
I am putting this PR on hold until I get explicit confirmation from mods for this.

Mcrtin:
"Am I allowed to use a macro that holds down melee attack?
Allowed
You are allowed to use a melee macro. However, you have to hold a button down to keep your macro active."
(https://forums.wynncraft.com/threads/game-forum-rules.111874/)

common/src/main/java/com/wynntils/mc/mixin/PlayerMixin.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022.
+ * Copyright © Wynntils 2022-2023.
```
magicus: please revert this spurious change.

Also, please check your own PRs after you create them, and/or make updates to them! ("self reviewing").

This is a trivial thing to notice and fix yourself, and pushing that burden on the reviewers is borderline arrogance.

common/src/main/java/com/wynntils/utils/wynn/ItemUtils.java
```diff
@@ -65,4 +69,13 @@ public static MutableComponent getNonGearDescription(ItemStack itemStack, String
         }
         return null;
     }
+
+    public static boolean canBeWielded(ItemStack itemStack) {
```
magicus, line 73: I am pretty certain we already do a similar check elsewhere, like in Quick Cast. Please reuse that code instead.

common/src/main/java/com/wynntils/models/spells/SpellModel.java
```diff
@@ -36,6 +44,8 @@ public class SpellModel extends Model {
     private final SpellSegment spellSegment = new SpellSegment();
 
     private SpellDirection[] lastSpell = SpellDirection.NO_SPELL;
+    private SpellDirection[] spellPrediction = SpellDirection.NO_SPELL;
```
magicus, line 47: What is the purpose of this spell prediction? How does it "predict"?

common/src/main/java/com/wynntils/utils/wynn/ItemUtils.java
Mcrtin: the checks there are done in place. I can make quick cast use this method instead. but then I'd have to make it 2 methods because quick cast displays different messages depending on which condition you are not fulfilling.

common/src/main/java/com/wynntils/models/spells/SpellModel.java
Mcrtin: It predicts using only the user (and quick cast) inputs for more up-to-date information. instead of using the server packets.

common/src/main/java/com/wynntils/mc/mixin/PlayerMixin.java
Mcrtin: this was spotless if i revert it spotless would just make a ci commit

Mcrtin: ... because i changed something in there applied spotless and then reverted it the spotless change stuck

magicus:
See this comment for the reason this PR is being closed:

https://github.com/Wynntils/Artemis/pull/2008#issuecomment-1677956738

PR #1946, opened 2023-08-06
feat: Add spell cooldown and fix right-click delay between spells

Mcrtin:


kristofbolyai (commented):
Okay so this fixes the bug where Rs don't register if we spam spells fast enough, since the delay wasn't added. But why should we have the spell cooldown?

Mcrtin:
the spell cooldown is only optional and it prevents you from completely draining you mana in an instant

kristofbolyai:
> the spell cooldown is only optional and it prevents you from completely draining you mana in an instant

Well what I see is that you can't hold down the button so you have to intentionally spam it, you can't do it by accident. So, I am not sure if that is needed. The fix itself is nice, I wanted to do this too.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -160,8 +163,8 @@ public void onTick(TickEvent e) {
             spellInProgress = SpellDirection.NO_SPELL;
         }
 
-        if (SPELL_PACKET_QUEUE.isEmpty()) return;
         if (--packetCountdown > 0) return;
+        if (SPELL_PACKET_QUEUE.isEmpty()) return;
```
kristofbolyai, line 167: set packetCooldown to 0 here otherwise you risk running into int limit

kristofbolyai:
Can you convince me to keep the spell cooldown? Otherwise I would ask you to revert that and repurpose this PR into what it is meant to be, a fix.

Mcrtin:
https://www.youtube.com/watch?v=CTJvkLAq8OI&t=834s around 11 minutes. And you actually can hold down the key to cast more spells.

kristofbolyai (approved):

kristofbolyai:
See, this is what we have been trying to express with magicus. Simple changes go a long way, easy to review, accept and make.

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -160,8 +163,8 @@ public void onTick(TickEvent e) {
             spellInProgress = SpellDirection.NO_SPELL;
         }
 
+        if (packetCountdown > 0 && --packetCountdown > 0) return;
```
magicus, line 166: I see this is merged now, and that it was almost this bad before, but this is really not an ideal piece of code.

Using ++ or -- should nearly always be done on a separate line. And having side effects in an if condition is nearly always bad. 

```
if (packetCountdown > 0) {
   packetCountdown--;
}
if (packetCountdown > 0) return;
```

would have been much clearer.

Mcrtin: now I know. Maybe I'll make a pr to extract side effects from all ifs

PR #1948, opened 2023-08-06
feat: Add minimap zoom and make the map zoom better

Mcrtin:
I think this pr needs a better name

### DIFF
--- a/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
+++ b/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
@@ -31,9 +31,9 @@ public class MinimapFeature extends Feature {
 
     @RegisterKeyBind
     public final KeyBind zoomIn =
-            new KeyBind("Zoom in Minimap", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95));
+            new KeyBind("Zoom in Minimap", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95f));
 
     @RegisterKeyBind
     public final KeyBind zoomOut =
-            new KeyBind("Zoom out of Minimap", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05));
+            new KeyBind("Zoom out of Minimap", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05f));
 }
--- a/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
+++ b/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
@@ -5,15 +5,18 @@
 package com.wynntils.features.map;
 
 import com.wynntils.core.consumers.features.Feature;
+import com.wynntils.core.consumers.features.properties.RegisterKeyBind;
 import com.wynntils.core.consumers.overlays.Overlay;
 import com.wynntils.core.consumers.overlays.RenderState;
 import com.wynntils.core.consumers.overlays.annotations.OverlayInfo;
+import com.wynntils.core.keybinds.KeyBind;
 import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.ConfigCategory;
 import com.wynntils.mc.event.RenderEvent;
 import com.wynntils.overlays.minimap.CoordinateOverlay;
 import com.wynntils.overlays.minimap.MinimapOverlay;
 import com.wynntils.overlays.minimap.TerritoryOverlay;
+import org.lwjgl.glfw.GLFW;
 
 @ConfigCategory(Category.MAP)
 public class MinimapFeature extends Feature {
@@ -25,4 +28,12 @@ public class MinimapFeature extends Feature {
 
     @OverlayInfo(renderAt = RenderState.PRE, renderType = RenderEvent.ElementType.GUI)
     private final Overlay territoryOverlay = new TerritoryOverlay();
+
+    @RegisterKeyBind
+    public final KeyBind zoomIn =
+            new KeyBind("Zoom in Minimap", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95));
+
+    @RegisterKeyBind
+    public final KeyBind zoomOut =
+            new KeyBind("Zoom out of Minimap", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05));
 }
--- a/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
+++ b/common/src/main/java/com/wynntils/features/map/MinimapFeature.java
@@ -31,9 +31,9 @@ public class MinimapFeature extends Feature {
 
     @RegisterKeyBind
     public final KeyBind zoomIn =
-            new KeyBind("Zoom in Minimap", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95f));
+            new KeyBind("Increase Minimap Zoom", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95f));
 
     @RegisterKeyBind
     public final KeyBind zoomOut =
-            new KeyBind("Zoom out of Minimap", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05f));
+            new KeyBind("Decrease Minimap Zoom", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05f));
 }
--- a/common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
+++ b/common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
@@ -17,6 +17,7 @@ import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.config.Config;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.features.map.MainMapFeature;
+import com.wynntils.screens.maps.AbstractMapScreen;
 import com.wynntils.services.map.MapTexture;
 import com.wynntils.services.map.pois.PlayerMiniMapPoi;
 import com.wynntils.services.map.pois.Poi;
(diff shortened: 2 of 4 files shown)


common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -93,6 +98,21 @@ public MinimapOverlay() {
...
     }
 
+    @SubscribeEvent
+    // Using TickEvent for smooth zooming
+    public void onTick(TickEvent event) {
+        long window = McUtils.window().getWindow();
+        boolean zoomOut = InputConstants.isKeyDown(window, InputConstants.KEY_EQUALS)
+                || InputConstants.isKeyDown(window, InputConstants.KEY_ADD);
+        boolean zoomIn = InputConstants.isKeyDown(window, InputConstants.KEY_MINUS)
+                || InputConstants.isKeyDown(window, GLFW.GLFW_KEY_KP_SUBTRACT);
```
kristofbolyai: This is not a screen; you can add keybinds. Do so

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -93,6 +98,21 @@ public MinimapOverlay() {
...
+        if (zoomIn && zoomOut) return;
```
kristofbolyai: I dont think this is needed, if they press both buttons we can just have take priority

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
Mcrtin: which keys should be the default ones? or should I add 4 keybinds?

kristofbolyai: the normal ones, not the numpad version.

Mcrtin: can I register them in the overlay?

kristofbolyai: in the feature only

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -93,6 +94,11 @@ public MinimapOverlay() {
                 new OverlaySize(DEFAULT_SIZE, DEFAULT_SIZE));
     }
 
+    public void scale(double scaler) {
+        scale.setValue((float) Math.clamp(0.1, 3, scale.get() * scaler));
+        System.out.println(scale.get());
```
kristofbolyai: Remove printline

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -93,6 +94,11 @@ public MinimapOverlay() {
                 new OverlaySize(DEFAULT_SIZE, DEFAULT_SIZE));
     }
 
+    public void scale(double scaler) {
```
kristofbolyai:
```suggestion
    public void scale(double modifier) {
```

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -93,6 +94,11 @@ public MinimapOverlay() {
...
+        scale.setValue((float) Math.clamp(0.1, 3, scale.get() * scaler));
```
kristofbolyai: extract 0.1 and 3 into static finals at the top of the class

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
kristofbolyai: Also technically you should call scale.touched() so this is persisted

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
Mcrtin: These are the same as in abstract map screen. Should I use them instead?

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
Mcrtin: Is it not like storage where it automatically calls that?! I think I noticed the changes not persisting so imma change that

Mcrtin: it definitely should automatically call touched or else sometimes they get saved (when you change the config) and other times they do not (saying this because most of the Config#setValue caller do not)

kristofbolyai: it will be fixed when it's refactored

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -42,6 +43,7 @@
 import java.util.stream.Stream;
 import net.minecraft.client.gui.Font;
 import net.minecraft.client.renderer.MultiBufferSource;
+import org.joml.Math;
```
magicus: No, don't do this. We have MathUtils.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -43,8 +43,8 @@ public abstract class AbstractMapScreen extends WynntilsScreen {
     private static final float BORDER_OFFSET = 6;
 
     // Zoom is the scaling of the map. The bigger the zoom, the more detailed the map becomes.
-    private static final float MIN_ZOOM = 0.1f;
-    private static final float MAX_ZOOM = 3f;
+    public static final float MIN_ZOOM = 0.1f;
+    public static final float MAX_ZOOM = 3f;
```
kristofbolyai: I would rather you just created these constants in the overlay class. Sharing this looks weird.

common/src/main/java/com/wynntils/features/map/MinimapFeature.java
```diff
@@ -25,4 +28,12 @@ public class MinimapFeature extends Feature {
 
     @OverlayInfo(renderAt = RenderState.PRE, renderType = RenderEvent.ElementType.GUI)
     private final Overlay territoryOverlay = new TerritoryOverlay();
+
+    @RegisterKeyBind
+    public final KeyBind zoomIn =
+            new KeyBind("Zoom in Minimap", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95f));
```
kristofbolyai:
```suggestion
            new KeyBind("Increase Minimap Zoom", GLFW.GLFW_KEY_EQUAL, false, () -> minimapOverlay.scale(0.95f));
```

common/src/main/java/com/wynntils/features/map/MinimapFeature.java
```diff
@@ -25,4 +28,12 @@ public class MinimapFeature extends Feature {
...
+
+    @RegisterKeyBind
+    public final KeyBind zoomOut =
+            new KeyBind("Zoom out of Minimap", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05f));
```
kristofbolyai:
```suggestion
            new KeyBind("Decrease Minimap Zoom", GLFW.GLFW_KEY_MINUS, false, () -> minimapOverlay.scale(1.05f));
```

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
magicus: I thought Spotless should have removed the unused import..?

Have you disabled running github actions on your fork?

magicus: Or maybe you unchecked the box to allow maintainers to edit the branch..?

Somehow our Spotless checks are failing to update this branch, anyway...

Mcrtin: i also ran spotless locally so it shouldn't be a github problem

magicus: Then why is this import still here..?

Mcrtin: i think i forgot to push

Mcrtin: nvm branch is up to date. the change should be online but I can't see the changes on github only the push msg

Mcrtin: now i can see the changes - i had the refresh the branch. Maybe this doesn't get updated because it is the only change I the file and spotless can't change an edited file to an unedited one.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -42,9 +42,12 @@
...
 
 public class MinimapOverlay extends Overlay {
     private static final int DEFAULT_SIZE = 130;
+    private static final float MIN_ZOOM = 0.1f;
+    private static final float MAX_ZOOM = 3;
```
kristofbolyai: Sorry, can you move this to MapRenderer, and use it here and in the map class aswell?

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
kristofbolyai: the change is there but you still didnt remove the import?

Mcrtin: it didn't remove the import because it was used. Before the class was using java.lang.Math without import and because I imported joml Math it switch to that one and didn't switch back.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -41,10 +41,6 @@
 public abstract class AbstractMapScreen extends WynntilsScreen {
     protected static final float SCREEN_SIDE_OFFSET = 10;
     private static final float BORDER_OFFSET = 6;
-
-    // Zoom is the scaling of the map. The bigger the zoom, the more detailed the map becomes.
```
kristofbolyai, line 45: Why did you not move the comment with the fields..?

Mcrtin: i thought it was not really needed

magicus (approved):

kristofbolyai (approved):

PR #1949, opened 2023-08-06
feat: Lootrun Time and Challenges functions

Rafii2198:


common/src/main/java/com/wynntils/models/lootrun/scoreboard/LootrunScoreboardPart.java
```diff
@@ -27,6 +27,8 @@ public class LootrunScoreboardPart extends ScoreboardPart {
     private static final Pattern DESTROY_PATTERN = Pattern.compile("^Destroy the objective!$");
     private static final Pattern DEFEND_PATTERN = Pattern.compile("^Defend for (\\d+)s!$");
 
+    private static final Pattern TIMER_PATTERN = Pattern.compile("^[-—] Time Left: (\\d+):(\\d+)$");
```
ShadowCat117: Ok, `"§f(.+):(.+)"` should work.

common/src/main/java/com/wynntils/models/lootrun/scoreboard/LootrunScoreboardPart.java
```diff
@@ -56,6 +58,11 @@ public void onSegmentChange(ScoreboardSegment newValue) {
         } else if (currentStateLine.matches(DEFEND_PATTERN, PartStyle.StyleType.NONE)) {
             Models.Lootrun.setState(LootrunningState.IN_TASK, LootrunTaskType.DEFEND);
         }
+
+        currentStateLine = content.get(1);
+        if (currentStateLine.matches(TIMER_PATTERN, PartStyle.StyleType.NONE)) {
```
ShadowCat117: Then down here, if you make a Matcher object like `Matcher timeMatcher = currentStateLine.getMatcher(TIMER_PATTERN);` and then use `timeMatcher.find()` and if that's true then you can use `timeMatcher.group(1)` to get the minutes and group 2 for seconds

common/src/main/java/com/wynntils/models/lootrun/scoreboard/LootrunScoreboardPart.java
```diff
@@ -56,6 +58,11 @@ public void onSegmentChange(ScoreboardSegment newValue) {
         } else if (currentStateLine.matches(DEFEND_PATTERN, PartStyle.StyleType.NONE)) {
             Models.Lootrun.setState(LootrunningState.IN_TASK, LootrunTaskType.DEFEND);
         }
+
+        currentStateLine = content.get(1);
```
ShadowCat117: I also had a few errors pop up from this whilst I was testing the regex by doing /class to pause the timer and come back, so you probably want a check that content does have an index 1

common/src/main/java/com/wynntils/models/lootrun/scoreboard/LootrunScoreboardPart.java
ShadowCat117: @Rafii2198 This should work now, sorry it took a while my regex isn't great (:

Rafii2198: I tried doing it, and it wouldn't work for some reason... but well original one works

Rafii2198:
I think I'm going to extend this PR with challenge number functions too, as they pretty much can be done together

Rafii2198:
```
{lootrun_time(false)} - {lootrun_time(true)}\n
{lootrun_challenges} - {lootrun_challenges(true)}
```

ShadowCat117:
> 
> 

Not at my PC today to do a full review but looking at this screenshot, I think it would be better for the timer to display MM:SS instead of total seconds

kristofbolyai:
> > 
> > 
> 
> Not at my PC today to do a full review but looking at this screenshot, I think it would be better for the timer to display MM:SS instead of total seconds

We only do seconds functions because people can just do div mod to get mm:ss

ShadowCat117:
> We only do seconds functions because people can just do div mod to get mm:ss

Ah ok so if we make a default overlay for this we'll do it there

kristofbolyai:
> > We only do seconds functions because people can just do div mod to get mm:ss
> 
> Ah ok so if we make a default overlay for this we'll do it there

Yes, that's right

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -126,6 +126,15 @@ public class LootrunModel extends Model {
 
     private Map<BeaconColor, Integer> selectedBeacons = new TreeMap<>();
 
+    private int timerOverall = 0;
+
+    @Persisted
+    private Storage<Map<String, Integer>> timerChallenge = new Storage<>(new TreeMap<>());
+
+    ;
```
magicus: oops

Rafii2198: I think the IntelliJ's Git integration was messing with my lines and I got lost and this came out by an accident, heh

Rafii2198:
Alright, I did a full lootrun with it, and it doesn't always work, mainly because that sometimes wynncraft adds the time slightly after the state is changed meaning I need to make it different

Rafii2198:
~~I made another attempt at doing it but for some reason spotlessJavaCheck is failing with format violations for some reason at the part I didn't even change making it impossible to build, and I don't get why.~~ Quick restart of IntelliJ fixed it for some reason...
Other than that, while running the game from IntelliJ, it now always works

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -126,6 +126,15 @@ public class LootrunModel extends Model {
 
     private Map<BeaconColor, Integer> selectedBeacons = new TreeMap<>();
 
+    private int timerOverall = 0;
+
+    @Persisted
```
magicus: Please move this to the end of the other Storages.

magicus: And the other three fields can be compacted without any newline separating them. (But leave one after the block of storages).

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
```diff
@@ -150,6 +150,32 @@ public FunctionArguments.Builder getArgumentsBuilder() {
...
+    public static class LootrunTimeFunction extends Function<Integer> {
+        @Override
+        public Integer getValue(FunctionArguments arguments) {
+            return Models.Lootrun.getTimer(arguments.getArgument("requiredTime").getBooleanValue());
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.OptionalArgumentBuilder(
+                    List.of(new FunctionArguments.Argument<>("requiredTime", Boolean.class, false)));
```
kristofbolyai:
```suggestion
                    List.of(new FunctionArguments.Argument<>("timeLeftForTask", Boolean.class, false)));
```
Perhaps this is a better name. Fix the function and i18n aswell

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
```diff
@@ -150,6 +150,32 @@ public FunctionArguments.Builder getArgumentsBuilder() {
...
+        public Integer getValue(FunctionArguments arguments) {
+            return Models.Lootrun.getChallenges(arguments.getArgument("max").getBooleanValue());
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.OptionalArgumentBuilder(
+                    List.of(new FunctionArguments.Argument<>("max", Boolean.class, false)));
+        }
+    }
```
kristofbolyai: CappedValue? So you don't need arguments.

kristofbolyai (approved):
Okay, I will merge this.

PR #1950, opened 2023-08-06
refactor: Implement IntelliJ suggestions

Mcrtin:
I mostly added <p> tags in docs, renamed overridden parameters and converted to records using IntelliJs suggestions.

### DIFF
--- a/.idea/inspectionProfiles/Artemis.xml
+++ b/.idea/inspectionProfiles/Artemis.xml
@@ -1399,7 +1399,7 @@
       <option name="useParameterizedTypeForCollectionMethods" value="true" />
       <option name="doNotWeakenToJavaLangObject" value="true" />
       <option name="onlyWeakentoInterface" value="true" />
-      <stopClasses>java.lang.String,java.util.List,java.util.Set,java.util.TreeMap,net.minecraft.network.chat.Component,java.util.Random,java.util.regex.Matcher</stopClasses>
+      <stopClasses>java.lang.String,java.util.List,java.util.Set,java.util.TreeMap,net.minecraft.network.chat.Component,java.util.Random,java.util.regex.Matcher,com.wynntils.utils.SystemUtils.ClipboardImage</stopClasses>
     </inspection_tool>
     <inspection_tool class="TypeParameterExtendsFinalClass" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="TypeParameterFindViewById" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
--- a/.idea/inspectionProfiles/Artemis.xml
+++ b/.idea/inspectionProfiles/Artemis.xml
@@ -861,6 +861,7 @@
     <inspection_tool class="JavaLangImport" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="JavaMapForEach" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
     <inspection_tool class="JavaStylePropertiesInvocation" enabled="false" level="INFORMATION" enabled_by_default="false" />
+    <inspection_tool class="JavadocBlankLines" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JavadocDeclaration" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JavadocLinkAsPlainText" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="JavadocReference" enabled="false" level="WARNING" enabled_by_default="false" />
@@ -1504,5 +1505,17 @@
     <inspection_tool class="YAMLSchemaValidation" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="YAMLUnresolvedAlias" enabled="false" level="ERROR" enabled_by_default="false" />
     <inspection_tool class="YAMLUnusedAnchor" enabled="false" level="WARNING" enabled_by_default="false" />
+    <inspection_tool class="unused" enabled="true" level="WEAK WARNING" enabled_by_default="true" editorAttributes="INFO_ATTRIBUTES" checkParameterExcludingHierarchy="false">
+      <option name="LOCAL_VARIABLE" value="true" />
+      <option name="FIELD" value="true" />
+      <option name="METHOD" value="true" />
+      <option name="CLASS" value="true" />
+      <option name="PARAMETER" value="true" />
+      <option name="REPORT_PARAMETER_FOR_PUBLIC_METHODS" value="true" />
+      <option name="ADD_MAINS_TO_ENTRIES" value="true" />
+      <option name="ADD_APPLET_TO_ENTRIES" value="true" />
+      <option name="ADD_SERVLET_TO_ENTRIES" value="true" />
+      <option name="ADD_NONJAVA_TO_ENTRIES" value="true" />
+    </inspection_tool>
   </profile>
 </component>
--- a/.idea/inspectionProfiles/Artemis.xml
+++ b/.idea/inspectionProfiles/Artemis.xml
@@ -1505,7 +1505,7 @@
     <inspection_tool class="YAMLSchemaValidation" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="YAMLUnresolvedAlias" enabled="false" level="ERROR" enabled_by_default="false" />
     <inspection_tool class="YAMLUnusedAnchor" enabled="false" level="WARNING" enabled_by_default="false" />
-    <inspection_tool class="unused" enabled="true" level="WEAK WARNING" enabled_by_default="true" editorAttributes="INFO_ATTRIBUTES" checkParameterExcludingHierarchy="false">
+    <inspection_tool class="unused" enabled="true" level="WARNING" enabled_by_default="true" editorAttributes="WARNING_ATTRIBUTES" inner_class="private" field="private" method="private" checkParameterExcludingHierarchy="false">
       <option name="LOCAL_VARIABLE" value="true" />
       <option name="FIELD" value="true" />
       <option name="METHOD" value="true" />
--- a/common/src/main/java/com/wynntils/commands/CompassCommand.java
+++ b/common/src/main/java/com/wynntils/commands/CompassCommand.java
@@ -271,7 +271,7 @@ public class CompassCommand extends Command {
         Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(location.asLocation());
 
         MutableComponent response = Component.literal(
-                        "Compass set to middle of " + territoryProfile.getFriendlyName() + " at ")
+                        "Compass set to middle of " + territoryProfile.friendlyName() + " at ")
                 .withStyle(ChatFormatting.AQUA);
         response.append(Component.literal(location.toString()).withStyle(ChatFormatting.WHITE));
         context.getSource().sendSuccess(response, false);
--- a/common/src/main/java/com/wynntils/commands/CompassCommand.java
+++ b/common/src/main/java/com/wynntils/commands/CompassCommand.java
@@ -271,7 +271,7 @@ public class CompassCommand extends Command {
         Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(location.asLocation());
 
         MutableComponent response = Component.literal(
-                        "Compass set to middle of " + territoryProfile.friendlyName() + " at ")
+                        "Compass set to middle of " + territoryProfile.getFriendlyName() + " at ")
                 .withStyle(ChatFormatting.AQUA);
         response.append(Component.literal(location.toString()).withStyle(ChatFormatting.WHITE));
         context.getSource().sendSuccess(response, false);
--- a/common/src/main/java/com/wynntils/core/components/Handler.java
+++ b/common/src/main/java/com/wynntils/core/components/Handler.java
@@ -9,7 +9,7 @@ package com.wynntils.core.components;
  * aspect of Minecraft functionality, and with some Wynncraft knowledge, they distribute
  * the MC content to the actual Wynncraft models. A handler is needed when there is not a
  * clear 1-to-1 relationship between Minecraft components and models.
- *
+ * <p>
  * Handlers are created as singletons in the {@link Handlers} holding class.
  */
 public abstract class Handler extends CoreComponent {
--- a/common/src/main/java/com/wynntils/core/components/Handler.java
+++ b/common/src/main/java/com/wynntils/core/components/Handler.java
@@ -9,7 +9,7 @@ package com.wynntils.core.components;
  * aspect of Minecraft functionality, and with some Wynncraft knowledge, they distribute
  * the MC content to the actual Wynncraft models. A handler is needed when there is not a
  * clear 1-to-1 relationship between Minecraft components and models.
- * <p>
+ *
  * Handlers are created as singletons in the {@link Handlers} holding class.
  */
 public abstract class Handler extends CoreComponent {
--- a/common/src/main/java/com/wynntils/core/net/NetResult.java
+++ b/common/src/main/java/com/wynntils/core/net/NetResult.java
@@ -9,6 +9,7 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.net.event.NetResultProcessedEvent;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
@@ -100,9 +101,10 @@ public abstract class NetResult {
                 });
     }
 
+    @SuppressWarnings("TryFinallyCanBeTryWithResources")
     private Consumer<InputStream> wrappingHandler(Consumer<InputStream> handler, Consumer<Throwable> onError) {
         return (inputStream) -> {
-            try (inputStream) {
+            try {
                 // FIXME: This is needed for patching class loading issue with Forge EventBus:
                 //        https://github.com/MinecraftForge/EventBus/issues/44
                 Thread.currentThread().setContextClassLoader(WynntilsMod.class.getClassLoader());
@@ -117,6 +119,12 @@ public abstract class NetResult {
                 WynntilsMod.warn("Failure in net manager [wrappingHandler], processing " + desc, t);
                 onError.accept(t);
                 onHandlingFailed();
+            } finally {
+                try {
+                    // We must always close the input stream
+                    inputStream.close();
+                } catch (IOException ignored) {
+                }
             }
         };
     }
--- a/common/src/main/java/com/wynntils/core/net/NetResult.java
+++ b/common/src/main/java/com/wynntils/core/net/NetResult.java
@@ -9,7 +9,6 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.net.event.NetResultProcessedEvent;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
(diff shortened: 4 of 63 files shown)


common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
```diff
@@ -88,10 +72,5 @@ public boolean equals(Object other) {
             CommandAlias that = (CommandAlias) other;
             return originalCommand.equals(that.originalCommand) && aliases.equals(that.aliases);
         }
```
kristofbolyai: Why did you keep the equals?

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
```diff
@@ -84,9 +84,9 @@ public BlockPos getPos() {
         }
 
         @Override
-        public void setCanceled(boolean canceled) {
-            super.setCanceled(canceled);
-            if (canceled) {
+        public void setCanceled(boolean cancel) {
+            super.setCanceled(cancel);
+            if (cancel) {
```
kristofbolyai: I doubt this is an intellij suggestion?

common/src/main/java/com/wynntils/models/marker/type/LocationSupplier.java
```diff
@@ -8,5 +8,5 @@
 
 @FunctionalInterface
 public interface LocationSupplier {
-    Location getLocation();
+    Location location();
```
kristofbolyai: Why the rename? `get` signals that this is not a constant record, but can be dynamic

kristofbolyai: Just revert `StaticLocationSupplier` so it will work. Don't overdo the intellij suggestions..

common/src/main/java/com/wynntils/models/particle/ParticleModel.java
```diff
@@ -68,9 +68,7 @@ public void onParticleAdded(ParticleAddedEvent event) {
                         case UNVERIFIED -> {
                             // Do nothing, we will wait for more particles
                         }
-                        case INVALID -> {
-                            invalidParticles.add(unverifiedParticle);
-                        }
+                        case INVALID -> invalidParticles.add(unverifiedParticle);
```
kristofbolyai: No. Either all have braces or none. All have to have it this case.

common/src/main/java/com/wynntils/models/worlds/bossbars/InfoBar.java
```diff
@@ -33,17 +33,17 @@ public InfoBar() {
     }
 
     @Override
-    public void onUpdateName(Matcher matcher) {
-        if (matcher.pattern().equals(BOMB_INFO_PATTERN)) {
-            BombType bombType = BombType.fromString(matcher.group("bomb"));
+    public void onUpdateName(Matcher match) {
```
kristofbolyai: If this is a cleanup PR, why rename random stuff? we called Matchers matcher like a 1000 places in the codebase

common/src/main/java/com/wynntils/overlays/custombars/BubbleTexturedCustomBarOverlay.java
```diff
@@ -33,15 +33,15 @@ protected float getTextureHeight() {
 
     @Override
     protected void renderBar(
-            PoseStack poseStack, MultiBufferSource bufferSource, float renderY, float barHeight, float progress) {
+            PoseStack poseStack, MultiBufferSource bufferSource, float renderY, float renderHeight, float progress) {
```
kristofbolyai: Again, random rename, why..

common/src/main/java/com/wynntils/screens/activities/WynntilsQuestBookScreen.java
```diff
@@ -340,10 +340,10 @@ private void renderDescription(PoseStack poseStack) {
     }
 
     @Override
-    protected void reloadElementsList(String searchText) {
+    protected void reloadElementsList(String searchTerm) {
```
kristofbolyai: This is getting changed in some other PR, no need to introduce merge conflicts

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
Mcrtin: The parameter is overridden. The name is from the parent

kristofbolyai (commented):
I mostly reviewed this. Just revert the code renames... if it is not smth like `UPPER_CAMEL` -> `lowerCamel` because the field isnt static, revert it.

common/src/main/java/com/wynntils/models/marker/type/LocationSupplier.java
Mcrtin: I didn't know it changes the names of the interfaces

common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
Mcrtin: Because they were custom made and not generated by intellij

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
kristofbolyai: Okay, fine, keep this one.

common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
kristofbolyai: I don't think so? They seem to just compare the every field.

Mcrtin:
This was just the first half. I didn't want to make too many changes.

Mcrtin:
Should I also rename the equals parameters? (Some are "o" and some "other")

common/src/main/java/com/wynntils/models/worlds/bossbars/InfoBar.java
Mcrtin: Intellij got the name from the parent class (I think bar)

Mcrtin:
Ill revert the overrie renames tmw

common/src/main/java/com/wynntils/models/worlds/bossbars/InfoBar.java
Mcrtin: i just changed one of them to match. the others where there before
image

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
magicus: Actually, I disagree. That is a particularly stupid suggestion from IntelliJ. We should turn it off in the Artemis analysis profile instead. 

The overriding method has a lot more special knowledge, that might warrant a different name.

And methods overridden from Vanilla will often have bad argument names due to the deobfuscation.

common/src/main/java/com/wynntils/models/worlds/bossbars/InfoBar.java
magicus: And, once again, don't do this. Please revert the rename changes.

common/src/main/java/com/wynntils/commands/CompassCommand.java
```diff
@@ -271,7 +271,7 @@ public int territory(CommandContext<CommandSourceStack> context) {
         Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(location.asLocation());
 
         MutableComponent response = Component.literal(
-                        "Compass set to middle of " + territoryProfile.getFriendlyName() + " at ")
```
magicus: Why?! Revert this. Getters should be getters.

Have you really used the Artemis or Artemis mini analysis profiles?

common/src/main/java/com/wynntils/core/components/Handler.java
```diff
@@ -9,7 +9,7 @@
  * aspect of Minecraft functionality, and with some Wynncraft knowledge, they distribute
  * the MC content to the actual Wynncraft models. A handler is needed when there is not a
  * clear 1-to-1 relationship between Minecraft components and models.
- *
+ * <p>
```
magicus: Revert this. We are not generating html, but reading javadoc in the source. This just hurts readability.

magicus: ... and all the other changes of the same kind.

common/src/main/java/com/wynntils/core/net/NetResult.java
```diff
@@ -103,7 +102,7 @@ private void doHandle(Consumer<InputStream> onCompletion, Consumer<Throwable> on
 
     private Consumer<InputStream> wrappingHandler(Consumer<InputStream> handler, Consumer<Throwable> onError) {
         return (inputStream) -> {
-            try {
+            try (inputStream) {
                 // FIXME: This is needed for patching class loading issue with Forge EventBus:
```
magicus: Nonono! This is a breaking change. This was deliberately not closed in the method. I thought this was commented, but it might have been where it was called instead.

common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
```diff
@@ -37,9 +37,9 @@ public void onCommandSent(CommandSentEvent e) {
         String message = e.getCommand();
 
         for (CommandAlias commandAlias : aliases.get()) {
-            if (commandAlias.getAliases().stream().anyMatch(alias -> Objects.equals(alias, message))) {
```
magicus, line 40: Why are you renaming getters?

magicus: Ah, I see, you turned it into a record. This is fine.

common/src/main/java/com/wynntils/features/tooltips/ItemStatInfoFeature.java
```diff
@@ -132,7 +133,7 @@ public void onTooltipPre(ItemTooltipRenderEvent.Pre event) {
         }
     }
 
-    private void updateItemName(GearInfo gearInfo, GearInstance gearInstance, LinkedList<Component> tooltips) {
+    private void updateItemName(GearInfo gearInfo, GearInstance gearInstance, Deque<Component> tooltips) {
```
magicus: I know IntelliJ keeps suggesting this, but I disagree. In my view, Deque is not a commonly used interface in Java. In short, it was introduced too late. So when we need to have a double-ended list, I prefer to specificially request a LinkedList. It's not that we ever have any other implementation of Deque anyway.

magicus:
I feel you are missing the fundamental reason to do this: it is to improve code readability and code quality. Not to satisfy a lot of IntelliJ markers.

I have trimmed away most of the stupid stuff that IntelliJ suggests in our Artemis profiles, but most of that which are left still has a lot of bad suggestions, and they all need to be considered if they should be applied or ignored.

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
Mcrtin: so should i change it back?

Mcrtin:
I used the artemis (not artemis mini) profile for this pr

common/src/main/java/com/wynntils/commands/CompassCommand.java
Mcrtin: this changed because TerritoryProfile is now a record. revert?

Mcrtin:
I edited the Artemis profile a bit. @magicus can you tell me what you think about that?

common/src/main/java/com/wynntils/models/territories/profile/TerritoryProfile.java
```diff
@@ -21,25 +21,22 @@
 import net.minecraft.ChatFormatting;
 import net.minecraft.core.Position;
 
-public class TerritoryProfile {
+public record TerritoryProfile(
```
magicus: I don't think this class gains by being converted to a record. Being a record is more than just fulfilling the formal requirements. It also conveys "i am just a stupid data carrier". Basically, the ideal record has no additional methods. Having one, or maybe two (for large records) is okay. But this class has so much special knowledge, that we send the wrong signals by turning it into a record.

We could perhaps do that, but then we'd need to move the logic away into the TerritoryModel or something like that. It might be a good refactoring, or it might not. But just this step is bad.

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
```diff
@@ -29,7 +29,7 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class ChatTabService extends Service {
-    private final ChatComponent FALLBACK_CHAT = new ChatComponent(McUtils.mc());
```
magicus, line 32: I think the proper fix would have been to make it static, no?

common/src/main/java/com/wynntils/utils/SystemUtils.java
```diff
@@ -47,13 +47,7 @@ public static int getMemMax() {
...
     }
 
-    private static final class ClipboardImage implements Transferable {
-        private final Image image;
-
-        private ClipboardImage(Image image) {
-            this.image = image;
-        }
-
+    private record ClipboardImage(Image image) implements Transferable {
```
magicus: Same here. This is not a record. You can't just apply the IntelliJ suggestions without thinking.

.idea/inspectionProfiles/Artemis.xml
```diff
@@ -861,6 +861,7 @@
     <inspection_tool class="JavaLangImport" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="JavaMapForEach" enabled="false" level="WEAK WARNING" enabled_by_default="false" />
     <inspection_tool class="JavaStylePropertiesInvocation" enabled="false" level="INFORMATION" enabled_by_default="false" />
+    <inspection_tool class="JavadocBlankLines" enabled="false" level="WARNING" enabled_by_default="false" />
```
magicus, line 864: this seems good

.idea/inspectionProfiles/Artemis.xml
```diff
@@ -1504,5 +1505,17 @@
     <inspection_tool class="YAMLSchemaValidation" enabled="false" level="WARNING" enabled_by_default="false" />
     <inspection_tool class="YAMLUnresolvedAlias" enabled="false" level="ERROR" enabled_by_default="false" />
     <inspection_tool class="YAMLUnusedAnchor" enabled="false" level="WARNING" enabled_by_default="false" />
+    <inspection_tool class="unused" enabled="true" level="WEAK WARNING" enabled_by_default="true" editorAttributes="INFO_ATTRIBUTES" checkParameterExcludingHierarchy="false">
```
magicus: I don't understand what this does..?

Mcrtin: Warning. : code fragments that might produce bugs or require enhancement. Weak Warning. : code fragments that can be improved or optimized (redundant code, duplicated code fragments, and so on).
https://www.jetbrains.com/help/idea/configuring-inspection-severities.html

One effect of this is that intellij complain when pushing

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
Mcrtin: idk how you feel about having statics in a service class

.idea/inspectionProfiles/Artemis.xml
Mcrtin: the unused inspection was level="WARNING" before (default)

Mcrtin: changed it back to warning but only if the member can't be accessed.

common/src/main/java/com/wynntils/core/net/NetResult.java
```diff
@@ -101,6 +101,7 @@ private void doHandle(Consumer<InputStream> onCompletion, Consumer<Throwable> on
                 });
     }
 
+    @SuppressWarnings("TryFinallyCanBeTryWithResources")
```
kristofbolyai, line 104: do you think we need these @magicus

common/src/main/java/com/wynntils/mc/event/PlayerInteractEvent.java
kristofbolyai: yes

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
kristofbolyai: if you checked and service class, you would find many statics. Same in models

common/src/main/java/com/wynntils/services/cosmetics/CosmeticsService.java
```diff
@@ -41,7 +41,7 @@ public class CosmeticsService extends Service {
                     RenderLayerParent<AbstractClientPlayer, PlayerModel<AbstractClientPlayer>>,
                     EntityModelSet,
                     WynntilsLayer>
-            ELYTRA_LAYER = (playerRenderer, playerModel) -> new WynntilsElytraLayer(playerRenderer, playerModel);
+            ELYTRA_LAYER = WynntilsElytraLayer::new;
```
kristofbolyai: This is like this so it's in symmetry with the cape layer. revert.

Mcrtin: i was asking myself that the whole time ^^ thx for the clarifcation

common/src/main/java/com/wynntils/core/net/NetResult.java
Mcrtin: image

kristofbolyai: Context matters. Afaik we only added the unchecked ones. By default, only we (mostly magicus) do these cleanups (with me reviewing) since this requires a much broader knowledge about the code base, than anyone else has right now. Even if we create and review these PRs, there is a chance we might break something. That's why we are strict reviewing this (and any other PRs).

Mcrtin: you also added deprecation

kristofbolyai: Yes. Deprecation has a meaning of "dont use this" (also it might be from a supermethod). Unchecked has a meaning of "these raw generics / casts look sketchy but we know what we are doing here". Adding 3 random suppress warning is just litter. Just like we disallowed Nonnull/Nullable annotations since they were not used consistently, this shouldn't be either.

magicus: I agree; these fills no purpose. This is for a check in IntelliJ that we explicitly do not say will pass.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -160,13 +160,7 @@ public GearMajorId deserialize(JsonElement jsonElement, Type jsonType, JsonDeser
...
     }
 
-    private static final class GearInfoDeserializer implements JsonDeserializer<GearInfo> {
-        private final List<GearMajorId> allMajorIds;
-
-        private GearInfoDeserializer(List<GearMajorId> allMajorIds) {
-            this.allMajorIds = allMajorIds;
-        }
-
+    private record GearInfoDeserializer(List<GearMajorId> allMajorIds) implements JsonDeserializer<GearInfo> {
```
magicus, line 163: This is not suitable as a record.

common/src/main/java/com/wynntils/models/ingredients/IngredientInfoRegistry.java
```diff
@@ -118,13 +118,8 @@ private void loadIngredients() {
...
     }
 
-    private static final class IngredientInfoDeserializer implements JsonDeserializer<IngredientInfo> {
-        private final Map<String, String> ingredientSkins;
-
-        private IngredientInfoDeserializer(Map<String, String> ingredientSkins) {
-            this.ingredientSkins = ingredientSkins;
-        }
-
+    private record IngredientInfoDeserializer(Map<String, String> ingredientSkins)
```
magicus, line 121: This is not suitable as a record.

magicus:
Actually, I'm going to close this. The whole PR is just a mess.

If you don't know how to clean up the code, make a simple PR which does something uncontroversial, e.g. adds missing "final" modifiers, so it can be properly reviewed. And then you can repeat with a second PR, for the next issue etc.

And, furthermore, please hold any such cleanups until @kristofbolyai is done with the 1.20 refactoring.

PR #1952, opened 2023-08-07
fix: Crash rich presence instead of disabling

ryanzhoudev:


magicus (approved):

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -129,7 +129,7 @@ protected void onConfigUpdate(Config<?> config) {
             if (!Services.Discord.isReady()) {
                 // Even though this is in the onConfigUpdate method, it is how the library is first loaded on launch
                 if (!Services.Discord.load()) {
```
magicus, line 131: Maybe add some comment that this will happen due to a severe config error for the SDK, and that crashing is needed to get the disabling handled properly.

magicus (approved):

PR #1953, opened 2023-08-07
refactor: avoid side effects in if statement

magicus:
Followup from #1946.

### DIFF
--- a/common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
+++ b/common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
@@ -161,7 +161,7 @@ public class QuickCastFeature extends Feature {
         // Clear spell after the 40 tick timeout period
         if (spellCountdown > 0) {
             spellCountdown--;
-            if (spellCountdown == 0) {
+            if (spellCountdown <= 0) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
```diff
@@ -159,11 +159,18 @@ public void onTick(TickEvent e) {
         if (!Models.WorldState.onWorld()) return;
 
         // Clear spell after the 40 tick timeout period
-        if (spellCountdown > 0 && --spellCountdown <= 0) {
-            spellInProgress = SpellDirection.NO_SPELL;
+        if (spellCountdown > 0) {
+            spellCountdown--;
+            if (spellCountdown == 0) {
```
ryanzhoudev: We can probably still keep the <= just in case

ryanzhoudev (approved):

common/src/main/java/com/wynntils/features/combat/QuickCastFeature.java
magicus: There is no other way to decrease spellCountdown? But sure, if you want to be extra defensive..

PR #1954, opened 2023-08-07
feat: Trade market Silverbull tax detection

ryanzhoudev:
Works on the item I tested in game. Freedom selling for total of 2stx 40le sells for 2stx 35le with silverbull on and 2stx 32le with silverbull off.
https://feedback.wynntils.com/posts/217/options-for-lowest-sell-offer-button-in-trademarket


common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceMatchFeature.java
```diff
@@ -44,6 +46,10 @@ public class TradeMarketPriceMatchFeature extends Feature {
     private static final int PRICE_SET_ITEM_SLOT = 12;
     private static final int PRICE_INFO_ITEM_SLOT = 17;
 
+    @Persisted
+    public final Config<Boolean> silverbullTax = new Config<>(false);
```
magicus: This syncing is not necessary, and problematic. You should have only one of these; remember, the config should really have been in the model, and will be moved there as soon as we can.  When that happen, we will have a hard time handling a "split" config like this.

magicus:
I'm a bit lost in the gameplay behind this. Do you get a lower sales tax if you have Silverbull shares or something like that?

Is there no way we can detect automatically if this is the case?

ryanzhoudev:
> I'm a bit lost in the gameplay behind this. Do you get a lower sales tax if you have Silverbull shares or something like that?
> 
> Is there no way we can detect automatically if this is the case?

Yeah, you get 2% tax discount with active Silverbull subscription. I don't think shares affect it. I am not sure about auto detection. I don't have the subscription, so I cannot check or test.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceMatchFeature.java
ryanzhoudev: What problems does this raise? We can think of this split config as just one config with two places that allow you access it. Since they will always be the same no matter what (synced on launch, on change, etc).

If we only have one, we will have to tell people to check one config or the other when they ask about how to set silverbull tax for -t and for the price match button.

ryanzhoudev:
In the `/use` menu, we have this Subscription: line we can check.
image

kristofbolyai:
> In the `/use` menu, we have this Subscription: line we can check.
> 
> image
> 
> 

Sadly adding a new container query to be run.. only on first join though, all classes share this

magicus:
Can we see an expiry date for active subscriptions? If so we might only need to recheck this ever so often.

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -38,14 +39,14 @@ public final class EmeraldModel extends Model {
     private static final Pattern M_PATTERN = Pattern.compile("(\\.?\\d+\\.?\\d*)\\s*(m|million)");
     private static final Pattern E_PATTERN = Pattern.compile("(\\d+)($|\\s|\\s*e|\\s*em)(?![^\\d\\s-])");
     private static final Pattern RAW_PRICE_PATTERN = Pattern.compile("\\d+");
-    public static final double TAX_AMOUNT = 1.05;
+    public static final double TAX_AMOUNT = Models.Character.isSilverbullSubscriber() ? 1.03 : 1.05;
```
kristofbolyai: Not really the way to do this. Since this is static final, referencing this class earlier than character model runs could break this. I suggest adding the tax values as private constants, and adding a getter. That will always work.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -55,6 +58,7 @@ public final class CharacterModel extends Model {
     private boolean inCharacterSelection;
     private boolean hasCharacter;
 
+    private Boolean silverbullSubscriber = null;
```
kristofbolyai: perhaps you should move this to after level. I think that is a more ergonomic approach to these variables. Don't forget to fix the toString.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -138,6 +149,19 @@ private void scanCharacterInfoPage() {
         query.executeQuery();
     }
 
+    private void scanCratesBombsCosmeticsPage() {
+        ScriptedContainerQuery query = ScriptedContainerQuery.builder("Crates, Bombs & Cosmetics Query")
+                .onError(msg -> WynntilsMod.warn("Error querying Crates, Bombs & Cosmetics: " + msg))
+
+                // Open /use menu
+                .then(QueryStep.sendCommand("use")
+                        .expectContainerTitle("Crates, Bombs & Cosmetics")
```
kristofbolyai: extract this to a final variable. I am not sure if this class should hold it though.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -138,6 +149,19 @@ private void scanCharacterInfoPage() {
         query.executeQuery();
     }
 
+    private void scanCratesBombsCosmeticsPage() {
+        ScriptedContainerQuery query = ScriptedContainerQuery.builder("Crates, Bombs & Cosmetics Query")
```
kristofbolyai: The key for the query is actually important, has to be unique. I suggest adding ": Parsing Silverbull Status", so we fulfill that.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: Also, if you do the change I suggested in emeraldModel, I see no reason why this has to be a boxed type.

kristofbolyai:
Thank you for commiting to working on this, and using your own money..

common/src/main/java/com/wynntils/models/character/CharacterModel.java
magicus: Or better only "Getting Silverbull Subscription"

ryanzhoudev:
> Can we see an expiry date for active subscriptions? If so we might only need to recheck this ever so often.

Yes but it is not very precise. It only tells me "1 week 5 days".

common/src/main/java/com/wynntils/models/character/CharacterModel.java
ryanzhoudev: Boxed type ensures we only check it once per game launch, but maybe I can improve this with magicus' idea of checking the remaining subscription duration.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
ryanzhoudev: The command or the container title? If the container title, do you also want me to move the existing scanCharacterInfoPage() one?

kristofbolyai: Title. and yes

magicus:
Sounds like it could be a bit tricky to parse, but if we can get this to a "don't need to recheck this before timestamp T", it would be nice.

ryanzhoudev:
image

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1034,7 +1034,7 @@
   "feature.wynntils.tradeMarketBulkSell.bulkSell3Amount.description": "How many items should be sold when the bulk sell 3 button is pressed?",
   "feature.wynntils.tradeMarketBulkSell.bulkSell3Amount.name": "Bulk Sell 3 Amount",
   "feature.wynntils.tradeMarketBulkSell.description": "Adds bulk sell buttons in the trade market UI",
-  "feature.wynntils.tradeMarketBulkSell.name": "Trade Market Bulk Sell Buttons",
```
magicus: I think we (and Wynntils) rather consistently refers to it as the "Trade Market", not just "Market". Is there any particular reason you want to change this?

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -152,6 +191,37 @@ private void parseCharacterContainer(ContainerContent container) {
...
+        if (!silverbullSubscriber.get()) return;
+
+        Matcher expiry = LoreUtils.matchLoreLine(rankSubscriptionItem, 1, SILVERBULL_DURATION_PATTERN);
+        if (!expiry.matches()) {
+            WynntilsMod.warn("Could not parse Silverbull subscription expiry from item: "
+                    + LoreUtils.getLore(rankSubscriptionItem));
+            silverbullExpiresAt.store(0L);
+            return;
+        }
+        int weeks = Integer.parseInt(expiry.group("weeks"));
```
magicus, line 215: You'll have to check if this regexp still matches your silverbull subscription page when you go from weeks and days to just days, and possibly when you are at the last few days (to see if it switches to include hours).

magicus (approved):
Looks good overall, but see my comment about the i18n changes...

common/src/main/resources/assets/wynntils/lang/en_us.json
ryanzhoudev: Half the market features do not have "Trade" in front, so I wanted to sync them. I picked just "Market..." because most of the names overflow with Trade in front

magicus: I agree that it is good to be consistent. But this is orthogonal to the silverbull tax issue, and this change is not the only one required to get to a consistent language. I recommend you revert these two changes, and we have a discussion in discord whether to use "Trade Market" or just "Market", and then open a PR where we fix all instances at once.

magicus (approved):
Looks good now!

PR #1955, opened 2023-08-07
feat: Port to 1.20.1

kristofbolyai:
This PR ports us to 1.20.1, a stepping stone for 1.20.2. All related changes are separated into commits, I highly suggest reviewing per commit.

DO NOT push merge commits here, it will most likely break it.. I will PR merges from time to time.


kristofbolyai:
With this, 1.20.1 porting is complete. Some bugs might still exist, but I think I got most of them.

common/src/main/java/com/wynntils/mc/mixin/PlayerTabOverlayMixin.java
```diff
@@ -18,11 +18,17 @@
...
+                    "render(Lnet/minecraft/client/gui/GuiGraphics;ILnet/minecraft/world/scores/Scoreboard;Lnet/minecraft/world/scores/Objective;)V",
+            at = @At("HEAD"),
+            cancellable = true)
+    private void render(
+            GuiGraphics guiGraphics, int width, Scoreboard scoreboard, Objective objective, CallbackInfo ci) {
         RenderEvent.Pre renderEvent =
-                new RenderEvent.Pre(poseStack, 0, McUtils.window(), RenderEvent.ElementType.PLAYER_TAB_LIST);
+                new RenderEvent.Pre(guiGraphics, 0, McUtils.window(), RenderEvent.ElementType.PLAYER_TAB_LIST);
         MixinHelper.post(renderEvent);
+
```
Mcrtin, line 31: why is there a change on this empty line?

kristofbolyai: I added it for readability. Why is that a problem..?

Mcrtin: nvm i thought the line was already there before.

kristofbolyai:
To note this here too, I changed the target to a 1.20.2 branch. Please review this PR, and merge it. I can do merges to the 1.20.2 branch, and port it to 1.20.2 once it becomes relevant.

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
```diff
@@ -1,32 +1,46 @@
...
+import net.minecraft.client.gui.GuiGraphics;
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.Cancelable;
 import net.minecraftforge.eventbus.api.Event;
 
+/**
+ * This event is fired when an item tooltip is rendered.
+ *
+ * However, starting Minecraft 1.20, there is no longer a concerete way of rendering item tooltips.
```
magicus: Typo in "concrete"

(Don't know why I can't make a suggestion... github moment)

common/src/main/java/com/wynntils/mc/mixin/GuiGraphicsMixin.java
```diff
@@ -0,0 +1,118 @@
...
+            ordinal = 0,
+            argsOnly = true)
+    private String renderItemDecorations(
+            String text,
+            Font font,
+            ItemStack itemStack,
+            int x,
+            int y,
+            String ignored,
+            @Share("wynntilsCountOverlayColor") LocalIntRef wynntilsCountOverlayColor) {
```
magicus, line 82: I have not seen this before. Is it from the mixin extension package? Does it create a "hidden" field which can be accessed only from those two methods?

kristofbolyai: https://github.com/LlamaLad7/MixinExtras/wiki/Local#mutability

It's a neat thing. I tried using it with Inject-cancel to simplify code instead of WrapOperation, but it did not work. I have not submitted an issue yet; I am not sure if it's intended to change method arguments as well. But from the examples it feels like it.

forge/src/main/java/com/wynntils/forge/mixins/ForgeGuiMixin.java
```diff
@@ -22,24 +22,28 @@ protected ForgeGuiMixin(Minecraft arg) {
...
+                guiGraphics, partialTick, this.minecraft.getWindow(), RenderEvent.ElementType.GUI));
     }
 
-    @Inject(method = "renderFood", at = @At("HEAD"), cancellable = true, remap = false)
-    private void onRenderFoodPre(int width, int height, PoseStack poseStack, CallbackInfo ci) {
+    @Inject(
+            method = "renderFood(IILnet/minecraft/client/gui/GuiGraphics;)V",
+            at = @At("HEAD"),
+            cancellable = true,
+            remap = false)
```
magicus, line 41: I see this is not changed now, but do you know why this and the next method is remap = false?

kristofbolyai: Yes. These are forge mixins, so most of the time we are mixing into forge added methods. These are not obfuscated, there is no need to map to intermediary (or to any other mapping).

magicus (commented):
Regard the sendSuccess changes: this takes a long construct and makes it really unwieldy.

What do you think about adding e.g. `McUtils.commandResponse(Context, Component)` which does this getSource.sendSuccess and lambda dance?

Or should we save it for after the porting is done?

forge/src/main/java/com/wynntils/forge/mixins/ForgeGuiMixin.java
magicus: Ah, so some of them are mixing into vanilla methods, but only needed on forge, and other are mixing into forge methods?

Maybe we should acknowledge the later with a comment?

common/src/main/java/com/wynntils/screens/maps/PoiCreationScreen.java
```diff
@@ -374,9 +377,8 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
 
         RenderUtils.disableScissor();
 
-        renderBackground(poseStack);
-        super.doRender(poseStack, mouseX, mouseY, partialTick);
-        // RenderUtils.renderDebugGrid(poseStack, GRID_DIVISIONS, dividedWidth, dividedHeight);
```
magicus, line 379: You don't think we should keep the debug grid lines? I think they amount to some level of effort to getting them right. (In fact, we should maybe have them not as commented-out code but disabled by a global variable only.)

common/src/main/java/com/wynntils/screens/overlays/placement/OverlayManagementScreen.java
```diff
@@ -239,12 +242,7 @@ public void doRender(PoseStack poseStack, int mouseX, int mouseY, float partialT
             }
         }
 
-        super.doRender(poseStack, mouseX, mouseY, partialTick); // This renders widgets
-        // This renders button tooltips
-        if (this.deferredTooltipRendering != null) {
```
magicus, line 244: This is not needed anymore? Is deferredTooltipRendering a dead variable then?

common/src/main/java/com/wynntils/screens/maps/PoiCreationScreen.java
kristofbolyai: I just removed the comment. The method is a renderutils one, can be called easily. This used to be a local method, hence the commented code.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -90,7 +90,7 @@ private ServerData getWynncraftServer() {
 
     private static void connectToServer(ServerData serverData) {
         ConnectScreen.startConnecting(
-                McUtils.mc().screen, McUtils.mc(), ServerAddress.parseString(serverData.ip), serverData);
+                McUtils.mc().screen, McUtils.mc(), ServerAddress.parseString(serverData.ip), serverData, false);
```
magicus, line 93: Just curious: what does the false represent?

common/src/main/java/com/wynntils/screens/overlays/placement/OverlayManagementScreen.java
kristofbolyai: This is called by MC now, after render call. The same variable works just like it (in fact we have technical debt of not using it).

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai: `isQuickPlay` or smth like that. You know, you can have a config file to auto join a server on launch.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
```diff
@@ -107,17 +107,26 @@ private static class WynncraftButton extends Button {
...
 
             if (serverIcon == null || serverIcon.getServerIconLocation() == null) {
                 return;
             }
 
-            RenderSystem.setShaderTexture(0, serverIcon.getServerIconLocation());
-
             // Insets the icon by 3
-            blit(matrices, this.getX() + 3, this.getY() + 3, this.width - 6, this.height - 6, 0, 0, 64, 64, 64, 64);
+            guiGraphics.blit(
```
magicus, line 118: In the previous commit, you changed a blit to a call to RenderUtils in the overlay management screen. I assumed the method were no longer accessible, but maybe there was another reason..?

forge/src/main/java/com/wynntils/forge/mixins/ForgeGuiMixin.java
kristofbolyai: Forge mixins only mixin into forge methods since the common mixins don't work on forge. Hence, a case for a forge mixin to mixin into not-forge code is really low. And the game doesn't compile without correct settings. So I think this is fine without explicit comments.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
magicus: Ah, thanks.

kristofbolyai:
> Regard the sendSuccess changes: this takes a long construct and makes it really unwieldy.
> 
> What do you think about adding e.g. `McUtils.commandResponse(Context, Component)` which does this getSource.sendSuccess and lambda dance?
> 
> Or should we save it for after the porting is done?

As a cleanover, or after this PR is merged. 1.20.2 has a separate branch, so that is fine. However I am not sure if I like using McUtils here, it does not feel intuitive.

common/src/main/java/com/wynntils/features/ui/WynncraftButtonFeature.java
kristofbolyai: That blit method was incompatible with GuiGraphics, and I rather converted it than fix it (also it was in a screen so that would just make it normal). This case, it was perfectly fine, and this is kind of an edge-case widget. So I left it.

magicus:
>  I am not sure if I like using McUtils here, it does not feel intuitive. 

Okay, let's skip it. I understand you, I had the same doubts. If I figure out a better place to put a convenience method I can fix this later on.

magicus (approved):

magicus:
The PR seemed very straightforward. Either it was an easy port, or you are very good at presenting the work you have done. :-)

kristofbolyai:
> The PR seemed very straightforward. Either it was an easy port, or you are very good at presenting the work you have done. :-)

A bit of both :) 

Last time I did porting was 1.19.3, learned quite a bit since then. Now I had a plan on how to port it, instead of just doing it.. I also learned from you and your 1.19.4 (I think you did that?). The update itself is pretty straightforward. Too bad item rendering is now all over the place, because someone forgot to check method usages.. We will have to live with 6 mixins instead of 2 :)

common/src/main/java/com/wynntils/mc/event/ItemTooltipRenderEvent.java
magicus:
```suggestion
 * However, starting Minecraft 1.20, there is no longer a concrete way of rendering item tooltips.
```

magicus (approved):

PR #1956, opened 2023-08-07
feat: Add new 2.0.3 minor IDs

magicus:
This might break the item chat tooltips for users on older versions of Artemis and Legacy


kristofbolyai (approved):
Does the WebAPI repo need this added?

magicus:
> Does the WebAPI repo need this added?

I don't think so, but maybe I'm forgetting something. Any particular file you were thinking of that needs updating?

I can say that Athena needs updating with this to support Legacy. But I think we've agreed not to do that. Athena needs no update for Artemis.

PR #1957, opened 2023-08-08
fix: Add exception in ItemHandler for spell cast indicators

P0keDev:
Fixes a bug relating to #1934, where updates to the shiny stat in lore could break weapon annotations. This is a general solution to issues caused by items being renamed due to spell casts, and overall makes the `ItemHandler` logic more robust.

I have tested this using a shiny mythic with the "mobs killed" stat to ensure it works properly


common/src/main/java/com/wynntils/core/keybinds/KeyBind.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022.
+ * Copyright © Wynntils 2022-2023.
```
magicus: I'm not sure what happened here, but this file has not been modified in 2023. 

I have fixed all correct updates to 2022-2023 in main and merged it here. I suggest you remove these manually; I don't think spotless will do it for you once it has put it in...

kristofbolyai:
I am fine with this PR, looks good to me. The decision is in magicus' hand. I am not fully up to date on this, but as far as I know, we are having problems with Handler-Model dependency.

P0keDev:
@magicus i have fixed the weird issues that happened with spotless

magicus:
I apologize I have not yet made up my mind about this. I need to spend a bit more quality time with the code to fully understand the pros and cons of the two competing solutions. I'll try to get it done today.

magicus:
I recommend closing this in favor of #1977.

P0keDev:
superseded by #2000

PR #1960, opened 2023-08-08
fix: Properly format major IDs in item guide

magicus:


magicus:
@kristofbolyai This was more or less impossible to do before. With StyledText, it was a breeze. I love it, thank you so much for building that!

common/src/main/java/com/wynntils/utils/mc/RenderedStringUtils.java
```diff
@@ -20,19 +20,17 @@ public static StyledText[] wrapTextBySize(StyledText s, int maxPixels) {
...
         int length = 0;
 
-        // FIXME: codes should not count toward the word length
-
         for (StyledText string : stringArray) {
             StyledText[] lines = string.split("\\\\n");
             for (int i = 0; i < lines.length; i++) {
                 StyledText line = lines[i];
-                if (i > 0 || length + font.width(line.getString()) >= maxPixels) {
+                if (i > 0 || length + font.width(line.getStringWithoutFormatting()) >= maxPixels) {
```
kristofbolyai: Are you sure? Bold will be counted as regular text now..

magicus: Hm. Yeah, that is probably wrong. Otoh, so is counting the two char long formatting code for bold as well. 


For a very particular length of text, these two errors will cancel. For all other, it will be wrong. 

Do you know of a way to get the width of bold text?

kristofbolyai: `font.width` actually applies bold, counts it properly, and skips formatting codes as real chars. To be honest, I am not quite sure what the FIXME means here, and if it still is an issue..

magicus: Ok, I'll revert this part and remove the FIXME.

kristofbolyai (approved):

PR #1961, opened 2023-08-08
feat: Transcription tooltips

ShadowCat117:
Tooltips


common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -85,19 +90,27 @@ public void onNpcDialogue(NpcDialogEvent event) {
...
                 .map(s -> ((Component) s.getComponent()))
                 .toList();
 
-        Managers.TickScheduler.scheduleNextTick(() -> {
-            NpcDialogEvent transcriptedEvent =
-                    new WynnTranscriptedNpcDialogEvent(transcriptedComponents, event.getType(), event.isProtected());
-            WynntilsMod.postEvent(transcriptedEvent);
-        });
+        if (showTooltip.get()) {
+            for (Component transcriptedComponent : transcriptedComponents) {
```
ShadowCat117, line 104: Not 100% sure on this, just sending `transcriptedComponents.get(0)` seemed to work fine, but since it is a list this seemed better.

kristofbolyai:
To be honest, I can't really comment on this PR. I didn't fully review the transcription PR, so I am not sure what happens here. magicus probably knows this the best.

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
magicus: getChatMessage().stream() sounds suspicious. Is get "message" (not messages!) returning a collection?

ShadowCat117: Yeah getChatMessage() is a List of Components

magicus: Apparently we can in theory send more than one Component as NPC message. I coded this myself, but I don't remember if this actually happened, or if it was just a safety measure. But yes, better to keep it this way. It does not hurt.

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -144,9 +157,23 @@ private StyledText getStyledTextWithTranscription(
                         defaultColor);
             }
 
+            Style style;
+
+            if ((transcribeGavellian || transcribeWynnic) && !showTooltip.get()) {
+                style = Style.EMPTY.withHoverEvent(new HoverEvent(
+                        HoverEvent.Action.SHOW_TEXT,
+                        Component.translatable("feature.wynntils.transcribeMessages.transcribedFrom")));
```
magicus: Maybe also append the original wynnic/gavellian text here? (And change the i18n to end in a colon or something like that)

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -144,9 +157,23 @@ private StyledText getStyledTextWithTranscription(
...
+            } else if ((transcribeGavellian || transcribeWynnic) && showTooltip.get()) {
```
magicus: Also, I think it is better to have this as a nested if: first check if we have any of the two transcriptions, and then decide the style  etc depending on the showTooltip config.

common/src/main/java/com/wynntils/features/utilities/TranscribeMessagesFeature.java
```diff
@@ -144,9 +157,23 @@ private StyledText getStyledTextWithTranscription(
...
+                style = Style.EMPTY.withHoverEvent(
+                        new HoverEvent(HoverEvent.Action.SHOW_TEXT, Component.literal(transcriptedText)));
+
+                transcriptedText = partText;
```
magicus: I can't help but wonder if this logic can be expressed a bit more straightfoward. We have this situation:
for each part, we extract the current partText and style. We should then either:
1) keep both (no transcription), or
2) set style to have a hover with the translation (and then have no changes in the partText)
3) set style to have a hover saying this is transcribed (and with the original partText, according to my suggestion), and replace the partText with the translation.

I think this could probably be expressed along these lines: 

```
if (have transcription) {
  Component hoverText = showTooltip.get() ? Component.literal(transcriptedText) : Component.translatable("feature.wynntils.transcribeMessages.transcribedFrom").append(partText)
  style  =  Style.EMPTY.withHoverEvent(new HoverEvent(HoverEvent.Action.SHOW_TEXT, hoverText);
  newPart = new StyledTextPart(showTooltip.get() ? partText : transcriptedText, style, null, Style.EMPTY);
} else {
   newPart = part
}
```

magicus (approved):

ShadowCat117:
I haven't simplified the StyledTextPart yet, I'll finish it up tomorrow

magicus:
I can go either way; take this PR as it is, or wait for you to change it. If you agree that you think it will be better, I'll wait for your change, but if you think it's just shuffling around the logic for no good, we can skip it.

ShadowCat117:
Ok, can take it as is. I can always sneak it into my next pr ;)

PR #1963, opened 2023-08-08
fix: Fix overlays not sticking to cursor when moving them

Mcrtin:
in OverlayManagementScreen we adapt the drag x and y to snap to edges but the distance we snap doesn't get saved so the Overlay snaps further away from the cursor every time you move it. Maybe the name "totalDrag" is a bit misleading.

### DIFF
--- a/common/src/main/java/com/wynntils/screens/overlays/placement/OverlayManagementScreen.java
+++ b/common/src/main/java/com/wynntils/screens/overlays/placement/OverlayManagementScreen.java
@@ -610,13 +610,14 @@ public final class OverlayManagementScreen extends WynntilsScreen {
 
     // Pair<dragX, dragY>
     private Pair<Double, Double> calculateDragAfterSnapping(double dragX, double dragY) {
+        if (!snappingEnabled) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/screens/overlays/placement/OverlayManagementScreen.java
```diff
@@ -612,6 +614,11 @@ private Pair<Double, Double> calculateDragAfterSnapping(double dragX, double dra
             return new Pair<>(dragX, dragY);
         }
 
+        dragX += snapOffsetX;
+        dragY += snapOffsetY;
+        snapOffsetX = dragX;
+        snapOffsetY = dragY;
```
kristofbolyai: Can you just create a variable "originalDragX/Y" and set snapOffset with that before returning? This code is currently really hard to understand from looking at it.

magicus (approved):
I think this seems reasonable, but @kristofbolyai decides.

kristofbolyai (approved):

PR #1965, opened 2023-08-08
perf: Make functions templates compiled and cached

kristofbolyai:
This PR greatly improves function speed, as they became an essential part of the mod, with many function based overlays. Now we compile a function tree only once, and just get the values. This saves us having to parse expressions more than one times. 

To give some statistics, Rafii, a heavy function user experienced a 7 ms of measured tick time improvement. That is 14%, since a tick is 50 ms.  The allocations we do is also infinitely less, reducing GC calls.

Reviewing per commit is recommended.

### DIFF
--- a/common/src/main/java/com/wynntils/core/consumers/functions/templates/Template.java
+++ b/common/src/main/java/com/wynntils/core/consumers/functions/templates/Template.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.consumers.functions.templates;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class Template {
+    private final List<TemplatePart> parts;
+
+    public Template(List<TemplatePart> parts) {
+        this.parts = parts;
+    }
+
+    public String getString() {
+        return parts.stream().map(TemplatePart::getValue).collect(Collectors.joining());
+    }
+}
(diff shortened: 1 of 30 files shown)


common/src/main/java/com/wynntils/core/consumers/functions/templates/Template.java
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under AGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.consumers.functions.templates;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class Template {
+    private final List<TemplatePart> parts;
```
Mcrtin, line 11: Idk if you like this but you can make this an unmodifiable list and then make this class a record.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -49,7 +49,7 @@ public Integer getValue(FunctionArguments arguments) {
         }
 
         @Override
-        public List<String> getAliases() {
```
magicus, line 52: I think it probably would have been better to keep the getAliases name for what the Functions override, and then introduced a new function for the caching version in the superclass. My arguments are that there are many subclass implementation, but just one or a few calls to the superclass caching version, so it will be less unnecessary changes to the code base. And secondly, I think `getAliases` sounds slightly better when the functions override it.

But if you disagree, you can keep it as it is as well. I don't feel too strongly about it.

magicus (approved):
This seems okay. I don't know the function system as well as you do, but I guess this is reasonable.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
kristofbolyai: I still want to make it protected.. but I changed the names.

PR #1971, opened 2023-08-09
fix: Update party model regexes for 2.0.3 hotfix 3

ryanzhoudev:


common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -78,7 +78,8 @@ public final class PartyModel extends Model {
 
     private static final Pattern PARTY_CREATE_SELF = Pattern.compile("§eYou have successfully created a party\\.");
 
-    private static final Pattern PARTY_INVITED = Pattern.compile("§eYou have been invited to join (.*)'s? party\\.");
+    private static final Pattern PARTY_INVITED =
+            Pattern.compile("\\s+§eYou have been invited to join (\\w{1,16})'s party!");
```
magicus: Please keep the `?` after the `s`, I'm pretty sure that still is valid for user names ending in s.

common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -122,23 +123,21 @@ public void onWorldStateChange(WorldStateEvent event) {
     public void onChatReceived(ChatMessageReceivedEvent event) {
         if (event.getMessageType() != MessageType.FOREGROUND) return;
 
-        StyledText coded = event.getOriginalStyledText();
```
magicus, line 125: I don't think this has any practical change, but it will open up for the possibility of you getting a message that has been modified by another listener.

Was there any specific reason to change this, or just that the code became shorter?

common/src/main/java/com/wynntils/models/players/PartyModel.java
ryanzhoudev: Oh yeah I didn't even notice I changed this. My bad. I think I just copied what I got from the game and replaced my username.

common/src/main/java/com/wynntils/models/players/PartyModel.java
ryanzhoudev: We shouldn't be getting modified messages since this has EventPriority.HIGHEST. But using our StyledText makes the color codes more consistent, no random reset codes anywhere.

magicus: Both event.getOriginalStyledText() and event.getStyledText() should return a StyledText, so there should be no difference in that regard..?

magicus (approved):

common/src/main/java/com/wynntils/models/players/PartyModel.java
ryanzhoudev: isn't .getOriginalStyledText exactly what the server sends us?

magicus: No, for that you need the originalComponent. The StyledText is always constructed by us, it is not a vanilla concept.

But this event is a bit confusing. We should probably give it some TLC.

PR #1975, opened 2023-08-10
fix: Transcribed NPC dialogue repeating

ShadowCat117:
Since the hover event contained the original text, when the NpcDialogEvent was triggered again, hasWynnicOrGavellian would return true


magicus (approved):
This looks somewhat akin to what I suggested as a pure refactor for readability. :) I did not realize it was needed for correctness...

ShadowCat117:
> This looks somewhat akin to what I suggested as a pure refactor for readability. :) I did not realize it was needed for correctness...

The check to not add the "Transcribed from...." tooltip to NPC dialogue was what fixed the issue. I had done the refactor last night for the tooltip pr but couldn't figure out why this issue was happening so didn't add it. Good night's sleep and the solution becomes clear :)

PR #1977, opened 2023-08-10
feat: Support shiny stats in ItemHandler

magicus:
This PR replaces #1957 and partially replaces #1941. The fix in #1941 does some attempt at handling she shiny value in itself, since it reads it for further processing, but it is not further integrated. This PR does not concern itself with using the shiny stat at all, but only makes sure that it does not break the item handler. So some variant of #1941 will be needed even after this fix,

This PR mixes ideas from both PRs. We relax the check for "same lore" in respect to shiny stats (and also durability) from #1941. But we make the changes in the ItemHandler, from #1957. However, there is no inverse model -> handler relationship here. Instead the Handler only knows about a very generic regexp to determine shiny values. Furthermore, there is now a framework in place for adding future "unstable" lines (lines that may change even if the item is essentially the same) to the lore parsing. I also took the opportunity to streamline (I hope; at least I thought so) the central code in ItemHandler.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
@@ -19,6 +19,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import net.minecraft.core.NonNullList;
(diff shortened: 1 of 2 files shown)


magicus:
The one thing I have not done, is to test this. :-) Or rather, I have tested it on the Trade Market. But I have no shiny weapon to try with.

P0keDev:
superseded by #2000

PR #1978, opened 2023-08-10
WIP: Create new map data provider architecture

magicus:
Creation of a new MapDataService, which will provide a formal way of specifying map features, categories and map icons.

I think the best way to understand this is to start by looking at the json file format. Here is an example:
https://github.com/magicus/Artemis/blob/mapdata-rewrite/common/src/main/resources/assets/wynntils/mapdata.json


magicus:
I'll close this in favor of #2158.

PR #1979, opened 2023-08-10
refactor: Rename Config.updateConfig() to store() to align with superclass

magicus:
Also remove dead code from Configurable.

### DIFF
--- a/common/src/main/java/com/wynntils/core/consumers/features/AbstractConfigurable.java
+++ b/common/src/main/java/com/wynntils/core/consumers/features/AbstractConfigurable.java
@@ -20,11 +20,6 @@ public abstract class AbstractConfigurable implements Configurable {
     }
 
     @Override
-    public void removeConfigOptions(List<Config<?>> options) {
-        configOptions.removeAll(options);
-    }
-
-    @Override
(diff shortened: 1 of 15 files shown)


common/src/main/java/com/wynntils/core/consumers/features/AbstractConfigurable.java
```diff
@@ -19,11 +19,6 @@ public void addConfigOptions(List<Config<?>> options) {
         configOptions.addAll(options);
     }
 
-    @Override
-    public void removeConfigOptions(List<Config<?>> options) {
-        configOptions.removeAll(options);
-    }
-
```
kristofbolyai, line 26: Unused?

kristofbolyai (approved):

common/src/main/java/com/wynntils/core/consumers/features/AbstractConfigurable.java
magicus: Yes, it is unused. My presumption was that it was part of a planned design that never came about, but possibly there is a bug hiding somewhere. It's hard to search for "where *should* this have been called" in IntelliJ, so I cannot really say. :-)

PR #1980, opened 2023-08-10
refactor: Rename ConfigHolder to PersistedMetadata

magicus:
Wrap access in Config since we are soon going to access this from the manager instead


kristofbolyai (approved):
I assume you tested reflection still works.

PR #1981, opened 2023-08-11
chore: Update url md5 for major ID file

P0keDev:
Requires https://github.com/Wynntils/WynntilsWebsite-API/pull/127 to be merged first

### DIFF
--- a/common/src/main/resources/assets/wynntils/urls.json
+++ b/common/src/main/resources/assets/wynntils/urls.json
@@ -129,7 +129,7 @@
   },
   {
     "id": "dataStaticMajorIds",
-    "md5": "a0013f56ac6434c7b37abdccf238d159",
+    "md5": "370a3208e21274ba247c64db3d597162",
     "url": "https://raw.githubusercontent.com/Wynntils/WynntilsWebsite-API/master/major_ids.json"
   },
   {


magicus (approved):

magicus:
I'd like to point out (to @kristofbolyai too) that updating the artemis md5 is not a strict requirement. What will happen when we update the API site is that at next startup, the clients will download the new urls.json from there, and use it instead of the bundled one. The updated md5 in this file is only useful in the very weird scenario where a client has downloaded the updated major_ids.json file, but has lost their downloaded urls.json, and somehow cannot re-download from the API site again (due to github outage, or the Great Firewall of China). Then Artemis will consider the downloaded major_ids.json to be invalid and remove it.

magicus:
Of course we can keep updating it like this, but we can also make a bulk update from the API copy every once in a while.

kristofbolyai:
@magicus the best would be automatic PRs, with [ci skip] so its in the next build

magicus:
Yeah, but how..? Maybe add a GHA cron job to Artemis that pulls urls.json from the API site? Like how we do with gears from Wynn API. I could probably do that.

magicus:
Proposed solution in #1982

PR #1982, opened 2023-08-11
[ci skip] chore: Add automatic updating of urls.json

magicus:
(The category `chore` is not really right, but it is not a user-visible change either, so...)

This will (hopefully!) pull the latest version of urls.json and automatically create a PR when it is updated. For obvious practical reasons, I have not tested this. Let's see how it fares with the next update to the API repo. It is based on existing GHA scripts so I have a fairly high level of confidence in it, though.

### DIFF
--- a/.github/workflows/update-urls.yml
+++ b/.github/workflows/update-urls.yml
@@ -16,7 +16,7 @@ jobs:
           token: ${{ secrets.PRIVATE_TOKEN }}
           ref: ${{ github.head_ref }}
 
-      - name: Update Gear Data
+      - name: Update Urls
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-urls.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ secrets.PRIVATE_TOKEN }}
+          ref: ${{ github.head_ref }}
+
+      - name: Update Gear Data
```
kristofbolyai:
```suggestion
      - name: Update Urls
```

kristofbolyai (approved):

PR #1984, opened 2023-08-11
refactor: Move the PersistedMetadata into a map in the PersistedManager

magicus:
This one is a bit more tricky. Briefly, I have moved logic from PersistedMetadata, so now it is only a simple holder (will convert it to a record in a follow-up). These methods were previously called by wrappers in Config, and have now been inlined there (and simplified).

The metadata itself is also no longer stored within the Config, but separately in a map in the PersistedManager.

There are still a fair amount of wrappers and redirects, but I will address all those in follow-up bugs. (Changing everything at once will make the PR harder to read and follow.)

There is still a lot of different ways to set the value of a config. This too will be addressed later on.

I've also continued to generalize abilities to PersistedValue from subclasses.

I have done the normal testing to verify that config changing, saving, loading and resetting works as before, for an arbitrarily selected config, so it is not DOA.

### DIFF
--- a/common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
+++ b/common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
@@ -41,7 +41,7 @@ public final class PersistedManager extends Manager {
 
         Map<PersistedValue<?>, PersistedMetadata<?>> newMetadatas = new HashMap<>();
 
-        getPersisted(owner, Config.class).stream().forEach(p -> {
+        getPersisted(owner, Config.class).forEach(p -> {
             Field configField = p.a();
             Config<?> configObj;
             try {
--- a/common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
+++ b/common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
@@ -4,16 +4,58 @@
  */
 package com.wynntils.core.persisted;
 
+import com.google.common.base.CaseFormat;
 import com.wynntils.core.components.Manager;
+import com.wynntils.core.components.Managers;
+import com.wynntils.core.consumers.overlays.Overlay;
+import com.wynntils.core.json.JsonManager;
+import com.wynntils.core.persisted.config.Config;
+import com.wynntils.core.persisted.config.NullableConfig;
 import com.wynntils.utils.type.Pair;
 import java.lang.reflect.Field;
+import java.lang.reflect.Type;
 import java.util.Arrays;
+import java.util.HashMap;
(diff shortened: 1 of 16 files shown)


kristofbolyai:
Fix the conflicts and I will review this.

magicus:
@kristofbolyai Fixed

common/src/main/java/com/wynntils/core/persisted/PersistedManager.java
```diff
@@ -4,16 +4,58 @@
...
+import java.util.TreeSet;
 import org.apache.commons.lang3.reflect.FieldUtils;
 
-public class PersistedManager extends Manager {
-    public PersistedManager() {
-        super(List.of());
+public final class PersistedManager extends Manager {
+    private final Map<PersistedValue<?>, PersistedMetadata<?>> metadatas = new HashMap<>();
+    private final Set<PersistedValue<?>> persisteds = new TreeSet<>();
+
```
magicus, line 28: Yeah, I know Qodana. It's for future use.

common/src/main/java/com/wynntils/core/persisted/config/Config.java
```diff
@@ -28,90 +31,138 @@ public void touched() {
...
+    }
 
-        boolean visible = !(this instanceof HiddenConfig<?>);
+    // FIXME: Old ways of setting the value. These should be unified, but since
+    // they have slightly different semantics, let's do it carfeully step by step.
 
-        String i18nKey = configInfo.i18nKey();
+    public void setValue(T value) {
+        if (value == null && !getMetadata().isAllowNull()) {
+            WynntilsMod.warn("Trying to set null to config " + getJsonName() + ". Will be replaced by default.");
```
kristofbolyai, line 48: This should also be fixed in the new manager, so we don't silently modify configs in an enum breaks..

kristofbolyai (approved):

common/src/main/java/com/wynntils/core/persisted/config/Config.java
magicus: I'm not sure I fully understand this. Do you mean that this piece of code is not very good, and we should fix it going forwards, so we do not reset configs to default? Or are you requesting a change in this PR?

kristofbolyai: I would like this to be fixed in the PR that integrates this whole system. But since we are doing it incrementally, I am not sure when that is. Either way, I think this just corrupts configs. Either fix it here or later, I don't mind either way.

Issue #1802, opened 2023-07-22
Custom Loading Screen Stays When Joining Another's Housing


Kepler-17c:
I can reliably reproduce it myself and I've seen others have the same issue when I invite them to my housing:

When you click on a housing invitation, you're teleported and the custom loading screen is activated. However the screen stays active until you press `Esc` to manually close it. It doesn't close automatically as it should.

geekazodium:
... yall I think I found the cause for the bug

geekazodium:
when you join housing the house name is set in the place where the world name SHOULD be so it doesn't match with the matcher that detects if the world has been changed.

geekazodium:
*working on a fix

geekazodium:
fix is now complete.

Issue #1970, opened 2023-08-09
 Change license from AGPL3 to LGPL3 


magicus:
## TL;DR

If you have been pinged here, you have contributed code to Artemis. If you agree with changing license to LGPL3 for your contributions to Artemis, **please type "I agree" in a comment below**. (If you think you recently did this: we have changed the new license from GPL to LGPL, so we need a new set of agreements.)

## Background

Due to historic reasons, Artemis was created with the same license as Legacy, the [GNU Affero General Public License v3 (AGPL3)](https://www.gnu.org/licenses/agpl-3.0.en.html). This is a slightly modified version of GPL3, which prohibits public use on a server without releasing the source code.

Furthermore, the legal status of of whether modding Minecraft counts as "linking with a non-free program", and are thus prohibited under GPL, is unclear at best. The [perception in the community](https://blog.modrinth.com/p/licensing-guide) seems to be that it is not allowed.

LGPL was designed for this particular purpose. It is [GPL with additions](https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License#Differences_from_the_GPL) that explicitly make it clear that it is allowed to link with non-free programs. 

I propose that we change the Artemis license from AGPL3 to the [GNU Lesser General Public License v3 (LGPL3)](https://www.gnu.org/licenses/lgpl-3.0.en.html).

Since the code currently present in the Artemis repo was contributed under the AGPL3 license, all contributors must agree to a change. If an author does not agree, or does not reply, we might be forced to remove those commits that said author have provided.

Note that this change do not apply to other Wynntils software, such as Hades and Athena that are running on the Wynntils servers. They will keep using AGPL.

## Process

If you have been pinged in this message, please type "I agree" (or, "I do not agree") in a comment below. An Artemis maintainer will then update the list below based on the responses.

If and when all authors have agreed, we will re-license the project. If a few individual authors do not agree or does not reply, the Artemis development team might decide to remove those contributions and still go ahead with the re-license, keeping only code from those who agree.

## Contributors

Here is a complete list of all contributors to the Artemis repo, [as provided by Github](https://github.com/Wynntils/Artemis/graphs/contributors). 

All contributors has accepted.

**Legend:** ❓ = has not replied,     ✅ = has accepted,     ❌ = has not accepted

| User  | Commits  | Approval  |
|---|---|---|
| @kristofbolyai | [875 commits](https://github.com/Wynntils/Artemis/commits?author=kristofbolyai) | ✅ |
[...]

PR #1985, opened 2023-08-11
fix: custom loading screen no longer loads indefinitely when joining housing

geekazodium:
world changes were not detected if the world name was a housing name.


magicus:
Fixes #1802

magicus (approved):

common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
```diff
@@ -147,11 +148,21 @@ public void update(PlayerDisplayNameChangeEvent e) {
         Component displayName = e.getDisplayName();
         StyledText name = StyledText.fromComponent(displayName);
         Matcher m = name.getMatcher(WORLD_NAME);
+        if (setWorldIfMatched(m)) return;
+        // must check in this order as housing name regex matches anything that WORLD_NAME would match, housing names
+        // need to exclude world names.
```
magicus, line 153: As in, this is a Wynncraft rule that disallows you from naming an island "Global WC13"?

magicus:
@geekazodium I just realized you had not contributed to Wynntils before. We are in the process of switching our license from AGPL to LGPL. Can you please confirm that you accept this change at #1970? Thank you in advance.

common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
geekazodium: yeah fair point you can just do that and it would match the first rule...

geekazodium: this was my best attempt at distinguishing between housing and some world

magicus: Well, if Wynn allows you do name your island exactly like a world, then there is no chance that we can figure them apart. Unless island names have a different color or something, which does not seem to be the case.

magicus: There were some idea floating around before that we can determine if you are on an island by looking at your coordinates. That might be an alternative approach, if this turns out to be problematic.

PR #1991, opened 2023-08-12
feat: Search in scrap menu

ryanzhoudev:


magicus:
How is the layout of the scrap menu? Do we need to adjust the "searchable content area" so we don't find items that are part of the UI?

ryanzhoudev:
Yes, the top bar (items 0-8) are ui.

magicus:
I think we should try to restore the addition of a "content area" definition in SearchableContainerType, which was there for a while during the PR review that introduced the fix in matchItems, instead of hard-coding yet another exception. Do you feel up for it, or should I create the abstraction so you just can add the scrap menu?

ryanzhoudev:
I would like to give it a try

common/src/main/java/com/wynntils/models/containers/type/ContainerBounds.java
```diff
@@ -0,0 +1,51 @@
...
+    private final List<Integer> slots = new ArrayList<>();
+
+    /**
+     * @param totalRows Total number of rows in the container
+     * @param startRow 0-indexed row of the first row containing items
+     * @param startCol 0-indexed column of the first column containing items
+     * @param endRow 0-indexed row of the last row containing items
+     * @param endCol 0-indexed column of the last column containing items
+     */
+    public ContainerBounds(int totalRows, int startRow, int startCol, int endRow, int endCol) {
```
magicus: Do you really need totalRows? Afaict it is only used to determine when to stop looping over slots, but for that you can use endRow, right?

ryanzhoudev: Oh yeah

magicus (approved):

PR #1994, opened 2023-08-12
feat: Add better bank search

Mcrtin:
image
Note: this pr does currently NOT add the legacy filtered search

current issues:

- [ ] the items don't persist/storage doesn't work
- [ ] the items don't get annotated - related: #1995
- [ ] scroll doesn't work because of ContainerScrollFeature
- [ ] calling methods of a different feature to quick-jump - in progress: #2008 
- [x] items only get updated once you change the page
- [x] page gets reset when you click on an item
- [x] search widget cursor is too long (or the widget is too short)
- [ ] items keep the status of their requirements; I don't intend to fix this because it is only a small problem

### DIFF
--- a/common/src/main/java/com/wynntils/models/containers/ContainerModel.java
+++ b/common/src/main/java/com/wynntils/models/containers/ContainerModel.java
@@ -9,10 +9,14 @@ import com.wynntils.core.components.Models;
 import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.storage.Storage;
 import com.wynntils.core.text.StyledText;
+import com.wynntils.mc.event.ContainerSetContentEvent;
+import com.wynntils.mc.event.ContainerSetSlotEvent;
+import com.wynntils.utils.mc.McUtils;
 import com.wynntils.utils.type.Pair;
-
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -69,6 +78,9 @@ public final class ContainerModel extends Model {
     @Persisted
     private final Storage<Integer> finalMiscBucketPage = new Storage<>(10);
 
+    @Persisted
+    private final Storage<Map<Integer, List<ItemStack>>> bank = new Storage<>(new HashMap<>(21));
```
Mcrtin, line 82: @kristofbolyai this storage doesn't keep the data when closing the game even though I'm calling persist(). What am I doing wrong?

kristofbolyai: Firstly we only allow storage treemaps, secondly whatever you want to do with predefined sizes wont work

Mcrtin: the predefined size doesn't really matter. how should I store the personal bank contents instead? Is the nested list ok? do I just have to change the HashMap to a TreeMap?

magicus:
See https://github.com/Wynntils/Artemis/pull/2008 for details.

PR #2000, opened 2023-08-12
fix: ItemHandler should handle spell cast renames properly

magicus:
This is more or less @P0keDev's solution in #1957, but by using dependency inversion we avoid putting a direct dependency from the ItemHandler to the SpellModel.

I also restructured the process in ItemHandler.onItemStackUpdate(), and hopefully made it easier to follow. Furthermore I kept the moving of isLoreSoftMatching from LoreUtils as done in #1977.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import net.minecraft.core.NonNullList;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
@@ -34,6 +35,11 @@ public class ItemHandler extends Handler {
     private final Map<Class<?>, Integer> profilingCounts = new HashMap<>();
     // Keep this as a field just of performance reasons to skip a new allocation in annotate()
     private final List<ItemAnnotator> crashedAnnotators = new ArrayList<>();
+    private final List<Pattern> knownMarkerNames = new ArrayList<>();
+
+    public void registerKnownMarkerNames(List<Pattern> markerPatterns) {
+        knownMarkerNames.addAll(markerPatterns);
+    }
 
     public static Optional<ItemAnnotation> getItemStackAnnotation(ItemStack itemStack) {
         if (itemStack == null) return Optional.empty();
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -97,63 +103,64 @@ private void onItemStackUpdate(ItemStack existingItem, ItemStack newItem) {
...
-
+        // We need to check if the name has changed, and/or the lore has changed
         StyledText originalName = ((ItemStackExtension) existingItem).getOriginalName();
         StyledText existingName =
                 StyledText.fromComponent(existingItem.getHoverName()).getNormalized();
         StyledText newName = StyledText.fromComponent(newItem.getHoverName()).getNormalized();
 
         if (newName.equals(existingName)) {
-            // This is exactly the same item, so copy existing annotation
+            // Tthe name is identical to the existing stack; now check the lore
```
ryanzhoudev:
```suggestion
            // The name is identical to the existing stack; now check the lore
```

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -172,6 +179,34 @@ private boolean isWildcardItem(ItemStack itemStack) {
...
+        List<StyledText> secondLines = LoreUtils.getLore(secondItem);
+        int firstLinesLen = firstLines.size();
+        int secondLinesLen = secondLines.size();
+
+        // Only allow a maximum number of additional lines in the longer tooltip
+        if (Math.abs(firstLinesLen - secondLinesLen) > 3) return false;
+
+        int linesToCheck = Math.min(firstLinesLen, secondLinesLen);
+        // Prevent soft matching on tooltips that are very small
+        if (linesToCheck < 3 && firstLinesLen != secondLinesLen) return false;
```
ryanzhoudev, line 197: Why are these needed? Are there specific items that meet these criteria?

ryanzhoudev (approved):

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: This is not new code; it is moved from LoreUtils. 

I am not sure if they are needed; I added this long time ago to not give false positives for items with very short lores. But it is possible that at this point, the "soft" matching could be replaced with a hard match, since that is basically how we treat the result of this method anyway. I'd rather do that separately, though, since it is always soooo scary to change anything here. Things can break in very unexpected ways.

P0keDev (approved):
Nice! Very clever solution for fixing the dependency issue

magicus:
> Nice! Very clever solution for fixing the dependency issue

Thanks! ❤️ Once I've realized this, I was a bit annoyed on myself for not thinking of it earlier. This is the way we have all the other handlers setup; including how we register annotator callbacks in the ItemHandler itself. :)

Let's just remember this going forward that we can resolve any backward dependencies by dependency inversions.

PR #2001, opened 2023-08-13
feat: Party Management Screen suggests friends in other servers

ryanzhoudev:
And also some really overdue regex fixes. They were never updated to be used with StyledText.


common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -84,18 +101,19 @@ public void onChatReceived(ChatMessageReceivedEvent event) {
...
-            }
+            onlineFriends.put(joinMatcher.group("username"), Integer.parseInt(joinMatcher.group("server")));
+            WynntilsMod.postEvent(new FriendsEvent.Joined(joinMatcher.group("username")));
+            return;
         }
-
-        if (tryParseFriendMessages(styledText)) {
+        Matcher leaveMatcher = styledText.getMatcher(LEAVE_PATTERN);
+        if (leaveMatcher.matches()) {
+            onlineFriends.remove(leaveMatcher.group("username"));
```
kristofbolyai: extract `leaveMatcher.group("username")`

common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -109,6 +127,31 @@ public void onChatReceived(ChatMessageReceivedEvent event) {
                 return;
             }
         }
+
+        if (expectingOnlineList && styledText.getMatcher(ONLINE_FRIENDS_HEADER).matches()) {
+            // List of online friends is sent in multiple messages
+            // When we detect the first message indicating the start of the friends list, we set a flag
+            processingOnlineList = true;
+            expectingOnlineList = false;
```
kristofbolyai: Perhaps an enum for state? Same with non online friend list?

common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -180,22 +231,40 @@ private void resetData() {
     public void requestData() {
         if (McUtils.player() == null) return;
 
-        if (System.currentTimeMillis() - lastFriendRequest < 250) {
-            WynntilsMod.info("Skipping friend list request because it was requested less than 250ms ago.");
-            return;
+        if (System.currentTimeMillis() - lastFriendRequest > REQUEST_RATELIMIT) {
```
Mcrtin, line 234: I think we have a class for exactly that use (called `ThrottledSupplyer`)

common/src/main/java/com/wynntils/models/players/FriendsModel.java
```diff
@@ -109,6 +132,33 @@ public void onChatReceived(ChatMessageReceivedEvent event) {
...
+            if (onlineFriendMatcher.matches()) {
+                String username = onlineFriendMatcher.group(1);
+                int server = Integer.parseInt(onlineFriendMatcher.group(2));
+
+                onlineFriends.put(username, server);
+                WynntilsMod.info("Friend " + username + " is online on WC" + server);
+                event.setCanceled(true);
+                return;
+            } else {
+                WynntilsMod.postEvent(new FriendsEvent.OnlineListed());
```
magicus: As I general rule, try to finish the internal processing before posting events. I don't think anything can go wrong in this case, but it is more robust if you reset the  onlineMessageStatus back to ListStatus.IDLE before sending out the event, in case the event handler calls back into this model etc.

common/src/main/java/com/wynntils/models/players/FriendsModel.java
ryanzhoudev: I don't really think it is worth it to be used here. The current system works by sending a command and setting a variable, then listening for a response in a separate event. We can't really easily call .get() on the chat event without doing some pretty heavy refactoring to this class. If someone still believes this change should happen lmk.

magicus (approved):

PR #2002, opened 2023-08-13
feat: Custom loading screen fallback

ryanzhoudev:


kristofbolyai (commented):
Bit of a weird approach to this; I would rather just make an universal fix for "expection screens" than do this timeout

ryanzhoudev:
Actually exception screens was the original idea, but it is rather difficult. The invalid session exception doesn't actually have a separate screen, it is just ProgressScreen or ConnectScreen, and those both occur in normal connections. We also cannot run any regex or anything on the text they display because users could be in a different language. I thought this would be the safest way to do things, since it is extremely unlikely that connecting takes 10+ seconds.

magicus:
So what problem is this supposed to solve? That the custom loading screen does not update properly? Or that it should be removed, and we don't notice that?

ryanzhoudev:
> So what problem is this supposed to solve? That the custom loading screen does not update properly? Or that it should be removed, and we don't notice that?

This solves the issue where the custom screen sits there and spins infinitely for certain exception screens. The easiest to test is the Invalid Session one. Just disable your DevAuth temporarily and try to join the Wynn server.

magicus:
I think we need to detect if Minecraft opens a new screen while we have our custom loading screen open, and if it does (and it is unexpected) we should close our custom screen.

kristofbolyai:
An event needs to be always posted I think, sometimes we do not close it correczky..

magicus:
Just to be clear: I think this is the wrong approach. I have not been able to spend time to fully understand the situation, but I think this is the case: If we would not have the custom screen, the user would have seen an updated text. Now they don't. This implies that either the current vanilla screen is updated, or it is replaced by a new one.

I don't know which of these cases happen, but both should be able for us to detect, and if it happens, close the custom screen and replace it with the vanilla screen. In fact, that was what I thought the custom screen were doing, but it is apparently lacking in this regard.

ryanzhoudev:
May come back to this after 1.20.2

magicus:
@DonkeyBlaster Maybe open an issue so we don't forget about it?

PR #2006, opened 2023-08-13
feat: Handle Shiny stats properly

magicus:
This is based on #1941, as contributed by @geekazodium. Due to the complexity of the changes needed, I am using #1941 as a base but will continue add the constructs needed.

Note that the functions added here have a bit of "demo" character. I do not really know what should be exposed, or how. Now at least it is easy to show all your shiny stats, or shiny stats for the item you are holding. I am assuming this is a good start.


P0keDev:
do we want to create a list somewhere of known shiny stats? this would allow us to, say, index shiny stats for the purpose of chat encoding

sadly this list would have to be 100% manually curated since i don't think the wynncraft api has any info on shininess

magicus:
We can try collecting such a list, but it is probably going to a bit hard to keep it up to date.

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -238,10 +248,21 @@ protected List<String> getAliases() {
...
+    public static class HeldItemShinyStatFunction extends Function<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            ItemStack itemStack = InventoryUtils.getItemInHand();
+            Optional<ShinyStat> shinyStatOpt = Models.Shiny.getShinyStat(itemStack);
+            if (shinyStatOpt.isEmpty()) return "";
+
+            return shinyStatOpt.get().name() + ": " + shinyStatOpt.get().value();
+        }
+    }
```
kristofbolyai: I would rather see this separated into a name and value function. The all function can stay.

kristofbolyai (commented):
Otherwise looks good

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
magicus: Should we introduce a new generic type, like capped? A "holder" type, or whatever, that has a string as a description, and a number. And then HeldItemShinyStat can return such a holder, and we can get name and value from that holder. I think it is a general construct enough that it can be useful in other situations as well.

I am slightly worried about the exploding matrix if we start making special functions for too much detail, like if I split this into two, then someone will want to have a "get helmet" version and then it too needs to versions, etc. Better to try and raise the abstraction level of our function language.

kristofbolyai: Okay, I guess that can work.

kristofbolyai (approved):
Fixed this one for you.

common/src/main/java/com/wynntils/core/components/Models.java
```diff
@@ -105,6 +106,7 @@ public final class Models {
     public static final SeaskipperModel Seaskipper = new SeaskipperModel(Item);
     public static final ShamanMaskModel ShamanMask = new ShamanMaskModel(WorldState);
     public static final ShamanTotemModel ShamanTotem = new ShamanTotemModel(WorldState);
+    public static final ShinyModel Shiny = new ShinyModel();
```
ShadowCat117: Why is this in the models with dependencies list but has no dependency, should it depend on the Item model?

kristofbolyai: It should. Nice catch. Ill push a fix.

ShadowCat117:
Looks good now :)

PR #2007, opened 2023-08-13
refactor: Organise textures directory

ShadowCat117:
Feel free to suggest any better names and/or locations, I doubt this is perfect :). 
[Here's a text document with where everything now is and what's been renamed.](https://github.com/Wynntils/Artemis/files/12329400/textures.txt)

This also makes the change world and disconnect buttons actually use their offset textures when hovering.

I'd recommend just reviewing [Texture.java](https://github.com/Wynntils/Artemis/blob/a8f531edc52ce99ed1eeb1e85a840c4ac40fcdf7/common/src/main/java/com/wynntils/utils/render/Texture.java) as 99% of the rest is just renaming variables, aside from moving the texture files


magicus:
Nice!

I was vaguely thinking if we are doing this wrong with the enum. Having to hard-code image sizes seem uncessary; we should be able to get that from the image file itself. And it would be nice if we only had to drop a new image file in the resource dir, and start using it directly. 

But I don't know how we should do that. The only thing I can think of is like `Texture.get("generic/arrow_up")` and I am not sure that is better, to have a lot of string literals. It will make stuff like what you are doing here even harder.

What do you think, now that you have worked with this a bit?

ShadowCat117:
I'm not sure if having to do `Texture.get("generic/arrow_up")` is a good idea, means whenever you want to use it you have to make sure you've got the correct path, but `Texture.ARROW_UP` is a lot easier to do, especially with IntelliSense.

It definitely would be nice to be able to just drop it in, but I'm just not sure if it's fully worthwhile given it's not particularly difficult to add something new to this. I think you can use `BufferedImage` to get the width and height so I can try using that in the constructor instead of manually inputting them.

ShadowCat117:
This seems to get the job done
```
Texture(String name) {
        this.resource = new ResourceLocation("wynntils", "textures/" + name);

        try {
            InputStream inputStream =
                    Texture.class.getClassLoader().getResourceAsStream("assets/wynntils/" + this.resource.getPath());

            if (inputStream != null) {
                BufferedImage texture = ImageIO.read(inputStream);

                this.width = texture.getWidth();
                this.height = texture.getHeight();

                inputStream.close();
            } else {
                WynntilsMod.error("Texture " + name + " is missing at location " + this.resource.getPath());
            }
        } catch (IOException e) {
            WynntilsMod.error("Failed to read texture file for " + name, e);
        }
    }
```

common/src/main/resources/assets/wynntils/textures/character_selection_ui/quests.png
kristofbolyai, line 1: This is used though?

kristofbolyai:
> This seems to get the job done
> 
> ```
> Texture(String name) {
>         this.resource = new ResourceLocation("wynntils", "textures/" + name);
> 
>         try {
>             InputStream inputStream =
>                     Texture.class.getClassLoader().getResourceAsStream("assets/wynntils/" + this.resource.getPath());
> 
>             if (inputStream != null) {
>                 BufferedImage texture = ImageIO.read(inputStream);
> 
>                 this.width = texture.getWidth();
>                 this.height = texture.getHeight();
> 
>                 inputStream.close();
>             } else {
>                 WynntilsMod.error("Texture " + name + " is missing at location " + this.resource.getPath());
>             }
>         } catch (IOException e) {
>             WynntilsMod.error("Failed to read texture file for " + name, e);
>         }
>     }
> ```

I would delay this and experiment with it separately, bufferedimages always behave a bit weirdly.

common/src/main/resources/assets/wynntils/textures/character_selection_ui/quests.png
ShadowCat117: There's 2 of these textures currently, one in character selection and one in the main textures directory, they're practically identical only one has a brown pixel on the actual scroll part which looks like a mistake honestly

magicus (approved):

common/src/main/java/com/wynntils/utils/render/Texture.java
```diff
@@ -6,180 +6,183 @@
...
-
-    // Quest State Icons
-    QUEST_CANNOT_START("quest_cannot_start_icon.png", 7, 7),
-    QUEST_CAN_START("quest_can_start_icon.png", 11, 7),
-    QUEST_FINISHED("quest_finished_icon.png", 11, 7),
-    QUEST_STARTED("quest_started_icon.png", 7, 7),
-
-    // Caves screen
-    CAVE_AVALIABLE("cave_avaliable.png", 7, 7),
+    // Character Selection
```
kristofbolyai: Just a small nit. Can you use `// region ` and `// endregion`? Just so it's easier to read this.

ShadowCat117: I've also added a new line above the "sub regions" to make that a bit easier to read, or should they be defined as seperate regions?

magicus (approved):

common/src/main/java/com/wynntils/utils/render/Texture.java
magicus: I think this is fine. Maybe the icon one is a bit too broad, but otoh having too small regions is also annoying when trying to use the IntelliJ structure view.

PR #2008, opened 2023-08-13
refactor: Extract quick jump behavior to model

Mcrtin:
Quick jump needed for #1994

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
@@ -11,26 +11,20 @@ import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.Config;
 import com.wynntils.core.persisted.config.ConfigCategory;
 import com.wynntils.mc.event.ContainerClickEvent;
-import com.wynntils.mc.event.ContainerSetContentEvent;
-import com.wynntils.mc.event.ScreenClosedEvent;
 import com.wynntils.mc.event.ScreenInitEvent;
 import com.wynntils.mc.event.SetSlotEvent;
 import com.wynntils.models.containers.type.SearchableContainerType;
-import com.wynntils.utils.mc.McUtils;
-import com.wynntils.utils.wynn.ContainerUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 import net.minecraft.ChatFormatting;
-import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
-import org.lwjgl.glfw.GLFW;
 
 @ConfigCategory(Category.INVENTORY)
 public class CustomBankPagesFeature extends Feature {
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -117,51 +69,31 @@ private void getCustomJumpDestinations() {
...
-    private List<Integer> getDefaultJumpDestinations() {
-        return switch (currentContainer) {
-            case BANK -> QUICK_JUMP_DESTINATIONS;
-            case BLOCK_BANK -> BLOCK_BANK_DESTINATIONS;
-            default -> HOUSING_DEFAULT_DESTINATIONS; // this has the lowest values, so it's the safest default
-        };
-    }
-
-    @SubscribeEvent
-    public void onScreenClose(ScreenClosedEvent e) {
```
ShadowCat117, line 136: This is required, otherwise when opening a different container from the bank screen the jump buttons are still visible, although due to your changes it instead throws an error. Right click potions whilst in a bank screen and you'll see it

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
```diff
@@ -323,16 +170,8 @@ private List<Integer> parseStringToDestinations(String destinationsStr, Searchab
...
-                switch (containerType) {
-                    case BANK -> defaultValues = QUICK_JUMP_DESTINATIONS;
-                    case BLOCK_BANK -> defaultValues = BLOCK_BANK_DESTINATIONS;
-                    case BOOKSHELF, MISC_BUCKET -> defaultValues = HOUSING_DEFAULT_DESTINATIONS;
-                    default -> {
-                        return null;
-                    }
-                }
+                List<Integer> defaultValues =
+                        Models.ContainerQuickJump.getJumpDestinations(Models.Container.getContainerType());
```
ShadowCat117: This will error as whilst in the config screen we aren't in one of the containers

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
Mcrtin: works perfectly fine but ill add a listener in the quick jump model

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
Mcrtin: thx for reviewing this

ShadowCat117:
Not exactly sure where the issue is, but misc bucket, bookshelf and block bank jump pages aren't working

Mcrtin:
you mean not the custom ones but the default ones?

ShadowCat117:
Custom ones, I'm clicking to go to page 4 and it takes me to page 9 and gets stuck there on bookshelf

https://github.com/Wynntils/Artemis/assets/8337467/de7386cd-4d8a-49d9-9157-e16536b3c252

common/src/main/java/com/wynntils/features/inventory/CustomBankPagesFeature.java
ShadowCat117: I'm still getting `Index 82 out of bounds for length 63`

ShadowCat117: Actually, seems like this is an issue outside of this pr.

Mcrtin:
I thought the default destinations are the default wynncraft destinations but they are the default wynntils jump destinations

ShadowCat117:
Yeah all containers use the same bank default destinations in vanilla, even though none of them have the same number of pages as the bank, so I just set some custom defaults

Mcrtin:
there is also a bug where the quick jump button buys pages because the final page in container model is unreliable because the defaults assume bought bank pages.

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -60,19 +66,21 @@ public final class ContainerModel extends Model {
...
     @Persisted
-    private final Storage<Integer> finalBlockBankPage = new Storage<>(12);
+    private final Storage<Integer> finalBlockBankPage = new Storage<>(1);
 
     @Persisted
-    private final Storage<Integer> finalBookshelfPage = new Storage<>(10);
+    private final Storage<Integer> finalBookshelfPage = new Storage<>(1);
 
     @Persisted
-    private final Storage<Integer> finalMiscBucketPage = new Storage<>(10);
```
Mcrtin: I had to change these or else the quickJumpModel might try to jump somewhere it can't endlessly (only an assumption; haven't experienced it yet)

ShadowCat117:
> there is also a bug where the quick jump button buys pages because the final page in container model is unreliable because the defaults assume bought bank pages.

It should never buy pages as it will detect if the current page is final before trying to click the next (buy) button

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
ShadowCat117: This shouldn't be required. With this, you won't be able to jump anywhere until you visit your final page. With them set to the max, you can jump anywhere that's valid and it only doesn't work if your final page is close to a vanilla destination and you have a custom destination there.

For example, if you own 16 pages and try to jump to 16, the closest will be 17 but since you don't own it nothing will happen. It's better than not being able to jump anywhere at all until you visit the final page

ShadowCat117:
I've just made https://github.com/Wynntils/Artemis/pull/2014 which I think should do everything you need it for in https://github.com/Wynntils/Artemis/pull/1994.

magicus:
@Mcrtin I am sorry to be this blunt, but you have still not listened to what we have been saying. You have not discussed with the development team what the purpose of #1994 is, why it should be needed, or what would be the best way to achieve this. Instead, you have repeatedly failed to listen or understand our instructions to make smaller PRs. I will be totally honest: there is no way that #1994 is going to be merged. I will not even try to review it.

Despite our best efforts to tell you to make small PRs with minimal changes, with a clearly defined goal, you have created yet another spaghetti monster.

You have not tried to read existing code, and understand how the system works. It is always easier to write new code than to read and understand existing code, but if you want to contribute you need to overcome such laziness. 

You have not checked your own code for trivial mistakes before asking us to review it. Or, possibly worse, you have checked but have failed to notice.

You have overall not listened to advice, and rather argued against the dev team members on the reviews.

This is not the way to build trust in a community. And at the end of the day, it is all about trust. We need to trust the code you provide, if we are to accept it. Right now, you have zero zilch nada trust from my PoV, and I'm pretty sure the rest of the dev team feels similar.

If you want to contribute to Wynntils, you need take a step back, listen to what we are saying, making simple contributions which solves a clear problem that everyone agrees needs to be solved, listen to and learn from the feedback you are given.

Or, you can fork the code base, and go create "The Ultimate Wynncraft Mod". The LGPL gives you that right. (But it does not give you the right to use the name Wynntils, mind you.)

PR #2009, opened 2023-08-13
feat: leq and geq aliases for lte gte functions

ryanzhoudev:
https://feedback.wynntils.com/posts/240/add-geq-and-leq-as-aliases-for-gte-and-lte-functions


ShadowCat117:
Given that we have "more_than_equals" is it worth adding "greater_than_equals" just for consistency?

ryanzhoudev:
Sure

magicus (approved):

PR #2010, opened 2023-08-13
feat: Undercut setting for price match buttons

ryanzhoudev:
https://feedback.wynntils.com/posts/217/options-for-lowest-sell-offer-button-in-trademarket


common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceMatchFeature.java
```diff
@@ -44,6 +47,9 @@ public class TradeMarketPriceMatchFeature extends Feature {
     private static final int PRICE_INFO_ITEM_SLOT = 17;
     private static final Component SILVERBULL_STAR = Component.literal(" ✮").withStyle(ChatFormatting.AQUA);
 
+    @Persisted
```
magicus, line 50: Maybe this should be a percentage, rather than a fixed amount of emeralds? (But if < 1 emerald, it rounds up so it is always one emerald.)

I think a reasonable default is about 3%, that is likely to be a psychologically reasonable difference for the buyer, but still not too low for the seller.

ryanzhoudev: I don't think that is really the point. If you have some item that isn't a commodity (so gear with unique IDs), people are going to care more about the IDs than the price differing by 3%. I would pay much more than 3% for a 80% overall item compared to a 50% overall item. The 1 emerald undercut though forces the trade market to choose your item when selling a commodity like a profession material/ingredient or scrap or runes or something of that sort. In that case all you need to do is beat them by 1 emerald.

ryanzhoudev: We can do a config though if you want the percentage setting available as well

magicus: Nah, skip it then.

magicus (approved):

PR #2012, opened 2023-08-14
fix: Activity tracker scoreboard missing spaces

ryanzhoudev:
Fixes https://discord.com/channels/394189072635133952/1135962118848512052

Previously:
image

Now:
image


kristofbolyai:
Are we sure this doesn't add extra spaces in coordinates?

magicus:
Yeah, this kind of reverts the fix we did previously. :-(

In worst case, we're stuck in a difficult situation. If wynn has lines that look like this:

"Go to the but"
"cher and ..."

as well as 
"Go to the"
"butcher and"

with no space at the end, then we cannot separate instances with cutoff words from where there should be a blank.

(Personally, I believe the former variant is a bug from Wynntils side; if they really do need to cut off words they should hyphenate them.)

Otoh, they maybe changed something recently, so they once again only have the last version only. If so, this patch applies.

Mcrtin:
you also added a space between "[1 Coconut]" and "," where it should not be

ryanzhoudev:
I will see if I can figure out a better system

ryanzhoudev:
May come back to this after vacation, no point leaving this open

PR #2013, opened 2023-08-14
feat: Armor and accessory durability functions

ShadowCat117:
[An example function](https://github.com/Wynntils/Artemis/files/12336396/durability.txt), only shows when you have something equipped


common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -19,6 +21,162 @@
 import net.minecraft.world.item.ItemStack;
 
 public class InventoryFunctions {
+    public static class AccessoryCurrentDurabilityFunction extends Function<Integer> {
```
magicus: You only need to have a single getDurability function; we have already helper generic functions for dealing with capped values (to get max and current value).

ShadowCat117: Held item has a function for current, max and CappedValue durability, or are they just outdated now?

magicus: We should not have individual functions returning the current/max part of capped values -- we should only return the capped value (and the function should not be named "capped"). There are probably a handful of functions that were introduced before we had full support for capped values in functions; ideally these should be removed but maybe we want to keep backwards compatibility.

We should not be adding new such functions.

Or did I misunderstand you?

ShadowCat117: Ah ok, yeah that's what I meant so we're just keeping them so older functions don't break

magicus: Yep. At some point I'd like to deprecate them and get rid of them. Maybe we need "upfixers" for functions too... or just warn folks that they will disappear in X days time, or whatever.

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -19,6 +21,50 @@
 import net.minecraft.world.item.ItemStack;
 
 public class InventoryFunctions {
+    public static class CappedAccessoryDurabilityFunction extends Function<CappedValue> {
```
magicus:
```suggestion
    public static class AccessoryDurabilityFunction extends Function<CappedValue> {
```

I don't think you need to include Capped in the name. Once again, it was used in some places when we introduced capped variables to distinguish them from the old version, but for new functions, this is just making the function name more verbose.

ShadowCat117:
[Updated my example function, has weapon now too](https://github.com/Wynntils/Artemis/files/12340284/durability.txt)

magicus (approved):

magicus:
Thanks! This looks very good now.

PR #2014, opened 2023-08-14
refactor: Extract quick jumping logic to ContainerQuickJumpModel

ShadowCat117:
Replaces https://github.com/Wynntils/Artemis/pull/2008


common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
```diff
@@ -0,0 +1,90 @@
...
+import org.lwjgl.glfw.GLFW;
+
+public class ContainerQuickJumpModel extends Model {
+    private static final int NEXT_PAGE_SLOT = 8;
+    private static final int PREVIOUS_PAGE_SLOT = 17;
+    private static final List<Integer> BUTTON_SLOTS = List.of(7, 16, 25, 34, 43, 52);
+    private static final List<Integer> QUICK_JUMP_DESTINATIONS = List.of(1, 5, 9, 13, 17, 21);
+
+    public ContainerQuickJumpModel(ContainerModel containerModel) {
+        super(List.of(containerModel));
```
Mcrtin: this model doesn't depend on ContainerModel

common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public void jumpToDestination(int currentPage, int pageDestination, int lastPage) {
```
Mcrtin: this method needs a better name

Mcrtin (commented):
This is basically #2008 without the automatic quick jump completion. I'm not saying it is worse but you'd have to write that system every time you want to jump somewhere so for consistency and easy use I'm leaning more towards #2008 on the other hand this is more adaptive (but I don't know what else you'd use this for.

common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.containers.type;
```
kristofbolyai:
```suggestion
package com.wynntils.models.containers;
```

Move this to the suggested package.

common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
kristofbolyai: Depending on container model makes sense even if that doesn't happen at the moment. I am find with keeping or removing this.

kristofbolyai:
> This is basically #2008 without the automatic quick jump completion. I'm not saying it is worse but you'd have to write that system every time you want to jump somewhere so for consistency and easy use I'm leaning more towards #2008 on the other hand this is more adaptive (but I don't know what else you'd use this for.

what? what is "automatic quick jump completion"?

Mcrtin:
when you call jumpToDestination it doesn't actually jump there, it just jumps closer so you have to call it every time the Screen inits (that means that it may jumped) and you have to keep track of the location you want to jump to

Mcrtin:
so you basically have to copy half of CustomBankPagesFeature just so you can quick-jump somewhere (there are 5 fields in the feature related to quick-jumping)

common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
ShadowCat117: Oops

common/src/main/java/com/wynntils/models/containers/type/ContainerQuickJumpModel.java
ShadowCat117: Well now it fully depends on it :)

ShadowCat117:
No longer needed since https://github.com/Wynntils/Artemis/pull/1994 is closed

PR #2023, opened 2023-08-18
feat: Custom Bank Page Names

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/54dc4ed0-085c-4f0a-ada8-54b04f777efa


kristofbolyai:
Check out upfixers for fixing configs

ShadowCat117:
> Check out upfixers for fixing configs

Is there no support for storage upfixing yet or can I just not find it? I moved final pages from the Container to the new Bank model but resetting them shouldn't be too much of a problem if it's not possible yet.

kristofbolyai:
Yeah storage upfixers don't have an implementation yet

magicus:
Nice! I'll have a look as soon as I can.

common/src/main/java/com/wynntils/core/persisted/upfixers/config/CustomBankQuickJumpsUpfixer.java
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.persisted.upfixers.config;
+
+import com.wynntils.core.persisted.upfixers.RenamedPrefixesUpfixer;
+import com.wynntils.utils.type.Pair;
+import java.util.List;
+
+public class CustomBankQuickJumpsUpfixer extends RenamedPrefixesUpfixer {
```
kristofbolyai, line 11: Checking `MapToMainMapRenamedConfigsUpfixer`, I think you can use `customBankPagesFeature.` -> `customBankQuickJumpsFeature` if only the feature name changed.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -132,6 +132,7 @@ public void onContainerClose(ContainerCloseEvent.Post event) {
     @SubscribeEvent
     public void onInventoryKeyPress(InventoryKeyPressEvent event) {
         if (event.getKeyCode() != GLFW.GLFW_KEY_ENTER) return;
+        if (Models.Bank.isEditingName()) return;
```
kristofbolyai: Can't you call something like `screen.getFocusedTextBox() == searchWidget`?

common/src/main/java/com/wynntils/core/persisted/upfixers/config/CustomBankQuickJumpsUpfixer.java
ShadowCat117: That was where I looked to see how upfixing worked, but it didn't seem to work for me, just deleted the configs. Unless I'm blind and am missing an obvious mistake in this

`private static final List<Pair<String, String>> RENAMED_PREFIXES =
            List.of(Pair.of("customBankPagesFeature.", "customBankQuickJumpsFeature."));`

common/src/main/java/com/wynntils/mc/mixin/AbstractContainerScreenMixin.java
```diff
@@ -34,6 +35,17 @@ private void renderPost(PoseStack client, int mouseX, int mouseY, float partialT
                 (AbstractContainerScreen<?>) (Object) this, client, mouseX, mouseY, partialTicks, this.hoveredSlot));
     }
 
+    @Inject(method = "renderLabels(Lcom/mojang/blaze3d/vertex/PoseStack;II)V", at = @At("HEAD"), cancellable = true)
+    private void renderLabels(PoseStack poseStack, int mouseX, int mouseY, CallbackInfo ci) {
+        ContainerRenderLabelEvent event =
```
kristofbolyai: I would love to see this an event that you can modify the rendered text, with `WrapOperation` like `ItemTooltipRenderEvent`. However that is a much more complex mixin. I can write it for you, or you can go on the adventure of figuring that out for yourself. (If it is too messy because of bad MC source code, let me know)

kristofbolyai: (Just checked the source, this is rather easy if you know what you do.)

ShadowCat117: I'll give it a go then :)

ShadowCat117: Having trouble getting it working, feel free to add it whenever you're free and I'll just look at how you've done it to see what I was doing wrong

kristofbolyai: There you go. Ping me on discord if you don't understand parts of it. https://github.com/Wynntils/Artemis/pull/2023/commits/11b7924e7da5ae22fb8352bc555062dd08afa86f

common/src/main/java/com/wynntils/core/persisted/upfixers/config/CustomBankQuickJumpsUpfixer.java
kristofbolyai: It should work? Maybe try debugging it.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,156 @@
...
+    private boolean editingName;
+    private int currentPage = 1;
+    private SearchableContainerType currentContainer;
+
+    public BankModel(ContainerModel containerModel) {
+        super(List.of(containerModel));
+    }
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent e) {
```
kristofbolyai: I see you are relying on the same event in a feature where you call the model. It is luck that it works, because the events are the same priority. I suggest you make this `@SubscribeEvent(priority = HIGHEST)`.

common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
```diff
@@ -106,6 +87,7 @@ public void onScreenInit(ScreenInitEvent e) {
 
     private void getCustomJumpDestinations() {
         String configDestinations;
+        SearchableContainerType currentContainer = Models.Bank.getCurrentContainer();
```
kristofbolyai: you don't use this?

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,156 @@
...
+    private final Storage<Map<Integer, String>> customBankPageNames = new Storage<>(new TreeMap<>());
+
+    @Persisted
+    private final Storage<Map<Integer, String>> customBlockBankPageNames = new Storage<>(new TreeMap<>());
+
+    @Persisted
+    private final Storage<Map<Integer, String>> customBookshelfPageNames = new Storage<>(new TreeMap<>());
+
+    @Persisted
+    private final Storage<Map<Integer, String>> customMiscBucketPageNames = new Storage<>(new TreeMap<>());
```
kristofbolyai, line 45: Maybe these should be configs.. but we don't support them here yet and I won't hold this feature back for that

common/src/main/java/com/wynntils/screens/container/widgets/ContainerEditButton.java
```diff
@@ -0,0 +1,76 @@
...
+import com.wynntils.screens.base.widgets.WynntilsButton;
+import com.wynntils.utils.render.FontRenderer;
+import com.wynntils.utils.render.RenderUtils;
+import com.wynntils.utils.render.Texture;
+import java.util.List;
+import net.minecraft.ChatFormatting;
+import net.minecraft.network.chat.Component;
+import org.lwjgl.glfw.GLFW;
+
+public class ContainerEditButton extends WynntilsButton {
```
kristofbolyai:
```suggestion
public class ContainerEditNameButton extends WynntilsButton {
```

Do we edit anything else?

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
kristofbolyai, line 1: This feature has to adapt for the new event.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public void updateFinalPage() {
+        switch (currentContainer) {
+            case BANK -> {
+                finalBankPage.store(currentPage);
+            }
+            case BLOCK_BANK -> {
+                if (currentPage > finalBlockBankPage.get()) {
+                    finalBlockBankPage.store(currentPage);
+                }
```
kristofbolyai, line 133: Why the special case?

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,156 @@
...
+            case BLOCK_BANK -> customBlockBankPageNames.get().getOrDefault(page, "");
+            case BOOKSHELF -> customBookshelfPageNames.get().getOrDefault(page, "");
+            case MISC_BUCKET -> customMiscBucketPageNames.get().getOrDefault(page, "");
+            default -> "";
+        };
+    }
+
+    public void saveCurrentPageName(String nameToSet) {
+        switch (currentContainer) {
+            case BANK -> {
```
kristofbolyai: If they fit 1 line perhaps remove `{}`. Same for every other switch.

kristofbolyai: (Only if all cases can live without `{}`)

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,156 @@
...
+            currentPage = 1;
+            return;
+        }
+
+        currentPage = Models.Container.getCurrentBankPage(screen);
+
+        editingName = false;
+    }
+
+    @SubscribeEvent
```
kristofbolyai: Just to be sure I would change the event priority here too

common/src/main/java/com/wynntils/screens/container/widgets/ContainerEditButton.java
```diff
@@ -0,0 +1,76 @@
...
+    // unused
+    @Override
+    public void onPress() {}
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (!isMouseOver(mouseX, mouseY)) return false;
+
+        if (button == GLFW.GLFW_MOUSE_BUTTON_LEFT) {
+            Models.Bank.setIsEditingName(!Models.Bank.isEditingName());
```
kristofbolyai: `toggleEditingName` in the model perhaps?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -214,16 +214,18 @@
...
-  "feature.wynntils.customBankPages.blockBankDestinations.name": "Block Bank Jump Destinations",
-  "feature.wynntils.customBankPages.bookshelfDestinations.description": "Which pages would you like your bookshelf quick jump buttons to jump to? Enter a list of six numbers separated with a comma.",
-  "feature.wynntils.customBankPages.bookshelfDestinations.name": "Bookshelf Jump Destinations",
-  "feature.wynntils.customBankPages.description": "Enable custom quick jump buttons in the bank, bookshelf & misc. bucket?",
-  "feature.wynntils.customBankPages.miscBucketDestinations.description": "Which pages would you like your misc. bucket quick jump buttons to jump to? Enter a list of six numbers separated with a comma.",
-  "feature.wynntils.customBankPages.miscBucketDestinations.name": "Misc. Bucket Jump Destinations",
-  "feature.wynntils.customBankPages.name": "Custom Bank Quick Pages",
+  "feature.wynntils.customBankPageNames.description": "Enable custom page names in bank containers?",
```
```suggestion
  "feature.wynntils.customBankPageNames.description": "Adds an option to set custom page names in bank containers",
```

common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
ShadowCat117: It's used in the switch and passing into `parseStringToDestinations`?

common/src/main/java/com/wynntils/models/containers/BankModel.java
ShadowCat117: Because block bank has different pages for your personal storage and each island storage. So don't want you having all 12 pages in personal storage, then swapping to island storage and it resetting back to page 1 as final.

When I get around to making the Island Model so we can detect which island we're on this can hopefully be handled better.

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
```diff
@@ -0,0 +1,112 @@
...
+                FontRenderer.getInstance().getFont().lineHeight,
+                null,
+                (ScreenExtension) screen);
+
+        editInput.setTextBoxInput(Models.Bank.getPageName(Models.Bank.getCurrentPage()));
+
+        screen.addRenderableWidget(editInput);
+    }
+
+    private void renderPageName(PoseStack poseStack, AbstractContainerScreen<?> screen) {
```
kristofbolyai: You didn't change this? You can use the event.

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
```diff
@@ -0,0 +1,112 @@
...
+        Models.Bank.saveCurrentPageName(editInput.getTextBoxInput());
+    }
+
+    @SubscribeEvent
+    public void onSlotClicked(ContainerClickEvent e) {
+        if (Models.Bank.getCurrentContainer() == null) return;
+        Models.Bank.toggleEditingName(false);
+    }
+
+    private void renderEditInput(AbstractContainerScreen<?> screen) {
```
kristofbolyai: addInputEditWidget? Or smth like that. This is not a render method.

common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
kristofbolyai: oh right. Sorry.

kristofbolyai: Although, maybe you could just call it there?

common/src/main/java/com/wynntils/models/containers/BankModel.java
kristofbolyai: Okay. sounds good.

common/src/main/java/com/wynntils/screens/container/widgets/ContainerEditNameButton.java
```diff
@@ -0,0 +1,76 @@
...
+    public void onPress() {}
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (!isMouseOver(mouseX, mouseY)) return false;
+
+        if (button == GLFW.GLFW_MOUSE_BUTTON_LEFT) {
+            Models.Bank.toggleEditingName(!Models.Bank.isEditingName());
+        } else if (button == GLFW.GLFW_MOUSE_BUTTON_RIGHT && !Models.Bank.isEditingName()) {
+            Models.Bank.saveCurrentPageName("");
```
kristofbolyai: Do we reset the name here?

kristofbolyai: Perhaps you should allow null in the model (use optionals) and setting to null signals no custom name.

ShadowCat117: Yes, would it be better to just remove it from the map?

kristofbolyai: Yeah. and return null also, wrapped in optionals.

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
ShadowCat117: Doing that now, you missed passing the instance into the original call so it wasn't working ;)

kristofbolyai: Oh, right. It doesn't complain until you run it.. I only compiled it.

common/src/main/java/com/wynntils/screens/container/widgets/ContainerEditNameButton.java
ShadowCat117: Is this what you had in mind? https://github.com/Wynntils/Artemis/pull/2023/commits/d845d092f5c6f2a1a8f1e909e3457236a01a1bf1

common/src/main/java/com/wynntils/core/persisted/upfixers/config/CustomBankQuickJumpsUpfixer.java
ShadowCat117: I got it, didn't remove it from the applied upfixers in config so it wouldn't upfix it again and just deleted the invalid configs

common/src/main/java/com/wynntils/screens/container/widgets/ContainerEditNameButton.java
kristofbolyai: Yes

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @SubscribeEvent
+    public void onSlotClicked(ContainerClickEvent e) {
+        if (Models.Bank.getCurrentContainer() == null) return;
+        Models.Bank.toggleEditingName(false);
+    }
+
+    private void addEditInput(AbstractContainerScreen<?> screen) {
+        if (editInput != null) return;
```
kristofbolyai, line 81: I am not sure this check is right. A screen get's reinited on resize too, not just open. Does it work well when resizing while editing a name?

ShadowCat117: Bank model sets editing to false in onScreenInit, so if you resize the screen it cancels the editing

ShadowCat117: Reason was so that if you scrolled to next page, you would then be editing the new page even though you opened the text input on previous page which just seemed like weird behaviour to me

kristofbolyai: So resizing resets editing? I guess thats acceptable

ShadowCat117: Yeah. I would imagine most people wouldn't resize mid edit but even if they did it's not exactly difficult to re enter the name

kristofbolyai (approved):
LGTM. Let's wait for magicus now.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,164 @@
...
+                    case BLOCK_BANK -> customBlockBankPageNames.get().get(page);
+                    case BOOKSHELF -> customBookshelfPageNames.get().get(page);
+                    case MISC_BUCKET -> customMiscBucketPageNames.get().get(page);
+                    default -> null;
+                };
+
+        return Optional.ofNullable(customPageName);
+    }
+
+    public void saveCurrentPageName(String nameToSet) {
```
ShadowCat117, line 93: Should I add .touched to each of these or is it not a good idea to call that a lot?

kristofbolyai: Storage .touched is delayed and batched, it is designed to be called a lot of time la (and in every place where you modify objects like lists. If you set a string or int or any other primitives, there is no need for touched

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
```diff
@@ -0,0 +1,92 @@
...
+import org.lwjgl.glfw.GLFW;
+
+@ConfigCategory(Category.INVENTORY)
+public class CustomBankPageNamesFeature extends Feature {
+    private TextInputBoxWidget editInput;
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent event) {
+        if (Models.Bank.getCurrentContainer() == null) return;
+        AbstractContainerScreen<?> screen = (AbstractContainerScreen<?>) event.getScreen();
```
magicus, line 33: Please add a newline here. We try to separate the block of "if this is the case, then return early" from real code with an empty line. At least I do that, and I think bolyai follows the same pattern.

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
```diff
@@ -0,0 +1,92 @@
...
+        if (Models.Bank.isEditingName()) {
+            event.setCanceled(true);
+            addEditInput(event.getScreen());
+        } else {
+            if (editInput != null) {
+                event.getScreen().removeWidget(editInput);
+                editInput = null;
+            }
+
+            if (Models.Bank.getPageName(Models.Bank.getCurrentPage()).isPresent()) {
```
magicus: Extract Models.Bank.getCurrentPage() to a local variable, this is a bit too much DRY warning.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -0,0 +1,188 @@
...
+        switch (currentContainer) {
```
magicus: Do something like 
```
Storage<Map<Integer, String>> pageNamesMap = switch (currentContainer) {
case BANK-> customBankPageNames;
case BLOCK_BANK -> customBlockBankPageNames;
...
}
pageNamesMap.get().put....
pageNamesMap.touched()
```

common/src/main/java/com/wynntils/features/inventory/CustomBankPageNamesFeature.java
ShadowCat117: Yeah I usually do try that, just missed this one :(

common/src/main/java/com/wynntils/models/containers/BankModel.java
magicus: ... and maybe break out the currentContainer -> pageNamesMap conversion, and then you can reuse it it `getPageName`.

edit: and in more functions, as well.

kristofbolyai (approved):

PR #2025, opened 2023-08-18
feat: Add NamedValue

magicus:
Required for #2006.


common/src/main/java/com/wynntils/utils/type/NamedValue.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.type;
+
+public record NamedValue(String name, int value) {
```
kristofbolyai, line 7: Should we make the value a double?

magicus: Let's start with having it an int, shall we? If it turns out to be insufficient, then we can change it.

kristofbolyai: Okay

kristofbolyai (approved):

PR #2026, opened 2023-08-19
fix: Tentative fix for NPC dialogue changes in latest Wynncraft hotfix

magicus:
I think this will work, but I have only tested it briefly in Detlas. At the very least, it needs more testing before we can re-enable the NPC overlay by default again.

If someone wants to take this for a spin, please go ahead! Try as different NPCs and dialogue situations as possible.

Otherwise, I believe this is pretty safe to accept into mainline; as long as `shouldSeparateNPC()` is false nothing should really change.


common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -100,45 +104,62 @@ public void removeNpcDialogExtractionDependent(Feature feature) {
...
-                handleIncomingChatScreen(message);
-                e.setCanceled(true);
-            }
-            // If we are not separating NPCs, just pass the chat screen right on. We can do
-            // nothing about it.
-            // FIXME: This is a temporary solution. We don't classify NPC messages correctly, they are marked as INFO.
-            //        Even if we don't separate NPC messages, we should still classify them correctly.
-            //        However, this hack allows us to disable the NPC dialog extraction feature (/overlay), and still
-            // get
-            //        the dialog in the chat window.
```
kristofbolyai, line 125: We still don't classify NPC messages correctly. I agree this is not the best place for the FIXME, but I think we need to have it somewhere

magicus: I'm not sure what you mean by this. NPC dialogues are sent as screens, and if we do not process the screens to pick out the NPC message, we send the entire message on as a chat. This can't be classified as anything really, since it is a conglomerate of all possible kinds of messages stoved into one, so it fallbacks to info.

A few NPC interactions is sent as normal messages to chat. Like in King's Recruit, where the NPCs says like "Don't stray too far!". These are correctly classified as NPC (at least last time I checked).

magicus: How do you think we should classify the chat screen messages then? As a new SCREEN type?

kristofbolyai: The examples you sent are classified correctly, yes. Other types are just "INFO". I guess that's fine, but a bit weird.

kristofbolyai: I think we could just classify them as NPC, even if we don't "process" them further. But it might not be reliable.

magicus:
I have tested it a bit now, and I think it is roughly okay. Much better than what we have right now at least :)

I just noticed that "protected" detection does not work. This makes option dialogues (like the one with the bonfire keeper) more clunky. I'll see if I can fix that too, or if it has to go into a separate PR.

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
magicus: I still don't get it, but let's take it in Discord instead.

magicus:
Another issue (which I don't know if it is new) is that NPC lines sent as normal chat are repeated whenever a new chat message appears.

magicus:
@kristofbolyai Is there a problem with this PR or can I integrate it?

kristofbolyai:
Didn't know it was ready

kristofbolyai (approved):

kristofbolyai:
Let's try again..

PR #2028, opened 2023-08-19
fix: fix beacon locations not resetting when rerolling

### LINKED ISSUE #2024: Rerolling doesn't clear possible task locations

kristofbolyai:
Fixes https://github.com/Wynntils/Artemis/issues/2024


common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -229,6 +232,14 @@ public void onWorldStateChanged(WorldStateEvent event) {
         timeLeft = 0;
     }
 
+    @SubscribeEvent
+    public void onCommandSent(CommandSentEvent event) {
+        if (event.getCommand().equals(LOOTRUN_BEACON_REROLL_COMMAND)) {
+            // Delay the clear to make sure the beacon locations are updated
+            Managers.TickScheduler.scheduleLater(() -> possibleTaskLocations = new HashSet<>(), 10);
```
magicus, line 239: Are you sure no particles will spawn within 10 ticks of this command? (Never used it, so I don't know.)

Also, if this is a new command, it should be added to the list of known Wynn commands so it can be tab expanded.

kristofbolyai: Well, it's more like a guess. Within 10 ticks wynn surely only sends the new ones only.. I could also do this the "hard" way and detect particle removals.

magicus: I mean, if new particles are spawned between 0 and 9 ticks, then they will be removed from the possibleTaskLocation list by this reset. Is that a potential problem?

kristofbolyai: No, since they are respawned every few seconds

kristofbolyai:
I will fix this later.

PR #2029, opened 2023-08-19
fix: Do not freeze/crash on invalid mention aliases

ryanzhoudev:
Adds a 1s timeout to the provided aliases field


kristofbolyai (commented):
I can't say I like using atomics and executors for this. It feels like we missuse them (we probably do). I am not sure what the best solution for this is though..

magicus:
Oooh, I don't like this. It seems a clunky and complicated workaround. I assume this is implemented since someone ran into this problem in real life?

Can you provide an example of a regex that causes breakage? Are we allowing users to input random regexps? If so, we should verify them when they are created. Or do we generate these regexps ourselves, from user input? If so, we must strengthen the generation so we do not produce invalid regexps.

kristofbolyai:
> Can you provide an example of a regex that causes breakage? Are we allowing users to input random regexps? If so, we should verify them when they are created. Or do we generate these regexps ourselves, from user input? If so, we must strengthen the generation so we do not produce invalid regexps.

I think "bol, test," will break since we split on "," and that causes and "" regex to be created which matches everything. Something like that.

ryanzhoudev:
> Oooh, I don't like this. It seems a clunky and complicated workaround. I assume this is implemented since someone ran into this problem in real life?
> 
> Can you provide an example of a regex that causes breakage? Are we allowing users to input random regexps? If so, we should verify them when they are created. Or do we generate these regexps ourselves, from user input? If so, we must strengthen the generation so we do not produce invalid regexps.

They are generated from user input. There isn't really a way to verify them when created, since technically the empty regex after the | is valid. But this way with the timeout, it guarantees any generated regex will either fail or execute somewhat quickly.

ryanzhoudev:
They are done with a simple "replace , with | " at the moment.

magicus:
So we need to harden up the user input, then.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -56,10 +60,26 @@ protected void onConfigUpdate(Config<?> config) {
...
+        } else {
+            // we can filter the provided aliases and remove the invalid parts, then tell the user
+            String[] splitAliases = aliases.get().split(",");
+            List<String> filteredAliases = Arrays.stream(splitAliases)
+                    .filter(alias -> !alias.isEmpty())
+                    .map(alias -> FILTERED_CHARACTERS.matcher(alias).replaceAll(""))
+                    .toList();
+
+            if (splitAliases.length != filteredAliases.size()) {
+                McUtils.sendErrorToClient(I18n.get("feature.wynntils.chatMention.aliasesParseFailed"));
```
kristofbolyai: The user doesn't know what got filtered?

ryanzhoudev: en_us for that mentions `Names should be separated with an single comma. Special characters (except underscore, dash, and space) will be filtered out.`, should we be more specific? This is really just meant so that people can put in their nicknames or whatnot, I don't think anyone is going to input `@#($*(@#$` and expect it to work

kristofbolyai: hm okay, i guess fine as is then

magicus:
This approach looks much better than the original timeout based. I have just had a cursory look through the code, it looked sane, but if you want a deeper review by me, let me know and I'll return to it.

ryanzhoudev:
Not opposed to a review and merge, I don't have anything else to add atm

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -115,8 +115,9 @@
   "feature.wynntils.chatCoordinates.name": "Chat Coordinates",
   "feature.wynntils.chatItem.description": "Adds the ability to share items in chat.",
   "feature.wynntils.chatItem.name": "Chat Item",
-  "feature.wynntils.chatMention.aliases.description": "Any aliases you wish to get mentioned for in addition to your username. Names should be separated with an single comma.",
+  "feature.wynntils.chatMention.aliases.description": "Any aliases you wish to get mentioned for in addition to your username. Names should be separated with an single comma. Special characters (except underscore, dash, and space) will be filtered out.",
   "feature.wynntils.chatMention.aliases.name": "Aliases",
+  "feature.wynntils.chatMention.aliasesParseFailed": "One or more aliases failed to parse. Aliases should be separated by a single comma. Special characters (except underscore, dash, and space) are not allowed.",
```
magicus: This mentions underscore, but the pattern is `"[^\\w -]"`. Which one is incorrect?

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -56,10 +60,26 @@ protected void onConfigUpdate(Config<?> config) {
...
+                    "(?<!\\[)\\b" + McUtils.mc().getUser().getName() + "\\b(?!:|])", Pattern.CASE_INSENSITIVE);
+        } else {
+            // we can filter the provided aliases and remove the invalid parts, then tell the user
+            String[] splitAliases = aliases.get().split(",");
+            List<String> filteredAliases = Arrays.stream(splitAliases)
+                    .filter(alias -> !alias.isEmpty())
+                    .map(alias -> FILTERED_CHARACTERS.matcher(alias).replaceAll(""))
+                    .toList();
+
+            if (splitAliases.length != filteredAliases.size()) {
```
magicus: I'm not sure I understand this. You check that there are as many items in both lists? How does that indicate incorrect characters?

common/src/main/resources/assets/wynntils/lang/en_us.json
ryanzhoudev: Neither, \\w is alphanumeric and underscore

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
ryanzhoudev: This only indicates a parse fail, so something like `,,`. Other characters are filtered out automatically and silently. I think bolyai's comment was also regarding this

magicus: Hm. For this to work, I think the filter and map calls need to be in the opposite order, right?

Have you tested that this work?

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -104,7 +125,6 @@ public void onChat(ChatMessageReceivedEvent e) {
                 partToReplace = last;
                 matcher = mentionPattern.matcher(lastPart);
             }
-
```
magicus, line 107: Can you restore this newline? I think it made sense.

common/src/main/resources/assets/wynntils/lang/en_us.json
magicus: Oh. I see. I thought it was just A-Z. TIL. Thanks!

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
ryanzhoudev: Even if the order was reversed, that wouldn't change the list length, it would just be an empty string where applicable

It's not intended/designed to catch when some invalid char is replaced, just when an entire alias parse fails

magicus: If you do a 
```
   .filter(alias -> !alias.isEmpty())
```
after replacing all invalid characters with the empty string, then I think it would, no? That means that an input string like "OK1,,$@,OK2" would result in the list "OK1", "OK2", which is different from the list "OK1", "", "$@", "OK2".

Otherwise I do not understand what you mean by "an entire alias parse fail".

kristofbolyai (commented):
If you check this report, https://athena.wynntils.com/crash/view/401ad7466e5118d39abc217a9986e43e, you can see that there are more edge cases than we thought of. At least, I don't think we handle this even here.. (not to mention I thought this was already merged..)

ryanzhoudev:
I am happy with this now, for real. All normal word strings, (usernames, etc) will be parsed normally. All strings containing non-word characters will instead just check for spaces around the given string, since we cannot check for word boundaries anymore. This seems to work for the crash bolyai linked.

common/src/main/java/com/wynntils/features/chat/ChatMentionFeature.java
```diff
@@ -77,32 +99,34 @@ public void onChat(ChatMessageReceivedEvent e) {
...
+                while (matcher.find()) {
+                    String match = partToReplace.getString(null, PartStyle.StyleType.NONE);
+
+                    String firstPart = match.substring(0, matcher.start());
+                    String mentionPart = match.substring(matcher.start(), matcher.end());
+                    String lastPart = match.substring(matcher.end());
+
+                    PartStyle partStyle = partToReplace.getPartStyle();
+
+                    StyledTextPart first = new StyledTextPart(firstPart, partStyle.getStyle(), null, Style.EMPTY);
```
magicus, line 114: @kristofbolyai Btw, this code here is very close to the functionality needed in the PR about trade market item names. I think we have an opportunity to make some more general functionality in StyledText to manipulate styled texts based on substrings.

Perhaps if we had a version of split that also returned the part that matched the split. So here we could have done like this pseudo-code:
```
                    StyledText[] splitStyledText = partToReplace.splitAndKeep(matcher.group());

                    changes.remove(partToReplace);
                  
                    changes.add(splitStyledText[0]);
                    changes.add(new StyledText(splitStyledText[1].getString(), new Style(mentionColor.get());
                    changes.add(splitStyledText[2]);

```

kristofbolyai: That would be a bit different than split, but I see what you are getting at. I plan to look at a styled text issue later today, maybe I check this too. But I also have a lot open so..

kristofbolyai:
Perhaps integrate this after https://github.com/Wynntils/Artemis/pull/2138?

magicus:
The new StyledText.partition() method does not help, since it works on StyledText and this code is based on StyledTextParts. However, let's accept this for now. @DonkeyBlaster will try to refactor this later on to do some simpler approach, along the lines of chopping off our own name using an initial regexp, to split the message in `<our name>` and `<message sent>`, and then process the latter part, and then finally just concatenate it all back together?

magicus (approved):

PR #2030, opened 2023-08-19
feat: Add CrowdSourcedDataManager (disabled, in preparation for opt-in crowd-sourced data)

kristofbolyai:
I recommend reviewing per commit. This system is missing the user interaction parts by design. I plan to follow up with a screen for this, after the base manager is accepted.


common/src/main/java/com/wynntils/core/telemetry/type/TelemetryType.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * This enum represents the type of telemetry that is being collected.
+ * Each data class has to implement {@link Comparable}.
+ */
+public enum TelemetryType {
+    LOOTRUN_TASK_LOCATIONS(LootrunTaskLocation.class);
+
+    private final Class<?> dataClass;
+
+    TelemetryType(Class<?> dataClass) {
```
magicus: I think you can do like this: `TelemetryType(Class<? extends Comparable> dataClass)` to enforce the requirement.

common/src/main/java/com/wynntils/core/telemetry/type/TelemetryGameVersion.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package com.wynntils.core.telemetry.type;
+
+/**
+ * This enum represents the game version that the telemetry is being sent from.
+ * Not every version may have a telemetry version, and telemetry versions may be skipped.
+ * This enum is only updated whenever relevant changes are made to the game.
+ */
+public enum TelemetryGameVersion {
+    VERSION_203_HOTFIX_4("2.0.3 Hotfix 4");
```
magicus: I think this is conceptually wrong, unfortunately. :-( You cannot deduce the version of Wynncraft based on the version of Wynntils. Someone can play an non-updated version of Wynntils on a more recent Wynncraft version.

In fact, as long as Wynncraft does not add a way to detect version in-game, I am afraid we'll need an Athena service (or on github) that returns the current Wynncraft game version, and that you needs to poll at startup.

kristofbolyai: I don't think so. This enum shows that this Wynntils version's crowd sourced data collectors were made for this version of Wynncraft. If you want to contribute data, update your mod. If we want to collect data, we can bump this version, if and when needed.

If someone plays on a version that has the correct enum, we can assume that the data they collect is correct. At least, if they don't try to fake the data..

magicus: I think you should send the current Wynntils version instead. I think we should have as much logic as possible about this crowd-sourcing in Athena, and make Wynntils itself as a client as thin as possible.

At any point in time, Athena will know the current Wynncraft version, and it will also know the latest released Wynntils. We can then determine if the version of Wynntils used is too old for the data to be of any use. We could in theory have different cut-offs for different types of data. So if lootruns just changed significantly, we need the bleeding edge latest Wynntils update, but if NPC data is stable we can use any version from alpha-123 and on.

common/src/main/java/com/wynntils/core/crowdsource/CrowdSourcedData.java
```diff
@@ -0,0 +1,97 @@
...
+ * A holder class for crowd sourced data.
+ * This class is only meant to be modified by the {@link CrowdSourcedDataManager}.
+ * If you need to access crowd sourced data, use the {@link CrowdSourcedDataManager}.
+ */
+public class CrowdSourcedData {
+    private final Map<CrowdSourcedDataGameVersion, Map<CrowdSourcedDataType, Set<Object>>> data = new TreeMap<>();
+
+    CrowdSourcedData() {}
+
+    CrowdSourcedData(Map<CrowdSourcedDataGameVersion, Map<CrowdSourcedDataType, Set<Object>>> deserializedData) {
```
magicus, line 33: I also don't see why you would need to code in the version as an argument here. We should not really persist crowd-source data between mod restarts, but send it off to Athena, right? 

(Which, in fact, I can't really see that you are doing -- is that planned as a future "enhancement"?)

common/src/main/java/com/wynntils/core/telemetry/type/TelemetryGameVersion.java
kristofbolyai: If I had any trust in Athena at this point..

If this was something we would immediately send to a server, yes, this sounds good. But this data is to be hand collected in discord at the moment.. Even is someone would code the athena part, it is really unstable to receive this much new info. 

I can see the concern though. Maybe keeping this until we don't have a proper server to receive this data..? Again, that could take a while, so I also don't want to leave this in a bad place either.

magicus:
Maybe we have a bit different views on the overall take of crowd-sourced data. To me, it is more like a "stream" that we regularly repeat back to Athena. Your implementation here seems to be targeted towards a "gather a whole lot of data, persist it locally (and then -- I presume -- at some point send it all to Athena)".

I am worried that such an approach will scale worse for data that is inherently more in flux or more voluminous than lootrun coordinates. Gathering nodes are legio. We might want to collect e.g. scavager hunt locations that change on a daily basis. Or what about getting trade market price information? 

I think we should handle this similar to a merge between statistics and storage -- gather data where applicable, freely and cheaply, and batch it together for a while (a few seconds or so), and then send it off to Athena.

Initially, the Athena implementation can just throw it in /dev/null, but then we will need to write a backend that can extract what we want from the incoming data torrent. (Personally, I'd rather not do that in php...)

magicus:
> Initially, the Athena implementation can just throw it in /dev/null, but then we will need to write a backend that can extract what we want from the incoming data torrent. (Personally, I'd rather not do that in php...)

Thinking a bit more about this, in fact we can make Athena only receive and persist the incoming data, and have an "off-line" system (written in java) that analyze the persisted data.

common/src/main/java/com/wynntils/core/telemetry/type/TelemetryGameVersion.java
magicus: Ah, I see. I did not understand that the data were to be hand collected. So this is basically a new take on my old "record NPC locations" mode in legacy.

Fair enough, let's do it like this for now then.

magicus (approved):

kristofbolyai:
I guess let's merge this now. Might need it for 2.0.4..

PR #2031, opened 2023-08-20
feat: Per character guild contribution

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/411c76a8-c2d1-4a66-9229-730f69186249

Also redirects the contribution message as that will appear a lot more because of this.


kristofbolyai:
Why the last commit? That touched was correct?

ShadowCat117:
Oh I misunderstood what you meant for when to call touched, thought you meant to only call it when an object like a list is set, but not when changing the primitives of the list

common/src/main/java/com/wynntils/features/utilities/PerClassGuildContributionFeature.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onCommandSent(CommandSentEvent event) {
+        if (Models.Guild.getGuildName().isEmpty()) return;
+
+        String characterId = Models.Character.getId();
+
+        if (characterId.equals("-")) return;
```
kristofbolyai: P sure we have Models.Character.hasCharacter which is more appropriate

common/src/main/java/com/wynntils/features/utilities/PerClassGuildContributionFeature.java
```diff
@@ -0,0 +1,62 @@
...
+
+        Matcher contributionMatcher = CONTRIBUTION_PATTERN.matcher(command);
+
+        if (contributionMatcher.matches()) {
+            int contributionAmount = Integer.parseInt(contributionMatcher.group(1));
+
+            if (contributionAmount < 0 || contributionAmount > 100) return;
+
+            classContributions.get().put(characterId, contributionAmount);
+
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/features/utilities/PerClassGuildContributionFeature.java
```diff
@@ -0,0 +1,62 @@
...
+import com.wynntils.utils.mc.McUtils;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UTILITIES)
+public class PerClassGuildContributionFeature extends Feature {
+    private static final Pattern CONTRIBUTION_PATTERN = Pattern.compile("guild xp (\\d+).*");
```
kristofbolyai: gu xp also exists, etc. Maybe listen to the response?

kristofbolyai: (just use event priorities so the redirector doesnt mess with you)

kristofbolyai (approved):
This is alright. Maybe it would be better to add a config to hide the message when we do it, but redirect it when the user does. Up to you.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -857,6 +859,8 @@
   "feature.wynntils.partyMembersOverlay.overlay.partyMembers.maxPartyMembers.description": "How many party members should be shown?",
   "feature.wynntils.partyMembersOverlay.overlay.partyMembers.maxPartyMembers.name": "Max Party Members",
   "feature.wynntils.partyMembersOverlay.overlay.partyMembers.name": "Party Members",
+  "feature.wynntils.perClassGuildContribution.description": "Remembers what contribution XP%% you set per class and applies when you select that class.",
```
kristofbolyai: xp%%?

ShadowCat117: That's how you get it to show %, using just one says format error
image

ShadowCat117:
> This is alright. Maybe it would be better to add a config to hide the message when we do it, but redirect it when the user does. Up to you.

I wanted it to be an input on character select screen or something but we don't know id there from what I can tell. Can look into the config option though

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: `\%`

ShadowCat117: That doesn't work either. The other %'s use %% like health, mana, mem_pct etc

kristofbolyai: Yeah but.. "Format error:"?

ShadowCat117: Just how I18n does it I guess
```
public static String get(String translateKey, Object... parameters) {
        String string = language.getOrDefault(translateKey);

        try {
            return String.format(string, parameters);
        } catch (IllegalFormatException var4) {
            return "Format error: " + string;
        }
    }
```

kristofbolyai: %% should be the way.. weird that it doesnt work?

ShadowCat117: %% does work. It's a single and \% that doesn't.

magicus:
We try to use the term "character" instead of "class" when referring to the possible characters you can play. Note that Wynncraft themselves have also changed terminology (the only remnant is the `/class` command).

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -22,20 +23,26 @@
...
+public class PerCharacterGuildContributionFeature extends Feature {
     private static final Pattern CONTRIBUTION_PATTERN =
             Pattern.compile("§3You will now contribute §b(\\d+)%§3 of your XP to §b.*§3.");
 
     @Persisted
-    private final HiddenConfig<Map<String, Integer>> classContributions = new HiddenConfig<>(new TreeMap<>());
+    private final HiddenConfig<Map<String, Integer>> characterContributions = new HiddenConfig<>(new TreeMap<>());
+
+    @Persisted
+    public final Config<Boolean> hideContributionMessage = new Config<>(true);
```
ShadowCat117, line 34: Should I have this as on or off by default? This is just for the one ran automatically by the feature, not all contribution messages

kristofbolyai: On by default that case.

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -0,0 +1,76 @@
...
+    private static final Pattern CONTRIBUTION_PATTERN =
+            Pattern.compile("§3You will now contribute §b(\\d+)%§3 of your XP to §b.*§3.");
+
+    @Persisted
+    private final HiddenConfig<Map<String, Integer>> characterContributions = new HiddenConfig<>(new TreeMap<>());
+
+    @Persisted
+    public final Config<Boolean> hideContributionMessage = new Config<>(true);
+
+    private boolean commandSentFromHere = false;
```
kristofbolyai:
```suggestion
    private boolean waitingForCommandResponse = false;
```

kristofbolyai: Change this and we can merge.

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -0,0 +1,76 @@
...
+        Matcher contributionMatcher = message.getMatcher(CONTRIBUTION_PATTERN);
+
+        if (contributionMatcher.matches()) {
+            if (commandSentFromHere && hideContributionMessage.get()) {
+                event.setCanceled(true);
+                commandSentFromHere = false;
+                return;
+            } else if (commandSentFromHere) {
+                commandSentFromHere = false;
+            }
```
kristofbolyai:
```suggestion
            if (commandSentFromHere) {
                 commandSentFromHere = false;
                
                if (hideContributionMessage.get()) {
                        event.setCanceled(true);
                }
                
                return;
            }
```

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -0,0 +1,78 @@
...
+import com.wynntils.models.character.event.CharacterUpdateEvent;
+import com.wynntils.utils.mc.McUtils;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UTILITIES)
```
magicus, line 25: I'm a bit behind in the reviewing here, but should this really be enabled by default? Now, I have not really been playing with a guild, so maybe this makes all sense and everyone assumes that this is how build contributions really should have worked from Wynncraft's side. But this will radically change how `/guild xp` work, and you will get no feedback whatsoever that this is the case.

I usually don't advocate having features turn off by default, but then again, most features are very noticeable. 

A compromise, or perhaps better solution, could be to follow up the "You will now contribute..." message with another message from Wynntils, saying like "This will only apply to the current character. <onClick runCommand=`/feature disable PerCharacterGuildContribution`>Click here</onClick> to disable this functionality", whenever the change was not triggered by us so commandSentFromHere  == true.

ShadowCat117: I did think about having it start as disabled but decided against it as imo it should work like this, as going from a 106 character to a brand new one you have to remember to run the command to contribute less, and vice versa.

I think the best compromise would be to not redirect the message and instead cancel it and replace it with what you suggested

ShadowCat117: Screenshot 2023-08-24 132426

How does this look? I've removed the redirect option too as it's not really needed if the message is hidden when we run the command

magicus: I think it looks good.

PR #2044, opened 2023-08-25
feat: New guild contribution message

ShadowCat117:
262971150-fc61f023-7ddf-4d5b-b6f6-628eb239bc5b

Also removes the redirect from previous feature as no longer needed due to cancelling the message when ran by the feature.

Suggested here https://github.com/Wynntils/Artemis/pull/2031#discussion_r1304014264

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
@@ -97,7 +97,8 @@ public class PerCharacterGuildContributionFeature extends Feature {
                 .withStyle(ChatFormatting.AQUA)
                 .withStyle(ChatFormatting.UNDERLINE)
                 .withStyle(style -> style.withClickEvent(new ClickEvent(
-                        ClickEvent.Action.RUN_COMMAND, "/wynntils feature disable PerCharacterGuildContributionFeature"))));
+                        ClickEvent.Action.RUN_COMMAND,
+                        "/wynntils feature disable PerCharacterGuildContributionFeature"))));
 
         contributionMessage.append(
                 Component.literal(" to disable this functionality.").withStyle(ChatFormatting.DARK_AQUA));
--- a/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
@@ -97,7 +97,7 @@ public class PerCharacterGuildContributionFeature extends Feature {
                 .withStyle(ChatFormatting.AQUA)
                 .withStyle(ChatFormatting.UNDERLINE)
                 .withStyle(style -> style.withClickEvent(new ClickEvent(
-                        ClickEvent.Action.RUN_COMMAND, "/feature disable PerCharacterGuildContributionFeature"))));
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -73,6 +78,30 @@ public void onChatReceived(ChatMessageReceivedEvent event) {
 
             characterContributions.get().put(Models.Character.getId(), contributionAmount);
             characterContributions.touched();
+
+            sendContributionMessage(contributionAmount);
         }
     }
+
+    private void sendContributionMessage(int contributionAmount) {
+        MutableComponent contributionMessage = Component.literal(ChatFormatting.DARK_AQUA + "You will now contribute "
```
ShadowCat117, line 87: This part is the exact same as the default message but the message we send appears before the default one so I decided to cancel the default and just add this here, but would it be better to not cancel it and just remove this first part and delay sending the message by a few ticks?

kristofbolyai (commented):
Ill let magicus review this, but perhaps keeping the redirector has value in the case this feature is disabled?

ShadowCat117:
> Ill let magicus review this, but perhaps keeping the redirector has value in the case this feature is disabled?

Maybe. If the feature is disabled the message won't be shown as often but I suppose if people didn't want it redirected they can always turn off the redirect.

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
```diff
@@ -73,6 +78,30 @@ public void onChatReceived(ChatMessageReceivedEvent event) {
...
+                + ".\n");
+
+        contributionMessage.append(Component.literal("This will only apply to your current character.\n")
+                .withStyle(ChatFormatting.DARK_AQUA));
+
+        contributionMessage.append(Component.literal("Click here")
+                .withStyle(ChatFormatting.AQUA)
+                .withStyle(ChatFormatting.UNDERLINE)
+                .withStyle(style -> style.withClickEvent(new ClickEvent(
+                        ClickEvent.Action.RUN_COMMAND, "/feature disable PerCharacterGuildContributionFeature"))));
```
magicus: I thought it was just `/feature disable PerCharacterGuildContribution`?

kristofbolyai: Also use /wynntils feature

ShadowCat117: Pretty sure it didn't work without Feature at the end but will double check

ShadowCat117: Yeah needs Feature on the end

magicus: Oh. That sounds like a bug to me..! But okay, let's do it like this for now and hope we'll remember to update this instance when fixing the bug.

magicus (approved):
I don't think it makes sense to keep the redirect, since it should happen rarely when the message is cancelled, and otherwise you'll want to see the response clearly.

common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
magicus: No, that is worse I think. This is fine.

PR #2046, opened 2023-08-26
fix: War message getting recruit prefix

ShadowCat117:
WarRank

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
@@ -35,7 +35,7 @@ public class GuildRankReplacementFeature extends Feature {
 
     // Test suite: https://regexr.com/7f8kh
     private static final Pattern GUILD_MESSAGE_PATTERN =
-            Pattern.compile("§3\\[(?:§b)?★{0,5}(?:§3)?(?:§o)?.{1,16}(?:§r)?(?:§3)?\\]");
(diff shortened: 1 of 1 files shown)


kristofbolyai (approved):

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -35,7 +35,7 @@ public class GuildRankReplacementFeature extends Feature {
 
     // Test suite: https://regexr.com/7f8kh
```
magicus, line 36: You did not update the test suite..?

magicus: There should be a negative match for WAR messages.

Also, we should really try to get an internal junit test suite instead of relying on regexr...

kristofbolyai: Also for INFO that case. However, this works because of the color matching so I rather merged it.

magicus: I'm not saying you did wrong that merged it. I'm just saying we should strive to keep our regex test suites up to date, when we find new cases that cause us to have to change them.

ShadowCat117: I was going to update it but the test that's already there doesn't seem to work and I don't have champion to test how it looks exactly with changed nickname, sorry should've mentioned it

PR #2051, opened 2023-08-28
feat: junit unit testing

geekazodium:
added example regex test
added junit to dependencies


kristofbolyai (commented):
Why did you add junit when it's already used? Also, I am not sure i like this system atm

kristofbolyai:
Do you plan to move more regexes to this system? If so, mark this as draft. But please hold on until I figure out how this should look..

geekazodium:
how do I mark something as a draft?

fabric/build.gradle
```diff
@@ -77,11 +77,6 @@ dependencies {
     modRuntimeOnly "me.djtheredstoner:DevAuth-fabric:${devauth_version}"
 
     testImplementation("net.fabricmc:fabric-loader-junit:${fabric_loader_version}")
-    testImplementation("org.junit.jupiter:junit-jupiter:${junit_version}")
-}
-
-test {
-    useJUnitPlatform()
```
kristofbolyai: Why do you think you should change how our testing works? We use fabric for JUnit, so we can test with game classes as well.

kristofbolyai:
To be honest, I am not sure what you did in this PR. We already had JUnit enabled, and we already had tests. All you had to do is copy one of the tests. Although, we would have to agree on a specific regex testing format in the issue beforehand, or anywhere else.

geekazodium:
> To be honest, I am not sure what you did in this PR. We already had JUnit enabled, and we already had tests. All you had to do is copy one of the tests. Although, we would have to agree on a specific regex testing format in the issue beforehand, or anywhere else.

fair enough :3

fabric/build.gradle
geekazodium: I don't know in my head it made more sense for the tests to be in common but now I'm not so sure...

geekazodium:
> To be honest, I am not sure what you did in this PR. We already had JUnit enabled, and we already had tests. All you had to do is copy one of the tests. Although, we would have to agree on a specific regex testing format in the issue beforehand, or anywhere else.

I'm still very new, and it appears I have misunderstood the issue, the point of it was to enter all of the regex tests into unit testing, I'll get to doing that now, hopefully this time I'm not goofing up again.

geekazodium:
so umm I'm just going to go with using spotless to mark any regex without tests and manually enter the tests from the website.

fabric/build.gradle
kristofbolyai: In regular projects yes. This is a game. We can't JUnit all 3 projects, that leaves fabric the best option.

kristofbolyai:
Okay so let's break this PR into steps:

1. Agree on the format used for testing regexes.
2. When agreed, move all regexes **with** regexr tests into the system.
3. When this is merged, add additional regex test cases, if needed.

We are on step 1. At the moment, I don't like the readability of how these are structured. Let's try to improve on it.
Also, I don't think the spotless rule has any value. You can keep it while porting, but after that I think it's easier to remove it than make it unnecessary complicated.

kristofbolyai:
Also the reflection approach might not be the best way. I don't think we can load model/manager classes in some cases in test environment, purely from the fact that the game is not launched. Maybe the tests should just have copies of the regex.

geekazodium:
> Okay so let's break this PR into steps:
> 
> 1. Agree on the format used for testing regexes.
> 2. When agreed, move all regexes **with** regexr tests into the system.
> 3. When this is merged, add additional regex test cases, if needed.
> 
> We are on step 1. At the moment, I don't like the readability of how these are structured. Let's try to improve on it. Also, I don't think the spotless rule has any value. You can keep it while porting, but after that I think it's easier to remove it than make it unnecessary complicated.

okay, I will get rid of the spotless rule once this is complete

about the regex testing format, idk this current implementation may not be a good idea because all the tests need to be done to give information of which case(s) are failing.

also, the reason I wanted to use the reflection approach is to make sure the regexes are the exact same as the ones that are actually used, so there's no requirement to copy the regex which would allow for the possibility of human error not being noticed during code review.

geekazodium:
maybe I should instead use a different method of assertions

geekazodium:
however I don't have the slightest idea where to put the test methods.

fabric/build.gradle
geekazodium: moved it back

Mcrtin:
> Also the reflection approach might not be the best way. I don't think we can load model/manager classes in some cases in test environment, purely from the fact that the game is not launched. Maybe the tests should just have copies of the regex.

Why not make the fields public?

kristofbolyai:
> > Also the reflection approach might not be the best way. I don't think we can load model/manager classes in some cases in test environment, purely from the fact that the game is not launched. Maybe the tests should just have copies of the regex.
> 
> Why not make the fields public?

Because it doesn't solve the problem? You still load the class... Also, making everything public just because you have a design problem is a poor choice.

geekazodium:
> > > Also the reflection approach might not be the best way. I don't think we can load model/manager classes in some cases in test environment, purely from the fact that the game is not launched. Maybe the tests should just have copies of the regex.
> > 
> > 
> > Why not make the fields public?
> 
> Because it doesn't solve the problem? You still load the class... Also, making everything public just because you have a design problem is a poor choice.

well what else is possible? I strongly disagree with the *ctrl+c ctrl+v* approach of accessing the regexes because if a single person forgets to copy the stuff to the tests it won't catch the mistake and is effectively pointless, so what do we do? reflection access or making the fields public.

also, what cases, do you mean when it comes to loading the class? I have moved the tests to fabric and the classes can still be loaded, if there is a case where it happens to be impossible to load the class I'm sure I can catch that when copying all of the tests. right....

common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
```diff
@@ -34,10 +34,12 @@ public class GuildRankReplacementFeature extends Feature {
     public final Config<RankType> rankType = new Config<>(RankType.NAME);
 
     // Test suite: https://regexr.com/7f8kh
+    // Test method at: test/java/TestRegexs$testGuildMessagePattern
```
magicus, line 37: When we move the tests from regexr.com into our test base, we should instead remove these comments completely. The assumption should then always be that the regex is tested in it's proper test location in the test suite.

These comments were only needed since we had external testing.

fabric/src/test/java/TestRegexs.java
```diff
@@ -0,0 +1,133 @@
...
+import com.wynntils.features.chat.GuildRankReplacementFeature;
+import com.wynntils.features.ui.BulkBuyFeature;
+import com.wynntils.models.items.annotators.gui.ArchetypeAbilitiesAnnotator;
+import java.lang.reflect.Field;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class TestRegexs {
```
magicus, line 14: Why put it in fabric and not common?

kristofbolyai: @magicus we junit test in fabric because of loaders. Just like styledtext/customcolor

fabric/src/test/java/TestRegexs.java
```diff
@@ -0,0 +1,133 @@
...
+                    expected,
+                    resultGroup,
+                    String.format(
+                            "asserted regex group should be equal to string,\ngroup:\t\t%s\nexpected:\t%s\nresult:\t\t%s%n",
+                            group, expected, resultGroup));
+            return this;
+        }
+    }
+
+    public MatcherAssertions assertMatchesAny(Pattern pattern, String string) {
```
magicus, line 92: I'd like to break out all helper methods to a separate "test lib" class, and then have the actual tests structured in a way that reasonably mirrors the code base. That it, it should be an easy to figure out, 1-to-1 mapping, on how to go to the test from the source code and vice versa.

magicus: I'm not sure what is the most modern and common junit preference for doing this, if we should have like a utility class with static members, or like an abstract superclass with protected helpers, and then subclass it for the actual tests. When I last used junit seriously, the latter was the mode du jour in junit coding.

fabric/src/test/java/TestRegexs.java
```diff
@@ -0,0 +1,133 @@
...
+        assertMatchesAny(pattern, "§6 - §a✔ §f10§7²");
+        assertMatchesAny(pattern, "§6 - §a✔ §f1483§7²");
+        assertMatchesAny(pattern, "§6 - §c✖ §f244§7²");
+        assertMatchesAny(pattern, "§6 - §c✖ §f1§7²");
+        assertMatchesAny(pattern, "§6 - §a✔ §f24§7²");
+    }
+
+    @Test
+    public void testGuildMessagePattern() {
+        Pattern pattern = TestPrivateFieldsUtil.getPrivateStaticRegexPattern(
```
magicus, line 52: This still looks a bit clunky, but I'm honestly not sure how to make it smoother. 

Maybe as a builder? Like
```
new RegExTester(GuildRankReplacementFeature.GUILD_MESSAGE_PATTERN)
.matching("§3[§b★★★★★§3§oDisco reroller§3]")
.matching("§3[§b★★★★§3§obol§r§3] ")
.withGroup(1, "bol")
.notMatching("banana pie with cream")
.test();
```

magicus: The point is that it should be an absolute minimal amount of boilerplate, so the actual test strings really shine through the code framework.

magicus: Also, as can be seen in this example, I'm not 100% sure I like the reflection lookup of the fields. I think it is better to make them public so the test can access them. The point in doing that is that you will get IntelliJ assist in going from the actual regex to the test and vice versa, by the "usage" linking.

But, this also breaks encapsulation, so I'm slightly at the fence. What do you say, @kristofbolyai?

kristofbolyai: Hm, I don't like either version. Let's think about it a but more later on

fabric/src/test/java/TestRegexs.java
magicus: Hm. I assume we could have the tests in common, even if we are just able to run the tests from fabric currently? I think it feels odd to put generic code in the fabric folder. If we already do so, I'd argue that it is wrong.

kristofbolyai: I am not sure. When we added it, it was accepted as is, so we never tried it.

fabric/src/test/java/TestRegexs.java
magicus: Ideally, we'd be able to expose the field to the test, and the test only. I don't think it is possible in Java, though. Maybe some `@annotations` on the field could help? But I think not.

Mcrtin: there is `@TestOnly` or `@VisibleForTesting`. VisibleForTesting is exactly what we need but it is only since 20.0.0 according to the docs:

> A member or type annotated with VisibleForTesting claims that its visibility is higher than necessary, only for testing purposes. In particular:
> 
> - If public or protected member/type is annotated with VisibleForTesting, it's assumed that package-private access is enough for production code.
> - If package-private member/type is annotated with VisibleForTesting, it's assumed that private access is enough for production code.
> - It's illegal to annotate private member/type as VisibleForTesting.
> 
> This annotation means that the annotated element exposes internal data and breaks encapsulation of the containing class; the annotation won't prevent its use from production code, developers even won't see warnings if their IDE doesn't support the annotation. It's better to provide proper API which can be used in production as well as in tests.
> Since:
> 20.0.0

kristofbolyai: So it doesn't do anything

Mcrtin: "Go to Preferences | Editor | Inspections and enable the inspection Java | General | Test-only class or method call in production code. Now you will get a warning if you try to call members with the @VisibleForTesting annotation from production code, but not from test code." - StackOverflow
isn't that what we wanted?

Edit: couldn't find the inspection ;/

geekazodium: so like.... now what?

kristofbolyai: Well, in the real world, you don't unit tests fields. You unit test behavior. Behavior of the model. But here, we want to bend the rules, which is understandable. But test suites are not made for this. So we will always need to compromise something. Code quality, testing "usefulness" or both.

kristofbolyai:
Supersed by #2303

magicus:
I had forgotten about this. That was why I had this nagging feeling that this had been already done...

PR #2052, opened 2023-08-28
refactor: Commands implement translatable for descriptions

ryanzhoudev:
Also added `bombs` as alias for bombbell


common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -26,16 +39,16 @@
...
-  "commands.wynntils.clearCaches.deleting": "Clearing caches and closing game in 5 seconds...",
-  "commands.wynntils.clearCaches.warn": "This will clear all API and update caches, and close Minecraft. Are you sure you want to continue?",
-  "commands.wynntils.debug.profile.avg": "Average time spent: %0.3f ms/call",
-  "commands.wynntils.debug.profile.cleared": "Performance data has been cleared",
-  "commands.wynntils.debug.profile.total": "Total time spent: %d ms; total number of calls: %d",
-  "commands.wynntils.reauth.tryReauth": "Disconnecting from Hades, and trying to reauthenticate...",
-  "commands.wynntils.reloadCaches.reloading": "Reloading caches...",
-  "commands.wynntils.statistics.clickHere": "Click here to confirm",
-  "commands.wynntils.statistics.warnReset": "This will reset all your collected statistics for this character. Are you sure you want to continue?",
+  "command.wynntils.map.description": "Manage map related settings",
```
kristofbolyai, line 42: Good change. `statistics` is also plural (the only key left like that). We should change that.

kristofbolyai (approved):
This looks good to me, but perhaps let magicus review this as well. This is most likely not hard to maintain for a few days.

Also, aren't you on vacation :)?

ryanzhoudev:
> Also, aren't you on vacation :)?

Yes, but the Japan portion was the more eventful portion, and it has just ended. So I sometimes have a few hours here and there, and I will try do something during that time. Nothing big though

kristofbolyai:
Let's merge this now, no need for it to sit here for so long.

PR #2053, opened 2023-08-28
fix: Adapt profession handling to 2.0.3 hotfix #5

kristofbolyai:
Updated relevant regexes. I did not get to test everything, but I added every adaptation needed for them to work most likely. I also removed two features which are either not relevant anymore or is replaced by wynn.


common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -22,41 +21,33 @@
...
-            "(§dx[\\d\\.]+ )?§7\\[\\+(§d)?(?<gain>\\d+)§f [ⓀⒸⒷⒿⒺⒹⓁⒶⒼⒻⒾⒽ]§7 (?<name>.+) XP\\] §6\\[(?<current>\\d+)%\\]");
+            "(§.x[\\d\\.]+ )?(§.)?\\[\\+(§d)?(?<gain>\\d+)§f [ⓀⒸⒷⒿⒺⒹⓁⒶⒼⒻⒾⒽ]§7 (?<name>.+) XP\\] §6\\[(?<current>[\\d.]+)%\\]");
 
     // §2[§a+1§2 Oak Wood]
     private static final Pattern PROFESSION_NODE_HARVEST_PATTERN =
             Pattern.compile("§2\\[§a\\+\\d+§2 (?<type>.+) (?<material>.+)\\]");
 
     // §dx2.0 §7[+§d28 §fⒺ §7Scribing XP] §6[56%]
     private static final Pattern PROFESSION_CRAFT_PATTERN = Pattern.compile(
-            "(§dx[\\d\\.]+ )?§7\\[\\+(§d)?(?<gain>\\d+) §f[ⓀⒸⒷⒿⒺⒹⓁⒶⒼⒻⒾⒽ] §7(?<name>.+) XP\\] §6\\[(?<current>\\d+)%\\]");
```
ShadowCat117, line 59: I think this pattern should still work, from what I can tell the crafting message hasn't changed

kristofbolyai: Yes. But this is future proofing if they add float precision later on. Everything else has it now.

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -22,41 +21,33 @@
...
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class ProfessionModel extends Model {
-    // §7[+36§f Ⓙ§7 Farming XP] §6[9%]
-    // §dx2.0 §7[+§d93§f Ⓙ§7 Farming XP] §6[9%]
+    // §7x1 [+3952§f Ⓒ§7 Woodcutting XP] §6[14.64%]
     private static final Pattern PROFESSION_NODE_EXPERIENCE_PATTERN = Pattern.compile(
-            "(§dx[\\d\\.]+ )?§7\\[\\+(§d)?(?<gain>\\d+)§f [ⓀⒸⒷⒿⒺⒹⓁⒶⒼⒻⒾⒽ]§7 (?<name>.+) XP\\] §6\\[(?<current>\\d+)%\\]");
+            "(§.x[\\d\\.]+ )?(§.)?\\[\\+(§d)?(?<gain>\\d+)§f [ⓀⒸⒷⒿⒺⒹⓁⒶⒼⒻⒾⒽ]§7 (?<name>.+) XP\\] §6\\[(?<current>[\\d.]+)%\\]");
```
ShadowCat117, line 42: Can't verify this works with x2 XP but if it doesn't can always do a follow up fix

kristofbolyai: Yes, i would assume it does though

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
ShadowCat117: Yeah that's fair enough

ShadowCat117:
Looks good then!

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -66,21 +57,18 @@ public class ProfessionModel extends Model {
 
     private static final int GATHER_COOLDOWN_TIME = 60;
```
ShadowCat117, line 58: Oh, bit late now but these timers aren't used anymore.

kristofbolyai: I guess, but that's still true at least. So ig it can stay.

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
ShadowCat117: As luck would have it I just happened to get in a world with one and can confirm it works

PR #2054, opened 2023-08-28
fix: Guild contribution message appearing in housing

ShadowCat117:
https://discord.com/channels/394189072635133952/1023381446733352990/1145526798482952252

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
@@ -6,6 +6,7 @@ package com.wynntils.features.utilities;
 
 import com.wynntils.core.components.Models;
 import com.wynntils.core.consumers.features.Feature;
+import com.wynntils.core.mod.event.WynncraftConnectionEvent;
 import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.Config;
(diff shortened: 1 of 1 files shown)


kristofbolyai (commented):
Hm. Feels like this should be fixed in the model.. I know it's hard handling the housing there and we have been having issues around that. Have we done any hacks like this before?

ShadowCat117:
> Hm. Feels like this should be fixed in the model.. I know it's hard handling the housing there and we have been having issues around that. Have we done any hacks like this before?

Fixed in the WorldState model as in maybe have a WorldState for housing or just a boolean for inHousing? I have been making bits of progress on an island model and that was one thing I wasn't sure if it was worth adding, so I can always revisit this after finishing that up.

Or if you'd like I can change this to my other solution of when world state changed if old and new state are both WORLD then have an `ignoreCharacterChange` similar to `waitingForCommandResponse` but for sending the command

kristofbolyai:
Ideally world state model should know when we are in a housing, and keep our character ID. Afaik we can only join a house when switching to a new housing when visiting and that changing servers. Otherwise, we always spawn outside of housing, so we can parse our character id.

ShadowCat117:
Ok, I'll continue working on the island model and see if I can find a better solution.

Can either leave this as is and just add a FIXME as a reminder for myself to change it once I have that working, or just not fix this yet. I personally don't mind as I don't have the toggle off to experience this issue

kristofbolyai (approved):
Let's merge this as is now, and adapt when we can properly fix this. Leaving this hanging has no gain..

PR #2037, opened 2023-08-21
feat: Add a custom trade market result screen

kristofbolyai:
This is the new project I am working on, mainly for creating custom screens for trade market, fixing the content book speed by keeping it open while navigating it and adding a base to use for custom ability trees.

Based on https://forums.wynncraft.com/threads/how-would-you-improve-the-trade-market-gui.310365. The thread has up-to-date videos of how this looks in-game.

image
image
image
image

PR #2056, opened 2023-08-28
feat: Refactor the item filter system by separating stat providers and stat filters

kristofbolyai:
This is basically required if we want to build a sane filter implementation. I need filters as a perquisite for #2037, but I need this for implementing filters. This means this PR is high priority.

What I have done:
Filters and stat providers are now separate. This means we can implement stat providers (like level, price, name, profession type) without doing any filter parsing. We have 2 default filters: Range and String. Both of those are actually better than they used to be. Range supports greater/less than (equals). String supports strict and lazy string matching by using `""`.

Edit:

This PR is the first of a few systematic changes. Next PR, I will handle how we present the helper tooltip, errors, general UX. However, this PR is needed to be merged before that.


common/src/main/java/com/wynntils/services/itemfilter/type/StatProviderAndFilterPair.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.itemfilter.type;
+
+import com.wynntils.models.items.WynnItem;
+import java.util.List;
+
+public record StatProviderAndFilterPair<T>(ItemStatProvider<T> statProvider, StatFilter<T> statFilter) {
+    public static StatProviderAndFilterPair<Object> fromPair(ItemStatProvider itemStatProvider, StatFilter value) {
```
kristofbolyai, line 11: Yes, I did just drop the generic type here. It is not strictly necessary here and keeping it would be really hard at this point.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchHelperWidget.java
```diff
@@ -29,14 +29,13 @@ public ItemSearchHelperWidget(
     @Override
     public List<Component> getTooltipLines() {
         List<Component> helpTooltip =
-                new ArrayList<>(List.of(Component.translatable("feature.wynntils.containerSearch.tooltip")));
+                new ArrayList<>(List.of(Component.translatable("screens.wynntils.itemSearchHelperWidget.name")));
```
ryanzhoudev, line 32: Maybe this should become `screen.wynntils...` for consistency. Or we can do the refactor in some other pr.

kristofbolyai: If other screens are like this at the moment , let's keep it that way.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1910,16 +1910,12 @@
...
-  "service.wynntils.itemFilter.level.usage": "<level> or <min>-<max>",
-  "service.wynntils.itemFilter.profession.description": "Filters items which can be used for the specified profession",
-  "service.wynntils.itemFilter.profession.invalidProfession": "%s: invalid profession",
-  "service.wynntils.itemFilter.profession.name": "Profession filter",
-  "service.wynntils.itemFilter.profession.usage": "<profession name>",
-  "service.wynntils.itemFilter.unknownFilter": "%s: unknown filter",
+  "service.wynntils.itemFilter.invalidFilter": "%s: invalid filter for type: %s",
+  "service.wynntils.itemFilter.level.description": "Level of the item",
+  "service.wynntils.itemFilter.level.name": "Level",
+  "service.wynntils.itemFilter.profession.description": "List of profession related to the item",
```
```suggestion
  "service.wynntils.itemFilter.profession.description": "List of professions related to the item",
```

ryanzhoudev (approved):
Seems sane, also appears to work in game

common/src/main/java/com/wynntils/services/itemfilter/type/StatFilterFactory.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+package com.wynntils.services.itemfilter.type;
+
+import java.util.Optional;
+
+/**
+ * A factory for creating {@link StatFilter} instances.
+ * @param <T>
+ */
+public interface StatFilterFactory<T> {
```
magicus, line 13: This PR reverted the long discussion we had, where the initial "ItemFilterFactory + ItemFilter" became "ItemFilter + ItemFilterInstance".

I think this PR makes the same mistake all over. The problem is that we get a discrepancy between the concepts of the code and the concepts of the user. You would not describe "lvl" as a "filter factory" to the user, instead you describe it as a "filter". The user realizes they can put in several filters, so the idea of multiple "filter instances" is inherent in the user conceptual view (even if they perhaps would not use that exact terminology).

By calling the filter instances for "filter", and the abstract concept of the filter for "filter factory", we get the word "filter" to mean to very different things. This is just bad, and a source of confusion.

Was there any particular reason why you did not simply just change "Item" to "Stat"? (I'm not sure I think even that rename was helpful; what else about an item could you filter than it's "stats"?) Or did you just disagree with the previous resolution of renaming "filter + filter factory" to "filter instance + filter"?

kristofbolyai: Perhaps move to discord? Seems like a better place to discuss

magicus: To be clear: I fully understand and support the idea of separating the parsing of filters with their specific types (numerical and string) with the value from the item ("stat") to compare them to

The thing I am not so happy about is how this is expressed in names and concept. 

For instance, "StatProviderAndFilterPair" is quite a mouthful, and still does not describe what it does, only how it is constructed. This is basically what was previously called "ItemFilterInstance", which I believe is a much better name. I think this can be refactored back to much of the original names and concepts, while still keeping how the filter is split between parser and value providers.

PR #2057, opened 2023-08-29
fix: add missing annotations to items on trade market

### LINKED ISSUE #501: Dungeon Keys Purchase not showing accurately in TM

SpaceMonkeyy86:
Currently, text overlays, highlights, and durability arcs for items will not be displayed when in certain trade market menus. This fixes that.

Remaining Issues:
- In the menu that shows what your current sell offers are, the durability arc for gathering tools and the text overlay for teleport scrolls to dungeons won't show. This is because this info is determined from the item lore, which is not present in this menu.
- Keys and teleport scrolls to the Lost Sanctuary will not have their annotations shown in some cases. This is because of a server-side issue where the item names and lore of these items are not properly formatted or may be missing.

Resolves #501


SpaceMonkeyy86:
I've run into trouble and I could use some help. When I use the debugger, it says the match fails for an item in the menu even though the pattern clearly matches the string if you copy/paste them into an online regex engine. I don't know what's going on here.

Mcrtin:
This, even if it's working, only fixes dungeon keys, but (almost) every `itemAnnotator` is affected by the prefix (e.g. 1x, 2x etc.) of the trade market. Hence we have to remove it before we annotate the item. Maybe that is something one could easily do when #2037 is implemented (or #2037 already fixes #501 idk).

kristofbolyai:
> I've run into trouble and I could use some help. When I use the debugger, it says the match fails for an item in the menu even though the pattern clearly matches the string if you copy/paste them into an online regex engine. I don't know what's going on here.

Use regex engines for java. Like `https://www.freeformatter.com/java-regex-tester.html`. Also, this fix won't be accepted this way. We either fix all items, or none. And adapting regexes is not the way to fix this. This is a much harder issue like it seems.

SpaceMonkeyy86:
I will fix all the items in this PR; I just chose to start with dungeon keys. I will probably change this to go with what @Mcrtin suggested and strip out the color codes and prefixes before matching.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -248,6 +251,27 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
         return annotation;
     }
 
+    private StyledText simplifyName(StyledText name) {
+        String str = name.getString();
+
+        final Pattern[] PATTERNS = {
+            Pattern.compile("^§6(?:Buying|Selling) [^ ]+ (.+)§6 for .+ Each$"),
+            Pattern.compile("^§6(?:Buying|Selling) [^ ]+ (.+)$"),
+            Pattern.compile("^§7§l[^ ]+x (.+)$")
```
Mcrtin: isn't `[^ ]` supposed to be a number? I think `\\d` should work fine.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -248,6 +251,27 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
...
+        };
+
+        for (Pattern p : PATTERNS) {
+            Matcher m = p.matcher(str);
```
Mcrtin: FYI you can call `matcher` directly from `StyledText`, although in this example it'd be more expensive because `getString` isn't cached.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
SpaceMonkeyy86: The problem is \\d doesn't match commas. You could write (\\d,\\d|\\d) but that only goes up to a million and [^ ] is easier to write and does the same thing. Although I suppose [0-9,]+ would also work.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
SpaceMonkeyy86: I did it this way because there was less code that could break. I might as well leave it because of the slight optimization like you mentioned and so all the logic would fit on screen.

kristofbolyai:
Hm. This is something only magicus can accept. That might take a while.

magicus:
I just gave this a cursory look, but it looks like the wrong approach to the problem. It's very hard-coded and specific, whereas the problem is more generic in that additional information might be present in name and lore.

SpaceMonkeyy86:
I agree that the problem itself is simple to describe, but it is also full of complications. You can't simply remove the start and end anchors from the patterns and find the first match in the string because most annotators have patterns that would be too general. `DungeonKeyAnnotator` matches (almost) any string ending with "Key." If we were to pass it an item with the name "Buying 1 Decrepit Sewers Key for 10E Each" (with the correct color codes and currency signs, of course), it would match "Buying 1 Decrepit Sewers Key" and fail to show the annotation because "Buying 1 Decrepit Sewers" isn't a dungeon name. `GearAnnotator` matches any non-empty string with a color code. `EmeraldAnnotator` matches anything with "Emerald" in it. I could go on but suffice it to say that the start and end anchors are required to eliminate false positives. Therefore, if we are passed an item with extra text around the true item name, we need to extract said true name from the given name. There are also some edge cases where certain color codes in the name may be missing and this approach lets us neatly handle those in one spot without further complicating the individual annotator patterns. I still wish we didn't have to hard-code the patterns for these kinds of items, but we have to hard-code all the other patterns, too, so it shouldn't make the system any more susceptible to server-side changes. These patterns in particular need to be more specific because we need to not just match the string but also capture the correct part of it.

magicus:
I'll look at this shortly (I hope).

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -251,6 +254,32 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
         return annotation;
     }
 
+    private StyledText simplifyName(StyledText name) {
+        String str = name.getString();
+
+        final Pattern[] PATTERNS = {
+            Pattern.compile("^§6(?:Buying|Selling) [^ ]+ (.+?)(?:§6)? for .+ Each$"),
```
magicus: This is knowledge that the ItemHandler should not really have.

I was going to suggest to have a look at `registerKnownMarkerNames` for how to do a dependency inversion, and move this to the TradeMarketModel. Then I realized we still do not have a TradeMarketModel, and maybe we should not create one in this PR...

I might be willing to accept this in ItemHandler for now, but as soon as we get the TradeMarketModel, it needs to go there instead.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -251,6 +254,32 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
...
+            Pattern.compile("^§6(?:Buying|Selling) [^ ]+ (.+)$"),
+            Pattern.compile("^§7§l[^ ]+x (.+)$")
```
magicus: Can you give an example of what this is supposed to match? Preferably as a comment in the source. The other two were obvious, this one much less so.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -251,6 +254,32 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
...
+        for (Pattern p : PATTERNS) {
+            Matcher m = p.matcher(str);
+
+            if (m.matches()) {
+                str = m.group(1);
+                break;
+            }
+        }
+
+        // Sometimes the text before the item name is already white, so the server omits this prefix
```
magicus: This only happens if you match the pattern and extract a partial name, no? So move this into the for...if statement.

magicus: In fact, you don't need a break there, do the entire  return StyledText.fromString(str), and if you fall through the for loop, just return the original `name`.

common/src/main/java/com/wynntils/models/items/annotators/game/DungeonKeyAnnotator.java
```diff
@@ -19,30 +18,44 @@ public final class DungeonKeyAnnotator implements ItemAnnotator {
     private static final Pattern DUNGEON_KEY_PATTERN =
             Pattern.compile("^(?:§[46])*(?:Broken )?(?:Corrupted )?(.+) Key$");
 
-    private static final Pattern LORE_PATTERN = Pattern.compile("§7(Grants access to the|Use this item at the)");
+    private static final String[] DUNGEONS = {
```
magicus: I don't like to have a hard-coded list like this. It will break when Wynn adds or renames a dungeon.

Is this since trade market items not always have a lore? In what circumstances might items on the trade market miss the lore? I think there are other annotators as well that will break, since they rely on the lore.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
SpaceMonkeyy86: This is a good idea. Should I just leave a comment explaining that it's temporary?

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
SpaceMonkeyy86: It matches the item you click on to cancel a partially filled sell offer. On the left is the item you're selling (unmodified name/lore) and on the right is this item, representing the number of items yet to sell, and the emeralds you made from the trade so far.

common/src/main/java/com/wynntils/models/items/annotators/game/DungeonKeyAnnotator.java
SpaceMonkeyy86: Yes, items that fit the second pattern (items in the "View Your Trades" menu) have their lore completely replaced by the trade information. Therefore, annotators cannot rely on lore to match items without some fallback option to match this type of item. I already fixed the other annotators with this problem, but dungeon keys are trickier because the pattern fits any item that ends in the word "key." As far as I know, there is no way to determine if a key item is a dungeon key or not given just the item name without checking each possible name, so this list is an unfortunate necessity.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: I missed this resolved conversation before. If you really are matching a number, do show this explicitly. `[\\d,]+` would be much better.

common/src/main/java/com/wynntils/models/items/annotators/game/DungeonKeyAnnotator.java
magicus: Ok. Like my comment above, this knowledge should really reside in a DungeonModel, but we do not have such either.

Are you interested in trying to add two new models? If so, you can go ahead and do that (ask if you need assistance on what is needed). Otherwise I think we can accept this PR as it is now, and I can create the new models and refactor it later.

common/src/main/java/com/wynntils/models/items/annotators/game/HorseAnnotator.java
```diff
@@ -29,7 +29,30 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
         if (!matcher.matches()) return null;
 
         Matcher tierMatcher = LoreUtils.matchLoreLine(itemStack, 0, HORSE_TIER_PATTERN);
-        if (!tierMatcher.matches()) return null;
+
+        // The lore may be missing if the item is on the trade market
+        if (!tierMatcher.matches()) {
+            int tier;
+
+            switch (matcher.group(1)) {
```
magicus: At least there is a HorseModel. :-) Create a new HorseTier inner enum in it, move the string literals and corresponding tier numbers there, and add a method for converting from string to tier number to HorseModel.

SpaceMonkeyy86: If I'm going to do this I might as well add models for the other things too.

common/src/main/java/com/wynntils/models/items/annotators/game/DungeonKeyAnnotator.java
SpaceMonkeyy86: Sure, I can add more models.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -251,6 +260,33 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
         return annotation;
     }
 
+    private StyledText simplifyName(StyledText name) {
+        String full = name.getString();
+
+        for (Pattern p : simplifiablePatterns) {
+            Matcher m = p.matcher(full);
```
magicus, line 267: The problem you are trying to solve here (to copy the surrounding formatting to a match group of a StyledText), should be solved in StyledText itself. My first reaction was "we already do this" but then I realize this is probably only done for StyledText.split(). 

@kristofbolyai is the main architect of StyledText, so I'd like for him to weigh in. My personal, initial idea is that we should have something like `StyledText.formattedSubstring()` that you could use like:
```
if (m.matches()) {
    return name.formattedSubstring(m.group(1));
}
```

and it will basically do what you do, but properly. (Your code will fail e.g. if there are consecutive formatting codes, like a color and then bold.), presumably it will look something like this pseudo-code:
```
public StyledText formattedSubstring(String substring) {
  StyledText[] parts = this.split(substring);
  if (parts.length < 2) return StyledText.EMPTY;
  StyledTextFormatting format = parts[1];
  return new StyledText(substring, format);
}
```

Or it can/needs to be a tweaked copy of the split() method...

magicus:
Apologies for the late review, I had missed your updates to this PR.

I think this looks basically good now, the only remaining issue is the StyledText match group formatting handling.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
kristofbolyai: Doesn't StyledText#substring already apply colors if asked? But only indexes on non-formatted text?

magicus: Then it can perhaps be used. But we still need a (convenience) method in StyledText that takes a snippet String (like from a match group) of the string, and returns it as a StyledText with the formatting the snippet would have had in the original StyledText. 

That functionality is still needed for this to work properly, and the place to implement that logic is inside StyledText, not here.

kristofbolyai: I am not sure if there is a proper implementation for that..?

magicus: You mean that we do not have that functionality yet, or if it is impossible to do? I am pretty sure the former is true and the latter is false, and I'm saying this needs to be done as part of this PR.

kristofbolyai: I mean that I don't think there can be an implementation like that. You can apply a specific style to a part, but to apply multiple, you would need to find the correct parts, split the unformatted text, apply the formats back. Reasonable, but kind of also bad. Also, this can only reasonably work on the first match it finds, which makes it only work reasonably on unique text parts.

magicus: What about the "splitAndKeep()" (or perhaps even better "separate()"?) method I suggested in Donkey's PR? Like split, but do not throw away the part matching the split pattern. That would be useful here. 

Also, here we know there can be only a single match, but if it was possible there could be more, the calling code would know what to do. Maybe concatenate all but the first returned parts, for instance, if we wanted to throw away everything before the matching pattern? Etc.

kristofbolyai: Sure.

SpaceMonkeyy86: How about a `StyledTextMatcher` that can match a string with color codes (§x) onto a `StyledText` and return `StyledText` as groups? (This would do the same as `formattedSubstring` but it could neatly handle multiple matches or groups without needing to use multiple matchers.)

SpaceMonkeyy86:
That's ok. Do you want me to request a review from you in the future?

magicus:
>  Do you want me to request a review from you in the future?

If you think I've missed that it is ready for review, ping me. (I'm not sure how the "request review" thingie works in altering me, but maybe it works as well)

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
magicus: I think the proper solution is to have a "StyledText.separate(String indexStr, StyleType type)" method; either bolyai or I will write one. We can either push this as-is now, and replace the functionality when it exists, or wait for it to be written and pushed to main first. It's a bit up to you which you prefer.

SpaceMonkeyy86: Let's merge now. I added a warning for the code.

magicus (approved):

magicus:
Thank you @TwinBuilderOne for your hard work! This turned out to be a nice design and a good addition to the mod.

PR #2059, opened 2023-08-29
fix: Mana bar skull texture does not render properly

ryanzhoudev:
lmao


kristofbolyai:
I can't type? I wonder why that was not copied from legacy, or what else could have happened..

kristofbolyai (approved):

PR #2061, opened 2023-08-29
feat: better help ui for filters

kristofbolyai:
image


common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchHelperWidget.java
```diff
@@ -24,20 +35,108 @@ public ItemSearchHelperWidget(
...
-                new ArrayList<>(List.of(Component.translatable("screens.wynntils.itemSearchHelperWidget.name")));
+        return tooltipPages.get(page);
+    }
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (!isMouseOver(mouseX, mouseY)) return false;
+
+        if (button == GLFW.GLFW_MOUSE_BUTTON_LEFT) {
+            page = MathUtils.clamp(page - 1, 0, tooltipPages.size() - 1);
```
ShadowCat117: Maybe just me but I think it might be nice if the pages looped, so left clicking on first page would take you to the last page and right clicking final page would take you back to the start

ShadowCat117:
Also the tooltip in the guide screen either needs wrapping or to render to the left of the mouse as on non widescreen it cuts off on gui scale 2+. Works fine in container though

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -170,6 +171,14 @@ public void onInventoryKeyPress(InventoryKeyPressEvent event) {
         tryAutoSearch(abstractContainerScreen);
     }
 
+    @SubscribeEvent
+    public void onInventoryMouseClick(InventoryMouseClickedEvent event) {
```
ShadowCat117, line 175: This needs `if (lastItemSearchHelperWidget == null) return;`

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchHelperWidget.java
kristofbolyai: Sure

kristofbolyai:
> Also the tooltip in the guide screen either needs wrapping or to render to the left of the mouse as on non widescreen it cuts off on gui scale 2+. Works fine in container though 

Fixed this for most screens.

ShadowCat117:
Looks good now :)

PR #2063, opened 2023-08-29
feat: add any filter

kristofbolyai:
This makes it easy to filter for items with x stat. Next task is adding new stats..

image


kristofbolyai:
I would also like to note that I think this PR shows how easy it is to add new filters with the new system.

HighCrit (approved):

PR #2069, opened 2023-08-31
feat: Rarity and quality filters

ShadowCat117:
filters


common/src/main/java/com/wynntils/services/itemfilter/statproviders/QualityStatProvider.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public Class<Integer> getType() {
+        return Integer.class;
+    }
+
+    @Override
+    public List<String> getAliases() {
+        return List.of("qual");
```
kristofbolyai:
```suggestion
        return List.of("qlt");
```

maybe better?

common/src/main/java/com/wynntils/services/itemfilter/statproviders/RarityStatProvider.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Override
+    public Class<String> getType() {
+        return String.class;
+    }
+
+    @Override
+    public List<String> getAliases() {
+        return List.of("rar");
```
kristofbolyai: Maybe remove this alias? There is no common shorthand for this. Maybe even the other one aswell

kristofbolyai (approved):
Aliases are weird but otherwise good!

common/src/main/java/com/wynntils/services/itemfilter/statproviders/QualityStatProvider.java
ryanzhoudev: Imo we should have "tier" as well, I feel like people say t3 or whatever a lot

kristofbolyai: Agreed, good idea

ShadowCat117: That's true I like that better

ryanzhoudev (approved):

PR #2070, opened 2023-08-31
feat: Migrate to Wynncraft territory API, fix territories having incorrect color in advancement mode

kristofbolyai:


common/src/main/java/com/wynntils/models/players/profile/GuildProfile.java
```diff
@@ -0,0 +1,43 @@
...
+            String prefix = jsonObject.getAsJsonPrimitive("prefix").getAsString();
+
+            // Color can be null, "", or a hex string
+            String colorString = jsonObject.has("color")
+                    ? jsonObject.getAsJsonPrimitive("color").getAsString()
+                    : "";
+
+            // If color is null or empty, use the name to generate a color
+            CustomColor color = !colorString.isBlank()
+                    ? CustomColor.fromHexString(colorString)
```
ShadowCat117, line 32: Could this just be `CustomColor.fromHexString` as if it doesn't match the hex pattern it'll return CustomColor.NONE and the edge case will catch it anyway

kristofbolyai: Could be, maybe it just doesn't highligh the edge cases as well. Im fine with it either way

ShadowCat117:
Seems good to me then :)

common/src/main/java/com/wynntils/models/territories/TerritoryModel.java
```diff
@@ -173,12 +166,12 @@ private void updateTerritoryProfileMap() {
                     allTerritoryPois = territoryProfileMap.values().stream()
                             .map(TerritoryPoi::new)
                             .collect(Collectors.toSet());
-                    // TODO: Add events if territories changed
                 },
                 onError -> {
                     errorCount++;
                     if (errorCount >= MAX_ERRORS) {
-                        WynntilsMod.error("Athena user lookup has repeating failures. Disabling future lookups.");
+                        WynntilsMod.error(
```
magicus, line 173: I think we can remove this entire errorCount thing. It was just a workaround to handle bad Athena behavior. Such a failure is much more unlikely with Wynncraft.

kristofbolyai: Perhaps.. I just kept it as a fallback.

common/src/main/java/com/wynntils/services/map/pois/TerritoryPoi.java
```diff
@@ -88,13 +89,16 @@ public void renderAt(
                 && McUtils.mc().screen instanceof GuildMapScreen guildMapScreen
                 && guildMapScreen.isResourceMode()) {
             color = territoryInfo.getResourceColor();
-        } else if (!isTerritoryInfoUsable() || territoryInfo.getGuildName().equals(territoryProfile.getGuild())) {
+        } else if (isTerritoryInfoUsable()) {
             // We know the guild name with it's color
-            color = territoryProfile.getGuildColor();
+            // Get the guild profile from the model, most likely it's already cached
```
magicus, line 94: Seems these two if cases only differs in whether you do `territoryInfo.getGuildName()` or `territoryProfile.getGuild()`. Maybe extract this as a 
`guildName = isTerritoryInfoUsable() ? `territoryInfo.getGuildName() : territoryProfile.getGuild()`?

kristofbolyai: I think I kept it this way to highlight the 3 cases: hybrid mode, advancement mode, and fallback. I don't remember fully though.

common/src/main/java/com/wynntils/models/territories/TerritoryModel.java
magicus: Fair enough. We do it on a timer, so we should not DOS Wynncraft in case of failures.

common/src/main/java/com/wynntils/services/map/pois/TerritoryPoi.java
magicus: I opened https://github.com/Wynntils/Artemis/pull/2110

PR #2075, opened 2023-09-01
fix: unescaped any char in pattern

HighCrit:
This fixes `3a3 4b4 5c5 => [3, 4, 5]` or `1 3/4 => [1, 0, 3]` from being seen as a coordinates. Unfortunately, `1 3/4` is still partially matched as `[1, 0, 3]/4`, but I don't see a clear path on how to resolve that without losing a large number of valid matches.

### DIFF
--- a/common/src/main/java/com/wynntils/utils/wynn/LocationUtils.java
+++ b/common/src/main/java/com/wynntils/utils/wynn/LocationUtils.java
@@ -15,7 +15,7 @@ public final class LocationUtils {
             "(?<x>[-+]?\\d{1,6})(?:\\.\\d+)?([^0-9.+-]{1,5}(?<y>[-+]?\\d{1,3})(?:\\.\\d+)?)?[^0-9.+-]{1,5}(?<z>[-+]?\\d{1,6})(?:\\.\\d+)?");
 
     private static final Pattern STRICT_COORDINATE_PATTERN = Pattern.compile(
-            "([-+]?\\d{1,6})(?:\\.\\d+)?([,\\s]{1,2}([-+]?\\d{1,3})(?:\\.\\d+)?)?[,\\s]{1,2}([-+]?\\d{1,6})(?:\\.\\d+)?");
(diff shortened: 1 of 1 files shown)


ryanzhoudev (approved):
I think this fix is an improvement and we can merge, but looking at this regex, I don't understand why we use `[^0-9.+-]` as the separator for coordinates. Why not just have a set such as [\s,/] or something, matching all whitespace, comma, and forward slash? The existing one still matches things like `3q3q0`. This would also allow us to solve `1 3/4` by creating a massive OR and enforcing all separators are the same in a given expression.

HighCrit:
Yeah, why can the coordinate pattern match: `554 aaaa33 aaaa342`?

kristofbolyai:
Hm. Perhaps make the pattern stricter? Match only space and ,?

HighCrit:
> Hm. Perhaps make the pattern stricter? Match only space and ,?

The strict coordinate pattern, which is used in chat messages, is already that strict (with my changes)

kristofbolyai:
> > Hm. Perhaps make the pattern stricter? Match only space and ,?
> 
> The strict coordinate pattern, which is used in chat messages, is already that strict (with my changes)

Hm. Then the non-strict matcher is like that only. I guess that can be as is?

kristofbolyai:
Hm wait. So we changed both of the use cases. Okay, I see. Btw, while at this, can you change it so `ChatCoordinatesFeature` uses the strict matcher for parsing? It uses the strict one for matching, the non-strict for parsing. That can result in edge cases..

HighCrit:
> Hm wait. So we changed both of the use cases. Okay, I see. Btw, while at this, can you change it so `ChatCoordinatesFeature` uses the strict matcher for parsing? It uses the strict one for matching, the non-strict for parsing. That can result in edge cases..

I don't quite get what you mean. It seems to already use the strict one for matching and the other one for parsing

kristofbolyai:
> > Hm wait. So we changed both of the use cases. Okay, I see. Btw, while at this, can you change it so `ChatCoordinatesFeature` uses the strict matcher for parsing? It uses the strict one for matching, the non-strict for parsing. That can result in edge cases..
> 
> 
> 
> I don't quite get what you mean. It seems to already use the strict one for matching and the other one for parsing

Yeah but isn't that kind of weird? I think we should use the strict parser for matching and parsing. To avoid weird edge cases.

kristofbolyai:
I will do the fix I want later, it doesn't matter in this PR.

PR #2076, opened 2023-09-01
feat: Scrolling in guild member list

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/b014a35e-25bb-4159-a6c5-f3b40af90c2b

Also reordered some variables and added missing test suits in the model

### DIFF
--- a/common/src/main/java/com/wynntils/models/containers/ContainerModel.java
+++ b/common/src/main/java/com/wynntils/models/containers/ContainerModel.java
@@ -7,6 +7,7 @@ package com.wynntils.models.containers;
 import com.wynntils.core.components.Model;
 import com.wynntils.core.components.Models;
 import com.wynntils.core.text.StyledText;
+import com.wynntils.models.containers.type.SearchableContainerType;
 import com.wynntils.utils.type.Pair;
 import java.util.List;
 import java.util.Optional;
@@ -27,8 +28,6 @@ public final class ContainerModel extends Model {
     private static final Pattern GUILD_BANK_PATTERN =
             Pattern.compile("[a-zA-Z ]+: Bank \\((?:Everyone|High Ranked)\\)");
 
-    // Test suite: https://regexr.com/7jgvf
-    private static final Pattern GUILD_MEMBER_LIST_PATTERN = Pattern.compile("[a-zA-Z ]+: Members");
 
     // Test suite: https://regexr.com/7jh1e
     private static final Pattern LOOT_CHEST_PATTERN = Pattern.compile("Loot Chest (§.)\\[.+\\]");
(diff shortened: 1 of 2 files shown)


kristofbolyai:
Hm, unrelated to your PR but it's strange seeing the advanced filters in that GUI :)

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -19,19 +19,27 @@
...
+    // Test suite: https://regexr.com/7jh15
     public static final Pattern ABILITY_TREE_PATTERN =
             Pattern.compile("(?:Warrior|Shaman|Mage|Assassin|Archer) Abilities");
 
     // Test suite: https://regexr.com/7b4lf
     private static final Pattern GUILD_BANK_PATTERN =
             Pattern.compile("[a-zA-Z ]+: Bank \\((?:Everyone|High Ranked)\\)");
 
+    // Test suite: https://regexr.com/7jgvf
+    private static final Pattern GUILD_MEMBER_LIST_PATTERN = Pattern.compile("[a-zA-Z ]+: Members");
```
kristofbolyai: didnt we have this pattern already for adding the search bar?

ShadowCat117: It's slightly different but also works

ShadowCat117:
> Hm, unrelated to your PR but it's strange seeing the advanced filters in that GUI :)

I did think that maybe we should display the old basic one in certain screens. Unless we start making screen specific filters and on this for example you could search by rank or contribution amount

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
kristofbolyai: Perhaps it's best not to duplicate the patterns then? I am not sure where it is put though..

kristofbolyai:
> > Hm, unrelated to your PR but it's strange seeing the advanced filters in that GUI :)
> 
> I did think that maybe we should display the old basic one in certain screens. Unless we start making screen specific filters and on this for example you could search by rank or contribution amount

If you add the code to support that, it could also be a config to turn them off/on, if we ever add specific filters. I tried once but it was a bit deeper rabbit hole and I didn't feel like doing it. Most likely if you refactor the feature in a sensible way it becomes simple.

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
ShadowCat117: It'll either have to be `return SearchableContainerType.getContainerType(StyledText.fromComponent(screen.getTitle())) == SearchableContainerType.MEMBER_LIST;` or adding a getter for the title pattern in SearchableContainerType

ShadowCat117:
> If you add the code to support that, it could also be a config to turn them off/on, if we ever add specific filters. I tried once but it was a bit deeper rabbit hole and I didn't feel like doing it. Most likely if you refactor the feature in a sensible way it becomes simple.

I can add it to my backlog of things I plan to work on :)

kristofbolyai (approved):

PR #2079, opened 2023-09-02
fix: Feature names no longer end in "Feature"

### LINKED ISSUE #2047: "/feature disable FooFeature" should not end in Feature
See #2044.

SpaceMonkeyy86:
When using the /feature or /wynntils feature commands to enable or disable features, the names of the features no longer end with the word "Feature."

Example: "/feature enable FooFeature" -> "/feature enable Foo"

Resolves #2047


Mcrtin:
I think this better to add "Feature" to the string in `FeatureManager#getFeatureFromString` if it is not already ending with it. That way you can still distinguish between features and overlays using the short name and you can still add "Feature" at the end and it will still be recognized. Also this doesn't work with overlays.

TwinBuilderOne:
> I think this better to add "Feature" to the string in `FeatureManager#getFeatureFromString` if it is not already ending with it.

This should already work because `FeatureManager#getFeatureFromString` calls `Feature#getShortName`.

SpaceMonkeyy86:
Changing `FeatureManager#getFeatureFromString` would make /feature stop working. What do you mean when you say it doesn't work with overlays?

common/src/main/java/com/wynntils/core/consumers/features/Feature.java
```diff
@@ -41,7 +41,8 @@ public String getTranslatedDescription() {
     }
 
     public String getShortName() {
-        return this.getClass().getSimpleName();
+        String str = this.getClass().getSimpleName();
+        return str.endsWith("Feature") ? str.substring(0, str.length() - 7) : str;
```
kristofbolyai: String#replace("Feature", "")

kristofbolyai:
`String#replaceAll` is not `String#replace`.

Mcrtin:
in `Overlay` is also a method called `getShortName`. Although rarely used, it'd be more consistent when also changing that.

SpaceMonkeyy86:
> `String#replaceAll` is not `String#replace`.

My bad...

common/src/main/java/com/wynntils/core/consumers/features/Feature.java
```diff
@@ -41,7 +41,7 @@ public String getTranslatedDescription() {
     }
 
     public String getShortName() {
```
magicus, line 43: Are you sure that changing this does not have any unwanted side effects?

Mcrtin: I checked that a while ago and as far as I can see this method (getShortName) is only used by getFeatureByName (or getOverlayByName) and that is only used by commands but someone should definitely verify that.

SpaceMonkeyy86: Yes, the only places this function is used are for /config and /feature and for logging.

magicus (approved):

PR #2082, opened 2023-09-02
fix: fix inventory emerald count being wrong, container not being counted

### LINKED ISSUE #1548: EmeraldCountFeature doesn't correctly keep track on the emeralds in the player's inventory
When slicing the itemStack by right-clicking the counter doesn't update.
It also doesn't count the emeralds on the cursor.
Because the sever only sends ClientboundContainerSetSlotPackets when the prediction of the client is false, a solution has to listen to both ServerboundContainerClickPacket(e.g. using mixins in MultiPlayerGameMode#handleInventoryMouseClick) and ClientboundContainerSetSlotPacket(e.g. using ContainerSetSlotEvent).

kristofbolyai:
Fixes https://github.com/Wynntils/Artemis/issues/1548.

**Review by commit.** Yes, all of these are needed for this to reliably work, Minecraft does have a lot of cases for ItemStack count updates, it's impossible to catch all of them. So we use a supplier.

I also moved to the tick method in emerald model, since we don't do any parsing there anymore (unlike when it was written) it is probably even faster than 90 set slots.

### DIFF
--- a/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
+++ b/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
@@ -151,7 +151,7 @@ public final class EmeraldModel extends Model {
     }
 
     public int getAmountInContainer() {
-        return containerEmeralds;
+        return containerEmeralds - inventoryEmeralds;
     }
 
     public String convertEmeraldPrice(String inputStr) {
--- a/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
+++ b/common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
@@ -8,7 +8,7 @@ import com.wynntils.core.components.Model;
 import com.wynntils.core.components.Models;
 import com.wynntils.mc.event.ContainerCloseEvent;
 import com.wynntils.mc.event.MenuEvent;
-import com.wynntils.mc.event.SetSlotEvent;
+import com.wynntils.mc.event.TickEvent;
 import com.wynntils.models.character.CharacterModel;
 import com.wynntils.models.emeralds.type.EmeraldUnits;
 import com.wynntils.models.items.ItemModel;
@@ -23,7 +23,6 @@ import java.util.Locale;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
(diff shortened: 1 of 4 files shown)


Mcrtin:
Instead of getting the emerald amount from the count, can we not calculate it through the lore? As far as I can remember, the emerald itemStack always displays its value and we can divide that by the EmeraldUnit. What am I missing?

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
```diff
@@ -50,6 +50,93 @@ public EmeraldModel(CharacterModel characterModel, ItemModel itemModel) {
...
+            for (ItemStack item : items) {
+                adjustBalance(null, item, false);
+            }
+        }
+    }
+
+    @SubscribeEvent
+    public void onMenuOpened(MenuEvent.MenuOpenedEvent e) {
+        String title = WynnUtils.normalizeBadString(e.getTitle().getString());
+        if (title.equals("Emerald Pouch")) {
```
ShadowCat117, line 86: Worth making this a constant? Think we do that in most other places

kristofbolyai:
> Instead of getting the emerald amount from the count, can we not calculate it through the lore? As far as I can remember, the emerald itemStack always displays its value and we can divide that by the EmeraldUnit. What am I missing?

Wynn is bugged and doesn't update the lore

common/src/main/java/com/wynntils/models/emeralds/EmeraldModel.java
kristofbolyai: Perhaps but this is a move, not a change.

ShadowCat117:
Inventory emerald count seems to be getting rendered on all containers

kristofbolyai:
Fixed the container count. I think this is a safe way to do it, since the container always has our inventory contents as well.

ShadowCat117 (approved):

common/src/main/java/com/wynntils/models/items/items/game/EmeraldItem.java
```diff
@@ -6,18 +6,19 @@
 
 import com.wynntils.models.emeralds.type.EmeraldUnits;
 import com.wynntils.models.items.properties.EmeraldValuedItemProperty;
+import java.util.function.Supplier;
 
 public class EmeraldItem extends GameItem implements EmeraldValuedItemProperty {
-    private final int amount;
+    private final Supplier<Integer> amountSupplier;
```
magicus, line 12: I am a bit skeptical to this. Up until now, item annotations have always been immutable. Now the value returned by getAmount() can suddenly start varying from call to call on the same EmeraldItem.

I hear what you say that there are a lot of cases where the item count is changed. But then maybe the actual amount should perhaps not be part of the EmeraldItem at all, but have like a EmeraldModel.getAmount(ItemStack item). It is a bit sad if the EmeraldItem cannot describe its actual value, and just the unit, but if we can't guarantee that it is unchanging, then I think that is a better path to take.

magicus: (But best would be to actually catch all count changes)

PR #2083, opened 2023-09-03
fix: CustomColor now properly handles hex string conversion (to/from), add unit tests

kristofbolyai:
Quite unfortunate since we always used hex color configs. However, it seems that only toHex was broken, so some values just couldn't be saved. This means we technically don't corrupt/change any colors. fromHexString also got alpha support now, for full support.


kristofbolyai:
I tried this on my config, produces 0 diffs. However, I did not set any colors except some POIs...

ryanzhoudev (approved):

PR #2084, opened 2023-09-03
fix: CME on laggy connections when kicking offline

ryanzhoudev:


kristofbolyai (approved):
Bit weird but I guess fine

common/src/main/java/com/wynntils/models/players/PartyModel.java
```diff
@@ -467,7 +467,8 @@ public void partyJoin(String playerName) {
      * Consider running {@link #requestData()} before this.
      */
     public void partyKickOffline() {
-        offlineMembers.forEach(this::partyKick);
```
magicus, line 470: Sorry for being late to the party, but maybe a comment would be nice, saying this is to avoid CME..? Otherwise it will look a bit odd to a future reader.

PR #2086, opened 2023-09-04
feat: add stat filters

kristofbolyai:
This should mean we have a full set of filters now :)


ShadowCat117:
They don't seem to work at all in the guides?

kristofbolyai:
> They don't seem to work at all in the guides?

Hm that might be the case..

kristofbolyai:
Fixed.

ShadowCat117:
Still doesn't work for ingredients, only item guide.

Looks good other than that though

kristofbolyai:
> Still doesn't work for ingredients, only item guide.
> 
> 
> 
> Looks good other than that though

Yeah, intentional atm

ShadowCat117:
> Yeah, intentional atm

Ok, is that going to be added in a follow up pr? If so I'll approve this

kristofbolyai:
> > Yeah, intentional atm
> 
> 
> 
> Ok, is that going to be added in a follow up pr? If so I'll approve this

Yeah perhaps, i'll check it out tomorrow

ShadowCat117 (approved):

ryanzhoudev (approved):

PR #2088, opened 2023-09-06
feat: make stat filters work on ingredients

kristofbolyai:


ShadowCat117 (approved):
Looks fine to me.

Something I did notice whilst testing this is we don't have support for negative inputs, so you can't search for healthBonus:-100 for example

kristofbolyai:
Hm. Okay. I will fix the negative inputs soon.

PR #2092, opened 2023-09-07
fix: Fix descriptions of text-based settings not displaying

### LINKED ISSUE #2091: The discription of the ConfigTile doesn't get displayed when hovering a text field
https://gyazo.com/5295c4083ec5706997eb949674507ae8
https://discord.com/channels/394189072635133952/424990456435310602/1149396907060375572

Mcrtin:
fixes: #2091

### DIFF
--- a/common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
+++ b/common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
@@ -8,7 +8,11 @@ import com.wynntils.core.persisted.config.Config;
 import com.wynntils.screens.base.TextboxScreen;
 import com.wynntils.screens.base.widgets.TextInputBoxWidget;
 import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.mc.ComponentUtils;
 import com.wynntils.utils.render.FontRenderer;
+import net.minecraft.network.chat.Component;
+
+import java.util.List;
 
 public class TextInputBoxSettingsWidget<T> extends TextInputBoxWidget {
     protected final Config<T> config;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/screens/settings/widgets/TextInputBoxSettingsWidget.java
```diff
@@ -17,6 +20,7 @@ protected TextInputBoxSettingsWidget(Config<T> config, TextboxScreen textboxScre
         super(0, 6, width, FontRenderer.getInstance().getFont().lineHeight + 8, null, textboxScreen);
         this.config = config;
         setTextBoxInput(config.get().toString());
+        tooltip = ComponentUtils.wrapTooltips(List.of(Component.literal(config.getDescription())), 150);
```
kristofbolyai, line 23: either this is not needed or this fixes something that you didnt describe..

ShadowCat117: This sets the tooltip to be displayed, no?

It might be better to have the tooltip as a parameter in the parent constructor, then it's easier to set tooltips for any other input widgets

kristofbolyai: The other change sets the tooltip to be rendered? this just sets some variable

kristofbolyai: i would have to read the code, this is kind of weird

ShadowCat117: Yeah this sets what the tooltip will be, in `TextInputBoxWidget` the variable is just null, it's only set here, that's why I think it would be better to do in the constructor

Mcrtin: the tooltip variable refers to the field I added in `TextInputBoxWidget`. I did it like this to avoid another constructor parameter (because it is just used once) and to allow it to be changed.

kristofbolyai (approved):
Hm, I see now. I didn't realize the inheritance between these classes..

PR #2097, opened 2023-09-11
fix: fix filters not accepting negative numbers

kristofbolyai:


common/src/main/java/com/wynntils/services/itemfilter/filters/RangedStatFilters.java
```diff
@@ -91,26 +92,36 @@ protected RangedStatValueStatFilter getRangedStatFilter(int min, int max) {
...
-            } else if (RANGE_PATTERN.matcher(inputString).matches()) {
-                String[] split = inputString.split("-");
-                int min = Integer.parseInt(split[0]);
-                int max = Integer.parseInt(split[1]);
+            }
+
+            matcher = RANGE_PATTERN.matcher(inputString);
+            if (matcher.matches()) {
+                int min = Integer.parseInt(matcher.group(1));
+                int max = Integer.parseInt(matcher.group(2));
```
ryanzhoudev, line 112: Yea this is much nicer, dunno why we used a split before

ryanzhoudev (approved):

common/src/main/java/com/wynntils/services/itemfilter/filters/RangedStatFilters.java
kristofbolyai: Yeah, my bad.

PR #2101, opened 2023-09-15
fix: Arrow Shield Model NPE when arrows not spawned

ryanzhoudev:


common/src/main/java/com/wynntils/models/abilities/ArrowShieldModel.java
```diff
@@ -117,9 +117,11 @@ public int getArrowShieldCharge() {
     }
 
     private void registerShield() {
-        spawnedArrowIds = collectedArrowIds;
-        collectedArrowIds = null;
-        WynntilsMod.postEvent(new ArrowShieldEvent.Created(spawnedArrowIds.size()));
+        if (collectedArrowIds != null) {
```
kristofbolyai, line 120: Is this the root cause? Is there not a race somewhere? I saw the exception log too, saw tick scheduler. Maybe we have a wrong assumption somewhere?

ryanzhoudev: It is possible that the cast is triggered multiple times within 400 milliseconds causing the collected list to become null then cascading to spawnedArrowIds, but this should catch that. There's no other instance of collectedArrowIds becoming null

kristofbolyai (approved):

PR #2104, opened 2023-09-18
fix: Action bar does not match 100% curse

### LINKED ISSUE #2100: Mana Bar and Powder Special Bar is not working properly when Curse is at 100%
Steps to replicate:
1. Get and use a weapon with Curse powder special
2. Attack mobs (or housing combat dummy) with main attack until Curse reaches 100%
3. Cast any spell 

Expected behaviour:
- Mana Bar will reflect on mana consumed.
- Powder Special Bar can show 100%

Actual behaviour:
image

ryanzhoudev:
Fixes #2100.

So the previous method of running a regex to split the actionbar into 3 parts using the color codes does not work with curse. Since StyledText will eliminate the duplicate blue code (usually there for 100% curse, and another one for mana), the 100% curse will be pushed to the mana matcher. In my solution here, we just split according to four consecutive spaces, which Wynn seems to always send to separate parts. Related regex fixes to accommodate this, as well as tests, are also included.


common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -54,16 +52,11 @@ public void onActionBarUpdate(ChatPacketReceivedEvent.GameInfo event) {
         }
         previousRawContent = content;
 
-        Matcher matcher = content.getMatcher(ACTIONBAR_PATTERN);
-        if (!matcher.matches()) {
-            WynntilsMod.warn("ActionBarHandler pattern failed to match: " + content);
-            return;
-        }
+        StyledText[] segments = content.split("    "); // Wynn will provide at least 4 spaces between each segment
```
magicus, line 64: Should we not do some kind of sanity check that we get exactly three segments? If we do not, we should either print a warning (and pad with empty segments if less than three, or drop/merge segments after the third), or just give up and return.

common/src/main/java/com/wynntils/models/characterstats/actionbar/ManaSegment.java
```diff
@@ -11,7 +11,8 @@
 import java.util.regex.Pattern;
 
 public class ManaSegment implements ActionBarSegment {
-    private static final Pattern MANA_PATTERN = Pattern.compile("§b✺ ([0-9]+)/([0-9]+)");
+    // https://regexr.com/7k9ab
+    private static final Pattern MANA_PATTERN = Pattern.compile("(?:§.  §.|§b {0,2})✺ (\\d+)/(\\d+)");
```
magicus: I'm not sure what I am seeing here. Why do we get these prefixed spaces? Are we correct in pushing them on to the segment, or should we have removed them in the handler?

ryanzhoudev: This is just the raw result from styledtext.split() on the entire actionbar

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
ryanzhoudev: We could, but I did a bunch of content and did not get any instance where actionbar crashed or had something other than 3 segments. I'll do it when I get home today, or tomorrow if I don't get time

kristofbolyai: > at least 4 spaces

What if they send 8? we offset our math and cause confusion. Oh.. I just read magicus' comment. Yes, I wonder the same thing.

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -158,4 +160,9 @@ public List<GearInfo> getWornGear() {
     private void centerSegmentCleared() {
         powderSpecialSegment.replaced();
     }
+
+    @SubscribeEvent
+    public void onHeldItemChanged(ChangeCarriedItemEvent event) {
+        powderSpecialSegment.replaced();
```
kristofbolyai, line 167: How is this a special case? I assume this is related to item handler's rename canceling?

ryanzhoudev: Currently the powder segment resets the charge data to 0 if we get the coordinates or something like that, but it is possible to not get the coordinates when switching weapons. But switching weapons always resets to 0. So this accounts for it.

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
ryanzhoudev: Oh good point, I'll see if I can find an example

common/src/main/java/com/wynntils/models/characterstats/actionbar/ManaSegment.java
magicus: So the action bar looks something like: `§x<left>    §y<center>    §z   §w<right>`?

I also find the `|` without parenthesis in regexps confusing, since I think it binds strangely. Does not ` §.|§b ` mean "space, followed by either § and any character, or "§b" and then space"? If so, this could just as well be ` §. `. (And why could we have anything but the `b`?)

magicus: I might be missing something here, but I think we might need to do some more processing "centrally" before sending out the segments to the segment handlers, and just not blindly split it. 

The other segment matchers looked okay, but this one does not.

magicus: Or does `(?:§.  §.|§b {0,2})` mean either match `§.  §.` or `§b {0,2})`? If so, why is not the first part `§.  §b`? This does not seem to rhyme with the example string you posted in ActionBarHandler...

ryanzhoudev: I think this will be easier to discuss in Discord, I'll @ you

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
magicus: Please add a short comment explaining this.

(In general, when someone asks a question in a review, it means the code was not clear to an (informed) reader. Hence, the reply should be stored with the source code, if future (informed) readers have the same question, not just in the PR.)

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -59,11 +59,15 @@ public void onActionBarUpdate(ChatPacketReceivedEvent.GameInfo event) {
             WynntilsMod.warn("ActionBarHandler pattern failed to match: " + content);
             return;
         }
+        StyledText[] segments = {
```
magicus: At this point, the patch has really boiled down to just a change in the ACTIONBAR_PATTERN, and this refactoring is not needed (nor helpful imho).

common/src/main/java/com/wynntils/models/characterstats/actionbar/SprintSegment.java
```diff
@@ -13,7 +13,8 @@
 import java.util.regex.Pattern;
 
 public class SprintSegment implements ActionBarSegment {
-    private static final Pattern SPRINT_PATTERN = Pattern.compile("§0 *§[246]\\[(§.*)§[246]\\] *");
+    // https://regexr.com/7k9an
+    private static final Pattern SPRINT_PATTERN = Pattern.compile("§[246]\\[(§.*)§[246]] *");
```
magicus:
```suggestion
    private static final Pattern SPRINT_PATTERN = Pattern.compile("§[246]\\[(§.*)§[246]]");
```

maybe..?

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -20,8 +20,8 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class ActionBarHandler extends Handler {
-    // example: "§c❤ 218/218§0    §7502§f S§7 -1580    §b✺ 1/119"
-    private static final Pattern ACTIONBAR_PATTERN = Pattern.compile("(?<LEFT>§[^§]+)(?<CENTER>.*)(?<RIGHT>§[^§]+)");
+    // https://regexr.com/7kfrm
+    private static final Pattern ACTIONBAR_PATTERN = Pattern.compile("(.+?) {4,}(.+?) {4,}(.+)");
```
magicus: What does `.+?` mean? It looks like invalid regexp syntax to me. I assume you mean just `.+`, that is, one or more of any character. If you mean zero or more, use `.*`.

I'm also thinking that someone cared enough to annotate the groups with names in the old pattern. (Maybe it was me, but I don't recall doing it). Even if it is not used, someone felt it was helpful, so maybe keep that?

kristofbolyai: ? can be added as an additional modifier to */+. It will match the **least** amount or characters possible.

magicus: Another thing that always makes me think twice in regexps like this is how the greediness works. I usually get it wrong, so I prefer coding safely. If you have this regexp `(.+) {4,}(.+)` and then you match the string "abc     def", where there is five spaces, will it match as (abc)(     )(def) with five spaces initially or (abc)(    )( abc) with four spaces initially? The answer depends on how you set the greediness behavior of the regexp engine.

To avoid this, you can instead code it like `(.+) {4,}([^ ].*)`, thus forcing the last part to never under any circumstances start with a space. While the regexp gets longer, I tend to prefer this kind of writing, since it eliminates any question about how stuff will match in "ambiguous" inputs.

(In your case, the corresponding change would also need to be done for the "end" of the middle segment)

ryanzhoudev: `.+?` is lazy matching, if we don't use lazy, the first matcher tries to steal spaces and only leave the ` {4,}` four spaces. Same goes for the middle one. Without it, the cg's sometimes ends up with extra trailing spaces. Last one does not need to be lazy since there are no spaces after the right segment

magicus: Ah, I see. Well, this solves the problem I was worried about. Turns out you can still learn new stuff about regexps after 25 years. ;-)

magicus (approved):

magicus:
This looks much better. If you want, you can reinstantiate the use of lazy matching; my suggestions was based upon my limited understanding of regexps.

The final segment patterns looks much clearer than before, nice improvement!

PR #2105, opened 2023-09-18
fix: Chat message filters for 2.0.3

### LINKED ISSUE #2103: Chat tab message types need to be updated to 2.0.3 format
Bunch of regexes in com.wynntils.handlers.chat.type.RecipientType.java

ryanzhoudev:
Fixes #2103.

It is way too much work to manually go through the private use section and do a full ^$ regex, so I just did ^ ending with .+
Fairly comprehensive tests exist if anyone wants to try though. The local messages is missing a case with a guild milestone, just keep that in mind.


magicus (approved):
This is utterly impossible to see with the naked eye if it is correct. I trust your diligence and test suites.

PR #2107, opened 2023-09-19
refactor: Use translatable for redirect messages

### LINKED ISSUE #626: Add I18n for InfoMessageFilterFeature redirects

ryanzhoudev:
Resolves #626. Please review somewhat carefully. I looked it over and didn't catch anything, but this was really mind numbing work so I didn't pay too much attention.


magicus (approved):
I glanced through this, and it looks sane.

But as you say, maybe we need to have a more thorough review that tries to check that the messages are actually moved correctly. Or, we'll just hope they are, and let players report bugs if they are weird. :)

I wonder if we can generalize this a bit, so we only have the part following `chatRedirect` of the key returned, and then also ideally check if it actually exists with an assert. Hm.

common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
```diff
@@ -494,13 +514,19 @@ public RedirectAction getAction() {
...
-            String emeraldString = matcher.group(2);
+            String emeraldString = ChatFormatting.GREEN + matcher.group(2) + ChatFormatting.LIGHT_PURPLE;
 
-            return StyledText.fromString(
-                    String.format("§dSold §7%s §dfor §a%s§d.", ingredientString.getString(), emeraldString));
+            return StyledText.fromComponent(Component.translatable(
+                            "feature.wynntils.chatRedirect.ingredientPouch.notification",
+                            formattedPlural.getString(),
+                            emeraldString)
+                    .withStyle(ChatFormatting.LIGHT_PURPLE));
```
kristofbolyai: I am not sure how this interacts. You set a color but also append char colors. Can you test this single line? Also perhaps, move the emeraldString into a styledText, which should resolve this doubt.

ryanzhoudev: image
Looks fine

kristofbolyai (approved):

PR #2109, opened 2023-09-19
fix: Fix emerald pouch's emerald displays overlapping

### LINKED ISSUE #2087: Opening Emerald Pouch gives Two Emerald Inventory Overlays
On Artemis v0.0.3-alpha.367

To reproduce, simply right-click on the emerald pouch in your inventory. The following should appear:

image

Expected behavior is to only have one overlay show up at a time, rather than have both conflict and look bad.

Mcrtin:
Removes the upper emerald display in the emerald pouch as proposed in #2087
image
Closes #2087


kristofbolyai (changes requested):
Not the way to fix this.. the container still has emeralds technically. Either just hide it in rendering or find a way to make it work. Text based emerald count would not work this way..

Mcrtin:
> Not the way to fix this.. the container still has emeralds technically. Either just hide it in rendering or find a way to make it work. Text based emerald count would not work this way..

I can move the upper display up until they no longer overlap. Is that a solution you'd find appropriate?

ryanzhoudev:
What if we changed how these rendered entirely? When you have a container open, start rendering from the middle. So the stacks counter is adjacent to each other. The one for the inventory would grow downwards, going from stx>le>blocks>emeralds, and the one for the open container would grow upwards. This way we will never have overlapping counters. Thoughts?

Mcrtin:
> What if we changed how these rendered entirely? When you have a container open, start rendering from the middle. So the stacks counter is adjacent to each other. The one for the inventory would grow downwards, going from stx>le>blocks>emeralds, and the one for the open container would grow upwards. This way we will never have overlapping counters. Thoughts?

Sounds good; I've also thought about that, but it's outside the scope of this pr.

Mcrtin:
image

magicus:
The screenshot looks a bit redundant. Is the idea that if you have additional emeralds outside the pouch, that will show up in the lower half (showing the inventory), but not the upper half (showing just what is inside the pouch)?

kristofbolyai (approved):
LGTM

Mcrtin:
> The screenshot looks a bit redundant. Is the idea that if you have additional emeralds outside the pouch, that will show up in the lower half (showing the inventory), but not the upper half (showing just what is inside the pouch)?

like in every other screen, the upper display shows the emeralds in the container and the lower one shows the emeralds in the container and your pouches. In this case (which is I think the most common one for pouches), all of my emeralds are in the only pouch I have in my inv; so it basically shows the count twice. So yeah, for me its a bit redundent.

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -89,8 +89,16 @@ public void onContainerRender(ContainerRenderEvent event) {
             int y = containerScreen.topPos;
             switch (emeraldCountType.get()) {
                 case TEXT -> renderTextCount(event.getPoseStack(), textX, y, topEmeralds);
-                case TEXTURE -> renderTexturedCount(
-                        event.getPoseStack(), textureX, y, topEmeralds, showZerosInEmeraldCount.get());
+                case TEXTURE -> {
+                    int displayedTextures = (int) (Arrays.stream(getEmeraldAmounts(topEmeralds))
```
magicus: Is this calculating the height of the textures that renderTexturedCount() is rendering? It feels a bit complex-y, and it is not immediately clear why we should calculate it like this. I get the feeling this is based on some other code elsewhere, but there is no connection to be found.

Mcrtin: I've moved the conditions into the `getEmeraldAmounts` method (and renamed it). It should be clearer now where the conditions come from.

magicus (approved):
Thanks. This looks much easier to follow.

PR #2114, opened 2023-09-21
fix: Login redirect not redirecting renamed players

ShadowCat117:
Should work but I don't have champion and an alt to test so if someone could verify that'd be great


ryanzhoudev (approved):
I tested with my alt, works fine. Though, in your regex link, can you make your new test match full?

ShadowCat117:
Done, and thanks for checking it works :)

PR #2120, opened 2023-09-24
fix: Chat Coordinates improvements

ShadowCat117:
Looks like #2075 broke a lot of scenarios when the coordinates should be detected

Coords

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
@@ -60,7 +60,7 @@ public class ChatCoordinatesFeature extends Feature {
                     LocationUtils.strictCoordinateMatcher(partToReplace.getString(null, PartStyle.StyleType.NONE));
 
             while (matcher.find()) {
-                Optional<Location> location = LocationUtils.parseFromString(matcher.group());
+                Optional<Location> location = LocationUtils.parseFromString(matcher.group(1));
 
                 if (location.isEmpty()) {
                     continue;
@@ -68,12 +68,8 @@ public class ChatCoordinatesFeature extends Feature {
 
                 String match = partToReplace.getString(null, PartStyle.StyleType.NONE);
 
-                int matchStart = matcher.group().charAt(0) == ' ' ? matcher.start() + 1 : matcher.start();
-                int matchEnd =
-                        matcher.group().charAt(matcher.group().length() - 1) == ' ' ? matcher.end() - 1 : matcher.end();
-
-                String firstPart = match.substring(0, matchStart);
-                String lastPart = match.substring(matchEnd);
+                String firstPart = match.substring(0, matcher.start(1));
+                String lastPart = match.substring(matcher.end(1));
 
                 PartStyle partStyle = partToReplace.getPartStyle();
 
--- a/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
@@ -68,8 +68,12 @@ public class ChatCoordinatesFeature extends Feature {
 
                 String match = partToReplace.getString(null, PartStyle.StyleType.NONE);
 
-                String firstPart = match.substring(0, matcher.start());
-                String lastPart = match.substring(matcher.end());
+                int matchStart = matcher.group().charAt(0) == ' ' ? matcher.start() + 1 : matcher.start();
+                int matchEnd =
+                        matcher.group().charAt(matcher.group().length() - 1) == ' ' ? matcher.end() - 1 : matcher.end();
+
+                String firstPart = match.substring(0, matchStart);
+                String lastPart = match.substring(matchEnd);
 
                 PartStyle partStyle = partToReplace.getPartStyle();
 
--- a/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
@@ -71,6 +71,11 @@ public class ChatCoordinatesFeature extends Feature {
                 String firstPart = match.substring(0, matcher.start(1));
                 String lastPart = match.substring(matcher.end(1));
 
+                if (firstPart.endsWith("[") && lastPart.startsWith("]")) {
+                    firstPart = firstPart.substring(0, firstPart.length() - 1);
+                    lastPart = lastPart.substring(1);
+                }
+
                 PartStyle partStyle = partToReplace.getPartStyle();
 
                 StyledTextPart first = new StyledTextPart(firstPart, partStyle.getStyle(), null, Style.EMPTY);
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
```diff
@@ -68,8 +68,12 @@ private static StyledText getStyledTextWithCoordinatesInserted(StyledText styled
...
                 String match = partToReplace.getString(null, PartStyle.StyleType.NONE);
 
-                String firstPart = match.substring(0, matcher.start());
-                String lastPart = match.substring(matcher.end());
+                int matchStart = matcher.group().charAt(0) == ' ' ? matcher.start() + 1 : matcher.start();
+                int matchEnd =
+                        matcher.group().charAt(matcher.group().length() - 1) == ' ' ? matcher.end() - 1 : matcher.end();
+
+                String firstPart = match.substring(0, matchStart);
+                String lastPart = match.substring(matchEnd);
```
kristofbolyai: I am not sure I understand what this does with relation to the regex change. Can you explain?

ShadowCat117: In the "My location is at x, y, z" for example, the matcher matches " x, y, z " so if you just use matcher.start() it'll appear as "My location is at[x, y, z] without a space between the "at" and [""

kristofbolyai: Hm. It's just that I really don't like using string manipulations, it is very easy to get them to work, but it is also very easy to not cover an edge case, and cause crashes/other issues. Do you think we could fix this in the regex itself?

ShadowCat117: Possibly, I did start with trying to fix the issue there but didn't have any luck though I'm by no means an expert in regex

kristofbolyai: Yeah, I see. I am not sure how I can help, but I would rather get the regex straight.

ShadowCat117: I'll give it another try, don't want to pull you away from the 1.20 port

ShadowCat117: Think I've got it working the same just with the regex and using group 1 instead. Wouldn't mind someone else giving it a test to make sure it's not parsing unexpected things but no rush on that, it's not a major issue

common/src/main/java/com/wynntils/utils/wynn/LocationUtils.java
```diff
@@ -12,10 +12,10 @@
 
 public final class LocationUtils {
     private static final Pattern COORDINATE_PATTERN = Pattern.compile(
-            "(?<x>[-+]?\\d{1,6})(?:\\.\\d+)?([^0-9.+-]{1,5}(?<y>[-+]?\\d{1,3})(?:\\.\\d+)?)?[^0-9.+-]{1,5}(?<z>[-+]?\\d{1,6})(?:\\.\\d+)?");
+            "\\s*(?<x>[-+]?\\d{1,6})(?:\\.\\d+)?([^0-9.+-]{1,5}(?<y>[-+]?\\d{1,3})(?:\\.\\d+)?)?[^0-9.+-]{1,5}(?<z>[-+]?\\d{1,6})(?:\\.\\d+)?\\s*");
 
-    private static final Pattern STRICT_COORDINATE_PATTERN = Pattern.compile(
-            "(?:^|\\s)([-+]?\\d{1,6})(?:\\.\\d+)?((,\\s?|\\s)([-+]?\\d{1,3})(?:\\.\\d+)?)?(,\\s?|\\s)([-+]?\\d{1,6})(?:\\.\\d+)?(?:\\s|$)");
+    private static final Pattern STRICT_COORDINATE_PATTERN =
+            Pattern.compile("(?:^|\\s)([-+]?\\d{1,6}(?:[\\s,]+[-+]?\\d{1,3}(?:[\\s,]+[-+]?\\d{1,6})?)?)(?:\\s+|$)");
```
kristofbolyai: Did you update all users of this matcher so they use the new groups?

kristofbolyai: Okay, I checked. Only 1 use case.

kristofbolyai: Perhaps limit the `[\\s,]` groups to a certain max count? Like 0-2 chars (inclusive)? This would match `100       ,    10 10` at the moment.

ShadowCat117: I think Minecraft auto removes more than 1 whitespace, so theoretically you should never be able to send that

kristofbolyai: Same with 1000,,,,,100 then

common/src/main/java/com/wynntils/utils/wynn/LocationUtils.java
```diff
@@ -37,15 +37,15 @@ public static Matcher strictCoordinateMatcher(String str) {
     }
 
     public static void shareLocation(String target) {
-        String locationString = "My location is at [" + (int) McUtils.player().position().x + ", "
```
magicus: Why these changes removing `[]`? Coordinates in Wynncraft are presented in brackets, so I think we should do the same.

kristofbolyai: The coord feature already adds then, so this was just a "double" bracket

magicus: Aha!

magicus: ... uh, but wait. 

This is added only if the recipient is also running Wynntils.

The proper fix should be that the coord highligher feature should only add `[...]` if there is not one already.

If you share your coordinates to someone not running Wynntils, they should see the brackets.

kristofbolyai: I mean, I am not sure brackets are essential if the coordinate is not clickable

magicus: It's back to what I said before: "Coordinates in Wynncraft are presented in brackets, so I think we should do the same." Consistency for the user experience. This was the reason I implemented the message this way long time ago.

ShadowCat117: Hmm, I do agree that it's nice to keep consistent with Wynn where we can, but I'd argue it's not necessary here as this is just a convenient way for people to send their location, it's not a server message.

When non-Wynntils users see players sending coords in chat normally they'll just see "go to x,y,z" for example, no brackets so I don't think we need a special case for this. Unless we start adding brackets everytime a user sends coords

magicus: Ok, let me turn this around the other way. I still think we should not add `[...]` around coordinates if there already are brackets. Let's say Wynncraft sends:

"Go back to [123,65,-789]"

Then we should not add additional brackets.

And if we do add this functionality, we can just as well keep the brackets here, to clearly mark the coordinates. I think of it as being a "good citizen", and try to always present coordinates in the same style.

ShadowCat117: Ok, I've just added them back and it'll remove the brackets if it finds coords with them at the start and end

magicus:
I think this looks good, but could you please post a screenshot of how the original test messages look with this version of the code?

ShadowCat117:
image

magicus:
Looks super!

magicus (approved):

PR #2124, opened 2023-09-25
fix: Activity tracker overlay missing spaces

ryanzhoudev:
So I tried this back in #2012, but we did not really have a conclusion other than that method was also problematic. I am going to try again.

### DIFF
--- a/common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
+++ b/common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
@@ -18,7 +18,7 @@ import net.minecraft.ChatFormatting;
 
 public class ActivityTrackerScoreboardPart extends ScoreboardPart {
     private static final SegmentMatcher TRACKER_MATCHER = SegmentMatcher.fromPattern("Tracked (.*):");
-    private static final Pattern SPACER_PATTERN = Pattern.compile("[^a-zA-Z\\[\\-\\d].*");
+    private static final Pattern SPACER_PATTERN = Pattern.compile("[^a-zA-Z\\[\\d-].*");
 
     @Override
     public SegmentMatcher getSegmentMatcher() {
--- a/common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
+++ b/common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
@@ -42,9 +42,18 @@ public class ActivityTrackerScoreboardPart extends ScoreboardPart {
         List<StyledText> taskLines = content.subList(1, content.size());
 
         for (StyledText line : taskLines) {
+            // If a comma is at the start of a line, we don't need the space from the previous line
+            if (line.getString(PartStyle.StyleType.NONE).startsWith(",")) {
+                nextTask.deleteCharAt(nextTask.length() - 1);
+            }
+            System.out.println(line.getString());
             nextTask.append(line.getString()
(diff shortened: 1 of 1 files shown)


ryanzhoudev:
This appears to work for all of the quest lines I tried. But I don't have every single stage for every single quest, so there might be one-offs. This also works with coordinates; the waypoint is still placed and spaces are added correctly.

It works by adding a space if the existing line does not end with a space. Then, on the next line, if it starts with a special character (we use a negated regex set), then it will delete the previously added space. This is to account for the fact that Wynn sometimes sends next lines with and without spaces, this ensures consistency during processing.

kristofbolyai (commented):
Hm. The code looks alright, I am just not sure how to explain this logic in the code..

common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
```diff
@@ -13,10 +13,12 @@
 import com.wynntils.handlers.scoreboard.type.SegmentMatcher;
 import java.util.List;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import net.minecraft.ChatFormatting;
 
 public class ActivityTrackerScoreboardPart extends ScoreboardPart {
     private static final SegmentMatcher TRACKER_MATCHER = SegmentMatcher.fromPattern("Tracked (.*):");
+    private static final Pattern SPACER_PATTERN = Pattern.compile("[^a-zA-Z\\[\\-\\d].*");
```
magicus: The traditional way of writing a character group with the literal `-` is to put it last. This way it can't mean a range, so regexp syntax stipulates it must be treated as a literal, even if it is not escaped.

Furthermore, I believe you do not need to escape `[`, since it cannot be mistaken for a character group start. (If you think it improves readability you might do so nevertheless.)

magicus:
Have  you verified that this does not add spaces in the middle of words, as before? (To me it seems this will fail just as much as the previous attempts). One of the "standard" NPCs in Detlas will by default say something with a split word, don't remember if it was The Apprentice or Yahya, one of I think.

common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
ryanzhoudev: We unfortunately do need to escape `[`, it says "Unclosed character class" when you don't. I will rearrange so that `-` is last

ryanzhoudev:
I checked both the potion quest and the Yahya quest. Neither have issues.

common/src/main/java/com/wynntils/models/activities/ActivityTrackerScoreboardPart.java
```diff
@@ -42,9 +44,20 @@ public void onSegmentChange(ScoreboardSegment newValue) {
         List<StyledText> taskLines = content.subList(1, content.size());
 
         for (StyledText line : taskLines) {
+            String unformatted = line.getString(PartStyle.StyleType.NONE);
+            Matcher spacerMatcher = SPACER_PATTERN.matcher(unformatted);
+            if (spacerMatcher.matches()) {
+                // There is a special character at the start of the line, we don't need the previous space
+                nextTask.deleteCharAt(nextTask.length() - 1);
```
magicus: If the very first character is a special character, it will try to delete past the start of the string, If `deleteCharAt` just ignores a -1 argument then all is fine, otherwise this will be bad. (If sending -1 is okay, please add a comment stating that it is okay)

magicus (approved):

magicus:
Let's hope this works better. If someone complains, we need to take a hard good look at the actual message that fails.

PR #2125, opened 2023-09-25
fix: Fix a batch of 11 bugs from the active Athena reports

kristofbolyai:
I looked through around 300 reports, these are the relevant ones. I believe this should make the mod close to "perfect", in terms of crashes, at the moment.

### DIFF
--- a/common/src/main/java/com/wynntils/models/marker/UserWaypointMarkerProvider.java
+++ b/common/src/main/java/com/wynntils/models/marker/UserWaypointMarkerProvider.java
@@ -15,12 +15,12 @@ import com.wynntils.utils.mc.type.Location;
 import com.wynntils.utils.mc.type.PoiLocation;
 import com.wynntils.utils.render.Texture;
 import com.wynntils.utils.type.Pair;
-import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Stream;
 
 public class UserWaypointMarkerProvider implements MarkerProvider<WaypointPoi> {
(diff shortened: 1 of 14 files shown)


common/src/main/java/com/wynntils/models/marker/UserWaypointMarkerProvider.java
```diff
@@ -15,12 +15,12 @@
...
 import com.wynntils.utils.render.Texture;
 import com.wynntils.utils.type.Pair;
-import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.stream.Stream;
 
 public class UserWaypointMarkerProvider implements MarkerProvider<WaypointPoi> {
-    private final Set<Pair<MarkerInfo, WaypointPoi>> markerInfoSet = new LinkedHashSet<>();
+    private final Set<Pair<MarkerInfo, WaypointPoi>> markerInfoSet = new CopyOnWriteArraySet<>();
```
magicus, line 23: Can you expand a bit about why this was needed? Preferably as a short comment in the code.

magicus (approved):
Great work! All fixes look good as far as I can tell.

common/src/main/java/com/wynntils/models/marker/UserWaypointMarkerProvider.java
kristofbolyai: Well, the marker provider implementations need to be race-proof.

kristofbolyai: https://github.com/Wynntils/Artemis/pull/2125/commits/0488ef8de25077db8aa29a79b49be1990d715caf

PR #2131, opened 2023-09-26
fix: Emerald count has wrong units when display zero is off

### LINKED ISSUE #2121: Emerald Count icons being wrong in other containers
In some inventory screens, the emerald count is correct but shows the wrong form of currency. shows proper currency while holding shift.

With Shift Loot run end chest
image
Without Shift
Loot run end chest
image
Item Identifier (this one should show 46le and 5eb)
image

ryanzhoudev:
Fixes #2121. Essentially, the logic checking if we should be displaying the zeros has been moved to the render method, and the emerald String[] method always returns an array of four now.


common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -96,7 +92,10 @@ public void onContainerRender(ContainerRenderEvent event) {
                 case TEXTURE -> {
                     int adjustedY;
                     if (displayBottom) {
-                        int displayedTextures = getFilteredEmeraldAmounts(topEmeralds).length;
+                        int displayedTextures = Arrays.stream(getRenderableEmeraldAmounts(topEmeralds))
+                                .filter(s -> !s.equals("0"))
+                                .toArray()
+                                .length;
```
kristofbolyai: Arguably this is weirder than the old way. Checking for equals "0" here and below feels odd.

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -152,13 +151,15 @@ private void renderTexturedCount(PoseStack poseStack, int x, int y, int emeralds
...
 
-        String[] emeraldAmounts = getFilteredEmeraldAmounts(emeralds);
+        String[] emeraldAmounts = getRenderableEmeraldAmounts(emeralds);
 
         int renderedCount = 0;
 
         for (int i = emeraldAmounts.length - 1; i >= 0; i--) {
             String emeraldAmount = emeraldAmounts[i];
 
+            if (!showZerosInEmeraldCount.get() && emeraldAmount.equals("0")) continue;
```
kristofbolyai, line 160: Do I understand correctly that this is the only actual line that needs changing? And only the left hand side of the AND, if we don't do anything else?

(the `isExcludedContainer` refactor is okay though)

ryanzhoudev: I am not really sure what you mean. The previous way worked by just dumping a bunch of strings into an array, and (wrongly) assuming that the render method would somehow automatically figure out what amount corresponded to what unit. We can either pad the array with zeros and filter in the render (what I did here) or we can change the array into a enum map and define the textures somewhere, and just have the getRenderableEmeraldAmounts method do the filtering again.

kristofbolyai: Hm, okay. I guess this looks good, perhaps we can refactor this rendering later on.

Mcrtin:
This bug originated from #2109 in the last commit.
You basically just have to implement a better/working solution for the readability problem.

kristofbolyai (approved):
If this fix works, let it be. However, I would not mind to revisit this feature later and clean it up. Not sure is it can be much better, but we will see.

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -96,7 +92,10 @@ public void onContainerRender(ContainerRenderEvent event) {
                 case TEXTURE -> {
                     int adjustedY;
                     if (displayBottom) {
-                        int displayedTextures = getFilteredEmeraldAmounts(topEmeralds).length;
+                        int displayedTextures = Arrays.stream(getRenderableEmeraldAmounts(topEmeralds))
+                                .filter(s -> !s.equals("0"))
```
Mcrtin: what about `showZerosInEmeraldCount.get()`?

ryanzhoudev: Didn't check because I only looked at the container in the issue, fixed now

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -94,16 +90,16 @@ public void onContainerRender(ContainerRenderEvent event) {
...
-                        int sizeY = displayedTextures * TEXTURE_SIZE + 2;
-                        int bottomY = containerScreen.imageHeight - TEXTURE_SIZE * 3 - 2;
-                        adjustedY = y - Math.max(sizeY - bottomY, 0);
-                    } else {
-                        adjustedY = y;
+                    if (displayBottom) { // ensure we don't overlap with bottom textures
+                        int topDisplayedTextureCount = Arrays.stream(getRenderableEmeraldAmounts(topEmeralds))
+                                .filter(s -> showZerosInEmeraldCount.get() || !s.equals("0"))
+                                .toArray()
+                                .length;
```
Mcrtin: nit: i believe there is a count method on stream; looks good otherwise

ryanzhoudev: I don't think it matters that much, count returns a long anyway so you'll have to cast

kristofbolyai: allocation wise might be a bit better, depending on how it is implemented in java itself

ryanzhoudev: mapToLong() then sum()

ryanzhoudev: Actually we can just mapToInt then sum. I will do that.

magicus: Why not use the `.count()` and cast? This sum business just look funky.

ryanzhoudev: It is the same internally, just uses mapToLong instead

magicus: What do you mean? When I check the JDK source code, the implementation for `count` I find is `evaluate(ReduceOps.makeRefCounting())`, which in turn seems to be highly optimized.

ryanzhoudev: OH I see count says *equivalent* to mapToLong

magicus (approved):

PR #2132, opened 2023-09-27
fix: fix mod reload looping if a feature crashes

kristofbolyai:


magicus (approved):

magicus:
I think there was a bug report associated with this problem as well. We should find it and close it.

magicus:
Actually I checked now, but could not find any such issue. I might have been dreaming, then.

PR #2135, opened 2023-09-28
feat: Hightlight duplicate cosmetics

ShadowCat117:
By default it will highlight cosmetics with names matching the one you're hovering and also cosmetics with names matching any of the cosmetics you've already selected to be scrapped. Can be configured to only show hovered or selected.

https://github.com/Wynntils/Artemis/assets/8337467/a39b44cd-7764-4b8b-8190-87516fe7924c


common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
```diff
@@ -0,0 +1,151 @@
...
+                        .getMenu()
+                        .slots
+                        .get(RETURN_SLOT)
+                        .getItem()
+                        .getHoverName()
+                        .equals(RETURN_TEXT)
+                || hoveredSlot.getItem().getHoverName().equals(Component.literal("§f"))) {
+            hoveredCosmetic = null;
+            return;
+        }
```
kristofbolyai: 1. The comment is a bit weird to read to me, might just be me though
2. Please clear this if condition up. It's long.
3. Extract `Component.literal("§f")` as a const?

common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
```diff
@@ -0,0 +1,151 @@
...
+        if (SearchableContainerType.SCRAP_MENU.getBounds().getSlots().contains(e.getSlot().index)) {
+            if (selectedCosmetics.contains(e.getSlot().getItem().getHoverName())
+                    && highlightCondition.get() != HighlightCondition.HOVER) {
+                RenderUtils.drawArc(
+                        e.getPoseStack(), selectedHighlightColor.get(), e.getSlot().x, e.getSlot().y, 200, 1f, 6, 8);
+            } else if (e.getSlot().getItem().getHoverName().equals(hoveredCosmetic)
+                    && highlightCondition.get() != HighlightCondition.SELECTED) {
+                RenderUtils.drawArc(
+                        e.getPoseStack(), hoveredHighlightColor.get(), e.getSlot().x, e.getSlot().y, 200, 1f, 6, 8);
+            }
```
kristofbolyai: The render call is the same, only the color changes. So that makes it easy to clean up this whole if/else chain.

common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onContainerSetContent(ContainerSetContentEvent.Post event) {
+        if (!onScrapMenu) return;
+
+        // Cosmetics have been scrapped and new cosmetic is being displayed, clear selected
+        if (event.getItems().get(RETURN_SLOT).getHoverName().equals(RETURN_TEXT)) {
+            selectedCosmetics.clear();
```
kristofbolyai: We don't have a general rule for this, perhaps we should.. Can you not use clear and set a new ArrayList instance instead? I think that is a bit cleaner, also less race prone (races logically can't happen, but some mod interaction could cause one.)

common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
```diff
@@ -0,0 +1,151 @@
...
+                if (slot == 1) {
+                    selectedCosmetics.clear();
+                }
+
+                break;
+            }
+
+            if (!selectedCosmetics.contains(selectedCosmetic)) {
+                selectedCosmetics.add(selectedCosmetic);
+            }
```
kristofbolyai: Perhaps selectedCosmetic can be a set? Then you can remove contains. Also, IMO it makes more sense since it represents a set of selected items.

kristofbolyai (commented):
Great feature, code is almost ready too.

ryanzhoudev (approved):
I tested in-game, works good even with Silverbull discount/autofill. I did not look at code too closely. Would be cool to see this expanded into something like the new TM stuff though, loading all pages then notifying if there are duplicates on other pages. (Obviously out of scope here)

ShadowCat117:
> I tested in-game, works good even with Silverbull discount/autofill. I did not look at code too closely. Would be cool to see this expanded into something like the new TM stuff though, loading all pages then notifying if there are duplicates on other pages. (Obviously out of scope here)

Oh I didn't know silverbull affected this, 1 thing might not fully work depending on how the 3 slots are displayed vs the 5, can you send a screenshot?

ryanzhoudev:
image

They just fill the first two slots with this item. Everything else is the same.

ShadowCat117:
> They just fill the first two slots with this item. Everything else is the same.

Thanks!

magicus (approved):
I looked briefly through this, and found nothing odd. Please await final approval from bolyai though.

common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
```diff
@@ -0,0 +1,154 @@
...
+    @Persisted
+    public final Config<CustomColor> selectedHighlightColor = new Config<>(CommonColors.ORANGE);
+
+    private static final Component ADD_REWARD_TEXT = Component.literal("§7Click on a reward to add it");
+    // The colored glass panes on the new reward screen
+    private static final Component DECORATIVE_ITEM_NAMES = Component.literal("§f");
+    private static final Component RETURN_TEXT = Component.literal("§7Return to Scrap Menu");
+    private static final int FINAL_SLOT = 53;
+    private static final int RETURN_SLOT = 18;
+    private static final List<Integer> SELECTED_COSMETIC_SLOTS = List.of(1, 2, 3, 4, 5);
```
kristofbolyai, line 50: Might be too much info for a feature. Do you want to move this logic into a model?

kristofbolyai: Although, moving only these constants would be weird in code. Since the events/logic is appropriate here. Okay, perhaps keep this for now, and if we ever extend it, move to a model. Unless you have a better idea.

kristofbolyai (approved):
Looks good. Not sure if we can do anything about the constants for now.

common/src/main/java/com/wynntils/features/inventory/HightlightDuplicateCosmeticsFeature.java
magicus: I think there is some kind of minimum threshold for creating a new model. If we have an existing model, we should prefer to enhance it rather than putting too much logic in a feature, but in absence of an existing model, I think we can accept some amount of "model-y" logic in a feature. This is of course a subjective judgement, but agree that it seems reasonable to keep this in the feature for now.

ShadowCat117: Potentially if this type of feature is used elsewhere, like in the TM as DonkeyBlaster suggested, then it could be moved to a Model but for now yeah it's fine here I think

PR #2137, opened 2023-09-28
fix: Fix corrupted lost sanctuary key overlay text showing as gray instead of red

SpaceMonkeyy86:
A follow-up to #2057. In that PR, I colored all Lost Sanctuary keys gray because the dungeon was removed, unaware that the corrupted version still exists. This PR reverts the corrupted keys to the correct color.

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/ItemTextOverlayFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/ItemTextOverlayFeature.java
@@ -237,12 +237,10 @@ public class ItemTextOverlayFeature extends Feature {
             String text = dungeon.getInitials();
 
             CustomColor textColor;
-            if (dungeon.isRemoved()) {
-                textColor = REMOVED_COLOR;
-            } else if (item.isCorrupted()) {
-                textColor = CORRUPTED_COLOR;
+            if (item.isCorrupted()) {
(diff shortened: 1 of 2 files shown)


ryanzhoudev (approved):

common/src/main/java/com/wynntils/features/inventory/ItemTextOverlayFeature.java
```diff
@@ -237,12 +237,10 @@ public TextOverlay getTextOverlay() {
...
-            if (dungeon.isRemoved()) {
-                textColor = REMOVED_COLOR;
-            } else if (item.isCorrupted()) {
-                textColor = CORRUPTED_COLOR;
+            if (item.isCorrupted()) {
+                textColor = dungeon.isCorruptedRemoved() ? REMOVED_COLOR : CORRUPTED_COLOR;
             } else {
-                textColor = STANDARD_COLOR;
+                textColor = dungeon.isRemoved() ? REMOVED_COLOR : STANDARD_COLOR;
             }
```
kristofbolyai, line 244: Perhaps it would be cleaner if you follow this pseudo code.

```
if item is removed
then return removed color

return iscorrupted ? corruptedcolor : normalcolor
```

kristofbolyai: Nevermind.., I just saw the change..

kristofbolyai (approved):

PR #2138, opened 2023-09-29
feat: Add StyledText#separate, support substring formatted-index splitting and use it

kristofbolyai:
This worked as expected in the TM. I think I added test cases for edge cases, but as always, it can be improved further..


magicus:
🐨 ❤️! 

A brief skimming of the code seems fine, but I want to check this more closely.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -428,19 +443,63 @@ public StyledText substring(int beginIndex, int endIndex) {
...
+        return separate(PartStyle.StyleType.NONE, indexes);
+    }
+
+    /**
+     * Splits this {@link StyledText} into multiple {@link StyledText}s at the given indexes.
+     * @param styleType the style type to use when calculating indexes
+     * @param indexes the indexes to split at, in ascending order
+     * @return the split {@link StyledText}s as an array
+     */
+    public StyledText[] separate(PartStyle.StyleType styleType, int... indexes) {
```
magicus: I were imagining a "split sibling" that would just keep the separating part as well. This is probably more clear and useful; thanks! 

However, I wonder if not "partition" would be a better name for this functionality.

kristofbolyai: Hm, you decide. Either name is fine by me.

magicus: Then I decide "partition", since this is what this function does :)

magicus (approved):

PR #2139, opened 2023-09-29
fix: Improve chat screen handling for game messages

### LINKED ISSUE #1951: Mysterious Obelisk message doesn't count as dialogue
image

magicus:
This fixes #1951. The problem was that the message "As you touch..." behaved in the same way as NPC dialogue (it disappeared and should not be stored as the last real chat line), but we did not detect it as such.

This entire code is somewhat finicky, but I believe this is a robust enough fix.

Also fixed so last real chat line (and others) get reset when reconnecting.

Also added junit test junk files to .gitignore...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,10 @@ classes/
 .gradle/
 .architectury-transformer/
 
-# Data generated while running Minecraft
+# Data generated while running Minecraft or tests
 run/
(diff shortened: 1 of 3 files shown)


Mcrtin:
Thx! seems straightforward.

kristofbolyai (approved):

PR #2146, opened 2023-09-30
feat: Add a GUI for crowd sourcing settings

kristofbolyai:
I would love to get this in while we are still on 1.19.4. Then I can update the 1.20.2 branch, so we will be ready for 2.0.4, when we will need this.

image
image


ShadowCat117:
I wonder if we should prompt the user to enable/disable crowd sourcing, like how we currently do for telemetry? Have it on by default, then two options in chat for disable or let me choose what to collect and that will open this screen

kristofbolyai:
> I wonder if we should prompt the user to enable/disable crowd sourcing, like how we do for telemetry? Have it on by default, then two options in chat for disable or let me choose what to collect and that will open this screen

Sure, I can add it

kristofbolyai:
Confirmation messages are now in.

ShadowCat117 (approved):
Looks good to me

PR #2151, opened 2023-10-01
refactor: Simplify component dependencies

magicus:
It turns out that we can simplify component dependency declaration. While the core idea of having components declare dependencies on "sibling" components (e.g. models to models), this was extended further than needed. In fact, only dependencies used during construction of the component needs to be declared. This simplifies our declaration, and it also makes it much clearer when you in fact need to declare a dependency -- it will be used shortly afterwards in the constructor!


kristofbolyai (approved):
In theory this looks good, but needs to be tested in-game.

PR #2154, opened 2023-10-01
feat: Waypoint manager redesign

ShadowCat117:
For PoiManagementScreen I wouldn't even bother reviewing differences as so much changed, just view it as a whole new thing.

Now includes:

- Selectable widgets, easily move and delete multiple pois at once. Can also be used to only export selected pois.
- Searching, filter displayed pois by searching by name.
- Icon filtering, filter displayed pois by only including ones with specific icon(s).
- Sorting, sort displayed pois by name, x, y or z. Includes ascending & descending

https://github.com/Wynntils/Artemis/assets/8337467/db0490cb-f6fa-4d0f-b45c-4d2eaae7eb43

Icon filtering is slightly overkill for the moment but when custom icons are a thing it'll hopefully be less so. The screen can also be used for map/minimap icon filtering which I have a branch ready for once the screen has been cleaned up here


common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
```diff
@@ -0,0 +1,285 @@
...
+
+        // The icons that will be displayed
+        usedIcons = new ArrayList<>(icons.keySet());
+
+        populateIcons();
+    }
+
+    @Override
+    public void doRender(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick) {
+        renderBackground(guiGraphics, mouseX, mouseY, partialTick);
```
kristofbolyai: I think this is left behind from 1.19.4. 1.20.2's `super#render` calls `renderBackground`. Perhaps you don't need this? Also, you might have to call `super#render` first now.

common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
```diff
@@ -0,0 +1,285 @@
...
+            if (mouseX >= scrollButtonRenderX
+                    && mouseX <= scrollButtonRenderX + dividedWidth
+                    && mouseY >= scrollButtonRenderY
+                    && mouseY <= scrollButtonRenderY + scrollButtonHeight) {
+                draggingScroll = true;
+            }
+        }
+
+        // Returning super.doMouseClicked(mouseX, mouseY, button) causes ConcurrentModificationException
+        return true;
```
kristofbolyai: Doesn't seem to happen to me, either way, I don't think this comment is needed.

ShadowCat117: Maybe 1.20.2 fixed it somehow. Only left it in incase someone knew how to fix it

kristofbolyai: Hm, okay. It does happen. It happens because when you click the include/exclude all buttons, you repopulate all widgets, while iterating over then. But I don't think you need to call this super, you do everything needed above.

kristofbolyai:
Okay. I started reviewing the code, but I might as well comment on the small changes I would request UI wise, then check the code.

- Scroll button: 
Really big on big screens, but that is a small issue. When scrolled fully down, it renders too low. Also, it feels really unresponsive. I know it was me who added all those bad scroll buttons in UIs. I hope to fix this issue, starting with the Custom TM Screen scroll button. Please use an implementation similar to that, as it is the smoothest scroll button we have (and the implementation we should port everywhere).
image

- Icon Filter Screen Buttons:
What do you think of rendering the POI icons in a scaled way when there is space? It looks a bit small to me.
image

- Sort by icon: I think this is really easy to add, and is a reasonable thing to do. This is not a required change though, unlike the scroll button.

ShadowCat117:
> Really big on big screens

Hmm, I didn't want it to be too small on big screens but I'll see what I can do

> Please use an implementation similar to that

I did use some code from the trade market pr but maybe there was something else I needed that doesn't fully work with old scrolling code

> What do you think of rendering the POI icons in a scaled way when there is space

Maybe, so we have a maximum and a minimum size they can be and try to fit as many as it can on each row?

> Sort by icon

That should be doable

kristofbolyai:
It might also be I have a small set of pois to scroll, still weird.

Sure, I guess that can work, just making sure we use the space we have

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        // Highlight to show inclusion
+        if (included) {
+            RenderUtils.drawRect(poseStack, CommonColors.RED.withAlpha(35), getX(), getY(), 1, width, height);
+        }
+    }
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
```
Mcrtin, line 83: There is an `onClick` method. Maybe it's applicable here.

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
```diff
@@ -0,0 +1,76 @@
...
+    public IconFilterWidget(
+            int x, int y, int width, int height, Texture icon, IconFilterScreen filterScreen, boolean included) {
+        super(x, y, width, height, Component.literal("Icon Filter Widget"));
+        this.icon = icon;
+        this.filterScreen = filterScreen;
+        this.included = included;
+
+        // Calculate x/y position of the icon to keep it centered
+        iconRenderX = (x + width / 2f) - icon.width() / 2f;
+        iconRenderY = (y + height / 2f) - icon.height() / 2f;
```
Mcrtin: I think this should be done at render time to keep these things together.

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
```diff
@@ -0,0 +1,76 @@
...
+
+        tooltip = included
+                ? List.of(Component.translatable(
+                        "screens.wynntils.iconFilter.filterExclude.tooltip", EnumUtils.toNiceString(icon)))
+                : List.of(Component.translatable(
+                        "screens.wynntils.iconFilter.filterInclude.tooltip", EnumUtils.toNiceString(icon)));
```
Mcrtin, line 51: this too? or if not then you can also make the transformation (`Lists.transform(tooltip, Component::getVisualOrderText)`) here.

common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        if (!draggingScroll && (usedIcons.size() > MAX_ICONS_TO_DISPLAY)) {
+            float scrollButtonRenderX = (int) (dividedWidth * 52);
+
+            if (mouseX >= scrollButtonRenderX
+                    && mouseX <= scrollButtonRenderX + dividedWidth
+                    && mouseY >= scrollButtonRenderY
+                    && mouseY <= scrollButtonRenderY + scrollButtonHeight) {
+                draggingScroll = true;
```
Mcrtin: you can call `setDragging(true)` to avoid having to keep track of it yourself.

magicus (commented):
I had a quick glance, and found nothing out of the ordinary. Not a full review though, bolyai is the GUI master and I'll leave it to him.

ShadowCat117:
Ok scrolling has hopefully improved, seems alright to me but since you have a bigger screen I can't really test that so whenever you've got the chance could you check that, I've also shrunk the scroll bar and it hopefully won't go over the bottom anymore.

I've also got it to increase the size of the icon widgets if you don't fill up a full page with the current icons. It still only increases them to the point where they will still fit on one page so there will be some space at the bottom sometimes but it's definetly better

common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
ShadowCat117: We do it this way with every other scroll bar so this is fine

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
ShadowCat117: It only needs to be calculated once, same with the tooltip so no need to do it at render time

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
ShadowCat117: I think `mouseClicked` is fine, `onClick` doesn't do anything in the super whereas this will

kristofbolyai:
> Ok scrolling has hopefully improved, seems alright to me but since you have a bigger screen I can't really test that so whenever you've got the chance could you check that, I've also shrunk the scroll bar and it hopefully won't go over the bottom anymore.

Seems like the scrolling texture is now in the correct place, but the dragging seems unreactive when you are scrolling up from down.

https://github.com/Wynntils/Artemis/assets/49001742/c15860c0-c704-41a5-ac39-0673770a9cef

kristofbolyai:
> I've also got it to increase the size of the icon widgets if you don't fill up a full page with the current icons. It still only increases them to the point where they will still fit on one page so there will be some space at the bottom sometimes but it's definetly better

Arguably, this also looks weird.. Kind of like the other end of the spectrum. I haven't read the code yet, but perhaps define a range which looks acceptable to you? GUI scaling will take care of the rest.

image

common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
```diff
@@ -0,0 +1,346 @@
...
+                backgroundX,
+                backgroundY,
+                0,
+                backgroundWidth,
+                backgroundHeight,
+                Texture.WAYPOINT_MANAGER_BACKGROUND.width(),
+                Texture.WAYPOINT_MANAGER_BACKGROUND.height());
+    }
+
+    public void toggleIcon(Texture icon) {
```
kristofbolyai: Let's keep the order of classes we usually use, that is:

- Constructors
- Overriden methods
- public
- protected
- private

common/src/main/java/com/wynntils/screens/maps/IconFilterScreen.java
```diff
@@ -0,0 +1,346 @@
...
+        // If there are less icons than the default per page, see how big we can make the icons whilst still staying on
+        // one page
+        if (usedIcons.size() < MAX_ICONS_PER_PAGE) {
+            int renderAreaWidth = (int) (dividedWidth * 36);
+            int renderAreaHeight = (int) (dividedHeight * 40);
+            int sizeToCheck;
+
+            // Maximum size should be fitting 2 icons on screen at once which will be roughly half of the width of the
+            // renderable area
+            for (int i = (DEFAULT_WIDGET_SIZE_MULTIPLIER + 1); i < MAX_WIDGET_SIZE_MULTIPLIER; i++) {
```
kristofbolyai: I think keep the for loop's i variable clean, and do the size multiplication inside the loop

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -11,40 +11,84 @@
...
+    private Map<Texture, Boolean> filteredIcons = new EnumMap<>(Texture.class);
+    private PoiSortButton activeSortButton;
+    private PoiSortButton iconSortButton;
+    private PoiSortButton nameSortButton;
+    private PoiSortButton xSortButton;
+    private PoiSortButton ySortButton;
+    private PoiSortButton zSortButton;
+    private PoiSortOrder sortOrder;
+    private TextInputBoxWidget focusedTextInput;
+    private TextInputBoxWidget searchInput;
```
kristofbolyai: Perhaps keep some kind of groupping here?

I like the approach I used in the `TradeMarketHolder` class in the custom TM PR. Basically just comment regions (but not actual // region ...)

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+                (int) (backgroundWidth - filterButtonWidth - 10),
+                20,
+                (s) -> {
+                    scrollOffset = 0;
+                    populatePois();
+                },
+                this,
+                searchInput);
+
+        this.addRenderableWidget(searchInput);
```
kristofbolyai, line 259: I think you forgot to copy the old searchInput's value on reinit, if it exists already

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+
+        // region sort buttons
+        int iconTitleWidth = (McUtils.mc()
+                        .font
+                        .width(StyledText.fromComponent(
+                                        Component.translatable("screens.wynntils.poiManagementGui.icon"))
+                                .getString()))
+                + 1;
+
+        int nameTitleWidth = (McUtils.mc()
```
kristofbolyai: I18n.get(key)

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+                0,
+                backgroundWidth,
+                backgroundHeight,
+                Texture.WAYPOINT_MANAGER_BACKGROUND.width(),
+                Texture.WAYPOINT_MANAGER_BACKGROUND.height());
+    }
+
+    public void selectPoi(CustomPoi selectedPoi) {
+        boolean add = true;
+        boolean updateState = false;
```
kristofbolyai: `updateState` is not really needed, `add` can hold the logic.. Also, please give it a better name

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+
+        customPois.get().remove(poiToDelete);
+        customPois.touched();
+
+        deletedPois.add(poiToDelete);
+        deletedIndexes.add(deletedPoiIndex);
+
+        undoDeleteButton.active = true;
+
+        // Handles deleting when the final poi is visible
```
kristofbolyai:
```suggestion
        // Handles scrolling when deleting the last poi
```

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+                customPois.get().indexOf(poiToSwap));
+
+        customPois.touched();
+
+        populatePois();
+    }
+
+    @Override
+    public boolean doMouseClicked(double mouseX, double mouseY, int button) {
+        // FIXME: Search bar cursor dissapears after interacting with it but still remains focused
```
kristofbolyai: I can look into this if needed

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+            if (child.isMouseOver(mouseX, mouseY)) {
+                return child.mouseClicked(mouseX, mouseY, button);
+            }
+        }
+
+        if (!draggingScroll && (pois.size() > maxPoisToDisplay)) {
+            if (mouseX >= scrollButtonRenderX
+                    && mouseX <= scrollButtonRenderX + (dividedWidth / 2)
+                    && mouseY >= scrollButtonRenderY
+                    && mouseY <= scrollButtonRenderY + scrollButtonHeight) {
```
kristofbolyai: Isn't there a mathutil for this? In range..? Perhaps create one, "inArea"?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+                draggingScroll = true;
+            }
         }
 
-        checkAvailablePages();
+        return true;
+    }
+
+    @Override
+    public boolean mouseReleased(double mouseX, double mouseY, int button) {
```
kristofbolyai: put this after mouseDragged please. Also put mousedragged after mouseclicked. Then scroll as last..

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+                .collect(Collectors.toList());
+
+        // No pois
+        if (pois.isEmpty()) {
+            return;
+        }
+
+        // Filter pois based on filtered icons
+        pois = pois.stream()
+                .filter(poi -> filteredIcons.getOrDefault(poi.getIcon(), true))
```
kristofbolyai, line 653: Why true..?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -67,79 +111,756 @@ public void onClose() {
...
+            row += 20;
+
+            poiManagerWidgets.add(poiWidget);
+
+            this.addRenderableWidget(poiWidget);
+        }
+    }
+
+    private List<CustomPoi> sortPois() {
+        List<CustomPoi> sortedPois = pois;
```
kristofbolyai: You don't actually copy the list so you just sort `pois`. If that is intended, I am not sure why this method returns anything.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -11,40 +11,84 @@
...
 import net.minecraft.client.resources.language.I18n;
 import net.minecraft.network.chat.Component;
 
-public final class PoiManagementScreen extends WynntilsScreen {
-    private final MainMapScreen oldMapScreen;
-
-    private Button nextButton;
-    private Button previousButton;
-    private Button undoDeleteButton;
+public final class PoiManagementScreen extends WynntilsScreen implements TextboxScreen {
```
kristofbolyai, line 47: This class generally needs to be reordered. Override, public, then private methods.

kristofbolyai (commented):
Looked through the whole code now.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: If you wouldn't mind I would appreciate that, probably something simple I just can't see

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: Ah yes, there's `isInside`, but it's not used anywhere. I think a few other places could use it so shall I change them in this too?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: true is to include it, so the default is to include all icons. It might not need the default honestly as `updateAllUsedIcons` should hopefully populate it with all icons

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
ShadowCat117: It works? Isn't that what passing searchInput into the constructor does?

```
if (oldWidget != null) {
    this.textBoxInput = oldWidget.textBoxInput;
    setCursorAndHighlightPositions(oldWidget.cursorPosition);
    this.renderColor = oldWidget.renderColor;
}
```

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: Perhaps, but only require it here

common/src/main/java/com/wynntils/screens/maps/widgets/IconFilterWidget.java
Mcrtin: Again, i don't see a reason why you make the transformation of the tooltip at render time when you are already doing everything else in the constructor.

ShadowCat117: Everywhere else we call `setTooltipForNextRenderPass` is done this way I don't see why this should be an exception

ShadowCat117:
I've capped it so it won't go past a certain size, this is the biggest they can be. May also change the icon to be scaled to make it look better, would you say this is acceptable? Can't really think of a better way to not make them too big but also not leave a lot of empty space

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: Yeah I missed that. Some other search box didn't support that, not sure where I saw it.

kristofbolyai (approved):
Looks good to me now. Hope you are up for reworking parts of this soon :)

kristofbolyai:
Why wasn't this merged..? I guess I didn't press the button..

PR #2156, opened 2023-10-01
fix: Fix container search arc rendering weirdly on items with durability arcs

kristofbolyai:


magicus (approved):
I think we're misusing the arc feature somewhat. We should really have some other way to highlight items. But let's fix it for now so it is not broken.

PR #2157, opened 2023-10-01
feat: Crowd source NPC and profession locations

kristofbolyai:
Sample data:

https://gist.github.com/kristofbolyai/e2ff441daf9ece39e0e8c4413da8a9bc


common/src/main/java/com/wynntils/models/label/LabelModel.java
```diff
@@ -0,0 +1,55 @@
...
+            if (info == null) continue;
+
+            WynntilsMod.postEvent(new LabelIdentifiedEvent(info));
+            return;
+        }
+    }
+
+    private void registerParsers() {
+        // Special order for optimization
+        registerParser(new NpcLabelParser());
```
magicus: This kind of feels a bit more like a Handler than a Model. Wynn uses entity labels for a variety of purposes, so the label itself does not really match to something we want to model as part of the world, but rather as a way we get info that we want to model.

If we need to have a certain order for the parsers it will make the inverse dependency registration a bit more tricky, but we can probably have like `Handlers.Label.registerParser(new NpcLabelParser, HIGH_PRIORITY)` or something like that.

kristofbolyai: We have LabelHandler. But it handles generic entity name tag change events.

I opted for a LabelModel since we do wynn specific parsing here.

common/src/main/java/com/wynntils/models/label/type/LabelInfo.java
```diff
@@ -0,0 +1,47 @@
...
+    protected final transient StyledText label;
+
+    protected final String name;
+    protected final String formattedLabelString;
+    protected final Location location;
+
+    protected LabelInfo(StyledText label, Location location) {
+        this.label = label;
+        this.name = label.getStringWithoutFormatting();
+        this.formattedLabelString = label.getString();
```
magicus: Do you really need formattedLabelString? What does it provide that label does not?

kristofbolyai: More of an "in case" if we ever need it during post processing

common/src/main/java/com/wynntils/models/label/LabelModel.java
magicus: Actually, we already *have* a LabelHandler. (I was a bit surprised, since I thought I had added one).

I think this functionality should be folded into that.

common/src/main/java/com/wynntils/models/label/type/LabelInfo.java
magicus: Can't you just call getString on the returned label then?

common/src/main/java/com/wynntils/models/label/LabelModel.java
magicus: We can do "wynn specific" parsing in a handler, as long as we have the dependency inversion pattern of e.g. ActionBarHandler, where we register wynn-specific callbacks.

I think this would be better redesigned so that LabelHandler has the additional responsibility of sending new labels with a list of registered parsers, and sending out a LabelIdentifiedEvent with a LabelInfo, when a parser returns non-null.

The actual parsers should reside in more specific models. The gathering stuff clearly belongs to ProfessionModel. We do not really have a NpcModel, something I have been pondering for a while. (I would like to connect NPC characters with NPC dialogue lines, and have a place to prepare for Voices of Wynn integration.) So maybe it is time to create one.

magicus: One way of thinking about this is that I believe a "model" should match a concept that any Wynncraft player will have. They will know stuff such as emeralds, rewards, damage, profession. Stuff like "label" is very much minecraft-centric. I ljust ooked through all our models, and there is one more that really should be a handler as well, and that is particle. It is the same situation -- particles are a minecraft concept; we need to separate them for all the different meanings they have in Wynncraft. The user model is that "this kind of particle effects mean it is a lootrun thingie, this other means it is an important quest place, etc".

common/src/main/java/com/wynntils/models/label/type/LabelInfo.java
kristofbolyai: That way GSON wouldn't serialize it.

magicus: Maybe we need a json type adapter for StyledText? That could come in handy elsewhere, too...

kristofbolyai: Hm, I could add it, but separately. This is probably needed tomorrow.

common/src/main/java/com/wynntils/core/crowdsource/type/CrowdSourcedDataType.java
```diff
@@ -6,13 +6,19 @@
...
 
 /**
  * This enum represents the type of crowd sourced data that is being collected.
  */
 public enum CrowdSourcedDataType {
-    LOOTRUN_TASK_LOCATIONS(LootrunTaskLocation.class);
+    LOOTRUN_TASK_LOCATIONS(LootrunTaskLocation.class),
+    NPC_LOCATIONS(NpcLabelInfo.class),
+    PROFESSION_NODE_LOCATIONS(GatheringNodeLabelInfo.class),
+    PROFESSION_STATION_LOCATIONS(GatheringStationLabelInfo.class);
```
ShadowCat117: Not sure I'm a fan of these names. I think they should be Crafting Stations

ShadowCat117:
Aside from the crafting station name and magicus' suggestion I think this is good

common/src/main/java/com/wynntils/core/crowdsource/type/CrowdSourcedDataType.java
kristofbolyai: I want to have the profession prefix. Profession Crafting Station should be a good choice.

common/src/main/java/com/wynntils/models/label/type/LabelInfo.java
kristofbolyai: Gonna add this now.

kristofbolyai:
Also adding an initial "right click to disable" option.
image

common/src/main/java/com/wynntils/models/profession/label/GatheringStationLabelInfo.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.profession.label;
+
+import com.wynntils.core.text.StyledText;
+import com.wynntils.models.profession.type.ProfessionType;
+import com.wynntils.utils.mc.type.Location;
+
+public class GatheringStationLabelInfo extends ProfessionLabelInfo {
```
ShadowCat117: Perhaps have this as crafting station too, you don't really gather at them

ShadowCat117 (approved):

PR #2158, opened 2023-10-02
feat: Add initial new MapData service [skip ci]

magicus:
This is the core of the new mapdata functionality. A few things are missing, most notably proper support for visibility/zoom levels. 

The new mapdata features are not currently used. The build-in providers are currently not populated. I have a WIP add-on patch to this which will enable those, for testing purposes, but we cannot enable it just yet since that would mean a regression in functionality.

I tried to split up the PR a bit in separate commits to show how the code is structured. It might be that some parts would not compile without following commits, even though I tried my best to get to free-standing commits.

Most likely the design will need some additional massaging when we really put it to the test, and finish the work of actually getting the current POIs into the system.


magicus:
A few words about the design. It is inspired by html+css -- we have attributes that "cascade" down a chain of providers, categories and then features. The idea is that the user should be able to turn on and off individual providers (which can come from us -- built-in, or json files "officially" distributed by us -- or be a third-party addition (user stored locations/waypoints is like a third-party addition in this sense). But they can also specify the order of providers, and attributes at a more specific level override attributes at a more general level.

So you can basically have a "texture pack" by incorporating a file which overrides the icon for e.g. "wynntils:service:item-identifier", while still getting the actual location for the identifiers from our official source.

magicus:
A provider contributes essentially three things: a list of features ("POIs"), a list of categories and a list of icons.

An icon is a mapping name -> base64-encoded jpeg.

A category is just a name-space prefix (with a user-friendly name)

A feature is more complicated -- it is either a point (location), a polyline (path) or a polygon (area). The latter two has not really been used so far, but my intention is to fold lootrunpaths into MapPath, and territories into MapArea. And it will allow for future uses. For instance, your totem's AoE could show up as a MapArea on the minimap, and the various effort to build path-finding support ("navsat") could utilize MapPath.

magicus:
All features belong to a category. Both features and categories has attributes, this allows for attribute inheritance both feature-wise and category-wise. Feature-wise always overrides category-wise.

Categories has an inherent hierarchy, so "wynntils:service:profession:tailoring" is a subcategory to "wynntils:service:profession", which in turn is a subcategory to "wynntils:service", and this in turn is our top official category "wynntils".

This way you can e.g. set visibility to "wynntils:service" to true, but ""wynntils:service:profession" to false, but then nevertheless set ""wynntils:service:profession:tailoring" to true, and then you will show all services in general, but not those related to professions, except tailoring, which will show nevertheless.

And not only can you set visibility, you can set/override any attribute, such as label color, icon, etc.

I think this will create a very powerful, flexible but yet comprehensible system for dealing with mapfeatures of all different kinds. Creating a good UI might be a bit of a challenge, though, but I have some ideas.

magicus:
This structure is also embedded in the json format. Basically, all kinds of mapfeatures can be provided as json (with one exception, and that is the "overlay" which must be dynamic. Currently this is the health bar of other players, but this is open for future extensions.)

Here is a demo json file I have been using: https://gist.githubusercontent.com/magicus/a2c810380a34a7474a2651400d36d72c/raw/online-mapdata.json

kristofbolyai:
I am excited to review this. However, it has to wait until tomorrow. I haven't slept in around 30 ish hours, so reviewing now wouldn't be of much use..

magicus:
Take your time. At this point, it will not actually be usable, so it is still mostly a "groundwork" kind of thing.

common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
```diff
@@ -8,38 +8,40 @@
...
-    SCRIBING_STATION("Scribing Station", Texture.SCRIBING_STATION),
-    SCROLL_MERCHANT("Scroll Merchant", Texture.SCROLL_MERCHANT),
-    SEASKIPPER("Seaskipper", Texture.SEASKIPPER),
-    TAILORING_STATION("Tailoring Station", Texture.TAILORING_STATION),
-    TOOL_MERCHANT("Tool Merchant", Texture.TOOL_MERCHANT),
-    TRADE_MARKET("Trade Market", Texture.TRADE_MARKET),
-    WEAPON_MERCHANT("Weapon Merchant", Texture.WEAPON_MERCHANT),
-    WEAPONSMITHING_STATION("Weaponsmithing Station", Texture.WEAPONSMITHING_STATION),
-    WOODWORKING_STATION("Woodworking Station", Texture.WOODWORKING_STATION);
+    ALCHEMISM_STATION("Alchemism Station", Texture.ALCHEMIST_STATION, "profession:alchemism"),
```
kristofbolyai, line 11: At first look it might be better to represent these values in code with some kind of actual class? That way it can be easier to work with/refactor?

But I am only a few classes in the review, so perhaps it is not needed at all...

common/src/main/java/com/wynntils/services/map/Label.java
```diff
@@ -47,9 +50,37 @@ public String getLevel() {
         return level;
     }
 
+    public int getCombatLevel() {
+        if (level == null) return 0;
```
kristofbolyai, line 54: Shouldn't we really do this parsing time? is `getLevel` even used..?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
```diff
@@ -0,0 +1,37 @@
...
+
+    // If this is MapFeatureIcon.NO_ICON_ID ("none"), then no icon will be displayed
+    // null means inherit
+    String getIconId();
+
+    // 1-1000, 1000 is highest priority (drawn on top of everything else)
+    // 0 means no value specified; inherit
+    int getPriority();
+
+    // 0 means no information is available, or suitable for all combat levels
```
kristofbolyai, line 23: Shouldn't this be split into 2 different values then..? Kind of weird, but imagine filtering for all POIs "my level", perhaps I don't want to include unspecified level ranges... Not too sure on how/if to fix the problem though

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
```diff
@@ -0,0 +1,37 @@
...
+    int getLevel();
+
+    MapVisibility getLabelVisibility();
+
+    CustomColor getLabelColor();
+
+    TextShadow getLabelShadow();
+
```
kristofbolyai, line 31: Perhaps these 3 (or 4 with the label id) could be a record on it's own? (Same below)

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapDecoration.java
```diff
@@ -0,0 +1,13 @@
...
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.mapdata.attributes.type;
+
+// Currently only for player health bar, but can be extended to more types of
+// overlays
+public interface MapDecoration {
+    boolean isVisible();
+
+    int getBarPercentage();
```
kristofbolyai, line 12: This feels too specific but it's hard to design for currently non-existent cases

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
```diff
@@ -0,0 +1,15 @@
...
+
+public enum MapVisibility {
+    ALWAYS,
+    NEVER
+
+    // FIXME: This needs to be more complex, and contain (at least):
+    // below zoom: first visible at, fully visible at
+    // above zoom: fully visible at, last visible at
+    // between zoom: both above
+}
```
kristofbolyai, line 15: I am not too sure what is the problem with implementing this. It feels like you've kind of got it figured out with the "variables". You can just define pre-set values as static fields here to use? And leave the option for creating a custom one.

common/src/main/java/com/wynntils/services/mapdata/attributes/FullCategoryAttributes.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    protected <T> T getAttribute(Function<MapAttributes, T> getter) {
+        for (String id = categoryId; id != null; id = getParentCategoryId(id)) {
+            Stream<MapAttributes> allAttributes = Services.MapData.getCategoryDefinitions(id)
+                    .map(MapCategory::getAttributes)
+                    .filter(Objects::nonNull);
+            Optional<T> attribute =
+                    allAttributes.map(getter).filter(Objects::nonNull).findFirst();
+            if (attribute.isPresent() && !(attribute.get() instanceof Integer i && i == 0)) {
```
kristofbolyai, line 29: This line kind of begs the question if we should use boxed integers in the attributes..? It would also solve differentiating between "no value" and "unknown value"

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
```diff
@@ -0,0 +1,201 @@
...
+import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.colors.CustomColor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
+public class CategoriesProvider extends BuiltInProvider {
+    private static final List<MapCategory> PROVIDED_CATEGORIES = new ArrayList<>();
+
+    public CategoriesProvider() {
```
kristofbolyai, line 23: Can you fix the inconsistent spacing?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        @Override
+        public String getName() {
+            return "Your position";
+        }
+
+        @Override
+        public MapAttributes getAttributes() {
+            return new MapAttributes() {
```
kristofbolyai: Why not use the abstract one?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/MapIconsProvider.java
```diff
@@ -0,0 +1,102 @@
...
+
+        @Override
+        public int width() {
+            return texture.width();
+        }
+
+        @Override
+        public int height() {
+            return texture.height();
+        }
```
kristofbolyai: Nit, but if you use "record style getters (without get prefix)" and "class style getters (with get prefix)" in the same class it looks weird. I think we are fine with using the record style.

common/src/main/java/com/wynntils/services/mapdata/type/MapFeature.java
```diff
@@ -0,0 +1,22 @@
...
+    String getFeatureId();
+
+    // Required.
+    String getCategoryId();
+
+    // Optional
+    MapAttributes getAttributes();
+
+    // Optional
+    List<String> getTags();
```
kristofbolyai, line 21: I have not seen "tags" yet, what are these?

common/src/main/java/com/wynntils/services/mapdata/MapDataService.java
```diff
@@ -0,0 +1,62 @@
...
+    public MapAttributes getFullFeatureAttributes(MapFeature feature) {
+        return new FullFeatureAttributes(feature);
+    }
+
+    public String getCategoryName(String categoryId) {
+        return getCategoryDefinitions(categoryId)
+                .map(MapCategory::getName)
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse("NAMELESS CATEGORY");
```
kristofbolyai: Can you make this a const so it is easier to access in the future? Also perhaps all caps is a little too much, what about "Nameless Category"?

kristofbolyai (commented):
Looks really exciting!

common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
magicus: You mean the tag "profession:alchemism"? It is basically open for all to define their own tag; we provide some by default. My idea was to not make it too strict, but rather have it easily extendable. As long as you have like a category with an icon (and ideally a user-friendly name), you can tag your map features with that category. Or you can replace name and/or icon for that category. So it's kind of a loose coupling, that is only based on matching names.

common/src/main/java/com/wynntils/services/map/Label.java
magicus: This is kind of hacky, to be backwards compatible. The data source we're using (map.wynncraft.com) mostly has level as an integer, but in a few places it looks like "60-70" or something like that. My plan is basically to keep this data source for labels as an alternative map data provider if you really want to use it, but by default instead offer a "curated" list of labels, where we can hand-polish some of the shortcomings of the official map.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: That is so heavily dependent on the data source. Let's say someone publishes a list of good places to grind mobs for an unusual ingredient. They might not bother to specify level ranges, but that does not mean it is suitable for all levels. Or if someone publishes a list of fancy names they are giving all entrances to Ragni. (Why not..?) Then it would not even be really clear if that would be suitable for all levels, or if the concept of level does not really apply.

I agree that it is a bit odd if you want to filter on a certain level. But it's kind of like when you sort quests based on distance -- what to do with the quests that do not have a location? In legacy, we used to sort them first as if they were the nearest quest. This was probably wrong, but so would sort them last be.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: Yeah... I went back and forth with having separate Label and Icon records, and then inlining them in the general attributes, extracting them again, and so on. 

My arguments for having them "inlined" is: 

1) We only get a single attribute class, which simplifies the overriding method. Otherwise you'd have to handle the case of someone wanting to only override label shadow but not color etc. The complexity in handling this (and making sure you do not think incorrectly in the code at any one step) rises sharply.

2) Icon and label were somewhat similar, but still in the end different enough that it did not make sense to unify them. But that would mean we should have an icon class and a label class with some redundancy nevertheless, and that would not feel satisfactory either. And if we tried to unify them, we'd get an even increased complexity with some AbstractMapAttributeAspect having basically color and visibility.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapDecoration.java
magicus: My vague idea here was that we could potentially add other kind of "decorations" to map icons, but whatever that should be, I don't know. But all such decorations would need to be very specialized anyway. So this is intentionally a "backdoor" to allow highly specialized code for certain aspects of certain map features.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
magicus: The problem here is mostly about trying to "reverse engineer" the current system. I think I designed it (or at least had a part in implementing it), but it is close to incomprehensible. :-) I remember I just tweaked a lot of values and tested zooming in and out until it looked kind of good. :-) I would like to replace this with a more logical system where you say like "begin fading in at zoom level X1, and reach full visibility at zoom level X2". That would force the map visibility to be a more complex type than a simple enum. I guess basically an interface implementing `float getVisibilityPercentage(float zoomLevel)` or something like that, and make `ALWAYS` return 1.0 regardless of zoomLevel, `NEVER` return 0.0 all the time, and have a way to express "sometimes visible" nicely.

magicus: It is not super difficult, I just got stuck when implementing this... I have a list somewhere, where I wrote down at which zoom levels the current implementation actually starts and ends showing or hiding different elements.

common/src/main/java/com/wynntils/services/mapdata/attributes/FullCategoryAttributes.java
magicus: I really really dislike API:s that use Integer or any other of the boxed value classes. That typically always end up of breaking the principle of least surprise, since now you can suddenly have null for what you thought was an integer. I have explicitly tried to make sure that the API as provided by the MapAttributes should have proper `int` for the (two) places where it returns numerical values.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
magicus: Done, thanks!

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
magicus: Good point. I think I had not created the abstract class when I wrote this code.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/MapIconsProvider.java
magicus: Yeah, that was odd. Don't know why width and height were defined without java bean prefix. All other attributes have them, so I changed width and height to align to them.

common/src/main/java/com/wynntils/services/mapdata/MapDataService.java
magicus: Fixed. I think this could need some additional polishing, how to best handle this extreme case (I'd almost like to call it an error, if we have categories without name).

common/src/main/java/com/wynntils/services/map/Label.java
kristofbolyai: I see. But technically this still could be in the constructor? Either way, perhaps add a comment for the future us.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: I only worry about this since this is a design decision we have to "fix" now, if we ever want to. Unless you want to go down the versioning rabbit hole..? Btw, does this system support any kind of versioning when parsing? That might be simple but really useful to explore.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: Okay, good reasoning, let's keep it as is.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapDecoration.java
kristofbolyai: Okay. Since this is internal, we can refactor it later. I only worry about creating a solid structure for the user-parsed JSON format.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
kristofbolyai: Yeah, I think we share the same idea implementation wise. I don't see what problems you can run into when coding it, but perhaps there is something.

Again, my only worry is releasing a version of the JSON format in a  way that users don't need to update the format many times.. Perhaps we should include a version field regardless if we support different versions or not.

common/src/main/java/com/wynntils/services/mapdata/attributes/FullCategoryAttributes.java
kristofbolyai: Okay, it was just an idea. I too dislike using boxed classes, but Optional, ErrorOr or any other wrapper class is too explicit for this kind of data structure. Perhaps this is where Kotlin's "nullability" safety would help..

common/src/main/java/com/wynntils/services/mapdata/MapDataService.java
kristofbolyai: Yes, definitely a weird case..

common/src/main/java/com/wynntils/services/mapdata/type/MapFeature.java
magicus: It's nothing that exists yet. :-) I planned for the future. The idea was that you should allow users to put free-form tags on map features. These need not be hierarchical or systematic like the categories. For instance, "favorite", or "investigate" or "avoid"..?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: That's a valid point. We might want to prescribe a version info for the json files...

magicus: I agree that this decision needs to be fixed now. I just don't understand how we could separate map features that do not have level information from "do not apply" to "we do not know". Let's take the Item Identifier in Llevigar for instance. He has no level information. Should he inherit the level information of Llevigar, since in practice, you should not go there unless you're of sufficient level? Or should he be marked "accessible for all", since every level can in fact interact with him? But then again, in *theory* even a level 1 character can go to Llevigar, it's just a bloody stupid idea.

common/src/main/java/com/wynntils/services/mapdata/type/MapFeature.java
kristofbolyai: Sounds good then!

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: This is 2 separate issues though.

1, The level "code" itself has to support 3 different values: actual integer, "all levels", "unknown".
2, We have to figure out what to put here when implementing the default providers. That is where the inheritance would come in (which I think is appropriate in your example).

This PR, we only basically need to care about the first issue, if we should support it, and if how.

magicus: Or take Light's Secret. It is missing level indicator on the official map. But that is really mostly an outage, it should be marked as level ~70. (Just as Ozoth's Spire). In this case, the absense of level information does not mean "accessible for all", but "we do not know", or even "should have had level information, but it is missing in the source material".

The trouble is, there is no way we can systematically and programmatically differentiate between these various reasons for a map feature missing level information. So the separation is between "has indication of level" and "do not have indication of level" -- and the reason for the former can never be known to us.

The latter state is preferably represented as 0, rather than having an Optional<Integer>, for reasons of simplicity. Since characters start out at level 1, level 0 is perfectly usable as an indicator for "does not have level information". With that said, the field is optional in json files, in which the value 0 is assumed.

common/src/main/java/com/wynntils/services/map/Label.java
magicus: Sure, I can cache this in the constructor, and make clear that the string version is theoretically inferior.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapDecoration.java
magicus: Yes, I agree. And I want a close match between json and internal java representation. But this is something that can never be expressed in json. Kind of special-extra-sauce that only our internal providers can have. So it is more open to extension.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
magicus: It probably mostly was that I was too tired when I got there, and then I started procrastinating going back to fix it, until I got scared of going back to it. :-) I also don't have a very good idea on how to express this in json, nor really what is a proper way to express this. For instance, should you maybe give a point value for the zoom appear/disappear factor, and have a separate "fudge" value to indicate when it should start/stop fading, or should you (as I said above) have two separate start and stop points? The point+offset value means basically the point is the mean value, and the start and stop becomes `point-offset` and `point+offset`.

This has the advantage that if you don't override the offset value, all fades will happen over the same scale factor distance. I think that would be good, but it ultimately depends on this "feels" right when zooming in the map. I'd have to test this with some of the existing map features.

magicus: And finally I worry a bit if I can treat the scale factor as linear. Maybe the offsetting needs to be scaled according to the scale factor as well, so it does not "feel" like the stuff fades in too quickly or too slowly at high (or low) zoom factors.

kristofbolyai: I agree that point+offset sounds good. I also think that it being linear is not a problem. When do you want to test this (or I could ask, when is this zooming format going to be finalized)?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: Perhaps we could just make -1 "accessible for all"?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
magicus: Ideally, I'd like to get this code that is done integrated as soon as possible. That was why this patch just has ALWAYS and NEVER, which are kind of trivial. I dislike long-open PRs, it just results in code rot.

I can't really say when I can spend time to continue fixing this. Right now it feels like the only time I have been able to spend on Wynntils has been on reviews and discussions. If you think this must be done for the first part, then we'll keep this PR open until I can get around to fix it. Or we can merge this as it is, and I can get back with a separate PR with the visibility functionality.

magicus:
To summarize the current status: (perhaps mostly as a worklist for me)

* Label should treat the int value of level better
* json files should have a version field
* We need to decide how to handle unknown/suitable-for-all on level in map attributes
* Visibility for zoom levels is missing; either fix here or as a follow-up

Did I miss something?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
kristofbolyai: I mean, this is not gonna be used yet. So it can be merged, but not "released/used" until it is fixed.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: Let's assume we do that. How should we present this on the map?

kristofbolyai (approved):
If you want to merge this right now, go ahead. But the things we discussed above needs to be fixed before this is used.

kristofbolyai:
@magicus I'll ping you here, in case you did not check Discord. If you did, and just not had time to respond, ignore this.

I am not sure if you wanted to finish up this PR by adding the log scaling zooming yourself, or if I should do it. If you don't have time, I am happy to take it up, but I might need a bit of guidance in the first few commits while I get myself oriented here, otherwise, our ideas might diverge. Anyways, respond when you can, here, or in Discord.

kristofbolyai:
As promised, I'll take over this PR and finish it for you. I hope I won't lost track of all of your ideas, nor mine :).

magicus:
Thanks!

PR #2159, opened 2023-10-02
WIP: Hacks needed to try out the new mapdata system

magicus:
This is really based on #2158. Here are some hacky extensions that will allow you to actually use and play around with the new mapdata system. 

There is also documentation (in the hijacked README file) with what I believe remains to be done, and various design notes (mostly for my own memory).


magicus:
Superseded by #2490.

PR #2162, opened 2023-10-03
feat: Lootrun Beacon Count now shows Red Challenges left

Rafii2198:
image


kristofbolyai:
Please target the 2.0.4 branch, we can live without this for a few days (and that way we don't have to fix the merge..)

Rafii2198:
I think I messed up. Let me redo the PR.

PR #2166, opened 2023-10-03
feat: Fabric loader lists conflicting mods

ryanzhoudev:


Rafii2198:
I think Creative Core is fine since some update and as incompatible versions are not available on 1.20.2 I think there is no need for listing them anymore. As for ImmediatlyFast, I am testing it, right config seems to fix the issue but well it is quite random

kristofbolyai (commented):
Immediately fast can be moved to conflicts imo. It only causes minor text rendering issues, arguably much better than avomod for example.

Can quilt/forge do this kind of conflict handling..?

ryanzhoudev:
> Can quilt/forge do this kind of conflict handling..?

Forge cannot, it only allows specifying dependencies. Quilt can specify the `breaks` section, but does not have a `conflicts` section

kristofbolyai (approved):

PR #2171, opened 2023-10-06
fix: Update class item patterns

ShadowCat117:
character


kristofbolyai (approved):

magicus:
Btw, whatever happened to the icons? They are all black now. Is it due to changes in Wynns original class selection screen, or a bug in Wynntils?

kristofbolyai:
> Btw, whatever happened to the icons? They are all black now. Is it due to changes in Wynns original class selection screen, or a bug in Wynntils?

1.20.2 thing, havent sorted it out yet.

PR #2177, opened 2023-10-08
fix: IF batching on hud text

HalcyonSuoh:


kristofbolyai:
Can you remove IF from the incompatibility lists from quilt/fabric?

HalcyonSuoh:
> Can you remove IF from the incompatibility lists from quilt/fabric?

failed to find IF entry in quilt.mod.json, other than that, sure

kristofbolyai:
> 
> 
> 
> 
> 
> 
> > Can you remove IF from the incompatibility lists from quilt/fabric?
> 
> 
> 
> failed to find IF entry in quilt.mod.json, other than that, sure

Might not be in there.

kristofbolyai (approved):

magicus:
I had not heard about IF before, but they seem to claim good interoperability, so I opened a bug on them to give them chance to fix the problem on their side.

https://github.com/RaphiMC/ImmediatelyFast/issues/126

However, I don't know what the actual problem was 😜 so if anyone wants to add such an explanation to the bug, I think it would be nice.

PR #2182, opened 2023-10-09
feat: Rewrite CommandManager to handle command suggestions and auto-completion better 

kristofbolyai:
This PR started as a simple bugfix (commit https://github.com/Wynntils/Artemis/pull/2182/commits/e31342c6513b6d1cfb5732e966d996295cfe8fa0, which took a long while to debug..). After I got it working, I noticed this implementation of client commands worked even worse than it used to work in 1.19.4. With this rewrite, we are utilizing the full power of the MC command suggestion system, hopefully making our commands easier to use.

The rewrite mostly focused on registering to the server command dispatcher, instead of only registering to our own client command dispatcher, and trying to patch into command suggestions, which is basically impossible because of the numerous ways MC code accesses it.

New:
image

Old, bugfix for autocompleting (partially worse than 1.19.4):
image

Before this PR: Autocomplete not working at all.


common/src/main/java/com/wynntils/core/consumers/commands/CommandManager.java
```diff
@@ -59,60 +67,36 @@ public CommandManager() {
         registerAllCommands();
     }
 
+    @SubscribeEvent
+    public void onCommandsAdded(CommandsAddedEvent event) {
+        for (Command command : commandInstanceSet) {
+            // Wynntils command is special,
+            // it registers every other command as a subcommand
+            if (command instanceof WynntilsCommand wynntilsCommand) {
```
magicus: Maybe we should store wynntilsCommand as a field when registering it, so you don't have to check for it in the for loop?

kristofbolyai: I mean, if you want to, okay. This runs once per server join, basically not related to performance in any way.

kristofbolyai: If this is the only thing that needs changing, I would merge this as I am off for today..

common/src/main/java/com/wynntils/features/commands/FilterAdminCommandsFeature.java
```diff
@@ -43,10 +44,14 @@ public void onCommandPacket(CommandsAddedEvent event) {
 
         RootCommandNode<SharedSuggestionProvider> newRoot = new RootCommandNode<>();
         for (CommandNode<SharedSuggestionProvider> child : root.getChildren()) {
-            if (!FILTERED_COMMANDS.contains(child.getName())) {
-                newRoot.addChild(child);
+            // Only add literal nodes, not argument nodes
```
magicus, line 47: This is the direct children to the root, yes? What are the argument nodes here? I thought it was only possible to have literals as the top-most children.

common/src/main/java/com/wynntils/core/consumers/features/FeatureCommands.java
```diff
@@ -44,4 +45,13 @@ public void discoverCommands(Feature feature) {
             }
         }
     }
+
+    public LiteralCommandNode<CommandSourceStack> getCommandNode() {
+        // Build the command node if it hasn't been built yet
+        if (commandNode == null) {
```
magicus, line 51: I think this feels a bit shaky. What if someone calls this too early? Then the builder will build, and the next `commandNodeBuilder.then` will fail (I assume). 

While slightly more complicated, I suggest adding some kind of finalizing method here, to build the commandNode, and call it in the end of FeatureManager.init().

magicus (commented):
Overall this looks nice. Deleted code is the best code! :)

common/src/main/java/com/wynntils/core/consumers/commands/CommandManager.java
magicus: It is not for performance as much as for readability.

magicus: I had some more comments below. We can merge this tomorrow, no hurry.

common/src/main/java/com/wynntils/features/commands/FilterAdminCommandsFeature.java
kristofbolyai: You can also have argument nodes, which are kind of weird, they don't resolve when trying to calculate autocomplete, freezing the whole process. It is added by creating a new root command node, I am not really sure why. Either way, this should work..

common/src/main/java/com/wynntils/core/consumers/features/FeatureCommands.java
kristofbolyai: Okay, sure.

magicus (approved):

PR #2185, opened 2023-10-11
feat: Add price stat, fix not being able to click items in TM

kristofbolyai:
Review by commit. The PR is actually not so big, just a lot of superclass changes..

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/GameItemAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/GameItemAnnotator.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.annotators.game;
+
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.text.StyledText;
+import com.wynntils.handlers.item.ItemAnnotation;
+import com.wynntils.handlers.item.ItemAnnotator;
+import com.wynntils.models.items.items.game.GameItem;
+import com.wynntils.utils.mc.LoreUtils;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.world.item.ItemStack;
+
+public abstract class GameItemAnnotator implements ItemAnnotator {
+    private static final Pattern PRICE_STR = Pattern.compile("§6Price:");
+
+    // Test suite: https://regexr.com/7lh2b
+    private static final Pattern PRICE_PATTERN = Pattern.compile(
+            "§[67] - (?:§f(?<amount>[\\d,]+) §7x )?§(?:(?:(?:c✖|a✔) §f)|f)(?<price>[\\d,]+)§7²(?: .+)?");
+
+    public abstract GameItem getAnnotation(
+            ItemStack itemStack, StyledText name, List<StyledText> lore, int emeraldPrice);
+
+    @Override
+    public final ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
+        List<StyledText> lore = LoreUtils.getLore(itemStack);
+        int emeraldPrice = parsePrice(lore);
+
+        return getAnnotation(itemStack, name, lore, emeraldPrice);
+    }
+
+    private int parsePrice(List<StyledText> lore) {
+        boolean foundPrice = false;
+        for (StyledText loreLine : lore) {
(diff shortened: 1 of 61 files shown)


kristofbolyai:
@magicus Important for you post-merge reviewing wise, I want your opinion on the item model changes. I am happy to refactor, however this needed to be out, and I found this to be the least intrusive and fitting way to add it.

magicus:
Yeah... I'm not really fond of this. :-/ We're kind of mixing in a very specific aspect (the price) with the very general API of parsers/items (an ItemStack and a name).

Let me look through this PR and the follow-ups more closely, and I'll see if I can suggest an alternative.

magicus:
Also, what was the hurry with this? I see there is a bug fix included, but as far as I can tell, it did not really depend on the rest of the PR..?

kristofbolyai:
> Also, what was the hurry with this? I see there is a bug fix included, but as far as I can tell, it did not really depend on the rest of the PR..?

Price sorting is really important for the TM screen, sadly I forgot about it originally. I also want to use the wynn TM filters too, but those are even harder to get working..

common/src/main/java/com/wynntils/models/items/items/game/EmeraldItem.java
```diff
@@ -12,7 +12,8 @@ public class EmeraldItem extends GameItem implements EmeraldValuedItemProperty {
     private final Supplier<Integer> amountSupplier;
     private final EmeraldUnits unit;
 
-    public EmeraldItem(Supplier<Integer> amountSupplier, EmeraldUnits unit) {
+    public EmeraldItem(int emeraldPrice, Supplier<Integer> amountSupplier, EmeraldUnits unit) {
+        super(emeraldPrice);
```
magicus, line 16: So this just looks very suspicious. In what way can the emeraldPrice be anything other than the amount of this item? And the latter is a supplier, but the former is fixed.

kristofbolyai: Supplier being there is another problem of the wynn tooltip..

There is a "Shady" merchant selling emerald for emerald :), either way, 0 emerald price means an "unpriced" item, basically the item is not sold for emeralds.

common/src/main/java/com/wynntils/models/items/annotators/game/GameItemAnnotator.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public final ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
+        List<StyledText> lore = LoreUtils.getLore(itemStack);
+        int emeraldPrice = parsePrice(lore);
+
+        return getAnnotation(itemStack, name, lore, emeraldPrice);
+    }
+
+    private int parsePrice(List<StyledText> lore) {
```
magicus, line 36: I am deeply worried about this. I spent a lot of time trying to make item annotators performant, and not scan more than absolutely necessary. Now suddenly *all* annotators will scan the entire lore, all of the time. Have you measured the performance hit of this?

magicus: I think the price scanning will need to be separated out of the annotators, and only done when in a trade market GUI which displays prices.

kristofbolyai: Yes, optimized it in a later PR. As expected it was bad, but only memory wise.

magicus: What about the performance?

common/src/main/java/com/wynntils/models/items/items/game/EmeraldItem.java
magicus: I agree that the supplier is a problem. Items should not suddenly change their value.

RIght, the Shady merchant... :-) Ok, fair enough, I agree that emeralds can have a price that is not the face value.

magicus:
My current thought about this is that we should introduce a separate priceAnnotation to the vanilla ItemStack. And only set it if we are parsing a TM screen.

The price, when set in the TM, is an aspect of items that is fully orthogonal to what type/kind it is.

There might be other solutions, but that is what I'm currently thinking.

common/src/main/java/com/wynntils/models/items/annotators/game/GameItemAnnotator.java
kristofbolyai: It was still sub milisecond per annotation.

kristofbolyai:
@magicus proposed a solution in dc

PR #2191, opened 2023-10-12
feat: Parse item prices lazily, add AmountStatProvider and TotalPriceStatProvider

kristofbolyai:
This should be the correct way.

Review by commit. I left notes in commit descriptions.


magicus:
I just had time for a quick glance now, will return to this later. Just an idea: do you think it would be better to let a WynnItem hold a back-reference to its "original" ItemStack? I think it could kind of make sense. And then you'd not need to pass the ItemStack around everywhere. It might also help with the EmeraldItem value calculation...

common/src/main/java/com/wynntils/models/items/WynnItemCache.java
```diff
@@ -14,6 +14,7 @@ public class WynnItemCache {
     public static final String TOOLTIP_KEY = "tooltip";
     public static final String SEARCHED_KEY = "searched";
     public static final String FAVORITE_KEY = "favorite";
+    public static final String EMERALD_PRICE_KEY = "emerald_price";
```
magicus: Maybe just PRICE = "price"? The emerald prefix feels a bit redundant, and a single word makes the key name fit better with the rest.

kristofbolyai:
> I just had time for a quick glance now, will return to this later. Just an idea: do you think it would be better to let a WynnItem hold a back-reference to its "original" ItemStack? I think it could kind of make sense. And then you'd not need to pass the ItemStack around everywhere. It might also help with the EmeraldItem value calculation...
> 
> 
> 
> 

Wouldn't it become kind of circular? You get a WynnItem from an ItemStack, but you also get the ItemStack from WynnItem.

But yes, it is cleaner.

magicus:
... or, once again, perhaps the wynnItemCache to the rescue?

kristofbolyai:
> ... or, once again, perhaps the wynnItemCache to the rescue?

How so...?

magicus:
Yes, it's circular, but that's fine. In some way a WynnItem and the ItemStack has a very intimate "two sides of the same coin" relationship, so it makes sense that you can go both ways. I don't see that as a problem.

However, it is not immediately clear at what level to put the responsibility of maintaining this link. Like, should it be part of an Annotation, and the ItemHandler should set it, or part of a WynnItem, and the ItemModel's responsibility? I think the latter makes more sense, but it will require some refactoring, to inject this in the proper place, or we will need to pass the ItemStack as constructor argument to all WynnItems and that will just be a pain, I think.

magicus:
I meant that we do not need to add a new field to WynnItem but can use the cache. I'm not 100% sure that is a good idea, though. It is not a cache -- we can't restore the value if it is lost (that is true for some of the other fields as well, though...). But then again, the cache will not be lost (unless the entire ItemStack gets lost). So it is perhaps a bit mis-named, it's more of a general key->value data store associated with each WynnItem.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/AmountStatProvider.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.wynntils.core.components.Models;
+import com.wynntils.models.items.WynnItem;
+import com.wynntils.models.items.WynnItemCache;
+import com.wynntils.models.trademarket.type.TradeMarketPriceInfo;
+import com.wynntils.services.itemfilter.type.ItemStatProvider;
+import java.util.List;
+import net.minecraft.world.item.ItemStack;
+
+public class AmountStatProvider extends ItemStatProvider<Integer> {
```
magicus: Maybe `TradeAmountStatProvider`?

kristofbolyai:
So, do you think that I should change WynnItem to include ItemStack here, or is this acceptable and you/i can add it when we need it next time?

magicus:
I think you should fix it here. Passing the ItemStack around the stat providers is also not ideal.

kristofbolyai:
The itemstack should be stored now.

common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -276,6 +278,12 @@ private ItemAnnotation calculateAnnotation(ItemStack itemStack, StyledText name)
 
         if (annotation == null) return null;
 
+        // Store the itemstack in the data for later use
+        if (annotation instanceof WynnItem wynnItem) {
+            wynnItem.getData().store(WynnItemData.ITEMSTACK_KEY, itemStack);
+        } else {
```
magicus: You can remove the else...

kristofbolyai: Oops. Left over debug code.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/PriceStatProvider.java
```diff
@@ -4,18 +4,27 @@
...
+import com.wynntils.models.trademarket.type.TradeMarketPriceInfo;
 import com.wynntils.services.itemfilter.type.ItemStatProvider;
 import java.util.List;
 
 public class PriceStatProvider extends ItemStatProvider<Integer> {
     @Override
     public List<Integer> getValue(WynnItem wynnItem) {
-        if (wynnItem instanceof GameItem gameItem && gameItem.hasEmeraldPrice()) {
-            return List.of(gameItem.getEmeraldPrice());
+        TradeMarketPriceInfo priceInfo = wynnItem.getData().getOrCalculate(WynnItemData.EMERALD_PRICE_KEY, () -> {
```
magicus, line 17: I don't think you need the block around the lambda here..?

kristofbolyai: You do, since you need to get the itemstack.

common/src/main/java/com/wynntils/models/trademarket/TradeMarketModel.java
```diff
@@ -72,4 +85,32 @@ public void setPresetFilter(int presetId, String filter) {
         presetFilters.get().put(presetId, filter);
         presetFilters.touched();
     }
+
+    public TradeMarketPriceInfo calculateItemPriceInfo(ItemStack itemStack) {
+        StyledText priceLine = LoreUtils.getLoreLine(itemStack, TRADE_MARKET_PRICE_LINE);
+
+        if (priceLine == null || !priceLine.matches(PRICE_STR)) {
+            WynntilsMod.warn("Trade Market item had an unexpected price line: " + priceLine);
```
magicus, line 93: This warning might just as well mean "we're not in the trade market"..? Or is there some mechanism I am missing that means we should only end up here for items on the TM?

kristofbolyai: Theoretically this code supports calculating price in other guis from any other models/features, if needed

common/src/main/java/com/wynntils/services/itemfilter/statproviders/PriceStatProvider.java
magicus: I don't get it. Why can't you just do `() -> Models.TradeMarket.calculateItemPriceInfo(wynnItem.getData().get(WynnItemData.ITEMSTACK_KEY))`?

kristofbolyai: Oh, yes, I can. But it arguably looks worse with spotless...

image

common/src/main/java/com/wynntils/models/trademarket/TradeMarketModel.java
magicus: Hm, what I mean is that do we not risk spamming the console with "Trade Market item had an unexpected price line: ..." if we try to access price stats outside the TM? As in, maybe we should not print a warning but just return.

If the first line match but the second does not, by all means -- print a warning! Then the price info has changed and our regexp fails. But if there is no "Price"-line, then it is not really an error.

kristofbolyai: Hm, I see what you mean. Code that is only supposed to run in the trade market (hence why it is in the TM model) got mixed up with some general purpose stuff... I can remove the log if you want, maybe not the best solution though.

magicus (approved):
This is sooo much better! I'm also glad this resulted in the change from "cache" to "data"; something has been nagging me about that "cache" since a long time but I did not really realize it until now. Now it is very clear that if we want to add "sweeping" fields on all WynnItems regardless of their place in the item hierarchy, it should go in as tags in the data storage.

If you address my concerns about spamming the log and simplifying the lambda (or explain to me like a 4yo why it's not possible), this is good to go.

And thanks again for doing the retroactive cleanup.

common/src/main/java/com/wynntils/models/trademarket/TradeMarketModel.java
magicus: I think it is an okay solution to just remove the log. If you try to filter your items on price outside TM, you will have to take the hit that we try to read the first line of the lore of all items in the container, just to be able to return an EMPTY price info.

I take it the stat providers are not actively doing anything unless the user explicitly tries to filter on them?

kristofbolyai: > I take it the stat providers are not actively doing anything unless the user explicitly tries to filter on them?

Yes, that is the case. These are only parsed when they are filtered for, on items that are already filtered.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/PriceStatProvider.java
magicus: Ok, fine, then it's just a matter of taste. I think the code above looks better but I accept that you prefer the current version. As long as the reason for the current code was not "oh, i did more stuff before, but I refactored and forget to simplify".

common/src/main/java/com/wynntils/models/trademarket/TradeMarketModel.java
kristofbolyai: Hm, wait a bit! I just read the code, and seems like I removed all "general" use case scenarios from it, and only works in the TM. That means that every TM item is expected to have that proper lore, so I think the warning is appropriate in both cases.

magicus: Ok, if this is the case, then the code is fine.

PR #2192, opened 2023-10-12
feat: Add support for Wynn's default sorting button in Trade Market screen

kristofbolyai:
image


common/src/main/java/com/wynntils/screens/trademarket/TradeMarketSearchResultHolder.java
```diff
@@ -73,6 +82,12 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
         // Reset the empty item count,
         // set content packets mean we are on a new page
         pageItemCount = 0;
+
+        if (event.getItems().size() >= SORT_MODE_SLOT) {
```
ShadowCat117, line 86: Why is the if needed here, isn't the results screen always going to have more than 53 items?

kristofbolyai: In theory not needed, just defensive code

ShadowCat117 (approved):
Hopefully less complaints about the screen after this :)

PR #2195, opened 2023-10-15
feat: New chat tabs screen

### LINKED ISSUE #1905: Chat tab screen needs to be ported to the grid layout
See #1870, which was closed when POI screen was implemented.

devmwang:
- Refactors chat tab screen to use grid system (Closes #1905)
- Add list of existing chat tabs (attempts to center vertically)
- Move delete buttons to chat tab list
- New icon for chat tab add button (renamed to chat tab settings button)
- Create new checkbox with scrolling text for constrained checkbox spaces

https://github.com/Wynntils/Artemis/assets/41971562/1e864f3c-1228-4a48-aa6c-868a9185f93c


ShadowCat117:
Might need a bit of tweaking, gets quite messy on gui scale 4

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
```diff
@@ -231,41 +254,50 @@ public void doRender(GuiGraphics guiGraphics, int mouseX, int mouseY, float part
 
         PoseStack poseStack = guiGraphics.pose();
 
+        dividedWidth = this.width / GRID_DIVISIONS;
+        dividedHeight = this.height / GRID_DIVISIONS;
```
ShadowCat117: I don't think these are required, doing it in the doInit should be enough

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabSettingsButton.java
```diff
@@ -18,8 +18,8 @@
 import net.minecraft.client.gui.GuiGraphics;
 import net.minecraft.network.chat.Component;
 
-public class ChatTabAddButton extends WynntilsButton {
-    public ChatTabAddButton(int x, int y, int width, int height) {
+public class ChatTabSettingsButton extends WynntilsButton {
+    public ChatTabSettingsButton(int x, int y, int width, int height) {
         super(x, y, width, height, Component.literal("Chat Tab Add Button"));
```
ShadowCat117:
```suggestion
        super(x, y, width, height, Component.literal("Chat Tab Settings Button"));
```

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabSettingsButton.java
```diff
@@ -33,13 +33,13 @@ public void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, float
         FontRenderer.getInstance()
                 .renderAlignedTextInBox(
                         poseStack,
-                        StyledText.fromString("+"),
+                        StyledText.fromString("\uD83D\uDDD7"), // Tabs Icon
```
ShadowCat117: Note for other devs, should we have a preference on writing unicode like this vs the actual character? If you copy the character from somewhere like https://www.compart.com/en/unicode/ and paste as plain text into Intellij it works fine. I try to use the actual character as it's more readable imo but I know we do it like this in a lot of other places

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
```diff
@@ -0,0 +1,132 @@
...
+
+        deleteButton.render(guiGraphics, mouseX, mouseY, partialTick);
+        moveUpButton.render(guiGraphics, mouseX, mouseY, partialTick);
+        moveDownButton.render(guiGraphics, mouseX, mouseY, partialTick);
+    }
+
+    @Override
+    protected void updateWidgetNarration(NarrationElementOutput narrationElementOutput) {}
+
+    private void incrementChatTabIndex() {
```
ShadowCat117: These 2 methods can probably be merged into 1 `updateChatTabIndex` with a parameter direction being 1 or -1

ShadowCat117: In fact, they can probably just be put into `setChatTabIndex` as something like `int index = Services.ChatTab.getTabIndex(chatTab) + direction;`

ShadowCat117:
Aside from a few small things and the gui scaling, very good :)

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
```diff
@@ -0,0 +1,132 @@
...
+            ChatTab chatTab,
+            float gridDivisions,
+            ChatTabEditingScreen parent) {
+        super((int) x, (int) y, width, height, Component.literal(chatTab.getName()));
+
+        this.chatTab = chatTab;
+        this.gridDivisions = gridDivisions;
+        this.parent = parent;
+
+        this.deleteButton = new Button.Builder(
```
Mcrtin, line 50: instead of rendering and listening for them individually, you can add them as renderable widgets. Then all the events get called without you having to worry about that and they also get rendered (maybe you have to call super in `renderWidget` for that).

kristofbolyai: You are in the widget itself..?

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
```diff
@@ -0,0 +1,132 @@
...
+                        Component.translatable("screens.wynntils.chatTabsGui.delete")
+                                .withStyle(ChatFormatting.DARK_RED),
+                        (button) -> Services.ChatTab.removeTab(chatTab))
+                .pos((int) (this.getX() + (this.width / this.gridDivisions * 17)), this.getY() + (this.height / 2) - 10)
```
Mcrtin, line 57: this calculation is a bit difficult to follow. can you extract some parts of this to vars (like `this.width / this.gridDivisions`)

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
kristofbolyai: Theoretically it could be passed up but we have never done it that way

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
kristofbolyai: arguably that would result in weirder code

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
kristofbolyai: Yes, these never change render time, but only when re-inited.

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabSettingsButton.java
kristofbolyai: Iirc, general standard is to type it out whenever it is displayed, etc, and use unicode when used in patterns/string matching.

kristofbolyai:
I can see an issue from the video:

seemingly creating/deleting (and maybe even moving) tabs doesn't focus them properly and leaves the tabs in a weird state.

Also, I think some colors need to be changed. Delete should be brighter red, and something needs to be done with the save buttons, but dark green is not really fitting/is hard to see.

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
Mcrtin: oh that's only available in screens, isn't it? ConfigTile does it like this too, but Minecrafts screen system isn't designed for that. Maybe we should start doing something about it like only using such classes only as builders for the actual widgets cuz in implementations like this stuff gets overlooked like forwarding the other events.

kristofbolyai: it doesn't get overlooked if it is tested/coded properly? not every event needs to be passed down if the widget doesnt need it

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private void decrementChatTabIndex() {
+        setChatTabIndex(Services.ChatTab.getTabIndex(chatTab) - 1);
+    }
+
+    private void setChatTabIndex(int index) {
+        Services.ChatTab.removeTab(chatTab);
+
+        int constrainedIndex = Math.max(0, Math.min(index, Services.ChatTab.getTabCount()));
```
Mcrtin: you can use clamp from MathUtils

Mcrtin:
what happens when you press tab in the screen? do you get an error?

ShadowCat117:
> what happens when you press tab in the screen? do you get an error?

Pressing tab cycles through the different input widgets

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
ryanzhoudev: Is init called when the window is resized? I'm pretty sure all the other screens also recalculate this in render

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabSettingsButton.java
ryanzhoudev: We tried copying the actual character here, Intellij just kept giving us the code though.

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
kristofbolyai: it is, since widgets also need to be readded at correct locations

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabSettingsButton.java
ShadowCat117: I can get it using this if I copy the actual character from here https://www.compart.com/en/unicode/U+1F5D7
image

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
ShadowCat117: I didn't recalculate it in the poi manager screen and I'm pretty sure it works

kristofbolyai:
Sorry, I still don't have time to fully review this. I will try to do it later tonight though :)

common/src/main/java/com/wynntils/screens/base/widgets/WynntilsCheckbox.java
```diff
@@ -0,0 +1,74 @@
...
+            int maxTextWidth,
+            CustomColor color) {
+        super(x, y, width, height, message, selected);
+        this.maxTextWidth = maxTextWidth;
+        this.color = color.asInt();
+    }
+
+    @Override
+    public void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick) {
+        Minecraft minecraft = Minecraft.getInstance();
```
kristofbolyai:
```suggestion
        Minecraft minecraft = McUtils.mc();
```

common/src/main/java/com/wynntils/screens/base/widgets/WynntilsCheckbox.java
```diff
@@ -0,0 +1,74 @@
...
+        RenderSystem.enableDepthTest();
+        Font font = minecraft.font;
```
kristofbolyai:
```suggestion
        Font font = FontRenderer.getInstance().getFont();
```

(don't forget the import)

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
```diff
@@ -416,4 +446,32 @@ public TextInputBoxWidget getFocusedTextInput() {
     public void setFocusedTextInput(TextInputBoxWidget focusedTextInput) {
         this.focusedTextInput = focusedTextInput;
     }
+
+    public void reloadChatTabsWidgets() {
+        chatTabsWidgets.clear();
```
kristofbolyai: usually reassigning a new collection is safer in terms of concurrent modifications (you can never be sure other mods don't change the "single threadedness" of rendering)

common/src/main/java/com/wynntils/screens/chattabs/ChatTabEditingScreen.java
```diff
@@ -416,4 +446,32 @@ public TextInputBoxWidget getFocusedTextInput() {
...
+        List<ChatTab> chatTabs = new ArrayList<>(Services.ChatTab.getChatTabs());
```
kristofbolyai: Do you need to duplicate the list? I don't see a reason not to just reference the service's list.

common/src/main/java/com/wynntils/screens/chattabs/widgets/ChatTabsWidget.java
```diff
@@ -0,0 +1,125 @@
...
+                chatTab.getCustomRegexString());
+        Services.ChatTab.addTab(newIndex, newChatTab);
+    }
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (deleteButton.mouseClicked(mouseX, mouseY, button)
+                || moveUpButton.mouseClicked(mouseX, mouseY, button)
+                || moveDownButton.mouseClicked(mouseX, mouseY, button)) {
+            Managers.TickScheduler.scheduleLater(parent::reloadChatTabsWidgets, 2);
```
kristofbolyai: Why schedule it later?

kristofbolyai:
Now that I reviewed the code, let's get to some design suggestions:

1, If we are editing a tab, can you perhaps render the name yellow/green? It would make it a bit more obvious that this is not a new tab.


Also, I don't fully understand what our modified checkbox does. It does not seem to be focused with tab, so I am not sure how it is any better, or what it is meant to do.

ryanzhoudev:
> Also, I don't fully understand what our modified checkbox does. It does not seem to be focused with tab, so I am not sure how it is any better, or what it is meant to do.

The WynntilsCheckbox renders scrollable names if they don't fit, not sure why the vanilla ones don't even though buttons do

kristofbolyai:
> Might need a bit of tweaking, gets quite messy on gui scale 4
> 
> 

Was this fixed/improved?

kristofbolyai:
Okay, so basically the code is alright now. I just want to change the UI still, a little bit.

How about renaming the "Save" button to "Add"?

Also, I want to have some kind of visual so the user can see that they are editing a "new" tab (atm it is only implicitly applied by having no yellow tab names). I don't have a concrete idea for this though.


After these 2 fixes, I think this should be fine :)

magicus:
> Also, I want to have some kind of visual so the user can see that they are editing a "new" tab (atm it is only implicitly applied by having no yellow tab names). I don't have a concrete idea for this though.

Create a heading and set it to "Create new tab" if creating new, and to "Editing tab 'Foo'" if editing an existing tab?

ShadowCat117:
> Was this fixed/improved?

Yes, looks much cleaner on all gui scales

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1751,14 +1751,20 @@
...
   "screens.wynntils.characterSelection.playButton.description": "You can also double click to play",
   "screens.wynntils.characterSelection.playButton.play": "[>] Play with the selected character",
+  "screens.wynntils.chatTabsGui.add": "Add",
+  "screens.wynntils.chatTabsGui.addAndClose": "Add and Close",
   "screens.wynntils.chatTabsGui.autoCommand": "Auto Command",
   "screens.wynntils.chatTabsGui.cancel": "Cancel",
   "screens.wynntils.chatTabsGui.consuming": "Consuming",
+  "screens.wynntils.chatTabsGui.create": "Create New Tab",
   "screens.wynntils.chatTabsGui.delete": "Delete",
+  "screens.wynntils.chatTabsGui.edit": "Editing Tab ",
```
kristofbolyai: Just a small nit, but you could use format strings.

```suggestion
  "screens.wynntils.chatTabsGui.edit": "Editing Tab %s",
```

and pass the string in the appropriate i18n.get or component.translatable call

kristofbolyai (approved):
Looks good to me! Thank you for working on this!

kristofbolyai:
I will wait a bit if anyone else wants to review this, if not, I'll merge.

ShadowCat117 (approved):
Looks much better now!

PR #2198, opened 2023-10-20
fix: Process not fully complete scoreboards, fix fallback mapping

### LINKED ISSUE #2196: Scoreboard field in lootruns (timer, challenges, etc...) does not update
**Version**
0.0.4-beta.35 and 0.0.3-alpha.380

**Brief Description**
The right side scoreboard UI does not update while there is a mission active (and possibly because the guild's weekly objective also appears)

Moreover, the latest beacon does not disappear upon completion.

**Steps to Reproduce**
1) have a guild weekly objective
2) start a mission in a lootrun

**Expected Behaviour**
the side scoreboard updates as usual (timer, number of challenges, etc...)

**Actual Behaviour**
the side scoreboard will not update until the objective of the mission is complete.
the latest beacon does not disappear upon completion.

kristofbolyai:
The code focuses on two fixes:
1, Handling parsing the scoreboard when (the last) element is not fully visible. Sometimes Wynn doesn't fully reset the scoreboard to a valid state, usually when content can't fit, but tries to show half of it. This version of the handler should still only parse valid states, but allow such rare cases to happen (which are not so rare in 2.0.4 because of lootrun missions).
2, The scoreboard handler assumed a 1 to 1 relation between scoreboard parts and segments, but the fallback part breaks this contract, since there can be more than 1 fallback parts (although very rarely). This was not actually an issue here, but fixing it was still logical.

Fixes #2196

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
@@ -287,15 +287,15 @@ public final class ScoreboardHandler extends Handler {
         }
 
         // Handle segment removals
-        for (Pair<ScoreboardPart, ScoreboardSegment> pair : oldSegments) {
+        for (Pair<ScoreboardPart, ScoreboardSegment> oldPair : oldSegments) {
             // Special case for the fallback part, don't call onSegmentRemove
-            if (pair.key() == FALLBACK_SCOREBOARD_PART) continue;
+            if (oldPair.key() == FALLBACK_SCOREBOARD_PART) continue;
 
-            Optional<Pair<ScoreboardPart, ScoreboardSegment>> oldSegmentOpt = scoreboardSegments.stream()
-                    .filter(oldPair -> oldPair.key() == pair.key())
+            Optional<Pair<ScoreboardPart, ScoreboardSegment>> segmentOpt = scoreboardSegments.stream()
+                    .filter(pair -> pair.key() == oldPair.key())
                     .findFirst();
-            if (oldSegmentOpt.isEmpty()) {
-                pair.key().onSegmentRemove(pair.value());
+            if (segmentOpt.isEmpty()) {
+                oldPair.key().onSegmentRemove(oldPair.value());
             }
         }
 
--- a/common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
@@ -4,7 +4,7 @@
  */
 package com.wynntils.handlers.scoreboard;
 
-import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -172,26 +172,38 @@ private boolean isScoreboardValid(List<ScoreboardLine> reconstructedScoreboard)
...
                 .equals(StyledText.fromString("À"))) {
-            return false;
+            // We did not find a new line at the start, so the scoreboard is invalid
+            return List.of();
         }
 
         // 4. Check for segment correctness
+        //    There are 2 error cases here:
+        //       - Fatal error: We find info that makes the current scoreboard invalid
+        //       - "Valid" error: We find an error, but it only makes the current scoreboard invalid, not the whole
```
ShadowCat117: What does the "not the whole" mean, is this meant to be it only makes the current segment invalid, not the whole scoreboard?

kristofbolyai: fixed the comment

ShadowCat117 (approved):
Looks fine to me

common/src/main/java/com/wynntils/handlers/scoreboard/ScoreboardHandler.java
```diff
@@ -172,26 +172,38 @@ private boolean isScoreboardValid(List<ScoreboardLine> reconstructedScoreboard)
...
                 .equals(StyledText.fromString("À"))) {
-            return false;
+            // We did not find a new line at the start, so the scoreboard is invalid
+            return List.of();
         }
 
         // 4. Check for segment correctness
+        //    There are 2 error cases here:
+        //       - Fatal error: We find info that makes the current scoreboard invalid
+        //       - "Valid" error: We find an error, but it only makes the current segment invalid, not the scoreboard
```
magicus, line 182: So what happens if we have a truncated part? Is it just removed from the scoreboard, or do we try to use the part that we were given? (I'm not fully following the code here...)

kristofbolyai: If we think it's a half part, we hide it, and dont process it. Usually after a scoreboard reset (/class) it doesn't get shown by wynn either. So I think we are fine.

magicus: So these half parts, they are just temporary glitches..?

kristofbolyai: Basically, almost like if Wynn doesn't want to intentionally do it, and is just a bug.

magicus: Ok. Sounds fine then.

PR #2200, opened 2023-10-20
fix: Disable editing of provided pois

ShadowCat117:


common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -417,7 +417,7 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
                     McUtils.mc().setScreen(PoiCreationScreen.create(this, new PoiLocation(gameX, null, gameZ)));
                 }
             } else if (KeyboardUtils.isAltDown()) {
-                if (hovered instanceof CustomPoi customPoi) {
+                if (hovered instanceof CustomPoi customPoi && !Services.Poi.isPoiProvided(customPoi)) {
```
ShadowCat117, line 420: This wasn't causing a crash but I think it's still worth including

kristofbolyai (approved):
Good temporary fix until the new map provider is out.

PR #2203, opened 2023-10-21
feat: Make the content book much faster

kristofbolyai:
Review by commit, each commit has it's reasoning/description. The new experience is much better and hopefully it does not break mysteriously.. 

I don't think the speedup needs to be measured, just compare them.

New version speed: 
[Kooha-2023-10-21-23-27-06.webm](https://github.com/Wynntils/Artemis/assets/49001742/739c9ec5-c949-45db-808b-725b94efd492)

Old version speed:
[Kooha-2023-10-21-23-32-32.webm](https://github.com/Wynntils/Artemis/assets/49001742/86f40284-12b2-4142-835e-35ca28e4b002)

### DIFF
--- a/common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
@@ -98,6 +98,9 @@ public class WynntilsContentBookFeature extends Feature {
     @Persisted
     public final Config<Boolean> displayOverallProgress = new Config<>(true);
 
+    @Persisted
+    public final Config<Boolean> cancelAllQueriesOnScreenClose = new Config<>(true);
+
     @SubscribeEvent
     public void onUseItem(UseItemEvent event) {
         if (McUtils.player().isShiftKeyDown() || !replaceWynncraftContentBook.get()) return;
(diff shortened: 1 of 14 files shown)


kristofbolyai:
This PR also superseds #2167, that implementation would be too specific..

ryanzhoudev (approved):

kristofbolyai:
Added a few new commits, they also have their own description. Basically this should make the content book queries working as expected without queries getting stuck in queues, getting added to queues too many times, etc. These issues were the actual problem before, not the slightly longer query times..

ShadowCat117 (approved):

ShadowCat117:
Oh also, not sure if you want to add it in here as it doesn't really need its own pr, the custom trade market screen is missing the `@ConfigCategory(Category.TRADEMARKET)` unless you have a tm related pr soon it can be done in

kristofbolyai:
I will fix it separately

magicus:
Nice! The backwards scrolling selection turned out to be much simpler than I anticipated. It should really have been there from day 1, but I could not really figure out how to do it efficiently.

kristofbolyai:
> Nice! The backwards scrolling selection turned out to be much simpler than I anticipated. It should really have been there from day 1, but I could not really figure out how to do it efficiently.

Yes, scrolling was easy. But the support for it in container queries was much harder.. (I already coded this a long time ago, but it wasn't working properly, but this time I managed to tackle it.)

magicus:
Ok, I see, the problem was that scrolling backwards required proper setslot support?

kristofbolyai:
> Ok, I see, the problem was that scrolling backwards required proper setslot support?

Yes, when scrolling backwards, there is no set content.

PR #2204, opened 2023-10-24
fix: Trade market buttons break with chat tabs

ryanzhoudev:
Chat tabs inject/modify at the location where the user presses enter and the game preps/parses the message. Trade market buttons now work when any tab is selected. Wynnic/Gavellian translations while using chat tabs is not affected.


common/src/main/java/com/wynntils/utils/mc/McUtils.java
```diff
@@ -104,7 +105,31 @@ public static void sendCommand(String command) {
...
+     */
+    public static void sendChat(String message) {
+        mc().getConnection().sendChat(message);
+    }
+
+    /**
+     * Simulates the player typing a message in chat. Respects ChatTabFeature settings.
+     * @param message The message to send.
+     */
+    public static void sendChatAsPlayer(String message) {
```
kristofbolyai: McUtils methods should absolutely not have any knowledge about Services or any other component types.

magicus: Agreed. It seems we have created a bit of a messy system for ourselves, if trade market, chat tabs and wynnic transcriptions all are intermingled. Ideally, we should try to tease them apart.

common/src/main/java/com/wynntils/mc/event/ChatScreenSendEvent.java
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+
+@Cancelable
+public class ChatScreenSendEvent extends Event {
```
kristofbolyai, line 11: We refer to dialogue windows as "chat screen". Why not just call it ChatSendEvent?

kristofbolyai (commented):
I assume this would also improve mod compatibility. There is something I can't really describe atm bugging me about this. Give me a moment.

common/src/main/java/com/wynntils/mc/event/ChatScreenSendEvent.java
ryanzhoudev: ChatSendEvent and ChatSentEvent? That seems kind of suspicious and a very easy way to cause problems in the future...

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
```diff
@@ -221,4 +221,26 @@ private boolean matchMessageFromEvent(ChatTab chatTab, ClientsideMessageEvent ev
 
         return event.getOriginalStyledText().matches(regex.get());
     }
+
+    /**
+     * Sends a chat message respecting chat tab autocommand settings.
+     * If no chat tab is actively selected, the message will be sent normally.
+     * @param message The message to send.
+     */
+    public void sendChat(String message) {
```
magicus, line 230: So this is like "send message but prefix with auto command if present"? Hm...

I think the name is a bit vague and generic.

common/src/main/java/com/wynntils/features/chat/InputTranscriptionFeature.java
```diff
@@ -118,22 +120,20 @@ public void onChatScreenKeyTyped(ChatScreenKeyTypedEvent event) {
...
+        updatedMessage = updatedMessage.substring(0, Math.min(updatedMessage.length(), MAX_CHAT_LENGTH));
+        if (updatedMessage.equals(message)) return; // No changes, we can pretend nothing happened.
 
-            if (!updatedMessage.equals(message)) {
-                event.setCanceled(true);
-                McUtils.sendChat(updatedMessage);
-            }
-        }
+        event.setCanceled(true);
+        Services.ChatTab.sendChat(updatedMessage);
```
magicus, line 136: So this will add a prefix, if present? I still think this interdependency is convoluted.

common/src/main/java/com/wynntils/mc/event/ChatScreenSendEvent.java
magicus: What would be the difference between the two events?

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
ryanzhoudev: Yes, it sends with prefix if there is one. But I think the name is fine, it has to be called via Services.ChatTab.sendChat() anyway, so it is pretty clear it is related to the chat tab prefix stuff.

common/src/main/java/com/wynntils/features/chat/InputTranscriptionFeature.java
ryanzhoudev: Yes. It would be a bit strange if suddenly the chat tabs were not respected just because the user decided to type in Wynnic. I am not sure of a way to make this less interdependent, hooking into the native chat-sending method isn't the best either since we can't control if the prefix is added or not and it will always be added

magicus:
Let me have a second look at this in the weekend.

magicus (approved):
This is fine. I like how the logic moved to the ChatTabService. We just need to keep in mind going forward that there are two different ways of sending chat, but the comments will (hopefully) help.

I also apologize for the late review.

PR #2207, opened 2023-10-24
refactor: Grid screens use abstract WynntilsGridLayoutScreen

ryanzhoudev:
- Grid layout screens extend WynntilsGridLayoutScreen to reduce duplication
- Changed arrows to be consistent on party member widgets in PartyManagementScreen
- Fixed buttons not clicking on all resolutions for ChatTabEditingScreen


ryanzhoudev:
PoiCreationScreen technically is also a grid layout style one, but already extends AbstractMapScreen. If everyone is on board with it, maybe I can refactor the entire WynntilsScreen to be grid instead? If not, this is fine. I will also use the abstract GridLayout one for the upcoming skill point loadout menu.

ShadowCat117:
PoiManagementScreen and IconFilterScreen are both grid too btw

ryanzhoudev:
Oh good to know thanks, I'll do those in a few hours

common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    protected void doInit() {
+        dividedHeight = this.height / GRID_DIVISIONS;
+        dividedWidth = this.width / GRID_DIVISIONS;
+    }
+
+    @Override
+    public boolean charTyped(char codePoint, int modifiers) {
+        return (focusedTextInput != null && focusedTextInput.charTyped(codePoint, modifiers))
```
kristofbolyai: Shouldn't it be `WynntilsScreen` that handles implementing `TextboxScreen` + the input methods that need to be overwritten?

ryanzhoudev: I don't think so, there are a lot of WynntilsScreens that don't have anything to do with textboxes

ShadowCat117:
> PoiCreationScreen technically is also a grid layout style one, but already extends AbstractMapScreen. If everyone is on board with it, maybe I can refactor the entire WynntilsScreen to be grid instead? If not, this is fine. I will also use the abstract GridLayout one for the upcoming skill point loadout menu.

I think this could be worth doing. In my opinion a lot of the screens, especially the ones recreated from legacy, could do with a redesign and the grid layout was definitely nice for when I did the poi manager

common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
kristofbolyai: It's still weird to have the code handling this in this class.. Even if a lot of screens don't use it in WynntilScreen, it still could be there though? Or maybe a new superclass, `WynntilsTextboxScreen`? I see no point in making a new class though, if WynntilsScreen can handle it. Of  course a few of the overriden methods would need to be adjusted to call super properly.

kristofbolyai:
> PoiCreationScreen technically is also a grid layout style one, but already extends AbstractMapScreen. If everyone is on board with it, maybe I can refactor the entire WynntilsScreen to be grid instead? If not, this is fine. I will also use the abstract GridLayout one for the upcoming skill point loadout menu.

Although useful, I wouldn't want to lock ourselves into always using grids, there are a few cases where it is not strictly necessary or even not needed at all. Having a good superclass for easy use of grids is nice, but it is definitely not the only we can/should write screens.

common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
kristofbolyai: If we could solve this comment, then we can probably merge this, I don't see any issues otherwise.

common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
```diff
@@ -0,0 +1,82 @@
...
+                if (children().get(i) instanceof TextInputBoxWidget textInputBoxWidget) {
+                    setFocusedTextInput(textInputBoxWidget);
+                    return true;
+                }
+            }
+        }
+
+        return (getFocusedTextInput() != null && getFocusedTextInput().keyPressed(keyCode, scanCode, modifiers))
+                || super.keyPressed(keyCode, scanCode, modifiers);
+    }
```
kristofbolyai: Well this code is still too specific for the class, as this is not a "grid" only implementation. Perhaps move this to WynntilsScreen too?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -390,12 +385,12 @@ protected void doInit() {
...
     @Override
     public void doRender(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick) {
-        PoseStack poseStack = guiGraphics.pose();
         super.doRender(guiGraphics, mouseX, mouseY, partialTick);
+        PoseStack poseStack = guiGraphics.pose();
         renderScrollButton(poseStack);
 
         // Uncomment when editing UI
-        //        RenderUtils.renderDebugGrid(poseStack, GRID_DIVISIONS, dividedWidth, dividedHeight);
+        // RenderUtils.renderDebugGrid(poseStack, GRID_DIVISIONS, dividedWidth, dividedHeight);
```
kristofbolyai: Perhaps could be a method in WynntilsGridScreen, and we could remove these comments.

ryanzhoudev:
At this point the screen really doesn't do much... the only notable thing is doInit() does some division and dumps the result into protected variables.

ShadowCat117:
> At this point the screen really doesn't do much... the only notable thing is doInit() does some division and dumps the result into protected variables.

If that's the case, I think it could be worth just including these in the base WynntilsScreen and then if a screen wants to use the grid layout it can.

kristofbolyai:
> At this point the screen really doesn't do much... the only notable thing is doInit() does some division and dumps the result into protected variables.

I think this is still a good refactor since now we handle the text input widgets in a superclass, and don't duplicate the code. As for the grid screens, I can see a few helper methods added for calculating the positions/sizes in the grid context, but that could be done later.

kristofbolyai (approved):

PR #2208, opened 2023-10-25
fix: Fix some specific lootrun tasks not getting parsed in Sky (and MH?)

kristofbolyai:
Seems like Sky has some weird cases..


kristofbolyai:
Dump of Sky data, which I used to fix this: https://gist.github.com/kristofbolyai/aa319237a0bd890145ef677345e9db9e

PR #2209, opened 2023-10-25
[ci skip] chore: small fixes to comments and features

kristofbolyai:
Some fixes that are not worth pushing a new build or creating a new PR..

### DIFF
--- a/common/src/main/java/com/wynntils/features/ui/CustomTradeMarketResultScreenFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/CustomTradeMarketResultScreenFeature.java
@@ -7,7 +7,9 @@ package com.wynntils.features.ui;
 import com.wynntils.core.components.Models;
 import com.wynntils.core.consumers.features.Feature;
 import com.wynntils.core.persisted.Persisted;
+import com.wynntils.core.persisted.config.Category;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/features/ui/CustomTradeMarketResultScreenFeature.java
```diff
@@ -23,6 +25,7 @@
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
+@ConfigCategory(Category.UI)
```
ShadowCat117, line 28: Not trade market?

kristofbolyai: Well all custom screens are here, as well as the feature itself is in this package

ShadowCat117: True, just most people I've seen searching to disable it have expected it to be in trade market category

kristofbolyai: Kind of have to rework categories either way..

PR #2213, opened 2023-10-27
fix: Fix custom quick jumps not working

ShadowCat117:
This will unfortunately make them slightly slower however it seems to be the best fix.

~~Issue was first reported on Discord [here](https://discord.com/channels/394189072635133952/1023381446733352990/1163569779072848034) but at the time I couldn't recreate it however today it started to happen and it would successfully call `clickOnSlot` for the next/previous buttons but they wouldn't actually be clicked.~~

Original report unrelated and has since resolved itself, this issue likely comes from 2.0.4 hotfix 2

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
@@ -123,7 +123,9 @@ public class CustomBankQuickJumpsFeature extends Feature {
         if (pageDestination > lastPage) {
             quickJumping = false;
             pageDestination = currentPage;
-        } else if (pageDestination != currentPage && ((e.getSlot() == NEXT_PAGE_SLOT && pageDestination > currentPage) || (e.getSlot() == PREVIOUS_PAGE_SLOT && pageDestination < currentPage))) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/inventory/CustomBankQuickJumpsFeature.java
```diff
@@ -183,6 +172,17 @@ public void onContainerSetEvent(ContainerSetContentEvent.Post e) {
 
             lastPage = currentPage;
         }
+
+        if (!quickJumping) return;
```
kristofbolyai, line 121: Are you sure the buttons are set by content, and not set by slot? Which happens earlier? If both happens, using the first would improve speed a bit, using the second packet would improve reliability

ShadowCat117: Just tried it with set slot, it seems to be almost the exact same speed

kristofbolyai: Use the one that is sent later then

kristofbolyai (approved):
LGTM

PR #2215, opened 2023-10-28
fix: fix some tooltip fitting bugs

kristofbolyai:
This improves the current state but still behaves pretty badly with universal scales...


kristofbolyai:
Tooltip wrapping also behaves weirdly with >1 universal scales..

PR #2216, opened 2023-10-28
feat: Center guild map when outside main area

ShadowCat117:
Border in video is just for a demo of the area you need to be outside of for it to trigger, doesn't actually show up with this

https://github.com/Wynntils/Artemis/assets/8337467/63172724-dcca-4d8c-811f-447de161414d

### DIFF
--- a/common/src/main/java/com/wynntils/screens/maps/GuildMapScreen.java
+++ b/common/src/main/java/com/wynntils/screens/maps/GuildMapScreen.java
@@ -20,6 +20,7 @@ import com.wynntils.services.map.pois.Poi;
 import com.wynntils.services.map.pois.TerritoryPoi;
 import com.wynntils.services.map.pois.WaypointPoi;
 import com.wynntils.services.map.type.TerritoryDefenseFilterType;
+import com.wynntils.utils.MathUtils;
 import com.wynntils.utils.colors.CommonColors;
 import com.wynntils.utils.mc.KeyboardUtils;
 import com.wynntils.utils.mc.McUtils;
@@ -45,6 +46,12 @@ import net.minecraft.sounds.SoundEvents;
 import org.lwjgl.glfw.GLFW;
 
 public final class GuildMapScreen extends AbstractMapScreen {
+    private static final int MAP_CENTER_X = -150;
+    private static final int MAP_CENTER_Z = -3000;
+    private static final int MAX_X = 1650;
+    private static final int MAX_Z = -150;
+    private static final int MIN_X = -2400;
+    private static final int MIN_Z = -6600;
     private boolean resourceMode = false;
     private boolean territoryDefenseFilterEnabled = false;
     private boolean hybridMode = true;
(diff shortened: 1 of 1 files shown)


kristofbolyai:
I think this could be a setting (default on) for the main map too, some people use that for territories as well (and legacy did it too).

ShadowCat117:
> I think this could be a setting (default on) for the main map too, some people use that for territories as well (and legacy did it too).

I was thinking of doing it for the main map, but problem is there are certain mapped areas outside the main map like the void, a better approach would probably be to detect if there are no map files for the area, in which case minimap could be hidden too

kristofbolyai:
> I was thinking of doing it for the main map, but problem is there are certain mapped areas outside the main map like the void, a better approach would probably be to detect if there are no map files for the area, in which case minimap could be hidden too

Sure. Wanna do it here or not?

ShadowCat117:
> Sure. Wanna do it here or not?

I think this method for guild map is fine, as I don't think you'll ever want to view guild map for areas that don't have any territories so I'm fine leaving this as is and then I'll see if I can do the main map as a seperate pr either tomorrow or sometime next week

common/src/main/java/com/wynntils/screens/maps/GuildMapScreen.java
```diff
@@ -55,7 +62,14 @@ public final class GuildMapScreen extends AbstractMapScreen {
...
     private BasicTexturedButton hybridModeButton;
 
-    private GuildMapScreen() {}
+    private GuildMapScreen() {
+        // When outside of the main map, center to the middle of the map
+        if (!MathUtils.isInside(
+                (int) McUtils.player().getX(), (int) McUtils.player().getZ(), MIN_X, MAX_X, MIN_Z, MAX_Z)) {
+            updateMapCenter(MAP_CENTER_X, MAP_CENTER_Z);
+            setZoom(0);
+        }
```
kristofbolyai: I don't think this should be decided in the constructor, since some screens hold references to screen instances long before they are opened. Create a `firstInit` boolean, and run this there.

common/src/main/java/com/wynntils/screens/maps/GuildMapScreen.java
```diff
@@ -55,7 +64,9 @@ public final class GuildMapScreen extends AbstractMapScreen {
     private BasicTexturedButton territoryDefenseFilterButton;
     private BasicTexturedButton hybridModeButton;
 
-    private GuildMapScreen() {}
+    private GuildMapScreen() {
+        this.firstInit = true;
```
kristofbolyai: Sorry for the nit, but could you move setting this to the field declaration? That would make it like other screens. Thanks

ShadowCat117: I think I've seen another screen do it this way but maybe I'm wrong, can't find it now to fix here

kristofbolyai (approved):
I am not a 100% sure about the const field placement/name, but let's go with it for now, as this is a good change.

ShadowCat117:
> I am not a 100% sure about the const field placement/name, but let's go with it for now, as this is a good change.

When I do the main map using map files to determine if there's nothing to display, I might add a config option disabled by default to always center to main area and I'll move the consts to AbstractMapScreen

PR #2222, opened 2023-10-29
feat: Add option to hide minimap and center main map in unmapped areas

ShadowCat117:
I imagine this may be annoying for those who are unable to download the map files but I'm not sure if there's anything that can be done about that

https://github.com/Wynntils/Artemis/assets/8337467/47c57fff-cc28-419a-a901-c28f09ebf32d


ShadowCat117:
~~https://github.com/Wynntils/Artemis/pull/2222/commits/60dd2d5ea979e2e1fb9babef0f018f682f56dd97 also fixes a minor issue not worthy of it's own pr where when you're near the edge of a map file, it won't render it
Tested turning the mask off and it wasn't rendering anymore than it should have~~

Incorrect fix

common/src/main/java/com/wynntils/utils/type/BoundingBox.java
```diff
@@ -31,6 +31,6 @@ public boolean contains(float x, float z) {
     public boolean intersects(BoundingBox other) {
         boolean xIntersects = Math.max(x1, other.x1) < Math.min(x2, other.x2);
         boolean zIntersects = Math.max(z1, other.z1) < Math.min(z2, other.z2);
-        return xIntersects && zIntersects;
+        return xIntersects || zIntersects;
```
kristofbolyai: I saw the bug with the minimap recently, but are you sure this is the correct fix? It seems like that when I take an intersection of 2 2d objects, I need to have intersections on both axes.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
```diff
@@ -86,6 +86,9 @@ public class MainMapFeature extends Feature {
     @Persisted
     public final Config<Float> poiScale = new Config<>(1f);
 
+    @Persisted
+    public final Config<Boolean> centerWhenUnmapped = new Config<>(false);
```
kristofbolyai: Default true should be fine imo

common/src/main/java/com/wynntils/utils/type/BoundingBox.java
ShadowCat117: Actually I just tried it again and it is rendering pois outside of the area with this change so no I don't think it actually is correct, I'll revert that

kristofbolyai: I think this is something about us only calculating map parts in our "rectangle" bounding box, but when we have map rotations enabled, we basically need to check the circle  enclosing the rectangle (which is easy to calculate since there is only 1 of them).

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
ShadowCat117: Oh yeah I was thinking of guild map centering condition

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -58,6 +58,9 @@ public class MinimapOverlay extends Overlay {
     @Persisted
     public final Config<Boolean> followPlayerRotation = new Config<>(true);
 
+    @Persisted
+    public final Config<Boolean> hideWhenUnmapped = new Config<>(true);
```
magicus: I'm not sure this should be true by default. This was the behavior of Legacy, and always found it annoying. Even if you don't have a minimap to show, the coordinates is relevant. I saw in your demo that you kept them visible, but then it just becomes visually odd. I also think it looks a bit weird with the minimap turning on and off. Ideally, we will map most of the quest areas, so we should only rarely end up in places without a map. But that will make the minimaps disappearance even more odd-looking.

ShadowCat117: I don't think it disappearing and reappearing is a problem, especially since Wynn has a lot of the fade to black transitions now, however the coordinates staying does look a bit weird, I've already moved my coordinates above on my instance in prep for this but I don't think we should just move it for everyone so maybe off by default is better

kristofbolyai: add an 'inmappedarea' function and we can fix that (here or new pr)

ShadowCat117: I don't follow, how will that fix this? I don't think we should hide coordinates when outside mapped areas if that's what you're suggesting

kristofbolyai: oh, I thought if we hide the map, we could hide the coordinates as well.

ShadowCat117: Unless we change this to 3 options, hide minimap, hide minimap & coordinates or neither

ShadowCat117: But I think coordinates should always be visible as that can be useful whereas a black map is pointless

magicus: I agree that coordinates is helpful. Moving them to above the minimap could definitely be a way to make it look less odd to just have the coordinates. And maybe having a tristate for the setting could make sense, at least if bolyai thinks the coordinates should be hidden too. :)

magicus: So, you can revert your latest fix and restore your original behavior, at least if you try to change the default positioning of the widgets so coordinates is above the minimap.

ShadowCat117: How's this? Territory overlay should be fine as in unmapped areas there's no territories anyway

magicus: I think it looks fine.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -490,6 +496,12 @@ private void renderMapBorder(PoseStack poseStack, float renderX, float renderY,
     @Override
     protected void onConfigUpdate(Config<?> config) {}
 
+    public enum UnmappedOption {
+        MINIMAP,
+        BOTH,
```
magicus:
```suggestion
        MINIMAP_AND_COORDS,
```

magicus (approved):

common/src/main/java/com/wynntils/overlays/minimap/CoordinateOverlay.java
```diff
@@ -43,7 +45,15 @@ protected void onConfigUpdate(Config<?> config) {
 
     @Override
     public String getTemplate() {
-        return this.shouldBeColored.get() ? TEMPLATE_COLORED : TEMPLATE;
+        if (Managers.Feature.getFeatureInstance(MinimapFeature.class)
+                        .minimapOverlay
+                        .hideWhenUnmapped
+                        .get()
```
kristofbolyai, line 51: Bit of cross-feature configurations.. I guess it is fine for the time being..

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -212,6 +212,21 @@ protected void doInit() {
...
+
+        if (firstInit) {
+            BoundingBox textureBoundingBox =
+                    BoundingBox.centered(mapCenterX, mapCenterZ, width / currentZoom, height / currentZoom);
+
+            // When in an unmapped area, center to the middle of the map if the feature is enabled
+            if (Managers.Feature.getFeatureInstance(MainMapFeature.class)
+                            .centerWhenUnmapped
+                            .get()
+                    && Services.Map.getMapsForBoundingBox(textureBoundingBox).isEmpty()) {
```
kristofbolyai, line 224: Didn't you add a method for `isPlayerInMappedArea`? Use that here.

kristofbolyai: Actually I don't know if that is what we want to do.. Although it is weird that the same config works differently on the guild map.

ShadowCat117: The one used for guild map is inside the main area as that's the only place with territories. Main map is if there's a map file for current location

common/src/main/java/com/wynntils/services/map/MapService.java
```diff
@@ -50,6 +53,18 @@ public List<MapTexture> getMapsForBoundingCircle(BoundingCircle circle) {
...
+        MinimapOverlay minimapOverlay = Managers.Feature.getFeatureInstance(MinimapFeature.class).minimapOverlay;
+        float width = minimapOverlay.getWidth();
+        float height = minimapOverlay.getHeight();
+        float scale = minimapOverlay.scale.get();
+
+        BoundingCircle textureBoundingCircle = BoundingCircle.enclosingCircle(BoundingBox.centered(
+                (float) McUtils.player().getX(), (float) McUtils.player().getZ(), width * scale, height * scale));
+
+        return !getMapsForBoundingCircle(textureBoundingCircle).isEmpty();
+    }
```
kristofbolyai: My last issue with this PR is this code. I don't like that we are referencing the overlay instance here, the whole method has a bit of "doesn't really fit the description" vibe. What if we simply check if the player x/y is not inside a mapped area? Or what if we just specify a constant box (eg. 100 in X-Y directions) and check that?

ShadowCat117: Perhaps could just set the width, height and scale to the default values of the minimap overlay

kristofbolyai: You know what? Just make it arguments of the function, and provide a default value..

kristofbolyai (approved):

Issue #2179, opened 2023-10-09
Comparing two pieces of gear will sometimes make the tooltips overlap


EpicPuppy613:
![image](https://github.com/Wynntils/Artemis/assets/61771471/7781a40f-e8cc-484e-b634-4a87f9ddde45)
Its very inconsistent where its perfectly fine with some comparisons, but others (like the one shown above) overlap and is sometimes hard to read.

Rafii2198:
This might be unrelated and could be caused by legendary tooltips (I will do proper checks at a later time) but it seems like if the second tooltip doesn't have enough horizontal space it tries to be rendered on the other side, but it goes to weird x positions.
![2023-10-09_02 09 59](https://github.com/Wynntils/Artemis/assets/26695327/2529b99d-a68b-4443-9bb5-e31a6443c81e)
On this example, you can see that I am comparing a chestplate in my inventory with my boots, the tooltip of the boots is going to the left of them and thus the chestplate tooltip is trying to render on the left of that tooltip, but it would not fit the screen and tries to render on the right but fails to determine right x position.

EpicPuppy613:
> This might be unrelated and could be caused by legendary tooltips (I will do proper checks at a later time) but it seems like if the second tooltip doesn't have enough horizontal space it tries to be rendered on the other side, but it goes to weird x positions.

Another theory for this might be because the tooltip is too big, so when it renders, it scales the entire screen up, but it doesn't account for the scaled up screen when offsetting the x position

EpicPuppy613:
I have spotted the main issue, the x offset is not accounting for the fact that the item name could be bold (perfect) or the percentage numbers on the identification. I might give a go at fixing this.
![image](https://github.com/Wynntils/Artemis/assets/61771471/acc217be-5ee4-412f-b133-1cc7f1c4341d)

EpicPuppy613:
I tried fixing it, but the tooltip system is just really goofy and stupid. I'll try to commit a partial fix soon, but this issue is definitely going to take a bit to fix.

ShadowCat117:
https://github.com/Wynntils/Wynntils/pull/2876 reworked the feature so they should no longer overlap

PR #2226, opened 2023-10-29
fix: properly position item tooltips when applying custom scaling

P0keDev:
Simple one-line fix. IIRC 1.19's tooltip code handled the tooltip location separately, but in 1.20 it is done after we apply our scaling, so the mouse coordinates have to be scaled accordingly


magicus (approved):

EpicPuppy613:
Potentially related to #2179

PR #2228, opened 2023-10-31
feat: Add Menu Command

Rafii2198:
As per this suggestion: https://feedback.wynntils.com/posts/273/add-a-command-to-open-content-book
I didn't add it as Wynntils command as perhaps it could be expanded in future to open each menu, like PoiManager, Quests, Settings, etc. and it would be just more convenient for the user.


common/src/main/java/com/wynntils/commands/MenuCommand.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    protected LiteralArgumentBuilder<CommandSourceStack> getCommandBuilder(
+            LiteralArgumentBuilder<CommandSourceStack> base) {
+        return base.executes(this::openMenu);
+    }
+
+    private int openMenu(CommandContext<CommandSourceStack> commandSourceStackCommandContext) {
+        Managers.TickScheduler.scheduleLater(() -> WynntilsMenuScreenBase.openBook(WynntilsMenuScreen.create()), 2);
```
kristofbolyai: Why the delay?

kristofbolyai: Okay, I see. Perhaps copy the comment too..

kristofbolyai (approved):

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -40,6 +40,7 @@
   "command.wynntils.lootrun.undoStandNear": "Stand near a previous spot to undo movements.",
   "command.wynntils.lootrun.undoSuccessful": "Successfully undid movements.",
   "command.wynntils.map.description": "Manage map related settings",
+  "command.wynntils.menu.description": "Opens the Wynntils Menu",
   "command.wynntils.quest.description": "List, show and track quests",
   "command.wynntils.reauth.tryReauth": "Disconnecting from Hades, and trying to reauthenticate...",
   "command.wynntils.reloadCaches.reloading": "Reloading caches...",
```
kristofbolyai: Actually, no, I don't like this. Move this command into the wynntils command as "wynntils menu"

kristofbolyai (changes requested):

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: I have no idea why this comment is here though, thanks Github Mobile.

common/src/main/java/com/wynntils/commands/MenuCommand.java
ShadowCat117: > overwriting the lootrun screen

Maybe change that to be more accurate for the screen it will open :)

common/src/main/resources/assets/wynntils/lang/en_us.json
Rafii2198: I tried, but it doesn't work for some reason. Should I commit so you can take a look? Basically suggestion and help command shows it but when I try to execute it, I am getting an error about unknown command.

kristofbolyai: Sure

Rafii2198: Alright it was very small thing that I missed

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -362,6 +365,12 @@ private int update(CommandContext<CommandSourceStack> context) {
         return 1;
     }
 
+    private int menu(CommandContext<CommandSourceStack> context) {
+        // Delay is needed to prevent chat screen overwriting the lootrun screen
```
kristofbolyai:
```suggestion
        // Delay is needed to prevent chat screen overwriting the menu screen
```

kristofbolyai (approved):

PR #2230, opened 2023-10-31
fix: Don't rely on coordinates to guess the lootrun locations

kristofbolyai:
Whenever doing a lootrun is Sky or MH, you can get too close to the other region's border, causing task predictions to act like they don't exist. This patch improves this as we don't strictly need to rely on regions to guess task locations. This makes gameplay more stable, but hurts lootrun data collection as we can't guess the location. Whenever we need data collection again, I will aim to fix that problem.

### DIFF
--- a/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
+++ b/common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
@@ -45,6 +45,7 @@ import com.wynntils.utils.mc.type.Location;
 import com.wynntils.utils.type.CappedValue;
 import com.wynntils.utils.type.Pair;
 import java.lang.reflect.Type;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -163,26 +165,55 @@ private void loadLootrunTaskLocations() {
...
-                        Location.containing(event.getParticle().position()),
-                        taskLocation.taskType()));
-                return;
+        boolean foundTaskLocation = false;
+        for (Set<TaskLocation> taskLocationsForLocation : taskLocations.values()) {
+            for (TaskLocation taskLocation : taskLocationsForLocation) {
+                if (PosUtils.closerThanIgnoringY(
+                        taskLocation.location().toVec3(), event.getParticle().position(), TASK_POSITION_ERROR)) {
+                    // Note: We do this re-allocation so we always display the correct location,
+                    //       even if it slightly changed, or our data is inprecise.
```
ryanzhoudev:
```suggestion
                    //       even if it slightly changed, or our data is imprecise.
```

ryanzhoudev (approved):

PR #2233, opened 2023-11-02
fix: PerCharacterGuildContribution not always triggering

ShadowCat117:
15 ticks seemed to be a reliable delay

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
@@ -4,6 +4,7 @@
  */
 package com.wynntils.features.utilities;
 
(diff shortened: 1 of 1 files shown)


kristofbolyai:
To be honest, I may prefer a systematic solution to this one.. Perhaps a "sendCommandWhenLoaded" in CommandManager? Although command manager has no knowledge on Wynn atm.. but we have many commands we run on world join, so perhaps we could enqueue them, delay them, etc in one place.

ShadowCat117:
> To be honest, I may prefer a systematic solution to this one.. Perhaps a "sendCommandWhenLoaded" in CommandManager? Although command manager has no knowledge on Wynn atm.. but we have many commands we run on world join, so perhaps we could enqueue them, delay them, etc in one place.

If the party model is going to need similar changes then that is probably worth doing instead

PR #2234, opened 2023-11-02
fix: Command cooldown related fixes

ryanzhoudev:
Including, but not limited to
- Models requesting data too quickly and stalling
- Per-character guild xp not setting properly


common/src/main/java/com/wynntils/utils/mc/McUtils.java
```diff
@@ -96,14 +96,6 @@ public static void sendSequencedPacket(PredictiveAction predictiveAction) {
         mc().gameMode.startPrediction(mc().level, predictiveAction);
     }
 
-    /**
-     * Sends the specified command to the server.
-     * @param command The command to send. The leading '/' should not be included.
-     */
-    public static void sendCommand(String command) {
-        mc().getConnection().sendCommand(command);
```
magicus, line 104: Can't you just keep this and let it do `Managers.Command.queueCommand` instead? It will both mean fewer changes, and that there will continue to be a simple to use util function to send commands.

ryanzhoudev: Imo that is kind of strange. There was a fairly significant change on wynncraft with the new command ratelimits, and our code should reflect that. It also doesn't make sense that a function that sends commands now just magically queues them, and it is also kind of strange that it is in McUtils of all places, since it is a custom implementation by us.

magicus:
Okay, I have done a better review now that I'm not half alseep. :-)

1) I have missed the rate limit change. That's on me, I haven't been on discord for a while. If the tick constant of 7 is correct, then then implementation seems like it will indeed be rate limiting our commands.

2) I am not sure this belongs in the CommandManager. This has -- up until now, at least -- been responsible for implementing our own commands. That is a quite different task than rate limiting packets sent to the server, even if both tasks contain the word "command". It might be the best place after all, but at least it needs some discussion.

3) The implementation could be made more efficient. If sufficient time has passed since the last command, then we should be able to send the command directly, and not enqueue it. In this implementation, all commands will be delayed by one tick, even if it is not necessary.

4) I am not convinced by the argument that we need to replace all calls to the McUtil function. Sending commands to the server is a kind of service the Wynntils framework provides. Up until now, this has been just a simple wrapper around vanilla code, but now it needs to be a bit more complex. The basic idea still remains though. This also ties in with 2. Maybe the entire enquing should/could be made in McUtil. It increases the level of complexity in that class, but it could make sense.

5) Neither am I completely happy with the name change to "queueCommand". I understand the wish to signal that it is not necessarily sent immediately, but in my mind, it just leaves the dangling question "well, how do i unqueue it then? do i need to send a process queue message as well?". To understand that this is not needed, I will need to have a much deeper implementation knowledge about the command enqueing system. And contrary, I don't think there ever was any implication that McUtils.sendCommand would do so immediately; of course the command will be sent whenever it was possible. 

So I still think `McUtils.sendCommand` is a perfectly fine way for code to send commands, regardless how it is implemented under the hood.

(In a similar vein, I have long time been miffed by the "enqueue" methods of the notification manager. I have been planning to fix this for a long time, but just never got around to it)

kristofbolyai:
Please note that while I agree using McUtils directly would be cleaner, referencing a handler/manager (or any other component) from there would cause unforeseen class loading order issues at any time we don't explicitly check for it..

magicus:
@kristofbolyai That is a valid point. The better the argument for having the entire implementation with the queue in the util class.

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
```diff
@@ -413,7 +414,7 @@ private void inviteFromField() {
 
     private void refreshAll() {
         Models.Party.requestData();
-        Models.Friends.requestData();
+        Managers.TickScheduler.scheduleLater(Models.Friends::requestData, 5);
```
kristofbolyai: Why?

kristofbolyai:
Some steps we can take:
1, Rename `CommandManager` to `ClientCommandManager`
2, Create a `CommandHandler` and implement queueing there

I know moving away from using `McUtils` is a bit unfortunate, but I also think having the queueing there breaks some of our abstractions (for example, do you event bus register `McUtils` for this..?).

What do you think @magicus @DonkeyBlaster?

ryanzhoudev:
Sounds fine to me, I will start if/when magicus agrees

common/src/main/java/com/wynntils/screens/partymanagement/PartyManagementScreen.java
ryanzhoudev: This was just testing to see if it was a ratelimit problem, removed now

magicus:
Those were good suggestions @kristofbolyai. I agree fully!

magicus:
(I must admit that I did not think about the fact that moving the queue to McUtils would need us to have that class starting to listen to events. That would, as you say, be a bad thing).

magicus (approved):
Now it looks great! Thanks!

PR #2243, opened 2023-11-09
fix: Round spell cost base values correctly

kristofbolyai:
Previously we ignored that these inverted values have to be rounded to negative infinity for positive values, and vice-versa.

I confirmed the validness of this fix by logging all changes between the old and new calculations, first confirmed it only changed for inverted stats (aka spell cost stats), then handpicked a few of them to check against a known-good base (Wynnbuilder).

### DIFF
--- a/common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
+++ b/common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
@@ -480,7 +480,8 @@ public class GearInfoRegistry {
                 }
                 // Range will always be stored such as "low" means "worst possible value" and
                 // "high" means "best possible value".
-                RangedValue range = StatCalculator.calculatePossibleValuesRange(baseValue, preIdentified);
+                RangedValue range = StatCalculator.calculatePossibleValuesRange(
+                        baseValue, preIdentified, statType.showAsInverted());
                 StatPossibleValues possibleValues = new StatPossibleValues(statType, range, baseValue, preIdentified);
                 list.add(Pair.of(statType, possibleValues));
             }
--- a/common/src/main/java/com/wynntils/models/stats/StatCalculator.java
+++ b/common/src/main/java/com/wynntils/models/stats/StatCalculator.java
@@ -11,25 +11,42 @@ import com.wynntils.models.stats.type.StatType;
 import com.wynntils.utils.MathUtils;
 import com.wynntils.utils.type.Pair;
 import com.wynntils.utils.type.RangedValue;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 
 public final class StatCalculator {
-    public static RangedValue calculatePossibleValuesRange(int baseValue, boolean preIdentified) {
+    public static RangedValue calculatePossibleValuesRange(int baseValue, boolean preIdentified, boolean isInverted) {
         if (preIdentified) {
             // This is actually a single, fixed value
             return RangedValue.of(baseValue, baseValue);
         } else {
             int low;
             int high;
+
             if (baseValue > 0) {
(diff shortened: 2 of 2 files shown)


kristofbolyai:
@magicus You should review this.

magicus:
Argh. I really hate those "negative positive" stats. 

If you have tested this, I guess it is fine then. Otoh, I thought I tested it quite thoroughly myself,  but it was apparently wrong anyway. 

Hm. Do you think you can write some unit test cases for this?

magicus:
Oh, wait, I see you use BigDecimal. That might have performance impact. How often do we call this? I'd rather see this done without resorting to BigDecimal.

kristofbolyai:
> Oh, wait, I see you use BigDecimal. That might have performance impact. How often do we call this? I'd rather see this done without resorting to BigDecimal.

This is called on API load, on boot. You can see how this doesn't matter in this context.

kristofbolyai:
> Argh. I really hate those "negative positive" stats.
> 
> If you have tested this, I guess it is fine then. Otoh, I thought I tested it quite thoroughly myself, but it was apparently wrong anyway.
> 
> Hm. Do you think you can write some unit test cases for this?

We could probably have unit tests but only for pre-determined values..

magicus:
Yes, that was kind of what I meant. Have some unit tests for edge case values, like when a value is very close to or exactly at the point where it will be rounded differently, and basically duplicate all lines and swap the sign after that.

We have basically four categories to test:
positive values when positive is good
negative values when positive is good
positive values when negative is good
negative values when negative is good

And for each category, we need to have like a value that is "just below" the threshold for rounding change, at the threshold, and "just above" the threshold.

kristofbolyai:
Do you think this could be merged now, without the added tests?

magicus:
As I said, I am not so happy about BigDecimal. Why do you need that?

kristofbolyai:
> As I said, I am not so happy about BigDecimal. Why do you need that?

But it is not used in runtime, only when loading (as I answered above, maybe you missed that). I am not sure if Java has some other way of rounding in different ways other than BigDecimal.

magicus:
Ah, sorry, I missed that. 

The canonical way to control rounding without BigDecimal is to add or subtract 0.5. This always makes my head spin, though. It's like the "one-off" problem but on the number line instead.

The complexity can certainly be alleviated by creating some util helper functions with explicit names like "roundTowardsPositiveInfinity" or something like that, or that mimicks the BigDecimal and takes a named enum for rounding mode. (In fact, we can probably reuse RoundingMode...)

kristofbolyai:
> Ah, sorry, I missed that.
> 
> The canonical way to control rounding without BigDecimal is to add or subtract 0.5. This always makes my head spin, though. It's like the "one-off" problem but on the number line instead.
> 
> The complexity can certainly be alleviated by creating some util helper functions with explicit names like "roundTowardsPositiveInfinity" or something like that, or that mimicks the BigDecimal and takes a named enum for rounding mode. (In fact, we can probably reuse RoundingMode...)

I would still stick to using BigDecimal, until we don't need to do it many times during "runtime" (gameplay). My main points being:
- The performance impact of using BigDecimal for calculating the stat ranges (once, during boot) is minimal. An async-profiler run collected ~78 samples in the `calculatePossibleValuesRange` method, in profile rounds with around ~8000 samples taken. This makes the performance "impact" (which is more of a boot time increase) virtually non-existent.
- Using a custom rounding util class would need a new set of unit tests

magicus (approved):
Okay. I accept. :) But if you (or someone else, perhaps a good intern/beginners job?) ever can be bothered to write some unit tests, that'd great.

PR #2244, opened 2023-11-10
fix: Display attack timer and territory name on hover

ShadowCat117:
Hover


magicus (approved):

kristofbolyai (approved):
Yeah, that variable was kind of messed up.. I thought I catched that mistake a while back but it seems I forgot to check usages!

PR #2247, opened 2023-11-12
chore: remove Debugify from the incompatible mod list (experimentally)

Julienraptor01:
Debugify seems to not break anymore and thus the break should be removed

i found it by helping a user within the FO server
i told them to replace the fabric loader overrides we use for FO and instead use a custom one where i remove the break on debugify from wynntils
they told it worked and they didn't seem to have the issue that existed in 1.19.3 that i found by looking in the other wynntils repo
(them)
image
(me)
image
(them)
image
(me)
image
(still me)
image
(them)
image
(still them)
image

here is a log of them successfully joining the server with wynntils, debugify and the break override : https://mclo.gs/1lRnWyz

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -3,7 +3,6 @@
     "Team Wynntils"
   ],
   "breaks": {
-    "debugify": "*",
     "exordium": "*",
     "memoryleakfix": "<1.1.2"
   },
--- a/quilt/src/main/resources/quilt.mod.json
+++ b/quilt/src/main/resources/quilt.mod.json
@@ -6,10 +6,6 @@
   "quilt_loader": {
     "breaks": [
       {
-        "id": "debugify",
-        "versions": "*"
-      },
-      {
         "id": "exordium",
         "versions": "*"
       },


kristofbolyai:
- How am I supposed to confirm it does not break anymore? I know it was an older issue that we placed debugify into breaks, but as far as I know it tends to break on Wynn even without Artemis installed. Artemis being a Wynn only mod, it might be worth keeping it a disallowed mod.

Julienraptor01:
> * How am I supposed to confirm it does not break anymore? I know it was an older issue that we placed debugify into breaks, but as far as I know it tends to break on Wynn even without Artemis installed. Artemis being a Wynn only mod, it might be worth keeping it a disallowed mod.

by testing yourself and also by the log that show it work
if you got issues with debugify on wynncraft, report them to isXander on the debugify repo, don't just add a break in the mod
debugify shouldn't break on any server with its default config so if something goes wrong, it's unintended behavior that should be fixed, not avoided

kristofbolyai:
Okay, let me ask someone to test internally.

Julienraptor01:
i've now tested myself for like 1h30 and i didn't got any crashes or anything (and i also discovered wynncraft which i had never played before 😄)

*Edit : i played for 3:45 straight without any crash or issue*

Rafii2198:
I have played for like a day and didn't notice anything out of ordinary. But I must put emphasis on the fact that the bugs caused by Debugify were extremely random. For example, we got **many** reports of crashes caused by something with packets, I personally never experienced that. To be fully sure, we would have to have more people test it.

Julienraptor01:
> I have played for like a day and didn't notice anything out of ordinary. But I must put emphasis on the fact that the bugs caused by Debugify were extremely random. For example, we got **many** reports of crashes caused by something with packets, I personally never experienced that. To be fully sure, we would have to have more people test it.

Maybe offer an alpha testing release or some sort to get many different people to try it or smth ?

Also i only noticed like 3 or 4 issues on GitHub that reported it, do you get your issues somewhere else ?

Rafii2198:
> Also i only noticed like 3 or 4 issues on GitHub that reported it, do you get your issues somewhere else ?

Discord mainly

> Maybe offer an alpha testing release or some sort to get many different people to try it or smth ?

Perhaps we can merge it now, but as soon as we get another report we will add it back. From what you are saying, Debugify got stuff fixed, so perhaps it is really not causing any issues anymore.
But it is up to devs.

magicus:
Maybe it was an older version of debugify that broke? It does not have to be black-and-white...

magicus:
This was introduced in https://github.com/Wynntils/Artemis/pull/2166. There is nothing in the PR description that indicates why the specific set of mods were listed. :-( @DonkeyBlaster You wrote the PR, do you remember?

(Also, in the future, please add descriptions to PRs, it is sooo helpful in situations like this when you try to determine why a certain change was made)

Rafii2198:
These were the mods listed on Discord that caused severe issues. https://discord.com/channels/394189072635133952/1128803191643967619/1128803191643967619
The full list is as follows:
- Exordium - Incorrectly affects overlays making them stutter as well as making most of hud opaque or being unable to see text in text boxes
- Debugify - Random crashes, inability to join server, other bugs.
- Optifine / OptiFabric - Breaks most of the game, can cause crashes. Use better alternatives instead.
- AvoMod - Random features of both mods will not functions properly when used together.
- Chat Patches - Chat related features break.
- WeirdChat - Causes many chat related features to not work correctly including base Wynncraft features as well as spell indicator. Use Emogg instead.
- Cloth API - Causes crashes on startup, if you are using Voices of Wynn use the mod Cloth Config instead.
- Memory Leak Fix -  Causing mixin crashes, use the latest version of this mod to fix it.

ryanzhoudev:
Rafii is right; I just put in as many changes from the incompatible mod list in the Discord as possible.

kristofbolyai:
> Perhaps we can merge it now, but as soon as we get another report we will add it back. From what you are saying, Debugify got stuff fixed, so perhaps it is really not causing any issues anymore.

This is basically what I think too. Let's do it.

kristofbolyai (approved):

PR #2249, opened 2023-11-13
feat: Jukebox menu scrolling

ryanzhoudev:


ShadowCat117:
I think it could be worth adding search support for it too.

Unless the vanilla one has searching, been a while since I used it

ShadowCat117:
There is also a jukebox on housing and it has basically the same layout minus the randomize option however the title is "Select Songs" so maybe support that too

magicus (approved):

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -50,12 +50,14 @@ public final class ContainerModel extends Model {
     private static final Pair<Integer, Integer> CONTENT_BOOK_PREVIOUS_NEXT_SLOTS = new Pair<>(65, 69);
     private static final Pair<Integer, Integer> GUILD_BANK_PREVIOUS_NEXT_SLOTS = new Pair<>(9, 27);
     private static final Pair<Integer, Integer> GUILD_MEMBER_LIST_PREVIOUS_NEXT_SLOTS = new Pair<>(10, 28);
+    private static final Pair<Integer, Integer> JUKEBOX_PREVIOUS_NEXT_SLOTS = new Pair<>(17, 8);
```
magicus, line 53: ... wait a sec. Don't you need to reference this as well? You need to add the jukebox screen as scrollable, somehow?

magicus: @DonkeyBlaster 

I did not see that you had auto merge turned on, so this comment arrived too late.

ryanzhoudev: Hmm.... I swear I did the normal implementation and everything and even played with it in-game. Guess not. I'll follow up on this PR with the search and fixes in a bit.

PR #2250, opened 2023-11-13
feat: Skill point loadouts

ryanzhoudev:
~~(video coming soon)~~ 2 videos!

https://github.com/Wynntils/Artemis/assets/57310593/e019a501-b25f-46a8-9b19-7200e2c6ecd9

In text, video 1:
- Save loadouts (no armour/accessories) and builds (armour and/or accessories)
- Loadouts displayed with just name and points
- Builds displayed with name and armour, or accessories if no armour present
- Types can be converted between each other, description of how is in the convert button tooltip
- Confirmation when overwriting existing build/loadout
- Points can be very quickly, much faster than legacy. Loads in 5's where possible using shift-clicks.

https://github.com/Wynntils/Artemis/assets/57310593/a33bb9c8-62e4-41f9-9fad-bd786f516500

Video 2:
- All builds/loadouts display with number of total points and level requirement, in red when not met
- Load button has a warning message when level requirement not met
- Low level players without tome access are parsed properly


common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
```diff
@@ -48,7 +48,8 @@ public static boolean openInventory(int slotNum) {
      * Clicks on a slot in the specified container. containerId and the list of items should correspond to the
      * same container!
      */
-    public static void clickOnSlot(int clickedSlot, int containerId, int mouseButton, List<ItemStack> items) {
+    public static void clickOnSlot(
+            int clickedSlot, int containerId, int mouseButton, List<ItemStack> items, boolean shiftClick) {
```
kristofbolyai: I would create a method with shift-click being off by default, to reduce the changes in the PR, and to make the method simpler to use.

common/src/main/java/com/wynntils/utils/mc/McUtils.java
```diff
@@ -104,4 +105,12 @@ public static void sendSequencedPacket(PredictiveAction predictiveAction) {
...
         mc().getConnection().sendChat(message);
     }
+
+    /**
+     * Closes invisible containers opened in the background, without closing the visible screen.
+     */
+    public static void closeBackgroundContainer() {
+        player().connection.send(new ServerboundContainerClosePacket(McUtils.player().containerMenu.containerId));
+        player().containerMenu = McUtils.player().inventoryMenu;
+    }
```
kristofbolyai: Maybe fits into container utils more, this is not so commonly used to be in McUtils

common/src/main/java/com/wynntils/features/utilities/SkillPointLoadoutsFeature.java
```diff
@@ -0,0 +1,46 @@
...
+import com.wynntils.core.consumers.features.Feature;
+import com.wynntils.mc.event.ScreenOpenedEvent;
+import com.wynntils.screens.base.widgets.WynntilsButton;
+import com.wynntils.screens.skillpointloadouts.SkillPointLoadoutsScreen;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraft.client.gui.screens.inventory.ContainerScreen;
+import net.minecraft.network.chat.Component;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class SkillPointLoadoutsFeature extends Feature {
```
kristofbolyai, line 20: Missing ConfigCategory

common/src/main/java/com/wynntils/features/utilities/SkillPointLoadoutsFeature.java
```diff
@@ -0,0 +1,46 @@
...
+    @SubscribeEvent
+    public void onCharacterInfoScreenOpened(ScreenOpenedEvent e) {
+        if (!(e.getScreen() instanceof ContainerScreen screen)) return;
+        if (!Models.Container.isCharacterInfoScreen(screen)) return;
+        if (screen.topPos == 0 || screen.width == 0) return; // Not done loading
```
kristofbolyai: ScreenOpenedEvent has .Pre and .Post, perhaps you need to use post?

common/src/main/java/com/wynntils/models/skillpoint/SavableSkillPointSet.java
```diff
@@ -0,0 +1,82 @@
...
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.skillpoint;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class SavableSkillPointSet {
```
kristofbolyai: This class really wants to be a record, doesn't it? You can create factory methods instead of the constructors.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import net.minecraft.world.item.ItemStack;
+import org.lwjgl.glfw.GLFW;
+
+public class SkillPointModel extends Model {
+    @Persisted
+    private final Storage<Map<String, SavableSkillPointSet>> skillPointLoadouts = new Storage<>(new LinkedHashMap<>());
```
kristofbolyai: The persisted wrappers default to `TreeMap`, and you can't change that at the moment. I assume you wanted  to keep the order here, so `TreeMap` should work just fine.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+            getAssignedSkillPoints(Skill.INTELLIGENCE),
+            getAssignedSkillPoints(Skill.DEFENCE),
+            getAssignedSkillPoints(Skill.AGILITY)
+        });
+    }
+
+    /**
+     * Saves the current equipped gear and provided skill points.
+     * @param name
+     * @param skillPoints
```
kristofbolyai: missing descriptions

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+        McUtils.inventory().armor.stream()
+                .filter(x -> !x.isEmpty())
+                .map(x -> x.getHoverName().getString())
+                .forEach(armourNames::add);
+        Collections.reverse(armourNames); // helmet to boots order
+
+        List<String> accessoryNames = new ArrayList<>();
+        for (int i : ACCESSORY_SLOTS) {
+            ItemStack itemStack = McUtils.inventory().getItem(i);
+            if (!itemStack.isEmpty() && !itemStack.getHoverName().getString().equals("§7Accessory Slot")) {
```
kristofbolyai: Might as well extract this to a const field: `§7Accessory Slot`

common/src/main/java/com/wynntils/models/skillpoint/SavableSkillPointSet.java
```diff
@@ -0,0 +1,82 @@
...
+     * Constructs a new SavableSkillPointSet representing a full build with gear.
+     */
+    public SavableSkillPointSet(int[] skillPoints, List<String> armourNames, List<String> accessoryNames) {
+        this.strength = skillPoints[0];
+        this.dexterity = skillPoints[1];
+        this.intelligence = skillPoints[2];
+        this.defence = skillPoints[3];
+        this.agility = skillPoints[4];
+        this.armourNames = new ArrayList<>(armourNames);
+        this.accessoryNames = new ArrayList<>(accessoryNames);
```
kristofbolyai: These should be immutable lists to prevent any weird cases. That also means you can cache `isBuild`. Whenever you convert a saved skill set to a build, just recreate this object.

common/src/main/java/com/wynntils/handlers/container/scriptedquery/QueryBuilder.java
```diff
@@ -43,6 +43,13 @@ public QueryBuilder then(QueryStep step) {
         return this;
     }
 
+    public QueryBuilder thenIf(boolean condition, QueryStep step) {
+        if (condition) {
+            steps.add(step);
+        }
+        return this;
+    }
```
kristofbolyai: I'll get back to this, as we've discussed.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+            getAssignedSkillPoints(Skill.AGILITY)
+        });
+    }
+
+    public void loadLoadout(String name) {
+        McUtils.closeBackgroundContainer();
+
+        ScriptedContainerQuery query = ScriptedContainerQuery.builder("Loading Skill Point Loadout Query")
+                .onError(msg -> WynntilsMod.warn("Failed to load skill point loadout: " + msg))
+                .then(QueryStep.useItemInHotbar(CharacterModel.CHARACTER_INFO_SLOT - 1)
```
kristofbolyai: I don't think we need the -1 anywhere else?

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+                        .processIncomingContainer((container) -> loadSkillPointsOnServer(container, name)))
+                .build();
+        query.executeQuery();
+    }
+
+    private void loadSkillPointsOnServer(ContainerContent containerContent, String name) {
+        // we need to figure out which points we can subtract from first to actually allow assigning for positive points
+        Map<Skill, Integer> negatives = new EnumMap<>(Skill.class);
+        Map<Skill, Integer> positives = new EnumMap<>(Skill.class);
+        for (int i = 0; i < 5; i++) {
```
kristofbolyai:
```suggestion
        for (int i = 0; i < Skill.values().length; i++) {
```

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+
+            // no difference automatically dropped here
+            if (difference > 0) {
+                positives.put(Skill.values()[i], difference);
+            } else if (difference < 0) {
+                negatives.put(Skill.values()[i], difference);
+            }
+        }
+
+        AtomicBoolean confirmationCompleted = new AtomicBoolean(false);
```
kristofbolyai: If  you used regular loops, you would not need atomics.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+            }
+        });
+
+        // Server needs 2 ticks, give one extra to be safe
+        Managers.TickScheduler.scheduleLater(this::populateSkillPoints, 3);
+    }
+
+    public void deleteLoadout(String name) {
+        skillPointLoadouts.get().remove(name);
+    }
```
kristofbolyai: can you put after the save method?

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+
+    /**
+     * Closes any open containers (but not the screen shown) then queries compass (and tome menus) depending on if
+     * tomes have been unlocked.
+     */
+    public void populateSkillPoints() {
```
kristofbolyai: can you put this to the public methods, like after loadLoadout?

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+        McUtils.closeBackgroundContainer();
+
+        Managers.TickScheduler.scheduleNextTick(() -> {
+            assignedSkillPoints.clear();
```
kristofbolyai: I would rather you just set it to a new instance, which is usually safer in case something holds a reference to the object, we won't crash with CCME.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+            calculateGearSkillPoints();
+            queryTotalAndTomeSkillPoints();
+        });
+    }
+
+    private void calculateGearSkillPoints() {
+        gearSkillPoints.clear();
+        craftedSkillPoints.clear();
```
kristofbolyai: Same here

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+                craftedGear.getIdentifications().forEach(x -> {
+                    if (x.statType() instanceof SkillStatType skillStat) {
+                        craftedSkillPoints.merge(skillStat.getSkill(), x.value(), Integer::sum);
+                    }
+                });
+            } else if (!itemStack.isEmpty()
+                    && !itemStack.getHoverName().getString().equals("§7Accessory Slot")) {
+                WynntilsMod.warn("Skill Point Model failed to parse accessory: "
+                        + LoreUtils.getStringLore(McUtils.inventory().getItem(i)));
+            }
```
kristofbolyai: You can extract the highlighted part in both loops into a single method (taking the ItemStack as the arg)

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    /**
+     * Queries the compass (and tome) menu for skill point data.
+     */
+    private void queryTotalAndTomeSkillPoints() {
+        totalSkillPoints.clear();
+        tomeSkillPoints.clear();
```
kristofbolyai: Same here, new instance should be better

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
kristofbolyai: Oh, I see now! I think you confused this variable. This variable is used for getting the player info item from the character menu, not to open the character info menu. So just introduce a new field with the correct index.
image

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+        }
+
+        tomesUnlocked = LoreUtils.getStringLore(content.items().get(TOME_SLOT)).contains("✔");
+        if (!tomesUnlocked) {
+            calculateAssignedSkillPoints();
+        }
+    }
+
+    private void processTomeSkillPoints(ContainerContent content) {
+        tomeSkillPoints.clear();
```
kristofbolyai: .clear

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,386 @@
...
+        return assignedSkillPoints.getOrDefault(skill, 0);
+    }
+
+    public int getAssignedSum() {
+        return assignedSkillPoints.values().stream().reduce(0, Integer::sum);
+    }
+
+    public Map<String, SavableSkillPointSet> getLoadouts() {
+        return skillPointLoadouts.get();
+    }
```
kristofbolyai: public methods should be before private ones

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -0,0 +1,577 @@
...
+                            Models.SkillPoint.saveCurrentSkillPoints(name);
+                            populateLoadouts();
+                            setSelectedLoadout(new Pair<>(
+                                    name, Models.SkillPoint.getLoadouts().get(name)));
+                            saveNameInput.setTextBoxInput("");
+                            this.setMessage(Component.translatable("screens.wynntils.skillPointLoadouts.save"));
+                            // in case user pressed on both buttons
+                            saveBuildButton.setMessage(
+                                    Component.translatable("screens.wynntils.skillPointLoadouts.saveBuild"));
+                        }
```
kristofbolyai: We've always just created new classes extending WynntilsButton, instead of using anonymous classes. It is a bit of boilerplate, but it improves this class' readability quite a bit.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -0,0 +1,577 @@
...
+                            .withStyle(ChatFormatting.RED)));
+        }
+    }
+
+    public Pair<String, SavableSkillPointSet> getSelectedLoadout() {
+        return selectedLoadout;
+    }
+
+    private void populateLoadouts() {
+        loadoutWidgets.clear();
```
kristofbolyai: better to set a new instance here, as well

common/src/main/java/com/wynntils/models/skillpoint/SavableSkillPointSet.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public int getSkillPointsSum() {
+        return strength + dexterity + intelligence + defence + agility;
+    }
+
+    public int getMinimumCombatLevel() {
+        return (int) Math.ceil(getSkillPointsSum() / 2.0) + 1;
+    }
+
+    public boolean isBuild() {
```
magicus: This is true if we have any armor or accessories..? Why store it as a boolean?

kristofbolyai: Technically you don't even need this method, since the boolean is already "cached"

common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
```diff
@@ -66,7 +66,31 @@ public static void clickOnSlot(int clickedSlot, int containerId, int mouseButton
...
+    }
+
     public static void closeContainer(int containerId) {
         McUtils.sendPacket(new ServerboundContainerClosePacket(containerId));
     }
+
+    /**
+     * Closes invisible containers opened in the background, without closing the visible screen.
+     */
+    public static void closeBackgroundContainer() {
```
magicus, line 92: This sounds scary...

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
```diff
@@ -0,0 +1,388 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.skillpoint;
```
magicus: I'm thinking this belongs in `models.character`...

common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
ryanzhoudev: Can't use queries without it, unless we remove the open container check. It's still "opened" in the background when our custom screen is on top

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
ryanzhoudev: Honestly I think we already have too much irrelevant stuff in there... At first I was under the impression that CharacterModel meant what we would previously call ClassModel, just the data about the player's currently selected class. But then slowly we added death location, Silverbull, veteran stuff, etc.

common/src/main/java/com/wynntils/models/skillpoint/SavableSkillPointSet.java
magicus: Sorry, my comments were a bit tarse, I am lying in bed in a strange position and had diffuculty typing. :)

What I meant was that it seemed to me that this value was set to false iff both armor and accessories where empty. But this can as easily be checked on the arrays, and it does not seem to be necessary to cache this in a boolean field; the isBuild() method can check the two arrays directly.

common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
magicus: I'm too tired to fully understand this now. But if @kristofbolyai says it is fine, I guess I am too.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
kristofbolyai: You keep the model separate, just move the package, as it's still character related. That way the code is not getting messier. At least, I think this is what magicus meant.

common/src/main/java/com/wynntils/utils/wynn/ContainerUtils.java
kristofbolyai: I have a round of reviewing specifically for the query system changes to do, I'll make sure to check if this is the best we can do.

common/src/main/java/com/wynntils/models/skillpoint/SkillPointModel.java
magicus: Yes. While we often has the format "models.foo" contains "FooModel", I don't think this is really necessary. The package grouping is on a somewhat higher level, and can contain many independent, but related, models.

I thought the CharacterModel contained info about our current skill points, but maybe I remembered wrong? The skill point loadout should be located next to the model containing info on our current skill points anyway. If it is not the CharacterModel, then whatever model it is.

kristofbolyai: > I thought the CharacterModel contained info about our current skill points, but maybe I remembered wrong?

There is only a FIXME about adding support for it :)

magicus: *lol* 

Then maybe that fixme can be removed now? Or, at least if the skillpoint model is flexible enough to allow us to read the current values of our skill points. (I'm thinking they should basically be exported by the model so they can e.g. be available as functions).

common/src/main/java/com/wynntils/commands/BombCommand.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
```
magicus: How come all these files have been touched by spotless?

PR #2251, opened 2023-11-15
fix: Revert game message dialogue handling

ryanzhoudev:
I don't mean to push this through just when magicus is away, but it has been unaddressed for a while now. This makes item bombs, crates, and certain other system messages become dialogue, which breaks things such as claiming item bombs. We should revisit in the future, probably with a hardcoded list of dialogue, since there is no way to differentiate between the different types of messages.


ShadowCat117:
[This was the original issue](https://github.com/Wynntils/Artemis/issues/1951), are there any other known messages that should be counted as game messages? Perhaps we should compile a list to find a pattern. I did previously have the idea of checking if there's another chat format present it isn't a game message but that's purely based on this one known message.

Since it does more damage than good for the moment though it's probably worth reverting for now

magicus (approved):

ryanzhoudev:
Off the top of my head, the messages near the llevigar meteor (star thief) quest are also in a similar style.

PR #2253, opened 2023-11-26
feat: Add the new chat gear encoding, gear sharing GUI, item record GUI

### LINKED ISSUE #2246: [Proposal] - Artemis Item Encoding Standard
EDIT: The originally proposed format can be found [here](https://gist.github.com/kristofbolyai/6a6505e41aa245be3b01f0c6edd5a4da). The format below is always updated to reflect the current standard.

# Artemis Item Encoding Standard
The purpose of this new standard is to provide grounds for a new system, used for encoding Wynncraft items into strings, and vice-versa. Unlike the chat item system, this format is not limited to identified gear items, and can encode items of any (supported) types.

## Encoding
Encoding is a 2 layer process. The **first layer** is responsible for translating between **byte arrays** and **encoded values (UTF-16, base64)**. The **second layer** is responsible for translating **any kind of game data to an array of bytes**.
### Encoding a byte array to an encoded value
This section describes how the standard encodes integers. This encoding is not to be changed once implemented.
**Encoding to UTF-16:**
- An UTF-16 character in the [Supplementary Private Use Area-A](https://en.wikipedia.org/wiki/Supplementary_Private_Use_Area-A) can encode any value between `0x0` and `0xFFFD`. If we borrow the first two characters (`U+100000-U+100001`) from [Supplementary Private Use Area-B](https://en.wikipedia.org/wiki/Supplementary_Private_Use_Area-B), we can encode exactly 2 bytes of data into a single character. If our number of bytes needs to be padded, we use the [Supplementary Private Use Area-B](https://en.wikipedia.org/wiki/Supplementary_Private_Use_Area-B) to encode the first value, with a `0xEE` byte for padding, which will be ignored when decoding blocks with lengths.

**Encoding to base64**:
- A byte array is can be encoded into base64 easily, as base64 is made to represent bytes as text.
## Format
The encoded string format is represented by different kind of blocks. The order of the blocks is unspecified, with the exception of the start and end blocks, which must be first and last, respectively. Also note that a "type" block must be present, representing the type of the item being encoded. See the format of each block [below](#versions).
### Stability of the format
The format of the encoding itself is not to be changed once implemented. However, the blocks themselves can change their format between versions and blocks may be added and removed. See more information about versioning [below](#versions).
### Block Formats
A block consists of a unique header id between the range of 0-255 (256 possible values, 1 byte). The next bytes are the block's data, which is decoded while reading, since a block's data length is not explicitly encoded. See [versions](#versions) for specific block formats, and the unique header id.
## Encoding data
Some blocks use pre-defined ways of encoding data. These are described in this section.
#### Encoding a string:
Encoding a string is done by encoding the string's ASCII representation into bytes, and terminating with a null byte.
#### Encoding a variable sized integer:
A variable sized integer is encoded the following way: 7 bits is stored in the first byte. If the data fits in 7 bits, set the highest bit to `0`. If the data does not fit in a single byte, set the highest bit to `1`, add a next byte, with the same process. Repeat until the data encoded into the required number of bytes. Zigzag encoding is used to handle negative values.
## Versions
Versions are only incremented if there is a breaking change in the format of 1 or more blocks, and/or if 1 or more blocks are added or removed.
#### Start block
**ID**: `0`<br>**Integer size:** 8 bits<br>**Data**: A single byte, encoding the version of the data.
#### End block
**ID**: `255`
#### Type block
**ID**: `1`
**Integer size:** 8 bits<br>**Data**: A single byte, representing the type of item being encoded. This character works as a mapping key, each type has an id character representing it.<br>**Description:** Each type represents a single type of item. In most cases these ids are separated in the same way as Artemis' item classes, however exceptions might be allowed in cases where it is logical.
##### Key to Type Mapping Table

| Key | Type | Required blocks | Optional blocks |
| ---- | ---- | ---- | ---- |
| `0` | Gear Item | Name | Identifications, Powders, Shiny, Reroll |
| `1` | Tome Item | Name | Identifications |
| `2` | Charm Item | Name | Identifications |
| `3` | Crafted Gear Item | Custom Gear Type, Durability, Requirements | Name, Damage, Defense, Custom Identifications, Powders |
| `4` | Crafted Consumable Item | Custom Consumable Type, Uses, Requirements | Effects, Name, Custom Identifications |
| `5` | Crafted Item from Recipe | TODO | TODO |

#### Name block
**Header**: `2`<br>**Data**: The name of item is encoded as an encoded string. 
#### Identifications block
**Header**: `3`<br>**Data**: The first byte contains the number of non-pre-identified ids (referred to as N) for the item. The second byte contains the type of the identification info in the following blocks, this is the identification type flag. The following bytes contain all identification info. The size of a single identification info depends on the identification type flag.
##### Identification Type Flag
The purpose of this flag is to make sure ID encoding fits multiple purposes and give the clients and users control above the stability and size of the encoded data. Choosing a longer, extended id encoding flag allows clients to decode the shared data without any external sources, such as APIs, and makes the encoded data "stable", even if the current item specifications change. Choosing a shorter, normal id encoding flag is preferred in situations where data only needs to be available in shorter periods, but a shorter encoding is preferred, such as in-game chat.
###### Normal encoding
The byte-flag of this encoding is 0.
- **Pre-identified stats**: Pre-identified stats are not encoded. Injecting them back is an implementation detail for the client.
- **Normal stats**: Each identification takes 2 bytes to encode. The first byte is the numerical key of the ID (from a single, open-source, shared, mutually agreed upon source). The second byte is the calculated internal roll of the item.
###### Extended encoding
The byte-flag of this encoding is 1. The next byte is the number of pre-identified stats.

- **Pre-identified stats**: The first byte is the numerical key of the ID (from a single, open-source, shared, mutually agreed upon source). The following bytes are the encoded variable sized integer of the base value. Internal roll is not sent, as it does not make sense for pre-identified stats.
- **Normal stats**: The first byte is the numerical key of the ID (from a single, open-source, shared, mutually agreed upon source). The following bytes are the encoded variable sized integer of the base value. The last byte is the calculated internal roll of the item.
#### Powder block
**ID**: `4`<br>**Data**: The first byte is the powder slots on the item. The next byte is the number of bytes. The following bytes are a binary blob, padded to fit the nearest byte with `0` bits. A powder is encoded in 5 bits, with the following math: `element * 6 + tier`. The elements follow an `ETWFA` order. 5 `0` bits are used to represent that no powder is present at the slot.
If it is absent, all powder slots are assumed to be unpowdered.
If it is present, but it's length does not match the number of powder slots of the item, it is assumed that the rest of the slots are unpowdered.
#### Rerolls block
**ID**: `5`<br>**Data**: A single byte encoding the number of rerolls.
#### Shiny block
**ID**: `6`<br>**Data**: The first byte is the id of the shiny stat (from a single, open-source, shared, mutually agreed upon source). ID `0` is reserved for "Unknown". The next bytes are the encoded variable sized integer of the shiny value.
#### Custom Gear Type block
**ID**: `7`<br>**Data**: The data is a single byte, containing the id of the type of the item. See the ID map below.
##### Gear Type map

| ID | Type |
|----|------|
| 0  | Spear |
| 1  | Wand |
| 2  | Dagger |
| 3  | Bow |
| 4  | Relik |
| 5  | Ring |
| 6  | Bracelet |
| 7  | Necklace |
| 8 | Helmet |
| 9 | Chestplate |
| 10 | Leggings |
| 11 | Boots |
| 12 | Weapon* |
| 13 | Accessory* |
* Fallback types

#### Durability block
**ID**: `8`<br>**Data**: The first byte is the overall effectiveness of the identifications (the percentage next to the name for crafted items). The next bytes are the encoded variable sized integer of the maximum value. The next bytes are the encoded variable sized integer of the current value.
#### Requirements block
**ID**: `9`<br>**Data**: The first byte is the level requirement. The second byte is the class requirement, represented with an id. The next byte is the number of skill requirements. A skill requirement encoded as an id byte, representing the skill (`ETWFA` order). The next bytes are the encoded variable sized integer of the requirement values.
##### Class Requirement map

| ID | Type |
|----|------|
| 0  | None |
| 1  | Mage |
| 2  | Archer |
| 3  | Warrior |
| 4  | Assassin |
| 5  | Shaman |
#### Damage block
**ID**: `10`<br>**Data**: The first byte is the id of the attack speed of the item. The next byte is the number of attack damages present on the item. An attack damage is encoded the following way: The first byte is the id of the skill (`ETWFAN`, where N represents Neutral). The next bytes are the encoded variable sized integer of the minimum damage. The next bytes are the encoded variable sized integer of the maximum damage.
##### Attack Speed map

| ID | Type |
|----|------|
| 0  | Super Fast |
| 1  | Very Fast |
| 2  | Fast |
| 3  | Normal |
| 4  | Slow |
| 5  | Very Slow |
| 6  | Super Slow |
#### Defense block
**ID**: `11`<br>**Data**: The next bytes are the encoded variable sized integer of the health value.  The next byte is the number of defense stats present on the item. An defense stat is encoded the following way: The first byte is the id of the skill (`ETWFA`). The next bytes are the encoded variable sized integer of the defense value. 
#### Custom Identifications block
**ID**: `12`<br>**Data**: The first byte is the number of identifications. The identifications are encoded the following way: The first byte is the id of the identification. The next bytes are the encoded variable sized integer of the max value. For crafted items, the max values can be used to calculate the minimum values (10% of the maximum, rounded) and the current values (from the overall effectiveness).
#### Custom Consumable Type block
**ID**: `13`<br>**Data**: The data is a single byte, containing the id of the type of the item. See the ID map below.
##### Consumable Type map
| ID | Type |
| ---- | ---- |
| 0 | Potion |
| 1 | Food |
| 2 | Scroll |
| 3 | Consumable* |
* Fallback types

#### Uses block
**ID**: `14`<br>**Data**: The first byte is the remaining uses for the item. The second byte is the maximum uses for the item.

#### Effects block
**ID**: `15`<br>**Data**: The first byte is the number of effects. An effect is encoded the following way: The first byte is the id of the effect. The next bytes are the encoded variable sized integer of the effect's value.
##### Consumable Effect map
| ID | Type |
| ---- | ---- |
| 0 | Heal |
| 1 | Mana |
| 2 | Duration |
### Shiny Stat Table
https://github.com/Wynntils/Static-Storage/blob/main/Data-Storage/shiny_stats.json
### Identification ID-map Table
https://github.com/Wynntils/Static-Storage/blob/main/Reference/id_keys.json

kristofbolyai:
This PR/branch is set-up to track the progress on implementing the new gear encoding system.

## Details
The new system is designed to be a successor to the old chat encoding, but overall, it brings many changes and improvements:
- Encoding is separated into two layers (item -> byte[], byte[] -> string) to allow encoding into any kind of formats, like UTF-16, base64, binary, etc. This also allows the mod to reuse this kind of encoding later on, and allows third-parties to support encoding/decoding of chat items.
- Improve the encoded string data density (1 UTF-16 character = 2 bytes). While the new encoded gear strings might not be shorter, they still have better data density compared to the old format.
- Support encoding/decoding crafted items (or any custom gear items), as well as support for sharing crafted items as "recipes" (in-game support coming later)
- Improve regular item encoding by introducing both a normal and an extended encoding. Normal encoding behaves like the old chat encoding, but except that it encoding gear identifications, and does not rely on order. The extended encoding also encodes base values, making the encoded item independent from API versions.
- Improve regular item encoding by sharing info from new game mechanics, like shiny item ids.

The implementation details can be found in https://github.com/Wynntils/Artemis/issues/2246.

## Progress
The milestones below should mostly refer to separate PRs, but some might be completed by a single one.
- [x] Create the byte encoding layer (byte[] -> string)
- [x] Create an universal data source/API for shiny stat ids (with id-key mapping) and use it
- [x] Create an universal data source/API for gear stat ids (with id-key mapping) and use it
- [x] Implement regular gear encoding/decoding
- [x] Implement charm/tome encoding/decoding
- [x] Implement crafted gear encoding/decoding
- [x] Implement crafted consumable encoding/decoding
- [x] Gear sharing GUI
- [x] Item Storage GUI


fabric/src/test/java/TestUnsignedByte.java
```diff
@@ -0,0 +1,40 @@
...
+public class TestUnsignedByte {
+    @BeforeAll
+    public static void setup() {
+        SharedConstants.tryDetectVersion();
+        Bootstrap.bootStrap();
+    }
+
+    @Test
+    public void simpleByteArray_toUnsignedByteWorks() {
+        UnsignedByte[] bytes = UnsignedByte.of(new byte[] {(byte) 214, 121, 11, 49, 43, 75});
```
magicus, line 21: It's kind of only the first value that actually tests that the unsigned handling works... :)

What would be the expected outcome of sending in a negative valued byte? Should it be tested?

kristofbolyai: 214 is basically a negative byte though. It would be converted into it's unsigned representation. But negative bytes are not used by the encoding, only unsigned bytes 0-255.

magicus: Yes, that is why I wrote that it is only 214 that is doing any real testing. I should clarify: I think you should add more "negative" bytes, and especially the borderline cases like 127, 128, 0, 1. 

And my second question was: what would be the result of UnsignedByte.of(new byte[] { -1, -2, -127, -128, -129 })? I think it would be a useful test case to add.

kristofbolyai: Added some test cases for them :)

kristofbolyai:
@magicus This PR should be ready for you to review, whenever you have the time for it. It's still in draft, but it is only missing screen related additions, everything else works now. I know it's huge to review, that's why I am asking you earlier, so you have time to go through all commits.

common/src/main/java/com/wynntils/screens/itemsharing/ItemSharingScreen.java
```diff
@@ -0,0 +1,244 @@
...
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.components.AbstractWidget;
+import net.minecraft.client.gui.components.Button;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.client.resources.language.I18n;
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.item.ItemStack;
+
+public class ItemSharingScreen extends WynntilsScreen {
```
magicus: I don't get it. This looks like a merge error. Was this not supposed to be added by the item sharing PR?

magicus:
Thanks for the heads-up! I'll look through it as soon as I can. I will probably need several passes.

Just a first reflection after a quick skimming: there seems to be two kinds of changes here, adapations to the existing models (presumably to support new needed functionality), and actual changes related to the item encoding. Do you think it would make sense to separate the first part into a separate PR? Both for ease of reviewing, and for code history clarity. Then if something breaks, we will know that it was not the item encoding itself but the preparations, or the other way round.

magicus:
Or it might be that the other changes are just spurious artifacts. I don't understand the ItemSharingScreen, and there might be other similar problems.

common/src/main/java/com/wynntils/screens/itemsharing/ItemSharingScreen.java
ShadowCat117: I targetted this branch for the item sharing GUI since it uses the new encoding

magicus: Aha, I see. I missed that part. Thanks.

kristofbolyai:
> Thanks for the heads-up! I'll look through it as soon as I can. I will probably need several passes.
> 
> 
> 
> Just a first reflection after a quick skimming: there seems to be two kinds of changes here, adapations to the existing models (presumably to support new needed functionality), and actual changes related to the item encoding. Do you think it would make sense to separate the first part into a separate PR? Both for ease of reviewing, and for code history clarity. Then if something breaks, we will know that it was not the item encoding itself but the preparations, or the other way round.

Every relevant major adaptation was already merged into main in separate PRs. What's here is mostly small, unspecific leftovers. 

Additionally, the PR was made with separated specifically commits, so each commit can be treated as a PR, as all of them work and compile. Obviously, I can't keep force pushing for review changes, but at least, the main bits should be separated.

magicus:
Ok, fine. I'll review per commit.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
```diff
@@ -4,17 +4,76 @@
...
+            // Set the 5 bits for the powder
+            for (int j = 0; j < 5; j++) {
+                // The index is reversed because the bits are stored in reverse order
+                int index = powderDataIndex + (4 - j);
+                boolean value = ((element * 6 + tier) >> j) % 2 != 0;
+                powderData[index] = value;
+            }
+        }
+
+        // Pad the data to the nearest byte with '1' bits
```
magicus: Why not pad with 0? Seems more natural...

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
```diff
@@ -4,17 +4,76 @@
...
+        return !data.powders().isEmpty();
+    }
+
+    @Override
+    protected byte getId() {
+        return ID;
+    }
+
+    private ErrorOr<UnsignedByte[]> encodePowderData(PowderData data) {
+        // Powders are encoded as bits, a powder needs 5 bits to encode
```
magicus, line 47: So if we ever get a way to find the tier of applied powders, we will solve this by retiring this transformer, and creating a new with another ID and using that instead? Is that the general thinking?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/ShinyDataTransformer.java
```diff
@@ -7,14 +7,39 @@
...
+    @Override
+    protected byte getId() {
+        return ID;
+    }
+
+    private static UnsignedByte[] encodeShinyData(ShinyData data) {
+        UnsignedByte[] shinyStatValueBytes =
+                UnsignedByteUtils.encodeVariableSizedInteger(data.shinyStat().value());
+
+        UnsignedByte[] bytes = new UnsignedByte[shinyStatValueBytes.length + 2];
```
magicus: All the other non-trivial encodes had nice comments describing the format. I'd like to see that here to, it was very helpful.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
kristofbolyai: No, it supports powder tier encoding. So we just assume every powder is tier 6. Although yes, the general thinking would be to up the version, if some unsolvable conflict happens.

common/src/main/java/com/wynntils/utils/UnsignedByteUtils.java
```diff
@@ -0,0 +1,79 @@
...
+            bytes[i / 8] = UnsignedByte.of(value);
+        }
+        return bytes;
+    }
+
+    public static UnsignedByte[] encodeString(String string) {
+        // Strings are encoded by encoding the char's ASCII value
+        // and is terminated by a 0 byte
+        UnsignedByte[] bytes = new UnsignedByte[string.length() + 1];
+        byte[] asciiBytes = string.getBytes(StandardCharsets.US_ASCII);
```
magicus, line 49: What will happen if the string contains a character that is not representable in US ASCII? Do we get a runtime exception?

common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -109,16 +131,94 @@ public void onChatReceived(ChatMessageReceivedEvent e) {
         e.setMessage(modified.getComponent());
     }
 
-    private StyledTextPart createItemPart(GearItem gearItem) {
-        Style style = Style.EMPTY
-                .applyFormat(ChatFormatting.UNDERLINE)
-                .withColor(gearItem.getItemInfo().tier().getChatFormatting());
+    private void decoodeChatEncoding(List<StyledTextPart> changes, StyledTextPart partToReplace) {
```
magicus: typo

common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -109,16 +131,94 @@ public void onChatReceived(ChatMessageReceivedEvent e) {
...
+            changes.add(first);
+            changes.add(replacedPart);
+            changes.add(last);
+
+            partToReplace = last;
+            matcher = Models.Gear.gearChatEncodingMatcher(lastPart);
+        }
+    }
+
+    private void decoodeDeprecatedChatEncoding(List<StyledTextPart> changes, StyledTextPart partToReplace) {
```
magicus: typo

common/src/main/java/com/wynntils/models/items/encoding/impl/block/EndDataTransformer.java
```diff
@@ -12,8 +12,6 @@
 import com.wynntils.utils.type.UnsignedByte;
 
 public class EndDataTransformer extends DataTransformer<EndData> {
-    public static final byte ID = (byte) 255;
```
magicus: I kind of like having a systematic constant like this, rather than a hard-coded value in a getter. I think you should basically revert this commit, but make the ID fields private (and keep the fix in the registry).

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomGearTypeTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+import com.wynntils.models.items.encoding.type.ItemTransformingVersion;
+import com.wynntils.utils.type.ArrayReader;
+import com.wynntils.utils.type.ErrorOr;
+import com.wynntils.utils.type.Pair;
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.List;
+
+public class CustomGearTypeTransformer extends DataTransformer<CustomGearTypeData> {
+    private static final List<Pair<GearType, Integer>> GEAR_TYPE_IDS = List.of(
+            new Pair<>(GearType.SPEAR, 0),
```
magicus: Maybe these ids should be stored in the GearType itself?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomIdentificationDataTransformer.java
```diff
@@ -0,0 +1,102 @@
...
+    public ErrorOr<CustomIdentificationsData> decodeData(
+            ItemTransformingVersion version, ArrayReader<UnsignedByte> byteReader) {
+        return switch (version) {
+            case VERSION_1 -> decodeCustomIdentificationData(byteReader);
+        };
+    }
+
+    @Override
+    public byte getId() {
+        return 12;
```
magicus: ... and, to complement my previous comment, I'd like to see an ID field introduced here as well. (The "No magic numbers in code" rule).

common/src/main/java/com/wynntils/models/items/encoding/impl/block/DamageDataTransformer.java
```diff
@@ -0,0 +1,128 @@
...
+            bytes.add(UnsignedByte.of(damageTypeId));
+
+            // The next byte is the length of the integer that the minimum damage can fit in.
+            // The next bytes are the minimum damage bytes, which are assembled into an integer.
+            UnsignedByte[] unsignedBytes =
+                    UnsignedByteUtils.encodeVariableSizedInteger(damage.b().low());
+            bytes.add(UnsignedByte.of((byte) unsignedBytes.length));
+            bytes.addAll(List.of(unsignedBytes));
+
+            // The next byte is the length of the integer that the maximum damage can fit in.
```
magicus: Just a reflection, maybe it is too late to change it now, and maybe it really doesn't matter, but there exists more compact representations for variable sized integers, that are assumed to be normally very low. I'm thinking of methods like UTF-8. Basically, the process is like: 
Store 7 bits of the source data in the target byte. Use the highest bit to mark if we're done. If it is 0, then we're done, and the byte is exactly the value we wanted. If it is 1, use the 7 bits we have, and grab the next 7 bits from the next byte. If that byte's highest bit is 0, then we're done (the value could be stored in 14 bits), otherwise continue.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/DefenseDataTransformer.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public ErrorOr<DefenseData> decodeData(ItemTransformingVersion version, ArrayReader<UnsignedByte> byteReader) {
+        return switch (version) {
+            case VERSION_1 -> decodeDefenseData(byteReader);
+        };
+    }
+
+    @Override
+    public byte getId() {
+        return 11;
```
magicus: Another idea is to have a "global address enum", which contains the ID for all these data types. One upshot is that you avoid the "magic number in code" problem, and at the same time you get a overview of existing ids, and what the next free number would be. It also feels a bit contrary to the free-standing design currently used, but then again, a central rule coordinating numbers so they don't collide is necessary, so maybe it's better to actually respresent it as a enum class.

magicus: In fact, the more I think about it, the better I like this idea. Don't restore the old ID fields, create a enum instead with the responsibility of owning the ID numbers.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/DefenseDataTransformer.java
```diff
@@ -0,0 +1,96 @@
...
+            bytes.addAll(List.of(unsignedBytes));
+        }
+
+        return ErrorOr.of(bytes.toArray(new UnsignedByte[0]));
+    }
+
+    private ErrorOr<DefenseData> decodeDefenseData(ArrayReader<UnsignedByte> byteReader) {
+        // The first byte is the length of the integer that the health can fit in.
+        // The next bytes are the health bytes, which are assembled into an integer.
+        int length = byteReader.read().value();
```
magicus: This is a common pattern, reading a length byte + a variable sized integer. I recommend introducing a helper method to do just this. Then, if you decide to switch to another method of implementing variable-sized integers, it's just a single method to update.

(Basically, make it the responsibility of decodeVariableSizedInteger to read in the initial length byte also)

common/src/main/java/com/wynntils/models/items/encoding/impl/block/DurablityDataTransformer.java
```diff
@@ -0,0 +1,82 @@
...
+        // The next bytes are the maximum durability bytes, which are assembled into an integer.
+        int max = data.durability().max();
+        UnsignedByte[] unsignedBytes = UnsignedByteUtils.encodeVariableSizedInteger(max);
+        bytes.add(UnsignedByte.of((byte) unsignedBytes.length));
+        bytes.addAll(List.of(unsignedBytes));
+
+        // The next byte is the length of the integer that the current durability can fit in.
+        // The next bytes are the current durability bytes, which are assembled into an integer.
+        int current = data.durability().current();
+        unsignedBytes = UnsignedByteUtils.encodeVariableSizedInteger(current);
```
magicus, line 56: And conversely, make a helper for writing variable length integers.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/RequirementsDataTransformer.java
```diff
@@ -0,0 +1,128 @@
...
+import com.wynntils.utils.type.ErrorOr;
+import com.wynntils.utils.type.Pair;
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+public class RequirementsDataTransformer extends DataTransformer<RequirementsData> {
+    private static final List<Pair<ClassType, Integer>> CLASS_TYPE_IDS = List.of(
+            Pair.of(ClassType.NONE, 0),
```
magicus, line 25: Same question here, maybe put the id in ClassType?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomGearTypeTransformer.java
kristofbolyai: I had a general rule when creating these transformers. If I added enums that are item encoding/transformer specific, I store the IDs there. However, if I use enums from other parts of the mod, like gear enums, I'll map them to an id in the transformer itself. This is mainly done to not have a lot of enum have some kind of "magic" ID. What do you think, should I add them to the enums, or keep it this way?

common/src/main/java/com/wynntils/utils/UnsignedByteUtils.java
kristofbolyai: I'll make it check and throw a readable exception. It'll be passed up to the transformer and nicely displayed to the user.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/DamageDataTransformer.java
kristofbolyai: I've added it! It's actually great, since we encode a lot of numbers that fit 7 bits, so we save a fair amount of bytes. Implementation took a while, but I think it should be stable now :)

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
magicus: I think I'm missing where the powder tier is encoded then..? Can you point me to it?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
```diff
@@ -0,0 +1,148 @@
...
+            Pair<Powder, Integer> powder = data.powders().get(i);
+            int element = powder.key().ordinal();
+            int tier = powder.value();
+            int powderDataIndex = i * 5;
+
+            // Set the 5 bits for the powder
+            for (int j = 0; j < 5; j++) {
+                // The index is reversed because the bits are stored in reverse order
+                int index = powderDataIndex + (4 - j);
+                boolean value = ((element * 6 + tier) >> j) % 2 != 0;
```
kristofbolyai, line 69: @magicus this line.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomGearTypeTransformer.java
magicus: I think we can add the "magic" IDs to the enum. Compare it with e.g. the API name. In a way that is a "magic" key for the Wynncraft API. And this ID is just another mapping between our data types and some other external interface. As long as we're not downing in such interfaces, I think it makes sense to store it as part of our definition.

kristofbolyai: Okay, sure. It's done.

common/src/main/java/com/wynntils/models/items/encoding/type/EncodingSettings.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.encoding.type;
+
+/**
+ * Settings for encoding items.
+ * @param extendedIdentificationEncoding Whether to use extended identification encoding. This only affects identifiable items.
+ * @param shareItemName Whether to share the item name between the item name and the item lore. This only affects crafted and custom items.
```
magicus: "share the item name between the item name and the item lore". Sounds weird, not sure what is meant.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomConsumableTypeDataTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+import com.wynntils.models.items.encoding.type.ItemTransformingVersion;
+import com.wynntils.utils.type.ArrayReader;
+import com.wynntils.utils.type.ErrorOr;
+import com.wynntils.utils.type.Pair;
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.List;
+
+public class CustomConsumableTypeDataTransformer extends DataTransformer<CustomConsumableTypeData> {
+    private static final List<Pair<ConsumableType, Integer>> CONSUMABLE_TYPE_IDS = List.of(
+            new Pair<>(ConsumableType.POTION, 0),
```
magicus: Here's another one of the type/enum situations. Marking for reference; I'll get back to the general discussion on how to handle these separately.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/EffectsDataTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+import com.wynntils.utils.type.ArrayReader;
+import com.wynntils.utils.type.ErrorOr;
+import com.wynntils.utils.type.Pair;
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.ArrayList;
+import java.util.List;
+
+public class EffectsDataTransformer extends DataTransformer<EffectsData> {
+    private static final List<Pair<ConsumableEffect, Integer>> CONSUMABLE_EFFECT_IDS = List.of(
+            new Pair<>(ConsumableEffect.HEAL, 0),
```
magicus: ... and here as well...

common/src/main/java/com/wynntils/models/items/ItemEncodingModel.java
```diff
@@ -13,6 +15,12 @@
 import java.util.regex.Pattern;
 
 public class ItemEncodingModel extends Model {
+    @Persisted
```
magicus, line 18: This should have been configs, but we still cannot do configs in models, right? How do we change the value of these?

magicus: Ah, I see, in the item sharing gui...

common/src/main/java/com/wynntils/models/items/encoding/type/DataTransformerType.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.encoding.type;
+
+public enum DataTransformerType {
+    START_DATA_TRANSFORMER((byte) 0),
```
magicus: Maybe make the constructor take an int and do the casting in a single line only in the constructor?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
magicus: The comment above is now incorrect...

common/src/main/java/com/wynntils/utils/UnsignedByteUtils.java
```diff
@@ -69,71 +70,55 @@ public static String decodeString(List<UnsignedByte> bytes) {
...
-            bytes = new UnsignedByte[8];
-            bytes[0] = UnsignedByte.of((byte) (value >> 56));
-            bytes[1] = UnsignedByte.of((byte) (value >> 48));
-            bytes[2] = UnsignedByte.of((byte) (value >> 40));
-            bytes[3] = UnsignedByte.of((byte) (value >> 32));
-            bytes[4] = UnsignedByte.of((byte) (value >> 24));
-            bytes[5] = UnsignedByte.of((byte) (value >> 16));
-            bytes[6] = UnsignedByte.of((byte) (value >> 8));
-            bytes[7] = UnsignedByte.of((byte) value);
+        // Use zig-zag encoding to encode negative numbers
```
magicus, line 73: TIL!

common/src/main/java/com/wynntils/models/gear/type/GearType.java
```diff
@@ -38,34 +39,39 @@ public enum GearType {
...
+            List.of(Items.CHAINMAIL_BOOTS, Items.IRON_BOOTS, Items.GOLDEN_BOOTS, Items.DIAMOND_BOOTS),
+            11),
+    MASTERY_TOME(null, Items.ENCHANTED_BOOK, 0, -1),
+    CHARM(null, Items.CLAY_BALL, 0, -1);
 
     private final ClassType classReq;
     private final Item defaultItem;
     private final int defaultDamage;
     private final List<Item> otherItems;
+    private final int id;
```
magicus: Maybe name this and related methods etc `encodingId` to clarify?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/PowderDataTransformer.java
magicus: Ah, thanks!

common/src/main/java/com/wynntils/models/items/encoding/DataTransformerRegistry.java
```diff
@@ -0,0 +1,196 @@
...
+            try {
+                ErrorOr<UnsignedByte[]> errorOrEncodedData = encodeData(version, itemData);
+                if (errorOrEncodedData.hasError()) {
+                    return ErrorOr.error(errorOrEncodedData.getError());
+                }
+
+                bytes.addAll(Arrays.asList(errorOrEncodedData.getValue()));
+            } catch (Exception e) {
+                WynntilsMod.error(
+                        "Failed to encode data class " + itemData.getClass().getSimpleName() + "!", e);
```
magicus: A dedup is hiding here, waiting :)

common/src/main/java/com/wynntils/models/items/encoding/DataTransformerRegistry.java
```diff
@@ -0,0 +1,196 @@
...
+
+                ErrorOr<ItemData> errorOrData = dataTransformer.decodeData(version, byteReader);
+
+                if (errorOrData.hasError()) {
+                    return ErrorOr.error(errorOrData.getError());
+                }
+
+                dataList.add(errorOrData.getValue());
+            } catch (Exception e) {
+                WynntilsMod.error("Failed to decode data block with id " + dataBlockId.value() + "!", e);
```
magicus: ... and here

magicus: I see this is a common pattern. Maybe introduce a `ErrorOr.errorAndLog()`, `ErrorOr.loggedError()` or something like that?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/CustomConsumableTypeDataTransformer.java
kristofbolyai: You are reviewing an older commit, this is fixed now.

common/src/main/java/com/wynntils/models/items/encoding/type/DataTransformerType.java
kristofbolyai: Sure.

common/src/main/java/com/wynntils/models/gear/type/GearType.java
kristofbolyai: Only here, or every other enum?

common/src/main/java/com/wynntils/models/items/encoding/DataTransformerRegistry.java
kristofbolyai: I think it's rather weird to call it like that. Maybe perhaps add a method like `errorInstance.log()`? Or `WynntilsMod#error(ErrorOr<>)`?

common/src/main/java/com/wynntils/utils/UnsignedByteUtils.java
kristofbolyai: The amount of black magic used here makes me a bit nervous, even though the method is tested throughly with many cases for longs, ints, signs, et..

common/src/main/java/com/wynntils/models/items/encoding/DataTransformerRegistry.java
magicus: I don't really know. I reacted to the duplication of the string. Either you can like create a `String errorMsg = "...."` and then call the log and ErrorOr with the string. Or you can create a common helper that basically does that. Whatever you want to call that, I don't really have any good suggestions.

magicus:
I just want to say that I have finished going through all individual commits (up until the last three new), but I also want to review the complete PR as well. I started doing that but was able to get only half-way through. I have been busy today and will be tomorrow as well; maybe I can get some time tomorrow night.

common/src/main/java/com/wynntils/screens/itemsharing/ItemSharingScreen.java
```diff
@@ -0,0 +1,260 @@
...
+            McUtils.sendMessageToClient(Component.translatable("screens.wynntils.itemSharing.copied")
+                    .withStyle(ChatFormatting.GREEN));
+        } else {
+            // TODO: Save item to storage
+        }
+    }
+
+    private void addSharingOptions() {
+        // region Checkbox options
+        if (wynnItem instanceof GearItem) {
```
ShadowCat117: This checkbox should probably be shown for charms and tomes too right? Or are they unlikely to change

kristofbolyai:
> I just want to say that I have finished going through all individual commits (up until the last three new), but I also want to review the complete PR as well. I started doing that but was able to get only half-way through. I have been busy today and will be tomorrow as well; maybe I can get some time tomorrow night.

Sounds good. Plan is to merge this and the skill point loadout PR this weekend, bump to Artemis 1.0.0 and introduce a new versioning system. We may fit the item vault UI in this PR, if it is ready this week, we will see about  that.

`Artemis [bumps with mc version].[bumps with major features].[bumps with major fixes]-beta.[bumps with everything else]`.`

EDIT: Check out https://github.com/Wynntils/Artemis/pull/2313

This versioning scheme would work with the current state of Athena and Artemis, and allows us to easily add version streams without managing separate branches, as users can prefer to update on major/minor/all/etc versions.

kristofbolyai:
I also realized that I meant to reply this 2 days ago. In fact, I opened Github Mobile and this comment welcomed me, unsent. Sorry about it!

Anyways, I want to target Sunday for the release, if you want to have a final look.

magicus:
Making a final look now. Maybe you should take this out of draft at some point..? ;-)

common/src/main/java/com/wynntils/screens/itemsharing/ItemSharingScreen.java
kristofbolyai: Yes. I'll patch it!

common/src/main/java/com/wynntils/utils/type/ErrorOr.java
```diff
@@ -41,6 +43,15 @@ public boolean hasError() {
         return error != null;
     }
 
+    public <T> ErrorOr<T> logged() {
```
magicus, line 46: This was a very elegant solution!

If we start to use this in other contexts, we can probably remove the throw clause and just do a no-op if there is no error.

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -256,7 +258,9 @@ public NamedValue getValue(FunctionArguments arguments) {
             Optional<ShinyStat> shinyStatOpt = Models.Shiny.getShinyStat(itemStack);
             if (shinyStatOpt.isEmpty()) return NamedValue.EMPTY;
 
-            return new NamedValue(shinyStatOpt.get().name(), shinyStatOpt.get().value());
+            // FIXME: The function system can't handle longs, so we have to cast to int
```
magicus, line 261: This irks me a bit. Are we planning to add longs to the function system? Can't we just remove the comment otherwise? Or are shinyStat values likely to overflow an int? If so, maybe we need to fix it. Otherwise, maybe it should return a int instead of a long.

Not a big thing, but it kind of bugs me.

common/src/main/java/com/wynntils/models/gear/type/ConsumableType.java
```diff
@@ -1,13 +1,19 @@
...
 
 public enum ConsumableType {
-    POTION,
-    FOOD,
-    SCROLL;
+    POTION(0),
+    FOOD(1),
+    SCROLL(2);
+
+    private final int id;
```
magicus: Should this be `encodingId` too?

common/src/main/java/com/wynntils/models/gear/type/GearType.java
```diff
@@ -95,6 +101,13 @@ public static GearType fromClassType(ClassType classType) {
         return null;
     }
 
+    public static GearType fromId(int id) {
```
magicus:
```suggestion
    public static GearType fromEncodingId(int encodingId) {
```

common/src/main/java/com/wynntils/models/items/encoding/impl/block/EndDataTransformer.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public ErrorOr<EndData> decodeData(ItemTransformingVersion version, ArrayReader<UnsignedByte> byteReader) {
+        // End data is always empty
+        return ErrorOr.of(new EndData());
+    }
+
+    @Override
+    public byte getId() {
+        return (byte) 255;
```
magicus: Return a `DataTransformerType` value?

common/src/main/java/com/wynntils/models/items/encoding/impl/block/RequirementsDataTransformer.java
magicus: I'm not sure why this was resolved..?

common/src/main/java/com/wynntils/utils/EncodedByteBuffer.java
```diff
@@ -0,0 +1,135 @@
...
+            UnsignedByte secondByte = UnsignedByte.of((byte) (values & 0xFF));
+
+            bytes.add(firstByte);
+            bytes.add(secondByte);
+
+            // Only 0xF0000-0xFFFFD are used
+            assert codePoint < 0xFFFFE : "Invalid code point: " + codePoint;
+        }
+
+        return fromBytes(bytes.toArray(new UnsignedByte[0]));
```
magicus: I think Java nowadays supports the more modern/logical:

```
        return fromBytes(bytes.toArray(UnsignedByte::new));
```

common/src/main/java/com/wynntils/models/items/encoding/impl/block/NameDataTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public ErrorOr<NameData> decodeData(ItemTransformingVersion version, ArrayReader<UnsignedByte> byteReader) {
+        return switch (version) {
+            case VERSION_1 -> decodeName(byteReader);
+        };
+    }
+
+    private ErrorOr<UnsignedByte[]> encodeName(String name) {
+        try {
+            return ErrorOr.of(UnsignedByteUtils.encodeString(name));
```
magicus, line 35: Wasn't the point that the name should not be byte-encoded, so at least it will show for non-Wynntils users? Or maybe I'm missing something in the grand scheme of things how the chat encoding packet is sent...

magicus (approved):
*phew* I have finally examined the complete PR in detail. Overall it looks good, nah, great! I added a few remarks, if you want to take care of them before pushing it would be nice, but they are not really blockers.

I am still a bit amazed at how the code exploded from like a single method in a feature to a huge framework. But then again, I cannot see how we could have done it differently either.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/RequirementsDataTransformer.java
magicus: To clarify, I mean that ClassType should have an `encodingId` field containing these numbers.

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
kristofbolyai: > Or are shinyStat values likely to overflow an int?

Could be the case. We have stats like "mobs killed". What if they add "experience collected"? These are sure candidates for overflow.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/NameDataTransformer.java
kristofbolyai: It was the point yes. But it turned out to be really messy, so I opted to just encode the name. Hopefully it is not a big issue, and if it is, we can fix it up with a new version.

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
magicus: Ok, let's leave the fixme for now then. Maybe open a bug to track that we need to handle this properly in the future.

common/src/main/java/com/wynntils/models/items/encoding/impl/block/NameDataTransformer.java
magicus: Perhaps if we put the name as part of the start or end packet it gets simpler? I mean, if we want a non-encoded string it is more part of the packet itself, not the individual data blocks. So maybe we could sneak it in in the start block, and when we encode the start block, we extract the name and send it separately..?

magicus: If we want to do something like that, it is probably best to do it now, since that changes the entire packet concept, rather than an individual data packet. Or, hm, maybe we should do both. That is, both keep this for the actual name, and then add some kind of "prefix" to the packet containing a "user representation", like `<start user representation byte>User visible name for non-wynntils users<the entire encoded packet>`.

kristofbolyai: Let me share my thinking. Normally, I would say, yes, let's do this, give everyone the best experience. But this case, it's really tricky to get the code quality/system right, without breaking major abstractions. Vanilla users feel left out, yes. However, they are basically left out either way, Wynncraft items tend to be special not because of rarity, but their stats. So knowing that "I am selling x mythic" is not any better than knowing "I am selling x" without knowing it's stats.

The whole `EncodedByteBuffer` concept relies on the fact that everything is encoded. This (sadly) includes names too. We even "gain" a little, we can encode 2 ASCII chars into a single Unicode character.

If this becomes a problem for Wynncraft, I'll personally work on fixing it. However, I think we should keep it as is for now.

PR #2254, opened 2023-11-27
fix: Detect soul points in hunted

ShadowCat117:
image
image2

Diamond axe is used for a lot of items, so potentially in the future we could detect if the player is in hunted mode and then determine if soul points are a nether star or diamond axe. Also noticed the fixme for veteran soul points so thought I'd add that too, though I can't verify for certain the non champion vet colours

### DIFF
--- a/common/src/main/java/com/wynntils/models/character/CharacterModel.java
+++ b/common/src/main/java/com/wynntils/models/character/CharacterModel.java
@@ -204,9 +204,7 @@ public final class CharacterModel extends Model {
 
         Matcher veteran = LoreUtils.matchLoreLine(rankSubscriptionItem, 0, VETERAN_PATTERN);
 
-        if (veteran.matches()) {
-            isVeteran = true;
-        }
+        isVeteran = veteran.matches();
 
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -194,6 +202,12 @@ private void parseCharacterContainer(ContainerContent container) {
     private void parseCratesBombsCosmeticsContainer(ContainerContent container) {
         ItemStack rankSubscriptionItem = container.items().get(RANK_SUBSCRIPTION_INFO_SLOT);
 
+        Matcher veteran = LoreUtils.matchLoreLine(rankSubscriptionItem, 0, VETERAN_PATTERN);
+
+        if (veteran.matches()) {
+            isVeteran = true;
+        }
```
kristofbolyai:
```suggestion
        isVeteran = veteran.matches();
```

kristofbolyai (approved):
Otherwise looks good.

magicus (approved):
LGTM

magicus (approved):

PR #2255, opened 2023-11-27
feat: Add universal support for encoding byte arrays

kristofbolyai:
EncodedByteBuffer is can be used for encoding and kind of byte[] data into formats like UTF-16 and base64. This is used as the base for the new gear chat encoding.

### DIFF
--- a/common/src/main/java/com/wynntils/core/encoding/EncodedByteBuffer.java
+++ b/common/src/main/java/com/wynntils/core/encoding/EncodedByteBuffer.java
@@ -4,25 +4,114 @@
  */
 package com.wynntils.core.encoding;
 
-// byte[] <-> UTF-16, Base64, etc.
-public final class EncodedByteBuffer {
-    private byte[] bytes;
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
 
-    private EncodedByteBuffer(byte[] bytes) {
+/**
+ * A buffer of bytes that can be encoded and decoded to various formats.
+ */
+public final class EncodedByteBuffer {
+    private static final int PRIVATE_USE_AREA_A_START = 0xF0000;
+    private static final int PRIVATE_USE_AREA_B_START = 0x100000;
+
+    private UnsignedByte[] bytes;
+
+    private EncodedByteBuffer(UnsignedByte[] bytes) {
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/core/encoding/EncodedByteBuffer.java
```diff
@@ -4,25 +4,114 @@
...
-// byte[] <-> UTF-16, Base64, etc.
+import com.wynntils.utils.type.UnsignedByte;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+
+/**
+ * A buffer of bytes that can be encoded and decoded to various formats.
+ */
 public final class EncodedByteBuffer {
```
magicus: I must have missed when we got this class. It feels more "utils" than "core" to me.

kristofbolyai: This PR targets a separate branch for gear encoding, so it doesn't exist on main.

magicus: Ah. Nevertheless, maybe consider moving encoding stuff to utils? Or are you planning to include much more in this package, so it really will make sense to be a part of core?

kristofbolyai: I've done some basic modeling in the PR this class was added, but I can agree with you. The main reason I put it in core is because I plan to use this class throughout Artemis for encoding different kind of data (like POIs maybe).

kristofbolyai: I think it's best leaving it here until the original branch is ready and check back on this. If I add nothing else here, I am going to move it.

magicus: Even fundamental types in Artemis (like Pair) have not resided in core. Maybe this was wrong, but if so we should move all fundamental types. Until we do that, I think this better belong with other more "general" types.

magicus: Ok, sounds like a fair solution.

kristofbolyai: I ended up moving the class here.

PR #2257, opened 2023-11-27
feat: Use ShinyStatTypes instead of strings for shiny stats

kristofbolyai:
Moving ShinyModel might not be necessary, but  I see no point in not including it in the real stats package, instead of the character stats package.


common/src/main/java/com/wynntils/models/stats/type/ShinyStatType.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.stats.type;
+
+public record ShinyStatType(int id, String key, String displayName, StatUnit statUnit) {
+    public static final ShinyStatType UNKNOWN = new ShinyStatType(0, "unknown", "Unknown", StatUnit.RAW);
```
magicus, line 8: Does the count of real shiny stats start at 1?

magicus (approved):
Looks fine. I agree that ShinyModel should live among stats.

common/src/main/java/com/wynntils/models/stats/type/ShinyStatType.java
kristofbolyai: Yes.

PR #2258, opened 2023-11-27
chore: Move to using Wynntils/Static-Storage

kristofbolyai:
With this PR, we fully moved to using [Static-Storage](https://github.com/Wynntils/Static-Storage) as our singular source of data, which I created to organize all of our data in a way we discussed earlier.

Edit: Do NOT merge this PR. I want to be available in case anything breaks.


kristofbolyai:
@magicus Aaand we just entered your territory :). I focused on making the static storage repository as close to our original idea as I could. Please leave any comments regarding the repository either here or in a new issue in the repo.

magicus:
I checked the repo now, and it looks great! Clean and well-separated into different parts. Thank you so much for doing this!

However, I do not fully understand this PR... I'll look through it commit-by-commit to see if I understand the intention better.

common/src/main/resources/assets/wynntils/urls.json
```diff
@@ -98,86 +90,86 @@
...
-    "url": "https://raw.githubusercontent.com/Wynntils/WynntilsWebsite-API/master/lootrun_tasks_named.json"
+    "url": "https://raw.githubusercontent.com/Wynntils/Static-Storage/main/Data-Storage/lootrun_tasks_named.json"
+  },
+  {
+    "id": "dataStaticShinyStats",
+    "md5": "2dbc0fdde50512b64828d5642bb480b6",
+    "url": "https://raw.githubusercontent.com/Wynntils/Static-Storage/main/Data-Storage/shiny_stats.json"
+  },
+  {
+    "id": "dataStaticAbilities",
```
magicus: I think we've used alphabetic sorting on `id` in the rest of the file. At least, the intention was to do it. Maybe it needs to be rechecked in other places as well.

magicus:
Ok, I understand the changes now. You cleaned out old parts of urls.json that are not used anymore. I guess we were hesitant to do that before since the file could be downloaded by older versions of Artemis, but if you're moving it then all bets are off anyway, so good chance to clean it up.

magicus (approved):

common/src/main/resources/assets/wynntils/urls.json
kristofbolyai: It's formatted in a random order in most places though :). I will fix it up.

kristofbolyai: It should sort automatically now with spotless, but I had to give up 2 space indentation, as I can only make it do 4 spaces. I think we should still keep it though :)

magicus: I like that you fix automatic tools to enforce sorting! However, I think this change needs be made in Static-Storage instead, since we will just grab urls.json from there if it differs.

Also, I wonder if you can have Spotless pipe it further down to gson, which is what is enforcing the 2 space indentation we have on all other json files. It would be a bit of a shame if we couldn't get that consistent.

kristofbolyai: Pushed a workflow in the Static-Storage repository. It also fixed the indentation. (And no, spotless only seems to format by the first rule it finds.)

PR #2259, opened 2023-12-02
feat: Migrate to using Item API v3 for items and ingredients

kristofbolyai:
This moves all of our item infrastructure to V3. Processing new API item data (charms, tomes, tools, materials) comes in later PRs.


common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -120,11 +120,11 @@ private void checkMajorIdCircles(Player player) {
         // Offset the radius slightly so multiple circles can be shown for each player
         // Only a few major IDs can actually be applied at the same time, but we make this general
         List<Pair<CustomColor, Float>> circles = validGear.stream()
-                .flatMap(gearInfo -> gearInfo.fixedStats().majorIds().stream().map(majorId -> switch (majorId.id()) {
-                    case "TAUNT" -> Pair.of(CommonColors.ORANGE.withAlpha(TRANSPARENCY), 12f);
-                    case "HERO" -> Pair.of(CommonColors.WHITE.withAlpha(TRANSPARENCY), 8f);
-                    case "ALTRUISM" -> Pair.of(CommonColors.PINK.withAlpha(TRANSPARENCY), 8.1f);
-                    case "GUARDIAN" -> Pair.of(CommonColors.RED.withAlpha(TRANSPARENCY), 7.9f);
+                .flatMap(gearInfo -> gearInfo.fixedStats().majorIds().stream().map(majorId -> switch (majorId.name()) {
```
magicus, line 123: Have Wynncraft dropped the concept of an "id" from a major id, as opposed to a name? Some of these have been renamed in the past, so that sounds like a bad idea from their side.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -147,45 +123,21 @@ private void loadGearRegistry() {
...
                 throws JsonParseException {
             JsonObject json = jsonElement.getAsJsonObject();
 
-            // Wynncraft API has two fields: name and displayName. The former is the "api name", and is
-            // always present, the latter is only present if it differs from the api name.
-            // We want to store this the other way around: We always want a displayName (as the "name"),
-            // but if it the apiName is different, we want to store it separately
+            // Wynncraft API name field includes "unnormalized" characters like ֎, we want to remove them
             String primaryName = json.get("name").getAsString();
-            String secondaryName = JsonUtils.getNullableJsonString(json, "displayName");
```
magicus, line 180: Did they remove the distinction of name/displayName? How do they handle the conflict where there were some item that originally was called "Foo" but now is called "Bar" in the user visible code (but still "Foo" in the gear view screen thingie), and there is a new item that is called "Foo" in user visible code.

Sorry for the handwaving, I have forgotten the details. But I know there were some problematic items.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -147,45 +123,21 @@ private void loadGearRegistry() {
...
-            if (secondaryName == null) {
-                String normalizedApiName = WynnUtils.normalizeBadString(primaryName);
-                if (!normalizedApiName.equals(primaryName)) {
-                    // Normalization removed a ֎ from the name. This means we want to
-                    // treat the name as apiName and the normalized name as display name
-                    secondaryName = normalizedApiName;
-                }
+            String normalizedApiName = WynnUtils.normalizeBadString(primaryName);
+            if (!normalizedApiName.equals(primaryName)) {
+                // Normalization removed a ֎ from the name. This means we want to
```
magicus: Can this still happen with the new API?

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -379,39 +379,34 @@ private Optional<String> parseQuest(JsonObject json) {
...
+
+            int healthBuff = JsonUtils.getNullableJsonInt(baseStats, "health");
+            List<Pair<Skill, Integer>> skillBonuses = parseSkillBonuses(identifications);
             String attackSpeedStr = JsonUtils.getNullableJsonString(json, "attackSpeed");
             Optional<GearAttackSpeed> attackSpeed = Optional.ofNullable(GearAttackSpeed.fromString(attackSpeedStr));
 
-            List<GearMajorId> majorIds = parseMajorIds(json);
-            List<Pair<DamageType, RangedValue>> damages = parseDamages(json);
-            List<Pair<Element, Integer>> defences = parseDefences(json);
+            Optional<GearMajorId> majorIds = parseMajorIds(json);
```
magicus, line 387: How come majorIds is not a list anymore? Can we ever only have a single major ID? Was this always the case but the old code was overly general?

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -379,39 +379,34 @@ private Optional<String> parseQuest(JsonObject json) {
...
+            return Optional.of(new GearMajorId(
+                    majorIdsJson.get("name").getAsString(),
+                    StyledText.fromString(majorIdsJson.get("description").getAsString())));
         }
 
         private List<Pair<Skill, Integer>> parseSkillBonuses(JsonObject json) {
             List<Pair<Skill, Integer>> list = new ArrayList<>();
             for (Skill skill : Skill.values()) {
-                String skillBonusApiName = skill.getApiName() + "Points";
+                String skillBonusApiName = "raw" + com.wynntils.utils.StringUtils.capitalizeFirst(skill.getApiName());
```
magicus, line 407: Okaaay... stat key names have been a constant source of error. Have you verified that this works properly on the new data? Or rather, more generally, have you checked that you do not encounter stat key names that you cannot properly parse? I remember that we had a problem before of just silently ignoring stat keys that we parsed incorrectly, since we did not actually parse every single field in the json object, instead we construct keys we want to look at to pick up.

I remember vaguely that I re-created the json and compared it to the one I read in. I don't remember if I kept that as part of the code, or if it was just debug code that I threw away. Anyway, if we don't have that right now, I recommend you re-instate it.

common/src/main/java/com/wynntils/models/gear/type/GearTier.java
```diff
@@ -13,7 +13,7 @@
 import net.minecraft.network.chat.Component;
 
 public enum GearTier {
-    NORMAL(ChatFormatting.WHITE, 0, 0.0f),
+    NORMAL(ChatFormatting.WHITE, 0, 0.0f, "common"),
```
magicus, line 16: If Wynn is renaming the base tier to "common", maybe we should follow suite? Or are there places in our code that assume the enum name is "normal"?

common/src/main/java/com/wynntils/models/wynnitem/type/ItemObtainType.java
```diff
@@ -4,16 +4,21 @@
...
     NORMAL_MOB_DROP("normal", "Unspecified Mob Drop"), // mob drops (and any loot chest, afaict)
+    CHALLENGE("challenge", "Challenge"), // at the moment, only Legendary Island
+    EVENT("event", "Event"), // like Bonfire, Heroes, etc.
+    LOOTRUN("lootrun", "Lootrun"), // lootrun
     UNKNOWN("never", "Unknown"), // more information is needed
 
-    // From crowd sourced data
-    BOSS_ALTAR("bossaltar", "Boss Altar"),
+    // From crowd sourced data or Wynncraft API
+    BOSS_ALTAR(List.of("bossaltar", "altar"), "Boss Altar"),
```
magicus, line 19: It's a bit sad if our data has a different key. At some point, we should update it to match Wynn's tag, so we don't need this list. Also, perhaps Wynn's list of obtain info is more complete now? Do we really still need our crowd sourced data?

magicus:
Overall I think this looks pretty good. Nice to see Wynn up their game on the quality of the API data.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
kristofbolyai: We used to have an external json mapping ids to names, descriptions. The API now provides the name-description mapping for each item, so there is no need for fetching the external source (also removed this PR). Since they don't include IDs, I opted to remove them (as they had no other use).

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
kristofbolyai: Yes, so basically the case is the same as it used to, except they fixed all mis-matched names. But items like `Cancer֎` still exist, since the old Cancer item is not called `Necrosis`. But The API mostly correctly returns all names, and those that are not normalised are handled below.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
kristofbolyai: https://github.com/Wynntils/Artemis/pull/2259#discussion_r1413412419

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
kristofbolyai: I am not sure if the old API had provided a list, or if this was just a design decision, but the new API's data structure quite clearly suggest that an item has no intention of having multiple major ids.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
kristofbolyai: No, the new naming scheme is quite clear now, and it does not seem to have edge-cases. I have not checked every misc stat by hand though, however I do plan to do that after I also ported ingredients. Then I would most likely write some kind of tool to check stat names, instead of scrolling a 100 pages of trade market :)

common/src/main/java/com/wynntils/models/gear/type/GearTier.java
kristofbolyai: I think "common" is not used anywhere except the api, so I think it's fine to not rename this, at least until we don't see more of these renames.

common/src/main/java/com/wynntils/models/wynnitem/type/ItemObtainType.java
kristofbolyai: I tried to get rid of the crowd-sourced, partially outdated data, but it is still much better than the info API provides, so we can not get away with it.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -466,11 +463,41 @@ private List<Pair<Element, Integer>> parseDefences(JsonObject json) {
...
+            JsonObject identificationsJson = json.get("identifications").getAsJsonObject();
+            boolean preIdentifiedItem = JsonUtils.getNullableJsonBoolean(json, "identified");
+
+            for (Map.Entry<String, JsonElement> entry : identificationsJson.entrySet()) {
+                StatType statType = Models.Stat.fromApiRollId(entry.getKey());
+
+                if (statType == null) {
+                    WynntilsMod.warn("Item DB contains invalid stat type " + entry.getKey());
+                    continue;
+                }
```
kristofbolyai, line 478: This re-imagined way of decoding the identifications was taken from IngredientInfoRegistry, it allows us to log every stat that we can't parse. This means I could fix all stats, without doing any in-game testing (apart from internal ids when checking other players' IDs, which was always a bit of a guessing game.)

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
magicus: I don't still get it. Using your examples, I thought this was the case: (using your example)

Item 1: displayName: `Necrosis`, internalRollName: `Cancer`
item 2: displayName: `Cancer`, internalRollName:`Cancer֎`

I fail to understand how they can tell us that an item should be shown to the user as `Necrosis`, but called `Cancer` in the internal roll, if they only have a single name provided per item.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
magicus: Great, that was kind of what I was looking for.

common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
kristofbolyai: Yes, it's not right. I brought this up with the API devs, hopefully we can get the old name/displayName system back.

kristofbolyai: We've managed to get it back into the API, this should be fixed now.

PR #2261, opened 2023-12-04
chore: Bump loom and other dependencies. Remove Quilt support.

kristofbolyai:
The bumps are basically self-explanatory, the new loom and fabric loader has some fixes that are generally useful.

It's time to remove explicit quilt support. Since they still do not have an updated API for 1.20.2, we can consider them not up-to-date from now on. Most likely we will still support quilt, with the fabric builds, but explicit support has to be dropped.


kristofbolyai:
I would also want to discuss moving to NeoForge here. We can either do this when releasing the mod for the next game version, or now, by bumping to 0.5-beta, and removing LexForge support.

NeoForge's event bus is greatly improved, not requiring us to do any black-magic to use it. Additionally, the event system itself has been improved greatly. With the new event bus being a hard-fork, it is understandable why we have to chose between the two forge loaders. NeoForge has been also working closely with Fabric to standardize some systems, and gave us new features like loader-bundled MixinExtras. NeoForge has marked a build stable for 1.20.2.

It also has to be said that we can't support newer LexForge versions anyways, so users are stuck using 48.0.13<=.

With this said, I vote to bump our version to 0.5-beta, and move from LexForge to NeoForge. This would be the most minimal impact to our users, as only Forge users would have anything to do. We would benefit in many ways, like not having to bundle MixinExtras, having the improved EventBus.

@magicus

gradle.properties
```diff
@@ -58,24 +51,20 @@ antiope_version=0.2.2
...
 ### DEVELOPMENT SUPPORT
 
 # DevAuth
 # Check for latest at https://github.com/DJtheRedstoner/DevAuth/releases
 devauth_version=1.1.2
 
-# Vineflower for Loom
-# Check for latest at https://github.com/Juuxel/loom-vineflower/releases
-loom_vineflower_version=1.11.0
-
```
kristofbolyai, line 72: Loom auto includes vineflower now.

magicus:
I hate this Minecraft community drama. :-( So, we can't support both NeoForge and LexForge?

Is NeoForge "winning", that is, are folks leaving LexForge and migrating over to NeoForge en masse? Last I checked my impression was that a the original Forge (LexForge now, apparently) still had quite a following, and retained a lot of the good developers.

kristofbolyai:
> I hate this Minecraft community drama. :-( So, we can't support both NeoForge and LexForge?
> 
> 
> 
> Is NeoForge "winning", that is, are folks leaving LexForge and migrating over to NeoForge en masse? Last I checked my impression was that a the original Forge (LexForge now, apparently) still had quite a following, and retained a lot of the good developers.

NeoForge seems to be more community friendly and a lot of suggestions make it into NeoForge, so it seems like it is taking priority.

magicus:
But if a large percentage is using LexForge, maybe it doesn't matter so much if NeoForge makes a lot of changes... Quilt also made a lot of changes after they forked, and look at them now.

Is it impossible to support both NeoForge and LexForge?

kristofbolyai:
The event buses are hard forked, so they work differently. We will have to monitor the situation now, I guess..

PR #2265, opened 2023-12-06
fix: Fix chat gear encoding after Item API V3

kristofbolyai:
This is a quick and dirty patch to fix chat encoding, until the new system is implemented. The second commit is a proper fix, an implementation detail that I overlooked.


common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
```diff
@@ -18,104 +18,108 @@
...
 
 public final class StatListOrderer {
-    // Legacy order was defined by Athena as a fixed list. This was missing several stat types; I have
-    // tried filling them in into "logical" places
+    // Legacy order was defined by Athena as a fixed list.
+    // This was missing several stat types;
+    // I have tried filling them in into "logical" places.
+    // NOTE: This list was changed to use Artemis keys,
+    //       as a hotfix to support Item API V3.
+    //       This will be removed once the new encoding is implemented.
```
magicus, line 26: I'm not sure we should remove it, even if it is not used for the chat. People might have been used to the way Wynntils Legacy ordered stats in the GUI, and want to keep it like that. Remember that this is the primary reason for the ordering of stats, the chat thing is mostly a side-effect.

kristofbolyai: We can reconstruct that legacy ordering as actual stat type lists, so renaming stats won't break it.

magicus: That sounds better.

PR #2266, opened 2023-12-07
fix: Fix ingredient durability being multiplied by 1000

kristofbolyai:


common/src/main/java/com/wynntils/models/ingredients/IngredientInfoRegistry.java
```diff
@@ -157,7 +157,9 @@ public IngredientInfo deserialize(JsonElement jsonElement, Type jsonType, JsonDe
 
             // Get items-only parts
             JsonObject itemIdsJson = JsonUtils.getNullableJsonObject(json, "itemOnlyIDs");
-            int durabilityModifier = JsonUtils.getNullableJsonInt(itemIdsJson, "durabilityModifier");
+            // Durability modifier is multiplied by 1000 in the API, so we divide it here
+            // (there is no known reason for this)
```
magicus, line 161: We should not report it as a bug on the API instead?

kristofbolyai: Zeer said it's not technically a bug, although he has no idea why it's like that.

magicus: 🤷

PR #2267, opened 2023-12-07
feat: Parse tomes as gears, show tome identification percentages

kristofbolyai:
This PR is best reviewed per commit, to understand the underlying refactors needed for it to work. Some commits have descriptions, to help understanding the commits. The size of the PR is not that huge as it seems, especially when going through commits 1 by 1.

Tomes are really close to actual gear items, but they need some adaptations, hence the refactors. In the end, the new system behaves in a much more adaptive way, allowing us to implement new item types effectively.

Some screenshots:
image
image
image
image
image
image


common/src/main/java/com/wynntils/models/gear/GearInfoRegistry.java
```diff
@@ -89,7 +68,7 @@ private void loadGearRegistry() {
         Download dl = Managers.Net.download(UrlId.DATA_STATIC_GEAR_ADVANCED);
         dl.handleJsonObject(json -> {
             Gson gson = new GsonBuilder()
-                    .registerTypeHierarchyAdapter(GearInfo.class, new GearInfoDeserializer())
+                    .registerTypeHierarchyAdapter(GearInfo.class, new GearInfoDeserizalier())
```
magicus: typo in the name

common/src/main/java/com/wynntils/models/gear/type/AbstractItemInfoDeserializer.java
```diff
@@ -0,0 +1,416 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.gear.type;
```
magicus: Not sure the type package is the right place for this. We usually reserve it for interfaces, enums and records.

common/src/main/java/com/wynntils/models/gear/type/AbstractItemInfoDeserializer.java
```diff
@@ -0,0 +1,416 @@
...
+        }
+
+        return List.copyOf(obtainInfo);
+    }
+
+    protected Optional<StyledText> parseLore(JsonObject json) {
+        String lore = JsonUtils.getNullableJsonString(json, "lore");
+        if (lore == null) return Optional.empty();
+
+        // Some lore contain like "\\[Community Event Winner\\]", fix that
```
magicus: Just noticed this: Is this still relevant in the v3 api?

common/src/main/java/com/wynntils/models/tooltip/IdentifiableItemInfo.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.tooltip;
```
magicus: move to `type` subpackage.

common/src/main/java/com/wynntils/models/tooltip/TooltipIdentificationDecorator.java
```diff
@@ -2,14 +2,13 @@
  * Copyright © Wynntils 2023.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
-package com.wynntils.models.gear.tooltip;
+package com.wynntils.models.tooltip;
```
magicus: Should also be in `type`, I think.

common/src/main/java/com/wynntils/models/tooltip/TooltipStyle.java
```diff
@@ -2,11 +2,11 @@
  * Copyright © Wynntils 2023.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
-package com.wynntils.models.gear.tooltip;
+package com.wynntils.models.tooltip;
```
magicus: `type`

common/src/main/java/com/wynntils/models/gear/type/AbstractItemInfoDeserializer.java
magicus: I think it better belongs in `models.wynnitem`, an "abstract" model I tried to create to unify similar item kinds. I think this has the same kind of vibe to it.

magicus:
Call me crazy, but I'm seriously thinking that the new `tooltip` model would be better of as a `handler` than a model. According to the javadoc guidelines:

```
 * a Model.
```

and

```
 * Handlers span the bridge between Minecraft and Wynncraft. They manage a certain
 * aspect of Minecraft functionality, and with some Wynncraft knowledge, they distribute
 * the MC content to the actual Wynncraft models. A handler is needed when there is not a
 * clear 1-to-1 relationship between Minecraft components and models.
```

In this case it is not really "dispatching", but I think it is the same core idea. We need to provide a service to several models to be able to build tooltips. That is not so much a model per se, as a support layer for the models.

magicus:
Overall, I think this look pretty nifty!

common/src/main/java/com/wynntils/models/gear/type/AbstractItemInfoDeserializer.java
kristofbolyai: Removed it, it's not.

kristofbolyai:
I also moved the TooltipModel to be a TooltipHandler. IMO it feels just as weird, but I can see how this is theoretically a Handler, it's just that we haven't made any of this kind.

magicus (approved):

PR #2271, opened 2023-12-09
feat: Parse charms as gears, show charm identification percentages 

kristofbolyai:
Review by commit. I tried to make it integrate as neatly as I could, but I am aware that some parts are a bit more shoehorned than we would like it to be ideally, but I do believe this is the nicest way to integrate this overall, with minor modifications.

Some screenshots:
Note that charm ids are inverted differently than spell stats. This means they also have 30-130 as the internal roll range, not taking the base value sign into account. They also have the best effects on the lowest internal rolls.
image
image
image
image

Other items:
image
image


common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -190,11 +190,11 @@ public static WynnItemParseResult parseItemStack(
                 String starString = statMatcher.group(5);
                 String statDisplayName = statMatcher.group(6);
 
+                // Skill bonuses looks like stats when parsing, ignore them
+                if (Skill.isSkill(statDisplayName)) continue;
```
kristofbolyai: This is a low hanging optimization, which I've done to make the performance impact of the regex-based display names lookups close to 0.

common/src/main/java/com/wynntils/models/stats/builders/MiscStatKind.java
```diff
@@ -44,30 +45,50 @@ public enum MiscStatKind {
...
-    GATHERING_XP("Gathering XP", StatUnit.PERCENT, "gatheringXP", "GATHERING_XP"),
-    DUNGEON_XP("Dungeon XP", StatUnit.PERCENT, "dungeonXP", "DUNGEON_XP");
+    SLAYING_XP("Slaying XP", StatUnit.PERCENT, "slayingXP", "SLAYING_XP", StatType.SpecialStatType.TOME_BASE_STAT),
+    GATHERING_XP(
+            "Gathering XP", StatUnit.PERCENT, "gatheringXP", "GATHERING_XP", StatType.SpecialStatType.TOME_BASE_STAT),
+    DUNGEON_XP("Dungeon XP", StatUnit.PERCENT, "dungeonXP", "DUNGEON_XP", StatType.SpecialStatType.TOME_BASE_STAT),
+
+    // Charm stats
+    LEVELED_XP_BONUS(
+            "XP from Lv. ${} content",
```
magicus, line 55: So this is more like a "stat template", where you can have different instances with specific levels? That really messes up our system, doesn't it? We'l need to store the level as part of the stat "key", not as the stat value...

kristofbolyai: It only had to be patched in StatModel for getting the display name from string and for getting a stat from display name. Although it is kind of hacky, I am not sure I can see a more obvious way forward atm.

magicus:
I'm honestly not sure I understood how you are handling the level-variable stats. I will need to read through the code again when I'm more sharp. Apart from that, it looks fine I guess, it was just more of a "standard" conversion.

common/src/main/java/com/wynntils/models/stats/StatModel.java
kristofbolyai, line 1: @magicus This class' changes show everything that needs to be changed to handle leveled stats.

magicus:
Okay, so you're using the levelRange as a requirement on the entire charm, and then you inject this when displaying a certain stat for that charm? This seems to be mandated by the API design, I guess. So if Wynn ever introduces more stats that have a level range, that level range will have to be the same for all stats on that particular charm. (Or they will have to redo the API.)

I guess this works.

magicus (approved):

kristofbolyai:
@magicus Do you have time to review this, or should I self-review and merge?

kristofbolyai:
I've self-reviewed this, should be good. I will merge this before the announcement on discord.

PR #2272, opened 2023-12-09
chore: Fix mod game version being wrongly published on Modrinth/Curseforge

kristofbolyai:


DevScyu (approved):

.github/workflows/beta-release.yml
```diff
@@ -135,6 +135,7 @@ jobs:
           name: Wynntils Artemis (${{ matrix.modloader }}) ${{ needs.changelog.outputs.tag }}
           version: ${{ needs.changelog.outputs.tag }}
           version-type: beta
+          game-versions: 1.20.2
```
magicus, line 138: It would be nice to read this off some gradle config file. It is easy to forget to change this value...

kristofbolyai: This is a temporary fix most likely, as we didn't need this before.

magicus: I'm guessing it doesn't hurt to have that information here, if Modrinth starts acting up again... But what will hurt is if we forget to update this number.

kristofbolyai: I usually do CTRL-Find our old version in the project, when I update to a new version. Hopefully we won't forget, and if we do, we'll fix this :)

PR #2275, opened 2023-12-10
refactor: Make IdentifiableItemInfo be an item property

kristofbolyai:
While I started working on gear encoding again, I realized, I am going to need to access `IdentifiableItemInfo` not just for tooltips, but for encoding any gear-like item. I originally just wanted to move the class from the tooltip handler, but while searching for the correct model, I realized, we could just make it an item property.


kristofbolyai:
Do you think this makes sense, @magicus?

magicus (approved):
I think this makes very much sense.

kristofbolyai:
> I think this makes very much sense.

Hope so :). I ended up modelling the whole based encoding on item properties..

PR #2280, opened 2023-12-11
refactor: Treat SkillStatType as a regular stat

kristofbolyai:
This refactor is not only needed for the gear encoding, I think it makes a lot of sense. I can see why skills were handled specially, as a fixed stat, but I think it is really obvious from this PR that it needed too much special handling, for no real benefit. Now that the V3 API makes it easy to tell the difference between "pre-identified" and not "pre-identified" stats, we can move skill stats back to every other stat. (Legacy API only had a difference between "pre-identified" items)

### DIFF
--- a/common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
+++ b/common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
@@ -8,6 +8,7 @@ import com.wynntils.models.stats.builders.MiscStatKind;
 import com.wynntils.models.stats.type.DamageStatType;
 import com.wynntils.models.stats.type.DefenceStatType;
 import com.wynntils.models.stats.type.MiscStatType;
+import com.wynntils.models.stats.type.SkillStatType;
 import com.wynntils.models.stats.type.SpellStatType;
 import com.wynntils.models.stats.type.StatListDelimiter;
 import com.wynntils.models.stats.type.StatListOrdering;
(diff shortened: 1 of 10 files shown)


kristofbolyai:
Throwing the ball to you, again @magicus. Sorry if it has been more PRs lately, but you are still father of the gear system in Artemis, which I've been doing surgery on for the last few days :).

common/src/main/java/com/wynntils/models/wynnitem/AbstractItemInfoDeserializer.java
```diff
@@ -416,8 +393,8 @@ protected List<Pair<StatType, StatPossibleValues>> parseVariableStats(JsonObject
             }
             // Range will always be stored such as "low" means "worst possible value" and
             // "high" means "best possible value".
-            RangedValue range = StatCalculator.calculatePossibleValuesRange(
-                    baseValue, preIdentifiedItem, statType.showAsInverted());
+            RangedValue range =
+                    StatCalculator.calculatePossibleValuesRange(baseValue, preIdentified, statType.showAsInverted());
```
kristofbolyai, line 397: This is actually a bugfix for a typo, I already fixed this either in the gear encoding or the charm PR, but it is really obviously broken in this case, so I promoted the fix here.

common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
```diff
@@ -183,13 +188,16 @@ private static List<StatType> createDefaultOrdering(
...
     private static List<StatType> createWynncraftOrdering(
+            List<SkillStatType> skillStats,
             List<MiscStatType> miscStats,
             List<DefenceStatType> defenceStats,
             List<DamageStatType> damageStats,
             List<SpellStatType> spellStats) {
         List<StatType> wynncraftOrdering = new ArrayList<>();
 
         // Wynncraft order seem to have grown a bit haphazardly
+        wynncraftOrdering.addAll(skillStats);
```
magicus, line 199: This will match the Wynncraft ordering? I'm assuming the skill stats where sorted in some canonical skill order so this would have been okay before, so this will still be okay..?

magicus (approved):

common/src/main/java/com/wynntils/models/stats/StatListOrderer.java
kristofbolyai: Wynn seems to do the ordering we have in the enum itself, with the added separator.

Issue #1791, opened 2023-07-21
Allow loading some resource packs ONLY when joining wynncraft


MenacingPerson:
For example, some packs like the Fungal Pack, Smaller Meteors/Ice Snake, Dark Mode GUI, etc

It would be a setting in wynntils that loads these alongside the wynncraft resource pack, and also hopefully on top of it.

The purpose is that they won't be loaded when not on wynncraft, as some like fungal pack can break other servers.

Using Wynntils auto-apply RPs, it would apply them on top of the wynntils one, which is not even possible using other mods currently.

geekazodium:
how do you suggest this work, my current idea is to just add a button to the texturepacks when displayed in the menu that allow you to toggle if a texturepack should only load if you are on wynncraft.
I'm suggesting something like this: (white shaded areas are where the buttons will appear)

![Screenshot 2023-08-13 165150](https://github.com/Wynntils/Artemis/assets/59382423/29aaeb7f-de48-4389-bec9-5c66bfa7bf90)

geekazodium:
alternative idea is to have text on one of the corners of the screen saying shift-click to make texturepack wynncraft only and shade in the texturepack you are hovering your cursor over be highlighted differently when you hold shift, and when a texture pack is toggled to wynncraft only it would show up with a different background color.

magicus:
Is this a feature request to basically "bundle" a bunch of other local resource pack that should be applied at the same time as the Wynncraft resource pack? Either when the server asks you to install the pack upon joining, or when you load Minecraft (if you have the load resource pack on start feature enabled)?

MenacingPerson:
> Is this a feature request to basically "bundle" a bunch of other local resource pack that should be applied at the same time as the Wynncraft resource pack?

Yes, exactly

But it applies them above the wynntils/wynncraft texture pack, so that the wynncraft textures can be overriden by the local resource packs

> my current idea is to just add a button to the texturepacks when displayed in the menu that allow you to toggle if a texturepack should only load if you are on wynncraft.

why a button? A config in the wynntils book should be enough, right?

geekazodium:
> why a button? A config in the wynntils book should be enough, right?

I was suggesting a an option in the texture packs menu as that would be where you would configure texture packs anyways.

however adding a config in the wynntils book is also fine, I don't know which idea would be better. (personally I'd prefer navigating this in the resourcepacks menu, as I don't quite like navigating the wynntils book in general but that's just me.)

[...]

PR #2281, opened 2023-12-12
feat: allows packs to be applied on top of wynn

geekazodium:
part 1/2 for #1791 
allows packs to be applied on top of wynncraft textures.
 "Using Wynntils auto-apply RPs, it would apply them on top of the wynntils one, which is not even possible using other mods currently."

I'm working on part 2/2 to actually auto apply resource packs.

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/AutoApplyResourcePackFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/AutoApplyResourcePackFeature.java
@@ -8,6 +8,7 @@ import com.wynntils.core.components.Services;
 import com.wynntils.core.consumers.features.Feature;
 import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.ConfigCategory;
+import com.wynntils.mc.event.PackGetFixedPositionEvent;
 import com.wynntils.mc.event.ResourcePackEvent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
@@ -27,5 +28,10 @@ public class AutoApplyResourcePackFeature extends Feature {
             // Use this resource pack as our preloaded pack
             Services.ResourcePack.setRequestedPreloadHash(packHash);
         }
+    }
+
+    @SubscribeEvent
+    public void onResourcePackFixedPosition(PackGetFixedPositionEvent event) {
+        if (event.getPack().getId().equals("server")) event.setFixedPosition(false);
     }
 }
--- a/common/src/main/java/com/wynntils/mc/event/PackGetFixedPositionEvent.java
+++ b/common/src/main/java/com/wynntils/mc/event/PackGetFixedPositionEvent.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright © Wynntils 2023.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.event;
+
+import net.minecraft.server.packs.repository.Pack;
+import net.minecraftforge.eventbus.api.Event;
+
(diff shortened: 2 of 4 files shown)


geekazodium:
update, this is currently broken, I need to stop making pull requests until it stops crashing

geekazodium:
I got a lot of errors saying "Signature is missing from Property textures" unsure if it's caused by changes, however feature seems to be functioning as normal.

geekazodium:
> I got a lot of errors saying "Signature is missing from Property textures" unsure if it's caused by changes, however feature seems to be functioning as normal.

update, probably not.

kristofbolyai:
> > I got a lot of errors saying "Signature is missing from Property textures" unsure if it's caused by changes, however feature seems to be functioning as normal.
> 
> update, probably not.

It's not.

kristofbolyai (commented):
I have to tell you, before pointing out all the code changes you need, do we know that this actually works fine internally? It seems really flaky to me to just enable moving around a server pack. I fear that MC hardcodes/expects it to be the "lowest" pack in some places. What do you think?

kristofbolyai:
What does the server pack unlocker mod do? If it is similar, maybe this works for now.

geekazodium:
> I have to tell you, before pointing out all the code changes you need, do we know that this actually works fine internally? It seems really flaky to me to just enable moving around a server pack. I fear that MC hardcodes/expects it to be the "lowest" pack in some places. What do you think?

I tested this and it seems to work perfectly fine when I loaded programmer art textures on top of wynncraft textures, I mean it is possible but the fact that I didn't see any broken/missing textures when I did this suggests otherwise. (I tested it with fabric)

geekazodium:
I'll quickly test this with forge as well
(am I dumb or does autoapply not work for forge???)

kristofbolyai:
I've checked server pack unlocker source, and it seems that they save the server pack, and load it "regularly". I think that is a safer bet as well, since you can't really rely on just "allowing" the order to be changed and hoping that it will work as expected.

geekazodium:
okay, that makes sense.

geekazodium:
I'll get to fixing that in a bit, currently kind of busy with another project right now, if someone else wants to do this that's fine, I don't really know how long I will be busy for.

geekazodium:
@kristofbolyai I'm really confused as to how simply unlocking it causes problems, I don't fully understand the texturepack handling of minecraft and I just... don't get it.

kristofbolyai:
> @kristofbolyai I'm really confused as to how simply unlocking it causes problems, I don't fully understand the texturepack handling of minecraft and I just... don't get it.

The minecraft code has some special handling for different kind of categories, mainly: build-in resource packs, server resource packs and downloaded/regular resource packs. For this reason, server pack unlocker removes server pack loading, and injects them back as regular resource packs.

A secondary reason why I would not want to add this, is that in 1.20.3, they reworked how server packs work and how they are treated internally. I don't want to find myself in deep problems whenever we move to the next game version.

kristofbolyai:
I'll close this PR now. If you think you have a suitable solution, please feel free to reopen it, but I don't think this solution should be merged.

geekazodium:
@kristofbolyai 
thank you so much for explaining, I hope I'm not being a burden to this project as I'm still very much a beginner in contributing to projects and have little understanding of the resource packs related parts of Minecraft.

kristofbolyai:
> @kristofbolyai 
> 
> thank you so much for explaining, I hope I'm not being a burden to this project as I'm still very much a beginner in contributing to projects and have little understanding of the resource packs related parts of Minecraft.
> 
> 

No problem! I hope that in the next MC version we'll use, this won't be a problem, or that we can fix it.

PR #2282, opened 2023-12-12
refactor: In preparation for charm identifications, add support for custom stat type ranges, rework the inverted stat system, other fixes to internal rolls

kristofbolyai:
Review per commit.

Charm identifications propose some new challenges, that the current system was not prepared to solve. This PR solves all of these issues, hopefully giving us a smooth release for charm identifications, when the API is fixed.

What changed:
- Stat types now provide their identification range, as charm ids can have custom ranges, like 80-115.
- Stat types now provide whether they are calculated as inverted values. This option was present before, but a new option, "display as inverted" was added, which enables us to easily invert ids that have negative values with positive meaning, which is also an issue for charm ids.
- We now use the API's provided range to fact-check our calculated ranges from the base values. This PR was also tested this way. Even if the API is not correct, it is better to fact-check against the API, than nothing.
- Fixed the internal name for some spell cost stats. This broke when migrating to V3, but only affected the other players' gear viewer.
- **Fixed internal range calculation for negative bases (and rare incorrect values in for positive bases). See the commit details and below for details .**
- **Fixed the perfect/increased/decreased roll chances (which was broken in legacy as well...)**

**Fixed internal roll calculations:**
Observe the two main issues:
- Negative base value internal rolls used to be flipped
- Some positive values were rounded incorrectly (See Aquamarine's Health Regen)

Old:
image
image
image
New: 
image
image
image

**New chance calculations:**
Observe the new perfect roll chance, that does not actually want to create a 3 star item, rather a maximum possible roll.
Observe the new increase/decrease chances, (that do not add up to 100%, further verifying that it is correct, since it accounts for the chance that the ID stays the sane).

image
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
@@ -70,6 +70,9 @@ public class ContainerSearchFeature extends Feature {
     public final Config<Boolean> filterInScrapMenu = new Config<>(true);
 
     @Persisted
+    public final Config<Boolean> filterInPetMenu = new Config<>(true);
+
+    @Persisted
     public final Config<CustomColor> highlightColor = new Config<>(CommonColors.MAGENTA);
 
     // If the guild bank has lots of custom (crafted) items, it can take multiple packets and a decent amount of time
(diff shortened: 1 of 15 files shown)


kristofbolyai:
I would like to order one more item stat PR for @magicus, please :)!

magicus:
I'll look at it tomorrow. If I don't, you have my blessing to intergrate anyway and I'll take the blame :)

kristofbolyai:
> I'll look at it tomorrow. If I don't, you have my blessing to intergrate anyway and I'll take the blame :)

I am rebasing the charm PR onto this PR, that should give you some time while I wait for the last API bugfix.

PR #2289, opened 2023-12-22
feat: Add new account types [skip ci]

DevScyu:
Add new account types for nametags


kristofbolyai (approved):
Looks good.

kristofbolyai:
Does this code crash on earlier patches? I hope not..

DevScyu:
I don't know if previous builds will crash if the client encounters a user with a account type that isn't in that enum. But I'm not planning on giving the new account types for a while, just wanted to get the client side out in case it does crash old builds

kristofbolyai:
> I don't know if previous builds will crash if the client encounters a user with a account type that isn't in that enum. But I'm not planning on giving the new account types for a while, just wanted to get the client side out in case it does crash old builds

I'll make sure to check that. In the meantime, let's merge this.

PR #2299, opened 2023-12-26
feat: Add support for custom crafted tooltips

kristofbolyai:
This is done to support rendering crafted items decoded from chat.

Some screenshots: 
image
image
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/tooltips/ItemStatInfoFeature.java
+++ b/common/src/main/java/com/wynntils/features/tooltips/ItemStatInfoFeature.java
@@ -19,6 +19,7 @@ import com.wynntils.mc.event.ItemTooltipRenderEvent;
 import com.wynntils.models.items.WynnItem;
 import com.wynntils.models.items.WynnItemData;
 import com.wynntils.models.items.properties.IdentifiableItemProperty;
+import com.wynntils.models.items.properties.NamedItemProperty;
 import com.wynntils.models.stats.StatCalculator;
 import com.wynntils.models.stats.type.StatActualValue;
 import com.wynntils.models.stats.type.StatListOrdering;
@@ -31,12 +32,14 @@ import com.wynntils.utils.wynn.ColorScaleUtils;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import net.minecraft.ChatFormatting;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 import org.lwjgl.glfw.GLFW;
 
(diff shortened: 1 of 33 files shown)


kristofbolyai:
I am not 100% sold on this code yet. I feel like it duplicates too much, although I find it really messy to combine crafted item and identifiable (gear, charm, tome) item tooltip generation into the same code. So perhaps, this is the best we can do for now.

kristofbolyai:
Should be ready now. I know there is a bit of duplication, happens when you have really different items with almost the same type of tooltips.. Maybe we can fix it eventually, but for now, this is what we need to progress.

common/src/main/java/com/wynntils/features/tooltips/ItemStatInfoFeature.java
```diff
@@ -90,37 +97,37 @@ public void onTooltipPre(ItemTooltipRenderEvent.Pre event) {
...
+                    Models.Item.asWynnItemProperty(event.getItemStack(), IdentifiableItemProperty.class);
+            if (identifiableItemPropertyOpt.isPresent()) {
+                tooltips =
+                        getIdentifiableItemTooltip(event.getItemStack(), wynnItem, identifiableItemPropertyOpt.get());
+            }
+
+            Optional<CraftedItemProperty> craftedItemPropertyOpt =
+                    Models.Item.asWynnItemProperty(event.getItemStack(), CraftedItemProperty.class);
+            if (craftedItemPropertyOpt.isPresent()) {
+                tooltips = getCraftedItemTooltip(event.getItemStack(), wynnItem, craftedItemPropertyOpt.get());
```
magicus, line 113: This is kind of an "else" situation, right? If we've already gotten a tooltip, we can't get here.

common/src/main/java/com/wynntils/handlers/tooltip/impl/crafted/CraftedTooltipComponent.java
```diff
@@ -0,0 +1,27 @@
...
+
+public abstract class CraftedTooltipComponent<T extends CraftedItemProperty> {
+    public abstract List<Component> buildHeaderTooltip(T craftedItem);
+
+    public abstract List<Component> buildFooterTooltip(T craftedItem);
+
+    protected MutableComponent buildRequirementLine(String requirementName, boolean fulfilled) {
+        MutableComponent requirement;
+
+        requirement = fulfilled
```
magicus, line 21: This checkmark/cross selection from a bool should be a helper function...

common/src/main/java/com/wynntils/features/tooltips/ItemStatInfoFeature.java
kristofbolyai: Yes, an item is either:
- identifiable: these are gear, tome, charms. items you identify/roll with the npc
- crafted items in essence also have stats/identifications, but they work differently, and we can't mix them with regular gear.

Since the relationship between crafted and normal gear is "so close yet so far", the tooltip building got into a bit of a messier situation. Overall, the code should be fine, code duplication is not big either.

common/src/main/java/com/wynntils/handlers/tooltip/impl/crafted/CraftedTooltipComponent.java
kristofbolyai: It's been like that for a bit, I've just copied the method. There are a few refactors the whole tooltip building system could take, I opted to only change what's really needed.

PR #2301, opened 2023-12-27
fix: Hero range visualizer not detecting

ryanzhoudev:


kristofbolyai (approved):

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -122,7 +122,7 @@ private void checkMajorIdCircles(Player player) {
         List<Pair<CustomColor, Float>> circles = validGear.stream()
                 .flatMap(gearInfo -> gearInfo.fixedStats().majorIds().stream().map(majorId -> switch (majorId.name()) {
                     case "Taunt" -> Pair.of(CommonColors.ORANGE.withAlpha(TRANSPARENCY), 12f);
-                    case "Saviour's Sacrifice" -> Pair.of(CommonColors.WHITE.withAlpha(TRANSPARENCY), 8f);
+                    case "Saviour’s Sacrifice" -> Pair.of(CommonColors.WHITE.withAlpha(TRANSPARENCY), 8f);
```
magicus, line 125: I told you it was no good having name strings instead of enums! :-) 

This sounds like a problem that can be reintroduced in the future.Maybe we can have some central mapping class where we can store the actual strings as something (like enums) that can be checked by the compiler?

kristofbolyai: I didn't make it an enum originally as there were no other use cases for major ids in the code. We can add it to the list of refactorings to do :)

PR #2303, opened 2023-12-29
chore: Add framework for unit testing of regexes

magicus:
I've been thinking about this for a long time, to have a simple way to store our regex test strings in the code base, and run them as unit tests. 

Here is a framework for doing that, and one test converted from regexr.com. Once this gets integrated, it's a bit of a copy/paste chore to move the remaining tests from regexr.com into the code base. Maybe we have some willing interns that could take that upon themselves? :-)


magicus:
... We should also add running of unit tests to our GHA testing; I don't think we're doing that currently...

kristofbolyai (approved):
This looks good, but I am not sure when we are going to be able to move all the test regexes.

PR #2304, opened 2023-12-29
feat: Item sharing GUI

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/195cb9c7-b535-4f1e-b324-df740c10d054

New sharing options menu, works the same as video just only shows 1 checkbox at a time and wider background for no scrolling button text
image


MenacingPerson:
Is it possible to make the buttons a bit bigger so they don't have to scroll?

common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -249,26 +273,14 @@ private List<StyledTextPart> createItemPart(WynnItem wynnItem) {
...
-            McUtils.sendErrorToClient(I18n.get("feature.wynntils.chatItem.chatItemUnidentifiedError"));
-            return;
-        }
-
-        Optional<WynnItem> wynnItemOpt = Models.Item.getWynnItem(hoveredSlot.getItem());
-        if (wynnItemOpt.isEmpty()) return;
-
-        // TODO: These settings should be persisted, and changeable in the custom share GUI
-        EncodingSettings encodingSettings = new EncodingSettings(false, true);
+    private void makeChatPrompt(WynnItem wynnItem, Optional<GearItem> gearItemOpt) {
```
kristofbolyai: I think it would be cleaner if you just got thte gearItemOpt again before it's used, instead of passing it down

common/src/main/java/com/wynntils/models/items/ItemEncodingModel.java
```diff
@@ -38,4 +46,8 @@ public ErrorOr<WynnItem> decodeItem(EncodedByteBuffer encodedByteBuffer) {
     public Pattern getEncodedDataPattern() {
         return ENCODED_DATA_PATTERN;
     }
+
+    public void toggleSetting(Storage<Boolean> setting) {
+        setting.store(!setting.get());
+    }
 }
```
kristofbolyai: I think this looks very weird, and doing it inline would be better.

common/src/main/java/com/wynntils/screens/itemsharing/ItemSharingScreen.java
```diff
@@ -0,0 +1,225 @@
...
+        tooltipX = this.width - sideGap - tooltipWidth - 30;
+        backgroundX = sideGap + 30;
+    }
+
+    private void shareItem(String target) {
+        if (WynntilsMod.isDevelopmentEnvironment()) {
+            WynntilsMod.info("Encoded item: " + encodedItem);
+            WynntilsMod.info("Encoded item UTF-16: " + encodedItem.toUtf16String());
+        }
+
```
kristofbolyai: I think you can remove this, most likely when debugging, we are going to use the chat setting.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1810,6 +1813,16 @@
   "screens.wynntils.itemSearchHelperWidget.availableStats": "Available Item Stats:",
   "screens.wynntils.itemSearchHelperWidget.page": "Page %d/%d",
   "screens.wynntils.itemSearchHelperWidget.switchPage": "Switch Pages with Left/Right Click",
+  "screens.wynntils.itemSharing.copied": "Copied item to clipboard",
+  "screens.wynntils.itemSharing.copy": "Copy to clipboard",
+  "screens.wynntils.itemSharing.extended.description": "Encode static stats, eg. Strength",
```
```suggestion
"screens.wynntils.itemSharing.extended.description1": "Encode base values and static stats, eg. Strength",
"screens.wynntils.itemSharing.extended.description2": "Note that items encoded in this mode will not follow item changes, stats are freezed in their current state.",
```

kristofbolyai (commented):
I think it looks good! The whole sharing background could be rendered to a bit bigger, so it would fit all the buttons correctly.

Also, since we have 2 settings, one only applies to crafted items, one to gear items only, you could hide to other one, depending on the current item.

ShadowCat117:
Updated original comment with resized sharing menu

magicus:
Oooohh! What a great idea! O_o

kristofbolyai (approved):
Looks good! Thank you for helping me out with this :)

common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
```diff
@@ -145,8 +151,26 @@ public void onChatReceived(ChatMessageReceivedEvent e) {
...
-    private void onInventoryPress(Slot slot) {
-        makeChatPrompt(slot);
+    private void onInventoryPress(Slot hoveredSlot) {
+        if (hoveredSlot == null) return;
+
+        // Special case for unidentified gear
+        Optional<GearItem> gearItemOpt = Models.Item.asWynnItem(hoveredSlot.getItem(), GearItem.class);
+        if (gearItemOpt.isPresent() && gearItemOpt.get().isUnidentified()) {
+            // We can only send chat encoded gear of identified gear
+            WynntilsMod.warn("Cannot make chat link of unidentified gear");
```
magicus, line 161: Why's that, though..? I mean, wouldn't it be nice if we could send unidentified gear as well? Sure, you could type the name and the recipient can look it up in the item DB, but sending it as a chat link would be much easier. Maybe something to consider for the new chat protocol, @kristofbolyai ?

PR #2305, opened 2023-12-30
chore: Move all regex testing from regexr.com to unit tests

magicus:
If something needs doing, do it yourself.

I have now ported all regex tests from regexr.com to our unit tests. (At least, those that had a regexr.com link in the source).

A handful of tests did not actually work, or did not work if I updated the regexr test with the current value of the pattern. I have just commented these out -- we should really investigate if the wynn format has changed, of if we have actually broken the regex pattern.

I made a few changes to the regexr tests if they failed, if they contained superfluous `§r` codes. I believe these changes are safe to do.

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/GuildRankReplacementFeature.java
@@ -33,11 +33,11 @@ public class GuildRankReplacementFeature extends Feature {
     @Persisted
     public final Config<RankType> rankType = new Config<>(RankType.NAME);
 
-    // Test suite: https://regexr.com/7f8kh
+    // Test in GuildRankReplacementFeature_GUILD_MESSAGE_PATTERN
     private static final Pattern GUILD_MESSAGE_PATTERN =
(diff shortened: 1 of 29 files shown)


magicus:
I chose to include the name of the test at the pattern for a few different reasons, even if it was systematically generated:

1) It reminds folks to update the test
2) It makes it easy to go to the test function by a copy and global search
3) It makes it clear which patterns have tests and which have not

kristofbolyai (approved):
DId you fix up the local/global chat patterns in the process? I don't think we've ever updated them to the new Unicode Wynn format.

magicus:
No, I did not change them. Someone already had.

magicus:
The commented-out tests should be reenabled, but we need to determine if the regexps have changed wrongly, or if the old test strings are no longer valid. I'll leave at least that part to someone else...

kristofbolyai:
> The commented-out tests should be reenabled, but we need to determine if the regexps have changed wrongly, or if the old test strings are no longer valid. I'll leave at least that part to someone else...

Ill make a follow up, don't worry about it.

magicus:
And then of course we should add tests for more regexes. :) Perhaps not all, some are really trivial, but there are many non-trivial ones that really should have testing, like all in MessageType, both background and foreground.

And hopefully this makes it easier to always add tests in the future when modifying or adding a regex.

PR #2307, opened 2023-12-30
chore: Don't run tests on "build"

kristofbolyai:
Turns out we already ran tests, when we built the mod. I've made sure we only run them separately now. (2 second runtime for tests was a little suspicious)


kristofbolyai:
@magicus

magicus (approved):

PR #2308, opened 2023-12-30
[ci skip] fix: Fix TestRegex failing tests

kristofbolyai:
I am wondering if we should pass strings through StyledText, and test those against the patterns as well. We still have a lot of reset characters in these patterns, and 99% percent of them could be removed... However, I am not sure if it is worth the effort; if it works, don't fix it.

### DIFF
--- a/fabric/src/test/java/TestRegex.java
+++ b/fabric/src/test/java/TestRegex.java
@@ -368,11 +368,9 @@ public class TestRegex {
     @Test
     public void GuildRankReplacementFeature_GUILD_MESSAGE_PATTERN() {
         PatternTester p = new PatternTester(GuildRankReplacementFeature.class, "GUILD_MESSAGE_PATTERN");
-        /* FIXME: These tests fail
-        p.shouldMatch("§3[§b★★★★★§3§oDisco reroller§3]");
-        p.shouldMatch("§3[§b★★★★★§3§oafKing§r§3]§");
-        p.shouldMatch("§3[§b★★★★§3§obol§r§3]");
-        */
+        p.shouldMatch("§3[§b★★★★★§3§oDisco reroller§3]§b");
+        p.shouldMatch("§3[§b★★★★★§3§oafKing§3]§b");
+        p.shouldMatch("§3[§b★★★★§3§obol§3]§b");
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


magicus (approved):
Did you just remove §r or did you check these with ingame strings?

kristofbolyai:
> Did you just remove §r or did you check these with ingame strings?

I've checked if the regexes were updated during the StyledText migrations, and if the reset chars have a color after them (which was a change introduced by StyledText)

magicus:
We should probably re-parse our strings, that is, take the constant strings in the tests, and pass them through `StyledText.fromString("test string).getString()` (or whatever it was called), and replace the literals with whatever comes out the other side. It's probably a quick hack to do that.

We only match on StyledText or stuff that has passed StyledText, but many of the strings here were copied from the game chat before the advent of StyledText.

magicus:
I just did this:
```
diff --git a/fabric/src/test/java/TestRegex.java b/fabric/src/test/java/TestRegex.java
index 9f7243829..721862001 100644
--- a/fabric/src/test/java/TestRegex.java
+++ b/fabric/src/test/java/TestRegex.java
@@ -2,6 +2,7 @@
  * Copyright © Wynntils 2023.
  */
 import com.wynntils.features.chat.GuildRankReplacementFeature;
 import com.wynntils.features.chat.MessageFilterFeature;
 import com.wynntils.features.redirects.ChatRedirectFeature;
@@ -78,12 +79,20 @@ public class TestRegex {
         }
 
         public void shouldMatch(String s) {
+            String ns = StyledText.fromString(s).getString();
+            if (!ns.equals(s)) {
+                System.out.println("MISMATCH: " + s);
+            }
             Assertions.assertTrue(
                     pattern.matcher(s).matches(),
                     "Regex failure: " + regexName + " should have matched " + s + ", but it did not.");
         }
 
         public void shouldNotMatch(String s) {
+            String ns = StyledText.fromString(s).getString();
+            if (!ns.equals(s)) {
+                System.out.println("MISMATCH: " + s);
+            }
             Assertions.assertFalse(
                     pattern.matcher(s).matches(),
                     "Regex failure: " + regexName + " should NOT have matched " + s + ", but it did.");
```

and I found a bunch of strings that should be updated...

magicus:
I can't be bothered to fix it right now, though...

kristofbolyai:
I assumed, since I've found more than a hundred reset characters overall..

PR #2310, opened 2024-01-03
feat: Skill point loadouts

kristofbolyai:
This is @DonkeyBlaster's PR, but squashed. Hopefully spotless won't be triggered for all the merge commits adding 2024 headers..


common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -0,0 +1,385 @@
...
+                .build();
+        query.executeQuery();
+    }
+
+    /**
+     * Closes any open containers (but not the screen shown) then queries compass (and tome menus) depending on if
+     * tomes have been unlocked.
+     */
+    public void populateSkillPoints() {
+        ContainerUtils.closeBackgroundContainer();
```
kristofbolyai, line 144: This should be okay for now.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
```diff
@@ -0,0 +1,74 @@
...
+
+public record SavableSkillPointSet(
+        int strength,
+        int dexterity,
+        int intelligence,
+        int defence,
+        int agility,
+        List<String> armourNames,
+        List<String> accessoryNames,
+        boolean isBuild) {
```
kristofbolyai: I think magicus requested that you make `isBuild` a method, basically revert it to as it was. Sorry for the confusion.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -0,0 +1,538 @@
...
+
+        convertButton =
+                new WynntilsButton(
+                        (int) (dividedWidth * 52),
+                        (int) (dividedHeight * 52),
+                        (int) ((dividedWidth * 59) - (dividedWidth * 52)),
+                        BUTTON_HEIGHT,
+                        Component.translatable("screens.wynntils.skillPointLoadouts.convert")) {
+                    @Override
+                    public void onPress() {
```
kristofbolyai: I think I previously commented on how these buttons should be separate classes, not anonymous ones. I think you only converted the first one, and forgot the others.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
```diff
@@ -0,0 +1,76 @@
...
+        return (int) Math.ceil(getSkillPointsSum() / 2.0) + 1;
+    }
+
+    public List<String> getArmourNames() {
+        return Collections.unmodifiableList(armourNames);
+    }
+
+    public List<String> getAccessoryNames() {
+        return Collections.unmodifiableList(accessoryNames);
+    }
```
kristofbolyai: These methods are not needed.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
```diff
@@ -0,0 +1,76 @@
...
+        int intelligence,
+        int defence,
+        int agility,
+        List<String> armourNames,
+        List<String> accessoryNames) {
+    /**
+     * Constructs a new SavableSkillPointSet representing just a loadout.
+     */
+    public SavableSkillPointSet(int[] skillPoints) {
+        this(skillPoints, new ArrayList<>(), new ArrayList<>());
```
kristofbolyai, line 23: Should be List.of() so it is an unmodifiable list

common/src/main/java/com/wynntils/features/utilities/SkillPointLoadoutsFeature.java
```diff
@@ -0,0 +1,48 @@
...
+    @SubscribeEvent
+    public void onCharacterInfoScreenOpened(ScreenOpenedEvent.Post e) {
+        if (!(e.getScreen() instanceof ContainerScreen screen)) return;
+        if (!Models.Container.isCharacterInfoScreen(screen)) return;
+
+        screen.addRenderableWidget(
+                new LoadoutScreenButton(screen.width / 2 - LoadoutScreenButton.BUTTON_WIDTH / 2, screen.topPos - 24));
+    }
+
+    private static final class LoadoutScreenButton extends WynntilsButton {
```
ShadowCat117, line 30: Missed this button too.

Also, do these buttons need to be separate classes, I think you can just use the button builder as from what I can tell none of the buttons do anything different from a regular button

kristofbolyai: This button specifically can be in the feature class, as it's injected by the feature, and not used on a custom screen. 

Do you want to use button builders in regular screens? Sure, we did that before, if there is nothing special about them.

ShadowCat117: I think using the builder is fine, for example the load button can just be this and (I think) it would work exactly the same, no need for a seperate LoadButton class

```
loadButton = new Button.Builder(
                        Component.translatable("screens.wynntils.skillPointLoadouts.load"),
                        (b) -> Models.SkillPoint.loadLoadout(selectedLoadout.key()))
                .pos((int) (dividedWidth * 35), (int) (dividedHeight * 52))
                .size((int) ((dividedWidth * 44) - (dividedWidth * 35)), BUTTON_HEIGHT)
                .build();
```

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
ryanzhoudev: Doesn't work with gson for some reason, when I try to save anything with List.of() it does not persist after relaunch

kristofbolyai:
`[20:25:00] [Render thread/WARN] (wynntils) Skill Point Model failed to parse tome: StyledText{' §6Requirements: - §a✔ §7Lv. Min: 105§6- §c✖ §7Raid Req: §f10§7/15'}`

Some overlooked error, should be easily to handle.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
kristofbolyai: Yeah, that might be the case. I'll just make sure to support it separately in a new PR eventually. Just use the immutable wrappers here for now, and i'll handle proper serialization support later.

common/src/main/java/com/wynntils/features/utilities/SkillPointLoadoutsFeature.java
kristofbolyai: Hm, it very well could be changed, it is a lot less lines, so maybe it's cleaner. Although having a separate class has it's cleanness too. Let's leave it up to DonkeyBlaster to decide.

ryanzhoudev: I will change it to a builder, having a separate class in a non-screen feels like a waste

oh wait no the actual convert/delete/loads I will keep as separate since they are screen parts, if the LoadoutScreenButton class is fine in the feature I will leave it

kristofbolyai (commented):
Looks good now! I'll merge this PR with the chat item one, when the time comes.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public int getSkillPointsSum() {
+        return strength + dexterity + intelligence + defence + agility;
+    }
+
+    public int getMinimumCombatLevel() {
+        return (int) Math.ceil(getSkillPointsSum() / 2.0) + 1;
+    }
+
+    public boolean isBuild() {
```
magicus, line 52: The traditional concept of a "build" in Wynncraft has also included a weapon. I'm not demanding a change in this PR, but maybe we should think of how this could be implemented in the future. That would make it more compatible with e.g. wynndata.

magicus (approved):
I've looked through this, some parts more cursory and other in depth. I think it looks good.

common/src/main/java/com/wynntils/models/character/type/SavableSkillPointSet.java
ryanzhoudev: Yup, the eventual goal is to integrate with bolyai's gear encoding standard and have exporting to/from wynnbuilder
(but the legacy portion is working already, so why not)

PR #2313, opened 2024-01-07
chore: Bump to release, add custom versioning

kristofbolyai:
Bumped `pre-alpha` => `alpha`.
Bumped `beta` => `release`.

New versioning: `Artemis v[bumps with mc version].[bumps with major features or fixes].[bumps with everything else]`.

To bump a major version: `chore(bump-mc)` or `chore!`.
To bump a minor version: `feat!` or `fix!`.
To bump a patch: any other conventional commit.

See https://github.com/kristofbolyai/Artemis/releases.
Screenshot if those releases get deleted:
image


kristofbolyai:
Don't merge this. This needs special care for the first release, before working automatically.

.github/.config.js
```diff
@@ -1,22 +1,74 @@
...
+    );
+
+    // Both of these are used in different places...
+    options.recommendedBumpOpts.whatBump = whatBump;
+    options.whatBump = whatBump;
+
+    return options;
+}
+
+module.exports = getOptions();
```
magicus: Missing newline at EOF

.github/workflows/autopromote.yml
```diff
@@ -7,9 +7,9 @@ on:
       - development
 
 jobs:
-  update-pre-alpha:
+  update-alpha:
```
magicus, line 10: Since we only have a single release stream now, I'm guessing the reason for the alpha is to provide a separate stream for testing out new Wynncraft versions on Wynncraft hero beta which introduce breaking changes. Am I correct?

If so, maybe we should name it "testing" instead of "alpha"? It feels a bit odd to "revert" to an alpha once we have done a series of proper releases.

magicus (approved):
Overall it looks good.

.github/workflows/autopromote.yml
kristofbolyai: I would like it to keep the name as it fits into the conventional commit standard, even if it does not matter fully at the moment.

PR #2317, opened 2024-01-09
fix: Held weapon not counted in gear skill points

ryanzhoudev:


magicus:
There is something odd going on with your setup that causes spotless to update year on all files. :-(

ryanzhoudev:
#2319

PR #2318, opened 2024-01-10
feat: Add "pf" as a shorter alias for "partyfinder"

afkfish:


ShadowCat117:
This isn't where you add custom aliases, this is for adding autofill to Wynn commands. `CommandAliasesFeature` is what you want, though by the looks of it it you can't just add a command alias to that easily as they're stored in a default config so the alias won't be added unless users delete that from their current config file

kristofbolyai:
If the user never edited their aliases, then it'll be updated, @JamieCallan117

ShadowCat117:
> If the user never edited their aliases, then it'll be updated, @JamieCallan117

It's a hidden config so only editable in the file. Weird that my config file has them then as I've never edited them before

afkfish:
> This isn't where you add custom aliases, this is for adding autofill to Wynn commands. `CommandAliasesFeature` is what you want, though by the looks of it it you can't just add a command alias to that easily as they're stored in a default config so the alias won't be added unless users delete that from their current config file

Ah thank you for the clarifying.

common/src/main/java/com/wynntils/features/commands/AddCommandExpansionFeature.java
```diff
@@ -158,6 +158,7 @@ private void addArgumentlessCommandNodes(RootCommandNode<SharedSuggestionProvide
         addNode(root, partyfinderNode);
 
         addAlias(root, partyfinderNode, "pfinder", AliasCommandLevel.SHORT_FORMS);
+        addAlias(root, partyfinderNode, "pf", AliasCommandLevel.SHORT_FORMS);
```
ShadowCat117: You don't this anymore

common/src/main/java/com/wynntils/features/commands/CommandAliasesFeature.java
```diff
@@ -30,7 +30,8 @@ public class CommandAliasesFeature extends Feature {
     public final HiddenConfig<List<CommandAlias>> aliases = new HiddenConfig<>(new ArrayList<>(List.of(
             new CommandAlias("guild attack", List.of("gu a", "guild a")),
             new CommandAlias("guild manage", List.of("gu m", "gu man", "guild m", "guild man")),
-            new CommandAlias("guild territory", List.of("gu t", "gu terr", "guild t", "guild terr")))));
+            new CommandAlias("guild territory", List.of("gu t", "gu terr", "guild t", "guild terr")),
+            new CommandAlias("partyfinder", List.of("pf", "pfinder")))));
```
ShadowCat117: No need to list pfinder, that's already a command in Wynn

common/src/main/java/com/wynntils/features/commands/AddCommandExpansionFeature.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
```
kristofbolyai:
```suggestion
 * Copyright © Wynntils 2022-2023.
```

kristofbolyai (approved):

PR #2321, opened 2024-01-10
feat: Max and min functions take argument list, docs improvements [skip ci]

ryanzhoudev:
Also updated https://wynntils-functions.donkeyblaster.tk/ with this


magicus (approved):

ShadowCat117:
@magicus I presume you've not seen Discord, but there's a reason for the [skip ci]

> it would be nice if we wouldn’t create hundreds of builds, so let’s batch some minor prs if they are not important fixes that need to be out asap, if it makes sense

> you can skip pr auto release in 2 ways:
> if you want a PR to be not included in the changelog: add [ci skip] before the pr name
> if you want the pr to be included jn the changelog (but still delayed): add [skip ci] to the end of the pr

PR #2323, opened 2024-01-10
fix: Closing item sharing screen breaks screen state

ryanzhoudev:
Closing sharing screen leaves background containers open.
https://discord.com/channels/394189072635133952/1023381446733352990/1194780496920584264

Since we didn't tell the server we moved away, we can just open it again and to the server it appears nothing happens.

This does not affect item record since that is our own custom screen and storage.


ShadowCat117 (approved):
This definitely makes the sharing screen better, kind of glad it caused a bug so it got added (:

kristofbolyai (approved):

PR #2325, opened 2024-01-11
fix: Custom mod overlays do not render on chat

ryanzhoudev:
POST does not render behind chat


common/src/main/java/com/wynntils/core/consumers/overlays/annotations/OverlayInfo.java
```diff
@@ -19,5 +19,8 @@
 
     RenderEvent.ElementType renderType();
 
-    RenderState renderAt() default RenderState.POST;
+    RenderState renderAt() default RenderState.PRE;
```
kristofbolyai: I surely don't want this to be changed.. We've set every overlay with the assumption this is Post, and specified PRE for specific cases. I've seen the chat problem, and most likely a better fix would be is to just patch the chat background to not be "solid".

kristofbolyai: Even if it seems working, we would have to test tens of overlays manually, and make sure it does not interact weirdly with anything, which is problematic, and WILL lead to bugs.

kristofbolyai: Not to mention, this does not fix any overlays' rendering issues that need to be rendered on Post for some reason.

common/src/main/java/com/wynntils/core/consumers/overlays/annotations/OverlayInfo.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
```
magicus: Revert this change, since the file is not edited.

kristofbolyai: Because of the revert commit, this will be added back.

magicus: Is that so? I thought spotless just diffed the file, but you are saying that it looks throught the commit history?

magicus:
```suggestion
 * Copyright © Wynntils 2022-2023.
```

magicus: Let's see what happens! :)

magicus (approved):

common/src/main/java/com/wynntils/core/consumers/overlays/annotations/OverlayInfo.java
magicus: Seems like the undoing stuck.

PR #2327, opened 2024-01-12
fix: Vanilla actionbar formatting [skip ci]

ryanzhoudev:
So in the first one, the padding is missing and the mana turns a weird color when curse is charged to full. Both are fixed here. Padding of 3 spaces taken from vanilla Wynn.

Delayed because very few people use vanilla actionbar and this doesn't actually hide any information, just looks better

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
@@ -19,6 +19,7 @@ import java.util.regex.Matcher;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class ActionBarHandler extends Handler {
+    private static final String VANILLA_PADDING = "\\s{4,}";
     private static final StyledText CENTER_PADDING = StyledText.fromString("§0               ");
     private static final String STANDARD_PADDING = "    ";
 
@@ -52,7 +53,7 @@ public final class ActionBarHandler extends Handler {
         }
         previousRawContent = content;
 
-        StyledText[] contentGroups = content.split(" {4,}");
+        StyledText[] contentGroups = content.split(VANILLA_PADDING);
 
         // Create map of position -> matching part of the content
         Map<ActionBarPosition, StyledText> positionMatches = new EnumMap<>(ActionBarPosition.class);
--- a/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
@@ -73,7 +73,6 @@ public final class ActionBarHandler extends Handler {
 
         StyledText newContentBuilder = StyledText.EMPTY;
         // vanilla segments have three spaces between each segment, regardless of content
-        //
         if (!lastSegments.get(ActionBarPosition.LEFT).isHidden()) {
             newContentBuilder = newContentBuilder.append(positionMatches.get(ActionBarPosition.LEFT));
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -70,18 +72,30 @@ public void onActionBarUpdate(ChatPacketReceivedEvent.GameInfo event) {
...
+            newContentBuilder = newContentBuilder.append(STANDARD_PADDING);
         } else {
             // Add padding
             newContentBuilder = newContentBuilder.append(CENTER_PADDING);
         }
         if (!lastSegments.get(ActionBarPosition.RIGHT).isHidden()) {
-            newContentBuilder = newContentBuilder.append(positionMatches.get(ActionBarPosition.RIGHT));
+            StyledText right = positionMatches.get(ActionBarPosition.RIGHT);
+
+            // Rare case where the user has 100% curse charge, which makes StyledText remove the color code
```
magicus: This seems like the wrong solution for two reasons.
1) It forces right to be aqua, assuming that it is always mana. I'm not sure it always is..? Can't it show different values e.g. when swimming? And even if it is, hard-coding it like this feels wrong.

2) We should have StyledText properly split up the string to include all formatting codes needed for each part. Iirc, we now have such a split function, but we did not at the time when the ActionBarHandler was written. So we should use that instead of this workaround.

And if we do not have such a function, we need to write one. It is the responsibility of StyledText to be able to include all relevant formatting codes.

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -56,32 +52,32 @@ public void onActionBarUpdate(ChatPacketReceivedEvent.GameInfo event) {
         }
         previousRawContent = content;
 
-        Matcher matcher = content.getMatcher(ACTIONBAR_PATTERN);
-        if (!matcher.matches()) {
-            WynntilsMod.warn("ActionBarHandler pattern failed to match: " + content);
-            return;
-        }
+        StyledText[] contentGroups = content.split(" {4,}");
```
kristofbolyai: Perhaps use `\\s{4,}` and also make this a constant field

common/src/main/java/com/wynntils/handlers/actionbar/ActionBarHandler.java
```diff
@@ -56,32 +52,32 @@ public void onActionBarUpdate(ChatPacketReceivedEvent.GameInfo event) {
...
 
         // Create map of position -> matching part of the content
         Map<ActionBarPosition, StyledText> positionMatches = new EnumMap<>(ActionBarPosition.class);
         Arrays.stream(ActionBarPosition.values())
-                .forEach(pos -> positionMatches.put(pos, StyledText.fromString(matcher.group(pos.name()))));
+                .forEach(pos -> positionMatches.put(pos, contentGroups[pos.ordinal()]));
```
kristofbolyai: This assumes the split always results in 3 parts. Is this always true?

ryanzhoudev: I cannot think of any actionbar where it is not true. Normally, we always have health + mana, then coordinates, "Housing Island", spring, or powder there. I will add a failsafe though

kristofbolyai (commented):
Looks okay to me, but I want magicus to have the final say.

magicus (approved):

PR #2329, opened 2024-01-15
fix: Custom connect screen does not revert on failure

### LINKED ISSUE #2128: Custom loading screen does not revert to vanilla on certain errors
#2002 for details.

ryanzhoudev:
Previously this did not work because when we cancelled/overwrote ConnectScreen, we stopped calling it's tick() method which is what tells Connection.class to do the handleDisconnection() work. So we will just do it manually now.

Also uses i18n for the messages now. Closes #2128


common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
```diff
@@ -12,29 +12,45 @@
...
             event.setCanceled(true);
+            loadingScreen.setMessage(I18n.get("feature.wynntils.customLoadingScreen.working"));
         }
         if (event.getScreen() instanceof ReceivingLevelScreen) {
             event.setCanceled(true);
-            loadingScreen.setMessage("Receiving terrain...");
+            loadingScreen.setMessage(I18n.get("feature.wynntils.customLoadingScreen.receivingTerrain"));
+        }
+
+        if (event.getScreen() instanceof DisconnectedScreen) {
```
magicus: This does not seem like the correct condition to set baseConnectScreen to null...? Or at least it should also be set to null when the ConnectScreen closes.

magicus:
Yeah, the ConnectScreen does too much non-GUI stuff... Mojang is behind on some refactoring there... Good find!

common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
ryanzhoudev: There is no actual way to check if the ConnectScreen closes, because, well, we got rid of it and have our own. But I think what I have here is fine now. Any screen that isn't our custom screen or some un-cancelled ConnectScreen will erase it. This allows it to work during our connection, and closes it as soon as the handshake is done (including ProgressScreen, ReceivingLevelScreen, etc), it'll disappear nicely.

magicus: Ok. Sounds good then. Maybe add a shortened version of this explanation as a comment?

magicus (approved):

kristofbolyai (approved):

kristofbolyai:
Thanks for spending time on this.

PR #2330, opened 2024-01-16
fix: Fix adding new custom bars duplicating config options [skip ci]

ShadowCat117:
Prevents this from happening

https://github.com/Wynntils/Artemis/assets/8337467/e29ebd3a-c67c-4c3d-8617-81b5103d6c6c

### DIFF
--- a/common/src/main/java/com/wynntils/commands/ConfigCommand.java
+++ b/common/src/main/java/com/wynntils/commands/ConfigCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.commands;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/commands/ConfigCommand.java
```diff
@@ -291,6 +291,7 @@ private int addOverlayGroup(CommandContext<CommandSourceStack> context) {
 
         Managers.Config.loadConfigOptions(true, false);
         Managers.Config.saveConfig();
+        Managers.Config.reloadConfiguration();
```
kristofbolyai, line 294: Doesn't loadConfiguration just above work? what is the difference? (I am on phone and need to check..)

ShadowCat117: reload loads the config file again and then calls that method again with both parameters as true, it works the same if I call that again after saving so shall I change it to that

kristofbolyai (approved):
Seems to work in-game.

PR #2331, opened 2024-01-17
fix: Use proper hooks for enabling/disabling Discord RPC, refactor territory checking [ci skip]

kristofbolyai:
Previously we called `DiscordService#load` for every config change, even if it was necessary. This is partially because `DiscordService#isReady` will always be false if Discord is closed. These new hooks will run iff we changed the enabled/disabled state of the feature (+on startup, if the feature is enabled). This makes config related operations much faster (as I noticed that `/config` commands hang the game for seconds...).

I would like to request help with testing some platforms, @JamieCallan117, @DonkeyBlaster. Please tick the table below if you found the feature working on the platform (please do an enable/disable cycle, start with a disabled state, run `/config reload`).

| OS | Test |
|--------|--------|
| MacOS (arch) | &check; |
| Linux - Ubuntu |  ? |
| Linux - Arch | &check; |
| Windows | &check;  |

### DIFF
--- a/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
+++ b/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
@@ -126,6 +126,11 @@ public class DiscordRichPresenceFeature extends Feature {
     }
 
     @Override
+    protected void onConfigUpdate(Config<?> config) {
+        tryUpdateDisplayedInfo();
+    }
+
+    @Override
     public void onEnable() {
         // This isReady() check is required for Linux to not crash on config change.
         if (!Services.Discord.isReady()) {
@@ -133,9 +138,18 @@ public class DiscordRichPresenceFeature extends Feature {
             if (!Services.Discord.load()) {
                 // happens when wrong version of GLIBC is installed and Discord SDK fails to load
                 Managers.Feature.crashFeature(this);
+            } else {
+                tryUpdateDisplayedInfo();
             }
         }
+    }
 
+    @Override
+    public void onDisable() {
+        Services.Discord.unload();
+    }
+
+    private void tryUpdateDisplayedInfo() {
         if (!Models.WorldState.onWorld() && Services.Discord.isReady()) return;
 
         if (displayLocation.get()) {
@@ -159,10 +173,5 @@ public class DiscordRichPresenceFeature extends Feature {
         } else {
             Services.Discord.setState("");
         }
-    }
-
-    @Override
-    public void onDisable() {
-        Services.Discord.unload();
     }
 }
--- a/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
+++ b/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
@@ -41,6 +41,7 @@ public class DiscordRichPresenceFeature extends Feature {
 
     @SubscribeEvent
     public void onCharacterUpdate(CharacterUpdateEvent event) {
+        if (!Services.Discord.isReady()) return;
         if (!Models.WorldState.onWorld()) return;
 
         if (displayCharacterInfo.get()) {
@@ -50,6 +51,7 @@ public class DiscordRichPresenceFeature extends Feature {
 
     @SubscribeEvent
     public void onXpChange(SetXpEvent event) {
+        if (!Services.Discord.isReady()) return;
         if (!Models.WorldState.onWorld()) return;
 
         if (displayCharacterInfo.get()) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -126,42 +126,43 @@ private void checkTerritory() {
...
+        if (!Services.Discord.isReady()) {
+            // Even though this is in the onConfigUpdate method, it is how the library is first loaded on launch
+            if (!Services.Discord.load()) {
+                // happens when wrong version of GLIBC is installed and Discord SDK fails to load
+                Managers.Feature.crashFeature(this);
             }
+        }
 
-            if (!Models.WorldState.onWorld() && Services.Discord.isReady()) return;
+        if (!Models.WorldState.onWorld() && Services.Discord.isReady()) return;
```
magicus: I'm trying to make sense of this.

1) If enabled at startup, onWorld is always false, so this returns if discord is ready, otherwise it continues below.

2) If enabled later, while in the hub, and discord is not ready, it continues below.

Both of these behaviors seems sus.

ShadowCat117:
Seems to work fine on Windows 👍

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
kristofbolyai: Most likely I have to separate the enabling/disabling logic from the config update method, and add it back (and also keep enable/disable separate). Overall, the aim is to reduce the native library calls, as they are slow.

kristofbolyai:
@JamieCallan117 Could you retest this new version?

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
kristofbolyai: Should be fixed now.

magicus: I still think it looks weird. Are you sure you don't mean:
```
 if (!(Models.WorldState.onWorld() && Services.Discord.isReady())) return;
 ```
?

Comment (kristofbolyai) [reply common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java]: Probably should be an or. Not on world, or discord is not ready.

Comment (magicus) [reply common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java]: If so, that is prhaps better expressed as:
```
 if (!Services.Discord.isReady()) return;
```

File: common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java [hunk 2]
```diff
@@ -88,6 +92,57 @@ public void onDisconnect(ConnectionEvent.DisconnectedEvent e) {
...
+    @Override
+    protected void onConfigUpdate(Config<?> config) {
+        tryUpdateDisplayedInfo();
+    }
+
+    @Override
+    public void onEnable() {
+        // This isReady() check is required for Linux to not crash on config change.
+        if (!Services.Discord.isReady()) {
+            // Even though this is in the onConfigUpdate method, it is how the library is first loaded on launch
```
magicus: This comment is no longer correct

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -88,6 +92,57 @@ public void onDisconnect(ConnectionEvent.DisconnectedEvent e) {
...
+            if (!Services.Discord.load()) {
+                // happens when wrong version of GLIBC is installed and Discord SDK fails to load
+                Managers.Feature.crashFeature(this);
+            } else {
+                tryUpdateDisplayedInfo();
```
magicus: I think you should do this even if discord isReady. I think the nicest way is to move this out to the end of the method, and add a return statement after the crash instead. I think that captures the logic nicely.

ShadowCat117:
> @JamieCallan117 Could you retest this new version?

Still seems to work fine

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -59,33 +61,89 @@ public void onXpChange(SetXpEvent event) {
...
+
+    @Override
+    public void onEnable() {
+        // This isReady() check is required for Linux to not crash on config change.
+        if (!Services.Discord.isReady()) {
+            // Load the Discord SDK
+            if (!Services.Discord.load()) {
+                // happens when wrong version of GLIBC is installed and Discord SDK fails to load
+                Managers.Feature.crashFeature(this);
+            }
```
magicus, line 113: Despite the name, this does not stop the code from running, it just adds the feature to the list of crashed features. So you'll need a return here so as not fall through to tryUpdateDisplayedInfo(). For now it seems like it should be okay since there are guards in that method, but that seems to be more by incident than design, so any change in that method might mess things up.

magicus (approved):
Almost there... :)

PR #2332, opened 2024-01-20
fix: fix hunted mode breaking skill point parsing [skip ci]

ryanzhoudev:
Also moved a few public methods up in the character model

### DIFF
--- a/common/src/main/java/com/wynntils/models/character/CharacterModel.java
+++ b/common/src/main/java/com/wynntils/models/character/CharacterModel.java
@@ -51,7 +51,7 @@ public final class CharacterModel extends Model {
             "§7Expiration: §f(?:(?<weeks>\\d+) weeks?)? ?(?:(?<days>\\d+) days?)? ?(?:(?<hours>\\d+) hours?)?");
     // Test in CharacterModel_VETERAN_PATTERN
     private static final Pattern VETERAN_PATTERN = Pattern.compile("§7Rank: §[6dba]Vet");
-    private static final String HUNTED_STRING = "to enable hunted";
+    private static final Pattern HUNTED_PATTERN = Pattern.compile("§cShift Right-Click to enable hunted");
 
     private static final int RANK_SUBSCRIPTION_INFO_SLOT = 0;
     public static final int CHARACTER_INFO_SLOT = 7;
@@ -127,8 +127,8 @@ public final class CharacterModel extends Model {
     }
 
     public boolean isHunted() {
-        return !LoreUtils.getStringLore(McUtils.inventory().getItem(SOUL_POINT_SLOT))
-                .contains(HUNTED_STRING);
+        Matcher m = LoreUtils.matchLoreLine(McUtils.inventory().getItem(SOUL_POINT_SLOT), 8, HUNTED_PATTERN);
+        return !m.matches();
     }
 
     @SubscribeEvent
--- a/common/src/main/java/com/wynntils/models/character/CharacterModel.java
+++ b/common/src/main/java/com/wynntils/models/character/CharacterModel.java
@@ -51,6 +51,7 @@ public final class CharacterModel extends Model {
             "§7Expiration: §f(?:(?<weeks>\\d+) weeks?)? ?(?:(?<days>\\d+) days?)? ?(?:(?<hours>\\d+) hours?)?");
     // Test in CharacterModel_VETERAN_PATTERN
     private static final Pattern VETERAN_PATTERN = Pattern.compile("§7Rank: §[6dba]Vet");
+    private static final String HUNTED_STRING = "to enable hunted";
 
     private static final int RANK_SUBSCRIPTION_INFO_SLOT = 0;
     public static final int CHARACTER_INFO_SLOT = 7;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -125,6 +126,11 @@ public boolean isVeteran() {
         return isVeteran;
     }
 
+    public boolean isHunted() {
+        return !LoreUtils.getStringLore(McUtils.inventory().getItem(SOUL_POINT_SLOT))
+                .contains(HUNTED_STRING);
```
kristofbolyai: Let's not use contains. Is there a reason not to match the whole string, either with a pattern or a hardcoded string?

ShadowCat117:
In the CharacterStats model there's a check for the soul point item being a nether star or diamond axe, since there's a way to check for hunted in this, could you edit that to use it too? :)

ShadowCat117:
Just had a thought, I'll check in a bit if you can't but how does the soul point look if on a hunted character? I imagine it wouldn't say click to enable hunted.

Yeah, it doesn't say that
image

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -352,20 +356,16 @@ private boolean checkTomesUnlocked(ContainerContent content) {
...
     }
 
-    private boolean verifyCharacterInfo(
-            ContainerContent content, Int2ObjectMap<ItemStack> changes, ContainerContentChangeType changeType) {
+    private boolean verifyChange(
+            ContainerContent content,
+            Int2ObjectMap<ItemStack> changes,
+            ContainerContentChangeType changeType,
+            int soulPointItemSlot) {
         // soul points resent last for both containers
```
magicus, line 364: I don't understand this comment. Is it a typo?

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -125,6 +126,11 @@ public boolean isVeteran() {
         return isVeteran;
     }
 
+    public boolean isHunted() {
+        Matcher m = LoreUtils.matchLoreLine(McUtils.inventory().getItem(SOUL_POINT_SLOT), 8, HUNTED_PATTERN);
```
magicus: Can't you just use the diamond axe check here? I'm not sure but I'm guessing this was what @JamieCallan117 meant... It will make the method much simpler.

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -98,7 +98,7 @@ private int getMaxSoulPoints() {
      */
     private int getCurrentSoulPoints() {
         ItemStack soulPoints = McUtils.inventory().getItem(8);
-        if (soulPoints.getItem() == Items.NETHER_STAR || soulPoints.getItem() == Items.DIAMOND_AXE) {
+        if (soulPoints.getItem() == (Models.Character.isHunted() ? Items.DIAMOND_AXE : Items.NETHER_STAR)) {
```
magicus: I think this change is kind of unnecessary. (Especially if you do the change I request to `isHunted`, then it becomes completely tautological.) 

In fact, we already know and trust that slot 8 is the soul points, so this is more of a sanity check/assertion. Had it been written today with the style we have now (and not copied verbatim from Legacy), it would have looked like:
```
 private int getCurrentSoulPoints() {
        ItemStack soulPoints = McUtils.inventory().getItem(8);
        if (soulPoints.getItem() != Items.NETHER_STAR &&  soulPoints.getItem() != Items.DIAMOND_AXE) return -1;

        return soulPoints.getCount();     
}
```

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
ryanzhoudev: No, it means the soul point item (nether star or diamond axe) will be sent as the last item in the stream of item updates, when some container is opened

magicus: Aha, re-sent, not resent. I get it.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -125,6 +125,10 @@ public boolean isVeteran() {
         return isVeteran;
     }
 
+    public boolean isHunted() {
```
kristofbolyai:
```suggestion
    public boolean isHuntedMode() {
```

Or something else, I think this name needs to be a bit better.

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -98,11 +98,9 @@ private int getMaxSoulPoints() {
      */
     private int getCurrentSoulPoints() {
         ItemStack soulPoints = McUtils.inventory().getItem(8);
-        if (soulPoints.getItem() == Items.NETHER_STAR || soulPoints.getItem() == Items.DIAMOND_AXE) {
-            return soulPoints.getCount();
-        }
+        if (soulPoints.getItem() != Items.NETHER_STAR && soulPoints.getItem() != Items.DIAMOND_AXE) return -1;
 
-        return -1;
+        return soulPoints.getCount();
```
kristofbolyai: Does this change do anything?

kristofbolyai (approved):
Looks good, but I think CharacterStatsModel could be reverted.

kristofbolyai (approved):

PR #2335, opened 2024-01-25
feat: Settings Screen Redesign

ShadowCat117:
[Final Textures](https://github.com/Wynntils/Artemis/pull/2335#issuecomment-2131426203)

- Filter by categories.
- Config widgets are now a consistent size. Uses scrolling text if too small.
- Export and import configs for all or just selected configurable.
- Searching will show features from selected category first, followed by any matching features from other categories.
- Selecting a feature with a matching config option to the search query will scroll to make that config option instantly visible.
- Quickly enable/disable features via checkbox.

https://github.com/Wynntils/Artemis/assets/8337467/ffc14dfd-1172-4903-8849-f65b0d9e40e0


ShadowCat117:
### See above for updated version.

Settings screen

- True scrolling, no longer pages.
- Filter by categories.
- Searching will show features from selected category first, followed by any matching features from other categories.
- Config options matching search query will be shown by number following feature name.
- Selecting a feature with a matching config option to the search query will scroll to make that config option instantly visible
- Quickly enable/disable features via checkbox.
- Import and export settings.

https://github.com/Wynntils/Artemis/assets/8337467/b151d372-80e4-4f47-ba86-5d21cd9e72c0

magicus:
@JamieCallan117 This looks like major improvements in usability!

ShadowCat117:
### No longer in this pr, was added in https://github.com/Wynntils/Artemis/pull/2502

Overlay Selection screen

- Search bar to search for an overlay.
- Filter by built-in or custom. (Info boxes/custom bars)
- Edit configurables on this screen.
- Preview overlay, with option to hide other overlays in preview.
- Buttons to add new info boxes and custom bars. (Commands still exist)
- Newly added overlays are automatically set as selected for instant edit access.
- Shift-Right click an info box/custom bar in list to delete it.
- Selected overlay, filter type and scroll position is persisted between screens.

https://github.com/Wynntils/Artemis/assets/8337467/9ed8e678-6979-4682-9e84-c9711b0dcf5c

magicus:
These changes all look great! I am very happy to see that you're taking the time needed to make these improvements.

kristofbolyai:
Up to you, but I have one idea for the overlay management screen. Do you think we could add user-specified/optionally custom naming for the custom overlays? It needs a bit of background work before we could do it though.

kristofbolyai:
Also, all of the new designs look beautiful! Thanks for sorting out all the UIs I "quickly" added so we could work on other core parts of the mods.

Rafii2198:
I don't know if it is the right way, but would simply adding a new config field for all custom overlays called `Name` would achieve this? The UIs could check if it is set and render what user put there and if it is empty then usual `Info Box X` for example would be shown instead and perhaps commands could use either the ID or the name if it is possible, that way 3rd party ways like custom screens in my modpack could easily add and manage custom overlays as currently it cannot be done without user input. If it is doable that way, then there would be no losses, I believe, and would be easily changeable when someone makes a typo or just wants to change the name as it is just a normal config field.

ShadowCat117:
> Up to you, but I have one idea for the overlay management screen. Do you think we could add user-specified/optionally custom naming for the custom overlays? It needs a bit of background work before we could do it though.

I have seen this requested a lot so yes it definitely could be done, and as Rafii said, I think a simple config field should get the job done

ShadowCat117:
The redesign is now going to be done in seperate parts so once a certain area is done it can be merged. This pr will focus on the "Wynntils" area, those being Settings, Overlays and Crowd Sourcing

ShadowCat117:
### No longer in this pr, was added in https://github.com/Wynntils/Artemis/pull/2502

Overlay Management Screen

- Option to render preview or just outline with name when editing.
- Option to hide non selected overlays.
- UI buttons can be moved to top/bottom of screen to make placing overlays in those areas easier.
- When rendering previews, a tooltip is shown with the name of the hovered overlay as some do not have a preview, or are not obvious what they are.

https://github.com/Wynntils/Artemis/assets/8337467/35f25dd7-deaa-45fa-bba9-8a41a327a664

JamieCallan117:
Crowd Sourcing screen

- No longer uses the book list layout.
- Send user a message upon copying to clipboard with amount of data collected.

Nothing much new here, just moving away from using the book for everything and giving it a bit of a data collection theme.
Temporary data types added just to show it supports scrolling once we start collecting more data types.

https://github.com/Wynntils/Artemis/assets/8337467/b9d3e2e7-08fb-40ec-85c2-96169595ffb8

JamieCallan117:
Also some pictures of the custom names for info boxes/custom bars being displayed

kristofbolyai:
Just letting you know I started reviewing this, but it is a bit too long to check in a single run. I'll be back.

common/src/main/java/com/wynntils/core/consumers/overlays/CustomNamedOverlay.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.consumers.overlays;
+
+import com.wynntils.core.persisted.config.Config;
+
+public interface CustomNamedOverlay {
```
kristofbolyai: CustomNameProperty?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -2082,8 +2082,13 @@
   "screens.wynntils.wynntilsContentBook.sort.distance.name": "Sort by Distance",
   "screens.wynntils.wynntilsContentBook.sort.level.description": "Quests are sorted by level, lowest first",
   "screens.wynntils.wynntilsContentBook.sort.level.name": "Sort by Level",
+  "screens.wynntils.wynntilsCrowdSourcing.collecting": "Collecting",
+  "screens.wynntils.wynntilsCrowdSourcing.copiedToClipboard": "Copied %d %s to clipboard",
```
```suggestion
  "screens.wynntils.wynntilsCrowdSourcing.copiedToClipboard": "Copied %d data points of %s to clipboard",
```

Maybe this is better?

common/src/main/java/com/wynntils/core/persisted/config/Config.java
```diff
@@ -49,7 +49,7 @@ public void setValue(T value) {
         this.userEdited = true;
     }
 
-    void restoreValue(Object value) {
+    public void setValueFromObject(Object value) {
```
kristofbolyai: No-no. Use setValue. This is not to be changed. The reason why it is not used is commented above the setValue method. TLDR, the system was rewritten, but never fully migrated to.

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
kristofbolyai, line 1: Not going to comment on each part, but all of the serialization/deserialization needs to move to ConfigManager.

kristofbolyai:
Can we fix the button size so the text doesn't overflow?
image

This UI's scrolling is way too hard with touchpads.
image

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
magicus: Yes. To clarify, you need to make functions to import settings from a json object (or a string), and conversely to export settings to a json object (or string). Then you can call this in the settings screen, and do the clipboard thing etc there.

ShadowCat117:
> Can we fix the button size so the text doesn't overflow?

Ah didn't realise there was a config with text long enough to overflow, I'll make it scrolling text as all of the buttons and input widgets are the same size for consistency. 

> This UI's scrolling is way too hard with touchpads.

It uses the same scrolling as the settings screen which is the same used on tm and poi manager so not sure why that screen specifically wouldn't work well, I can double check the implementation but I don't have a laptop currently set up for me to test it out on

common/src/main/java/com/wynntils/core/persisted/config/ConfigManager.java
```diff
@@ -262,4 +266,61 @@ public Stream<Config<?>> getConfigsForOwner(PersistedOwner owner) {
...
+                    // If the config data contains this config option, then it can be imported
+                    if (configData.containsKey(configOptionName)) {
+                        Object configOptionValue = configData.get(configOptionName);
+                        setConfigValue(configOption, configOptionValue);
+                    }
+                }
+            }
+
+            return true;
+        } catch (JsonSyntaxException ex) {
```
kristofbolyai, line 300: Isn't there more types of exception that can be thrown in the code above? If `setConfigValue` fails to due to type miss-match, or some other issue, we need to catch that as well.

kristofbolyai:
Scrolling feels good now! Also checked the button sizes, they look good.

common/src/main/java/com/wynntils/core/persisted/config/ConfigManager.java
ShadowCat117: `tryParseStringValue` catches a generic exception if it fails which then won't call `setValue` so I think it should be fine? I have now added a check for configData being null though as that wasn't checked

kristofbolyai: Yes, you are right.

kristofbolyai:
Essentially, this looks good to me, if we can get the textures added.

magicus (approved):
I've glanced through all code, and studied some parts in more particular. I think this looks good.

ShadowCat117:
I've removed the overlay and crowd sourcing redesigns as one was added separately and the other I'll do another time as not a fan of how it was done here.

There's also a few other fixes but no major changes from the original but might be worth giving it another review anyway.

I also failed at squashing so can mostly ignore the first ~15 commits

kristofbolyai:
So do we have textures now, or what happened?

ShadowCat117:
> So do we have textures now, or what happened?

Still no textures, Rafii has started some of them but no promises

magicus:
> I also failed at squashing so can mostly ignore the first ~15 commits

At some points it's easier to make a diff relative to main, and apply that diff to a new branch and create a new PR. I think that is easier than trying to play with git's rewrite functionality. (Rebasing/squashing an open PR is also taboo; that makes the reviews hopeless to follow later on.)

Rafii2198:
The finished textures
image

kristofbolyai (approved):
Looks good! We can merge this now, if no other texture changes are pending.

Rafii2198:
Yeah, all textures are done, I think there should be an announcement announcing the change as it is pretty impactful to some at least.

ShadowCat117:
To be honest, if MapData is nearly complete it could be saved for 1.3.0 I don't mind

ShadowCat117:
Would save having 2 announcements close to each other

kristofbolyai:
I forgot to comment, but I'll think until tomorrow, too tired for the announcement today. Mapdata itself is done, but porting will take weeks, so don't worry about that.

ShadowCat117:
Alright, I still need to do the UI for MapData anyway and that'll probably take me a few days

PR #2337, opened 2024-02-05
fix: Make sure all location abbreviations are unique (for scrolls and seaskipper tags)

ryanzhoudev:
Workaround for common duplicate scroll abbreviations
https://feedback.wynntils.com/posts/279/for-the-nemract-and-nesaak-scrolls

Merge together with https://github.com/Wynntils/Static-Storage/pull/53

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
@@ -1,9 +1,11 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.items.annotators.game;
 
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.components.Services;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.handlers.item.ItemAnnotation;
 import com.wynntils.handlers.item.ItemAnnotator;
@@ -42,7 +44,13 @@ public final class TeleportScrollAnnotator implements ItemAnnotator {
                     .toUpperCase(Locale.ROOT);
             return new TeleportScrollItem(destination, true);
         } else {
-            String destination = scrollName.substring(0, 2);
+            String destination = Services.Destination.getAbbreviation(scrollName);
+            if (destination == null) {
+                // fallback
+                destination = scrollName.substring(0, 2);
+                WynntilsMod.warn(
+                        "TeleportScrollAnnotator: No destination found for " + scrollName + ", using fallback");
+            }
             return new TeleportScrollItem(destination, false);
         }
     }
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2024.
+ * Copyright © Wynntils 2022-2023.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.items.annotators.game;
@@ -42,27 +42,7 @@ public final class TeleportScrollAnnotator implements ItemAnnotator {
                     .toUpperCase(Locale.ROOT);
             return new TeleportScrollItem(destination, true);
         } else {
-            String destination =
-                    switch (scrollName) {
-                            // Originally Ne
-                        case "Nemract" -> "Nr";
-                        case "Nesaak" -> "Ns";
-
-                            // Originally Th
-                        case "Thesead" -> "Ts";
-                        case "Thanos" -> "Tn";
-
-                            // Originally Al
-                        case "Almuj" -> "Aj";
-                        case "Aldorei Town" -> "At";
-
-                            // Originally Ma
-                        case "Maex" -> "Mx";
-                        case "Mage Island" -> "Mi";
-
-                        default -> scrollName.substring(0, 2);
-                    };
-
+            String destination = scrollName.substring(0, 2);
             return new TeleportScrollItem(destination, false);
         }
     }
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
@@ -42,7 +42,26 @@ public final class TeleportScrollAnnotator implements ItemAnnotator {
                     .toUpperCase(Locale.ROOT);
             return new TeleportScrollItem(destination, true);
         } else {
-            String destination = scrollName.substring(0, 2);
+            String destination = switch (scrollName) {
+                // Originally Ne
+                case "Nemract" -> "Nr";
+                case "Nesaak" -> "Ns";
+                
+                // Originally Th
+                case "Thesead" -> "Ts";
+                case "Thanos" -> "Tn";
+                
+                // Originally Al
+                case "Almuj" -> "Aj";
+                case "Aldorei Town" -> "At";
+
+                // Originally Ma
+                case "Maex" -> "Mx";
+                case "Mage Island" -> "Mi";
+                
+                default -> scrollName.substring(0, 2);
+            };
+
             return new TeleportScrollItem(destination, false);
         }
     }
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.items.annotators.game;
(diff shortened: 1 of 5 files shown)


kristofbolyai (commented):
I don't think the annotator should hold this kind of info, as it mixes responsibility between different "API" levels. When I get to a PC, I'll try to suggest a better place for this to exist.

ryanzhoudev:
> I don't think the annotator should hold this kind of info, as it mixes responsibility between different "API" levels. When I get to a PC, I'll try to suggest a better place for this to exist.

Maybe we just pass the full name into TeleportScrollItem, and have a thing in the constructor that determines this abbreviation?

ShadowCat117:
Just as an FYI, this issue probably also occurs for the Seaskipper passes as there can be Nesaak and Nemract, less of a problem though as I'd imagine most people will use either of the screens

magicus:
This is kind of a PlaceNameModel thing to know. I think bolyai can find a good place to put it.

Also, one remark: I think it is better if you use the next non-conflicting significant letter, so "Almuj" would be "Am", and "Aldorei Town" would be "Ad". That would make the naming more consistent.

And if you really want to keep two-word names using initials, then e.g. "Mage Island" should be "MI" with a capital "I". (I think that is a worse approach, just to be clear.)

kristofbolyai:
magicus commented what I imagined last night.

If we could have a list of all scroll location names, we could pre-calculate the letters with a specific user preference, and without needing manual updates ever. As for the way on how to get the shortened letter representation, magicus' idea is the same as mine. Start with the first letter, skip the next ones while there are conflicting ones.

kristofbolyai:
So, essentially, what needs to be done here is a model where we (pre-)load all town/location names. For unknown names, we can calculate on the fly, still avoiding duplicates. I don't think we have a way to automatically collect location names, so most likely this is going to have to be a manual data source.

magicus:
I think the place name json file might be a good start.

ryanzhoudev:
> Also, one remark: I think it is better if you use the next non-conflicting significant letter, so "Almuj" would be "Am", and "Aldorei Town" would be "Ad". That would make the naming more consistent.

I don't think consistency should be a priority here - we're working with humans, not computers after all. Almuj makes more sense to me as "Aj" since the "j" is very distinct. "Am" is possible to confuse with Ahmsord. "At" would make sense for Aldorei Town, and I don't think there are any duplicates of that. Nemract - I think "Nr" would be better than "Nm" since "ract" is it's own syllable.

ryanzhoudev:
> > Also, one remark: I think it is better if you use the next non-conflicting significant letter, so "Almuj" would be "Am", and "Aldorei Town" would be "Ad". That would make the naming more consistent.
> 
> I don't think consistency should be a priority here - we're working with humans, not computers after all. Almuj makes more sense to me as "Aj" since the "j" is very distinct. "Am" is possible to confuse with Ahmsord. "At" would make sense for Aldorei Town, and I don't think there are any duplicates of that. Nemract - I think "Nr" would be better than "Nm" since "ract" is it's own syllable.

Though, we could actually automate this if we used the next non-conflicting letter...

common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
```diff
@@ -42,7 +44,13 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
                     .toUpperCase(Locale.ROOT);
             return new TeleportScrollItem(destination, true);
         } else {
-            String destination = scrollName.substring(0, 2);
+            String destination = Services.Destination.getAbbreviation(scrollName);
+            if (destination == null) {
+                // fallback
+                destination = scrollName.substring(0, 2);
```
ShadowCat117: Maybe the fallback can be done inside the service? It would mean we'd lose which annotator failed though so up to you

common/src/main/java/com/wynntils/services/destination/DestinationService.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Get the abbreviation of a location
+     * @param location Full location name of the destination (eg. "Ragni")
+     * @return Abbreviation of the location (eg. "Ra") or null if not found
+     */
+    public String getAbbreviation(String location) {
+        if (!destinations.containsKey(location)) {
+            return null;
+        }
+        return destinations.get(location);
```
ShadowCat117, line 38: I don't think the containsKey check is needed as it should return null if the key isn't found

common/src/main/java/com/wynntils/models/items/annotators/game/TeleportScrollAnnotator.java
ryanzhoudev: Yeah, this might actually be required otherwise the substring might end up with a dupe anyway. I'll change it

common/src/main/java/com/wynntils/services/destination/DestinationService.java
ryanzhoudev: Doesn't it cause npe?

ShadowCat117: Though if you do move the fallback here, then it would be done inside the if

ShadowCat117: > Doesn't it cause npe?

I believe it just returns null, might need to double check

common/src/main/java/com/wynntils/services/destination/DestinationService.java
```diff
@@ -0,0 +1,71 @@
...
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.components.Managers;
+import com.wynntils.core.components.Service;
+import com.wynntils.core.net.Download;
+import com.wynntils.core.net.UrlId;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DestinationService extends Service {
```
magicus, line 16: This is not a `Service`, it is a `Model`. If you ever gets unsure, read the description in `Model.java` and `Service.java`:

```
 * a Model.
 ```
and
```
 * Services provide support to consumers for anything that is not specifically
 * related to the Wynncraft model, either since it is an external service
 * (3rd party library, or system/environment interaction), or since it is something
 * that we augment the Wynncraft world with, but the Wynncraft servers has no knowledge
 * about. In other words, this is a "misc" category, where everything consumers need
 * that is neither a manager, handler or manager can be placed.
```

**Review by magicus — APPROVED** [review]
Looks good. 

For a time I was thinking this should be a Model rather than a Service, but I think you are right that this is better treated as a Service; the abbreviation is an additional functionality we provide to Wynncraft.

File: common/src/main/java/com/wynntils/services/destination/DestinationService.java [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+            while (destinations.containsValue(abbreviation)) {
+                if (location.charAt(i) == ' ') i++;
+                abbreviation = location.charAt(0) + "" + location.charAt(i);
+                i++;
+            }
+            WynntilsMod.warn("DestinationService: No destination found for " + location
+                    + ", using fallback abbreviation " + abbreviation);
+            return abbreviation;
+        }
+        return destinations.get(location);
```
kristofbolyai: I would move the generation code into a new method, and also put a `computeIfAbsent` for the map, so it does not have to be generated.

common/src/main/java/com/wynntils/services/destination/DestinationService.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private void loadDestinations() {
+        destinations = new HashMap<>();
+        Download dl = Managers.Net.download(UrlId.DATA_STATIC_DESTINATIONS);
+        dl.handleReader(reader -> {
+            Destination result = WynntilsMod.GSON.fromJson(reader, Destination.class);
+            for (Destination.DestinationDetail destination : result.destinations) {
+                destinations.put(destination.location, destination.abbreviation);
```
kristofbolyai: What we generally do when loading anything from net (which is basically async, since the user can trigger a reload) is to use a temp map to load all destinations into, then set the actual field instance. This way, you still have the old destinations if the download fails, plus, you are safe for concurrency.

common/src/main/java/com/wynntils/services/destination/DestinationService.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Finds the next non-conflicting abbreviation for a location, excluding spaces
+     * @param location Full location name of the destination (eg. "Ragni")
+     */
+    private String findNextAbbreviation(String location) {
+        String abbreviation = location.substring(0, 2);
+        int i = 2;
+        while (destinations.containsValue(abbreviation)) {
+            if (location.charAt(i) == ' ') i++;
```
kristofbolyai: This can go out of bounds. Also, you might as well make this a loop to check for multiple spaces, while you are at it.

common/src/main/java/com/wynntils/services/destination/DestinationService.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Download dl = Managers.Net.download(UrlId.DATA_STATIC_DESTINATIONS);
+        dl.handleReader(reader -> {
+            Destination result = WynntilsMod.GSON.fromJson(reader, Destination.class);
+            for (Destination.DestinationDetail destination : result.destinations) {
+                newDestinations.put(destination.location, destination.abbreviation);
+            }
+        });
+
+        destinations = newDestinations;
```
kristofbolyai: You need to do this in the handlerReader lamba, after the for loop.

kristofbolyai (approved):
Looks good now! Thanks!

PR #2338, opened 2024-02-07
feat: add scaling to waypoint distance markers

lith-x:
Adds scaling to waypoint distance markers, decently simple change. Felt like I needed this when I saw the marker was way too big for my liking, so I did a quick implementation. The comment about the "4f" was a guess but that seems to be what that magic value is pointing to. First pull request, hopefully I haven't done anything too offensive coding-wise lol

edit: reading back on the comments, I was making a lot of assumptions in general that could be just wrong, kinda unsure about them now


kristofbolyai:
Hi there! Thanks for contributing to the project! Please allow us a few days to review. Thank you!

lith-x:
Sounds good I'll keep an eye out!

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
```diff
@@ -139,62 +142,66 @@ public void onRenderGuiPost(RenderEvent.Post event) {
...
             } else {
                 displayPositionX = intersectPoint.x;
                 displayPositionY = intersectPoint.y;
 
+                // pointer position is determined by finding the point on circle centered around displayPosition
+                double angle = Math.toDegrees(StrictMath.atan2(
+                                displayPositionY - event.getWindow().getGuiScaledHeight() / 2,
+                                displayPositionX - event.getWindow().getGuiScaledWidth() / 2))
+                        + 90f;
+                float radius = icon.width() / 2 + 8f;
```
kristofbolyai, line 184: Theoretically, this should be scaled, as well?

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
```diff
@@ -236,12 +243,15 @@ private Vec2 getBoundingIntersectPoint(Vec3 position, Window window) {
         if (isInBound(position, window)) return null;
         Vec3 centerPoint = new Vec3(window.getGuiScaledWidth() / 2, window.getGuiScaledHeight() / 2, 0);
 
+        // compensate for scaling, 4f is half the pointer icon height (used to anchor render positions)
+        float scaleFactor = 4f * (1 - scale.get());
```
kristofbolyai:
```suggestion
        float scaleFactor = Texture.POINTER.height() * (1 - scale.get());
```

kristofbolyai:
Looks good to me, in-game!

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
lith-x: This is basically only used for pointerDisplayPositionX and Y vars, which needs to be kept unscaled, since it's the point I'm using to scale from/to keep the bounding box visually intact during the scaling process.

magicus (approved):
A superficial glance of the code look okay, but I did not go into details if the math makes sense.

common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
kristofbolyai: Okay.

kristofbolyai (approved):
I've also checked the math and tested in-game, I think it looks good.

kristofbolyai:
I've left this out, but thank you for your contribution!

PR #2343, opened 2024-02-14
fix: Parse set data to fix skill point model saving skill requirements incorrectly

ryanzhoudev:
https://github.com/Wynntils/Static-Storage/pull/67

- [x] GearItem implements Optionals of SetInstance and SetInfo
- [x] Add a method into GearItems to check if they have requirements met
- [x] SetModel holds list of item slots and true count and collects set instances
- [x] SetInstance holds name of set (SetInfo), items that are/aren't part of set, active item count, current theoretical bonuses
- [x] GearAnnotator needs to also check wynncraft count when annotating, and take max of the two counts - otherwise when gear is just sent, some gear hasn't arrived yet when annotation starts, resulting in bad true count (actually i am gonna cascade setinstance data)


magicus:
Yay! 🎉  Great to see you start working on fixing this.

common/src/main/java/com/wynntils/models/character/type/SetInfo.java
```diff
@@ -0,0 +1,34 @@
...
+    private final int wynncraftCount;
+    private int trueCount = 0;
+    private final ItemStack relevantItem;
+
+    public SetInfo(int wynncraftCount, ItemStack relevantItem) {
+        this.wynncraftCount = wynncraftCount;
+        this.relevantItem = relevantItem;
+    }
+
+    public int getWynncraftCount() {
```
magicus: Is Wynncraft count counting rings twice, or the other way around? Or do I misunderstand this? A comment describing the differences between the "count" functions would be good.

common/src/main/java/com/wynntils/models/character/type/SetInfo.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.character.type;
+
+import net.minecraft.world.item.ItemStack;
+
+public class SetInfo {
```
magicus: At first I thought "SetInfo" was what you called "SetInstance", but it seems you are using the names the other way around. For me, "SetInfo" sounds like a name for a generic information about the sets that exists in Wynncraft, while a SetInstance might be an actual instance that you are currently wearing.

But the purpose of this class seem to be only to handle the duplicate ring issue, right? Then I suggest the name should be more direct, like DuplicateSetItemDetector.

common/src/main/java/com/wynntils/models/character/type/SetInfo.java
ryanzhoudev: If you have the same ring on both slots it is only counted once

common/src/main/java/com/wynntils/models/character/type/SetInfo.java
ryanzhoudev: I left it as SetInstance. The class also collects some item relevant to a set so it can be parsed later.

ryanzhoudev:
If this happens to be approved, please don't merge, just let me know - I'll do it myself and will consider adding a [skip ci] based on how far along I am with the next fix

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -52,13 +57,20 @@ public class SkillPointModel extends Model {
     private static final String EMPTY_ACCESSORY_SLOT = "§7Accessory Slot";
     private static final int CHARACTER_INFO_SOUL_POINT_SLOT = 62;
     private static final int TOME_MENU_SOUL_POINT_SLOT = 89;
+    private static final Pattern SET_PATTERN = Pattern.compile("§a(.+) Set §7\\((\\d)/\\d\\)");
+    private static final String SET_BONUS_HEADER = "§aSet Bonus:";
+    private static final Pattern BONUS_SKILL_POINT_PATTERN =
+            Pattern.compile("§[ac]([+-]\\d+) §7(Strength|Dexterity|Intelligence|Defence|Agility)");
```
ShadowCat117: Since we have the TestRegex setup now, should we be adding all future patterns to it?

ShadowCat117 (approved):
Looks good and works well in game :)

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
kristofbolyai: We don't necessarily test every regex, but yes, if we want tests, we do it there, not regexr.

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -292,16 +312,75 @@ private void loadSkillPointsOnServer(ContainerContent containerContent, String n
...
+                        }
+                        continue;
+                    }
+                    Matcher m = BONUS_SKILL_POINT_PATTERN.matcher(line.getString());
+                    if (!m.matches()) continue;
+
+                    int value = Integer.parseInt(m.group(1));
+                    Skill skill = Skill.fromString(m.group(2));
+                    setBonusSkillPoints.merge(skill, value, Integer::sum);
+                }
```
kristofbolyai: This really should be integrated into the item models we already have.

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,90 @@
...
+    public SetInfo getSetData(String setId) {
+        return setData.get(setId);
+    }
+
+    private void loadSetData() {
+        Download dl = Managers.Net.download(UrlId.DATA_STATIC_ITEM_SETS);
+        dl.handleReader(reader -> {
+            Map<String, Map<String, List<?>>> tempData =
+                    WynntilsMod.GSON.fromJson(reader, new TypeToken<Map<String, Map<String, List<?>>>>() {}.getType());
+            tempData.forEach((key, value) -> {
```
kristofbolyai: Why not create a class that represents the JSON value?

kristofbolyai: We often use private "decoder" classes so that we don't need to use such big type definitions, and GSON can still deserialize the data.

common/src/main/java/com/wynntils/utils/wynn/InventoryUtils.java
```diff
@@ -54,6 +61,26 @@ public static ItemStack getItemInHand() {
...
+        for (StyledText loreLine : LoreUtils.getLore(itemStack)) {
+            if (loreLine.startsWith("§c✖") && loreLine.contains("Min: ")) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
     public enum MouseClickType {
```
kristofbolyai: This method does not fit this util class. But I don't think it should be moved, instead you should be able to to make GearItem/CraftedGearItem able to tell us if all requirements are met. Perhaps the start would be creating a `RequirementItemProperty`.

kristofbolyai (changes requested):
Although functionally this works greatly, I have to hold this PR back, even though I understand that this is meant to be a "quick" fix.

There are 2-3 code parts that need to be integrated into GearModel, and into our item models (to be clear, I am talking about `GearItem`, etc.) directly. I wouldn't want us to merge this PR since it would create quite a bit of technical debt for us in the future, not to mention, if other features would start to use the code, the breakage would even be higher in the said future.

I think the implementation into the gear models should be straightforward, but please let me know if you have any questions, so we can discuss instead of guessing and inevitably wasting time.

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public Set<SetInstance> getUniqueSetInstances() {
+        return Set.copyOf(setInstances.values());
+    }
+
+    private void loadSetData() {
+        Download dl = Managers.Net.download(UrlId.DATA_STATIC_ITEM_SETS);
+        dl.handleReader(reader -> {
+            Map<String, RawSetInfo> rawSets =
+                    WynntilsMod.GSON.fromJson(reader, new TypeToken<Map<String, RawSetInfo>>() {}.getType());
```
kristofbolyai:
```suggestion
                     Managers.Json.GSON.fromJson(reader, new TypeToken<Map<String, RawSetInfo>>() {}.getType());
```

We should use the JsonManager GSON instance. (I am aware that some other models use it incorrectly)

kristofbolyai: Also, we while you are here, we usually extract the type tokens to a separate variable, although this is just a nit (but you might as well do it with the other change.)

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,108 @@
...
+            rawSets.forEach((setName, rawSetInfo) -> {
+                List<Map<StatType, Integer>> bonuses = rawSetInfo.bonuses.stream()
+                        .map(bonusPair -> {
+                            Map<StatType, Integer> bonusMap = new HashMap<>();
+                            bonusPair.forEach((statName, statValue) -> {
+                                StatType statType = Models.Stat.fromApiName(statName);
```
kristofbolyai: So it looks like this is a dependency for loading this URL. This should work serially, like `IngredientInfoRegistry#onDataLoaded` (so just call this from that event).

Also, implement `reloadData` and call this method there too.

common/src/main/java/com/wynntils/models/items/items/game/GearItem.java
```diff
@@ -33,13 +37,23 @@ public class GearItem extends GameItem
...
                 ShinyItemProperty,
-                IdentifiableItemProperty<GearInfo, GearInstance> {
+                IdentifiableItemProperty<GearInfo, GearInstance>,
+                SetItemProperty,
+                RequirementItemProperty {
     private final GearInfo gearInfo;
     private final GearInstance gearInstance;
-
-    public GearItem(GearInfo gearInfo, GearInstance gearInstance) {
+    private final Optional<SetInfo> setInfo;
```
kristofbolyai: Let's address `SetInfo` first. You should be able to put this in `GearInfo`, load gear data serially, just like I mentioned before. That way, you won't have a race when trying to get info from a model that may be still loading.

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -48,9 +51,10 @@ public final class GearModel extends Model {
 
     private final GearChatEncoding gearChatEncoding = new GearChatEncoding();
     private final Map<GearBoxItem, List<GearInfo>> possibilitiesCache = new HashMap<>();
+    private final Map<String, List<SetInstance>> activeSetsCache = new HashMap<>();
 
-    public GearModel() {
-        super(List.of());
+    public GearModel(SetModel setModel) {
```
kristofbolyai: This dependency is not needed, since you don't use the set model from the constructor. I think you felt that this model depends on the set data, but this is not the way to ensure the data is loaded, since this just ensures that `SetModel`'s constructor ran, which loads the data **asynchronously**.

common/src/main/java/com/wynntils/models/gear/type/SetInfo.java
```diff
@@ -0,0 +1,26 @@
...
+     * @return A map of stat names to the bonus value for that stat
+     */
+    public Map<StatType, Integer> getBonusForItems(int numberOfItems) {
+        return bonuses.get(Mth.clamp(numberOfItems, 1, items.size()) - 1);
+    }
+
+    @Override
+    public String toString() {
+        return "SetInfo{" + "name='" + name + '\'' + ", bonuses=" + bonuses + ", items=" + items + '}';
+    }
```
kristofbolyai: Records, by default, implement "pretty" `toString` methods, is there a special reason why you override it?

common/src/main/java/com/wynntils/models/gear/type/SetInstance.java
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.gear.type;
+
+import com.wynntils.core.components.Models;
+import com.wynntils.models.stats.type.StatType;
+import java.util.Map;
+
+// wynncraftCount can eventually be removed when Wynncraft fixes their double ring bug...
```
kristofbolyai: This is actually backwards. True count can be removed when Wynn fixes their issue, probably along with the non-automatically updated set data, since it won't be accurate. We should rely on in-game parsed (true) data, unless the API has set info data.

common/src/main/java/com/wynntils/models/items/items/game/GearItem.java
```diff
@@ -33,13 +37,23 @@ public class GearItem extends GameItem
...
+    private final Optional<SetInstance> setInstance;
```
kristofbolyai: And now, let's address `SetInstance`. It took me a while to get a grip on the issue, but I think I basically got it now.

Let's remove `trueRoll` from `SetInstance`. That means that you can populate `SetInstance` when parsing everything `GearInstance` related, purely from the tooltip (even if you get wrong data when joining a world, items are going to get resent shortly, with correct tooltips).  You can remove every hacky, reference-based changes (if there is any left) in `SetModel` too.  While you are at it, move `Optional<SetInstance>` into `GearInstance`.

This means that most of the changes are gone, but we still have set instance data. For actually getting the true roll data, get it through `SetModel`. This should also mean that whenever Wynn fixes this bug, we only have to remove parts of the model, without touching anything else.

kristofbolyai (commented):
We are headed in the right direction. There are a few changes needed with how the info/instance is added to the item, so we don't break our abstractions. I feel like you almost wanted to do it this way at first, but struggled with the data-dependencies.

I'll probably need to look into the GearSlot related methods in a separate round, although those are much closer to good, as-is.

common/src/main/java/com/wynntils/models/gear/type/SetInfo.java
ryanzhoudev: Was originally a class, intellij didn't clean up after i used the convert to record thing

common/src/main/java/com/wynntils/models/gear/SetModel.java
ryanzhoudev: Sorry I am not too certain on what you mean by this. I kind of just copied what is going on in the IngredientInfoRegistry. Not sure where `reloadData` would go. My attempt is the "serially load gear/set infos" commit. It seems to work, but not sure if you want the syntax some other way.

kristofbolyai: `CoreComponent`'s (Models, Services, etc.) have a method called `reloadData`, which gets called in special cases, like `wynntils reloadcache`. Basically we just override it below the constructor, call `loadSetData` (or anything else) and that's it (and also call it from the constructor, if applicable).

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -358,7 +387,7 @@ private boolean checkTomesUnlocked(ContainerContent content) {
 
     private boolean verifyChange(
             ContainerContent content,
-            Int2ObjectMap<ItemStack> changes,
+            Int2ObjectFunction<ItemStack> changes,
```
kristofbolyai, line 380: Although this change is not harmful, is there a reason for it, or why was it changed?

common/src/main/java/com/wynntils/models/containers/PlayerInventoryModel.java
```diff
@@ -42,6 +43,36 @@ public CappedValue getIngredientPouchSlots() {
         return new CappedValue(getUsedIngredientPouchSlots(), MAX_INGREDIENT_POUCH_SLOTS);
     }
 
+    /**
+     * @return -1 if the item is not equipped, otherwise the slot number
+     */
+    public int getEquippedItemSlot(ItemStack itemStack) {
+        int[] accessorySlots = {9, 10, 11, 12};
```
kristofbolyai: Can you make this private static final?

Nit: While you are here, could you move this method below `unregisterWatcher`?

common/src/main/java/com/wynntils/models/containers/PlayerInventoryModel.java
```diff
@@ -42,6 +43,36 @@ public CappedValue getIngredientPouchSlots() {
...
+        for (int i = 0; i < 4; i++) {
+            ItemStack equipped = McUtils.inventory().armor.get(i);
+            if (ItemStack.isSameItem(equipped, itemStack)) {
+                return i + 5;
+            }
+        }
+        if (ItemStack.isSameItem(McUtils.player().getItemInHand(InteractionHand.MAIN_HAND), itemStack)) {
+            return McUtils.inventory().selected;
+        }
+        return -1;
```
kristofbolyai: Let's add some newlines.

```suggestion
        for (int slot : accessorySlots) {
            ItemStack equipped = McUtils.inventory().getItem(slot);
            if (ItemStack.isSameItem(equipped, itemStack)) {
                return slot;
            }
        }
        
        for (int i = 0; i < 4; i++) {
            ItemStack equipped = McUtils.inventory().armor.get(i);
            if (ItemStack.isSameItem(equipped, itemStack)) {
                return i + 5;
            }
        }
        
        if (ItemStack.isSameItem(McUtils.player().getItemInHand(InteractionHand.MAIN_HAND), itemStack)) {
            return McUtils.inventory().selected;
        }
        
        return -1;
```

common/src/main/java/com/wynntils/models/containers/PlayerInventoryModel.java
```diff
@@ -42,6 +43,36 @@ public CappedValue getIngredientPouchSlots() {
...
+        for (int slot : accessorySlots) {
+            ItemStack equipped = McUtils.inventory().getItem(slot);
+            if (ItemStack.isSameItem(equipped, itemStack)) {
+                return slot;
+            }
+        }
+        for (int i = 0; i < 4; i++) {
+            ItemStack equipped = McUtils.inventory().armor.get(i);
+            if (ItemStack.isSameItem(equipped, itemStack)) {
+                return i + 5;
```
kristofbolyai: `+ 5`? 
(Add the comment in code too, not only here).

common/src/main/java/com/wynntils/models/gear/GearChatEncoding.java
```diff
@@ -207,9 +207,9 @@ public GearItem fromEncodedString(String encoded) {
 
         // create chat gear stack
         // Note that the chat procotol does not allow for shiny stats to be transferred
-        GearInstance gearInstance =
-                GearInstance.create(gearInfo, identifications, powderList, rerolls, Optional.empty());
-        return new GearItem(gearInfo, gearInstance);
+        GearInstance gearInstance = GearInstance.create(
+                gearInfo, identifications, powderList, rerolls, Optional.empty(), false, Optional.empty());
+        return new GearItem(gearInfo, gearInstance); // TODO: populate setInfo but not setInstance
```
kristofbolyai: Is this TODO actually up-to-date? 

All GearInfo's should already have proper SetInfos. For setInstance, we simply don't care here, no comment needed.

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -80,24 +85,49 @@ public boolean canBeGearBox(GearInfo gear) {
...
     }
 
+    // For "real" gear items eg. from the inventory
     public GearInstance parseInstance(GearInfo gearInfo, ItemStack itemStack) {
         WynnItemParseResult result = WynnItemParser.parseItemStack(itemStack, gearInfo.getVariableStatsMap());
         if (result.tier() != gearInfo.tier()) {
             WynntilsMod.warn("Tier for " + gearInfo.name() + " is reported as " + result.tier());
         }
 
+        // probably migrate this to GearInstance?
```
kristofbolyai: Let's migrate this to the call just above this one :) `WynnItemParser#parseItemStack` already does the requirements parsing, adding a check if just a matter of adding/checking a group in the regex. You can return the value in `WynnItemParseResult`.

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -80,24 +85,49 @@ public boolean canBeGearBox(GearInfo gear) {
...
+        boolean meetsRequirements = true;
+        for (StyledText loreLine : LoreUtils.getLore(itemStack)) {
+            if (loreLine.startsWith("§c✖") && loreLine.contains("Min: ")) {
+                meetsRequirements = false;
+                break;
+            }
+        }
+
+        // TODO: parse set here!!
```
kristofbolyai: It seems pretty important to parse the set instance here, no?

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,154 @@
...
+                Matcher nameMatcher = SetModel.SET_PATTERN.matcher(line.getString());
+                if (nameMatcher.matches() && nameMatcher.group(1).equals(setName)) {
+                    trueCount++;
+                }
+            }
+        }
+
+        Optional<WynnItem> wynnItem =
+                Models.Item.getWynnItem(McUtils.player().getItemInHand(InteractionHand.MAIN_HAND));
+        if (wynnItem.isPresent() && wynnItem.get() instanceof GearItem gearItem && gearItem.meetsActualRequirements()) {
```
kristofbolyai: This is basically duplicated code. Perhaps add a method in `PlayerInventoryModel`, `getWornItems` or something along the lines?

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,154 @@
...
+            if (entry.getValue().items().contains(itemName)) {
+                return entry.getKey();
+            }
+        }
+        return null;
+    }
+
+    public void updateSetInstance(GearSlot slot, SetInstance instance) {
+        setInstances.put(slot, instance);
+    }
```
kristofbolyai: What if we just didn't save this data in the model? It is a weird dependency, having the annotator call this model.

Do you see anything wrong with just storing the set instance in the GearItem, and just collecting all worn items and checking them in `getTrueCount`?

kristofbolyai: Also, you basically don't even use this map anymore from what I can see. Collecting the unique worn sets can be done by collecting it from the gearItems. (to clarify, `getUniqueSetNames` should stay, but should be modified).

kristofbolyai: Also, with the map removed, `GearSlot` itself can go.

common/src/main/java/com/wynntils/models/gear/type/SetInstance.java
```diff
@@ -0,0 +1,55 @@
...
+        this.setInfo = setInfo;
+        this.activeItems = activeItems;
+        this.wynnCount = wynnCount;
+    }
+
+    public void update(SetInfo setInfo, Map<String, Boolean> activeItems, int wynnCount) {
+        this.setInfo = setInfo;
+        this.activeItems = activeItems;
+        this.wynnCount = wynnCount;
+    }
```
kristofbolyai: The method is unused now. In fact, this is now an immutable class (as it should be, like GearInstance is), so you can make this a record, and remove every method.

common/src/main/java/com/wynntils/models/items/annotators/game/GearAnnotator.java
```diff
@@ -34,6 +48,109 @@ public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
...
+                        setInfo.get(), getActiveItems(setInfo.get().name()), wynnCount));
+                GearSlot slot;
+                if (gearInfo.type() == GearType.RING) {
+                    slot = equippedItemSlot % 9 == 0 ? GearSlot.RING1 : GearSlot.RING2;
+                } else {
+                    slot = GearSlot.fromGearType(gearInfo.type());
+                }
+                Models.Set.updateSetInstance(slot, setInstance.get());
+            }
+        }
```
kristofbolyai: And with the changes done in the previous suggestions, you can remove this whole part. GearInfo already has the set info from "upstream", and gear instance should be populated "upstream" too.

kristofbolyai: Aaand, that means `getActiveItems` and `getWynncraftCount` goes as well.

common/src/main/java/com/wynntils/models/items/items/game/CraftedGearItem.java
```diff
@@ -156,6 +158,11 @@ public GearTier getGearTier() {
         return GearTier.CRAFTED;
     }
 
+    @Override
+    public boolean meetsActualRequirements() {
+        return false; // TODO
```
kristofbolyai: Should this be fixed in this PR? I assume the calculation wouldn't work correctly if we hold a weapon with a skill bonus, but we cannot use it? Does crafted items "count" for skill points?

common/src/main/java/com/wynntils/models/items/properties/RequirementItemProperty.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.properties;
+
+public interface RequirementItemProperty {
+    /**
+     * @return true if the item's lore has no X's in it, for the requirements section
```
kristofbolyai:
```suggestion
     * @return true if the item is usable by the player
```

kristofbolyai: Although your comment is true, it is an implementation detail, and probably not the info that we should hold here.

common/src/main/java/com/wynntils/screens/gearviewer/GearViewerScreen.java
kristofbolyai, line 1: Seems like you can remove all the changes here.

kristofbolyai (commented):
Phew! After the changes, I really don't see anything else holding this PR back.

Integrating set data into our item models were not done on purpose, since it required too much time back then, and we did not need it. I think you can see why noone wanted to give it thought and time back then, but now that we need it, it's obvious how niche the implementation really is.  Overall, I am really happy with how the result will turn out. (In fact, I could only be happier if the API provided set data, and wynn fixed their bug.)

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
ryanzhoudev: Intellij kept doing a highlight + suggestion about weakening the type so I thought might as well

common/src/main/java/com/wynntils/models/items/items/game/CraftedGearItem.java
ryanzhoudev: Crafteds should be handled separately, i'll check for this function in a bit

common/src/main/java/com/wynntils/models/containers/PlayerInventoryModel.java
ryanzhoudev: wiki.vg/Inventory, helmet starts at 5 for normal inventory numbers. Added as a comment too

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -91,22 +88,28 @@ public void saveCurrentSkillPoints(String name) {
...
+            if (wynnItemOptional.isEmpty()) continue;
+            WynnItem wynnItem = wynnItemOptional.get();
+
+            if (wynnItem instanceof GearTypeItemProperty gear) {
+                if (gear.getGearType().isArmor()) {
+                    armourNames.add(itemStack.getHoverName().getString());
+                } else if (gear.getGearType().isAccessory()) {
+                    accessoryNames.add(itemStack.getHoverName().getString());
+                } else if (gear.getGearType().isWeapon()) {
+                    weapon = itemStack.getHoverName().getString();
```
kristofbolyai: Usually we don't use Component#getString, because it can produce unexpected output when Wynn changes some small thing in their back-end. Use StyledText#getStringWithoutFormatting

common/src/main/java/com/wynntils/models/containers/PlayerInventoryModel.java
```diff
@@ -51,6 +58,37 @@ public void unregisterWatcher(InventoryWatcher watcher) {
...
+        }
+
+        Optional<WynnItem> wynnItem =
+                Models.Item.getWynnItem(McUtils.player().getItemInHand(InteractionHand.MAIN_HAND));
+        if (wynnItem.isPresent()
+                && ((wynnItem.get() instanceof GearItem gearItem && gearItem.meetsActualRequirements())
+                        || (wynnItem.get() instanceof CraftedGearItem craftedGearItem
+                                && craftedGearItem.meetsActualRequirements()))) {
+            returnable.add(McUtils.player().getItemInHand(InteractionHand.MAIN_HAND));
+        }
```
kristofbolyai: Use `ItemModel#asWynnItemProperty`

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -184,4 +211,34 @@ public GearInfo getGearInfoFromApiName(String apiName) {
     public Stream<GearInfo> getAllGearInfos() {
         return gearInfoRegistry.getGearInfoStream();
     }
+
+    private Map<String, Boolean> getActiveItems(String setName) {
+        Map<String, Boolean> activeItems = new HashMap<>();
+
+        List<String> itemNames = Models.PlayerInventory.getEquippedItems().stream()
+                .map(itemStack -> itemStack.getHoverName().getString())
```
kristofbolyai: Same as before, it's preferable to use StyledText

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -184,4 +211,34 @@ public GearInfo getGearInfoFromApiName(String apiName) {
     public Stream<GearInfo> getAllGearInfos() {
         return gearInfoRegistry.getGearInfoStream();
     }
+
+    private Map<String, Boolean> getActiveItems(String setName) {
```
kristofbolyai: Rename to `getActiveSetItems`

common/src/main/java/com/wynntils/models/gear/GearModel.java
```diff
@@ -80,24 +86,44 @@ public boolean canBeGearBox(GearInfo gear) {
...
     public GearInstance parseInstance(GearInfo gearInfo, ItemStack itemStack) {
         WynnItemParseResult result = WynnItemParser.parseItemStack(itemStack, gearInfo.getVariableStatsMap());
         if (result.tier() != gearInfo.tier()) {
             WynntilsMod.warn("Tier for " + gearInfo.name() + " is reported as " + result.tier());
         }
 
+        SetInstance setInstance = null;
+        if (result.tier() == GearTier.SET && gearInfo.setInfo().isPresent()) {
+            SetInfo setInfo = gearInfo.setInfo().get();
+            setInstance = new SetInstance(setInfo, getActiveItems(setInfo.name()), getWynncraftCount(setInfo.name()));
```
kristofbolyai: Overall, it feels like that this is still misplaced. I think it should be `WynnItemParser` parsing the "wynncraft count", **along with parsing the equipped items from the tooltip**. I know, it won't be "accurate" for the duplication case, but it shouldn't be! Here, we are parsing what Wynn tells us..., and eventually, it will be correct.

So, I think, you should just parse and include the SetInstance in `WynnItemParseResult`. That would also mean that `getActiveItems` and `getWynncraftCount` can basically go, and all parsing is done with regexes.

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * @return A Set of all equipped set names
+     */
+    public Set<String> getUniqueSetNames() {
+        Set<String> returnable = new HashSet<>();
+        for (ItemStack itemStack : Models.PlayerInventory.getEquippedItems()) {
+            Optional<WynnItem> wynnItem = Models.Item.getWynnItem(itemStack);
+            if (wynnItem.isPresent()
+                    && wynnItem.get() instanceof GearItem gear
```
kristofbolyai: You can use `ItemModel#asWynnItem` here

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,136 @@
...
+            TypeToken<Map<String, RawSetInfo>> type = new TypeToken<>() {};
+            Map<String, RawSetInfo> rawSets = Managers.Json.GSON.fromJson(reader, type.getType());
+            rawSets.forEach((setName, rawSetInfo) -> {
+                List<Map<StatType, Integer>> bonuses = rawSetInfo.bonuses.stream()
+                        .map(bonusPair -> {
+                            Map<StatType, Integer> bonusMap = new HashMap<>();
+                            bonusPair.forEach((statName, statValue) -> {
+                                StatType statType = Models.Stat.fromApiName(statName);
+                                if (statType == null) {
+                                    WynntilsMod.warn("Unknown stat type: " + statName);
```
kristofbolyai: Shouldn't you put a continue here? I don't think null should make it into the bonus map.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -170,17 +174,46 @@ public static WynnItemParseResult parseItemStack(
...
+                if (mark.contains("✖")) {
+                    allRequirementsMet = false;
+                }
+            }
+
             Matcher setBonusMatcher = normalizedCoded.getMatcher(SET_BONUS_PATTEN);
             if (setBonusMatcher.matches()) {
                 // Any stat lines that follow from now on belongs to the Set Bonus
                 // Maybe these could be collected separately, but for now, ignore them
+                // TODO: we can actually do this now
```
kristofbolyai: So, basically, I want you to do this, instead of what you do now. See https://github.com/Wynntils/Artemis/pull/2343/files#r1553784277

common/src/main/java/com/wynntils/models/gear/SetModel.java
```diff
@@ -0,0 +1,135 @@
...
+    public int getTrueCount(String setName) {
+        int trueCount = 0;
+
+        for (ItemStack itemStack : Models.PlayerInventory.getEquippedItems()) {
+            for (StyledText line : LoreUtils.getLore(itemStack)) {
+                Matcher setMatcher = SetModel.SET_PATTERN.matcher(line.getString());
+                if (setMatcher.matches() && setMatcher.group(1).equals(setName)) {
+                    trueCount++;
+                }
+            }
```
kristofbolyai: Even for true count, we can use SetInfo/SetInstance, since the set name itself is not bugged. (And doing matchers is always costly to an extent).

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -72,21 +76,29 @@ public final class WynnItemParser {
...
+
+    // Test in WynnItemParser_MISC_REQ_PATTERN
+    private static final Pattern MISC_REQ_PATTERN = Pattern.compile("^§(c✖|a✔)§7 (.+)$");
 
     private static final Pattern EFFECT_HEADER_PATTERN = Pattern.compile("^§(.)Effect:$");
 
     private static final Pattern POWDER_MARKERS = Pattern.compile("[^✹✦❋❉✤]");
 
-    public static final Pattern SET_BONUS_PATTEN = Pattern.compile("^§aSet Bonus:$");
+    private static final Pattern SET_ITEM_PATTERN = Pattern.compile("^§[a7]- §([28])(.+)");
```
kristofbolyai: Can you put this below the set bonus pattern so it is in the real order? Also, add a comment like "- Morph Emerald" / this pattern is used to check what items are equipped from the set.

common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -273,6 +358,10 @@ public static WynnItemParseResult parseItemStack(
             }
         }
 
+        System.out.println("Figured out item " + itemStack.getHoverName().getString() + " has " + activeItems);
+        System.out.println(
```
kristofbolyai: debug

kristofbolyai (commented):
Two comments left, then it's good to go

kristofbolyai (approved):
Great! Took a while, but it's ready now!

PR #2348, opened 2024-02-25
feat: Skill point loadouts smooth scrolling

ryanzhoudev:


ShadowCat117:
Minor nitpick, but can you make scrolling work with click and drag too?

ryanzhoudev:
Oh yeah that was planned but I completely forgot

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -509,6 +450,60 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
         return super.doMouseClicked(mouseX, mouseY, button);
     }
 
+    /*
+    This is rather complicated
```
magicus: Can you rewrite this to summarize the behavior rather than how you arrived at it?

common/src/main/java/com/wynntils/screens/skillpointloadouts/widgets/ScrollBar.java
```diff
@@ -0,0 +1,75 @@
...
+                getX(),
+                getY(),
+                0,
+                getWidth(),
+                getHeight());
+    }
+
+    /*
+    Here we go again...
+    Usually this is called when dragY reaches *about* 0.50.
```
magicus: Same here.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
ryanzhoudev: Yes, but I would like to keep the process to get to it somewhere. It will be useful for if/when smooth scrolling is implemented into the new book design. Any suggestions on where to put it instead?

magicus: I don't know really. Maybe we should have a wiki page for that?

magicus: But if others are ok with having this in source, I'll accept it.

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
```diff
@@ -126,6 +126,7 @@ public void deleteLoadout(String name) {
     }
 
     public void loadLoadout(String name) {
+        if (name == null) return; // no loadout selected
```
kristofbolyai: Is this even clickable, or rather, callable, if name is not null?

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -29,17 +30,22 @@
...
 import net.minecraft.client.resources.language.I18n;
 import net.minecraft.network.chat.Component;
+import org.lwjgl.glfw.GLFW;
 
 public final class SkillPointLoadoutsScreen extends WynntilsGridLayoutScreen {
+    private static final int MAX_LOADOUTS_PER_PAGE = 11;
     private List<LoadoutWidget> loadoutWidgets = new ArrayList<>();
 
     private boolean firstInit = true;
+    private final List<Pair<Supplier<String>, Function<Skill, Integer>>> summaryParts = new ArrayList<>();
```
kristofbolyai, line 48: Why supplier? Also if it ends up as a StyledText, why save it as StyledText?

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
kristofbolyai: Let's save this in Discord, if there is a need to check this again, but keep a shorter explanation of the math here.

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
ryanzhoudev: Oh yeah, I was exploring having loadouts be available by pressing enter, but this doesn't work so well with the name text box. Didn't remove this after.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
ryanzhoudev: Needs to be supplier since the skill point sums used in the I18n %s parts can change when the screen is still open. If they aren't suppliers they will get stuck on the first parse result

common/src/main/java/com/wynntils/models/character/SkillPointModel.java
ShadowCat117: I think clicking the selected loadout for a 2nd time to load it could be nice if you do want an alternate method of loading

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
kristofbolyai: And why not make a method for repopulating a list of cached strings when they change? I think that should be a bit easier logic to have overall.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -509,6 +450,47 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
         return super.doMouseClicked(mouseX, mouseY, button);
     }
 
+    // baseYPosition - dividedHeight * maxScrollOffset * scrollPercent = dividedHeight * 52
+    // Solve for maxScrollOffset
+    // Full explanation in #artemis-dev
+    @Override
+    public boolean mouseScrolled(double mouseX, double mouseY, double scrollX, double scrollY) {
+        // 11 loadouts are fully displayed from 9 to 56
+        // 12th one is very slightly cut off (it needs 57)
```
kristofbolyai, line 460: 56/57 what? pixels? or are we talking about "grid pixels"?

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
```diff
@@ -509,6 +450,47 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
...
+        return super.mouseScrolled(mouseX, mouseY, scrollX, scrollY);
+    }
+
+    @Override
+    public boolean keyPressed(int keyCode, int scanCode, int modifiers) {
+        if (keyCode == GLFW.GLFW_KEY_DELETE && deleteButton.active) {
+            deleteButton.onPress();
+            return true;
+        } else if (keyCode == GLFW.GLFW_KEY_END) {
+            mouseScrolled(0, 0, 0, Float.NEGATIVE_INFINITY);
```
kristofbolyai: I can't say I like doing this call. Maybe separate the scroll logic to a separate method from mouseScrolled, then call that here?

common/src/main/java/com/wynntils/screens/skillpointloadouts/widgets/ScrollBar.java
```diff
@@ -0,0 +1,56 @@
...
+                getHeight());
+    }
+
+    // dragY = dividedHeight * 8 + dividedHeight * 48 * (-scrollY / 50)
+    // Discard constant after solving, it comes from the height of the scrollable area
+    // Full explanation in #artemis-dev
+    @Override
+    protected void onDrag(double mouseX, double mouseY, double dragX, double dragY) {
+        super.onDrag(mouseX, mouseY, dragX, dragY);
+        parent.mouseScrolled(mouseX, mouseY, 0, (-25 * dragY) / (24 * dividedHeight));
```
kristofbolyai: Same goes for here. Use a separate method.

kristofbolyai: Also, shouldn't we call this before super?

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
ryanzhoudev: I don't think that is much better than what the suppliers do now. If we do the cache and repopulation, we need to get the model to post an event, have the feature subscribe, and call some public method in the screen to repopulate cache. But the supplier way is essentially the exact same as if you had manually coded all the summary lines into doRender(), and I don't think it's expensive to just get a number from a model.

common/src/main/java/com/wynntils/screens/skillpointloadouts/widgets/ScrollBar.java
```diff
@@ -0,0 +1,56 @@
...
+    // Discard constant after solving, it comes from the height of the scrollable area
+    // Full explanation in #artemis-dev
+    @Override
+    protected void onDrag(double mouseX, double mouseY, double dragX, double dragY) {
+        parent.doScroll((-25 * dragY) / (24 * dividedHeight));
+        super.onDrag(mouseX, mouseY, dragX, dragY);
+    }
+
+    @Override
+    protected void updateWidgetNarration(NarrationElementOutput narrationElementOutput) {}
```
magicus, line 55: I thought we had like a `WynntilsWidget` or something like that, that inherited `AbstractWidget`, implemented this function, and did some other nice stuff, and that we were supposed to base all widgets on it?

ryanzhoudev: Doesn't seem like it. We have WynntilsButton, but that doesn't really make sense to use here imo, and everything else is either AbstractWidget or TextInputBoxWidget

magicus: Ok, fine then.

magicus (approved):
This looks good as far as I can tell, but bolyai is the expert here.

common/src/main/java/com/wynntils/screens/skillpointloadouts/SkillPointLoadoutsScreen.java
kristofbolyai: Okay, let's do this (for now, at least).

common/src/main/java/com/wynntils/screens/skillpointloadouts/widgets/DeleteButton.java
```diff
@@ -22,5 +22,6 @@ public void onPress() {
         Models.SkillPoint.deleteLoadout(parent.selectedLoadout.key());
         parent.setSelectedLoadout(null);
         parent.populateLoadouts();
+        parent.mouseScrolled(0, 0, 0, 0); // force scrollPercent refresh
```
kristofbolyai:
```suggestion
        parent.doScroll(0); // force scrollPercent refresh
```

kristofbolyai (approved):
Works greatly. Although, I wonder, day by day, if we should build a general UI toolkit..

PR #2351, opened 2024-02-26
feat: Allow overriding urls.json source link and cache with Java properties

### LINKED ISSUE #2350: [Proposal] - Allow overriding urls.json source link with a java properties
### Problem Statement
In recent years, users from China and other countries with censored or filtered internet access have encountered difficulties in downloading online resources for both Artemis and Wynntils mods.

For Artemis specifically, managing networking and URL access has been a challenge. While we have made improvements such as utilizing a unified `urls.json` resource to store all resources as key-value pairs, this resource currently lacks easy override functionality, as it's not designed for modification.

### Previous Workarounds
To address this issue, the Chinese community has commonly resorted to two main workarounds:
1. Creating custom builds of Artemis.
2. Developing add-on mods that mix into Artemis to incorporate patched `urls.json` files, redirecting to CDNs accessible in China.

### Proposed Solution
Our goal has always been to support all communities interested in using Artemis. While the ideal solution would involve migrating our data to AWS, ensuring accessibility in China, this migration has been delayed due to resource constraints within the web team.

As an interim solution, we propose implementing the ability to override the `urls.json` file source link using java properties. This solution offers several benefits:
- **Ease of Implementation**: The proposed change is straightforward to implement.
- **User-Friendly**: End users will find it easy to utilize this feature.
- **Safety**: This approach is as secure as previous methods involving custom builds or add-on mods.
- **Minimal Disruption**: Regular users will experience minimal to no impact from this change.

kristofbolyai:
See https://github.com/Wynntils/Artemis/issues/2350 for details.

Implemented:
- `wynntils.url.list.override.link` property, which is an override for the original `urls.json` link. **Note: If you don't disable caching, and you already have a urls.json cache, the override link won't take priority on the first run.**
- `wynntils.url.list.ignore.cache` property, which disables reading `urls.json` from cache.

### DIFF
--- a/common/src/main/java/com/wynntils/core/net/UrlManager.java
+++ b/common/src/main/java/com/wynntils/core/net/UrlManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.core.net;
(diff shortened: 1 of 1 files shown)


TUsama:
Thank you so much for your efforts! This feature is incredibly helpful for Chinese players.
I wanted to share a suggestion. Currently there are various non-official Minecraft launchers in China, some launcher can hide their JVM arguments setting in the corner for some weird reasons, Moreover,  most players don't even know what JVM argument it is(at least for Chinese players). This could potentially make this solution less convenient in certain situations.  However, I genuinely appreciate the work you've put into this!

kristofbolyai:
@Tusama do you have a suggestion for replacing JVM arguments?

TUsama:
I think turning this into an in-game config would be a great idea since every player needs to configure their Wynntils. This could be more user-friendly.

kristofbolyai:
> I think turning this into an in-game config would be a great idea since every player needs to configure their Wynntils. This could be more user-friendly.

I specifically don't want this to be a setting that users mess with by accident. Java properties are hidden enough not to be changed without actually knowing what you are doing.

I also think that if a Minecraft launcher has hidden the JVM properties so much that it is hard to find, then the launcher should not be used. Changing JVM settings has become a necessity a long time ago for Minecraft.

kristofbolyai:
Also, @TUsama do you have the resources available to you to test this PR? For example, building it, providing a url override and then asking a Chinese user to test it.

TUsama:
I currently don't have the condition for testing, but I'm willing to assist in finding Chinese players for testing. However, it's not guaranteed that I'll be able to find someone.

TUsama:
I gave this build a test and it just perfectly worked😊. can't wait to see this PR merged.

magicus (approved):
This seems as a good solution to me. It will also, incidentally, be useful during development as an easy way to setup an alternative json. I'm guessing this is not by pure chance. :)

kristofbolyai:
> This seems as a good solution to me. It will also, incidentally, be useful during development as an easy way to setup an alternative json. I'm guessing this is not by pure chance. :)

Yes, I am sure it could be used during development. However the main focus was the Chinese users. 

(Actually, I am a bit against using this in development environments as this could lead to us forgetting to commit the update the urls.json for Artemis, and release a build without it.)

PR #2353, opened 2024-02-29
refactor: Turn SearchableContainerType into InteractiveContainerType [skip ci]

ShadowCat117:
Also add a few new searchable & scrollable containers.

Thought this would make it easier to add new scrollable/searchable containers in the future as all the information is in one place instead of split across the enum and ContainerModel as all searchable screens are scrollable, but not vice versa. Also since it has the knowledge of the previous button in the enum, perhaps someone could implement the backwards searching from legacy ;)


magicus:
Give me a few days to look at this before integrating, please. (I think it looks good but I need to check more carefully)

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -201,58 +215,80 @@ private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
...
-                currentSearchableContainerType.getNextItemSlot(),
+                currentInteractiveContainerType.getNextItemSlot(),
                 abstractContainerScreen.getMenu().containerId,
                 GLFW.GLFW_MOUSE_BUTTON_LEFT,
                 abstractContainerScreen.getMenu().getItems());
     }
 
-    private SearchableContainerType getCurrentSearchableContainerType(StyledText title) {
-        SearchableContainerType containerType = SearchableContainerType.getContainerType(title);
+    private InteractiveContainerType getCurrentInteractiveContainerType(StyledText title) {
```
kristofbolyai: I wonder if this whole method could be a return an exhaustive switch expression (so that it forces implementing new types here when adding new enums). Looking at it, I think it should be possible.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -68,15 +68,15 @@ public void onScreenInit(ScreenInitEvent e) {
...
 
         if (Models.Container.isAccountBankScreen(screen)) {
-            currentContainer = SearchableContainerType.ACCOUNT_BANK;
+            currentContainer = InteractiveContainerType.ACCOUNT_BANK;
         } else if (Models.Container.isBlockBankScreen(screen)) {
-            currentContainer = SearchableContainerType.BLOCK_BANK;
+            currentContainer = InteractiveContainerType.BLOCK_BANK;
         } else if (Models.Container.isBookshelfScreen(screen)) {
-            currentContainer = SearchableContainerType.BOOKSHELF;
+            currentContainer = InteractiveContainerType.BOOKSHELF;
```
kristofbolyai: Could the enum have a `Predicate<Screen>` so you can get the type easily? If you also add an `isBank` field, you can also get rid of the code duplication here.

kristofbolyai: I see that the enum already has a pattern for the screens. Is that necessary, seeing that most of them are seemingly also in `ContainerModel`? 

I think we have two options here:
1, Move all patterns to `ContainerModel`, add `Predicate<Screen>` for the enum
2, Remove all duplicated patterns from `ContainerModel`, and only keep them in `InteractiveContainerType`. Adding a `Predicate<Screen>` for enums is trivial in this case.

I would lean towards the second option, as I've never liked `ContainerModel` having so much patterns + is...Screen methods. I think we could create two enums to fit all purposes. `InteractiveScreenType` and `StaticScreenType` (or what fits best here) and store all related info that way.

However, let's consult with @magicus before doing this change (and feel free to do the other suggestions I've recommended).

kristofbolyai (commented):
Really great start for a cleanup! I have 2-3 recommendations that should clean up a whole lot of duplicated code once again.

common/src/main/java/com/wynntils/models/containers/BankModel.java
magicus: I would need to study the code to know exactly what I'm talking about (and I'm running out of time rn, sorry), but I think I also prefer 2 here, if I understand it correctly. I never liked all the patterns, it felt very much like an anti-pattern (no pun intended). Somehow we need to bundle everything we know about a container into a separate class, patterns to detect it, what buttons to press to scroll it, etc. I think that is what 2 says.

magicus: Quickly scrolling through the code... let me add another note: I don't think we should have an *enum* at all. We should have subclasses and/or instances that represent everything we need to know about a screen. (Regardless of if we can interact with it or not, just as long as wee need to know what kind of screen it is.) 

If this PR is the place to make such a fix, I don't know.

ShadowCat117: I've done the predicate for now, I've also moved a lot of logic out of ContainerModel, especially some that was more relevant to BankModel.

I think the idea of having subclasses/instances to contain all relevant information about any screen we do anything with is interesting, probably not something I'd do here but definitely worth looking into it at some point

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -53,11 +58,17 @@ public class BankModel extends Model {
     private final Storage<Map<String, Map<Integer, String>>> customCharacterBankPagesNames =
             new Storage<>(new TreeMap<>());
 
+    public static final int LAST_BANK_PAGE_SLOT = 8;
+
+    // Test in BankModel_PERSONAL_STORAGE_PATTERN
+    private static final Pattern PERSONAL_STORAGE_PATTERN =
+            Pattern.compile("^§0\\[Pg\\. (\\d+)\\] §8[a-zA-Z0-9_ ]+'s?§0 (.*)$");
```
kristofbolyai, line 65: How come this is not in the enum? It could also return a matcher to fix the problem?

common/src/main/java/com/wynntils/models/containers/BankModel.java
kristofbolyai: I think it's fine to have the enum in this PR, since it's already a major code improvement. We can separate it to instances later, when we have a solid use case for it.

common/src/main/java/com/wynntils/models/containers/BankModel.java
ShadowCat117: This one is just to find current page number for the various banks it's not its own container

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -72,6 +72,21 @@ public class ContainerSearchFeature extends Feature {
     @Persisted
     public final Config<Boolean> filterInPetMenu = new Config<>(true);
 
+    @Persisted
```
magicus, line 75: Is there no way we can make this more general? Most of the other code is fine to improve in steps, but adding a bunch of settings mean that any kind of future improvement to this will require an upfixer.

But then again our settings type system might be too limited to make this better. :(

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -201,61 +215,51 @@ private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
...
+            if (type.isScreen(screen)) {
+                containerType = type;
+            }
         }
 
-        return null;
+        if (containerType == null || !containerType.isSearchable()) return null;
+
+        return switch (containerType) {
+            case ACCOUNT_BANK -> filterInBank.get() ? InteractiveContainerType.ACCOUNT_BANK : null;
```
magicus, line 245: (With an improved config system, this mapping would not be necessary.)

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
magicus: One thing we could do would be to instead have a config like: filterInMenues = { ALL, NONE, SOME }, and if it is set to SOME, we can have a String menusToFilterIn = "content_book, bookshelf" or something like that, which we parse as being a space and/or comma-separated list of the enums.

I imagine most people will want to have it at ALL anyway. If they do, when a new menu type is added, it will be supported automatically. If you prefer to only have it in a few, or none, when a new menu type is added, it will be off by default if you have NONE or SOME.

A somewhat more complex solution would be to have ALL, NONE, INCLUDE, EXCLUDE and a menusFilterList, where INCLUDE would work as SOME above, but EXCLUDE would mean "all but those in the menusFilterList".

Even more ideally, we would add some GUI support for creating the menusFilterList. That, I think, would be the ideal way of doing this.

magicus: Hm. Basically, I think I'm saying that we should support Set<Enum> as a valid configuration type in the GUI. I don't think that would be too hard to do, and it will probably help us elsewhere.

kristofbolyai: We could also have a "list selection" config that could have a supplier for the allowed values.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -53,11 +58,17 @@ public class BankModel extends Model {
...
     private static final int MAX_CHARACTER_BANK_PAGES = 10;
+    private static final StyledText LAST_BANK_PAGE_STRING = StyledText.fromString(">§4>§c>§4>§c>");
 
     private boolean editingName;
     private int currentPage = 1;
-    private SearchableContainerType currentContainer;
+    private InteractiveContainerType currentContainer;
 
     public BankModel() {
```
magicus, line 73: Maybe this should be renamed `StorageModel` now that it supports so much more than the original bank? (And the `type.isBank()` etc as well.)

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
magicus: I might be nitpicking, but I think what we need in general is to support a "set" (unordered) rather than a "list" (ordered"). The latter is much harder to implement GUI-wise, and I don't think that is something we would need in general. 

But yes, it seems like a good approach, for a Set (or, perhaps, some other collection type), we need to have a supplier for allowed values. For enum this can be created automatically, but we could also supply a generic supplier to the config value.

magicus:
I think I've sprinkled mostly random comments in this PR. (The "drive-by" style...) I'm not sure if any of these have been helpful.

I've now gone through the entire PR in a more consistent matter. I think it looks good in general. There is more to be done, for sure, but this is a good start. The only thing I'm slightly more hesitant to just say okay to is all the new settings, but then again -- we don't have a Set framework for configs, that will require new major functionality, and it does not seem reasonable to withhold this PR awaiting that. And in the end, anything is solvable by upfixers (it's just no fun writing those).

As for the enum/classes split; I think this might do just fine. We have basically customized each enum entry so much that it is like a separate instance. This is perhaps stretching the enum idea almost too far, but then again, it works.

So I'm basically fine with this being integrated, as long as bolyai is too.

common/src/main/java/com/wynntils/models/containers/BankModel.java
kristofbolyai: We might confuse it with our own Storage types, but since it's a model, we could tell the difference. If you do the rename, just make sure to add a comment to clarify that this is not `StorageManager`.

common/src/main/java/com/wynntils/models/containers/BankModel.java
kristofbolyai: Okay.

kristofbolyai (approved):
Code looks good to me. If you do the rename, check if there are any storage items that needs to be fixed (we can't upfix storages..).

common/src/main/java/com/wynntils/models/containers/BankModel.java
magicus: Might be a bad idea. Keep it as "Bank" for now.

ShadowCat117:
I think I'll keep it as BankModel for now until we can upfix storage as custom page names are stored there and I don't want those to get reset for everyone

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
ShadowCat117: When I was redesigning the settings screen I was thinking about different config types and lists was one of them. It could have an button labelled "Edit" which would then replace all the listed config options with all the inputs for the list, say bank quick jumps for example, it would have "Account Bank Quick Jumps" with an edit button, clicking that would then show 6 input boxes to enter the 6 destinations, or the ranged slider if we were to add that back

kristofbolyai:
@JamieCallan117 Is this PR okay to merge, or are you waiting for something else?

ShadowCat117:
I think it's fine unless we do want to do something with the configs, like the ALL, SOME or NONE option

kristofbolyai:
> I think it's fine unless we do want to do something with the configs, like the ALL, SOME or NONE option

Let's merge it then. I don't want to fiddle with configs, until we do changes to them specifically.

Issue #2356, opened 2024-03-03
Make translation feature work with NPC dialogues


kristofbolyai:
### Note to future self: Make sure to test all behavior with Chat Tabs Feature disabled!

# The Issue

This issue is used to track the progress on making the mod handle translating NPC dialogues, **and most importantly, displaying the translated version, either on the overlay, or in chat.**

Most of the work needs to be done in `ChatHandler` itself, not `TranslationService`, most likely in the following order:
- [x] Fix some minor bugs in dialogue parsing (in `ChatHandler`)
- [x] Create some kind of `NpcDialogueModel`, and move the responsibility of tracking what dialogue is currently displayed there, from `NpcOverlay`. 
    - [x] The model needs to have knowledge of "NPC dialogue extraction dependents", signal it to `ChatHandler`.
    - [x] The model has the responsibility of distributing the dialogues between consumers (if the overlay is disabled, the dialogue should fall back to the chat).
    - [x] Add the ability to use NPC dialogue extraction with a chat-based display mode 
- [x] Make `TranslationFeature` work with `NpcDialogueModel` to easily distribute translated data
- [x] Optionally, explore integrating Wynnic/Gavellian transcription into the model
- [x] Fix the translation cache getting corrupted quite frequently

magicus:
A semi-related thought -- I wonder if there exists any new flashy LLM-based translation services that can be used as an alternative backend.

magicus:
And a second semi-related thought -- this would pave the way for allowing an easy (or at least easier) integration of Voices of Wynn into Wynntils.

kristofbolyai:
> A semi-related thought -- I wonder if there exists any new flashy LLM-based translation services that can be used as an alternative backend.

DeepL has been suggested before, I am thinking of implementing it.

kristofbolyai:
> And a second semi-related thought -- this would pave the way for allowing an easy (or at least easier) integration of Voices of Wynn into Wynntils.

Yes, we will now have knowledge of the current dialogues, so it's definitely a step.

magicus:
An additional bullet or two that I've thought about for a long time:

[...]

PR #2357, opened 2024-03-03
fix: Fix some bugs in dialogue parsing (in ChatHandler) [ci skip]

kristofbolyai:
First task of #2356.

### fix: Use ticks for measuring slow packet difference, fix tick scheduler manager having an inconsistent order with the tick events [50de318](https://github.com/Wynntils/Artemis/commit/50de31847c1413d18843ffc14aec931e47abd406) 

The logic was correct, but using `System.currentTimeMillis` is too "inaccurate" for tick timing in many cases. Since the packet difference is actually measured in ticks, we might as well use that here as well.
I've bumped the timeout ticks from 10 to 20, since the previous value was too low for some cases.

This also fixes the tick scheduler manager having an inconsistent order with the tick events, which also caused the same issue.
The new (and correct) order of operations is:
1. TickAlwaysEvent
2. Run the tick scheduler (highest TickAlwaysEvent priority)
3. TickEvent

This fixes the issue where tick events could schedule a task for the next tick, but the tick scheduler would run it at the same tick event, since the tick always event was run after the tick event.

### fix: Fix chat handler "missing" a tick when collecting dialogue [7ffea3c](https://github.com/Wynntils/Artemis/commit/7ffea3c2abc9e47ce2457a516e4201f3aaf71493) 

This bug resulted in the handler not collecting chat lines correctly rarely, while in dialogue, dumping a lot of duplicated background text.

### fix: Fix the "last" bad tick logic in ChatHandler [fc3d4b6](https://github.com/Wynntils/Artemis/pull/2357/commits/fc3d4b60359a08c83e0df9f16f49e9ef4f76fcd5)

This commit extends the commit before it, as there was a wrong assumption about tick/packet order that left me not to change the equality in the tick event.

### DIFF
--- a/common/src/main/java/com/wynntils/core/mod/TickSchedulerManager.java
+++ b/common/src/main/java/com/wynntils/core/mod/TickSchedulerManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.core.mod;
@@ -10,6 +10,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class TickSchedulerManager extends Manager {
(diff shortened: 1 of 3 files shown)


magicus (approved):
This looks good to me. Thanks for starting to work on the dialogue translation!

Is the reason that this is a draft PR that it needs more testing, or that it needs to go in together with the other PRs for the dialogue fix?

kristofbolyai:
> This looks good to me. Thanks for starting to work on the dialogue translation!
> 
> 
> 
> Is the reason that this is a draft PR that it needs more testing, or that it needs to go in together with the other PRs for the dialogue fix?

It's a draft because there is still a really small chance that we "miss" a tick. I think the wait time might need to be adjusted, as it worked almost always at home, but on mobile data, it happened quite frequently (which is good for reproduction).

kristofbolyai:
I've managed to fix the issue, it seems like that I had a wrong assumption about the packet/tick order. If the equality is allowed in the tick event, it collects packets for exactly two ticks, unlike what it did (3 ticks, so packets could slip in and cause the wrong code to run).

PR #2358, opened 2024-03-04
feat!: Make NpcDialogueFeature work in a chat-based mode, add legacy chat mode, fix many NPC dialogue related bugs, make the overlay work once again

kristofbolyai:
Second task of #2356.

## Main task
The main purpose of this PR is to separate dialogue logic into the newly created `NpcDialogueModel`, and add a way to display NPC dialogues in chat, if the overlay is disabled, with NPC dialogue extraction enabled. This will eventually allow us to nicely integrate translations into the dialogue model, but this PR also improves gameplay-experience in itself, since it removes 99% of chat spamming that happens during NPC dialogues (with NPC dialogue extraction turned off).

### refactor: Create NpcDialogueModel, move timing/currently displayed dialogue logic from NpcDialogueOverlay to the model, prepare NpcDialogueFeature for an optional chat-based displaying mode [aa53d78](https://github.com/Wynntils/Artemis/commit/aa53d7816fd8347ac66ae2a5a6cca74ea4056920)

This commit does the heavily lifting. I've created `NpcDialogueModel` to handle all timing/displaying logic, which was previously done in the overlay. I've also moved all auto-progress related code from the overlay to the feature itself, since it has to work with the upcoming chat-based dialogue approach as well.

<hr>

### fix: Fine-tune alignments in NpcDialogueOverlay [967b7ed](https://github.com/Wynntils/Artemis/commit/967b7edc864b89dc8b4e0002b7badc847fe5ef3e)
Nothing major, but the helper texts looked off when auto-progress was enabled. Should look better now.

<hr>

### feat: Add config upfixers [87fb981](https://github.com/Wynntils/Artemis/pull/2358/commits/87fb9816f1d29051e3cc41ae3997fbce2855d276)
Previous commits renamed the feature, as well as moved configs out of the overlay to the parent feature, we need to upfix it.

<hr>

### fix: Properly calculate render height for empty lines, align multiple dialogues correctly [5dee07c](https://github.com/Wynntils/Artemis/pull/2358/commits/5dee07c90dff91d5c50e004bc635da8925063fa8) 

Previously we've had both the `FontRenderer` and the overlay calculate render height incorrectly, or only have specific methods calculate the heights correctly. I've added a fix so all code paths work the expected way.

#### Old behavior
image
image

#### New, fixed behavior
image
image

<hr>

### feat: Display dialogues in order on NpcDialogueOverlay [3a00b17](https://github.com/Wynntils/Artemis/pull/2358/commits/3a00b1789a07dd0dcd5feab07892292443a4f0d9) 

Previously we've always displayed normal dialogues first, then all confirmationless dialogues. It did not make much sense, since newer dialogue could end up further up on the overlay.

#### Old behavior
image

#### New, fixed behavior
image

<hr>

### feat: Add the ability to display dialogues in chat Wynncraft-like [12f8fe0](https://github.com/Wynntils/Artemis/pull/2358/commits/12f8fe05f6893428699fb8db02d62a84ed9d9f6e) 

This commit opens up a huge amount of new features we could do. To start with, I've implemented what Wynn does, but with our own parsing, so spamming does not happen.

To achieve this:
- I've had to modify `ChatComponent` so we can remove old messages.
- I've also had to modify `NotificationManager` to add the ability to remove old messages (and fix an old bug with chat-only notifications).

The result speaks for itself.

#### Showcase of the new systems
**Selections, with overlay enabled:**
https://github.com/Wynntils/Artemis/assets/49001742/b095cd56-ac58-483f-9efa-43747df0c4b5

**Chat-only behavior, with auto-progress:**
https://github.com/Wynntils/Artemis/assets/49001742/e28d4e34-9a11-4c47-8808-9bb823be9ed8

<hr>

### feat: Add a setting for the legacy Wynncraft dialogue system [96105db](https://github.com/Wynntils/Artemis/pull/2358/commits/96105dbbed33979c5c8d91b63567a7755d2096e9)

With all the improvements from the previous commits, we can simply add back the old system, as a config option :).

#### Demo
[Kooha-2024-03-04-20-34-26.webm](https://github.com/Wynntils/Artemis/assets/49001742/f8bf4abd-ccd7-4814-b6a6-08cc19f76db6)

<hr>

### feat: Display dialogues in normal mode as a single component list [7c3bf5f](https://github.com/Wynntils/Artemis/pull/2358/commits/7c3bf5f78392d03ce0c522e91a49ce8ea9139f5f)

This allows us to always update the currently displayed dialogue, whenever they change. This allows us to clear confirmationless dialogues when they time out, or update them when they become normal dialogues. This will also be useful when adding translation/transcription support.

<hr>

### fix: Also fix legacy mode confirmationless -> normal dialogue behaviour [39a22f5](https://github.com/Wynntils/Artemis/pull/2358/commits/39a22f56e9c238dbb4a26e379b6503fe1adf1e68)

Same problem as the last commit, we needed to fix confirmationless -> normal dialogue transitions.

<hr>

### fix: Separate queued/immediate commands, execute queued commands when we are safely outside of dialogues [25bd0cc](https://github.com/Wynntils/Artemis/pull/2358/commits/25bd0ccec4906af094a21e2feba87a0f6bd304b5) 

Previously we executed commands like /friend list in dialogues, which sent us a background message, so we couldn't parse it, and it appeared as spam to the user. I've separated "queued"/immediate commands and made sure queued commands are executed whenever we are outside a dialogue. Immediate commands, like chat commands, aliases, /class, /hub are still ran immediately (which should have been the case previously).

<hr>

### refactor: Rename CommandHandler methods to make their intent clear
[a17fab3](https://github.com/Wynntils/Artemis/pull/2358/commits/a17fab3b794ccb7729f2eaf5d83daa44bffc468c)

<hr>

### fix: fix small logic mistake in CommandHandler [19e9fd2](https://github.com/Wynntils/Artemis/pull/2358/commits/19e9fd2fe399972c99d35880bf4718727fb79b96)

<hr>

### fix: Detect when non-standard confirmationless dialogues happen, so we don't spam the chat [0a39a1d](https://github.com/Wynntils/Artemis/pull/2358/commits/0a39a1d47f6f7b816e02333dd6219b2bdee9096a) 

This is not 100% job done, since we still can't parse non-standard confirmationless dialogues, but at least we can track them, and backtrack to not display any spam.

<hr>

### fix: Don't run on tick when the overlay is enabled [d384457](https://github.com/Wynntils/Artemis/pull/2358/commits/d384457520cf1f2e5b50e1e3a6e7d124d070c08a)


common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -451,6 +440,6 @@ private RecipientType getRecipientType(StyledText codedMessage, MessageType mess
     }
 
     private boolean shouldSeparateNPC() {
```
magicus, line 460: You might consider inlining this at this point...

magicus:
I think this looks good. Just one question; I could have found the answer to this myself but it is tricky to browse files in a PR and I'm running a bit short on time.

Is there more logic in ChatHandler that should be moved to the NpcDialogueModel, now that it exists? I mean, ideally, the ChatHandler should be a "handler" that just distributes the work to models, like how the ItemHandler work. This has not been the case since the NPC dialogue stuff has been like the raison d'être for the ChatHandler, and it has had nowhere else to offload this work to.

kristofbolyai:
> I think this looks good. Just one question; I could have found the answer to this myself but it is tricky to browse files in a PR and I'm running a bit short on time.
> 
> 
> 
> Is there more logic in ChatHandler that should be moved to the NpcDialogueModel, now that it exists? I mean, ideally, the ChatHandler should be a "handler" that just distributes the work to models, like how the ItemHandler work. This has not been the case since the NPC dialogue stuff has been like the raison d'être for the ChatHandler, and it has had nowhere else to offload this work to.

I think there could be a fair part to be offloaded to the model, but it is not currently a goal at the moment.

magicus:
I might have a look at that as a follow up also.

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
kristofbolyai: I think it can stay until we refactor NPC logic out of there.

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
```diff
@@ -397,18 +394,17 @@ public float calculateRenderHeight(List<TextRenderTask> toRender) {
...
 
     public float calculateRenderHeight(List<StyledText> lines, float maxWidth) {
-        int sum = 0;
-        for (StyledText line : lines) {
-            sum += font.wordWrapHeight(line.getString(), (int) maxWidth);
-        }
-        return sum;
+        return lines.stream()
+                .mapToInt(line -> (int) calculateRenderHeight(line, maxWidth))
+                .sum();
```
kristofbolyai: This int cast was stupid from me, fixed in https://github.com/Wynntils/Artemis/pull/2358/commits/2316d197cbbedbe2fef72c10a8371fe139a07b63.

magicus:
Lots of improvements, I see! Nice! I'll look through the code as soon as I can.

Just a random thought: I see you have been working with the selection style chats. One thing I always wanted to do was to implement these as Screens, so we could get the mouse cursor active and actually *click* on the answer we wanted to give. (This could even include changing these answers into buttons).

That was the original reason I implemented the "protection" detection -- if you are in an unprotected situation, you cannot pop up a Screen, since you would need to be able to do stuff like run away or defend yourself. (Also, the logic came out quite wrong on this -- instead of showing "Protected" on almost all dialogs, we should have like a "⚠️ Beware of enemies!" on the few non-protected dialogs.

magicus:
Not saying you need to fix any of that in this PR, just sharing some of my thinking.

kristofbolyai:
Protection detection is still only 80-90% consistent, so it needs to get fixed properly before that. It shouldn't be a lot of work though, so maybe I can include that here.

common/src/main/java/com/wynntils/mc/extension/ChatComponentExtension.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.extension;
+
+import net.minecraft.network.chat.Component;
+
+public interface ChatComponentExtension {
+    void deleteMessage(Component component);
```
magicus, line 10: Ohhh... I wonder if this framework can help us restore BACKGROUND chat messages (grayscale versions sent during a chat) with FOREGROUND versions (normally colored that are "restored" when the chat is ended). That is currently actually a regression compared to vanilla Wynncraft, that we do not restore colors afterwards.

Just thinking out loud.

kristofbolyai: We could, rather easily.. I was just not sure if I should take that step here, or check if these changes are stable by themselves.

I think I won't do this, until I fix up client messages with chat tabs, since currently we don't really have great support for chat tabs even in our own code. (client side messages are sent only to the currently focused tab)

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
```diff
@@ -326,10 +326,7 @@ public void renderTexts(PoseStack poseStack, float x, float y, List<TextRenderTa
         float currentY = y;
         for (TextRenderTask line : lines) {
             renderText(poseStack, x, currentY, line);
-            // If we ask Mojang code the line height of an empty line we get 0 back so replace with space
```
magicus, line 329: Are you sure line is never empty? Or is the bug fixed in Vanilla?

magicus: Edit, ah, I see, the check has moved.

kristofbolyai: It's been integrated below.

magicus (approved):
I think this looks good.

common/src/main/java/com/wynntils/mc/extension/ChatComponentExtension.java
magicus: Yeah, not suggesting you do it here. Keep it as a separate change.

PR #2362, opened 2024-03-09
fix: Make translations work with NPC dialogues

kristofbolyai:
Third part of #2356.

## Showcase
Transcribed, then translated dialogue
image

Normal translation
image

Display original
image


common/src/main/java/com/wynntils/models/npcdialogue/event/NpcDialogueProcessingEvent.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Event that is fired before processing an NPC dialogue.
+     * This event can be used to add any pre-processing to the dialogue.
+     */
+    public static class Pre extends NpcDialogueProcessingEvent {
+        private final NpcDialogue dialogue;
+
+        // This is a future that can be used to add any post-processing to the dialogue.
+        // This is used for things like translation, transcription, etc.
+        private CompletableFuture<List<StyledText>> postProcessedDialogueFuture;
```
magicus, line 28: I assume this is since translation takes time. Nevertheless, we have not really used async code in Artemis, and I'm a bit scared about what that means. I guess this is just exposing something that already existed in the translation service, but nevertheless, it looks like this is creating more exposure for it.

common/src/main/java/com/wynntils/services/translation/TranslationService.java
```diff
@@ -27,8 +31,13 @@ public void reloadData() {
         CachingTranslationProvider.loadTranslationCache();
     }
 
-    public void shutdown() {
-        CachingTranslationProvider.saveTranslationCache();
+    @SubscribeEvent
+    public void onWorldStateChange(WorldStateEvent event) {
+        if (event.getNewState() == WorldState.WORLD)
+            ;
```
magicus: Huh?

common/src/main/java/com/wynntils/services/translation/TranslationService.java
```diff
@@ -68,9 +77,19 @@ public void resetTranslator() {
...
+    }
+
     public enum TranslationServices {
-        GOOGLEAPI(GoogleApiTranslationProvider.class),
-        PIGLATIN(PigLatinTranslationProvider.class);
+        GOOGLEAPI(GoogleApiTranslationProvider.class);
+
+        // This is a demo service, not used in production.
+        // Users accidentally enabling this service could cause confusion and frustration.
+        // PIGLATIN(PigLatinTranslationProvider.class);
```
magicus, line 98: Unfortunately this will make IntelliJ complain that`PigLatinTranslationProvider` is unused, but maybe we can live with that.

common/src/main/java/com/wynntils/features/utilities/TranslationFeature.java
```diff
@@ -120,12 +120,17 @@ public void onNpcDialogue(NpcDialogueProcessingEvent.Pre event) {
     }
 
     private StyledText unwrapCoding(String origCoded) {
-        // Some translated text (e.g. from pt_br) contains Á. This will be stripped later on,
-        // so convert it to A (not ideal but better than nothing).
-        // FIXME: Check if Á should be À
+        // Some translated text (e.g. from pt_br) contains special characters.
+        // These will need to be stripped or converted, which is not ideal but better than nothing.
```
magicus, line 129: Just thinking out loud, is there some way we can preserve these, by encoding them differently? E.g. replace it with `§&Aacute;` and then search for that string and replace it back with `Á` after all stripping is done?

Or make the stripping before the translation, so we don't need to strip after translating, when  `Á` might be a valid character.

Or, if we cannot strip before translation, do it the opposite way, as soon as we can, replace  `Á` with `§_` or something like that, and then strip these markers?

common/src/main/java/com/wynntils/models/npcdialogue/NpcDialogueModel.java
```diff
@@ -31,21 +30,16 @@ public class NpcDialogueModel extends Model {
     private final Set<Feature> dialogExtractionDependents = new HashSet<>();
 
     private NpcDialogue currentDialogue = NpcDialogue.EMPTY;
-    private final List<ConfirmationlessDialogue> confirmationlessDialogues = new ArrayList<>();
+    private ConcurrentLinkedQueue<NpcDialogue> confirmationlessDialogues = new ConcurrentLinkedQueue<>();
```
magicus, line 37: Why ConcurrentLinkedQueue? In what way can this be concurrently modified?

I'm not questioning this change per se, I am just curious how this happens. (And maybe asking that the access pattern should be documented so we understand any race situations.)

magicus (approved):
Overall I think this looks good. Thank you for cleaning this entire area up!

I've left a bunch of comments. I think most are actually questions, rather than requests for changes. But as usual, questions in reviews usually indicate that some comment is missing, and they are better replied to as comments in the code rather than as replies in the PR. :)

common/src/main/java/com/wynntils/features/utilities/TranslationFeature.java
kristofbolyai: These characters cannot be put in chat either way, so we need to strip after translation, since they are changed in the Wynn resource pack, and render weirdly. I also don't think we can use Unicode codes for these characters, as essentially, they will render the same font.

magicus: Ah, I see. If we ever are going to fix that, we'd need to combine it with a patch to the Wynn resource pack. I'll leave it to the Portuguese speaking users to file a bug if they deem it important to fix.

common/src/main/java/com/wynntils/models/npcdialogue/event/NpcDialogueProcessingEvent.java
kristofbolyai: Yes, making this async is required for translations. I've tried to design this future so that it is as unexposed as it can be, but still usable and extendable in the future.

common/src/main/java/com/wynntils/services/translation/TranslationService.java
kristofbolyai: It was meant to be a return. Saving on any world state change, except world, as saving when joining both can cause slight lag. Also, it is theoretically impossible to have new, unsaved translations when joining a new world, as we've saved it before..

common/src/main/java/com/wynntils/services/translation/TranslationService.java
kristofbolyai: I did not want to remove the code, but unfortunately, we can't have conditional enums.

common/src/main/java/com/wynntils/models/npcdialogue/event/NpcDialogueProcessingEvent.java
magicus: I understand. We'll just have to take care when working with this.

common/src/main/java/com/wynntils/models/npcdialogue/NpcDialogueModel.java
kristofbolyai: If we translate a confirmationless dialogue, then we need to change the list. However, translation is async.

I will make sure to comment this.

PR #2364, opened 2024-03-10
feat: Custom Seaskipper screen redesign

ShadowCat117:
This was mainly about making the departure board support scrolling so if future destinations are added it will work but I decided to merge the two screens as well since they didn't really need to be separate.

- Map side functions exactly like the original, just slightly smaller.
- Pois are coloured red when you cannot afford to travel there, or grey if they are inaccessible.
- Departure board supports scrolling.
- Departure board sorted alphabetically so you have an idea of where the destination you want will be listed.
- Made some more text translatable

https://github.com/Wynntils/Artemis/assets/8337467/a0113960-8c1b-4a39-8582-fcddf03ecf2d


kristofbolyai (approved):
Looks good to me. For some reason clicking on POIs/the travel button did not work for the first time, but I can't reproduce it again. I've tried resizing while having a selected location, still no luck.

Maybe make the currently selected POI green?

ShadowCat117:
> For some reason clicking on POIs/the travel button did not work for the first time, but I can't reproduce it again. I've tried resizing while having a selected location, still no luck.

Hmm ok I'll see if I can reproduce that when back home. 

> Maybe make the currently selected POI green?

I'll do that too

ShadowCat117:
I wasn't able to reproduce it. Something that did happen once was I clicked galleons graveyard and it thought I had selected nemract, but the `isSelected` method seems fine, so I might take a closer look tomorrow

kristofbolyai:
Okay, let me know when you want to merge this then.

ShadowCat117:
Still couldn't figure out the issue you had so I'm happy to merge this now and if some reports of it come in maybe they can give an indication on the cause

kristofbolyai:
I'll merge this with a fix for some chat issues later (or separately if I can't fix that other issue today).

magicus (approved):
Looks very nice!

PR #2365, opened 2024-03-11
feat: Add support for events in string-parsed StyledTexts [skip ci]

kristofbolyai:
This should allow us to reconstruct fully working clickable/hoverable components after translation.

Also reworked how events are collected. Previously, it was only collected, when we ran `getString(INCLUDE_EVENTS)`.

I can't justify doing that, it was really stupid. In fact, I am glad I had to be the one to track it down. To be clear, this caused the events not be collected, unless a text was printed as text, which the `toString` method gladly does. This left me wondering why debug mode works, and production-run does not. Let me tell you the obvious: the debugger runs `toString`...

## EDIT:
I've realised that our support for events is pretty low, as we have never used them in production code before. I've added logic to make events inherit like other styles elements, like bold, so they don't get added for every part, like they used to, but only get "repeated" on reset.

### DIFF
--- a/common/src/main/java/com/wynntils/core/text/PartStyle.java
+++ b/common/src/main/java/com/wynntils/core/text/PartStyle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
+ * Copyright © Wynntils 2023-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.core.text;
@@ -160,7 +160,7 @@ public final class PartStyle {
                 styleString
                         .append(STYLE_PREFIX)
                         .append("[")
-                        .append(owner.getParent().addClickEvent(clickEvent))
+                        .append(owner.getParent().getClickEventIndex(clickEvent))
                         .append("]");
             }
 
@@ -169,7 +169,7 @@ public final class PartStyle {
                 styleString
                         .append(STYLE_PREFIX)
                         .append("<")
-                        .append(owner.getParent().addHoverEvent(hoverEvent))
+                        .append(owner.getParent().getHoverEventIndex(hoverEvent))
                         .append(">");
             }
         }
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/core/text/PartStyle.java
```diff
@@ -111,7 +111,7 @@ public String asString(PartStyle previousStyle, StyleType type) {
         // If the current color is NONE, we NEED to try to construct a difference,
         // since there will be no color formatting resetting the formatting afterwards.
         if (previousStyle != null && (color == CustomColor.NONE || previousStyle.color.equals(color))) {
-            String differenceString = this.tryConstructDifference(previousStyle);
+            String differenceString = this.tryConstructDifference(previousStyle, type == StyleType.INCLUDE_EVENTS);
```
magicus, line 114: Will type ever be anything different?

magicus (approved):
I think this looks good. I've forgotten too much details about the StyledText implementation to be 100% sure, but it seems reasonable.

common/src/main/java/com/wynntils/core/text/PartStyle.java
kristofbolyai: Yes, normal. This type is only going to by used by translations atm.

magicus: So "normal" actually means "I don't care about events, so you can remove them to optimize performance", and `INCLUDE_EVENTS` means: "do this properly and correct"?

kristofbolyai: Yes. Normal is also the universal Minecraft format, while include events is our creation. Hence why it was never used before.

kristofbolyai: Although removing events is not fully correct, we do remember them and add it to components, just not in their string representations.

magicus: So maybe we should rename these options to `FULL` and `STRIP_EVENTS` instead, to better capture their behavior?

kristofbolyai: I mean, it's been like this ever since we had StyledText, and NORMAL means normal Minecraft behavior, so I don't think we need to.

magicus: Hm, okay.

PR #2366, opened 2024-03-11
fix: Refactor multi-line message handling in ChatHandler, fix multi-line messages not appearing twice, even if they should

kristofbolyai:
This change is much-much "larger" than the reader might think at first look.

Using StyledText, we can split the chat-screen message dump properly, without splitting at all new-lines. This makes it so components that use newlines regularly are kept as they were. This enables us to do a much better job at processing dialogues. In fact, this was the last key step needed before we could have replaced the background components with foreground ones. This is now possible easily, since we have 1-1 mapping for background-foreground components, so theoretically, "only" the replacing code needs to be written. I've deliberately not done that here, as I want to roll out these updates in small batches, so we can backtrack easily, if there is ever a need to do so.


kristofbolyai:
This seems to work from testing, also the concept looks good, so I'll merge this.

PR #2367, opened 2024-03-11
fix: Wait even more in CommandHandler when trying to execute a command during a dialogue [skip ci]

kristofbolyai:
It is reasonable to wait more, since we wouldn't be able to parse the response anyway.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/command/CommandHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/command/CommandHandler.java
@@ -14,9 +14,11 @@ import java.util.Queue;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class CommandHandler extends Handler {
+    private static final int TICKS_PER_EXECUTE = 7;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/handlers/command/CommandHandler.java
```diff
@@ -14,9 +14,11 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class CommandHandler extends Handler {
+    private static final int TICKS_PER_EXECUTE = 7;
+    private static final int NPC_DIALOGUE_WAIT_TICKS = 40;
```
magicus, line 18: This feels a bit arbitrary. Is it to wait for the player to finish the dialogue interaction, or for wynn to send all parts of the current dialogue screen? I'm not actually sure I understand how this is a fix, or what it is supposed to fix.

kristofbolyai: It's supposed to wait for us to fully finish a dialogue.

PR #2368, opened 2024-03-12
feat: Add StatusEffect function [skip ci]

joaovictorbarrera:
I decided to add a function to retrieve the information from status effect overlay so I can make info boxes that tell me if a certain status is active.

For example, if Radiance is off and needs to be recast.

This can be incredibly helpful for the warring community, as radiance cycling is difficult for beginners.


joaovictorbarrera:
mock up info box:

{if_str(contains_str(status_effects(); "Radiant"); "&aRadiance Active&r"; if_str(contains_str(status_effects(); "Radiance"); "&bSECOND GUARD RADIANCE&r";"&cCAST RADIANCE&r"))}

kristofbolyai (commented):
Hello, nice first try, but I can see you didn't run development environment, as you are missing the translation keys for this function. I assume you just build the mod and tested it.

Either way, please add them (en_us.json).

kristofbolyai:
Also since this function will always be used by contains, might as well add a function with a string argument that checks if you have the specified effect on. If you are up to it.

magicus:
> might as well add a function with a string argument that checks if you have the specified effect on. If you are up to it.

Either that, or even better **just** have such a function. I'm not sure how useful it is to have a function returning a list of all effects. The only usage for that I can see is the one example you provided, which is better expressed as a check on a specified effect.

joaovictorbarrera:
> Also since this function will always be used by contains, might as well add a function with a string argument that checks if you have the specified effect on. If you are up to it.

There are many more possibilities which would, for example, regex the time of an effect and other custom displays that could make use of the full list. I can maybe add a shorthand contains_status_effect(string) but not substitute it.

magicus:
Did you intend to close this? I am sorry if we scared you away, that was not the intention! I believe the essence of what you are trying to do here is sound and good. It just needs a bit more of tweaking. We'll help you get this ready for integration.

joaovictorbarrera:
> her



> Did you intend to close this? I am sorry if we scared you away, that was not the intention! I believe the essence of what you are trying to do here is sound and good. It just needs a bit more of tweaking. We'll help you get this ready for integration.

I will fix and reopen it !

joaovictorbarrera:
> Hello, nice first try, but I can see you didn't run development environment, as you are missing the translation keys for this function. I assume you just build the mod and tested it.
> 
> Either way, please add them (en_us.json).

Yea Im not quite sure what the dev env is in this project, I just compiled it and tested it xd

joaovictorbarrera:
I can get the translations for traditional and simplified Chinese in as well, would that be appreciated? I am quite close to a Chinese person that plays wynn.

kristofbolyai:
> I can get the translations for traditional and simplified Chinese in as well, would that be appreciated? I am quite close to a Chinese person that plays wynn.

We can merge it, but it might cause issues on our end.., so I can't guarantee that we can keep it. Crowdin doesn't like to sync with Github properly...

joaovictorbarrera:
Ok it should be done now, let me know if there's more I need to fix

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -87,6 +89,45 @@ public Double getValue(FunctionArguments arguments) {
         }
     }
 
+    public static class StatusEffectsFunction extends Function<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            List<StatusEffect> statusEffectsList = Models.StatusEffect.getStatusEffects();
```
ShadowCat117: You can use streams on both of these so you'll only need 1 list. For the contains function filter it to only include the query and return if it's empty or not

joaovictorbarrera: Got it. Give me a bit

joaovictorbarrera: Ok refactored

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
```diff
@@ -438,6 +438,8 @@ private void registerAllFunctions() {
         registerFunction(new CharacterFunctions.SoulpointTimerMFunction());
         registerFunction(new CharacterFunctions.SoulpointTimerSFunction());
         registerFunction(new CharacterFunctions.SprintFunction());
+        registerFunction(new CharacterFunctions.StatusEffectsFunction());
+        registerFunction(new CharacterFunctions.ContainsStatusEffectFunction());
```
kristofbolyai: Please maintain alphabetic order here.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -87,6 +87,39 @@ public Double getValue(FunctionArguments arguments) {
...
+            return String.join("\n", statusEffectsList);
+        }
+    }
+
+    public static class ContainsStatusEffectFunction extends Function<Boolean> {
+        @Override
+        public Boolean getValue(FunctionArguments arguments) {
+            String query = arguments.getArgument("query").getStringValue();
+            return !Models.StatusEffect.getStatusEffects().stream()
+                    .filter(statusEffect -> statusEffect.asString().getString().contains(query))
```
kristofbolyai: Isn't it more logical to match the whole string? Since we are talking about a list of effects, we want to do contains on the list, not the string.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -87,6 +87,39 @@ public Double getValue(FunctionArguments arguments) {
...
+        public String getValue(FunctionArguments arguments) {
+            List<String> statusEffectsList = Models.StatusEffect.getStatusEffects().stream()
+                    .map(statusEffect -> statusEffect.asString().getString())
+                    .toList();
+
+            return String.join("\n", statusEffectsList);
+        }
+    }
+
+    public static class ContainsStatusEffectFunction extends Function<Boolean> {
```
kristofbolyai: Perhaps rename this to "StatusEffectActive"? Don't forget to rename the localization keys.

kristofbolyai (commented):
Should be good after these changes. :)

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
joaovictorbarrera: The issue is each effect contains an icon, cooldown, many parts that are irrelevant and hard to keep track of when you are writing info boxes, it's a lot easier to just match part of the string for the purposes of using it

kristofbolyai: And there is no method for getting only the name part as a string for status effect? Maybe there should be one then.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
joaovictorbarrera: Queued

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
joaovictorbarrera: Queued

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
joaovictorbarrera: The line looks like this: *ICON* Radiant (00:15)
After every second, it will change to lets say *ICON* Radiant (00:14)

I need the function to simply match if the word Radiant is present, it would be a lot harder if it was trying to match the entire effect name rather than just part of it

It sure is a bit weird since status_effect_active("a") would most likely return True if any effect had the letter "a" in them, but... how complicated do we want to make this

joaovictorbarrera: > And there is no method for getting only the name part as a string for status effect? Maybe there should be one then.

I could work on that

kristofbolyai: > > And there is no method for getting only the name part as a string for status effect? Maybe there should be one then.
> 
> 
> 
> I could work on that

Afaik and remember, the StatusEffect class holds everything parsed nicely, so it's only a getter that you need. If it is not parsed there, don't spend time trying to parse it correctly.

joaovictorbarrera: > > > And there is no method for getting only the name part as a string for status effect? Maybe there should be one then.
> > 
> > 
> > I could work on that
> 
> Afaik and remember, the StatusEffect class holds everything parsed nicely, so it's only a getter that you need. If it is not parsed there, don't spend time trying to parse it correctly.

Apparently, it is. Thank you. I'll do some testing as to what exactly it holds, I'll just use that if it's good.

joaovictorbarrera:
sec i need to fix localization keys

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
```diff
@@ -87,6 +87,40 @@ public Double getValue(FunctionArguments arguments) {
...
+
+    public static class StatusEffectActiveFunction extends Function<Boolean> {
+        @Override
+        public Boolean getValue(FunctionArguments arguments) {
+            String query = arguments.getArgument("query").getStringValue();
+            return !Models.StatusEffect.getStatusEffects().stream()
+                    .filter(statusEffect ->
+                            statusEffect.getName().getStringWithoutFormatting().equals(query))
+                    .toList()
+                    .isEmpty();
```
kristofbolyai: One final change :) You can use anyMatch here, so you don't need to do ! and .isEmpty. (also it's slightly faster).

kristofbolyai (approved):
After the last change, this looks good. 

Developers, please test this in development environment, before merging.

common/src/main/java/com/wynntils/functions/CharacterFunctions.java
joaovictorbarrera: queued

magicus (approved):
Looks good! Thank you for your contribution to Wynntils!

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1694,6 +1694,11 @@
   "function.wynntils.soulpointTimerS.name": "SP[s]",
   "function.wynntils.sprint.description": "Your remaining sprint stamina",
   "function.wynntils.sprint.name": "Sprint",
+  "function.wynntils.statusEffectActive.argument.query": "String to check for",
```
```suggestion
  "function.wynntils.statusEffectActive.argument.query": "Case-sensitive effect name string to check for",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1694,6 +1694,11 @@
...
+  "function.wynntils.statusEffectActive.description": "Checks if [string] is contained in the status effect list",
```
```suggestion
  "function.wynntils.statusEffectActive.description": "Checks if an effect is contained in the status effect list",
```

ryanzhoudev (approved):
Just thought the docs could be a bit more descriptive. Other than that, tested in game, works great. Thanks!

joaovictorbarrera:
Here is a link as to what I'm trying to achieve with these changes:

https://discord.com/channels/394189072635133952/424990456435310602/1216905589502902312
(message sent in #general of wynntils discord server)

PR #2369, opened 2024-03-15
fix: Staff join messages not redirecting [skip ci]

ryanzhoudev:
Probably merge just before scrolling

### DIFF
--- a/fabric/src/test/java/TestRegex.java
+++ b/fabric/src/test/java/TestRegex.java
@@ -216,6 +216,10 @@ public class TestRegex {
         p.shouldMatch("\uE023 §#44aa3300v8j§a has just logged in!");
         // champion nickname
         p.shouldMatch("\uE017 §#ffe60000§ocharlie268IsAWizard§6 has just logged in!");
+        // moderator nickname, new style
+        p.shouldMatch("\uE01F §6§o§<1>Navi§e has just logged in!");
+        // music, new style
+        p.shouldMatch("\uE020 §3Texilated§b has just logged in!");
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


ShadowCat117:
Looks good, not sure if we want to add more test cases for the other ranks, I have all(?) the characters used for each but not the colour code so we might not be able to make them 100% accurate

Admin ""
Artist ""
Builder ""
CMD ""
Dev ""
GM ""
Hybrid ""
Item ""
Media ""
Owner ""
QA ""
Web ""

ryanzhoudev:
They're in the enum as well, but I don't want to add "fake" test cases we created. The ones we have now are copied from the game

ShadowCat117:
Ah, I didn't notice the enum. Best to keep the test cases as real examples then

ShadowCat117 (approved):

fabric/src/test/java/TestRegex.java
```diff
@@ -216,6 +216,10 @@ public void ChatRedirectFeature_LoginRedirector_FOREGROUND_PATTERN() {
         p.shouldMatch("\uE023 §#44aa3300v8j§a has just logged in!");
         // champion nickname
         p.shouldMatch("\uE017 §#ffe60000§ocharlie268IsAWizard§6 has just logged in!");
+        // moderator nickname, new style
+        p.shouldMatch("\uE01F §6§o§<1>Navi§e has just logged in!");
```
kristofbolyai, line 220: Why are we using INCLUDE_EVENTS here? Does the method even use that for getting the string? (If so , why?)

magicus: wdym?

kristofbolyai: &<1> is an event (clickable or hoverable, I don't remember). I am asking why this method is matching clickable events, as we never do this, and only use regular MC-styles.

PR #2373, opened 2024-03-16
feat: Add exponential interpolation map zooming [skip ci]

kristofbolyai:
_I wonder which PR's perquisite is this change._ 

Touchpad demo:

https://github.com/Wynntils/Artemis/assets/49001742/2b889d30-5d50-42a1-9eba-a624ac9f6eb4


common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -49,7 +49,7 @@ public class MinimapOverlay extends Overlay {
     private static final int DEFAULT_SIZE = 130;
 
     @Persisted
-    public final Config<Float> scale = new Config<>(1f);
```
kristofbolyai, line 52: The upfixer is missing on purpose. Users will have to reset this value, as there is no 1 to 1 mapping between the new and old values (and rounding to new values is almost just as bad as a full reset).

kristofbolyai:
With this PR done, @magicus, how do you think we should account for zooming in MapData?

We have two options:
- Still go by the old zoom levels, which are percentage based, 0.3 means the map is rendered with a 30% scale.
- Go by the new step system, which allows defining levels on an unified integer base.

common/src/main/java/com/wynntils/utils/render/MapRenderer.java
```diff
@@ -33,10 +33,21 @@
 
 public final class MapRenderer {
     // Zoom is the scaling of the map. The bigger the zoom, the more detailed the map becomes.
+
+    // ZOOM_STEPS is the number of steps the zoom can take. (0 to ZOOM_STEPS - 1)
```
magicus: I'm not sure I'm following this. I see you introduced the term "zoom steps", I assume this is to separate it from the value that you multiply with when drawing the map. This is ... good, I guess. 

But I don't understand how 40, 0.1 and 5 relates. Or what MIN_ZOOM_LOG etc is supposed to do.

I would like to stipulate that we have a min zoom level of 1, and a max zoom level of 100. (Or the other way around, I'm not sure which is most natural.) Then the other values should be calculated based from this, to see how we should multiply the map, what zoom level to start at, etc.

magicus: Also, I don't think we should represent zoom as an int. Even if we go from 1 to 100, we should be able to have a zoom of e.g. 2.5. How much the zoom value changes when you scroll the mouse well is kind of independent from everything else. (In fact, maybe it should be adjustable depending on how "fast" zoom you want?)

In a perfect world, this might also make us support a touchpad-style pinch-to-zoom. That'd be cool!

kristofbolyai: I've just pushed the second iteration of this code.

I've made zoom steps represent map zoom as a percentage between 1-100 (0% zoom makes no sense).
I've also fine-tuned the default values for scrolling so that it feels like to old map.
I've also added float-steps, making this world perfect, as we happen to support touchpad-style pinch-to-zoom by concept. :)

(Check PR description for demo.)

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -368,16 +369,21 @@ protected void centerMapAroundPlayer() {
 
     protected void centerMap() {
         updateMapCenter(MAP_CENTER_X, MAP_CENTER_Z);
-        setZoom(0);
+        adjustZoomStep(0);
```
kristofbolyai: FIXME: This is not correct. There should be a setZoomStep method, and an adjustZoomStepByDelta method.

kristofbolyai:
@magicus Any chance you can find time to get this reviewed?

magicus:
Sorry for the slow input on this. 

I tested this branch, and discovered that the zoom of the map is dependent on the GUI scale. This means that on GUI scale 1, when max zoomed out, the entire continent is just like 25% of the visible space which is way too much zoomed out, and contrary, it is hard to zoom in enough to see all details.

And conversely, at GUI scale 5, when max zoomed out, you can hardly see the entire Province of Wynn, and when zooming in you get to absurd levels where you see like 20 blocks across. 

I think that needs to be fixed as part of this PR, so there is a clear understanding what zoom level 1 and zoom level 100 means, e.g. on zoom level 1 you should see like (almost) the entire continent, and on zoom level 100 you should see a quarter of Detlas.

magicus:
I'm also not happy with the naming. I think the new 1-100 scale should be what we use everywhere except when finally rendering, so it should just be called "zoom" (not anything with "steps"). To get a value we can send into the rendering engine, we can use something like "zoom factor". But right now, I think we are calculating this in multiple steps, where at some point we invert this using 1/zoom etc. I would like to see us send the "zoom" (1-100) value as far down as possible in the call stack, until we actually need to convert this into something that makes sense for the rendering.

edit: We could also call it "zoom level"; I realize that is how I think of it.

magicus:
On the positive side, however, is that the zooming feels very good. It has a natural feel, and is just perfectly tuned to be fast enough while still being precise enough.

I tried looking for how the GUI scale comes into play, but did not figure it out. In the worst case, we can't do anything about it (which would be a shame); if so we might have to set our expectations based on a supposed commonly used value (2-3?).

magicus:
I added functionality to show the current zoom level when pressing shift.

ShadowCat117:
> I added functionality to show the current zoom level when pressing shift.

Doesn't have to be done here, but I'd love to see a slider for zoom level in the future akin to this
bbzUc

magicus:
Agree. My hack of showing the zoom level as an overlay is kind of a step on the way. Presenting it as a slider instead of a number is clearer, and it will also allow direct manipulation. It's just much more work to implement a slider widget in Minecraft. If you feel up to it, please go ahead! You can create a new branch based on this, so you can use the new 1-100 zoom levels.

ShadowCat117:
Might save it for when the map screen gets redone. I've seen a few suggestions and have some ideas of my own for improvements but after the filter screen I'm taking a small break from ui work to do some of the fuy.gg features

kristofbolyai:
@magicus 

https://github.com/Wynntils/Artemis/pull/2373/commits/46f333e0e0a5a8b43fe7b4e57825f1aa0f994479 fixes the zoom level not being stable for different GUI scales.

Below is demonstrated how the map basically fits the same way for different scales. POIs and text are unaffected.

1 GUI Scale:


2 GUI Scale:


5 GUI Scale:

kristofbolyai:
@magicus 

https://github.com/Wynntils/Artemis/pull/2373/commits/dc8fbd0198e9e1ecf908ea73626a7e18d5865d99 renames zoom steps to zoom levels. While I understand your intent of only having `zoomLevel` in our code, it's practically impossible, as we have 58 usages for `zoomRenderScale` just in `AbstractMapScreen`.

magicus (approved):
I've made a bunch of changes. Now I am happy with this PR. Are you okay with it?

magicus:
> we happen to support touchpad-style pinch-to-zoom by concept. :)

Do you know what would be required to add this in reality too? I guess it needs some digging into GLFW...

magicus:
It seems to be stuck here https://github.com/glfw/glfw/issues/90 (since 2013), with a solution here: https://github.com/glfw/glfw/pull/2419 that has not been merged since 2023. :-(

kristofbolyai:
> > we happen to support touchpad-style pinch-to-zoom by concept. :)
> 
> 
> 
> Do you know what would be required to add this in reality too? I guess it needs some digging into GLFW...

I think we talk about a different thing. I mainly tested Mac's double finger scrolling. Not sure how to support real pinching (I forgot how pinching was supposed to work..)

magicus:
Ah, I see. Yes, the double-finger scroll works fine. And I guess we'll have to wait another decade or so for pinch-zooming to be implemented in GLFW :-).

magicus:
@kristofbolyai To be clear, I am awaiting a review from you as well, since I changed so much in the end.

kristofbolyai (commented):
Looks good to me.

PR #2375, opened 2024-03-16
refactor: Implement proposed fixes for MapData [skip ci]

kristofbolyai:
This PR expects that https://github.com/Wynntils/Artemis/pull/2373 is merged, and we agreed upon using zoom steps for our zooming value.

Quoting https://github.com/Wynntils/Artemis/pull/2158#issuecomment-1762887872:

### [feat: Make map features have a range of level, add foundation for custom (dynamic only) and json-parsed map visibility](https://github.com/Wynntils/Artemis/commit/401857af1fb1b6a9974451a97ac7e8ae7dc43bc4) 
- [x] Label should treat the int value of level better
- [x] We need to decide how to handle unknown/suitable-for-all on level in map attributes
- [x] Visibility for zoom levels is missing; either fix here or as a follow-up

### [feat: add base versioning for forward-compatibility in JsonProvider (and also add a missing local-file provider loading method)](https://github.com/Wynntils/Artemis/pull/2375/commits/bd351f056fb930c1d7eccf4da6108a6ed03db89d) 
- [x] JSON files should have a version field

I expect this PR to be controversial, at least partially. I am okay with that, in fact, I hope we can agree on something even better, in the end.


magicus:
In a first quick glance, this looks good. But I need to take a better look when I'm more sharp and concentrated.

My main worry is if we are going to be able to retrofit the existing "zoom levels" into the new format. Have you experimented anything with that? Things like town names that does not show when zoomed in very much, but neither when zoomed out very much. Some discrepancy wrt to the current situation is okay, but overall "feel" of zoom visibility must be able to be expressed.

kristofbolyai:
> In a first quick glance, this looks good. But I need to take a better look when I'm more sharp and concentrated.
> 
> 
> 
> My main worry is if we are going to be able to retrofit the existing "zoom levels" into the new format. Have you experimented anything with that? Things like town names that does not show when zoomed in very much, but neither when zoomed out very much. Some discrepancy wrt to the current situation is okay, but overall "feel" of zoom visibility must be able to be expressed.

It's really easy to port it, you can count the steps with a scroll keybind, so you don't even need to code/debug anything.

magicus:
Interesting approach. I'd still like to see both the old and the new on-screen at the same time to double-check that they fade in/out at about the same time.

magicus:
Also, my weekend was a bit hectic. I thought I would have time to review this and the other map PR, but I had not. I'll try to do it as soon as I can, but it might take a few days.

kristofbolyai:
> Also, my weekend was a bit hectic. I thought I would have time to review this and the other map PR, but I had not. I'll try to do it as soon as I can, but it might take a few days.

Start with the other PR first. This is only changes that have no effect yet, since we have no data supplying to MapData yet (so there is also nothing to test..)

magicus:
Let's get #2373 sorted out first, then I will come back and look at this. (Basically, it is just the zoom levels that still worry me.)

magicus:
Once https://github.com/Wynntils/Artemis/pull/2373 is merged, I'll have another look at this.

kristofbolyai:
> Let's get #2373 sorted out first, then I will come back and look at this. (Basically, it is just the zoom levels that still worry me.)

And @magicus, whenever you have time, here is the next step.

kristofbolyai:
@magicus Hi. I've sent this ping to you on both Github and Discord. Please respond to the first one you see, and ignore the second one.

I've been approached by some people from our community, working on the same project. Coincidentally, both would benefit from the new mapdata system. Also considering that the 1.2.0 release is now done, I'd like to finish up map data too.

The reason why I am pinging is you to ask about the development process for this feature. I have a 3 options in mind, so please pick whichever suits you the best.
1, You still want to manage the map data refactor. I still write the code. Only downside is that the review cycle will get really long.
2, I take over management of the refactor. I won't wait for explicit review from you, but I'll label every map-data PR or tag you, whichever you prefer, so you can do any post-review suggestions. Before the "final" release, I'll try to get your overall opinion on the change, if it's possible.
3, You'll have time to work on the feature in the upcoming time-period. If so, I'll take up another big project.

Some details about the "project" I've mentioned above:
_**This part was asked to not be public, and is only shared in the discord message.**_

common/src/main/java/com/wynntils/services/map/Label.java
```diff
@@ -6,19 +6,33 @@
...
 
 public class Label {
     private final String name;
     private final int x;
     private final int z;
     private final int layer;
-    private final String level;
 
-    public Label(String name, int x, int z, int layer, String level) {
+    // level is provided as a string for some data sources
```
magicus: Can we get around this and make sure we only use integer values? I think the source of this was the Wynncraft map label thing, but that is apparently very hand-crafted, and if they write bad data we can adjust it, I think. I know I was very keen in old times to make sure we match the official data exactly, but my opinion has changed in that case, and I think now it is better if we provide a curated solution.

common/src/main/java/com/wynntils/services/map/Label.java
```diff
@@ -6,19 +6,33 @@
...
+    private final String levelString;
+    private final transient RangedValue level;
```
magicus: How does a ranged value apply? If a place is "level 70", do you want to store that as a range 70-130? Or does the range signify an uncertainty in the lowest level, like "level 30-40", so a more exact value like "level 30" would be the range 30-30?

Otherwise, I think the level should be just an integer, telling the lowest recommended level. That is the usual case for this.

magicus: (We should actually be able to use this, and e.g. hide everything that is above your level, or mark it as dangerous with a ⚠️ symbol or whatever. That is really a missed opportunity.)

common/src/main/java/com/wynntils/services/mapdata/attributes/AbstractMapAttributes.java
```diff
@@ -27,8 +28,8 @@ public int getPriority() {
     }
 
     @Override
-    public int getLevel() {
-        return 0;
+    public RangedValue getLevelRange() {
```
magicus: I think we never fully agreed on how to handle this. Does this mean "skill level is not relevant for this marker", or "the minimum level is not known/provided" or "this is available to all players"? Should we separate these concerns?

common/src/main/java/com/wynntils/services/map/Label.java
kristofbolyai: I think there is a value in specifying a max level, where the area is no longer interesting to the player.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
```diff
@@ -20,8 +21,12 @@ public interface MapAttributes {
...
     int getPriority();
 
-    // 0 means no information is available, or suitable for all combat levels
-    int getLevel();
+    // the range of combat levels for which this feature is suitable
+    // If one of the bounds of the range is undefined, Integer.MIN_VALUE or Integer.MAX_VALUE will be used
+    // (eg. Corkus has a minimum level of 80, but no maximum level)
+    // NONE (0-0) or a range of (Integer.MIN_VALUE-Integer.MAX_VALUE) means suitable for all combat levels
+    // null means no information is available
+    RangedValue getLevelRange();
```
magicus: I see you have answered my question above here already. 

I am not sure I agree with this. First off, max-level. Is there some real-world examples where a map marker would have a max value? I mean, even a level 100 player might go to Ragni from time to time. 

Since players start out at level 1 (right?), we can be tougher here and say that if you mean "suitable for all levels", you should use 1.

That leaves us with two other cases:
a) min level not known (probably will be the likely case for most player created content, like someone sharing a set of their set of personal good but not well-known grinding locations)
b) min level does not apply, like for instance for the province labels ("Gavel", "Wynn", etc)

and one way to express this:
1) set the min level to 0

I think this is okay, that having a min level of 0 leaves some vagueness if the min level is not known, or not applicable.

magicus: I remember that we discussed this quite a lot, but I don't remember any arguments or what lead you to make this implementation.

common/src/main/java/com/wynntils/services/mapdata/attributes/impl/OffsetMapVisiblity.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.mapdata.attributes.impl;
+
+import com.wynntils.services.mapdata.attributes.type.MapVisibility;
+
+public class OffsetMapVisiblity implements MapVisibility {
```
magicus: Given our discussion in discord, I think this class should basically be replaced by something that takes in a min, max and fade values, and interpolates based on them. I'm not sure if we should remove the interface, I think it could be good to keep it since that leaves open for hard-coded (non-json-based) internal mapdata providers behaving differently. But we should in practice only use the min-max-fade based one.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: I went back and re-read the discussion. I think it boiled down to basically you asking for a way to separate level unknown from level-open-for-all or level-does-not-apply, and me saying that it is hard to to, and even if we separate it, it's hard to know how to use it.

I worry that making this complicated just leaves a way for people to make bad data. I would argue that we should just have these possible values:
1: min level of 1, that is, suitable for everyone
>1: only suitable for the recommended min level
0: level unknown or not applicable

and if we want to show only stuff that are below my level, we can have an option "include markers with unknown or n/a level" so people can chose if that should be included or not.

magicus: And we should ignore the bad data on the Wynncraft map labels, and just set the integer 60 for the string "60-70". So level just becomes a single, trivial integer.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapVisibility.java
```diff
@@ -4,12 +4,12 @@
...
-    // above zoom: fully visible at, last visible at
-    // between zoom: both above
+@FunctionalInterface
+public interface MapVisibility {
+    /**
+     * This method is used to determine the visibility of a feature at a given zoom level.
+     * @param zoomStep The current zoom step of the map. You can read more about zoom steps in the {@link com.wynntils.utils.render.MapRenderer} class.
+     * @return A float between 0 and 1, where 0 means the feature is not visible and 1 means the feature is fully visible. Values in between are used to determine the transparency of the feature.
+     */
+    float getVisibility(int zoomStep);
```
magicus: Given how the zoom PR ended up, we should use the term "zoom level" and have it as a float everywhere.

```suggestion
    float getVisibility(float zoomLevel);
```

common/src/main/java/com/wynntils/services/map/Label.java
magicus: I see what you mean, but I think that is problematic. 

1) I think such a judgement will be much more subjective than specifying a min level. If you go to places above your level, you'll die almost instantly by mobs. Otoh, I can sometimes go to too simple places just because it is more peaceful there. How do you determine that level?

2) The official sources have a min level specified, but not a max level. That means we can not get such data for official sources.

magicus: I mean, consider Detlas. What would be the max level there? Is it 100, since you need to go there to start A Journey Beyond? Or is it 50, since it starts to fade in relevance after that?

kristofbolyai: Okay. Let's do integers as levels for now.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: Had to change this up a little bit. You forgot to deal with inheritance. I think my version should work for you.

kristofbolyai:
@magicus I think I did everything we've discussed.

common/src/main/java/com/wynntils/services/map/Label.java
```diff
@@ -41,29 +47,44 @@ public LabelLayer getLayer() {
...
-        } catch (NumberFormatException e) {
-            return 0;
-        }
-    }
-
     public Location getLocation() {
         return new Location(x, 0, z);
     }
 
+    public static final class LabelDeserializer implements JsonDeserializer<Label> {
```
magicus, line 58: When we migrate to the new mapdata providers, I think we can just enforce having label as an integer instead of a string. I don't recall the details, but apparently thenew system reuses the services.map.Label and PoiService..? Maybe we should split it off so we can have a fresh start with an integer instead of a string for the json.

common/src/main/java/com/wynntils/services/mapdata/attributes/impl/FadingMapVisiblity.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public FadingMapVisiblity(float min, float max, float fade) {
+        this.min = min;
+        this.max = max;
+        this.fade = fade;
+    }
+
+    @Override
+    public float getVisibility(float zoomLevel) {
+        // A feature stats fading in at min - fade, is fully visible at min,
```
magicus, line 22: I'd like to see this somewhat different. The change I would like to see is that the fading should be "centered" around min and max, instead of extending the visibility. That is, if you change the fade it should still feel like the icon is shown for roughly the same interval. With this code, increasing the fade will make the icon visibile for a much longer period, and that will likely make people set a shorter visible interval to get the desired effect.

So basically what I am saying is that the max/min specifies the interval where the icon is shown with at least 50% visibility.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
```diff
@@ -20,7 +20,10 @@ public interface MapAttributes {
     // 0 means no value specified; inherit
     int getPriority();
 
-    // 0 means no information is available, or suitable for all combat levels
+    // the minimum combat level for which this feature is suitable for
+    // 1 means suitable for all levels (players start at level 1)
+    // 0 means inherit
+    // -1 means no information is available, or the feature is suitable for all levels
```
magicus, line 26: So can we set -1 as the combat level at the "root"? So if someone adds a data point and leaves out the level, the default value will be "no info".

kristofbolyai: Every other method here specified their default as inherit. I think this method should do the same, and we'll specify the real default value somewhere else (which will be called when the root is "inherit".

common/src/main/java/com/wynntils/utils/type/RangedValue.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
```
magicus, line 2: See if you can revert this.

common/src/main/java/com/wynntils/services/map/Label.java
kristofbolyai: I've been thinking about dropping the Label class. I've actually had to figure out how a "label" turns into a map feature. I thought we just implement the map
feature interface on the label class, but no. It is wrapped in a "service provider". I wonder that if the service class should not use **this** label class, but just a temporary deserialization class.

common/src/main/java/com/wynntils/services/mapdata/attributes/impl/FadingMapVisiblity.java
kristofbolyai: Is that desirable behavior, though? Personally, I would expect min max values a point where they either start or stop fading. Now, it's a mix of both, which I agree can be a bit confusing, but overall, I thought this makes the most sense, without having to read "documentation" for the user.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: Ok. I've apparently forgotten like everything about this. :) Where do we specify the root? Can we make sure this is set to -1?

kristofbolyai: I am a bit lost in that regard still. Here is what I've found. It seems that `FullFeatureAttributes` and `FullCategoryAttributes` expect inherit values to be null, or 0. I would assume these classes would provide/get the default values. Somewhere, sometime. I think i'll have to figure this out, when I am working on porting to map data.

(For reference, here is a code snippet from `FullFeatureAttributes`:
```java
    protected <T> T getAttribute(Function<MapAttributes, T> getter) {
        // Check if the feature has overridden this attribute
        if (attributes != null) {
            T attribute = getter.apply(attributes);
            if (attribute != null && !(attribute instanceof Integer i && i == 0)) {
                return attribute;
            }
        }

        // Otherwise get it from the category
        MapAttributes categoryAttributes = Services.MapData.getFullCategoryAttributes(feature.getCategoryId());
        if (categoryAttributes == null) return null;

        return getter.apply(categoryAttributes);
    }
```

common/src/main/java/com/wynntils/utils/type/RangedValue.java
kristofbolyai: Only by force pushing. Do you want me to?

common/src/main/java/com/wynntils/services/map/Label.java
magicus: Hm, I'm vaguely recalling this logic. I think the Label class was created to match the Wynncraft web map data with the place labels. When we switch to proper MapData, this will be redone as a normal json file. I think that was one of my actual test cases when designing the MapData structure/json format, that this should be possible.

kristofbolyai: Okay. So what if we commit this now, and when I am actually porting the sources, I'll do whatever is required to clean this up and adopt the new system.

common/src/main/java/com/wynntils/services/mapdata/attributes/impl/FadingMapVisiblity.java
magicus: I think that is desirable, yes. Here are my two arguments:

1) This decouples fading from the actual max/min point. Otherwise you will probably need to adjust max/min at the same time you adjusts fading, to get your desired effect.

2) If you don't select the midpoint of the fading range, then what do you select? When the point starts to fade or stops to fade? It is kind of impossible to know.

My idea is to think of this as without fading. Then there is a single instance at which the visibility kicks in (or out). But the `fade` argument makes this a little less abrupt. But if you slide the fading up or down, it should still center around the actual value.

common/src/main/java/com/wynntils/utils/type/RangedValue.java
magicus: Yes, I think it is okay to force-push to a PR branch to get rid of a thing like this. A bit annoying that this is the only way to fix this, but in this case I think it's fine. I assume you'll make an interactive rebase and remove all edits to this file?

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: This is the one point that is still worrying me. I think we need to get to the bottom on how this is supposed to work, so we get this part right.

Maybe it is bad that we use 0 for inheritance on the other integer values as well. Perhaps we need a way to say if an integer value is present or not, basically having nullable `Integer`s instead of `int`. That is harder to match to a json format, though.

common/src/main/java/com/wynntils/services/map/Label.java
magicus: Okay, I think that is good. That is, for the Label part. We still need to figure out the level-0 thing.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: So, do you have any idea on how to handle this? Nullable ints are not so bad to handle in GSON, afaik.

magicus:
This has been superseded by https://github.com/Wynntils/Artemis/pull/2489.

PR #2376, opened 2024-03-17
feat: Add favorite filter [skip ci]

ShadowCat117:
Bank
Unidentified favourite items don't display the favourite star but are still counted as favourites (Fixed in [fc2763d](https://github.com/Wynntils/Artemis/pull/2376/commits/fc2763d0a156018ea063b3d3c74e5cf8be607003))
TM
IngGuide
CharmGuide


kristofbolyai:
> Unidentified favourite items don't display the favourite star but are still counted as favourites

Is it hard to fix in this PR?

kristofbolyai (commented):
The items you have implemented the named property for, usually have item stacks that create fake tooltips. Can you make those tooltips use these names, so at least it is a little bit more unified? That should make it more obvious to change, if these names ever change.

ShadowCat117:
> Is it hard to fix in this PR?

I didn't really look into the cause, so it might be a simple fix. I'll have a look

ShadowCat117:
image

ShadowCat117:
> The items you have implemented the named property for, usually have item stacks that create fake tooltips. Can you make those tooltips use these names, so at least it is a little bit more unified? That should make it more obvious to change, if these names ever change.

Unless I'm misunderstanding what you meant, I'm not sure if it can use the same name as the tooltip will need colours and these don't use colours to make it match what gets stored in the favourites list. The names that the guide item stacks currently use are the same just with colour

kristofbolyai:
> > The items you have implemented the named property for, usually have item stacks that create fake tooltips. Can you make those tooltips use these names, so at least it is a little bit more unified? That should make it more obvious to change, if these names ever change.
> 
> Unless I'm misunderstanding what you meant, I'm not sure if it can use the same name as the tooltip will need colours and these don't use colours to make it match what gets stored in the favourites list. The names that the guide item stacks currently use are the same just with colour

Yeah, I see. I think we could post-process and color it, but that's not needed here.

kristofbolyai (approved):

kristofbolyai:
Looks good! Looking forward to the UI!

PR #2378, opened 2024-03-21
feat: Migrate to wiki.gg [skip ci]

ShadowCat117:
All works as far as I can tell.

Unfortunately wiki.gg encodes the `'` in the quest list so that has to be changed back. [Fandom list](https://wynncraft.fandom.com/index.php?title=Special:CargoExport&format=json&tables=Quests&fields=Quests._pageTitle) [wiki.gg list](https://wynncraft.wiki.gg/index.php?title=Special:CargoExport&tables=Quests&&fields=Quests._pageTitle&order+by=&limit=2000&format=json)

### DIFF
--- a/common/src/main/java/com/wynntils/models/activities/quests/QuestModel.java
+++ b/common/src/main/java/com/wynntils/models/activities/quests/QuestModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.activities.quests;
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import net.minecraft.world.phys.Vec3;
 import net.minecraftforge.eventbus.api.EventPriority;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/models/activities/quests/QuestModel.java
```diff
@@ -132,15 +134,16 @@ public void openQuestOnWiki(QuestInfo questInfo) {
...
 
             Managers.Net.openLink(UrlId.LINK_WIKI_LOOKUP, Map.of("title", wikiName));
-            return;
+        } else {
+            ApiResponse apiResponse =
+                    Managers.Net.callApi(UrlId.API_WIKI_QUEST_PAGE_QUERY, Map.of("name", questInfo.getName()));
+            apiResponse.handleJsonArray(json -> {
+                String pageTitle = WIKI_APOSTROPHE
+                        .matcher(json.get(0).getAsJsonObject().get("_pageTitle").getAsString())
+                        .replaceAll("'");
```
kristofbolyai: Isn't there a String#replace method with plain string inputs?

ShadowCat117: Yes, for some reason I was thinking it would only replace the first occurrence

kristofbolyai: Iirc the method is named replace so one would assume the same, especially after using other languages..

kristofbolyai (approved):
PR the urls.json change to Static-Storage please.

kristofbolyai:
Also, I think you should increment the version of urls.json both here and in Static-Storage. (as we modified a link)

PR #2382, opened 2024-03-25
fix: Fix NpcDialogueFeature legacy mode not displaying text when translations are off

kristofbolyai:


common/src/main/java/com/wynntils/models/npcdialogue/NpcDialogueModel.java
```diff
@@ -140,14 +140,21 @@ public void handleDialogue(List<Component> chatMessage, boolean protectedDialogu
...
             final NpcDialogue oldDialogue = dialogue;
 
+            // Should never happen, but just in case
+            if (styledTexts.isEmpty()) {
+                WynntilsMod.warn("Dialogue processing returned an empty list.");
+                WynntilsMod.postEvent(new NpcDialogueProcessingEvent.Post(oldDialogue, oldDialogue.currentDialogue()));
+                return;
+            }
+
             // If the dialogue is the same as the old one, don't do anything
```
magicus, line 153: This comment seems to not apply anymore..?

kristofbolyai: We still don't do any processing, but yes, I overlooked that..

PR #2386, opened 2024-04-03
fix: Fix item record crashing if there is a faulty item, fix item saving mode

kristofbolyai:
This PR adds the proper save flags for saved items, which is extended encoding. We did not do this previously, slowly breaking the saved items with API changes. Unfortunately, we cannot make the item record keep up with item changes, all items in there need to be saved as-is at the time of recording. I've also added a fallback cleanup mechanism for the future, quarantining faulty items and trying to re-add items that parse (after a patch or other change).

### DIFF
--- a/common/src/main/java/com/wynntils/screens/itemsharing/SavedItemsScreen.java
+++ b/common/src/main/java/com/wynntils/screens/itemsharing/SavedItemsScreen.java
@@ -71,6 +71,10 @@ public final class SavedItemsScreen extends WynntilsContainerScreen<SavedItemsMe
     @Override
     protected void doInit() {
         super.doInit();
+
+        // If our encoding breaks for some reason, we need to clean up the item record to prevent crashes
+        Services.ItemRecord.cleanupItemRecord();
+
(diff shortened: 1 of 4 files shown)


magicus:
I'm not sure I understand the problem you are trying to solve. What is EncodingSettings? If I have known, then I have forgotten by now... :-)

kristofbolyai:
> I'm not sure I understand the problem you are trying to solve. What is EncodingSettings? If I have known, then I have forgotten by now... :-)

It's whether we encode that current stat ranges in the encoding. We must, because the API might update. Basically, extended encoding works without any API data (for identifications). The second setting is saving item name, we might as well do it always.

EDIT: Basically, we did not use the correct encoding, but the one meant for one-time use in chat (shorter), but as the API is updating items with new stats, the decoding fails (with this PR, the item is theoretically decodable in the future any time, as if it was frozen in time. That is how the item vault is meant to work)

magicus (approved):
I'm still not sure I understand but I guess that is fine 🤷

PR #2389, opened 2024-04-06
feat: Item Filter GUI

ShadowCat117:
Tested it a fair amount and I can't find any issues with it but I would appreciate it if someone else would give it a thorough test as I may have overlooked something.

It's a little hacky when populating the widgets with new values from the search bar and vice versa but I couldn't think of a better way to keep them in sync.

The first commit makes it so that now only screens that can use the advanced search will use it, instead they will have the original search widget.

Something I did want to add was a way to filter what filters are available on a per screen basis, for example the item guide screen has no use for the profession filter, or only the trade market can use the price filter etc.

https://github.com/Wynntils/Artemis/assets/8337467/4d699694-1fcc-4c59-aae0-cca95d3744d4


ShadowCat117:
@magicus  The first commit for this is a partial implementation of what you mentioned before about having different classes for each container instead of the enum, I haven't fully implemented it yet as I wanted to get your opinion on what I've done so far so for now it's just for the searchable/scrollable containers and personal storage.

Future things I would add to this approach would be a property for `FullscreenContainerProperty` for containers that replace the inventory as well, such as the ability tree and content book, then instead of checking for ability tree in the item lock feature, we won't lock the slot if the screen is a fullscreen one.
Or for crafting stations that all have the same slots for materials and ingredients, we could detect what profession it is and display a tooltip of what you will be crafting based on the items given and that could use a `CraftingStationContainer` similar to `PersonalStorageContainer`

kristofbolyai:
It's too late to review this today, but it is really incredible, watching the showcase video it seems everything is handled beautifully.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
```diff
@@ -215,91 +235,114 @@ private void tryAutoSearch(AbstractContainerScreen<?> abstractContainerScreen) {
...
     }
 
-    private InteractiveContainerType getCurrentInteractiveContainerType(Screen screen) {
-        InteractiveContainerType containerType = null;
-
-        for (InteractiveContainerType type : InteractiveContainerType.values()) {
-            if (type.isScreen(screen)) {
-                containerType = type;
+    private SearchableContainerProperty getCurrentSearchableContainer() {
+        // FIXME: Do something about this
```
kristofbolyai: I guess you could put these in a `Map<Class<? extends SearchableContainerProperty>, Supplier<Boolean>>` but that wouldn't look much nicer, we would just move the code.

common/src/main/java/com/wynntils/models/containers/BankModel.java
```diff
@@ -68,24 +69,20 @@ public class BankModel extends Model {
 
     private boolean editingName;
     private int currentPage = 1;
-    private InteractiveContainerType currentContainer;
+    private PersonalStorageType currentContainer = null;
```
kristofbolyai: Maybe this field and the getter requires a rename? `storageContainerType`?

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
```diff
@@ -64,11 +108,42 @@ public boolean isLootOrRewardChest(String title) {
...
+        registerWynncraftContainer(new HousingJukeboxContainer());
+        registerWynncraftContainer(new HousingListContainer());
+        registerWynncraftContainer(new JukeboxContainer());
+        registerWynncraftContainer(new LobbyContainer());
+        registerWynncraftContainer(new MiscBucketContainer());
+        registerWynncraftContainer(new PetMenuContainer());
+        registerWynncraftContainer(new ScrapMenuContainer());
+        registerWynncraftContainer(new TradeMarketFiltersContainer());
+        registerWynncraftContainer(new TradeMarketPrimaryContainer());
+        registerWynncraftContainer(new TradeMarketSecondaryContainer());
```
kristofbolyai, line 149: Maybe we could keep it alphabetical by type, kind of like features. But we can keep it as is, really doesn't matter.

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerProperty.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.containers.type;
+
+public interface SearchableContainerProperty extends ScrollableContainerProperty {
+    ContainerBounds getBounds();
+
+    boolean supportsAdvancedSearch();
```
kristofbolyai, line 10: Did we ever have a way to disable the advanced search with a config? If yes, do you know if that works? (If we have no such config, ignore this :)).

common/src/main/java/com/wynntils/models/containers/type/wynncontainers/AbilityTreeContainer.java
```diff
@@ -0,0 +1,38 @@
...
+import com.wynntils.models.containers.type.ScrollableContainerProperty;
+import com.wynntils.models.containers.type.WynncraftContainer;
+import java.util.regex.Pattern;
+
+public class AbilityTreeContainer extends WynncraftContainer implements ScrollableContainerProperty {
+    private static final Pattern NEXT_PAGE_PATTERN = Pattern.compile("§7Next Page");
+    private static final Pattern PREVIOUS_PAGE_PATTERN = Pattern.compile("§7Previous Page");
+
+    public AbilityTreeContainer() {
+        super(Pattern.compile("(?:Warrior|Shaman|Mage|Assassin|Archer) Abilities"));
```
kristofbolyai: Although this specific case may not have a test at the moment, can we change the constructor pattern to extract the pattern into a static final field? Otherwise, these regexes become un-testable by our framework. Please change every container class.

common/src/main/java/com/wynntils/screens/base/widgets/ItemSearchButton.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package com.wynntils.screens.base.widgets;
+
+import com.wynntils.screens.itemfilter.ItemFilterScreen;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraft.client.gui.components.Tooltip;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.network.chat.Component;
+
+public class ItemSearchButton extends WynntilsButton {
```
kristofbolyai: `ItemSearchButton` is a bit too broad, I had to read the code to know what this does. I propose renaming to `ItemSearchUIButton`/`ItemFilterUIButton`.

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
```diff
@@ -0,0 +1,1279 @@
...
+    public void setFocusedTextInput(TextInputBoxWidget focusedTextInput) {
+        this.focusedTextInput = focusedTextInput;
+    }
+
+    public void addFilter(Pair<ItemStatProvider<?>, String> newFilter) {
+        filters.add(newFilter);
+        usedProviders.add(newFilter.a());
+
+        createFilterQuery();
+        populateQuery();
```
kristofbolyai: Do you not need to call `populateSorts` here and in `removeFilter`? If not, maybe you could include a comment about them below the two calls.

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
```diff
@@ -0,0 +1,1279 @@
...
+        } else {
+            for (AbstractWidget widget : sortButtons) {
+                this.removeWidget(widget);
+            }
+
+            sortButtons = new ArrayList<>();
+        }
+    }
+
+    private void parseFilters(String input) {
```
kristofbolyai: I would assume this is the hacky part you mentioned before. At first sight, my first idea/question is why not just use `ItemFilterService#createSearchQuery`? You get back a parsed object, from that, you can easily get all providers/sorts/whatever you need. Would this work?

common/src/main/java/com/wynntils/screens/itemfilter/widgets/IntegerValueWidget.java
```diff
@@ -0,0 +1,508 @@
...
+            } else if (greaterThanInputQuery.isEmpty() && greaterThanMatcher.matches()) {
+                greaterThanInputQuery = greaterThanMatcher.group(2);
+                greaterThanEqual = greaterThanMatcher.group(1).equals(">=");
+                this.greaterThanPercentage = value.endsWith("%") && supportsPercentage;
+            } else if (lessThanInputQuery.isEmpty() && lessThanMatcher.matches()) {
+                lessThanInputQuery = lessThanMatcher.group(2);
+                lessThanEqual = lessThanMatcher.group(1).equals("<=");
+                this.lessThanPercentage = value.endsWith("%") && supportsPercentage;
+            }
+        }
```
kristofbolyai: Can't you call `updateValues`?

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
kristofbolyai: I overall think  that the widget's `updateValues` do too much parsing, I assume you know this as well. I wonder if you would still need this parsing if'd done the suggestion above, since `ItemSearchQuery` has the StatFilters too, perhaps it could be passed to the widget to make parsing easier since you would only have to decide how you represent the filter in the widget itself, without needing to parse it.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/GearTypeStatProvider.java
```diff
@@ -16,4 +18,23 @@ public List<String> getValue(WynnItem wynnItem) {
 
         return List.of(gearTypeItemProperty.getGearType().name());
     }
+
+    @Override
+    public List<String> getValidInputs() {
+        // Can't use all values of the enum as WEAPON, ACCESSORY, MASTERY_TOME and CHARM
+        // don't work with this filter, ItemTypeStatProvider should be used for tomes and charms
```
kristofbolyai, line 25: `ItemTypeStatProvider` doesn't work for WEAPON, ACCESSORY?

common/src/main/java/com/wynntils/services/itemfilter/statproviders/ItemTypeStatProvider.java
```diff
@@ -13,4 +34,32 @@ public class ItemTypeStatProvider extends ItemStatProvider<String> {
...
+                IngredientItem.class.getSimpleName().replace("Item", ""),
+                InsulatorItem.class.getSimpleName().replace("Item", ""),
+                MaterialItem.class.getSimpleName().replace("Item", ""),
+                MultiHealthPotionItem.class.getSimpleName().replace("Item", ""),
+                PotionItem.class.getSimpleName().replace("Item", ""),
+                PowderItem.class.getSimpleName().replace("Item", ""),
+                RuneItem.class.getSimpleName().replace("Item", ""),
+                SimulatorItem.class.getSimpleName().replace("Item", ""),
+                TeleportScrollItem.class.getSimpleName().replace("Item", ""),
+                TomeItem.class.getSimpleName().replace("Item", ""));
```
kristofbolyai: We will forgot to update this. Can we get all inheritors of WynnItem and display all, even if not all will be "real" items?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -2166,28 +2199,49 @@
   "service.wynntils.itemFilter.sortingNotSupported": "Sorting is not supported for this container",
   "service.wynntils.itemFilter.stat.actual.description": "%s stat of the item",
   "service.wynntils.itemFilter.stat.countedItem.description": "Count of an item (e.g. health potion), this is not the same as the amount of items in a stack.",
+  "service.wynntils.itemFilter.stat.countedItem.name": "Item Count",
```
kristofbolyai, line 2209: I see you added these name i18ns. Did you add an assert for them in the code?

common/src/main/java/com/wynntils/models/containers/type/SearchableContainerProperty.java
ShadowCat117: I don't think so, when the advanced search was added it just replaced the original basic search

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
ShadowCat117: No, because those 2 methods can only be called from the value widgets so you won't be viewing the sorts at the moment

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
ShadowCat117: I think `parseFilters` is fine, it's the `updateValues` I'm not a fan of especially on the `IntegerValueWidget`. I'll look into using the services method, maybe that will be better

common/src/main/java/com/wynntils/services/itemfilter/statproviders/GearTypeStatProvider.java
ShadowCat117: No, you can only do `itemType:gear`, there is no `WeaponItem` or `AccessoryItem`

common/src/main/java/com/wynntils/services/itemfilter/statproviders/ItemTypeStatProvider.java
ShadowCat117: I was going to do that originally but I couldn't find a good way to do it

common/src/main/resources/assets/wynntils/lang/en_us.json
ShadowCat117: No, the providers for skills and actual stats don't use i18n

kristofbolyai: Oh, okay.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/GearTypeStatProvider.java
kristofbolyai: Maybe we should implement that. Probably not here though.

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
kristofbolyai: If you could add a comment in the code, we can mark this as completed.

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
ShadowCat117: I tried using this but was unable to get it to work for filters. Sorts and plain queries were fine but after removing the parsing from the constructors and always calling `updateValues` I'm happier with how it works. The `ignoreUpdate` field in the widgets still bothers me a bit but I don't see how it can be avoided, unless I add a `setTextBoxInputWithoutConsuming`..

kristofbolyai: I'll try to make a patch to make this work.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/ItemTypeStatProvider.java
magicus: That seems to be tricky, https://stackoverflow.com/questions/205573/at-runtime-find-all-classes-in-a-java-application-that-extend-a-base-class. 

Since we own the code, can't we change the WynnItem supertype to register themselves automatically?

magicus:
I'd like to echo bolyai: This looks fantastic, judging from the video! But it is quite a bit to review. I'll try to get to it, but I can't promise when.

ShadowCat117:
> I'd like to echo bolyai: This looks fantastic, judging from the video! But it is quite a bit to review. I'll try to get to it, but I can't promise when.

It's mostly the first commit I'd like your opinion on so if you only have time for that don't worry about the screen itself :)

kristofbolyai:
Yes magicus, if you review the screen refactors, I can review the widgets and UI, if you don't have time for it.

common/src/main/java/com/wynntils/features/inventory/ContainerSearchFeature.java
ShadowCat117: I think this looks a lot better than all the if statements

kristofbolyai: Indeed it looks better, all is well then.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/ItemTypeStatProvider.java
kristofbolyai: I suggest getting `ItemHandler`'s `ItemAnnotator`s, they have the same name as the items, you just replace the "Annotator" part. Otherwise, they should exist whenever an item exists. You can probably Make two subclasses for `ItemAnnotator`, `GameItemAnnotator`, `GuiItemAnnotator` (and leave the last 3 edge cases as normal implementations).

common/src/main/java/com/wynntils/screens/itemfilter/widgets/StringValueWidget.java
```diff
@@ -4,138 +4,439 @@
...
+        for (int i = 0; i < MAX_OPTIONS_PER_PAGE; i++) {
+            currentValue = i + scrollOffset;
+
+            if (currentValue > filters.size() - 1) break;
+
+            if (filters.get(currentValue).statFilter() == null) {
+                widgets.add(new StringWidget(getX(), renderY, "", false));
+            } else {
+                String currentInput = filters.get(currentValue).statFilter().asString();
+                boolean strict = currentInput.startsWith("\"") && currentInput.endsWith("\"");
```
kristofbolyai: You could check for `instanceof StringStatFilter`, then call `isStrict`.

common/src/main/java/com/wynntils/screens/itemfilter/ItemFilterScreen.java
```diff
@@ -670,6 +670,10 @@ public void setFiltersForProvider(ItemStatProvider<?> provider, List<StatProvide
         }
 
         updateQueryString();
+
+        if (provider == selectedProvider) {
+            valueWidget.onFiltersChanged(filterPairs);
+        }
```
kristofbolyai: I can't seem to understand why this is needed. It's also the reason as to why `BooleanValueWidget` needs the null check for filters.

ShadowCat117: It's so when you interact with the provider button to add the any filter or remove all filters it updates the widget

ShadowCat117: This has been removed now

ShadowCat117:
Closing for now. Will re open a pr for the `ContainerModel` changes shortly then a new one for the filter GUI.

PR #2390, opened 2024-04-06
feat: Add a custom ability tree

kristofbolyai:


ShadowCat117:
Could be cool to have the custom UI one day but the main appeal of this was the loadouts which is now done in https://github.com/Wynntils/Wynntils/pull/4183 so will close this as it's VERY outdated

PR #2392, opened 2024-04-07
feat: Add OR support for item filters [ci skip]

kristofbolyai:
Prequisite for https://github.com/Wynntils/Artemis/pull/2389.

image
image
image
image
image


ShadowCat117 (approved):
I actually had multiple of the same filters be sorted in a list like this by accident when working on the screen, guess I need to reimplement that :')

kristofbolyai:
> I actually had multiple of the same filters be sorted in a list like this by accident when working on the screen, guess I need to reimplement that :')

I noticed that your screen allowed "rarity: x rarirty:y" but that didn't work. I added it, but then adding separator support was easy.

PR #2397, opened 2024-04-09
feat: Bomb Bell Relay Keybinds [skip ci]

ryanzhoudev:
image

Not tested with multiple bombs actually existing yet. If you are confident the comparing stuff works, feel free to merge but otherwise I will be able to test it within the next 2 days. 

Allowed, reference:
https://discord.com/channels/394189072635133952/405350459126317065/1224567174946099210


common/src/main/java/com/wynntils/features/chat/BombBellRelayFeature.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @RegisterKeyBind
+    private final KeyBind relayGuildKeybind =
+            new KeyBind("Relay Bomb to Guild", GLFW.GLFW_KEY_UNKNOWN, true, this::relayToGuild);
+
+    private String getAndFormatLastBomb() {
+        BombInfo lastBomb = Models.Bomb.getLastBomb();
+        if (lastBomb == null) return null;
+
+        // Intentionally not localized as the majority of the playerbase speaks English
```
magicus: I don't think that makes sense. We mostly have English translations, but we always use i18n hooks. Remember that you can use placeholders for variables in the i18n texts, so this is not any more tricky to translate than any other text.

kristofbolyai: But, let's say I am german, I play with German language, but I am in an English guild. Obviously I want to send the message as english.

magicus: Ah, I see. Right, this is the message sent to other players. Hm. Do we have any other places where we sent out chat messages like this? Are they kept in English, or are they mistakenly translated? Like coordinate sharing etc.

common/src/main/java/com/wynntils/features/chat/BombBellRelayFeature.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @RegisterKeyBind
+    private final KeyBind relayGuildKeybind =
+            new KeyBind("Relay Bomb to Guild", GLFW.GLFW_KEY_UNKNOWN, true, () -> relayTo("g"));
+
+    private String getAndFormatLastBomb() {
+        BombInfo lastBomb = Models.Bomb.getLastBomb();
+        if (lastBomb == null) return null;
+
+        // Intentionally not localized as the majority of the playerbase speaks English
```
magicus:
```suggestion
        // This is not localized as it is sent to other players
```

common/src/main/java/com/wynntils/features/chat/BombBellRelayFeature.java
kristofbolyai: They are all hard coded afaik.

ryanzhoudev: Coordinate ones look okay 
image

magicus: Ok, good. Maybe we should add a comment at those places too, to describe why they are not localized.

kristofbolyai (approved):
LGTM

magicus (approved):

ShadowCat117:
This might be just be personal preference, but I think having clickable text after the bomb bell to share with party/guild could be nice for when multiple bombs are thrown at once.

Still keep the keybind as most of the time it'll be 1 bomb at a time and you'll only want to share the most recent

ryanzhoudev:
image

Works fine with multiple bombs.

ryanzhoudev:
> This might be just be personal preference, but I think having clickable text after the bomb bell to share with party/guild could be nice for when multiple bombs are thrown at once.
> 
> Still keep the keybind as most of the time it'll be 1 bomb at a time and you'll only want to share the most recent

I feel like you could just double tap the bind (like I sort of did in the image), there's no rate limit except the one the server provides for /p and /g. I am not too fond of modifying the chat and opening that can of worms at the moment, so maybe we could revisit in the future if there is demand?

ShadowCat117:
I was thinking more just sending a seperate client side message after the bomb thrown message, like the click to claim item bomb messages rather than modifying the original as that's not a good idea, not sure if we have a BombThrownEvent but that should be easy enough to add.

It would be more useful when bombs are thrown in pairs, prof xp and speed or loot and combat xp are ones I often see thrown at the same time and only being able to share the most recent, you would have to manually send a message if you also wanted to convey that there was another bomb thrown.

PR #2398, opened 2024-04-09
feat: Above ground function

ryanzhoudev:
I put this in Combat functions since it will primarily be used for acrobat/battle monk. Also doesn't really feel like it fits anywhere else very well.


common/src/main/java/com/wynntils/functions/CombatFunctions.java
```diff
@@ -40,4 +42,35 @@ protected List<String> getAliases() {
             return List.of("adavg");
         }
     }
+
+    public static class HeightFunction extends Function<Double> {
+        @Override
+        public Double getValue(FunctionArguments arguments) {
+            // iteratively find the first non-air block below the player
+            double endY = (int) Math.ceil(McUtils.player().position().y) - 1;
```
magicus: I'm pretty certain Minecraft vanilla keeps a height map where you can look up the highest non-air-block. Don't remember the name of the function though, but google is likely your friend.

ryanzhoudev: I originally tried that, but it's not suitable for two reasons:
- It returns an int, so it doesn't work with slabs/carpets which is significant since acrobat wants 1.25 blocks off the ground
- It doesn't play well with caves/verticality at all, there is no way to specify which "pocket" of air you want to look at

kristofbolyai: Okay, but if you have to produce this value yourself, place it in a model. Functions are no such place to do almost any calculation.

magicus (commented):
The code itself looks good, but I am not happy about the name. As you say, you could be in a cave but 5 blocks up and thus have "height" or "altitude" 5. My natural instinct is to take both these concept to be related to some fixed point, like Y = 0 or y = 64 or so.

I don't have any foolproof replacement to suggest. Maybe "airborne height"? "air below"? "ground distance"?

ryanzhoudev:
> The code itself looks good, but I am not happy about the name. As you say, you could be in a cave but 5 blocks up and thus have "height" or "altitude" 5. My natural instinct is to take both these concept to be related to some fixed point, like Y = 0 or y = 64 or so.
> 
> I don't have any foolproof replacement to suggest. Maybe "airborne height"? "air below"? "ground distance"?

I wouldn't mind changing it to just above_ground_level, I already have the agl abbreviation in there. Thoughts?

magicus:
Sounds good!

magicus:
Or maybe just `above_ground`?

ryanzhoudev:
Sure, I'll do above_ground as the primary and agl + the full name as the aliases

kristofbolyai:
> Or maybe just `above_ground`?

Sounds like a boolean to me.

ryanzhoudev:
> > Or maybe just `above_ground`?
> 
> Sounds like a boolean to me.

Kind of, but isn't it strange to have that as a function since you are always above some type of ground? I don't know what people would immediately think of it. Maybe `blocks_above_ground` then?

magicus:
Go with "blocks above ground". It is unambiguous.

magicus (approved):

PR #2399, opened 2024-04-10
fix: Use the new dropMeta info for showing item guesses [skip ci]

ryanzhoudev:


ryanzhoudev:
Actually this might not work for certain event boxes like the Hero's Mask thing from the current event.. I'll look into the api v3's dropMeta component

kristofbolyai:
> Actually this might not work for certain event boxes like the Hero's Mask thing from the current event.. I'll look into the api v3's dropMeta component

The API is actually kind of unstable. The `dropMeta` coordinates are surely off, the "type" might be okay.

kristofbolyai:
Perhaps fix the duplications here, if you want to.
image

magicus:
If we are not 100% sure we can trust the API, I think it is better *not* to exclude potential items due to drop type. In the end, the user can themselves reason if any of the provided alternatives are not possible, but if we exclude something that would have been possible then they're in for a surprise.

magicus:
The duplications are a completely different issue? (But I agree that it looks bad and should be fixed)

common/src/main/java/com/wynntils/models/wynnitem/type/ItemObtainType.java
```diff
@@ -29,10 +29,22 @@ public enum ItemObtainType {
     QIRA_HIVE_MERCHANT("hive", "Qira Hive Merchant", true),
     QUEST("quest", "Quest"),
     RAID("raid", "Raid"),
-    SECRET_DISCOVER("discovery", "Secret Discovery"),
+    SECRET_DISCOVERY("discovery", "Secret Discovery"),
     SPECIAL_MOB_DROP("specialdrop", "Specific Mob Drop"),
     UNOBTAINABLE("unobtainable", "Unobtainable");
 
+    public static final List<ItemObtainType> BOXED_ITEMS = List.of( // All sources that possibly drop boxed items
```
kristofbolyai:
```suggestion
    // All sources that possibly drop boxed items
    public static final List<ItemObtainType> BOXED_ITEMS = List.of(
```

kristofbolyai (approved):

PR #2400, opened 2024-04-10
fix: Don't render lootrun path chests on air or barriers [skip ci]

ShadowCat117:
Requested at https://discord.com/channels/394189072635133952/405350459126317065/1227709549553778698

### DIFF
--- a/common/src/main/java/com/wynntils/services/lootrunpaths/LootrunRenderer.java
+++ b/common/src/main/java/com/wynntils/services/lootrunpaths/LootrunRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
+ * Copyright © Wynntils 2023-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.services.lootrunpaths;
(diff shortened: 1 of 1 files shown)


ryanzhoudev (approved):

ryanzhoudev:
Though I think this is more of a `fix:`

ShadowCat117:
> Though I think this is more of a `fix:`

Hmm could go either way, I'll go with fix

common/src/main/java/com/wynntils/services/lootrunpaths/LootrunRenderer.java
```diff
@@ -138,6 +140,10 @@ private static void renderChests(PoseStack poseStack, LootrunPathInstance lootru
         float blue = ((float) FastColor.ARGB32.blue(color)) / 255;
 
         for (BlockPos chest : chests) {
+            // Wynncraft requested that chest highlights are not rendered on these blocks
+            BlockState block = McUtils.mc().level.getBlockState(chest);
+            if (block.is(Blocks.BARRIER) || block.is(Blocks.AIR)) continue;
```
magicus, line 145: Does it mean that a chest that is not present at the moment will not be rendered? I'm not sure that is always a good thing. I have loaded lootruns that I downloaded just to get inspiration on where to look for chests.

... but then again, it was some time ago I played Wynncraft, maybe that does not make sense anymore with the new personal chest cooldown.

ShadowCat117: It won't render the outline for barriers or air, it was requested to stop things like this where barriers are being shown
image

ShadowCat117: Since you can place "chests" to be highlighted anywhere, even if they aren't actually chests

magicus (approved):

PR #2401, opened 2024-04-11
fix: Make blocks_above_ground function handle an edge-case hang

mbnkwm:
Some situations where it would previously enter an infinite loop are:
- when you're above the void in sky islands
- when you're entering or leaving a housing balloon
- using teleport scrolls

### DIFF
--- a/common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
+++ b/common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
@@ -91,7 +91,7 @@ public final class CharacterStatsModel extends Model {
             endY--;
 
             // stop checking beyond the minimum build height as there will never be any blocks below it
-            if (endY < McUtils.mc().level.getMinBuildHeight()) break;
+            if (endY < McUtils.mc().level.getMinBuildHeight()) return -1;
         }
 
         // add the floor height to the result to account for half-blocks
--- a/common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
+++ b/common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
@@ -90,9 +90,8 @@ public final class CharacterStatsModel extends Model {
                 .isAir()) {
             endY--;
 
-            if (endY < -64) {
-                break;
-            }
+            // stop checking beyond the minimum build height as there will never be any blocks below it
+            if (endY < McUtils.mc().level.getMinBuildHeight()) break;
         }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -89,6 +89,10 @@ public double getBlocksAboveGround() {
                         McUtils.player().blockPosition().getZ()))
                 .isAir()) {
             endY--;
+
+            if (endY < -64) {
+                break;
+            }
```
ShadowCat117:
```suggestion
            if (endY < -64) break;
```

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -89,6 +89,10 @@ public double getBlocksAboveGround() {
                         McUtils.player().blockPosition().getZ()))
                 .isAir()) {
             endY--;
+
+            if (endY < -64) {
```
ShadowCat117: Could also be worth extracting the -64 to a constant and a comment about why this check needs to happen

kristofbolyai:
I'll try to produce a version of this without iteration (or iteration handled by MC)

ShadowCat117:
@kristofbolyai Are we good to merge this or do you want to wait until you've produced a different version, better to have a fix out soon imo

magicus (approved):
Let's put out a fix quickly so we don't hang; then you can rewrite later to be better.

kristofbolyai:
> Let's put out a fix quickly so we don't hang; then you can rewrite later to be better.

Give me a moment please. I am profiling my changes.

common/src/main/java/com/wynntils/models/characterstats/CharacterStatsModel.java
```diff
@@ -89,6 +89,9 @@ public double getBlocksAboveGround() {
                         McUtils.player().blockPosition().getZ()))
                 .isAir()) {
             endY--;
+
+            // stop checking beyond the minimum build height as there will never be any blocks below it
+            if (endY < McUtils.mc().level.getMinBuildHeight()) break;
```
kristofbolyai: I think you should return -1 here. Otherwise the function produces an "infinite" value, which is hard to check for with Wynntils functions.
```suggestion
            if (endY < McUtils.mc().level.getMinBuildHeight()) return -1;
```

kristofbolyai:
Thanks for your PR @mbnkwm!

PR #2402, opened 2024-04-12
fix: Parse awakened bar & mask correctly [skip ci]

ShadowCat117:
Parse boss bar correctly & reverts mask type to correct mask after the effect expires

https://github.com/Wynntils/Artemis/assets/8337467/e7e3066b-8505-414a-911e-cc06e95c9f4d

### DIFF
--- a/common/src/main/java/com/wynntils/models/abilities/bossbars/AwakenedBar.java
+++ b/common/src/main/java/com/wynntils/models/abilities/bossbars/AwakenedBar.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.abilities.bossbars;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/models/abilities/bossbars/AwakenedBar.java
```diff
@@ -10,7 +10,7 @@
 import java.util.regex.Pattern;
 
 public final class AwakenedBar extends TrackedBar {
-    private static final Pattern AWAKENED_PATTERN = Pattern.compile("§fAwakening §7\\[§f(\\d+)/(\\d+)§7]");
+    private static final Pattern AWAKENED_PATTERN = Pattern.compile("§fAwakened §7\\[§f(\\d+)/(\\d+)§7]");
```
magicus, line 13: Just checking, did Wynn change their string or was this always wrong?

ShadowCat117: There's reports of it not working as far back as July last year so I presume it was changed before then, I've only just leveled a shaman myself so can't confirm when it was changed

kristofbolyai: This used to work baaack then.

kristofbolyai (approved):
LGTM

PR #2406, opened 2024-04-15
feat: Add beacons/POIs for territories with queued war timers

kristofbolyai:
image
image

### DIFF
--- a/common/src/main/java/com/wynntils/models/territories/TerritoryAttackTimer.java
+++ b/common/src/main/java/com/wynntils/models/territories/TerritoryAttackTimer.java
@@ -1,28 +1,36 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.territories;
 
+import com.wynntils.core.components.Models;
+import com.wynntils.models.territories.profile.TerritoryProfile;
+import com.wynntils.models.territories.type.GuildResourceValues;
+import java.util.Optional;
 import net.minecraft.ChatFormatting;
 
 public final class TerritoryAttackTimer {
+    private final TerritoryProfile territoryProfile;
+
     private final String territory;
     private final int minutes;
     private final int seconds;
 
-    private String defense;
+    private GuildResourceValues defense;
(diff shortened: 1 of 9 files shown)


common/src/main/java/com/wynntils/models/territories/TerritoryAttackTimer.java
```diff
@@ -1,28 +1,38 @@
...
         this.minutes = minutes;
         this.seconds = seconds;
-        this.defense = "Unknown";
+        this.defense = null;
     }
 
     public String asString() {
-        return ChatFormatting.GRAY + territory + ChatFormatting.YELLOW + " (" + defense + ")" + ChatFormatting.AQUA
-                + " " + timerString();
+        ChatFormatting defenseColor = this.defense == null ? ChatFormatting.YELLOW : this.defense.getDefenceColor();
```
ShadowCat117, line 31: Noticing we're use `defense` in some places and `defence` in others, I think we should be using `defense`. Definitely for a seperate pr though as it's used a lot elsewhere

ShadowCat117 (approved):

common/src/main/java/com/wynntils/models/territories/TerritoryAttackTimer.java
magicus: This is a recurring issue. :-( I think Wynncraft tried to standardize on one of them some time ago, and I think it was "defence". Imho we should use what they have in the name of attributes in their API.

kristofbolyai: One could also say, use MC source names (mojmap).

ShadowCat117: > This is a recurring issue. :-( I think Wynncraft tried to standardize on one of them some time ago, and I think it was "defence". Imho we should use what they have in the name of attributes in their API.

Unfortunately I think Wynn uses a mix of both

magicus: > Unfortunately I think Wynn uses a mix of both

They have done so historically but I believe they recently standardized on one of them. Was what I was trying to say. :-)

PR #2407, opened 2024-04-15
feat: Add Enabled Template to TextOverlay [skip ci]

Rafii2198:
This feels too simple, I assumed it would require much work, but I tried to do everything in game you can do and found no issues at all.

This adds a bit of customization to TextOverlays, the enabled template works the same way it does in custom bars, basically a function that returns a boolean which controls if this overlay will be rendered or not.
Benefits: Users do not need to recreate whole overlays just to change their display rules as well as makes Info Boxes a lot shorter as now you don't have to include the whole checking, then converting all the data to one type, then combining all that data which just makes the function exponentially longer and harder to read, with this you can keep a simple function and just throw another simple function into Enabled Template and as a side effect it makes the visual formatting with : a lot more common to use as currently, because many functions are only used in certain places like lootruns, there is just no way to use them at all, and they simplify the function a lot and make it more readable which in turn it makes it easier to get into and feel a lot better to write.


kristofbolyai:
Code looks alright at first, but I am a bit worried about "feature creep". Are we sure we wan't to make using functions harder? I know, this is not harder for most, but those people who barely find what to paste and where, I am not sure. However, it's not that bad that I even consider denying this PR outright, so, does anyone here anything for or against this change?

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
```diff
@@ -65,7 +69,7 @@ protected TextOverlay(int id) {
 
     @Override
     public void render(PoseStack poseStack, MultiBufferSource bufferSource, float partialTicks, Window window) {
-        if (!Models.WorldState.onWorld()) return;
```
magicus, line 68: I am worried about performance here. Our functions are much faster now (thanks bolyai!), but calling them once per render for every overlay... I dunno.

I thought we had some kind of buffering mechanism to calculate what to render which was called less frequently. This check should be there too.

magicus:
I think this might make sense. But, we might have to build/discuss some more around this. 

First, if people have trouble copy/pasting, this will make it harder. So maybe we need a good import/export functionality? Perhaps some way of expressing an overlay as a json object with a "condition" and "value" field for the two functions?

Secondly, how does this play with preview? If the condition is not valid, the text box should show up in preview anyway, but then the function it calculates is perhaps not valid?

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
```diff
@@ -32,6 +33,9 @@ public abstract class TextOverlay extends DynamicOverlay {
     @Persisted(i18nKey = "overlay.wynntils.textOverlay.fontScale")
     public final Config<Float> fontScale = new Config<>(1.0f);
 
+    @Persisted(i18nKey = "overlay.wynntils.textOverlay.enabledTemplate")
+    public final Config<String> enabledTemplate = new Config<>("string_equals(world_state;\"WORLD\")");
```
magicus: My suggestion is that the default is an empty template, and that the code use the current check if it finds an empty template.

magicus:
Also, we need error checking and handling. What if the template returns an integer instead of a boolean? What if it has syntax errors? How do we detect this, handle this, inform the user? For starter, I think the function needs to be validated when entered. (Also, how do we do this with the actual function to display?)

ShadowCat117:
> Code looks alright at first, but I am a bit worried about "feature creep". Are we sure we wan't to make using functions harder? I know, this is not harder for most, but those people who barely find what to paste and where, I am not sure. However, it's not that bad that I even consider denying this PR outright, so, does anyone here anything for or against this change?

I'm presuming this is so that functions don't need to be wrapped in some form of `{if_str(condition;actualfunction;"")}` which I would say is a minor improvement for larger functions. Custom bars have the enabled template since there's no way to hide the bar texture otherwise, but info boxes can use the `if_str` I mentioned so it's not strictly necessary for them to have an enabled template.

Including it will make the actual functions a tiny bit cleaner, but at the cost of potentially making them more difficult to understand for newer users

Rafii2198:
I don't think it will make it harder on newer users, you have to remember that in order to achieve it now in info boxes you would have to put a considerate amount of work for it to work and make it hard to read.
Take this small and tidy function as an example:
```
Challenges: {lootrun_challenges}\n
Time: {lootrun_time}
```
I think we all can agree that it is super easy to read, but then the user decides they want to make it only visible while they are in a lootrun they would have to change it to
```
{if_str(string_equals("NOT_RUNNING";lootrun_state);"";concat("Challenges: ";format_capped(lootrun_challenges);"\nTime: ";string(lootrun_time)))}
```
Now the function got a lot bigger and harder to read, now the function was fairly simple and thus the new is not as bad, but let's say you want to make the time into minutes and seconds, not just seconds, it will get much bigger because you can't use the : formatting which simplifies it a lot. What am I trying to say is that this approach makes it harder for new users, as it gets messy pretty fast. In comparison with Enabled Template, you don't have to change anything at all, instead just put `not(string_equals("NOT_RUNNING";lootrun_state))` there and your job is done, you can now expand the function to display more stuff without the need of formatting the data, converting types and joining it together.
I believe it will make it easier to get into not harder and definitely easier to debug, as you don't have to this whole converting and so on, there are fewer chances of creating an error and current approach usually also output whole function and says there is some error which is just confusing but as with enabled template everything is more tidy you will get more precise errors and less of them too.

Not only that, in case of other text overlays like for example Combat XP, if a user wants to see it only if they gain XP so it doesn't show if they idle to make their screen cleaner, currently the only option is not only to recreate the whole function for themselves from ground up, they have again to the whole thing with if_str and converting values, formatting them and joining and with enabled template they don't have to do any of that, they just need change enabled template in Combat XP Overlay itself, no need to recreate whole overlay just to have such simple functionality. 

I genuinely think overall it will make it easier for everyone, it will make functions shorter, easier to read thus easier to understand and easier to debug which in turn mean it is easier to get into and learn while with other overlays it adds a simple functionality that users do not need to touch but if they wanted to they don't have to recreate whole overlays.


Now to address @magicus's concerns:
First and foremost, I copied the code from Custom Bars, so anything regarding these changes should also apply to those too.
As for performance, I didn't run a profiler, so I don't know the specific data, it is more anecdotal, but I didn't notice any changes when playing. As for your suggestion to make it empty, I think it will cause an issue where people would think that you can't have these things displayed in the hub as it will give off the impression of basically no rules, but they can indeed work in the hub, I have a clock and fps info box that would work it there well, of course if you think it is not an issue, I can implement the change for Text Overlays and Custom Bars.

As for error handling, this is already an issue with Custom Bars as the Enabled Template doesn't have any feedback if there is an error, and it is also the case for Value Template. I think making it have some feedback it would require some additional work that I think would be out of scope of this PR as it is a completely different issue that is not exclusive to just this enabled template, perhaps we could include that with the book redesign to have the error be written under all fields that accept functions. But the error does not cause major issues, it just means it won't render, basically in order for it to render the template need to return `true` if anything else is returned it won't render including different types or errors.
Normally for info boxes and text template in custom bars, if the function have an error it will just return an error message

And in preview it always render no matter what, which makes sense, the preview is means to see how everything looks together, if things are not overlapping or are aligned so seeing these overlays is essential.

magicus:
Just want to add that I am vaguely (but maybe not firmly) in favor of this. If nothing else, it aligns with the custom bars.

It might be part of a bigger picture though. The functions we have are adequate for simple expressions, but when things start getting complicated it gets messy. Maybe we need a multi-line code editor for the functions? Or embedding a scripting language? Just throwing out ideas here.

I'm not saying we should integrate this PR as it is right now, but I think the concern it tries to address is valid -- that the function expressions can get too complex, and that one way they would be simpler if they could be split into two parts: what to display, and when/if to display it.

Rafii2198:
We just had a discussion about this pr on [Discord](https://ptb.discord.com/channels/394189072635133952/946471329857675307/1229776091879575572) and another person had similar thoughts, and it is from a perspective of someone fairly new to functions. Ideally, we would have an in game editor which would solve most issues, but this still improves writing functions a lot, instead of having to make a convoluted package of the display itself and the condition to display, these two things are now separated, making everything simple as you don't have to make tons of adaptations just to make the condition to work.
And again outside of info boxes it just adds more customizability with little to no drawback, definitely a lot more benefits in many places

kristofbolyai:
So, would we need this, if we had an editor?

Rafii2198:
It still adds customization to the rest of overlays, like all TextOverlays can be customized, not just info boxes. And for info boxes it lets people write less code that in turn is more understandable, like no matter how easy would writing functions be at the end of the day stuff like this makes it hard to get into as there is just much knowledge and thinking needed in order to begin, having an ability to split the function into 2 parts lets people tackle it one by one so not all at once and display does not have to be made in a way to fit the condition as they are 2 separate things.
This by itself makes getting into functions easier and offers universal customizability for other overlays, which also makes it easier to get into. Editor by itself can't solve everything

kristofbolyai:
Oh, this is directly TextOverlay? That's even more change than I anticipated. Anyways, I'll have to sleep on this a day or two, just to get my thought together.

Rafii2198:
Yes this affects all TextOverlays, my initial idea was just to add some customizability that Custom Bars have so that people do not need to recreate whole overlays just to have a simple condition to show them. As a side effect, this affected info boxes too as they are TextOverlay too, which made it so you no longer need to write really long functions to display simple info with a condition.

By having small bit of functions in many overlays, people can get into learning functions easily as they do not need to write whole function themselves but just a simple condition and for info boxes they don't have to force themselves to create really repetitive functions that are hard to read and easily make mistakes in, they can do a simple function and then do a simple condition, even if we had an editor, the function would still be long and repetitive it would only make it easier to read and prevent mistakes.

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
```
kristofbolyai, line 2: Did you change this yourself?

kristofbolyai: It should be 2023-2024.

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
```diff
@@ -127,4 +131,10 @@ private float getTextScale() {
 
     @Override
     protected void onConfigUpdate(Config<?> config) {}
+
+    public boolean isRendered() {
+        ErrorOr<Boolean> enabledOrError = Managers.Function.tryGetRawValueOfType(enabledTemplate.get(), Boolean.class);
+        return !enabledOrError.hasError() && enabledOrError.getValue();
+    }
+    ;
```
kristofbolyai:
```suggestion
```

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
kristofbolyai: The speedup I've done is mainly done by caching the function string -> function structure object calculation, so basically the onRendered call runs a `HashMap#get` and a string comparison, which should be fine.

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
Rafii2198: Oh, is it changed automatically? I had no clue, so I changed it myself

common/src/main/java/com/wynntils/overlays/custombars/CustomBarOverlayBase.java
```diff
@@ -19,7 +19,7 @@ public abstract class CustomBarOverlayBase extends BarOverlay {
     public final Config<String> valueTemplate = new Config<>("");
 
     @Persisted(i18nKey = "feature.wynntils.customBarsOverlay.overlay.customBarBase.enabledTemplate")
-    public final Config<String> enabledTemplate = new Config<>("string_equals(world_state;\"WORLD\")");
```
kristofbolyai, line 22: You should change the `isRendered` method here, the same way you did for `TextOverlay`

common/src/main/java/com/wynntils/core/consumers/overlays/TextOverlay.java
kristofbolyai: Yes. It's done by a formatting bot on GH, or the git hook the README mentions. Either way, no need to touch it. We are lazy programmers :)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -302,7 +302,7 @@
   "feature.wynntils.customBarsOverlay.description": "Adds the ability to add custom bars to the screen as overlays.",
   "feature.wynntils.customBarsOverlay.name": "Custom Bars",
   "feature.wynntils.customBarsOverlay.overlay.bubbleTexturedCustomBar.name": "Custom Bubble Bar %d",
-  "feature.wynntils.customBarsOverlay.overlay.customBarBase.enabledTemplate.description": "A template returning a boolean value, which determines whether the bar should be rendered or not. This template should not have curly brackets.",
+  "feature.wynntils.customBarsOverlay.overlay.customBarBase.enabledTemplate.description": "A template returning a boolean value, which determines whether this bar should be rendered or not. This template should not have curly brackets.",
```
```suggestion
  "feature.wynntils.customBarsOverlay.overlay.customBarBase.enabledTemplate.description": "A template returning a boolean value, which determines whether this bar should be rendered or not. This template should not have curly brackets. If empty, the overlay is going to render when the player is in a Wynncraft world.",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1781,6 +1781,8 @@
   "overlay.wynntils.overlay.userEnabled.name": "Enabled",
   "overlay.wynntils.overlay.verticalAlignmentOverride.description": "Set this if you want to override the position vertical alignment",
   "overlay.wynntils.overlay.verticalAlignmentOverride.name": "Vertical Alignment Override",
+  "overlay.wynntils.textOverlay.enabledTemplate.description": "A template returning a boolean value, which determines whether this overlay should be rendered or not. This template should not have curly brackets.",
```
```suggestion
  "overlay.wynntils.textOverlay.enabledTemplate.description": "A template returning a boolean value, which determines whether this overlay should be rendered or not. This template should not have curly brackets. If empty, the overlay is going to render when the player is in a Wynncraft world.",
```

magicus:
I think there were a lot of questions/ideas left dangling. We should probably create an issue to track them. (Sorry, too lazy/short of time to do that right now)

* What about previews? How does that work with enabled/disabled?
* Can we get better error handling/reporting for all our overlay scripts?
* Should we have a better code editor (multiline, allow line breaks in syntax?)
* Can we create some way for simple import/export of complete overlays (including both content and enable functions), so users can share them more easily? (We might even setup a repository..?)

kristofbolyai:
> * What about previews? How does that work with enabled/disabled?

Works as expected.

> * Can we get better error handling/reporting for all our overlay scripts?

Sure, if anyone finds time for it.

> * Should we have a better code editor (multiline, allow line breaks in syntax?)

Yes, we should. We can't force anyone to do UI work, I am sure it will happen eventually.

> * Can we create some way for simple import/export of complete overlays (including both content and enable functions), so users can share them more easily? (We might even setup a repository..?)

Not until the UI is done. Probably can be done in the same PR. We already have the Discord channel for sharing exports.

PR #2408, opened 2024-04-16
feat: Add RevealNicknamesFeature

kristofbolyai:
image


ShadowCat117 (approved):

kristofbolyai:
I am not merging this until I figure out a `StyledText` style-generation bug. It's really subtle, but can cause noticeable issues.
image

ShadowCat117:
Didn't even notice that, I don't think I experienced it when trying out the feature either

kristofbolyai:
> Didn't even notice that, I don't think I experienced it when trying out the feature either

Yeah, but I get it, and I dumped all sibling-style pairs and I can see that the produced style is different. However, I couldn't get further because of time constraints.

kristofbolyai:
Fixed version:

image

fabric/src/test/java/TestStyledText.java
```diff
@@ -530,8 +530,12 @@ public void styledText_shouldProduceCorrectComponent() {
 
         StyledText styledText = StyledText.fromComponent(component);
 
+        // The reconstructed componenet differs in that StyledText always adds components as a sibling to an empty
+        // component
         Assertions.assertEquals(
-                component, styledText.getComponent(), "StyledText.getComponent() returned an unexpected value.");
+                component.toFlatList(),
+                styledText.getComponent().toFlatList(),
+                "StyledText.getComponent() returned an unexpected value.");
```
kristofbolyai, line 538: This is not fully ideal, but at least we produce a more correct component (at least, it behaves like the original one, even if the structures differ).

fabric/src/test/java/TestStyledText.java
```diff
@@ -836,7 +840,7 @@ public void styledText_inheritesHoverEvents() {
 
         Assertions.assertEquals(
                 component.getStyle().getHoverEvent(),
-                styledText.getComponent().getStyle().getHoverEvent(),
+                styledText.getComponent().toFlatList().get(0).getStyle().getHoverEvent(),
```
kristofbolyai, line 843: This test was wrong.

magicus (approved):
I am not too worried about the structure change in the StyledText.

PR #2410, opened 2024-04-16
fix: Use proper url encoding [skip ci]

kristofbolyai:
The previous method encoded the string by the `application/x-www-form-urlencoded` standard, which differs from the HTML specification, so spaces were encoded as "+" instead of "%20" which caused issues when using V3 Wynncraft API.


ShadowCat117:
Is causing issues with opening the ??? quest on the wiki. Seems to work with everything else though

kristofbolyai:
This should work, @JamieCallan117.

ShadowCat117 (approved):
Works fine now, thanks for this :)

PR #2414, opened 2024-04-18
feat: Add ServerUptimeInfoOverlay

kristofbolyai:
Works by pressing the keybind for the player list, of by the enabled function the user sets.

image


common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
```diff
@@ -510,12 +512,14 @@ private void registerAllFunctions() {
         registerFunction(new LootrunFunctions.LootrunLastSelectedBeaconColorFunction());
         registerFunction(new LootrunFunctions.LootrunRedBeaconChallengeCountFunction());
         registerFunction(new LootrunFunctions.LootrunStateFunction());
+
```
ShadowCat117:
```suggestion
```

common/src/main/java/com/wynntils/functions/WorldFunctions.java
```diff
@@ -62,7 +62,44 @@ public String getValue(FunctionArguments arguments) {
 
         @Override
         protected List<String> getAliases() {
-            return List.of("world_uptime", "uptime");
```
ShadowCat117, line 65: Hmm if we're going to remove aliases I think we either need an upfixer to replace it or warn people when/before the feature goes live

kristofbolyai: Sadly, we don't have the ability to do this. A simple find and replace won't suffice for all cases, and can break "text" in info boxes. In a perfect world, we would have a function object -> string generator and edit the functions as objects, almost like the Filter UI does.

I'll make the new function's argument optional, and replace the null value there with the current world. That should "fix" this.

ShadowCat117 (approved):

PR #2415, opened 2024-04-20
feat: Add soulpoint subcommand and aliases for server commands, click-to-switch server texts, fix formatting issues, command no longer hands the game

kristofbolyai:
The timeouts were removed since the ServerListModel now automatically updates every 15 seconds, no need for the manual force updates now, and the running the command feels much faster now.

image
image
image

### DIFF
--- a/common/src/main/java/com/wynntils/commands/ServersCommand.java
+++ b/common/src/main/java/com/wynntils/commands/ServersCommand.java
@@ -47,6 +47,7 @@ public class ServersCommand extends Command {
 
         LiteralCommandNode<CommandSourceStack> infoBuilder = Commands.literal("info")
                 .then(Commands.argument("server", StringArgumentType.word()).executes(this::serverInfo))
+                .executes(this::serverInfo)
                 .build();
 
         LiteralCommandNode<CommandSourceStack> soulpointsBuilder = Commands.literal("soulpoints")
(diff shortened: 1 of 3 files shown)


ShadowCat117 (approved):
Looks good, one thing I will note if you want to fix it here there's been a bug for a while if you just run `/servers info` without an argument it sends you back to the lobby, I'm not quite sure what causes it

kristofbolyai:
> Looks good, one thing I will note if you want to fix it here there's been a bug for a while if you just run `/servers info` without an argument it sends you back to the lobby, I'm not quite sure what causes it

/servers is a Wynn command. So that, I guess.

ShadowCat117:
> /servers is a Wynn command. So that, I guess.

Ahh that would do it

kristofbolyai:
> > /servers is a Wynn command. So that, I guess.
> 
> Ahh that would do it

Fixed it anyways

magicus:
The updating every 15 seconds, were that done before? I can't figure it out from this PR. (Also, are we caching results in Athena or polling Wynncraft every 15 second?)

kristofbolyai:
> The updating every 15 seconds, were that done before? I can't figure it out from this PR. (Also, are we caching results in Athena or polling Wynncraft every 15 second?)

I recently introduced it, when I added a server uptime overlay. We use athena (for the better or for the worse, Athena performance is really bad recently..)

PR #2417, opened 2024-04-20
feat: Item Filter GUI

ShadowCat117:
Forgot to show StatValue widgets but they have a % checkbox next to them

https://github.com/Wynntils/Artemis/assets/8337467/fbebe7d4-7917-445c-ac66-65a69bf45973


magicus:
Maybe you should even break out the item annotator refactoring as a separate change?

kristofbolyai:
@JamieCallan117 Could you resolve the merge conflicts? After that, I can review this.

kristofbolyai:
It seems to me that you forgot to add the "remove" button for the filters :)

Also, I can't seem to do "any" for number/string types, not even by selecting single, then typing *.

common/src/main/java/com/wynntils/screens/itemfilter/type/FilterWidgetFactory.java
```diff
@@ -0,0 +1,241 @@
...
+import com.wynntils.screens.itemfilter.widgets.numeric.SingleStatValueFilterWidget;
+import com.wynntils.services.itemfilter.filters.PercentageStatFilter;
+import com.wynntils.services.itemfilter.filters.RangedStatFilters;
+import com.wynntils.services.itemfilter.type.StatProviderAndFilterPair;
+import com.wynntils.services.itemfilter.type.StatValue;
+import com.wynntils.utils.type.CappedValue;
+import java.util.Map;
+import java.util.function.Function;
+
+public final class FilterWidgetFactory {
```
kristofbolyai: Maybe rename to NumericFilterWIdgetFactory? I was looking for boolean/string types here. Or what if you moved them here?

ShadowCat117:
> It seems to me that you forgot to add the "remove" button for the filters :)
> 
> Also, I can't seem to do "any" for number/string types, not even by selecting single, then typing *.

I decided to not add a remove and instead just leaving it blank is enough but I could add it if you think it's necessary. As for the any filter, it's done by middle clicking the provider button, though I guess inputting * should also work

kristofbolyai (commented):
Not done with the code yet, but from what I've seen, it will be fine.

kristofbolyai:
> > It seems to me that you forgot to add the "remove" button for the filters :)
> > Also, I can't seem to do "any" for number/string types, not even by selecting single, then typing *.
> 
> I decided to not add a remove and instead just leaving it blank is enough but I could add it if you think it's necessary. As for the any filter, it's done by middle clicking the provider button, though I guess inputting * should also work

I think it is not obvious that you should leave it blank, so I think a remove button is a lot more obvious.

What if you added an "any" state for the filter type button? I think that is the most "obvious". Maybe even include a tooltip for that button, so users know what it is.

common/src/main/java/com/wynntils/screens/itemfilter/type/FilterWidgetFactory.java
ShadowCat117: String is in the `createFilterWidget` method, though that's only for the providers without a list of valid inputs.

Boolean didn't seem worth adding to it as it's literally just

```
if (provider.getType().equals(Boolean.class)) {
    if (filterPairs.isEmpty()) {
        widgets.add(new BooleanFilterWidget(null, this));
    } else {
        widgets.add(new BooleanFilterWidget(filterPairs.get(0), this));
    }
} 
```

ShadowCat117:
I've added this text for when there are no widgets for the current provider to hopefully make it more clear how to add an any filter. I don't really want to add another button for adding it when the provider button already does that and since this is a UI that's meant to make using the filter system easier, I think it's simpler to have that one button to do it, instead of select provider, click add filter, type "*".
image

common/src/main/java/com/wynntils/screens/itemfilter/type/FilterWidgetFactory.java
kristofbolyai: What do you think about the rename, then?

common/src/main/java/com/wynntils/screens/itemfilter/widgets/AnyFilterWidget.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public boolean mouseClicked(double mouseX, double mouseY, int button) {
+        if (removeButton.isMouseOver(mouseX, mouseY)) {
+            return removeButton.mouseClicked(mouseX, mouseY, button);
+        }
+
+        return false;
+    }
+
```
kristofbolyai, line 62:
```suggestion
@Override
```

kristofbolyai: Applies to every class.

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
```diff
@@ -255,6 +257,183 @@ public void renderAlignedTextInBox(
...
+                    switch (horizontalAlignment) {
+                        case CENTER -> -(textLength / 2f) + (renderWidth / 2);
+                        case RIGHT -> 0.0f;
+                        default -> textLength - renderWidth;
+                    };
+
+            double currentTimeInSeconds = (double) Util.getMillis() / 1000.0;
+            double e = Math.max((double) maxScrollOffset * 0.5, 3.0);
+            double f =
+                    Math.sin(1.5707963267948966 * Math.cos(6.283185307179586 * currentTimeInSeconds / e)) / 2.0 + 0.5;
```
kristofbolyai: This feels like Mojang/Mojmap code. Can you extract the big numbers into constants? They are `90 degs` and `360 degs` in radian. Also, can you name the fields to something that makes sense?

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
```diff
@@ -255,6 +257,183 @@ public void renderAlignedTextInBox(
                 1f);
     }
 
+    public void renderScrollingString(
```
kristofbolyai: Perhaps these methods should be put below the number renderText methods.

kristofbolyai (commented):
Finished the code review, not many changes left.

common/src/main/java/com/wynntils/screens/itemfilter/type/FilterWidgetFactory.java
ShadowCat117: Probably could do with a more specific name, not quite sure right now I'll give it some thought

common/src/main/java/com/wynntils/utils/render/FontRenderer.java
ShadowCat117: Yeah was copied from the default `renderScrollingString` missed those two

common/src/main/java/com/wynntils/screens/itemfilter/type/FilterWidgetFactory.java
ShadowCat117: I think it's best to just call it Numeric and take the string case out of it

magicus (approved):
I have mostly glanced through the UI code, but I have not noticed anything I reacted to. (But wait for bolyai's final review as well)

kristofbolyai (approved):
Looks good! But I found some changes/bugs I want to do/fix with filters, before we release this to the public.

We have two options here. Release this after my fixes, or combine this with the fuy.gg minor version change. Normally, I would have given this PR a minor version, but since the other PR is so close, I don't want to. I also don't want to hold back your work. So maybe the best is to release this (soon) and we'll just make the fuy.gg patch a minor version change, and include this PR in it's announcement. Although, if you just want to combine it with fuy.gg, it's fine by me. Your decision.

Anyways, please don't merge this PR, until I have time to PR my fixes.

magicus:
Release management woes! :-) Maybe we shold just stick to a single, monotonically increasing version number?

ShadowCat117:
I'm happy for this to go into the fuy branch since as you said it's so close and they do both make sense for a version change

kristofbolyai:
> Release management woes! :-) Maybe we shold just stick to a single, monotonically increasing version number?

I quite like our versioning now, once we actually get to use it. Major version for each MC version is really convenient.

kristofbolyai:
> I'm happy for this to go into the fuy branch since as you said it's so close and they do both make sense for a version change

Worst case scenario, if you find the raid features too extensive, we include them separately. That is, if you still plan to work on those. If not, I'll most likely not do them right now. Although, you'll have a bit to implement them, while I work on the remaining lootrun stats and, of course, the territory management ui.

ShadowCat117:
I still plan on doing the raid stuff, hopefully either later today or tomorrow. I may split up the features a bit though as some will take longer than others, so I'll just do a basic raid detection model first then timers and a mythic tome drystreak seperately. And also if we do want to include the showmemydps functionality I'll do that seperately too

kristofbolyai:
> I still plan on doing the raid stuff, hopefully either later today or tomorrow. I may split up the features a bit though as some will take longer than others, so I'll just do a basic raid detection model first then timers and a mythic tome drystreak seperately. And also if we do want to include the showmemydps functionality I'll do that seperately too

Showmemydps might be too much, we will see. I'll add the lootrun dry streaks as "statistics", we can probably use that for mythic boxes too.

PR #2418, opened 2024-04-21
feat: Introduce provider types for ItemStatProviders and SearchableContainers [skip ci]

kristofbolyai:
The filter types introduced here are debatable, but the main point of this PR is adding support for special filter types for specific containers (like territory management item filters, or guild member filters). However, I find it logical that containers know what provider types they want to use, and I prefer this over providers knowing what screens they "support". 

The types itself are only used by the helper widget now, and the filter UI, when it's finished. The ItemFilterService also filters the providers by the supported types.

### DIFF
--- a/common/src/main/java/com/wynntils/models/containers/type/wynncontainers/BlockBankContainer.java
+++ b/common/src/main/java/com/wynntils/models/containers/type/wynncontainers/BlockBankContainer.java
@@ -19,6 +19,6 @@ public class BlockBankContainer extends PersonalStorageContainer {
 
     @Override
     public List<ItemProviderType> supportedProviderTypes() {
-        return List.of(ItemProviderType.COUNTED);
+        return List.of();
     }
 }
--- a/common/src/main/java/com/wynntils/models/containers/type/wynncontainers/BlockBankContainer.java
+++ b/common/src/main/java/com/wynntils/models/containers/type/wynncontainers/BlockBankContainer.java
@@ -6,6 +6,8 @@ package com.wynntils.models.containers.type.wynncontainers;
 
 import com.wynntils.models.containers.type.PersonalStorageContainer;
 import com.wynntils.models.containers.type.PersonalStorageType;
+import com.wynntils.services.itemfilter.type.ItemProviderType;
+import java.util.List;
 import java.util.regex.Pattern;
 
 public class BlockBankContainer extends PersonalStorageContainer {
@@ -16,7 +18,7 @@ public class BlockBankContainer extends PersonalStorageContainer {
     }
 
     @Override
-    public boolean supportsAdvancedSearch() {
-        return false;
+    public List<ItemProviderType> supportedProviderTypes() {
+        return List.of(ItemProviderType.COUNTED);
     }
 }
--- a/common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
+++ b/common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.screens.base.widgets;
--- a/common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
+++ b/common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
+ * Copyright © Wynntils 2023-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.screens.guides.charm;
@@ -9,6 +9,7 @@ import com.wynntils.core.components.Models;
 import com.wynntils.core.components.Services;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.screens.guides.WynntilsGuideScreen;
+import com.wynntils.services.itemfilter.type.ItemFilterType;
 import com.wynntils.services.itemfilter.type.ItemSearchQuery;
 import com.wynntils.utils.colors.CommonColors;
 import com.wynntils.utils.render.FontRenderer;
@@ -30,7 +31,9 @@ public final class WynntilsCharmGuideScreen
     private List<GuideCharmItemStack> allCharmItems = List.of();
 
     private WynntilsCharmGuideScreen() {
-        super(Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"));
+        super(
+                Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"),
+                List.of(ItemFilterType.GEAR, ItemFilterType.GEAR_INSTANCE));
     }
 
     public static Screen create() {
--- a/common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
+++ b/common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
@@ -33,7 +33,7 @@ public final class WynntilsCharmGuideScreen
     private WynntilsCharmGuideScreen() {
         super(
                 Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"),
-                List.of(ItemProviderType.GEAR, ItemProviderType.GEAR_INSTANCE));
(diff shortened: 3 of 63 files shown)


common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
```
ShadowCat117, line 2: I presume the changes to this class were reverted?

common/src/main/java/com/wynntils/services/itemfilter/statproviders/SkillStatProvider.java
```diff
@@ -38,4 +39,10 @@ public List<Integer> getValue(WynnItem wynnItem) {
                 .map(StatActualValue::value)
                 .toList();
     }
+
+    @Override
+    public List<ItemFilterType> getFilterTypes() {
+        // Skill stats are either fixed in gear
```
ShadowCat117, line 45: Feels like there should be an "or" after this

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
kristofbolyai: Yeah, probably still in the commit history..

common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
```diff
@@ -30,7 +31,9 @@ public final class WynntilsCharmGuideScreen
     private List<GuideCharmItemStack> allCharmItems = List.of();
 
     private WynntilsCharmGuideScreen() {
-        super(Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"));
+        super(
+                Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"),
+                List.of(ItemFilterType.GEAR, ItemFilterType.GEAR_INSTANCE));
```
ShadowCat117: All guides should also use GENERIC

common/src/main/java/com/wynntils/services/itemfilter/statproviders/SkillStatProvider.java
kristofbolyai: I thought crafted items were "identifiable", but they are not. That was the or part

common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
kristofbolyai: Why? I think you don't need more than I listed.

common/src/main/java/com/wynntils/screens/guides/ingredient/WynntilsIngredientGuideScreen.java
```diff
@@ -31,7 +32,9 @@ public final class WynntilsIngredientGuideScreen
     private List<GuideIngredientItemStack> allIngredientItems = List.of();
 
     private WynntilsIngredientGuideScreen() {
-        super(Component.translatable("screens.wynntils.wynntilsGuides.ingredientGuide.name"));
+        super(
+                Component.translatable("screens.wynntils.wynntilsGuides.ingredientGuide.name"),
+                List.of(ItemProviderType.INGREDIENT));
```
ShadowCat117: And this should use professions

common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
ShadowCat117: Generic includes level and favourite

common/src/main/java/com/wynntils/models/containers/type/wynncontainers/BlockBankContainer.java
```diff
@@ -16,7 +18,7 @@ public BlockBankContainer() {
     }
 
     @Override
-    public boolean supportsAdvancedSearch() {
-        return false;
+    public List<ItemProviderType> supportedProviderTypes() {
+        return List.of(ItemProviderType.COUNTED);
```
ShadowCat117: We don't parse any items in the block bank to WynnItem's so unless you're planning to do that in the future I don't think this is needed

kristofbolyai:
https://github.com/Wynntils/Artemis/pull/2418/commits/a703369c49ed21bb862691ee44f16bd6a38aca83 @JamieCallan117

ShadowCat117:
> [a703369](https://github.com/Wynntils/Artemis/commit/a703369c49ed21bb862691ee44f16bd6a38aca83) @JamieCallan117

Yup, those look good

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
magicus: But you never reverted the year change..?

kristofbolyai: I can't really do that, without triggering spotless, or doing a force push. I can revert it by force-pushing main, if you want to.

common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
```diff
@@ -30,7 +31,9 @@ public final class WynntilsCharmGuideScreen
     private List<GuideCharmItemStack> allCharmItems = List.of();
 
     private WynntilsCharmGuideScreen() {
-        super(Component.translatable("screens.wynntils.wynntilsGuides.charmGuide.name"));
+        super(
```
magicus, line 34: This formatting looks really odd! Is this the doing of Spotless?

kristofbolyai: It is. There are a few instances, where it is not perfect...

https://github.com/Wynntils/Artemis/blob/main/common/src/main/java/com/wynntils/utils/render/buffered/BufferedFontRenderer.java#L68

common/src/main/java/com/wynntils/services/itemfilter/statproviders/SkillStatProvider.java
magicus: Yeah, but the comment still reads funny. 
```suggestion
        // Skill stats are fixed in gear
```
or remove the comment completely, since it is unclear if it says anything the code does not say already.

common/src/main/java/com/wynntils/screens/guides/charm/WynntilsCharmGuideScreen.java
magicus: *sigh* :-(

common/src/main/java/com/wynntils/services/itemfilter/statproviders/SkillStatProvider.java
kristofbolyai: Right, I missed that last night. I'll just do a force push with the year change then..

magicus:
Sorry I was a bit late in reviewing this. :(

Overall I like the approach of "inverting" the knowledge about the filters. I think we have managed to build a new quite good abstraction around containers. The only thing I'm not so sure about is the renaming to "AbstractWynncraftContainer". Instead of making the name longer, I think we should go the other way, and just make it "Container". Just as "Feature" is the base class for all features, "Container" should be the base class for all our container implementations. The word is specific enough in our context to use it as a term for "Wynncraft containers".

kristofbolyai:
> Overall I like the approach of "inverting" the knowledge about the filters. I think we have managed to build a new quite good abstraction around containers. The only thing I'm not so sure about is the renaming to "AbstractWynncraftContainer". Instead of making the name longer, I think we should go the other way, and just make it "Container". Just as "Feature" is the base class for all features, "Container" should be the base class for all our container implementations. The word is specific enough in our context to use it as a term for "Wynncraft containers".

Sure. I can also do that.

magicus:
Actually, let me fix it. I also want to do some more design changes to `models/containers`.

kristofbolyai:
> Actually, let me fix it. I also want to do some more design changes to `models/containers`.

Your change then, I am off until afternoon anyways.

common/src/main/java/com/wynntils/screens/base/widgets/SearchWidget.java
magicus: I see the problem. Too much effort, we'll have to live with this.

PR #2419, opened 2024-04-22
refactor: Container design rehaul [skip ci]

magicus:
The container model has grown organically for some while now. It has reached maturity now, and some finishing design changes was needed.

I have:
1) enforced our rule of "type" only containing interfaces, enums and records; and hence moved all container implementations to a new "containers" package
2) renamed the base class to "Container"
3) Removed the ContainerProperty which was unused
4) Converted ContainerBonds to a record
5) Made the award and personal storage container base classes abstract, and sorted out them and their subclasses in separate packages
6) Separated out inventory handling to a separate `models.inventory` and renamed the model `InventoryModel`


common/src/main/java/com/wynntils/models/containers/type/ContainerBounds.java
```diff
@@ -46,6 +40,6 @@ private boolean isInBounds(int slot) {
      * @return A list of all non-UI slots in the container
      */
     public List<Integer> getSlots() {
-        return Collections.unmodifiableList(slots);
+        return slots;
```
ShadowCat117, line 43: Shouldn't this still be unmodifiable?

magicus: Yes, it still is. It is created as a final unmodifiable in line 30.

ShadowCat117: Ah yes, I missed that

ShadowCat117 (approved):

magicus:
That was quick! Maybe @kristofbolyai should have had a chance to take a look at it as well? 

@kristofbolyai Let me know if you have any comments and I'll address them in a follow-up.

kristofbolyai:
> That was quick! Maybe @kristofbolyai should have had a chance to take a look at it as well?
> 
> @kristofbolyai Let me know if you have any comments and I'll address them in a follow-up.

Looks good to me overall! I've had a chance to glance through it on phone, before merging, but I did not want to approve until I checked on PC. Now that I've done so, I can say, it looks good to me.

PR #2421, opened 2024-04-22
feat: Parse territory items, add territory item properties, advanced search for territory list [skip ci]

kristofbolyai:
Before adding the Custom UI for the territory list, I added everything needed for advanced filters for territories, so it can be used regardless of the custom screen.


common/src/main/java/com/wynntils/models/items/annotators/gui/TerritoryAnnotator.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Collections;
+import java.util.Deque;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.minecraft.world.item.ItemStack;
+
+public class TerritoryAnnotator implements ItemAnnotator {
```
ShadowCat117: This can be a GuiItemAnnotator now

kristofbolyai: Ah. I changed the item, I didn't know annotators were separate too..

common/src/main/java/com/wynntils/services/itemfilter/statproviders/territory/TerritoryNameStatProvider.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.services.itemfilter.statproviders.territory;
+
+import com.wynntils.models.items.items.gui.TerritoryItem;
+import java.util.List;
+
+public class TerritoryNameStatProvider extends TerritoryStatProvider<String> {
```
ShadowCat117, line 10: Is this necessary for the ui? If not, does the plain query not handle the name  in some cases?

kristofbolyai: The real search would also search for the (HQ)/[!] parts. Your question is good though, Ill think about it until tomorrow.

common/src/main/java/com/wynntils/services/itemfilter/type/ItemProviderType.java
```diff
@@ -63,5 +62,40 @@ public enum ItemProviderType {
...
+    PROFESSION,
+
+    /**
+     * Special filter for territory items.
+     */
+    TERRITORY(true);
+
+    /**
+     * Whether this item provider type is special.
+     * Special providers are not included in {@link ItemProviderType#ALL}.
```
ShadowCat117: ALL has been removed, perhaps "not included in normalTypes()"

common/src/main/java/com/wynntils/models/territories/type/GuildResource.java
```diff
@@ -1,17 +1,17 @@
...
 public enum GuildResource {
-    EMERALD(ChatFormatting.GREEN, "Emeralds", ""),
-    ORE(ChatFormatting.WHITE, "Ores", "Ⓑ"),
+    EMERALDS(ChatFormatting.GREEN, "Emeralds", ""),
+    ORE(ChatFormatting.WHITE, "Ore", "Ⓑ"),
     WOOD(ChatFormatting.GOLD, "Wood", "Ⓒ"),
-    FISH(ChatFormatting.AQUA, "Fishes", "Ⓚ"),
-    CROPS(ChatFormatting.YELLOW, "Crops", "Ⓙ");
+    FISH(ChatFormatting.AQUA, "Fish", "Ⓚ"),
+    CROP(ChatFormatting.YELLOW, "Crop", "Ⓙ");
```
magicus, line 14: Being a guid noob, is this some well-established order of resources? (Seeing as you reordered it).

kristofbolyai: Didn't reorder it, but just renamed the enums/strings so it matches ingame. Emeralds is plural, everything else is not. The order is the same as every in-game tooltip, but it has not changed.

common/src/main/java/com/wynntils/services/itemfilter/type/ItemProviderType.java
```diff
@@ -63,5 +62,40 @@ public enum ItemProviderType {
...
+     */
+    private final boolean isSpecial;
+
+    ItemProviderType() {
+        this(false);
+    }
+
+    ItemProviderType(boolean isSpecial) {
+        this.isSpecial = isSpecial;
```
magicus: "special" makes me nervous. Why do you need this? Is it just so the personal storage container and guild bank should not have the territory filter?

kristofbolyai: Nothing, except the territory menus should have them.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/territory/TerritoryNameStatProvider.java
kristofbolyai: The normal searching will work for the names, but this provider will also provide an exact name. I don't mind removing it, but also don't mind keeping it..

common/src/main/java/com/wynntils/models/territories/type/GuildResource.java
magicus: Ah, I see, the diff looked like a reordering. :) Should we have this alphabetical instead, or is the in-game tooltip to be considered canonical?

common/src/main/java/com/wynntils/services/itemfilter/type/ItemProviderType.java
magicus: Can we solve this without introducing the concept of "special"? (It is a very vague concept)

Perhaps just add an explicit filter-out in `normalTypes()`, and let that method "know" about what is normal, and conversely, rename `specialTypes` to `territoryTypes` and let it "know" about which type(s) to return?

common/src/main/java/com/wynntils/services/itemfilter/statproviders/territory/TerritoryNameStatProvider.java
ShadowCat117: Ah I see, are you wanting to have the names in the valid inputs list the filter ui uses? I see no reason to remove it if that's the case

common/src/main/java/com/wynntils/models/territories/type/GuildResource.java
kristofbolyai: > Ah, I see, the diff looked like a reordering. :) Should we have this alphabetical instead, or is the in-game tooltip to be considered canonical?

It's most definitely should be like the in-game tooltip, we use this for reconstructing territory-like tooltip in multiple places.

common/src/main/java/com/wynntils/services/itemfilter/statproviders/territory/TerritoryNameStatProvider.java
kristofbolyai: > Ah I see, are you wanting to have the names in the valid inputs list the filter ui uses? I see no reason to remove it if that's the case

Perhaps we could use it like that, yes.

kristofbolyai:
@magicus I did your reviews, and changed the target to main, so we include the ItemProviderType changes in main. I'll do the UI in fuygg branch, but I think this can go into main now (with the user info fix).

kristofbolyai:
Anyways, I'll merge this now, so I can release the other fix. Please let me know if you have any reviews, and I'll make sure to include it in the fuy_gg branch.

PR #2422, opened 2024-04-23
fix: Make Wynntils much smarter about user info requests to greatly reduce Athena load

kristofbolyai:
## Important
Please review this PR ASAP, but let me merge it, and release it with an announcement. From the limited logs we have about Athena, we think that the major performance hit Athena's been taking recently is due to this route. This PR should reduce requests by 2/3x, meaning, we should get much, much less requests per minute (I don't have exact numbers, but thousands less..).

## The changes

This PR should make error handling of user requests much better, fix some bugs where players got stuck in a fetching state, and make sure that we do the least amount of requests possible. 

Here is the data I recorded pre- and post- fix, it can be seen that we load the same amount of users, in around 2/3 times less requests. The duplication rate is also 0, meaning we don't do any unnecessary calls, for failed users, but we also don't refetch user data we already have.

```
Pre-fix:

[10:15:21] [Render thread/INFO] (wynntils) PlayerModel: 19 users loaded, 0 ghosts, 0 errors, 0 user failures.
[10:15:21] [Render thread/INFO] (wynntils) PlayerModel: 165 fetches, 111 unique users, duplication rate: 0.49x

[10:16:21] [Render thread/INFO] (wynntils) PlayerModel: 28 users loaded, 27 ghosts, 0 errors, 0 user failures.
[10:16:21] [Render thread/INFO] (wynntils) PlayerModel: 241 fetches, 145 unique users, duplication rate: 0.66x

Post-fix:

[10:10:43] [Render thread/INFO] (wynntils) PlayerModel: 15 users loaded, 8 ghosts, 0 errors, 0 user failures.
[10:10:43] [Render thread/INFO] (wynntils) PlayerModel: 54 fetches, 53 unique users, duplication rate: 0.02x

[10:11:50] [Render thread/INFO] (wynntils) PlayerModel: 29 users loaded, 29 ghosts, 0 errors, 0 user failures.
[10:11:50] [Render thread/INFO] (wynntils) PlayerModel: 120 fetches, 120 unique users, duplication rate: 0.0x
```

#### What the data means
users loaded = wynntils user profiles loaded
ghosts = player ghosts we found, irrelevant to athena
fetches = number of requests to athena
unique users = loaded users that either have an wynntils profile, or not

## Left to do
We still run some requests for NPC players. Sadly, I can see no way of filtering them out currently. I'll think about if we can fix that issue, but this change should be enough for now.

#### Edit
We now have a way to filter out NPCs, starting Wynncraft 2.1. The code is already implemented here, for the future.

### DIFF
--- a/common/src/main/java/com/wynntils/models/players/PlayerModel.java
+++ b/common/src/main/java/com/wynntils/models/players/PlayerModel.java
@@ -33,6 +33,7 @@ import net.minecraft.world.scores.PlayerTeam;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class PlayerModel extends Model {
+    private static final String ATHENA_USER_NOT_FOUND = "User not found";
     private static final Pattern GHOST_WORLD_PATTERN = Pattern.compile("^_(\\d+)$");
 
     // If there is a failure with the API, give it time to recover
@@ -46,6 +47,7 @@ public final class PlayerModel extends Model {
     private static final int MAX_USER_ERRORS = 3;
 
     private final Map<UUID, WynntilsUser> users = new ConcurrentHashMap<>();
+    private final Set<UUID> usersWithoutWynntilsAccount = ConcurrentHashMap.newKeySet();
     private final Set<UUID> fetching = ConcurrentHashMap.newKeySet();
     private final Map<UUID, Integer> ghosts = new ConcurrentHashMap<>();
     private final Map<UUID, String> nameMap = new ConcurrentHashMap<>();
@@ -139,7 +141,7 @@ public final class PlayerModel extends Model {
     private void loadUser(UUID uuid, String userName) {
         // Avoid fetching the same user multiple times
         if (fetching.contains(uuid)) return;
-        if (users.containsKey(uuid)) return;
+        if (users.containsKey(uuid) || usersWithoutWynntilsAccount.contains(uuid)) return;
 
         if (errors.size() >= MAX_ERRORS) {
             // Athena is having problems, skip this
@@ -157,6 +159,13 @@ public final class PlayerModel extends Model {
         ApiResponse apiResponse = Managers.Net.callApi(UrlId.API_ATHENA_USER_INFO, Map.of("uuid", uuid.toString()));
         apiResponse.handleJsonObject(
                 json -> {
+                    if (json.has("message") && json.get("message").getAsString().equals(ATHENA_USER_NOT_FOUND)) {
+                        // This user does not exist in our database, stop requesting it
+                        fetching.remove(uuid);
+                        usersWithoutWynntilsAccount.add(uuid);
+                        return;
+                    }
+
                     if (!json.has("user")) return;
 
                     WynntilsUser user = WynntilsMod.GSON.fromJson(json.getAsJsonObject("user"), WynntilsUser.class);
@@ -189,7 +198,9 @@ public final class PlayerModel extends Model {
     }
 
     private void clearUserCache() {
+        fetching.clear();
         users.clear();
+        usersWithoutWynntilsAccount.clear();
         nameMap.clear();
     }
 
--- a/common/src/main/java/com/wynntils/models/players/PlayerModel.java
+++ b/common/src/main/java/com/wynntils/models/players/PlayerModel.java
@@ -52,7 +52,8 @@ public final class PlayerModel extends Model {
     private final Map<UUID, Integer> ghosts = new ConcurrentHashMap<>();
     private final Map<UUID, String> nameMap = new ConcurrentHashMap<>();
 
-    // Count of errors in the last minute, to avoid spamming the API
+    // The size of this set is the count of errors in the last ERROR_TIMEOUT_MINUTE minutes.
+    // This is used to avoid spamming the API.
     private final TimedSet<Object> errors =
             new TimedSet<>(ERROR_TIMEOUT_MINUTE, TimeUnit.MINUTES, true, ConcurrentHashMap::newKeySet);
     private final Map<UUID, Integer> userFailures = new ConcurrentHashMap<>();
@@ -161,8 +162,8 @@ public final class PlayerModel extends Model {
                 json -> {
                     if (json.has("message") && json.get("message").getAsString().equals(ATHENA_USER_NOT_FOUND)) {
                         // This user does not exist in our database, stop requesting it
-                        fetching.remove(uuid);
                         usersWithoutWynntilsAccount.add(uuid);
+                        fetching.remove(uuid);
                         return;
                     }
 
--- a/common/src/main/java/com/wynntils/models/players/PlayerModel.java
+++ b/common/src/main/java/com/wynntils/models/players/PlayerModel.java
@@ -71,13 +71,13 @@ public final class PlayerModel extends Model {
 
     public boolean isLocalPlayer(String name) {
         // Wynn uses TeamNames for player names that are online
-        return !isNpc(StyledText.fromString(name))
+        return !isNpcName(StyledText.fromString(name))
                 && McUtils.mc().level.getScoreboard().getTeamNames().contains(name);
     }
 
     public boolean isNpc(Player player) {
         StyledText scoreboardName = StyledText.fromString(player.getScoreboardName());
-        return isNpc(scoreboardName);
+        return isNpcName(scoreboardName) || isNpcUuid(player.getUUID());
     }
 
     public boolean isPlayerGhost(Player player) {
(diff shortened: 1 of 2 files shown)


DevScyu (approved):

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -120,18 +139,33 @@ public void onAddPlayerToTeam(PlayerTeamEvent.Added event) {
...
         fetching.add(uuid); // temporary, avoid extra loads
         nameMap.put(uuid, userName);
 
         ApiResponse apiResponse = Managers.Net.callApi(UrlId.API_ATHENA_USER_INFO, Map.of("uuid", uuid.toString()));
         apiResponse.handleJsonObject(
                 json -> {
+                    if (json.has("message") && json.get("message").getAsString().equals(ATHENA_USER_NOT_FOUND)) {
+                        // This user does not exist in our database, stop requesting it
+                        fetching.remove(uuid);
+                        usersWithoutWynntilsAccount.add(uuid);
```
magicus:
```suggestion
                        usersWithoutWynntilsAccount.add(uuid);
                        fetching.remove(uuid);
```

Will slightly help with a possible race condition. (Maybe it can't happen, but I like to think in code this way)

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -120,18 +139,33 @@ public void onAddPlayerToTeam(PlayerTeamEvent.Added event) {
...
+                        return;
+                    }
+
                     if (!json.has("user")) return;
```
magicus: If this happens maybe we should add it to usersWithoutWynntilsAccount too? Or handle it as a athena-user-error?

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -143,15 +177,30 @@ private void loadUser(UUID uuid, String userName) {
...
-                    if (errorCount >= MAX_ERRORS) {
-                        WynntilsMod.error("Athena user lookup has repeating failures. Disabling future lookups.");
+                    errors.put(System.currentTimeMillis());
+
+                    // Save user failures
+                    userFailures.computeIfAbsent(uuid, k -> 0);
+                    userFailures.compute(uuid, (k, v) -> v + 1);
+
+                    // Only log the error once
+                    if (errors.size() == MAX_ERRORS + 1) {
```
magicus: Can this happen? If errors.size() == MAX_ERRORS we will exit early. (I think the `+1` needs to go)

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -143,15 +177,30 @@ private void loadUser(UUID uuid, String userName) {
...
+                        WynntilsMod.error(
+                                "Athena user lookup has repeating failures. Disabling future lookups temporarily.");
+                    }
+
+                    // Only log the error once
+                    if (userFailures.get(uuid) == MAX_USER_ERRORS + 1) {
```
magicus: Same here.

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -31,24 +33,39 @@
...
+    private static final int MAX_USER_ERRORS = 3;
+
     private final Map<UUID, WynntilsUser> users = new ConcurrentHashMap<>();
+    private final Set<UUID> usersWithoutWynntilsAccount = ConcurrentHashMap.newKeySet();
     private final Set<UUID> fetching = ConcurrentHashMap.newKeySet();
     private final Map<UUID, Integer> ghosts = new ConcurrentHashMap<>();
-    private int errorCount;
     private final Map<UUID, String> nameMap = new ConcurrentHashMap<>();
 
+    // Count of errors in the last minute, to avoid spamming the API
```
magicus:
```suggestion
    // The size of this set is the count of errors in the last ERROR_TIMEOUT_MINUTE minutes.
    // This is used to avoid spamming the API.
```

magicus:
Remember that `TimedSet` is a bit of a misnomer. Or rather, there is no automatic background purging, instead we check if items has timed out at various operations. I am not sure that calling `errors.size()` is one of those. If I am correct, the set will never shrink once it hits its limit.

Have you tested this? (E.g. by faking errors from Athena)

To be on the safe side, we should probably add a `refresh()` (or whatever) method to TimedSet, and call it before checking the size.

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -75,13 +92,15 @@ public Stream<String> getAllPlayerNames() {
...
     public void reset() {
-        errorCount = 0;
+        errors.clear();
+        userFailures.clear();
     }
 
     @SubscribeEvent
     public void onWorldStateChange(WorldStateEvent event) {
         if (event.getNewState() == WorldState.NOT_CONNECTED) {
             clearUserCache();
```
magicus: Do we really need to do this? I understand we need to clear the ghost cache if we change world, but the user cache (and the without-account cache) are about what information Athena has about a player, and is independent on the world.

kristofbolyai:
> Remember that `TimedSet` is a bit of a misnomer. Or rather, there is no automatic background purging, instead we check if items has timed out at various operations. I am not sure that calling `errors.size()` is one of those. If I am correct, the set will never shrink once it hits its limit.
> 
> 
> 
> Have you tested this? (E.g. by faking errors from Athena)
> 
> 
> 
> To be on the safe side, we should probably add a `refresh()` (or whatever) method to TimedSet, and call it before checking the size.

TimedSet does have the option to auto-purge on calls, if enabled. This case, it is. I'll double check, though. 

Yes, I tested responses from Athena (and also tested on mobile data, to emulate really unstable connections).

common/src/main/java/com/wynntils/models/players/PlayerModel.java
kristofbolyai: Thought about this. Historically we cleared it so the users get fresh cosmetics without full restarts. Maybe we should just require a restart...

common/src/main/java/com/wynntils/models/players/PlayerModel.java
kristofbolyai: If this happens, it's probably some other error. We should remove the user from fetching though, and count it as an error..

magicus:
> TimedSet does have the option to auto-purge on calls, if enabled

If I understand the code correctly, this only purges if you create an iterator. So it has worked fine for our use cases up until now, but I don't think it will work for size().

The *proper* solution is probably to wrap all calls exposed by Set, and make sure we purge before all of them. I don't think it is worth it, though, and a short-cut with providing an explicit way to purge and letting the programmer know when to do that would be good enough.

common/src/main/java/com/wynntils/models/players/PlayerModel.java
kristofbolyai: Fixed.

common/src/main/java/com/wynntils/models/players/PlayerModel.java
kristofbolyai: Changed it so we don't clear it.

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -31,35 +33,51 @@
...
         // Wynn uses TeamNames for player names that are online
-        return !isNpc(StyledText.fromString(name))
+        return !isNpcName(StyledText.fromString(name))
                 && McUtils.mc().level.getScoreboard().getTeamNames().contains(name);
     }
 
     public boolean isNpc(Player player) {
         StyledText scoreboardName = StyledText.fromString(player.getScoreboardName());
-        return isNpc(scoreboardName);
+        return isNpcName(scoreboardName) || isNpcUuid(player.getUUID());
```
magicus, line 80: Is not isNpcUuid enough?

kristofbolyai: After Wynn v2.1, it is. Not right now.

magicus: I see. I've lost track of the Wynn versioning scheme.

kristofbolyai:
Fixed the TimedSet issues now. Anything left, or can I merge this, when I get home, @magicus?

magicus (approved):
I think this looks good now. Thanks!

magicus:
If you announce this, you might also want to add a note that the zoom level for the minimap config is changed and will be lost if it was customized.

PR #2423, opened 2024-04-23
fix: Fix minimap disabling/crashing when changing the minimap's configs

kristofbolyai:
The overlook happened from me when reviewing https://github.com/Wynntils/Artemis/pull/2373/commits/f5101064c4b0ca80532b9d93fc5632e3151760a1. The check to make sure we don't recurse in the configs were removed, because I did not explicitly comment why it was there in the first place, and @magicus thought it could be clean up.


magicus:
Ah, I see. Should have thought of that. Sorry. :(

PR #2424, opened 2024-04-23
feat: Add ProfessionHighlightFeature

kristofbolyai:
Although this feature is not in fuy.gg, but Wynntils Legacy/avomod2, I just added it due to a request. I've improved the original version with a per-container config option.

image


common/src/main/java/com/wynntils/models/containers/containers/ProfessionCraftingContainer.java
```diff
@@ -0,0 +1,31 @@
...
+package com.wynntils.models.containers.containers;
+
+import com.wynntils.models.containers.Container;
+import com.wynntils.models.containers.type.ContainerBounds;
+import com.wynntils.models.containers.type.HighlightableProfessionProperty;
+import com.wynntils.models.profession.type.ProfessionType;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+public class ProfessionCraftingContainer extends Container implements HighlightableProfessionProperty {
```
ShadowCat117, line 14: Crafting containers were one of the types I thought could be added, but didn't serve a purpose at the time so I left them out. I was thinking more of having an abstract CraftingStation like PersonalStorage and Reward, then each crafting container having an associated ProfessionType. That would contain a list of the material and ingredient slots for anything we would need them for.

If you want to leave it like this for now that's fine as what I suggested doesn't serve a purpose yet, unless you wanted to modify this to say on the crafting screen it would instead become a toggle to only highlight ingredients in your inventory that are relevant to the current crafting station. So if I open the alchemism station, I would toggle the button to highlight alchemism ingredients

common/src/main/java/com/wynntils/models/containers/containers/ProfessionCraftingContainer.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public ProfessionCraftingContainer() {
+        super(TITLE_PATTERN);
+    }
+
+    @Override
+    public ContainerBounds getBounds() {
+        // Technically this includes more than just the ingredient slots,
+        // but ContainerBounds can't select only a subset of slots.
+        return new ContainerBounds(0, 0, 2, 4);
```
ShadowCat117: I don't see much point in highlighting ingredients already placed into the crafting station, it makes more sense to me for the inventory to be highlighted

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
```diff
@@ -0,0 +1,229 @@
...
+    private final Storage<Map<String, ProfessionType>> selectionPerContainer = new Storage<>(new TreeMap<>());
+
+    @SubscribeEvent
+    public void onScreenInit(ScreenInitEvent event) {
+        Screen screen = event.getScreen();
+
+        if (!(screen instanceof AbstractContainerScreen<?> containerScreen)) return;
+
+        if (!(Models.Container.getCurrentContainer()
+                instanceof HighlightableProfessionProperty highlightableProfessionProperty)) {
```
ShadowCat117, line 59:
```suggestion
        if (!(Models.Container.getCurrentContainer() instanceof HighlightableProfessionProperty)) {
```

ShadowCat117:
I would also suggest to either make `PersonalStorageContainer` implement `HighlightableProfessionProperty` or make account/character bank and misc bucket implement it

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
kristofbolyai: Spotless won't allow this :c

ShadowCat117: :(

common/src/main/java/com/wynntils/models/containers/containers/ProfessionCraftingContainer.java
kristofbolyai: I won't complicate it further, until we have an idea what to use it for. Made the inventory highlighted though.

ShadowCat117 (approved):
Looks better now. I may look into the other changes I suggested after doing the raid stuff for the fuy features but this is good for now :)

common/src/main/java/com/wynntils/models/ingredients/IngredientModel.java
```diff
@@ -23,8 +24,8 @@ public class IngredientModel extends Model {
             ChatFormatting.AQUA, 3);
     private final IngredientInfoRegistry ingredientInfoRegistry = new IngredientInfoRegistry();
 
-    public IngredientModel() {
-        super(List.of());
+    public IngredientModel(WynnItemModel wynnItem) {
```
magicus, line 30: Why do you need this? (Same for the RewardModel)

kristofbolyai: Oh, yeah! I forgot to comment for this here. I planned to, though :). Anyways, `xInfoRegistry` (Ingredient/Charm/Tome) uses `WynnItem` models methods for converting item material to item stack. `IngredientModel`'s already been depending on `WynnItem` implicitly, by only loading ingredients after loading material conversions and item obtain data. https://github.com/Wynntils/Artemis/pull/2424/commits/48eafc1efd6299377fb6ae624836c30fb2771ae5 should clear this up even further.

kristofbolyai: As to why this was fixed now, it's because I saw a log where this happened. This is a really rare race case, so it's not been a problem before, but it should be stable now.

magicus: I see. Yeah, that was a tricky one. Thanks for adding the description.

kristofbolyai: ~~Actually, this is still race-prone... If the materials load before the ingredients, they won't load... I'll properly patch it here~~, and I'll work on a data dependency system later... We could make dependency graphs and use parallelism for many requests.

EDIT: It's not actually race prone, but I want to work on a better dependency system regardless...

magicus (approved):

PR #2425, opened 2024-04-23
fix: Fix crops resource name causing TerritoryAnnotator to crash [skip ci]

ShadowCat117:
Snippet of the problematic item log `"text":"Ⓙ +4320 Crops per Hour"`


kristofbolyai:
To make sure, do the other ones match..? I've thought I've checked, but it's unfortunate that we added this at off-season times.

ShadowCat117:
My guild only has an emerald and a crops territory right now so I can only confirm those

kristofbolyai:
Should be alright.

image

kristofbolyai (approved):

PR #2426, opened 2024-04-23
chore: Update mod dependency and ci versions [skip ci]

kristofbolyai:
The mod dependency versions fix a few things, like the new `palantir-java-format` supporting modern switch cases better. DevAuth fixes an issue with OAuth2 being very frequent.

On the CI side, I've updated every action, so we only use the Node 20 stack. I don't except anything to break, but we won't know for sure, until we release a version. I've also fully fixed up the alpha branch script and ported the changes from the release script to it.


.github/workflows/release.yml
```diff
@@ -153,7 +153,7 @@ jobs:
     needs: [build, changelog]
     steps:
       - name: Publish Release
-        uses: eregon/publish-release@v1
+        uses:  eregon/publish-release@v1.0.6
```
kristofbolyai:
```suggestion
        uses: eregon/publish-release@v1.0.6
```

.github/workflows/alpha-release.yml
```diff
@@ -16,77 +16,82 @@ jobs:
       changelog: ${{ steps.changelog.outputs.changelog }}
       release_id: ${{ steps.release.outputs.id }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4.1.3
```
magicus: Can't we just specify the major version?

magicus: When checking more I see that you have specified patch release version numbers for all actions. Is you idea to make the build more stable and not risk breakage if any of the actions change? Otherwise I think the common usage is to only specify major version for GHA actions.

kristofbolyai: I did this at night, and just completely forgot about the fact that we can use major versions only. Pushed a patch.

.github/workflows/alpha-release.yml
```diff
@@ -16,77 +16,82 @@ jobs:
...
           git-user-name: 'WynntilsBot'
           git-user-email: 'admin@wynntils.com'
           pre-commit: ./.github/.pre-commit.js
           config-file-path: ./.github/.config.js
+          version-file: ./version.json
           skip-version-file: true
           skip-git-pull: true
           pre-release: true
           pre-release-identifier: alpha
+          release-count: 5
```
magicus, line 46: What does this mean?

.github/workflows/alpha-release.yml
```diff
@@ -16,77 +16,82 @@ jobs:
...
       release_id: ${{ steps.release.outputs.id }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
           token: ${{ secrets.PRIVATE_TOKEN }}
 
-      - uses: actions/setup-node@v3
+      - name: Set up version.json
```
magicus: If this is used by the changelog, maybe move it so we do the installations first, and then do this just before the changelog step. I stared at this for a long time trying to figure out why node would need this...

.github/workflows/alpha-release.yml
kristofbolyai: It's copied from release, it's basically how many releases it keeps in CHANGELOG.md

.github/workflows/alpha-release.yml
```diff
@@ -100,10 +105,10 @@ jobs:
     name: Release Github
     if: ${{ needs.changelog.outputs.skipped != 'true' }}
     runs-on: ubuntu-latest
-    needs: [build, changelog]
+    needs: [ build, changelog ]
     steps:
       - name: Publish Release
-        uses: eregon/publish-release@v1
+        uses:  eregon/publish-release@v1.0.6
```
magicus: Actually, this pins this release as well, since `@1` will allow for any minor or patch release after that. Is this what you want? (For lesser-known mods, it could be a point, but I would like to know that it was intentional).

If you want to keep it, please revert the extra space.

.github/workflows/release.yml
```diff
@@ -153,7 +153,7 @@ jobs:
     needs: [build, changelog]
     steps:
       - name: Publish Release
-        uses: eregon/publish-release@v1
+        uses: eregon/publish-release@v1.0.6
```
magicus, line 156: Same here as above.

gradle.properties
```diff
@@ -10,7 +10,7 @@ minecraft_version=1.20.2
 # Check for latest at https://maven.architectury.dev/architectury-plugin/architectury-plugin.gradle.plugin
 architectury_plugin_version=3.4-SNAPSHOT
 # Check for latest at https://maven.architectury.dev/dev/architectury/loom/dev.architectury.loom.gradle.plugin
-architectury_loom_version=1.4-SNAPSHOT
+architectury_loom_version=1.5-SNAPSHOT
```
magicus, line 13: I know fabric-loom recently released 1.6, but maybe architectury is lagging behind a bit?

kristofbolyai: Arch loom 1.6 is beta only.

.github/workflows/release.yml
kristofbolyai: It's intentional for smaller actions.

.github/workflows/alpha-release.yml
magicus: Ah.

.github/workflows/alpha-release.yml
magicus:
```suggestion
        uses: eregon/publish-release@v1.0.6
```

.github/workflows/alpha-release.yml
kristofbolyai: Fixed.

magicus (approved):

PR #2427, opened 2024-04-24
feat: Add TowerStatsFeature and TowerStatsOverlay

kristofbolyai:
## Overlay
image

## Chat log
image


common/src/main/java/com/wynntils/functions/WarFunctions.java
```diff
@@ -22,4 +26,126 @@ public Double getValue(FunctionArguments arguments) {
...
+            if (initialTowerStateOpt.isEmpty()) return -1;
+
+            return initialTowerStateOpt.get().defense();
+        }
+    }
+
+    public static class InitialTowerDamageFunction extends Function<CappedValue> {
+        @Override
+        public CappedValue getValue(FunctionArguments arguments) {
+            // Technically this is a RangedValue, but functions don't support RangedValue
```
magicus: Should we fix that?

kristofbolyai: This should be fine for the time being imo.

magicus: Ok.

common/src/main/java/com/wynntils/functions/generic/RangedFunctions.java
magicus, line 1: Thanks. I think this is better to actually expose our "reality" in functions.

common/src/main/java/com/wynntils/functions/generic/StringFunctions.java
```diff
@@ -43,6 +44,33 @@ public FunctionArguments.RequiredArgumentBuilder getRequiredArgumentsBuilder() {
         }
     }
 
+    public static class FormatRangedFunction extends GenericFunction<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            RangedValue value = arguments.getArgument("value").getRangedValue();
+            return StringUtils.integerToShortString(value.low()) + "/" + StringUtils.integerToShortString(value.high());
```
magicus: We represent RangedValues as `<low>-<high>` in our toString method. I think we should do the same here. That is kind of the essential difference between a ranged value ("between low and high") and a capped value `<cur>/<max>` ("the value cur out of max").

magicus:
Some of the more general changes (like in `StringUtil.java`, and the function value changes) perhaps should be extracted into a separate PR that targets main directly?

kristofbolyai:
> Some of the more general changes (like in `StringUtil.java`, and the function value changes) perhaps should be extracted into a separate PR that targets main directly?

I think it's not necessary in this case, since main wouldn't be using any of the function changes. As for the StringUtil class, I think there is no rename there, just a int -> long type change, which should work implicitly. Also, in theory, main won't get anything "new", except for the item filter UI, until this PR is finished (which I am pushing hard for, as you can see, so it can be done before I  have all my exams).

common/src/main/java/com/wynntils/functions/generic/StringFunctions.java
kristofbolyai:
```suggestion
            return StringUtils.integerToShortString(value.low()) + "-" + StringUtils.integerToShortString(value.high());
```

Indeed! I just copied the method, and forgot to change it. Thanks!

magicus (approved):

PR #2429, opened 2024-04-25
feat: Handle territory timers in a more advanced way to make sure they always appear as accurate as they can

kristofbolyai:


magicus (approved):

common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java
```diff
@@ -13,30 +13,51 @@
...
+ * - Use chat messages when we can, as they are more reliable
+ * - Use scoreboard messages when we don't have a chat message
+ * - If a chat message is received for a territory, we stop using the scoreboard for that territory
+ * - We also take territory defenses from chat messages
+ */
 public final class GuildAttackTimerModel extends Model {
     private static final Pattern GUILD_ATTACK_PATTERN = Pattern.compile("§b- (.+):(.+) §3(.+)");
     private static final Pattern GUILD_DEFENSE_CHAT_PATTERN = Pattern.compile("§3.+§b (.+) defense is (.+)");
+    private static final Pattern WAR_MESSAGE_PATTERN = Pattern.compile(
+            "§3\\[WAR\\]§c The war for (?<guild>.+) will start in (?<remaining>.+) (?<type>minutes|seconds)\\.");
```
ShadowCat117: Minor nitpick but the guild group should be called territory

ShadowCat117:
Code looks good, can't test it but I don't see why it wouldn't work. Only exception is I think you need to handle refunds to remove timers

kristofbolyai:
> Code looks good, can't test it but I don't see why it wouldn't work. Only exception is I think you need to handle refunds to remove timers

Refunds are handled.

common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java
kristofbolyai: Thanks!

ShadowCat117:
> Refunds are handled.

Oh I see, the captured message appears if a different guild takes the territory so no need to check for the refund message

kristofbolyai:
> > Refunds are handled.
> 
> Oh I see, the captured message appears if a different guild takes the territory so no need to check for the refund message

Yeah. At least I really think so, because all of my logs have a captured message right before the refund message...

PR #2430, opened 2024-04-25
feat: Add guild bank redirector in ChatRedirectFeature

kristofbolyai:
image


ShadowCat117 (approved):
Great idea, this is a better solution than filtering the messages entirely or adding a specific chat tab

kristofbolyai:
Ended up changing the color for community opinion:
https://i.imgur.com/GyyX8jD.png

magicus:
As a side note, the Chat redirector has really outgrown itself as a feature. I think we should have a model (*) for this, were we can register the parsers. It is encoding a lot of "knowledge" about Wynncraft messages and what they mean.

(*) or maybe another Component; I'm not sure...

kristofbolyai:
Probably the extension of existing chat types in some ways.

PR #2431, opened 2024-04-25
feat: Add ExtendedSeasonLeaderboardFeature

kristofbolyai:
# Wynncraft Default (original)

# Default with Wynntils

# Short scores - optional config


ShadowCat117:
I think you need to change labelInfos to a map with the season as the key and the labelinfo as the value.

Changing to season 16 immediately shows the issue

magicus:
The code (incl the LabelHandler changes) look fine overall, but I guess you must address Jamie's complaints first. :)

ShadowCat117:
Issue is still present :( Might be related but you don't seem to be adding anything to `liveLabels`

image

kristofbolyai:
Oh... It looked like it's working, but I guess I didn't do the math..

common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
```diff
@@ -130,8 +130,12 @@ private void writeToJson() {
...
 
         storages.forEach((jsonName, storage) -> {
-            JsonElement jsonElem = Managers.Json.GSON.toJsonTree(storage.get(), storageTypes.get(storage));
-            storageJson.add(jsonName, jsonElem);
+            try {
+                JsonElement jsonElem = Managers.Json.GSON.toJsonTree(storage.get(), storageTypes.get(storage));
+                storageJson.add(jsonName, jsonElem);
+            } catch (Throwable t) {
+                WynntilsMod.error("Failed to save storage " + jsonName, t);
+            }
```
kristofbolyai, line 138: Only used this for debugging, but I think it should be kept, to reduce the risk of "resetting" storage files.

kristofbolyai:
@JamieCallan117 Should be fixed now

common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
ShadowCat117: Yes, I have seen some people reporting their storage file getting reset so hopefully this will help prevent that

ShadowCat117 (approved):
Looks good and works perfectly now! :)

PR #2432, opened 2024-04-27
feat: Add an overall mode the statistics page

kristofbolyai:
image
image
image


ShadowCat117 (approved):
Looks good!

Didn't realise the statistics list was so empty I kind of forgot we had it, hopefully we can track some more things with the fuy features like wars and raids to hopefully give it some more use :)

kristofbolyai:
> Looks good!
> 
> Didn't realise the statistics list was so empty I kind of forgot we had it, hopefully we can track some more things with the fuy features like wars and raids to hopefully give it some more use :)

Yes... It was added as a system, with little implemented features overall. Hopefully we can fill it up in the future.

PR #2434, opened 2024-04-27
feat: Add statistics functions, add overall support for statistics command

kristofbolyai:
image
image
image
image


common/src/main/java/com/wynntils/functions/StatisticFunctions.java
```diff
@@ -0,0 +1,150 @@
...
+
+            return statisticKind.getFormattedValue(value);
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.RequiredArgumentBuilder(List.of(
+                    new FunctionArguments.Argument<>("statisticKey", String.class, null),
+                    new FunctionArguments.Argument<>("value", Number.class, null)));
+        }
```
ShadowCat117, line 104: I'm a little confused on why this function needs these 2 arguments, I think you meant so that you can pass in the character or overall value for the 2nd argument, but wouldn't it be simpler to have a boolean for character or overall?

common/src/main/java/com/wynntils/functions/StatisticFunctions.java
```diff
@@ -0,0 +1,150 @@
...
+        public String getValue(FunctionArguments arguments) {
+            String statisticKey = arguments.getArgument("statisticKey").getStringValue();
+            int value = arguments.getArgument("value").getIntegerValue();
+
+            StatisticKind statisticKind;
+            try {
+                statisticKind = StatisticKind.from(statisticKey);
+            } catch (IllegalArgumentException e) {
+                return "-";
+            }
```
ShadowCat117: I don't think this is working as intended
image

common/src/main/java/com/wynntils/functions/StatisticFunctions.java
kristofbolyai: It's not meant to be a boolean. It's meant to be a numeric value to be formatted. Each statistic key has a it's own formatter.

image

common/src/main/java/com/wynntils/screens/statistics/widgets/StatisticButton.java
```diff
@@ -51,21 +52,20 @@ public void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, float
         CustomColor backgroundColor = getButtonBackgroundColor();
         RenderUtils.drawRect(poseStack, backgroundColor, this.getX(), this.getY(), 0, this.width, this.height);
 
-        int maxTextWidth = this.width - 2;
         FontRenderer.getInstance()
-                .renderText(
+                .renderScrollingText(
```
ShadowCat117, line 56: I was thinking of implementing this to the activity screens and other places it could be used once it was merged into main so glad to see it here :)

common/src/main/java/com/wynntils/functions/StatisticFunctions.java
ShadowCat117: Ah yeah that makes sense now

common/src/main/java/com/wynntils/functions/StatisticFunctions.java
ShadowCat117: Actually, it's probably because I stuck a different stat into the 2nd argument, this try catch should work but one should be needed for the `getFromattedValue`

kristofbolyai: I forgot to do the null check after I changed from using valueOf to from.

ShadowCat117 (approved):

PR #2435, opened 2024-04-28
feat: Add mythic pull dry streaks and some new lootrun statistics

kristofbolyai:


common/src/main/java/com/wynntils/handlers/item/ItemHandler.java
```diff
@@ -124,6 +128,22 @@ public void onContainerSetContent(ContainerSetContentEvent.Pre event) {
         }
     }
 
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onEntitySpawn(SetEntityDataEvent event) {
+        Entity entity = McUtils.mc().level.getEntity(event.getId());
+        if (!(entity instanceof ItemEntity itemEntity)) return;
+
+        // Item entities can have item's that need to be annotated
```
kristofbolyai:
```suggestion
        // Item entities can have an item that need to be annotated
```

common/src/main/java/com/wynntils/features/MythicFoundFeature.java
```diff
@@ -36,26 +38,54 @@ public void onMythicFound(MythicFoundEvent event) {
...
+                .append(itemName.getComponent())
+                .append(Component.literal(" after ")
+                        .withStyle(ChatFormatting.LIGHT_PURPLE)
+                        .append(Component.literal(Models.LootChest.dryPulls.get() + " pulls")
+                                .withStyle(ChatFormatting.GOLD)))
+                .append(Component.literal(" without a mythic.").withStyle(ChatFormatting.LIGHT_PURPLE)));
+    }
+
+    private static void sendNormalDryStreakMessage(StyledText itemName) {
+        McUtils.sendMessageToClient(Component.literal("Dry streak broken! Found a ")
```
ShadowCat117:
```suggestion
        McUtils.sendMessageToClient(Component.literal("Dry streak broken! Found an ")
```

To make it the same as the other message

common/src/main/java/com/wynntils/models/containers/LootChestModel.java
```diff
@@ -23,23 +25,33 @@
...
 import net.minecraft.client.gui.screens.Screen;
 import net.minecraft.core.BlockPos;
+import net.minecraft.network.syncher.SynchedEntityData;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.entity.item.ItemEntity;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public final class LootChestModel extends Model {
```
ShadowCat117: I think most of the changes here are more suited to be in `LootrunModel` rather than here

kristofbolyai: Yeah, perhaps. Originally I wanted it here because it was checking the chest inventory.

ShadowCat117:
I'm not sure if the dry pull tracking is fully accurate.

What I interpret the feature should do is at the end of a lootrun check the completion message for pull count, which the model already does, then only after the chest has been collected from, the dry streak should then be incremented as only now you have gone x pulls without a mythic, currently it looks like it will increment the dry count on completion, so if you get a mythic on your first run it will show a dry streak of x pulls when in reality it should be 0. (Unless we want to increment the counter on rerolls).

common/src/main/java/com/wynntils/services/statistics/StatisticsCollectors.java
```diff
@@ -45,4 +49,19 @@ public void onLootrunFailed(LootrunFinishedEvent.Failed event) {
                 StatisticKind.LOOTRUNS_CHALLENGES_COMPLETED, event.getChallengesCompleted());
         Services.Statistics.addToStatistics(StatisticKind.LOOTRUNS_TIME_ELAPSED, event.getTimeElapsed());
     }
+
+    @SubscribeEvent
+    public void onMythicFoundEvent(MythicFoundEvent event) {
+        Services.Statistics.increaseStatistics(StatisticKind.MYTHICS_FOUND);
+
+        Services.Statistics.addToStatistics(
+                StatisticKind.LOOTRUNS_CHALLENGES_WITHOUT_MYTHIC, Models.Lootrun.dryPulls.get());
```
ShadowCat117: This should have a check to make sure the mythic was found in a lootrun and not outside one

kristofbolyai: The event is only fired on lootrun/lootchest finds, but yes, we should check for lootrun.

common/src/main/java/com/wynntils/services/statistics/type/StatisticKind.java
```diff
@@ -11,20 +11,23 @@
...
     SPELLS_CAST(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
+    // endregion
 
     // region Lootruns
-
+    MYTHICS_FOUND(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
     LOOTRUNS_COMPLETED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
     LOOTRUNS_FAILED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
     LOOTRUNS_CHALLENGES_COMPLETED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.ADVANCED),
+    LOOTRUNS_CHALLENGES_WITHOUT_MYTHIC(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.ADVANCED),
```
ShadowCat117: Maybe rename this to PULLS instead of CHALLENGES as it's storing the dry pull count and they aren't 1:1 due to purple and dark grey beacons

kristofbolyai:
> I'm not sure if the dry pull tracking is fully accurate.
> 
> What I interpret the feature should do is at the end of a lootrun check the completion message for pull count, which the model already does, then after the chest has been collected from the dry streak should then be incremented as only now you have gone x pulls without a mythic, currently it looks like it will increment the dry count on completion, so if you get a mythic on your first run it will show a dry streak of x pulls when in reality it should be 0. (Unless we want to increment the counter on rerolls).

Hm, I think that's up to interpretation. I think we should count the LR's pulls somewhere, when we find a mythic, but being 100 dry after finding a mythic in the same run is not logical, so we include it before the rewards. As for the statistic, yes, it's only added to when actually finding a mythic, so it is not a "live" count. Sadly the statistic system is currently only capable for this kind of data. In the future, we may able to change this to a more advanced version.

common/src/main/java/com/wynntils/services/statistics/type/StatisticKind.java
kristofbolyai: I would have to rename the upper one as well, but I can't, since we can't up-fix it. The localizations already say pulls for this reason. Ill change the new one's name, and add a comment for the old one.

ShadowCat117:
> Sadly the statistic system is currently only capable for this kind of data. In the future, we may able to change this to a more advanced version.

Ah ok, this should be fine for now then

ShadowCat117 (approved):

PR #2438, opened 2024-04-29
refactor: Move onlinemembers and player command json handling to models

ShadowCat117:


kristofbolyai (commented):
Maybe you should return a deserialized API object, and turn it into a component at command level? Otherwise, this refactor has no practical usage.

common/src/main/java/com/wynntils/commands/OnlineMembersCommand.java
```diff
@@ -52,91 +49,57 @@ public LiteralArgumentBuilder<CommandSourceStack> getCommandBuilder(
...
 
-                    JsonObject guildMembers = json.getAsJsonObject("members");
+            try {
+                guild = completableFuture.get();
+            } catch (InterruptedException | ExecutionException e) {
+                completableFuture.whenComplete((result, throwable) ->
+                        McUtils.sendMessageToClient(Component.literal("Error trying to parse guild online members")
+                                .withStyle(ChatFormatting.RED)));
+                return;
+            }
```
kristofbolyai: This will be blocking on the main thread, until the request completes. How about adding a `whenComplete` hook? (Obviously, for all cases).

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -38,6 +41,9 @@
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 public class GuildModel extends Model {
+    private static final Gson GUILD_GSON = new GsonBuilder()
+            .registerTypeHierarchyAdapter(Guild.class, new Guild.GuildDeserializer())
+            .create();
     private static final Gson GUILD_PROFILE_GSON = new GsonBuilder()
             .registerTypeHierarchyAdapter(GuildProfile.class, new GuildProfile.GuildProfileDeserializer())
             .create();
```
kristofbolyai: I think you are safe to combine these two.

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
```diff
@@ -5,12 +5,12 @@
...
-    RECRUITER("Recruiter", "★"),
-    CAPTAIN("Captain", "★★"),
-    STRATEGIST("Strategist", "★★★"),
+    OWNER("Owner", "★★★★★"),
     CHIEF("Chief", "★★★★"),
-    OWNER("Owner", "★★★★★");
+    STRATEGIST("Strategist", "★★★"),
+    CAPTAIN("Captain", "★★"),
+    RECRUITER("Recruiter", "★"),
+    RECRUIT("Recruit", "");
```
kristofbolyai: Was there a reason for this?

kristofbolyai (commented):
I would also rename all "type" to have an `Info` appendix. `GuildInfo`, `GuildMemberInfo`, etc.

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
ShadowCat117: For the online members command to iterate over the ranks in this order rather than the reverse. Easier than reversing the list before the loop

kristofbolyai: Can you verify `GuildRank#ordinal` not used anywhere?

common/src/main/java/com/wynntils/models/players/PlayerModel.java
```diff
@@ -190,6 +200,21 @@ private void loadUser(UUID uuid, String userName) {
                 });
     }
 
+    public CompletableFuture<com.wynntils.models.players.type.PlayerInfo> getPlayer(String username) {
+        CompletableFuture<com.wynntils.models.players.type.PlayerInfo> future = new CompletableFuture<>();
```
kristofbolyai: Seems like IDEA had a stroke when refactoring this...

kristofbolyai: Or.. MC already has a `PlayerInfo` class. Rename to `WynnPlayerInfo` perhaps?

common/src/main/java/com/wynntils/models/players/type/GuildRank.java
ShadowCat117: Ah good thing you asked, I thought I checked everywhere but I missed `GuildRank.values()[stars]` in the replacement feature. Will just have to reverse the list then

common/src/main/java/com/wynntils/models/players/PlayerModel.java
ShadowCat117: Yeah already has one and both are already used in the model

common/src/main/java/com/wynntils/models/players/type/WynnPlayerInfo.java
```diff
@@ -0,0 +1,89 @@
...
+
+                    try {
+                        guild = completableFuture.get();
+                    } catch (InterruptedException | ExecutionException e) {
+                        McUtils.sendMessageToClient(Component.literal("Error trying to parse player guild")
+                                .withStyle(ChatFormatting.RED));
+                        return null;
+                    }
+
+                    Optional<String> guildJoinedTimestampOpt = guild.guildMembers().stream()
```
kristofbolyai: guild could be null at this point, right? You should throw then. Also, since you need to do changes anyway, can you rename the deserializer classes to match the record names?

kristofbolyai (approved):

PR #2439, opened 2024-04-29
feat: Add TerritoryManagementScreen

kristofbolyai:
[Kooha-2024-05-05-12-03-23.webm](https://github.com/Wynntils/Artemis/assets/49001742/e11b1318-cf6b-446e-bccb-b8e437750b65)


ShadowCat117:
Not a big deal but updating the search widget should probably update the quick filters/sorts

kristofbolyai:
> Not a big deal but updating the search widget should probably update the quick filters/sorts

I think we can keep them desynced. So the quick filters save their state, so the user can type advanced queries, but still can come back to the quick filters not being reset.

common/src/main/java/com/wynntils/features/ui/CustomTerritoryManagementScreenFeature.java
```diff
@@ -0,0 +1,118 @@
...
+                    customScreenOpened = true;
+                }
+            }
+            case DISABLED_IF_SHIFT_HELD -> {
+                if (!shiftClickedTerritoryItem) {
+                    event.setOpenScreen(true);
+                    customScreenOpened = true;
+                }
+            }
+        }
```
ShadowCat117: Since they're all setting to true, maybe make a `shouldOpenScreen` from the switch and then set it based on that. Something like

```
boolean shouldOpenScreen = switch (shiftBehaviorConfig.get()) {
    case ENABLED_IF_SHIFT_HELD -> shiftClickedTerritoryItem;
    case DISABLED_IF_SHIFT_HELD -> !shiftClickedTerritoryItem;
    default -> true;
};

event.setOpenScreen(shouldOpenScreen);
customScreenOpened = shouldOpenScreen;
```

kristofbolyai: Sure. Although the trade market screen does the same exact thing, I might as well change it.

ShadowCat117 (approved):
I think it looks good! Not really an eco person myself so can't think of any other functionality to add but if it does what the fuy version does that's probably enough

kristofbolyai:
> I think it looks good! Not really an eco person myself so can't think of any other functionality to add but if it does what the fuy version does that's probably enough

I've made my best efforts to talk to people who do eco, I think this should be fine.

PR #2442, opened 2024-04-29
fix: Fix TerritoryModel#getTerritoryProfileFromShortName not having a stable order

kristofbolyai:


ShadowCat117 (approved):
Hopefully will fix the issue :)

PR #2444, opened 2024-05-01
perf: Multiple `StyledText` optimizations

kristofbolyai:
# Test
It's hard to fairly test `StyledText`, as it's used widely across the mod. The results below were acquired from rendering the "Test" overlay page for 10 seconds, on both the old and the new versions. From general gameplay profiling, I've found that most of the time is spent on `StyledText` <-> text rendering, so this should be a (somewhat) representative result.

# Results
| Type | Old | New |
|--------|--------|--------|
| Memory | 267 MB | 74 MB (-70%) |
| CPU | 150 | 75 (-50%) |

Overall performance improvements are from 50% to 70% in both CPU cycles and memory usage.

# Verification
The changes made are trivial, but overall I've found no rendering artifacts.


magicus (approved):
Looks good.

PR #2445, opened 2024-05-03
fix: Make GuildAttackTimerModel use string territory names instead of profiles

kristofbolyai:
It is not correct to use profiles, as they can vary when guilds take territories. Also added a fail-safe if a timer is refunded, but the scoreboard is not updated fast enough. In reality, this might not even be needed, but there is no harm in keeping it.


ShadowCat117 (approved):
Looks fine to me

kristofbolyai:
I've also just tested this, and it seems to work. I've been spending a considerable amount of time playing on the new version (+merged the performance fixes), and I can safely say I have found it pretty stable, except for these timers, which should be fixed now.

PR #2446, opened 2024-05-03
feat: Raid progress overlay and print times feature

ShadowCat117:
Did many runs of each raid whilst working on it but only 3 after everything was complete so should all work but would recommend at least 1 other person to try it out for multiple runs.

Total time includes the power up rooms and the time from intro cutscene to start of challenge 1 so that's why they don't add up

I'm also not the best at functions so if someone can create a better default then by all means suggest one.


common/src/main/java/com/wynntils/models/raid/event/RaidFinishedEvent.java
kristofbolyai, line 1: Maybe `RaidEndedEvent` is better? Finished is closer to completed, ended is more generic.

common/src/main/java/com/wynntils/functions/RaidFunctions.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    public static class RaidRoomTimeFunction extends Function<Integer> {
+        @Override
+        public Integer getValue(FunctionArguments arguments) {
+            if (Models.Raid.getCurrentRaid() == null) return -1;
+
+            RaidRoomType roomType =
+                    RaidRoomType.fromName(arguments.getArgument("roomName").getStringValue());
```
kristofbolyai, line 85: It's a bit unfortunate that we can't provide a list of possible values for functions at the moment. Although it wouldn't be a big change, we would have to go through many functions to fill in the possible values. Aaand if we do that, we might as well add support for "enum-functions". So most likely out of scope for now.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -0,0 +1,209 @@
...
+                WynntilsMod.postEvent(new RaidFinishedEvent.Failed(
+                        currentRaid, (int) ((System.currentTimeMillis() - raidStartTime) / 1000)));
+            } else {
+                return;
+            }
+
+            currentRaid = null;
+            currentRoom = null;
+            roomTimers.clear();
+        }
```
kristofbolyai:
```suggestion
        if (currentRaid == null) {
            currentRaid = Raid.fromTitle(styledText);

            if (currentRaid != null) {
                // In a raid, set to intro room and start timer
                currentRoom = RaidRoomType.INTRO;
                raidStartTime = System.currentTimeMillis();
            }
            return;
        } 
        
        if (styledText.equals(RAID_COMPLETE)) {
            int timeTaken = (int) ((System.currentTimeMillis() - raidStartTime) / 1000);
            // Raid has been completed, post event with time taken in seconds
            WynntilsMod.postEvent(new RaidFinishedEvent.Completed(currentRaid, timeTaken));

            checkForNewPersonalBest(currentRaid, timeTaken);
            
            currentRaid = null;
            currentRoom = null;
            roomTimers.clear();
        } else if (styledText.equals(RAID_FAILED)) {
            // Raid failed, post event with time elapsed in seconds
            WynntilsMod.postEvent(new RaidFinishedEvent.Failed(
                    currentRaid, (int) ((System.currentTimeMillis() - raidStartTime) / 1000)));
          currentRaid = null;
          currentRoom = null;
          roomTimers.clear();
        }
```

else -> return is a bit odd, so maybe 3 lines of code duplication is better. Alternatively, set a boolean and return if it was not either case.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @SubscribeEvent
+    public void onLabelChange(EntityLabelChangedEvent event) {
```
kristofbolyai, line 103: Is there a reason why you didn't use `LabelHandler` and created a boss fight label parser? We already have NPC label parsing, so you could use that to check if the label is a blacksmith NPC. Or is there something I am missing?

common/src/main/java/com/wynntils/models/raid/type/Raid.java
kristofbolyai, line 1: Nit;

I dislike 1 word enums. Would you like `RaidKind` or `RaidType` better?

common/src/main/java/com/wynntils/overlays/RaidProgressOverlay.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class RaidProgressOverlay extends TextOverlay {
+    private static final String TEMPLATE =
+            "{concat(\"§6§l§n\";current_raid;\"\\n\\n§dChallenge 1: \";if_str(eq(raid_room_time(\"challenge_1\");-1);\"§7--:--\";concat(\"§b\";leading_zeros(int(div(raid_room_time(\"challenge_1\");60));2);\":\";leading_zeros(int(mod(raid_room_time(\"challenge_1\");60));2)));\"\\n§dChallenge 2: \";if_str(eq(raid_room_time(\"challenge_2\");-1);\"§7--:--\";concat(\"§b\";leading_zeros(int(div(raid_room_time(\"challenge_2\");60));2);\":\";leading_zeros(int(mod(raid_room_time(\"challenge_2\");60));2)));\"\\n§dChallenge 3: \";if_str(eq(raid_room_time(\"challenge_3\");-1);\"§7--:--\";concat(\"§b\";leading_zeros(int(div(raid_room_time(\"challenge_3\");60));2);\":\";leading_zeros(int(mod(raid_room_time(\"challenge_3\");60));2)));\"\\n\\n§4Boss: \";if_str(eq(raid_room_time(\"boss_fight\");-1);\"§7--:--\";concat(\"§b\";leading_zeros(int(div(raid_room_time(\"boss_fight\");60));2);\":\";leading_zeros(int(mod(raid_room_time(\"boss_fight\");60));2)));\"\\n\\n§5Total: \";concat(\"§b\";leading_zeros(int(div(raid_time;60));2);\":\";leading_zeros(int(mod(raid_time;60));2)))}";
```
kristofbolyai: We may not do this for other text overlays (my bad!), but if you have this template saved with actual new-lines, I think we should start using [Java Text Blocks](https://docs.oracle.com/en/java/javase/15/text-blocks/index.html) to include the actual, multi-line version if this, if we ever need to edit it, or if someone wants to use this as their base for their function.

kristofbolyai: Edit: I see you already used text blocks below :)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1793,6 +1811,12 @@
   "function.wynntils.professionXpPerMinuteRaw.argument.profession": "The profession to check",
   "function.wynntils.professionXpPerMinuteRaw.description": "The raw amount of XP you gained for the specified profession in the last minute",
   "function.wynntils.professionXpPerMinuteRaw.name": "Profession XP Per Minute (Raw)",
+  "function.wynntils.raidPersonalBestTime.argument.raidName": "The name of the raid to get the best time for",
+  "function.wynntils.raidPersonalBestTime.description": "The fastest time taken to beat the specified raid in seconds",
+  "function.wynntils.raidPersonalBestTime.name": "Raid Personal Best Time",
+  "function.wynntils.raidRoomTime.argument.roomName": "The room type to get the time taken to complete for. One of CHALLENGE_1, CHALLENGE_2 or CHALLENGE_3",
+  "function.wynntils.raidRoomTime.description": "How long did it take to complete the specified raid room in seconds",
```
```suggestion
  "function.wynntils.raidRoomTime.description": "How long it took to complete the specified raid room in seconds",
```

kristofbolyai (commented):
I've made my best efforts to look through this once carefully. I know this is a pain to work on, as you have to send almost an equal time playing as programming.

Essentuan:
I have a few questions/comments on this. Everyone I've spoken to has preferred powerup rooms and the boss intermission to be included in each challenge & boss timer. Same with the way the time is formatted, they prefer mm:ss:ms. Also the state of this will be broken if a player switches world or gets disconnected then rejoins after their party has completed a room.

kristofbolyai:
Valid comments IMO. Would it be hard to have a raid info object with both version of times included? This implementation feels more casual, the fuygg is more speedrun-y.

Essentuan:
If there would be an easy way to add it from fuy.gg I could do it.

ShadowCat117:
Only problem is with this tracking method, the boss intermission detection is not accurate as NOG and NOL don't render the boss fight label until you get in range which is a few blocks away so challenge 3 would be longer and the boss fight would be shorter.

Milliseconds are fine to add though.

> Also the state of this if a player will be broken if a player switches world or gets disconnected then rejoins after their party has completed a room.

Ah, I had presumed if a player disconnected they'd be unable to rejoin. Guess I'll have to try it out

kristofbolyai:
> If there would be an easy way to add it from fuy.gg I could do it. 

I think this PR has a solid foundation to do it, I am mainly just asking because I don't know how timers are currently saved.

kristofbolyai:
> Only problem is with this tracking method, the boss intermission detection is not accurate as NOG and NOL don't render the boss fight label until you get in range which is a few blocks away so challenge 3 would be longer and the boss fight would be shorter.

Can't you guess when you enter the boss intermission room when you "switch" rooms after having 3 missions completed?

Essentuan:
> > If there would be an easy way to add it from fuy.gg I could do it.
> 
> I think this PR has a solid foundation to do it, I am mainly just asking because I don't know how timers are currently saved.

uhm I technically store the 'structure' of the raid (Each room, their time, when they started, when they were completed) so that if the raid ever changes I can update each raid without affecting pbs.

Essentuan:
> Only problem is with this tracking method, the boss intermission detection is not accurate as NOG and NOL don't render the boss fight label until you get in range which is a few blocks away so challenge 3 would be longer and the boss fight would be shorter.

You could use when the player teleports out of the intermission room (thats what I do) but if you want I have replays of all 3 raids saved and I could send you them if you wanna look into other methods.

ShadowCat117:
> Can't you guess when you enter the boss intermission room when you "switch" rooms after having 3 missions completed?

> You could use when the player teleports out of the intermission room

I think both of these would require knowledge of the coordinates for the 3rd power up room which I'd rather avoid as they could change in a patch. Unless we're fine with that?

kristofbolyai:
> > Can't you guess when you enter the boss intermission room when you "switch" rooms after having 3 missions completed?
> 
> 
> 
> > You could use when the player teleports out of the intermission room
> 
> 
> 
> I think both of these would require knowledge of the coordinates for the 3rd power up room which I'd rather avoid as they could change in a patch. Unless we're fine with that?

If you can detect a power up
room from labels (or however you detect it now), you could count them perhaps? 

I'll probably have to take a good look again tomorrow, I am only on phone now..

Essentuan:
> I think both of these would require knowledge of the coordinates for the 3rd power up room which I'd rather avoid as they could change in a patch. Unless we're fine with that?

I doubt they would change the location of the room unless they redid the whole raid. But if the positions of the rooms did ever change there would certainly be a hero beta so you could prepare the changes ahead of time. (Maybe you could load the coords from some external source so it can be updated without a new version of wynntils? Though that seems like a bit too much imo)

ShadowCat117:
> If you can detect a power up room from labels (or however you detect it now), you could count them perhaps?

Using the blacksmith as TCC power up labels aren't visible right away from the tp, but that's probably out of range at the exit position, same as the power ups.

> (Maybe you could load the coords from some external source so it can be updated without a new version of wynntils? Though that seems like a bit too much imo)

We could put the info in static-storage, what do you think about that @kristofbolyai? The label info could be used for when no data has been downloaded, otherwise it can use the coords from there

kristofbolyai:
> We could put the info in static-storage, what do you think about that @kristofbolyai? The label info could be used for when no data has been downloaded, otherwise it can use the coords from there

I think it is too complicated to do both, so we should just decide on using either option. I want to give labels one last try, otherwise, we can use coords.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ShadowCat117: I didn't because that seemed more for matching patterns whereas here it's looking for one exact match. The blacksmith could be done using it though

kristofbolyai: If you don't want to convert both, just don't do the change, so it's less chaotic overall.

ShadowCat117:
Updated overlay with milliseconds and intermission timer

kristofbolyai:
> Updated overlay with milliseconds and intermission timer 

Maybe the format should be "mm:ss.mmm"?

kristofbolyai (approved):
Code looks good. I think you should position the overlay at the same place the lootrun overlay is currently, to make sure it doesn't interact with the hades overlay.

common/src/main/java/com/wynntils/services/statistics/type/StatisticKind.java
```diff
@@ -29,7 +29,18 @@ public enum StatisticKind {
...
+
+    // region Raids
+    NEST_OF_THE_GROOTSLANGS_FAILED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
+    NEST_OF_THE_GROOTSLANGS_TIME_ELAPSED(CustomStatFormatters.TIME, StatisticType.ADVANCED),
+    ORPHIONS_NEXUS_OF_LIGHT_FAILED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
+    ORPHIONS_NEXUS_OF_LIGHT_TIME_ELAPSED(CustomStatFormatters.TIME, StatisticType.ADVANCED),
+    THE_CANYON_COLOSSUS_FAILED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
+    THE_CANYON_COLOSSUS_TIME_ELAPSED(CustomStatFormatters.TIME, StatisticType.ADVANCED),
+    THE_NAMELESS_ANOMALY_FAILED(CustomStatFormatters.FORMATTED_NUMBER, StatisticType.COUNT),
+    THE_NAMELESS_ANOMALY_TIME_ELAPSED(CustomStatFormatters.TIME, StatisticType.ADVANCED);
```
kristofbolyai: Actually, shouldnt't we have "succeeded" stats :)?

ShadowCat117: I didn't add them since Wynn already tracks them and you can see it on your stats page. Can still add it here if you think it's worthwhile

kristofbolyai: I think we might as well do it.

ShadowCat117:
> I think you should position the overlay at the same place the lootrun overlay is currently, to make sure it doesn't interact with the hades overlay.

I think I did originally put it there, but the chat messages from the raid was blocking it so I moved it up. I'll try find a middle ground

ShadowCat117:
Actually I don't think there really is any middle ground, only option would be to move it to the other side. I'll just line it up with the lootrun one then

ShadowCat117:
Don't merge this yet, I want to change the print times to show each room instead of just overall time as the overlay vanishes as soon as the raid ends so you don't get a chance to see it all complete

kristofbolyai (approved):
Ready to merge now?

ShadowCat117:
Just trying to clean up the printer but if you're ok with it like that then yes. I can just improve it for 1.2.1

kristofbolyai:
> Just trying to clean up the printer but if you're ok with it like that then yes. I can just improve it for 1.2.1

Go ahead then. We can wait for it.

ShadowCat117:
Fastest tna you've ever seen
image

kristofbolyai (approved):
Looks good! Thanks!

PR #2448, opened 2024-05-05
feat: Spell Recast Functions [skip ci]

ryanzhoudev:
https://feedback.wynntils.com/posts/22/add-a-recast-spell-counter-to-info

https://github.com/Wynntils/Artemis/assets/57310593/ae3e3ec1-3f2c-4922-906d-f3e5284f67c8


common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
```diff
@@ -444,6 +444,9 @@ private void registerAllFunctions() {
         registerFunction(new CombatFunctions.AreaDamageAverageFunction());
         registerFunction(new CombatFunctions.AreaDamagePerSecondFunction());
         registerFunction(new CombatFunctions.BlocksAboveGroundFunction());
+        registerFunction(new CombatFunctions.RepeatedSpellNameFunction());
+        registerFunction(new CombatFunctions.RepeatedSpellCountFunction());
+        registerFunction(new CombatFunctions.RepeatedSpellTicksFunction());
```
kristofbolyai: These are kind of weird names IMO. How about `LastSpellName`, `LastSpellRepeatCount` and `TicksSinceLastSpell`?

common/src/main/java/com/wynntils/models/spells/SpellModel.java
```diff
@@ -106,6 +111,47 @@ public void onSubtitleSetText(SubtitleSetTextEvent e) {
...
+        }
+        lastSpellName = e.getSpellType().getName();
+        repeatedSpellCount = 1;
+    }
+
+    @SubscribeEvent
+    public void onTick(TickEvent e) {
+        if (lastSpellName.isEmpty()) return;
+        ticksSinceCast++;
+        if (ticksSinceCast >= 60) {
```
kristofbolyai: You could make the 60 a constant.

common/src/main/java/com/wynntils/core/consumers/functions/FunctionManager.java
ryanzhoudev: I thought about those and they were what I chose at first, but LastSpellName implies that it would work all the time, even when a spell "burst" is not in progress. Not sure if this is behaviour that you want, but I kept the scope here purely to spell "bursts"

kristofbolyai: We could make it generic, no? So it handles both bursts and not bursts

common/src/main/java/com/wynntils/functions/CombatFunctions.java
```diff
@@ -52,4 +52,64 @@ protected List<String> getAliases() {
...
+        public String getValue(FunctionArguments arguments) {
+            return arguments.getArgument("burst").getBooleanValue()
+                    ? Models.Spell.getLastBurstSpellName()
+                    : Models.Spell.getLastSpellName();
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.OptionalArgumentBuilder(
+                    List.of(new FunctionArguments.Argument<>("burst", Boolean.class, true)));
```
kristofbolyai: false is probably a better default, isn't it? (for everything)

ryanzhoudev: Not sure. I don't really see a use case for non-burst other than just extra information, but the burst one is actually used to time recasts or cycles or whatever. If you really want to though, I can do false

ryanzhoudev: Unless you meant to invert the boolean

kristofbolyai: I think the function name itself is more logical with burst mode off?

ryanzhoudev: Ah yea true

kristofbolyai (approved):

PR #2449, opened 2024-05-05
feat: Add WeeklyConfigBackupFeature

kristofbolyai:
This should prevent accidental data loss, if something happens (no matter if we were the mod crashing or not). Some mods crash on tick, causing us to leave open files on windows, deleting them.

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/WeeklyConfigBackupFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/WeeklyConfigBackupFeature.java
@@ -14,6 +14,7 @@ import com.wynntils.core.persisted.storage.Storage;
 import com.wynntils.mc.event.TitleScreenInitEvent;
 import com.wynntils.utils.FileUtils;
 import java.io.File;
+import java.util.concurrent.TimeUnit;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 
 @ConfigCategory(Category.UTILITIES)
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/wynntils/features/utilities/WeeklyConfigBackupFeature.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Persisted
+    private final Storage<Long> lastBackup = new Storage<>(0L);
+
+    @SubscribeEvent
+    public void onTitleScreenInit(TitleScreenInitEvent.Post event) {
+        long currentTime = System.currentTimeMillis();
+        long lastBackupTime = lastBackup.get();
+
+        if (currentTime - lastBackupTime >= BACKUP_INTERVAL_DAYS * 24 * 60 * 60 * 1000) {
```
ryanzhoudev: TimeUnit.DAYS.toMillis() could be cleaner here

common/src/main/java/com/wynntils/features/utilities/WeeklyConfigBackupFeature.java
```diff
@@ -0,0 +1,78 @@
...
+import com.wynntils.core.persisted.config.Category;
+import com.wynntils.core.persisted.config.ConfigCategory;
+import com.wynntils.core.persisted.storage.Storage;
+import com.wynntils.mc.event.TitleScreenInitEvent;
+import com.wynntils.utils.FileUtils;
+import java.io.File;
+import java.util.concurrent.TimeUnit;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UTILITIES)
```
ShadowCat117: Maybe the Wynntils category?

kristofbolyai: Sure, although it fits both.

ShadowCat117 (approved):

common/src/main/java/com/wynntils/features/utilities/WeeklyConfigBackupFeature.java
ShadowCat117: Would be nice if we could add multiple categories, I think someone has mentioned that before.

kristofbolyai: magicus did, we mostly need UI refactors for that.

PR #2451, opened 2024-05-05
feat: Add war statistics and WarSinceFunction

kristofbolyai:


ShadowCat117 (approved):
Looks good.

Don't need to but since guild owner is also being stored, I wonder if functions for how many territories you've taken from a specific guild could also be fun to add, maybe as an additional optional argument for this current function too

kristofbolyai:
> Don't need to but since guild owner is also being stored, I wonder if functions for how many territories you've taken from a specific guild could also be fun to add, maybe as an additional optional argument for this current function too

I want to add a whole screen dedicated to war history later on, hence why I save the full data.

ShadowCat117:
> I want to add a whole screen dedicated to war history later on, hence why I save the full data.

Ooo that sounds good

PR #2456, opened 2024-05-05
fix: Fix PlayerCommand not using the correct date format [skip ci]

kristofbolyai:
The day itself might be off by a day.. Currently we use "dynamic" months, not the regular 30 day months. I guess this is a way to do it...


magicus:
I don't understand this at all. Can you explain what problem you want to solve?

kristofbolyai:
> I don't understand this at all. Can you explain what problem you want to solve?

The player command times are much longer than what `SimpleDateFormatter` can produce. It can only work with dates up to a month. Many players have been playing the game for years, or have been in guilds for years. So we needed a date formatter, that behaves like `SimpleDateFormatter`, but for full dates.

magicus:
So it's really not as much a "date" formatter as a "time period" formatter? Like "2 years 4 months 10 days and 9 seconds"?

PR #2466, opened 2024-05-07
fix: Some lootrun drops not included in box possibilities [skip ci]

ryanzhoudev:
Apparently some lootrun-exclusive rewards are boxes and some are shown as an Unidentified Item.
image


ShadowCat117:
I'd like https://github.com/Wynntils/Artemis/pull/2467 to be in the next version so I'll change that one to trigger ci instead as there shouldn't be any problems with it

ShadowCat117 (approved):

PR #2467, opened 2024-05-07
fix: Parse shaman mask correctly when only one mask available [skip ci]

ShadowCat117:
Was removed in https://github.com/Wynntils/Artemis/pull/2402, thought that subtitle pattern had been removed but it only occurs when you have one mask ability


kristofbolyai:
@JamieCallan117 I added skip-ci, but I'll do a release today, even if I don't get to push my fix.

kristofbolyai (approved):
So this is kind of like the tutorial mode for spell casting, I assume? When you are lower leveled, you have less masks, so the title is different. I see.

ShadowCat117:
Sort of, but even when fully levelled if you only unlock one mask in the tree it still shows the "Mask of the X" subtitle instead of the arrowed one that shows when you have 2 or 3.

PR #2472, opened 2024-05-08
feat: Settings smooth scrolling [skip ci]

ShadowCat117:
Since the textures for https://github.com/Wynntils/Artemis/pull/2335 are going to take longer than hoped I'm going to take some of the non texture improvements from that pr and implement them separately, starting with improved scrolling for the settings screen. This isn't actually the same scrolling that pr uses as that scrolls 1 "row" at a time whereas this is smooth.

https://github.com/Wynntils/Artemis/assets/8337467/8ed8dc2b-6659-4003-bcf1-ca1feb0b98b4


common/src/main/java/com/wynntils/screens/settings/widgets/ConfigTile.java
```diff
@@ -86,29 +81,34 @@ private void renderDisplayName(PoseStack poseStack) {
 
     @Override
     public boolean mouseClicked(double mouseX, double mouseY, int button) {
-        double actualMouseX = mouseX - getRenderX();
-        double actualMouseY = mouseY - getRenderY();
+        // Prevent interaction when the tile is outside of the mask from the screen, same applies to drag and released
+        if ((mouseY <= 21 || mouseY >= 205)) return false;
```
kristofbolyai: Can we make this a constant in the screen? It's also hardcoded in other widgets.

kristofbolyai (commented):
Looks/feels good. Only a single request to do.

common/src/main/java/com/wynntils/screens/settings/widgets/ApplyButton.java
```diff
@@ -22,7 +23,9 @@ public ApplyButton(WynntilsBookSettingsScreen screen) {
                 14,
                 Component.translatable("screens.wynntils.settingsScreen.apply"),
                 List.of(Component.translatable("screens.wynntils.settingsScreen.apply.description")
-                        .withStyle(ChatFormatting.GREEN)));
+                        .withStyle(ChatFormatting.GREEN)),
+                0,
+                McUtils.mc().screen.height);
```
ShadowCat117, line 28: This and CloseButton don't need to worry about the mask but they are `GeneralSettingsButton`'s so they need it, when they get replaced with the side tags in the redesign it'll be fine

kristofbolyai (approved):

PR #2475, opened 2024-05-10
feat: Colour territory attack time if player is inside territory [skip ci]

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/a7ae0451-66e3-49e4-8f58-a3ab1b96d7d2

Makes it a little more obvious when you have multiple queued like this
image


kristofbolyai:
Maybe we don't need bold now?

ShadowCat117:
Was asked to keep the bold
image

kristofbolyai (approved):
Okay then.

PR #2476, opened 2024-05-10
fix: Fix transcription tooltip option not working correctly [skip ci]

ShadowCat117:
Currently Wynnic/Gavellian text is always transcribed and coloured regardless of the tooltip config when it should only change the actual text if the tooltip config is disabled.

Discovery condition was removed as since the model is almost always out of date due to it taking significantly longer to query the content book it was basically never working.

### DIFF
--- a/common/src/main/java/com/wynntils/models/wynnalphabet/WynnAlphabetModel.java
+++ b/common/src/main/java/com/wynntils/models/wynnalphabet/WynnAlphabetModel.java
@@ -9,8 +9,6 @@ import com.wynntils.core.components.Models;
 import com.wynntils.core.text.PartStyle;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.core.text.StyledTextPart;
-import com.wynntils.models.activities.discoveries.DiscoveryInfo;
-import com.wynntils.models.activities.type.ActivitySortOrder;
 import com.wynntils.models.wynnalphabet.type.TranscribeCondition;
 import com.wynntils.utils.StringUtils;
 import com.wynntils.utils.mc.McUtils;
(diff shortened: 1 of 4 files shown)


kristofbolyai (approved):

common/src/main/java/com/wynntils/models/wynnalphabet/WynnAlphabetModel.java
```diff
@@ -351,21 +343,6 @@ private boolean hasTranscriber(WynnAlphabet transciberToFind) {
         return false;
     }
 
-    private boolean hasCompletedDiscovery(WynnAlphabet discoveryToCheck) {
```
magicus, line 354: I think it is a bit sad to see that option go. I can say that I have never managed to complete any of these discoveries, and I've actually played with this setting on to kind of keep a bit of the suspense.

Can't we move this check to the character model instead, so we query the discoveries when the player enters a new world? Then it would be quick to check for this condition. I think we try to parse the quest book at that time anyhow, right?

ShadowCat117: I think(?) we only scan the content book when the user opens the relevant screen, so discoveries would only be checked when the discoveries screen is opened. It could be done every world change but given the amount of discoveries it wouldn't be quick and I imagine we'd get complaints again about the inventory being replaced by the content book

magicus: I think that basically brings me back to my old idea of essentially caching the quest book locally, so we track any changes (and has a full-rescan fallback in case of issues). Having the quest book data instantaneuosly available without cost would probably open up a lot of cool stuff, such as automatically indicating when you are near a starting point for a quest that you have not done, etc.

magicus: So I agree, let's keep this removed for now, but if we do add the content book cache, we should re-implement this.

kristofbolyai: We'll have to figure out a way to have time to do the query, we already do quite a lot on class join. But otherwise, great idea.

Issue #2388, opened 2024-04-05
WorldStateModel needs to correctly handle Housing Islands


ShadowCat117:
There are 2 issues caused by the lack of support for housing islands in the model. The main issue only occurs with guild housings as far as I'm aware.

1. Incorrect world name. A minor issue that causes your current world name to be displayed as the name of the housing island. This affects things like Discord rich presence, bomb bell command and possibly more. Fixed in https://github.com/Wynntils/Artemis/pull/2477
2. World state not being updated to WORLD when switching worlds via travelling to guild housing. This is the main issue, if the custom loading screen is enabled it will never leave the "Receiving terrain" stage and upon pressing esc you will be unable to do almost anything except move. Certain overlays do not work, you cannot chat or leave the island via the usual menu. The only option is to leave the server. If the loading screen is disabled you are able to leave and rejoin the housing the usual way, but certain overlays are still missing until doing so. Fixed in https://github.com/Wynntils/Artemis/pull/2826

### How to reproduce
Issue 1. Join any housing island.
Issue 2. Have a player be on a guild housing island in a seperate world to you and join via the public housing list. Have not tested joining via invite.

Demo of joining island with loading screen disabled, followed by enabled

https://github.com/Wynntils/Artemis/assets/8337467/cc81f969-e270-4b52-823e-68a4c44a9d52

PR #2477, opened 2024-05-10
fix: Don't overwrite current world name when entering housing [skip ci]

ShadowCat117:
Should fix the minor issue from https://github.com/Wynntils/Artemis/issues/2388 but doesn't fix the main issue.

### DIFF
--- a/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
+++ b/common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
@@ -7,7 +7,6 @@ package com.wynntils.features.utilities;
 import com.wynntils.core.components.Handlers;
 import com.wynntils.core.components.Models;
 import com.wynntils.core.consumers.features.Feature;
-import com.wynntils.core.mod.event.WynncraftConnectionEvent;
 import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.Config;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/features/utilities/PerCharacterGuildContributionFeature.java
ShadowCat117, line 1: I believe this is the only place to have a special case to prevent the contribution message from appearing each time you entered housing but there may be other places I'm unaware of

common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
```diff
@@ -147,18 +153,20 @@ public void update(PlayerDisplayNameChangeEvent e) {
...
         Matcher housingNameMatcher = name.getMatcher(HOUSING_NAME);
-        setWorldIfMatched(housingNameMatcher);
+        setWorldIfMatched(housingNameMatcher, true);
     }
 
-    private boolean setWorldIfMatched(Matcher m) {
+    private boolean setWorldIfMatched(Matcher m, boolean housing) {
         if (m.find()) {
-            String worldName = m.group(1);
+            String worldName = housing ? currentWorldName : m.group(1);
```
kristofbolyai, line 165: This assumes that even when joining other worlds by joining houses, you get to see the updated world name for a small bit. I only assume this is correct because of the duplicate messages this PR aims to fix, which I assume is triggered by first, the world change, then by the the world changing to "housing".

ShadowCat117: No, this doesn't fix the changing worlds housing issue, only it changing the world name to the name of the housing, which then makes features that use current world name not work correctly. Server uptime overlay, bomb bell command/overlay, discord rich presence etc. 

So since `setState` checks the world names aren't equal before posting the world change event, it won't incorrectly post the event as the `newWorldName` is still set to `currentWorldName`

kristofbolyai: Okay, so it's not a full fix, but at least it patches the issue. Sure.

kristofbolyai (approved):

PR #2479, opened 2024-05-10
fix: Don't return null in profession last harvest functions [skip ci]

ShadowCat117:
Returning null makes the functions unusable until you harvest something, you can't even check for the "??" returned
image


kristofbolyai (approved):

magicus:
I think our Functions should never return null. Do we have any other instances of that, or was these the only ones?

ShadowCat117:
> I think our Functions should never return null. Do we have any other instances of that, or was these the only ones?

I believe these were the only ones

PR #2480, opened 2024-05-10
feat: Add custom names for info boxes and custom bars [skip ci]

ShadowCat117:
This is another feature that is in https://github.com/Wynntils/Artemis/pull/2335 but does not rely on textures at all so may as well release it separately.
Settings
OverlaySelection
OverlayManagement


common/src/main/java/com/wynntils/overlays/custombars/CustomBarOverlayBase.java
```diff
@@ -7,12 +7,16 @@
...
+import com.wynntils.core.consumers.overlays.CustomNameProperty;
 import com.wynntils.core.consumers.overlays.OverlaySize;
 import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.config.Config;
 import com.wynntils.utils.type.ErrorOr;
 
-public abstract class CustomBarOverlayBase extends BarOverlay {
+public abstract class CustomBarOverlayBase extends BarOverlay implements CustomNameProperty {
+    @Persisted(i18nKey = "feature.wynntils.customBarsOverlay.overlay.customBarBase.customName")
+    public final Config<String> customName = new Config<>("");
```
kristofbolyai, line 18: Maybe custom overlay name / overlay name fits better? I assume there would be an UI element for changing this, and the config would be hidden then?

ShadowCat117: Is there a need for it to be a hiddenconfig with a seperate ui element? Having it as a regular string config seems fine to me as it's with all of the other configs for overlays

kristofbolyai: Hm. I guess we can discuss that when we get to that. But we are going to have separate UI to change this, right?

ShadowCat117: I'm not following, why does this need a seperate ui?

kristofbolyai: I imagine the reworked setting screen would allow the custom name to be changed, in some way. Like double clicking to edit the name.

ShadowCat117: Hm, I suppose it could be done that way instead. If you think that's the best way to do it then I'll close this pr and take the overlay selection/management redesigns and make a new pr with those and include it in there as those textures are mostly fine

kristofbolyai: We can merge this as is now, and you can do the change when you do that PR. How does that sound?

ShadowCat117: That works

kristofbolyai (approved):

PR #2482, opened 2024-05-11
fix: Fix ItemGuessFeature tooltip line being fully underlined incorrectly [skip ci]

kristofbolyai:
# Old
image

# Fixed
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/tooltips/ItemGuessFeature.java
+++ b/common/src/main/java/com/wynntils/features/tooltips/ItemGuessFeature.java
@@ -1,9 +1,10 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.features.tooltips;
 
+import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Models;
 import com.wynntils.core.components.Services;
 import com.wynntils.core.consumers.features.Feature;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/tooltips/ItemGuessFeature.java
```diff
@@ -91,12 +97,14 @@ private List<Component> getTooltipAddon(GearBoxItem gearBoxItem) {
 
             guesses.append(StyledText.fromString("§7: ").getComponent());
 
-            Optional<MutableComponent> itemsComponent = itemsForLevel.stream()
-                    .reduce((i, j) -> i.append(Component.literal(", ").withStyle(ChatFormatting.GRAY))
-                            .append(j));
+            MutableComponent itemsComponent = Component.empty();
+            itemsComponent.append(itemsForLevel.get(0));
```
ShadowCat117, line 102: Won't this fail if `itemsForLevel` is empty? Maybe in the above check it should just append something like "Unknown item" or just skip to the next level

kristofbolyai: It should never be empty, but I forgot to add the continue after the logging above.

magicus:
What does the underlining signify?

kristofbolyai:
> What does the underlining signify?

Whether the item is favorited.

magicus:
Ah, I see. I have not used favourites that much. Maybe we should insert a yellow ☆ in front instead? I think that would be more visible and intuitive?

kristofbolyai:
> Ah, I see. I have not used favourites that much. Maybe we should insert a yellow ☆ in front instead? I think that would be more visible and intuitive?

We already have a star in front of the item itself, so underlining should be okay.

magicus:
Okay. :-)

PR #2483, opened 2024-05-11
feat: Backwards container searching and resume search after match [skip ci]

### LINKED ISSUE #1714: Bank Search Behaviour
In Legacy it was possible to search forward *and* backwards using the arrow buttons. It was also possible to continue search on the next page, even though the current page has a match.

Currently Artemis only has forward search, which means I always have to skip to the first page, then run the search from there. It also blocks search when the current one has a match, which requires me to jump to the next page before continuing the search.

So parity/improvement points are:

- add forward/backwards search on arrow buttons
- fix block by matches on current page

ShadowCat117:
I'm not certain how legacy handled the direction, whether it was by actually pressing the next/previous buttons but I think this works nicely.

Pressing enter will default to forward search unless on last page. Hovering the next/previous page slot will determine direction otherwise.

https://github.com/Wynntils/Artemis/assets/8337467/b37a2ad4-c8e1-4c0e-8171-f8d3f941f130

Closes https://github.com/Wynntils/Artemis/issues/1714


kristofbolyai:
Perhaps it should just do a full search, no backwards searching needed. Example: You are on page 5, search for something. It goes to the last page, 10, for example, then it uses the bookmarks to go to page 1, searches until page 5.

kristofbolyai:
But I am not saying you should not keep this, maybe just add last page -> first page support, and search until you re-hit the starter page. (And this conveniently will return you to the page you started the search on, if no items are found).

ShadowCat117:
Will only work for personal storage containers as they're the only things with the quick jumping. Could be interesting to have for them though

kristofbolyai:
> Will only work for personal storage containers as they're the only things with the quick jumping. Could be interesting to have for them though

Okay. Do you want to do that now, or different PR?

ShadowCat117:
> Okay. Do you want to do that now, or different PR?

I'll do it now, don't think it should be too difficult

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/7d29b492-22e8-4bf6-bff2-2862c0fbc443

There is a bit of overlap now with this, the quick jumping feature and the model, so I think I'll merge the quick jumping logic into the model in a later pr

ShadowCat117:
Don't merge yet, seems to be an issue with auto search not continuing for non personal storage containers

ShadowCat117:
Thinking about it, maybe it's not a good idea. I forgot to make it loop backwards, but that will act weirdly when your final page isn't one of the destinations.

If you start on page 1 and you search backwards with your final page as 19, it'll jump to 17 then go forward to 19, only to then go backwards towards 1

kristofbolyai:
> Thinking about it, maybe it's not a good idea. I forgot to make it loop backwards, but that will act weirdly when your final page isn't one of the destinations.
> 
> If you start on page 1 and you search backwards with your final page as 19, it'll jump to 17 then go forward to 19, only to then go backwards towards 1

Oh.. Yeah, this would only work then, if you had forwards search only, and no backwards search. Should we still do this then..? This method is slower, but searches 100% of the container. Your method can be faster (if guided by the user), but won't be "autonomous".

ShadowCat117:
Perhaps a good alternative is shift+enter to jump back to page 1 and start the search there

kristofbolyai:
> Perhaps a good alternative is shift+enter to jump back to page 1 and start the search there

Okay, sure.

kristofbolyai (approved):
Looks good to me.

kristofbolyai:
Maybe merge this with #2485?

PR #2484, opened 2024-05-11
feat: Redirect Merchant Messages [skip ci]

cmkohnen:
Filter Merchant Messages the likes of "Scroll Merchant: Thank you for your business. Come again!", as they can spam the chat quite quickly when buying multiple items


kristofbolyai (commented):
Can you move this into Message Redirect Feature? (Actually, I don't remember the feature name, but you will see what I am talking about. I want to make this change redirect to notification overlay, by default.

kristofbolyai (approved):
Thanks!

PR #2485, opened 2024-05-12
feat: Show item type for chat items, show perfect/defective

kristofbolyai:
image
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
@@ -26,15 +26,18 @@ import com.wynntils.models.items.WynnItem;
 import com.wynntils.models.items.encoding.type.EncodingSettings;
 import com.wynntils.models.items.items.game.GearItem;
 import com.wynntils.models.items.properties.GearTierItemProperty;
+import com.wynntils.models.items.properties.IdentifiableItemProperty;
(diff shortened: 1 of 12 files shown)


ShadowCat117 (approved):
Looks good!

magicus:
I wonder about the use of `makeObfuscated` here. The idea is that it should be used when we have control over the rendering, since each letter has a chance of getting replaced with a weird one. This works okay if they just flicker in and out of existence, but when sent to chat it gets permanent, so if you are unlucky one of the letters disappear and will never re-appear.

Instead I think you should just make defective items bold and dark red.

The rainbow variant is okay. While it will not move, the idea comes across and it will always be visible.

kristofbolyai:
> I wonder about the use of `makeObfuscated` here. The idea is that it should be used when we have control over the rendering, since each letter has a chance of getting replaced with a weird one. This works okay if they just flicker in and out of existence, but when sent to chat it gets permanent, so if you are unlucky one of the letters disappear and will never re-appear.
> 
> 
> 
> Instead I think you should just make defective items bold and dark red.
> 
> 
> 
> The rainbow variant is okay. While it will not move, the idea comes across and it will always be visible.

The chance of obfuscation is 0 here, I just used the bold/red part of the formatting.

magicus:
Ah, I see. Then it is all good! 👍

PR #2487, opened 2024-05-13
fix: Detect insulators and simulators in mythic found feature [skip ci]

ShadowCat117:
They are technically mythics so should be counted

### DIFF
--- a/common/src/main/java/com/wynntils/features/MythicFoundFeature.java
+++ b/common/src/main/java/com/wynntils/features/MythicFoundFeature.java
@@ -55,22 +55,23 @@ public class MythicFoundFeature extends Feature {
         }
 
         // Lootrun rewards
+        boolean validLootrunMythic = false;
         Optional<GearItem> gearItem = Models.Item.asWynnItem(itemStack, GearItem.class);
         if (gearItem.isPresent()) {
-            sendLootrunDryStreakMessage(
-                    StyledText.fromComponent(event.getMythicBoxItem().getHoverName()));
-            return;
+            validLootrunMythic = true;
         }
 
         Optional<InsulatorItem> insulatorItem = Models.Item.asWynnItem(itemStack, InsulatorItem.class);
         if (insulatorItem.isPresent()) {
-            sendLootrunDryStreakMessage(
-                    StyledText.fromComponent(event.getMythicBoxItem().getHoverName()));
-            return;
+            validLootrunMythic = true;
         }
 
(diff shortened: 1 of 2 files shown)


ShadowCat117:
Also I should note, I think unid shiny items are parsed as `UnknownGearItem` so the feature probably won't detect those either. I did try fixing it a while ago but couldn't get a good regex pattern for it

kristofbolyai:
> Also I should note, I think unid shiny items are parsed as `UnknownGearItem` so the feature probably won't detect those either. I did try fixing it a while ago but couldn't get a good regex pattern for it

Perhaps make an issue?

common/src/main/java/com/wynntils/features/MythicFoundFeature.java
```diff
@@ -52,11 +54,25 @@ public void onMythicFound(MythicFoundEvent event) {
...
+            sendLootrunDryStreakMessage(
+                    StyledText.fromComponent(event.getMythicBoxItem().getHoverName()));
+            return;
+        }
+
+        Optional<SimulatorItem> simulatorItem = Models.Item.asWynnItem(itemStack, SimulatorItem.class);
+        if (simulatorItem.isPresent()) {
+            sendLootrunDryStreakMessage(
+                    StyledText.fromComponent(event.getMythicBoxItem().getHoverName()));
         }
```
kristofbolyai: Is this the only place this check is used? Does the mythic found event fire for these items?

ShadowCat117: Oops yeah that does need it too

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -307,11 +309,26 @@ public void onEntitySpawn(SetEntityDataEvent event) {
...
+                    dryPulls.store(0);
+                    return;
+                }
 
-                if (gearItem.getGearTier() == GearTier.MYTHIC) {
+                Optional<SimulatorItem> simulatorItemOpt = Models.Item.asWynnItem(itemStack, SimulatorItem.class);
+                if (simulatorItemOpt.isPresent()) {
                     WynntilsMod.postEvent(new MythicFoundEvent(itemStack, true));
                     dryPulls.store(0);
                 }
```
kristofbolyai: Repeat yourself twice, but the third time :). Can you add a "mythicFound" boolean and do the remaining work there? Maybe do this in the feature too.

kristofbolyai (approved):

PR #2492, opened 2024-05-13
fix: Don't send unexpected disconnect on IO thread

magicus:
I got an unexpected disconnect while debugging mapdata, and it crashed Minecraft.

I am not sure this is the correct solution. As bolyai points out in https://github.com/Wynntils/Artemis/commit/c6fa5191fb6287bb64a3e7889b0422ca9fe7855c#r141951957, if this happens too early will get into another mess instead.


kristofbolyai:
I'll sleep on this to decide, it's not a critical fix anyways.

magicus:
Agreed, it's not critical. In fact, I'd like to see if I could even figure out a way to reproduce this situation. Maybe the correct solution is to remove the mixin and event.

magicus:
Consider this PR more as a starting point for a discussion on how we should handle unexpected stream closure, rather than a proper fix.

kristofbolyai:
See ticket-1584 on discord.

magicus:
@kristofbolyai You refreshed this? Think we should move forward with it?

kristofbolyai:
> @kristofbolyai You refreshed this? Think we should move forward with it?

Mainly I've done the merge because I've updated our CI to prepare for 1.21, and needed a format-and-build to pass, which this PR was convenient for. I do think this issue happens enough times to move forward, but I have not spent enough time thinking about a proper solution. Partially, that is because, when this bug happens, there is an interaction with Fabric API, that makes it think we don't initialize a screen properly, so it crashes the game (because we open the Custom Loading Screen from a different screen, or something like that..).

kristofbolyai:
I think this PR fixes more random crashes than it causes. If we have a disconnect failure before manager init, something went really wrong anyways.

kristofbolyai (approved):

PR #2493, opened 2024-05-13
feat: Add visibility handling to MapData [skip ci]

magicus:
I don't recall the last time I messed up a patch like I have done with this. :) So I'm starting all over again. This PR drops the level changes, and focuses on the visibility handling. It also adds a WaypointProvider, which was missing before. (I really wanted to make sure the visibility stuff worked as it should with user-defined waypoints, and collected loot chest locations.)

I also included MapFeaturePoiWrapper which I originally wrote as a kind of debug function, but I now realize will be essential in moving from the old world to the new. It is not used yet, but I'd rather have it in mainline.

This is not 100% done, not even for visibility. But it is far along enough that I want this reviewed and committed to mainline.


magicus:
It was a bit annoying that `customPoiConfig.touched()` did not result in `onConfigUpdate` being called, so I had to inject calls to `updateWaypoints` everywhere. The change to do that was technically simple, but I did not understand the consequences so I did not dare doing that as part of this PR. If/when we fix the config stuff properly, we should be able to remove this and rely on `touched()` being enough.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
```diff
@@ -30,6 +35,10 @@ public CategoriesProvider() {
         for (Label.LabelLayer layer : Label.LabelLayer.values()) {
             PROVIDED_CATEGORIES.add(new PlaceCategory(layer));
         }
+        for (int tier = 1; tier <= 4; tier++) {
```
kristofbolyai: Maybe `LootChestType.size()` should be used here, just in-case this is not updated, when a new loot chest tier is introduced.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
```diff
@@ -66,11 +77,121 @@ public String getIconId() {
...
+        public String getName() {
+            return "Personal Waypoints";
+        }
+
+        @Override
+        public MapAttributes getAttributes() {
+            return new AbstractMapAttributes() {
+                @Override
+                public int getPriority() {
+                    return 500;
```
kristofbolyai: How come waypoints are not max (1000) priority? Afaik, they render above everything else right now.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
```diff
@@ -0,0 +1,154 @@
...
+                    case CHEST_T4 -> 4;
+                    default -> 0;
+                };
+        if (tier == 0) {
+            String iconId = MapIconsProvider.getIconIdFromTexture(customPoi.getIcon());
+            PROVIDED_FEATURES.add(new WaypointLocation(
+                    customPoi.getLocation().asLocation(), customPoi.getName(), iconId, customPoi.getVisibility()));
+        } else {
+            PROVIDED_FEATURES.add(new FoundChestLocation(customPoi.getLocation().asLocation(), tier));
+        }
```
kristofbolyai, line 54: This really wants us to not collect loot chest locations to the personal POI list, right? Most likely we should move all loot chest locations to a new list of storage, and separate it here.

kristofbolyai (approved):
Overall I think this is good to go.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
magicus: The short answer is that the mapdata replacements are roughly equivalent, but need more scrutiny and fixes before we can be sure they will fully replace the current Poi counterparts.

In this case, you are likely correct that this should be 1000. I'll fix it, but we need to check all other mapdata sources as well before switching over.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
magicus: Yes, I think that was a huge mistake from out part (copied over from Legacy). The stuff we collect automatically should be kept separate from manually added waypoints. 

For instance, automatically collected chests might be checked over time if they still exist or needs to be removed, etc. With the new looting system, I guess it would make sense to store the timestamp of when such a chest was last visited, etc.

I was thinking about just ditching all these chests completely, but my assumption here is that anything with a tier X chest icon means it is automatically added (not necessarily correct) and that we should take ownership of it. I guess a more proper way to do this would be to run some kind of conversion that splits out chests from the rest of the waypoints, and ask the user what to do with them.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
magicus: Unlikely to happen, and that will break the textures as well... But sure, it might be more clear to read.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
kristofbolyai: Do you think we should do the separation now? We can just leave all old chests in the user waypoints, and collect them into a new list now. I know, it's less than ideal, but I don't think we want to deal with anything else. The users can delete these chests from their custom waypoints in even the old waypoint manager. We will render the newly loot chests, but they won't be editable for a short period.

kristofbolyai:
So, basically, I think we can merge this, and I (or you, if you have time, which I assume you don't) can follow up with the user waypoint-loot chest POI separation.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
magicus: Yes, we should start doing a separate collection as soon as possible. Then we could correlate the two sets and delete the "user waypoints" that match coordinate and icon type. I think that would be the best. But the sooner we start collecting, the better.

Otoh, I would also like to have the collected data in the new mapdata json format, so until that is finished, I think we should hold off collecting. Or, perhaps, we can store it as a HiddenConfig or Storage for the time being, and then convert it?

kristofbolyai: I am not sure how to make it be in the new format. But we can't upfix storage, so perhaps let's not do it atm.

magicus: Yeah, we can discuss that further on discord.

PR #2494, opened 2024-05-13
refactor: Better handling of MapData attributes [skip ci]

magicus:
I'm not still sure this is exactly how this should be done, but lets take that as a separate discussion from the visibility changes. This PR depends on #2493.


kristofbolyai:
So, this is basically from the other PR I did. What do we want to discuss here? This code works in-game. We've agreed to remove Label/LabelDeserializer during the migration, and make this change just for porting.

However, since we don't strictly need this for the migration, do we need even need this change? I think this is the question.

magicus:
This PR is not so much as a code change, as it is a marker for an unresolved design question.

The question boils down to:
* Should we differentiate between "unknown", "not applicable" and "suitable for all levels" for level information? All of these three? Some of them? None of them?
* If so, how should we express these values? 
* And how should we express that the value is inherited?

They are kind of separate questions but also highly intermingled.

I'm not sure myself what the best solution would be, but I am sure that we need to resolve this now, so we have a finalized design, before starting to migrate to MapData.

magicus:
I *think* the best way forward is the following:

1) Use `1` to mean "suitable for all levels"

2) Use Optional<Integer> to express the `getLevel()` function. If the optional is not present, this means "inherit". That corresponds to the `"level"` field being absent from the json file.

3) Use `0` to mean "unknown" or "not applicable", and don't differentiate further between those.

4) Set the default value to be `0`, so any json file that does not express a `"level"` field will inherit this.

kristofbolyai:
As other mapdata fields don't use Optional for "inherit", I am not convinced this one should. We either change all fields, or keep this one to the original proposal (1 - all level, 0 - inherit, -1 unknown). The current change has -1 unknown/all levels. I am not sure if that should be kept. I think that edge case should be handled by UIs, not mapdata.

EDIT: I agree that the default should be unknown (in this case -1, not 0, but that is just an impl. detail).

magicus:
>  We either change all fields, 

Yes, that might very well be what we should. I have been thinking along those lines.

kristofbolyai:
> > We either change all fields,
> 
> Yes, that might very well be what we should. I have been thinking along those lines.

So, should I go and implement the optionals and see how it looks?

magicus:
> So, should I go and implement the optionals and see how it looks?

I had a look at `MapAttributes`. That means adjusting just this too:
```
    // 1-1000, 1000 is highest priority (drawn on top of everything else)
    // 0 means no value specified; inherit
    int getPriority();
```

Maybe we should just return `Integer`s and stay with the style that "null means inherit". 

Or we should disallow null for all values (including String and CustomColor etc), and make them all `Optional`. I guess the latter is a more modern way, but if it results in too much boilerplate maybe we should not go down that route. Otoh, it will really make it clear that each individual attribute is optional. So maybe that is better, after all.

TL;DR: Go ahead and try to change all methods in MapAttributes return an `Optional` wrapper, and change all null checks into checks for `isPresent()` instead. And then we can see if this seems good. I hope it does, since it will make the code clearer and more resilient.

kristofbolyai:
So, do we like this, @magicus? Honestly, it might a bit too much. These are the times Java could really learn from other languages, like Rust.

magicus:
I kind of like it. But I'm not 100% sure. I'll need to sleep on it.

However, there are things we can do to make it better. Like `hasIcon(iconId.orElse(null)))`; this should be folded into the `hasIcon` method which should take the optional as argument, etc.

I am also personally a bit skeptical to the `map` function of Optional. I understand how it provides a more functional expression, but I find it hard to read. Maybe it is just lack of practice from my side, but I think a traditional `if (foo.isEmpty())` and a two-pronged return to be easier to read and understand.

kristofbolyai:
> Like `hasIcon(iconId.orElse(null)))`; this should be folded into the `hasIcon` method which should take the optional as argument, etc.

Sure. Although Optional arguments are a bit unconventional. 

> I am also personally a bit skeptical to the `map` function of Optional. I understand how it provides a more functional expression, but I find it hard to read. Maybe it is just lack of practice from my side, but I think a traditional `if (foo.isEmpty())` and a two-pronged return to be easier to read and understand.

Sure, we can get rid of the sugar syntax, I don't mind it either way.

magicus:
> Although Optional arguments are a bit unconventional.

Good point. It might be better to pass the `atttributes` to `hasIcon`.

common/src/main/java/com/wynntils/services/map/Label.java
```diff
@@ -41,29 +47,44 @@ public LabelLayer getLayer() {
...
     }
 
+    public static final class LabelDeserializer implements JsonDeserializer<Label> {
+        @Override
+        public Label deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context)
+                throws JsonParseException {
+            JsonObject jsonObject = json.getAsJsonObject();
+            JsonElement levelObject = jsonObject.get("level");
+
+            int level = -1;
```
magicus: I think this part keeps the old solution of having -1 instead of an empty Optional for a missing label. In fact, looking more closely, I think `getLevel()` and the object `level` field should return/be Optional<Integer>.

common/src/main/java/com/wynntils/services/map/pois/LabelPoi.java
```diff
@@ -130,13 +130,13 @@ public void renderAt(
                         getTextShadow(),
                         1f);
         if (hovered) {
-            String level = label.getLevel();
-            if (!level.isEmpty()) {
+            int level = label.getLevel();
```
magicus: ... And here we should check that getLevel is not empty first, before checking if it is >= 1.

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
```diff
@@ -63,17 +64,22 @@ public String getName() {
...
-            return (int) (icon.getWidth() * scale);
+        if (hasIcon()) {
+            Optional<MapIcon> icon = Services.MapData.getIcon(iconId.get());
+            if (icon.isPresent()) {
+                return (int) (icon.get().getWidth() * scale);
+            }
         }
 
-        if (hasLabel(label)) {
+        if (hasLabel(label.orElse(null))) {
```
magicus: Can you do the same thing with `hasLabel` as you did with `hasIcon`?

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
```diff
@@ -63,17 +64,22 @@ public String getName() {
...
             // Use label for measurements
-            return (int) (FontRenderer.getInstance().getFont().width(attributes.getLabel()) * scale);
+            return (int) (FontRenderer.getInstance()
+                            .getFont()
+                            .width(attributes.getLabel().orElse(""))
```
magicus: I wonder if `attributes.getLabel().orElse("")` is something that should be as a helper function. Not sure where, though. But it sort of feels like it would make sense to have a way to get a guarenteed string, even if it is empty. Otoh, if this is the only place this operation is done, then there is perhaps no big point.

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
```diff
@@ -63,17 +64,22 @@ public String getName() {
 
     @Override
     public int getWidth(float mapZoom, float scale) {
-        String iconId = attributes.getIconId();
-        String label = attributes.getLabel();
+        Optional<String> iconId = attributes.getIconId();
```
magicus: I have the feeling that these might be pushed down to much closer their actual use.

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
```diff
@@ -83,15 +89,17 @@ public int getWidth(float mapZoom, float scale) {
 
     @Override
     public int getHeight(float mapZoom, float scale) {
-        String iconId = attributes.getIconId();
-        String label = attributes.getLabel();
+        Optional<String> iconId = attributes.getIconId();
```
magicus: These as well I believe can move closer to their actual use.

common/src/main/java/com/wynntils/services/mapdata/attributes/DerivedAttributes.java
```diff
@@ -8,50 +8,51 @@
...
         return getAttribute(MapAttributes::getIconId);
     }
 
     @Override
-    public int getPriority() {
-        Integer integer = getAttribute(MapAttributes::getPriority);
-        return integer == null ? 0 : integer;
+    public Optional<Integer> getPriority() {
+        Optional<Integer> integer = getAttribute(MapAttributes::getPriority);
+        return integer.isEmpty() ? Optional.of(0) : integer;
```
magicus: Should we really do this, or just return ` getAttribute(MapAttributes::getPriority)`? (The same question applies to `getLevel` below.)

Judging from symmetry reason with the other methods, this looks a bit sus.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
```diff
@@ -6,32 +6,36 @@
...
-    int getPriority();
+    // 1-1000, 1000 is the highest priority (drawn on top of everything else)
+    // empty means no value specified; inherit
+    Optional<Integer> getPriority();
 
-    // 0 means no information is available, or suitable for all combat levels
-    int getLevel();
+    // the minimum combat level for which this feature is suitable for
+    // 0 means suitable for all levels
+    // -1 means no information is available
```
magicus:
```suggestion
    // 1 means suitable for all levels
    // 0 means no information is available, or level is not applicable
```

magicus:
It seems like we are not (yet?) using the priority information from MapData. I looked for it to see that the case of an empty optional was properly handled, but could not find any instance. We just need to make sure that when we do handle it that we give it a proper default value (e.g. 500) if it returns empty.

magicus:
I think this is finally starting to look like all pieces are put together. Yes, it becomes more verbose with Optional, but it also much more clear, and we do not need to forget to check for null somewhere along the line.

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
kristofbolyai: If the need comes up when I'm implementing anything related, I'll add the method.

common/src/main/java/com/wynntils/services/mapdata/attributes/DerivedAttributes.java
kristofbolyai: You already commented about not having a default priority. This should be set to it. This 0 is basically the fallback value, which is correct for level, but not correct here. I'll set the default 500 here.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
kristofbolyai: I agree with the comment change, but why the level change? I think having 0 and -1 as special cases is much more common than 1, even if in this case, 1 can be used, as it's the default level. But overall, I think, 1 does not suggest a special case, so it's not worth using it here.

common/src/main/java/com/wynntils/services/map/pois/LabelPoi.java
```diff
@@ -130,13 +131,13 @@ public void renderAt(
                         getTextShadow(),
                         1f);
         if (hovered) {
-            String level = label.getLevel();
-            if (!level.isEmpty()) {
+            Optional<Integer> level = label.getLevel();
+            if (level.isPresent()) {
```
magicus:
```suggestion
            if (level.isPresent() && level.get() >= 1) {
```

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MapAttributes.java
magicus: I thought we had this discussion before in discord, and that we had resolved it..?

I really dislike using -1 as a special number for level. I think 0 much better fills the role of saying "level do not apply, or we don't know". You can never be level 0, so it is free to use for this purpose. The reason we're adding "1 means suitable for all levels" is not about demarcating 1 as a special number, but about helping users to understand how they should express "suitable for all".

common/src/main/java/com/wynntils/services/mapdata/attributes/DerivedAttributes.java
magicus: There's still something odd with this. We need a default value, yes, but it seems to be put in the wrong place. I think the default value should be provided by FullFeatureAttributes if it fails to dig up any other value from the inherited map attributes.

I'll have a look again at the code in the IDE, this PR view makes me crazy.

magicus:
I have some ideas that are easier if I just code than try to explain. I'll hijack this PR for a while, I'll let you know when I'm done.

magicus:
This became quite the refactoring! 😮 

But now I am finally fully happy with how mapdata attributes work. The insight that *all* MapAttribute values must be Optional was the key to unlocking a better way of thinking of this.

Along the way, I also realized what we did was resolving the actual values from the inherited values. The new name ResolvedMapAttributes further drove clarity on how this works. So now  I actually understand how the map attribute inheritance works. :-) It has gone from incomprehensible black magic to rather straightforward.

magicus:
I also throw in a better generalization of MapDecoration in the mix... It's still not used, though...

magicus:
I have now also tested it with my mapdata-debug additions, and verified that I can indeed override selected attributes, or selected values of a visibility attribute, using json files. (I discovered a couple of bugs in that process...)

magicus:
@kristofbolyai At this point, I am fully happy with this and want to merge it. It might be tricky to review as a diff in Github, I recommend looking at the actual source in the IDE to get a better understanding of how things ended up after this change.

kristofbolyai (approved):
I think this turned out to be nice! At this point, with the refined inheritance, I am thinking if optionals have any use, or they are just "useless" wrappers. I have a feeling they might have been left behind from the original idea, but with this refactor, it might not need to kept.

Either way, I'll try to port 1 or 2 providers to mapdata, and we'll see how they turn out.

magicus:
> At this point, with the refined inheritance, I am thinking if optionals have any use, or they are just "useless" wrappers. I have a feeling they might have been left behind from the original idea, but with this refactor, it might not need to kept.

I am not sure what you mean here really. The Optional is what kind of made this possible. The alternative would be to use null everywhere instead, but that easily becomes error prone. The optional allow any given "layer" in the inheritance to chose which parts of a MapAttribute that they implement. Maybe I am stupid, but I can't see how we can keep that functionality and get rid of the Optionals. (Unless, of course, we use null for this instead...)

magicus:
> Either way, I'll try to port 1 or 2 providers to mapdata, and we'll see how they turn out.

I have already ported several of them. We basically only have to remove the old ones and enable the new ones. 

Or did you mean convert the old json files to the new format?

kristofbolyai:
> > Either way, I'll try to port 1 or 2 providers to mapdata, and we'll see how they turn out.
> 
> 
> 
> I have already ported several of them. We basically only have to remove the old ones and enable the new ones. 
> 
> 
> 
> Or did you mean convert the old json files to the new format?

Yes, I meant converting the old ones. In my head, I assumed, this is necessary, basically porting is done when we get rid of "MapPoiWrapper".

magicus:
Well, yes and no. We can also port gradually, kind of... I have hooks in place (in my testing branch) that reads the current json files but creates MapData constructs. It is a simple transition, and it would allow to start using the new model and remove some old code, and give it some testing. I suggest we start there.

Also, the Wrapper class was supposed to be a stop-gap solution, but after the latest round of cleaning up it turned out to be pretty okay. The rendering method will need to live somewhere. But I agree we will not be using the rest of the heavy Poi framework, so we should try to transmogrify it into a class that basically only contains the render method.

kristofbolyai:
> > At this point, with the refined inheritance, I am thinking if optionals have any use, or they are just "useless" wrappers. I have a feeling they might have been left behind from the original idea, but with this refactor, it might not need to kept.
> 
> 
> 
> I am not sure what you mean here really. The Optional is what kind of made this possible. The alternative would be to use null everywhere instead, but that easily becomes error prone. The optional allow any given "layer" in the inheritance to chose which parts of a MapAttribute that they implement. Maybe I am stupid, but I can't see how we can keep that functionality and get rid of the Optionals. (Unless, of course, we use null for this instead...)

My thinking was that, since we provide default values for all the optionals when resolving them, what's the point in having them be an optional. However, that doesn't work as I imagined it would (as the default values are not directly inherited as a superclass, rather, they are just provided by a "dummy" default implementation. So let's just go with what's here, and if anything comes to mind when porting, I'll bring it up.

kristofbolyai:
> Well, yes and no. We can also port gradually, kind of... I have hooks in place (in my testing branch) that reads the current json files but creates MapData constructs. It is a simple transition, and it would allow to start using the new model and remove some old code, and give it some testing. I suggest we start there.
> 
> 
> 
> Also, the Wrapper class was supposed to be a stop-gap solution, but after the latest round of cleaning up it turned out to be pretty okay. The rendering method will need to live somewhere. But I agree we will not be using the rest of the heavy Poi framework, so we should try to transmogrify it into a class that basically only contains the render method.
> 
> 

I think that many of the POI sources need to be adapted in Static-Storage, so generators create the new format. That will probably be the first step for me. After separating loot chests from custom pois.

magicus:
> My thinking was that, since we provide default values for all the optionals when resolving them, what's the point in having them be an optional. 

Yes, I see what you mean. ResolvedMapAttributes could return values that are not wrapped in optional. The thought crossed my mind. The flipside is that if it does that, it cannot implement MapAttributes. There were some nice synergy from having DerivedMapAttributes implement MapAttributes, and then having ResolvedMapAttributes extend that. But it is perhaps possible to change that.

PR #2495, opened 2024-05-14
WIP: MapData visibility inheritance refactoring

magicus:
This is just to publish my current work when playing around and trying to find a better model for mapdata visibility inheritance.


magicus:
@kristofbolyai If you want, you can take a look at this and see if you can figure out a way to inherit individual parts of the visibility object. Basically, I think we should have a "root visibility" of 0, 100, 6. And then if you just want to change the min value to 32, you should basically do 
```
var vis = new MapVisibiliy() { 
    @override 
    public Optional<Integer> getMin() {
        return 32;
    }
}
```

or in json just having like: 
```
...
"iconVisibility": {
   "min": 32
} 
....
```

or something like that.

I think we need to have something along the lines of how attributes work, were each level adds just the "overrides" it needs, and then we have a "collect all" function that traverses the tree upwards and builds a complete structure.

And we might want to separate between a top-level MapVisibility object that we can rely on always returning `int`s, and the overriding versions that return `Optional<Integer>` since they might not return anything at all, which mean that they will override.

magicus:
... Thinking of which, this sounds an awful lot like the level discussion of being able to not return a value at all, and meaning that you want to inherit (vs returning a value of 0 meaning "not applicable").

magicus:
To be clear, I think we need to resolve this before we can finalize the MapData structure. But I will not have time to work on it for a while. So we did get a step closer with my previous PR, but not all the way. :(

magicus:
Eh, what happened? Must have mis-clicked.

magicus:
Ah, right, the target branch was deleted. And now I can't re-target this PR. Brilliant, github. :-/

kristofbolyai:
I'll try to figure out a solution.

magicus:
Replaced by https://github.com/Wynntils/Artemis/pull/2500 which is opened against main.

PR #2497, opened 2024-05-14
fix: fix damage stat collecting neutral damage only

kristofbolyai:


magicus:
Should we collect the elemental damages individually as well? I imagine it could be interesting to know, for instance if you have received a lot more of thunder damage than other types, it indicates that at least historically you'd have been better off with wearing protection from thunder.

kristofbolyai:
> Should we collect the elemental damages individually as well? I imagine it could be interesting to know, for instance if you have received a lot more of thunder damage than other types, it indicates that at least historically you'd have been better off with wearing protection from thunder.

This is damage you deal, so it's not important for defenses. I think the most important part is to just collect the total damage, whether we collect them separately also, is just a matter of someone coding it.

magicus:
Ah, I see. Hm, I really feel the statistics system is underutilized... We should collect damage received as well. Oh well.

PR #2500, opened 2024-05-15
refactor: MapData visibility inheritance refactoring [skip ci]

magicus:
This is just to publish my current work when playing around and trying to find a better model for mapdata visibility inheritance.


magicus:
@kristofbolyai If you want, you can take a look at this and see if you can figure out a way to inherit individual parts of the visibility object. Basically, I think we should have a "root visibility" of 0, 100, 6. And then if you just want to change the min value to 32, you should basically do 
```
var vis = new MapVisibiliy() { 
    @override 
    public Optional<Integer> getMin() {
        return 32;
    }
}
```

or in json just having like: 
```
...
"iconVisibility": {
   "min": 32
} 
....
```

or something like that.

I think we need to have something along the lines of how attributes work, were each level adds just the "overrides" it needs, and then we have a "collect all" function that traverses the tree upwards and builds a complete structure.

And we might want to separate between a top-level MapVisibility object that we can rely on always returning `int`s, and the overriding versions that return `Optional<Integer>` since they might not return anything at all, which mean that they will override.

... Thinking of which, this sounds an awful lot like the level discussion of being able to not return a value at all, and meaning that you want to inherit (vs returning a value of 0 meaning "not applicable").

kristofbolyai:
The JSON format can be solved by using a custom GSON deserailiser. What about using "builder" pattern, like we do for styles? So you could say `MapVisibility.DEFAULT.withMin(10)`. Adding this is easy and is mostly just moving the default to the visibility class. We could also make it a record.

magicus:
Problem is that you don't know the parent, and would like to just override one variable. And the parent could perhaps be changed by another override on the category.

Much like how the entire Attributes work. That's why I am saying that this would be perfectly simple if we just inline the three variables in the Attributes, instead of wrapping them in a Visibility object. Maybe that is just what we need to do, to be able to push ahead with this. It will make the attributes a bit flat, with too many objects and some code duplication, but at least it will solve the problem quickly.

Otoh, I really would like to get this right. Maybe I'm overly protected. This has become my "baby" and I might need to kill my darling. I dunno. :-/

kristofbolyai:
Hm, okay, I'll take a closer look.

magicus:
I'm pretty sure we will need something like `DerivedAttributes` and `FullFeatureAttributes`, but for `MapVisibility` instead. I'm not sure if we can re-use anything, or if we just must copy and modify these. Anyway, I suggest you start looking there for inspiration.

kristofbolyai:
I'll also note the first attempt at doing this here, in case we ever want to come back and read the PRs again. https://github.com/magicus/Artemis/pull/24

magicus:
I think this is ready to go in now. I don't really know who should review this since this is produced by both @kristofbolyai and me, but I guess if we both approve of each others work it is fine. :-)

kristofbolyai (approved):
Okay. Let's go with this.

PR #2502, opened 2024-05-15
feat: Overlay selection & management screen redesign

ShadowCat117:
Probably the last thing to come separately from https://github.com/Wynntils/Artemis/pull/2335 as these textures are at a quality I'm happy to see released.

It's basically the same implementation from that pr but now using the same smooth scrolling as the settings screen and also the custom name config was changed to be hidden and replaced with left clicking the selected overlay to "replace" the button in the list with a text input widget which can be saved by pressing enter.


common/src/main/java/com/wynntils/screens/overlays/selection/CustomBarSelectionScreen.java
```diff
@@ -0,0 +1,242 @@
...
+import net.minecraft.client.gui.components.Button;
+import net.minecraft.client.gui.components.Tooltip;
+import net.minecraft.client.gui.components.events.GuiEventListener;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.network.chat.Component;
+import net.minecraft.sounds.SoundEvents;
+import org.lwjgl.glfw.GLFW;
+
+public class CustomBarSelectionScreen extends WynntilsScreen {
+    // The order of these matters, they must match that in CustomBarsOverlayFeature
```
kristofbolyai, line 34: Why is that?

common/src/main/java/com/wynntils/screens/overlays/selection/widgets/OverlayButton.java
```diff
@@ -0,0 +1,305 @@
...
+            WynntilsMod.error("Failed to delete, overlay group not found for overlay " + overlay.getJsonName());
+            return;
+        }
+
+        // Delete the overlay
+        Managers.Overlay.removeIdFromOverlayGroup(overlayGroupHolder, overlayId);
+
+        // Reload config
+        Managers.Config.loadConfigOptions(true, false);
+        Managers.Config.saveConfig();
```
kristofbolyai: The other two use cases also call `ConfigManager#reloadConfig`. Why do we not that here?

kristofbolyai (commented):
Comments on usability:
- We need a remove button for the custom bars/info overlays
- Selecting built-in/custom should unfocus the currently selected overlay, if they are not of the selected type

Optional, but "maybe we should do it now":
- Add a way to resize an overlay, preserving it's aspect ratio (CTRL, perhaps?)

kristofbolyai:
This is all, though! Looks really good :)

common/src/main/java/com/wynntils/screens/overlays/selection/CustomBarSelectionScreen.java
ShadowCat117: Because when the bar is actually added, it needs the index of it in the list of OverlayGroupHolder's so it uses `barTypeIndex` to track which bar type you currently have selected so that the same index can be used when getting the group for the specific bar type

common/src/main/java/com/wynntils/screens/overlays/selection/widgets/OverlayButton.java
ShadowCat117: Think I just forgot it :)

ShadowCat117:
> We need a remove button for the custom bars/info overlays

Shift-right clicking the overlay button currently deletes them

> Add a way to resize an overlay, preserving it's aspect ratio (CTRL, perhaps?)

Sure I can try and do that

common/src/main/java/com/wynntils/screens/overlays/selection/CustomBarSelectionScreen.java
kristofbolyai: You could use the class of the overlay, which the holder provides from a getter. That way, you can save the overlay type in this variable, and we won't break it ever by forgetting to update this.

kristofbolyai:
> > We need a remove button for the custom bars/info overlays
> 
> 
> 
> Shift-right clicking the overlay button currently deletes them

We have nice buttons for adding them, probably we can slot a remove currently selected button.

common/src/main/java/com/wynntils/screens/overlays/selection/CustomBarSelectionScreen.java
ShadowCat117: I would like to think we won't forget to update this as if any new bar types are ever added they will probably use this screen to add it and will notice it not working then

kristofbolyai: Okay, let's skip this for now.

kristofbolyai: Even if we add a new type, it probably won't cause issues with the indexing, as it'll be added after the last holder.

kristofbolyai:
Let me know if you are done with all changes you want to do here.

ShadowCat117:
I haven't tried the aspect ratio yet, I'll do that tomorrow

common/src/main/java/com/wynntils/screens/overlays/selection/CustomBarSelectionScreen.java
ShadowCat117: I've added a comment to the feature just in case

ShadowCat117:
Not been able to get the aspect ratio working, I can revisit it a later time if nobody else has done it by then.

I think that and then eventually adding a "Minecraft" tab to the selection screen to allow for moving and disabling things like the hotbar and xp bar are the only things these screens still need

kristofbolyai:
One last thing. The free move/edit buttons don't return to the overlay screen correctly, resetting any unsaved changes to the overlay + it has to be selected again. (unlike with the preview button).

ShadowCat117:
> One last thing. The free move/edit buttons don't return to the overlay screen correctly, resetting any unsaved changes to the overlay + it has to be selected again. (unlike with the preview button).

It remains selected for me, not sure why that's happening. As for saving the changes what would be the best way to handle that do you think, either a force save happens when you click edit/free move or I can try and ignore calling `populateConfigs()` when returning from the management screen

kristofbolyai:
> As for saving the changes what would be the best way to handle that do you think, either a force save happens when you click edit/free move or I can try and ignore calling `populateConfigs()` when returning from the management screen

I think saving is fine.

kristofbolyai (approved):

PR #2503, opened 2024-05-15
fix: Parse unidentified shiny items correctly [skip ci]

### LINKED ISSUE #2488: Unidentified Shiny Items are not annotated correctly
GearItem currently expects the shiny symbol to be before the "Unidentified".

image

kristofbolyai:
I think this is all we need to fix https://github.com/Wynntils/Artemis/issues/2488.

image

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/annotators/game/GearAnnotator.java
+++ b/common/src/main/java/com/wynntils/models/items/annotators/game/GearAnnotator.java
@@ -4,6 +4,7 @@
  */
 package com.wynntils.models.items.annotators.game;
 
+import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Models;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.handlers.item.GameItemAnnotator;
@@ -17,7 +18,7 @@ import net.minecraft.world.item.ItemStack;
 
 public final class GearAnnotator implements GameItemAnnotator {
     private static final Pattern GEAR_PATTERN = Pattern.compile(
-            "^(?<rarity>§[5abcdef])(?<unidentified>Unidentified )?(?:§f⬡ )?(?:§[5abcdef])?(?:Shiny )?(?<name>.+)$");
+            "^(?:(?<unidrarity>§[5abcdef])(?<unidentified>Unidentified ))?(?:§f⬡ )?(?<idrarity>§[5abcdef])?(?:Shiny )?(?<name>.+)$");
 
     @Override
     public ItemAnnotation getAnnotation(ItemStack itemStack, StyledText name) {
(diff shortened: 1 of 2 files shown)


kristofbolyai:
Actually, it's not how it works..

common/src/main/java/com/wynntils/models/items/annotators/game/GearAnnotator.java
```diff
@@ -16,8 +16,8 @@
 import net.minecraft.world.item.ItemStack;
 
 public final class GearAnnotator implements GameItemAnnotator {
-    private static final Pattern GEAR_PATTERN =
```
magicus, line 19: Maybe add a regexp test?

kristofbolyai: My gradle setup is slightly broken and I can't run tests, and I have yet to fix that...

ShadowCat117:
Doesn't work for id'd shinies and marks them as common. I think that's the solution I tried and ended up with this result
image

kristofbolyai:
I'll just separate the unid and not unid regexes then.

ShadowCat117:
Unfortunately I think that's the only solution :/

kristofbolyai:
Actually, I think I can do 1 pattern.

kristofbolyai:
@JamieCallan117 Should work now.

common/src/main/java/com/wynntils/models/items/annotators/game/GearAnnotator.java
magicus: Oh, that's really bad. :( I think this is a case where a test really would be beneficial, especially seeing as how this has been going back and forth with different strings that failed in one way or another.

I had a similar problem with a dev environment that I had customized yet were broken in some way. What I did was just clone the repo to another dir and open as a separate project in IntelliJ. I recommend you do that, and just run the test.

kristofbolyai: I just pushed the tests, and if they fail, we will see it on the github workflow.

ShadowCat117 (approved):
Looks good, just add the `// Test in GearAnnotator_GEAR_PATTERN` and it's good to go

kristofbolyai:
> Just add the `// Test in GearAnnotator_GEAR_PATTERN` and it's good to go

We still do that?

ShadowCat117:
I think all of the other test cases have it

magicus:
I'm not quite sure why we still do that, but I think we argued that it helps folks remember to update tests, since we are not really into the habit of always running tests, or that all regexes has tests to it is second nature to always update them.

magicus (approved):
And thanks for adding the test!

PR #2505, opened 2024-05-17
fix: Detect TCC failure correctly [skip ci]

ShadowCat117:
Two raids with fail titles, one with a chat message and one with neither 🙃 
image


kristofbolyai (approved):

kristofbolyai:
Wynn moment :)

PR #2507, opened 2024-05-18
feat: Content book QoL changes, speed fixes, (better) partial caching 

kristofbolyai:
With this PR, we cache all of the currently parsed content book (quests, mini-quests, all discoveries, caves) per character. We can't yet figure out, when to re-parse the book, and when there were no changes, but this PR is a step in the direction for that. Theoretically, we have enough hints in the chat to mark a class outdated, but that is a too complex change here.

All filter buttons are now also saved into storage, to allow for easier usage of the content book. Also, switching from different types of activity screens will now cancel all background queries correctly, avoiding really long wait times.

### DIFF
--- a/common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/WynntilsContentBookFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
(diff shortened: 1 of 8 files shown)


ShadowCat117 (approved):
Looks good and definitely feels nicer to use

magicus:
Nice! A step in the direction of having the quest book as a Storage per class. I'm pretty sure we could do awesome things if we have a fast, guaranteed up-to-date representation of the Content book that is always stored.

PR #2510, opened 2024-05-20
fix: NPC dialogues not getting transcribed [skip ci]

kristofbolyai:
This is less than ideal, but the last patch made it so the NPC overlay messages cannot be transcribed at all. Since adapting the code to send a chat message with the transcription (but only when the overlay is enabled, as the other two methods are unaffected) is very hard, I think this is a nice compromise.


ShadowCat117:
What I had it do before was if the overlay was enabled it would just send a seperate message in chat that could be hovered. And leave the original message in the overlay. 

Could that not be done? Or will it not work if the translation feature is enabled

kristofbolyai:
I don't want to complicate it further, and yes, compatibility wise this is a better option.

ShadowCat117:
Ok, bit unfortunate that it has to be this way but better than it not transcribing at all

ShadowCat117 (approved):

PR #2511, opened 2024-05-20
chore: Add Storage upfixers [skip ci]

kristofbolyai:
If I had a dollar, every time we needed storage upfixers... I could probably eat a menu in McDonalds.

Anyways, this was really simple to add, thanks to every previous refactor of the persisted values, by magicus. This will be useful in the future for keeping our code clean, allowing us to do necessary refactors. It'll also be helpful when we are migrating map-data.

The upfixers however can't interact between storage types, you can't move configs to storage. That's a whole different story :).

I've tested this by applying a storage upfixer to a renamed list of crowd sourced data, which is our most complex storage type. I've also tested applying a config upfixer, to make sure they still work.

### DIFF
--- a/common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
+++ b/common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
@@ -12,13 +12,17 @@ import com.wynntils.core.components.Manager;
 import com.wynntils.core.components.Managers;
 import com.wynntils.core.mod.event.WynncraftConnectionEvent;
 import com.wynntils.core.persisted.Persisted;
+import com.wynntils.core.persisted.PersistedValue;
+import com.wynntils.core.persisted.upfixers.UpfixerType;
 import com.wynntils.utils.mc.McUtils;
 import java.io.File;
 import java.lang.reflect.Field;
 import java.lang.reflect.Type;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -37,6 +41,8 @@ public final class StorageManager extends Manager {
     private final Map<String, Storage<?>> storages = new TreeMap<>();
     private final Map<Storage<?>, Type> storageTypes = new HashMap<>();
     private final Map<Storage<?>, Storageable> storageOwner = new HashMap<>();
+
+    private JsonObject storageObject;
 
     private long lastPersisted;
     private boolean scheduledPersist;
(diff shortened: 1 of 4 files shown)


ShadowCat117 (approved):
Looks good to me but probably wait for magicus' review as I'm not too well versed in the config/storage system

common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
```diff
@@ -60,6 +66,16 @@ public void initFeatures() {
 
         readFromJson();
 
+        // Now, we have to apply upfixers, before any storage loading happens
+        // FIXME: Solve generics type issue
```
magicus, line 70: What is the generics type issue? Do config upfixers have it?

magicus (approved):

magicus:
> If I had a dollar, every time we needed storage upfixers... I could probably eat a menu in McDonalds.

🤣 

I think that is closer to the core point here than you perhaps imagined... We have run into this from time to time, but not often enough to make this fix highly prioritized.

> Anyways, this was really simple to add, thanks to every previous refactor of the persisted values, by magicus. 

You're welcome. :) I made like the easy 60% of the unification between Config and Storage, but then I got stuck on the hard remaining 40%. I'm glad at least this part was not so difficult.

It would be great if we could finish up that refactoring. I recently ran into a problem were config.touch() did not trigger the onConfigUpdate() method as I had expected. (I don't recall right now what I was trying to do, but I had changed a value in a config that was a collection, and that did not automatically update it so I had to call touch()). I started looking into it but quickly descended down the rabbit hole of all different kind of "set value" functions that I was too scared to unify, and this somehow went through a path that persisted the value but did not call updating. I dared not change this either due to unforeseen consequences, so I ended up making a separate call to update the config. And it all became a mess and I never even posted a PR.

magicus:
Anyway, this PR looks fine as far as I can tell from remembering the code. :)

common/src/main/java/com/wynntils/core/persisted/storage/StorageManager.java
kristofbolyai: Yes, this is the same thing. In fact, you wrote this :). It could be solved by upfixers taking in `List<? extends PersistedValue<?>>` instead of `List<PersistedValue<?>>` but I did not want to do that. Java generics, I guess..

kristofbolyai:
> It would be great if we could finish up that refactoring. I recently ran into a problem were config.touch() did not trigger the onConfigUpdate() method as I had expected. (I don't recall right now what I was trying to do, but I had changed a value in a config that was a collection, and that did not automatically update it so I had to call touch()). I started looking into it but quickly descended down the rabbit hole of all different kind of "set value" functions that I was too scared to unify, and this somehow went through a path that persisted the value but did not call updating. I dared not change this either due to unforeseen consequences, so I ended up making a separate call to update the config. And it all became a mess and I never even posted a PR.

Oh yeah. Changing anything config related is almost guaranteed to break something related to the onConfigUpdate methods... I am not sure if they should be kept in the long run, when they are used, they are almost always problem source.

PR #2515, opened 2024-05-21
feat: Add Guild Functions

cmkohnen:
Implement a few functions related to guilds. This enables to do for example the following:
Bildschirmfoto vom 2024-05-21 18-55-47


common/src/main/java/com/wynntils/functions/GuildFunctions.java
```diff
@@ -0,0 +1,88 @@
...
+public class GuildFunctions {
+    public static class CappedGuildLevelPercentageFunction extends Function<CappedValue> {
+        @Override
+        public CappedValue getValue(FunctionArguments arguments) {
+            return new CappedValue(Models.Guild.getGuildLevelPercentage(), 100);
+        }
+
+        @Override
+        protected List<String> getAliases() {
+            return List.of("capped_guild_level_percentage");
```
ShadowCat117: This goes for all the functions, but the default for this will automatically be `capped_guild_level_percentage` so you don't need to do this.

common/src/main/java/com/wynntils/functions/GuildFunctions.java
```diff
@@ -0,0 +1,88 @@
...
+        @Override
+        public Integer getValue(FunctionArguments arguments) {
+            return Models.Guild.getGuildLevelPercentage();
+        }
+
+        @Override
+        protected List<String> getAliases() {
+            return List.of("guild_level_percentage");
+        }
+    }
```
ShadowCat117: We don't really do these functions anymore as you can get the current percentage from the CappedValue function using `current(capped_guild_level_percentage))`

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1409,6 +1409,8 @@
   "function.wynntils.bps.name": "Player Speed",
   "function.wynntils.bpsXz.description": "Player speed in blocks per second excluding vertical movement",
   "function.wynntils.bpsXz.name": "Player Speed (Without Y)",
+  "function.wynntils.cappedGuildLevelPercentage.description": "The amount of XP aquired by your Guild to reach the next Guild Level",
```
ShadowCat117:
```suggestion
  "function.wynntils.cappedGuildLevelPercentage.description": "The XP progress of your guild",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1699,6 +1701,16 @@
   "function.wynntils.generic.z.argument.location": "The location to extract the coordinate from",
   "function.wynntils.generic.z.description": "Extracts the Z coordinate from a location",
   "function.wynntils.generic.z.name": "Z Coordinate",
+  "function.wynntils.guildLevel.description": "The Level of the Guild you are in",
+  "function.wynntils.guildLevel.name": "Guild Level",
+  "function.wynntils.guildLevelPercentage.description": "The amount of XP aquired by your Guild to reach the next Guild Level",
+  "function.wynntils.guildLevelPercentage.name": "Guild Level Percentage",
+  "function.wynntils.guildMember.description": "Whether or not you are in a Guild",
+  "function.wynntils.guildMember.name": "Guild Member",
+  "function.wynntils.guildName.description": "The Guild you are currently in",
```
ShadowCat117:
```suggestion
  "function.wynntils.guildName.description": "The name of the Guild you are currently in",
```

common/src/main/java/com/wynntils/functions/GuildFunctions.java
```diff
@@ -0,0 +1,88 @@
...
+import com.wynntils.core.consumers.functions.arguments.FunctionArguments;
+import com.wynntils.models.players.type.GuildRank;
+import com.wynntils.utils.type.CappedValue;
+import java.util.List;
+
+public class GuildFunctions {
+    public static class CappedGuildLevelPercentageFunction extends Function<CappedValue> {
+        @Override
+        public CappedValue getValue(FunctionArguments arguments) {
+            return new CappedValue(Models.Guild.getGuildLevelPercentage(), 100);
```
ShadowCat117: You should return `CappedValue.EMPTY` if a guild was not found, otherwise it'll be returning -1/100

ShadowCat117 (approved):
Looks good to me, thanks for the contribution!

cmkohnen:
I just noticed some implementation details, that can be improved, marking as draft

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -192,36 +190,37 @@ public void onContainerClick(ContainerClickEvent e) {
...
+            WynntilsMod.info("Skipping silverbull subscription query ("
+                    + (silverbullExpiresAt.get() - System.currentTimeMillis()) + " ms left)");
+        }
+        // Open Guild Menu
+        queryBuilder.then(QueryStep.clickOnSlot(GUILD_MENU_SLOT)
+                .expectContainerTitle(ContainerModel.GUILD_MENU_NAME)
+                .processIncomingContainer(Models.Guild::parseGuildContainer));
 
-        query.executeQuery();
+        queryBuilder.build().executeQuery();
```
cmkohnen: Okay, so the problem I ran into is that running multiple Queries has probably a race condition, so runnig three queries resulted simoultaneously resulted in them failing most of the time.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -71,7 +74,8 @@ public final class CharacterModel extends Model {
     private boolean reskinned;
     private int level;
 
-    private boolean isVeteran = false;
+    @Persisted
+    private final Storage<Boolean> isVeteran = new Storage<>(false);
```
cmkohnen, line 78: Also this wouldnt get checked if silverbull is still active, so I made it persistent

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -117,13 +132,36 @@ public void parseGuildInfoFromGuildMenu(ItemStack guildInfoItem) {
...
+        Matcher levelProgressMatcher = LoreUtils.matchLoreLine(membersItem, 0, LEVEL_PROGRESS_MATCHER);
+        if (!levelMatcher.matches()) {
+            WynntilsMod.warn("Could not parse guild level progress from item: " + LoreUtils.getLore(membersItem));
+            return;
+        }
+        long current = Long.parseLong(levelProgressMatcher.group("current").replace(",", ""));
+        long required = Long.parseLong(levelProgressMatcher.group("required").replace(",", ""));
+        guildLevelProgress = new CappedValue((int) (((double) current / required) * 100d), 100);
+
+        WynntilsMod.info("Successfully parsed guild info for guild " + guildName);
```
cmkohnen, line 164: Parsing this screen has multiple benefits:
1. It sometimes takes a few seconds for the InfoBar to update on world change, so this is faster in 90% of the cases.
2. Having this in place allows for even more functions in the future (e. g. amount of held territories, boost level, members completed their objective, etc.) and more importantly to track tributes, which are currently not calculated in the territory management screen.

cmkohnen (commented):
This took a bit longer than expected and the complexity of this pr grew a bit, but the functionality has remained the same.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
magicus: Queries should be placed in queue, so only one is executed at a time. If this is not working properly, it is a bug with the query system. With that said, your solution here seems more elegant than the old one.

magicus (approved):
From what I can tell this looks good, but you'll need a review from someone more experienced in the area too.

common/src/main/java/com/wynntils/functions/GuildFunctions.java
```diff
@@ -0,0 +1,50 @@
...
+            return Models.Guild.getGuildLevel();
+        }
+    }
+
+    public static class GuildMemberFunction extends Function<Boolean> {
+        @Override
+        public Boolean getValue(FunctionArguments arguments) {
+            return !Models.Guild.getGuildName().isEmpty();
+        }
+    }
```
kristofbolyai: Is this strictly needed? You could use the GuildName function here. Generally, we don't want to add such functions that essentially do the same, as 2-3 functions chained together.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: I agree with this, good catch.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -192,36 +190,37 @@ public void onContainerClick(ContainerClickEvent e) {
...
+                    .then(QueryStep.clickOnSlot(COSMETICS_BACK_SLOT)
+                            .expectContainerTitle(ContainerModel.CHARACTER_INFO_NAME));
+        } else {
+            WynntilsMod.info("Skipping silverbull subscription query ("
+                    + (silverbullExpiresAt.get() - System.currentTimeMillis()) + " ms left)");
+        }
+        // Open Guild Menu
+        queryBuilder.then(QueryStep.clickOnSlot(GUILD_MENU_SLOT)
+                .expectContainerTitle(ContainerModel.GUILD_MENU_NAME)
+                .processIncomingContainer(Models.Guild::parseGuildContainer));
```
kristofbolyai: Can you click on the guild item, if you are not in a guild? Can you "expect" a container?

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -192,36 +190,37 @@ public void onContainerClick(ContainerClickEvent e) {
...
+            queryBuilder
+                    .then(QueryStep.clickOnSlot(COSMETICS_SLOT)
+                            .expectContainerTitle(ContainerModel.COSMETICS_MENU_NAME)
+                            .processIncomingContainer(this::parseCratesBombsCosmeticsContainer))
+                    .then(QueryStep.clickOnSlot(COSMETICS_BACK_SLOT)
+                            .expectContainerTitle(ContainerModel.CHARACTER_INFO_NAME));
+        } else {
+            WynntilsMod.info("Skipping silverbull subscription query ("
+                    + (silverbullExpiresAt.get() - System.currentTimeMillis()) + " ms left)");
+        }
```
kristofbolyai, line 213: It may be beneficial to add new lines before and after this block to reduce the cognitive load on future readers.

common/src/main/java/com/wynntils/models/players/GuildModel.java
kristofbolyai: I originally did not want to add more queries on class join, to reduce the time it needed for us to "init" the class, on very laggy servers/connections. I am still entirely convinced that we should add more queries. But I know that we can parse a lot from this container. So, the best thing we can do is optimize the amount of "clicks" needed by the container query.

common/src/main/java/com/wynntils/functions/GuildFunctions.java
cmkohnen: Is there a way to return a boolean value if a string is empty or insert literals into the expression? Cause I couldn't find any functionaility in this regard.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
cmkohnen: This is something not actually tested by me, as I don't have an alt account nor want to leave my guild.
But my assumption is, that it would "fail" the parsing when the container isn't presented, with everything else already parsed at this point.

common/src/main/java/com/wynntils/functions/GuildFunctions.java
ShadowCat117: Yes, you would use `eq_str(guild_name;"")` to check for no guild

common/src/main/java/com/wynntils/models/players/GuildModel.java
cmkohnen: An alternative way of accessing this container would be "/guild manage", however I found that this was a bit unreliable because commands such as "/friend list" and "/use" would be run at the same time

common/src/main/java/com/wynntils/functions/GuildFunctions.java
cmkohnen: ah ty

common/src/main/java/com/wynntils/models/character/CharacterModel.java
ShadowCat117: This is the item for no guild and clicking it does nothing
image

common/src/main/java/com/wynntils/models/players/GuildModel.java
kristofbolyai: Running commands are no better, delay wise. Let's go with this for now.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: Making this work without expected failure is a blocking change required for this PR. The query system should have the necessary conditional query steps that allow you to fix this.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -192,36 +190,42 @@ public void onContainerClick(ContainerClickEvent e) {
...
 
-                // Open /use menu
-                .then(QueryStep.sendCommand("use")
-                        .expectContainerTitle(ContainerModel.COSMETICS_MENU_NAME)
-                        .processIncomingContainer(this::parseCratesBombsCosmeticsContainer))
-                .build();
+        // Open Guild Menu
+        queryBuilder.conditionalThen(
+                (container) -> !Models.Guild.getGuildName().isEmpty(),
```
kristofbolyai, line 218: How would the model have a guild name, if the guild menu is yet to be parsed?

You should use the container to check for the guild item being an "active" guild item.

cmkohnen: Correct me if I'm wrong, but the parsing of the guild name happens already in `parseCharacterContainer()`. This predicate will get executed when the guild container should be opened, meaning after the main character container has already been parsed (and the guild item too, from which guild name and rank are obtained). Checking the guild item again, would add unnecessary complexity, no?

kristofbolyai: Yes, that's right, actually. Can you add a short version of this is a comment? It's hard to expect any reader to know which guild related info is parsed here and which is parsed before.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: I think we don't want for the container to close, we just start the next query after we sent the close container packet. I'll check more thoroughly later.

kristofbolyai (approved):
Looks good.

PR #2516, opened 2024-05-22
feat: Highlight bombs active in current world [skip ci]

cmkohnen:
A small change highlighting bombs in the current world:
Bildschirmfoto vom 2024-05-22 14-39-51


ShadowCat117 (approved):
I was thinking of doing this during the dxp weekend but hadn't gotten around to it yet, so thanks! :)

PR #2517, opened 2024-05-22
fix: Account for timezone in player commands [skip ci]

ShadowCat117:
Before
image

After
image


magicus:
Ooohhh, time zone stuff is difficult. Can you explain a bit more what you are fixing and why it is correct?

ShadowCat117:
> Ooohhh, time zone stuff is difficult. Can you explain a bit more what you are fixing and why it is correct?

The api returns a timestamp of when the player was last online in UTC and the commands were just comparing that to current time but not taking the timezone into account, so this just does the same thing we do for territory held as that also displays the "acquired" time in UTC

magicus:
Hm. So the server gives you a valid date+time in UTC, right? Do we say in the Model when reading this that the time is in UTC?

Something still feels a bit wrong here. I don't like that the conversation is done in the presentation layer. Maybe there is some better `java.time` function that can give you the elapsed time between the current time and the value given in a proper, timezone-annotated time object?

ShadowCat117:
I suppose it could be in the deserializer, but this is the same as how `TerritoryProfile` handles timezones only I'm using `System.currentTimeMillis()` instead of `new Date().getTime()`

```
private long getTimeHeldInMillis() {
    return new Date().getTime() - this.getAcquired().getTime() + getTimezoneOffset();
}

private long getTimezoneOffset() {
    return ((long) new Date().getTimezoneOffset() * 60 * 1000);
}
```

magicus:
Ok, I'll accept this for now, but I think there is room for improvement here. We should be sure to always include timezone info and then this should not be something we need to think about.

magicus (approved):

kristofbolyai:
Oh, we don't even parse the date in the model? We just keep it as a a string?

magicus:
Oh, is *that* the problem? 😮

common/src/main/java/com/wynntils/models/players/type/GuildInfo.java
```diff
@@ -21,11 +26,13 @@ public record GuildInfo(
         int xpPercent,
         int territories,
         long wars,
-        String createdTimestamp,
+        Date createdTimestamp,
         int totalMembers,
         int onlineMembers,
         List<GuildMemberInfo> guildMembers) {
     public static class GuildDeserializer implements JsonDeserializer<GuildInfo> {
+        private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.ROOT);
```
magicus: I think this should be a locale specifying UTC. I am not sure Locale.ROOT does that..? Or if it does, please say something like:
`// Wynncraft sends times in UTC, which is the timezone specified for Locale.ROOT`

common/src/main/java/com/wynntils/models/players/type/WynnPlayerInfo.java
```diff
@@ -22,12 +27,14 @@ public record WynnPlayerInfo(
...
         String server,
-        String lastJoinTimestamp,
+        Date lastJoinTimestamp,
         String guildName,
         String guildPrefix,
         GuildRank guildRank,
-        String guildJoinTimestamp) {
+        Date guildJoinTimestamp) {
     public static class WynnPlayerInfoDeserializer implements JsonDeserializer<WynnPlayerInfo> {
+        private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.ROOT);
```
magicus: ... same here ...

common/src/main/java/com/wynntils/models/players/type/GuildInfo.java
```diff
@@ -9,9 +9,14 @@
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
+import com.wynntils.core.WynntilsMod;
 import java.lang.reflect.Type;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
```
magicus: `java.util.Date` is severely outdated. You should only use classes and functionality from the new `java.time` package.

common/src/main/java/com/wynntils/commands/PlayerCommand.java
```diff
@@ -154,18 +149,14 @@ private int lookupPlayerLastSeen(CommandContext<CommandSourceStack> context) {
...
-                                System.currentTimeMillis() - joinedDate.getTime() + getTimezoneOffset();
-
-                        response.append(Component.literal(" was last seen ").withStyle(ChatFormatting.GRAY))
-                                .append(Component.literal(DATE_FORMATTER.format(differenceInMillis))
-                                        .withStyle(ChatFormatting.GOLD)
-                                        .append(Component.literal("ago").withStyle(ChatFormatting.GRAY)));
-                    } catch (ParseException e) {
-                        WynntilsMod.error("Error when trying to parse player last join", e);
-                    }
+                    long differenceInMillis = System.currentTimeMillis()
```
magicus, line 146: I'm pretty certain that there is a better way to get the elapsed time from a proper java.time timestamp and the current system time, than doing manual calculations and correcting for time zone etc. I don't recall the API details, but have a look at e.g. `Duration`.

common/src/main/java/com/wynntils/models/players/type/GuildInfo.java
ShadowCat117: The locale isn't relating to timezone here

> locale – the locale whose date format symbols should be used

magicus: Ok. I got a bit confused there. Anyway, here is a better way to do this:
https://stackoverflow.com/questions/6543174/how-can-i-parse-utc-date-time-string-into-something-more-readable

magicus: You should return this as an `Instant`, not a `Date`. An `Instant` basically means "timestamp with suitable metadata" in java.time. The second major component is `Duration`, which describes a period of time between two instants.

I guess we have been a bit sloppy with using these properly, but we really really should be using only the new java.time API. I would even like to add a code check rule for Wynntils that bans java.util.Date. It is just bad.

ShadowCat117: Alright, I'll see if I can fix up all the places that uses it currently, it's only a few places

magicus: I have a vague memory about some comment claiming that we needed to use an outdated `Date` construct due to one reason or another. I'm not sure that reason is valid (it can be that it just needs some more effort to convert to a proper API call), but maybe it is -- if so, you can keep it like that. But if you can fix it too, that'd be great. `<insert meme here>`

ShadowCat117: How's https://github.com/Wynntils/Artemis/pull/2517/commits/203fe2752c261954ed365d2e903cf47de936e57b? If that looks good I'll apply that elsewhere

kristofbolyai: LGTM

ShadowCat117:
java.util.Date has been eradicated

common/src/main/java/com/wynntils/services/lootrunpaths/LootrunPathFileParser.java
```diff
@@ -121,10 +124,12 @@ public static LootrunSaveResult writeJson(UncompiledLootrunPath activeLootrun, F
...
-            json.addProperty(
-                    "date",
-                    DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, Locale.US)
-                            .format(new Date()));
+            ZonedDateTime time = ZonedDateTime.ofInstant(Instant.now(), ZoneId.systemDefault());
+
+            DateTimeFormatter formatter =
+                    DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM).withLocale(Locale.US);
+
+            json.addProperty("date", time.format(formatter));
```
kristofbolyai, line 132: Can this load the old format properly? I think this is the place @magicus mentioned, where we needed to have a kind of backward-compatibility.

common/src/main/java/com/wynntils/utils/LongDateFormatter.java
```diff
@@ -4,53 +4,45 @@
...
-            Calendar.HOUR_OF_DAY, "h",
-            Calendar.MINUTE, "m",
-            Calendar.SECOND, "s");
-
-    public LongDateFormatter() {
-        setCalendar(Calendar.getInstance());
-        setTimeZone(TimeZone.getTimeZone("UTC"));
-    }
 
+public class LongDateFormatter extends DateFormatter {
```
kristofbolyai: Maybe we can merge this with SimpleDateFormatter? EIther by adding a "long" boolean in the constructor, or by adding a parameter in the method.

kristofbolyai (commented):
Looking really good overall!

common/src/main/java/com/wynntils/services/lootrunpaths/LootrunPathFileParser.java
ShadowCat117: I think I did check that the format was the same but I'll give it a double check.

Do we support legacy lootruns I can't remember, they might be different

kristofbolyai: I think we can load them, if they don't have notes.

ShadowCat117: The paths I have on my main instance load fine so they should be compatible

common/src/main/java/com/wynntils/utils/LongDateFormatter.java
ShadowCat117: Sure, we'll just have the one DateFormatter

common/src/main/java/com/wynntils/services/lootrunpaths/LootrunPathFileParser.java
ShadowCat117: In fact, I think this is the only place we actually use the date and it's just for saving, it's never read anywhere

magicus (approved):
Thanks for cleaning this up! Looks much better now.

kristofbolyai (approved):
Thank you!

PR #2518, opened 2024-05-22
feat: Add Ophanim Bar Overlay [skip ci]

cmkohnen:
Adds an Ophanim Bar Overlay

Bildschirmfoto vom 2024-05-22 17-59-47
Bildschirmfoto vom 2024-05-22 18-06-18

This includes some changes to the parent class to allow ability bars to render arbitrary text.

### DIFF
--- a/common/src/main/java/com/wynntils/models/abilities/bossbars/OphanimBar.java
+++ b/common/src/main/java/com/wynntils/models/abilities/bossbars/OphanimBar.java
@@ -4,6 +4,7 @@
  */
 package com.wynntils.models.abilities.bossbars;
 
+import com.wynntils.core.WynntilsMod;
 import com.wynntils.handlers.bossbar.TrackedBar;
 import com.wynntils.models.abilities.type.OphanimOrb;
 import java.util.ArrayList;
@@ -11,6 +12,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import net.minecraft.ChatFormatting;
 
 public class OphanimBar extends TrackedBar {
     // Test in OphanimBar_OPHANIM_PATTERN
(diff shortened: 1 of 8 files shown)


magicus:
Ophanim? I really must go and play Wynncraft a bit again, apparently stuff is happening there that I've completely missed.

kristofbolyai:
> Ophanim? I really must go and play Wynncraft a bit again, apparently stuff is happening there that I've completely missed.

This is a spell variation/archetype in the mage ability tree from the 2.0 update. I assume you haven't played mage yet :)

magicus:
Have basically not played at all since 2.0. :-) While I started working on Wynntils out of mostly self-interest, at this point it is more of a pet project in and of itself for me.

I do hope to go back to playing Wynncraft at some time, and then I'll be glad that there is a great mod supporting it. ;-)

common/src/main/java/com/wynntils/models/abilities/type/OphanimOrb.java
```diff
@@ -0,0 +1,44 @@
...
+        return healthState.getColorCode() + "⏺";
+    }
+
+    public enum HealthState {
+        HEALTHY("§b"),
+        DAMAGED("§c"),
+        DYING("§e"),
+        DEAD("§7");
+
+        private final String colorCode;
```
kristofbolyai: It would be better to store the underlying ChatFormatting here, and convert it into string, when needed.

common/src/main/java/com/wynntils/models/abilities/bossbars/OphanimBar.java
```diff
@@ -0,0 +1,62 @@
...
+
+        OphanimOrb.HealthState healthState = null;
+        int start = 0;
+        while (orbMatcher.find(start)) {
+            String colorCode = orbMatcher.group("color");
+            if (colorCode != null) {
+                healthState = OphanimOrb.HealthState.fromColorCode(colorCode);
+            }
+
+            orbs.add(new OphanimOrb(healthState));
```
kristofbolyai, line 62: What if the color code was null? `healthState` is then either null or in the state if the last orb. I think you need to handle that. If the regex "doesn't allow" colorCode to be null, just add a comment or a `WynntilsMod.warn` call to make sure we prevent bugs, if anything get changed.

cmkohnen: We expect the color code to not exist sometimes.

This has to do with `Component` to `StyledText` conversion. Wynncraft sends a color code before every orb icon, but the conversion to StyledText removes redundant code formatting. So if three consecutive Orbs are of red color for example, only the first one has a color code prepended. This logic works around that by caching the health state of the last orb and only if a new color code prefix is present inside the pattern update it.

So in theory the healthState should not be null as we expect the first orb to have a color code prefix.

Will definitely add a warning and maybe some explanatory comments as well

kristofbolyai: > Will definitely add a warning and maybe some explanatory comments as well

This is the most important part :). I know StyledText, in fact, I wrote it, but still, one can't comprehend all of it's "side-effects" when reading code using it, like here.

kristofbolyai (approved):
Looks good! Thank you.

ShadowCat117 (approved):
Looks good!

PR #2520, opened 2024-05-23
feat: Allow showing territory owner [skip ci]

cmkohnen:
Yet another small change allowing to show the current territory's owner in the territory overlay, matching the Info Bar.

### DIFF
--- a/common/src/main/java/com/wynntils/overlays/minimap/TerritoryOverlay.java
+++ b/common/src/main/java/com/wynntils/overlays/minimap/TerritoryOverlay.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.overlays.minimap;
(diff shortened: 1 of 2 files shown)


ShadowCat117:
You might want to change the function to handle when you are not inside a territory, otherwise you'll end up with `[]`

ShadowCat117:
I also wonder if it would be worth adding a config to show the full guild name instead, maybe then put the name on a separate line

cmkohnen:
> You might want to change the function to handle when you are not inside a territory, otherwise you'll end up with `[]`

oops, thanks for noticing.

The three options now are:
Bildschirmfoto vom 2024-05-23 18-21-43
Bildschirmfoto vom 2024-05-23 18-21-59
Bildschirmfoto vom 2024-05-23 18-22-06

With just the territory being the default still

common/src/main/java/com/wynntils/overlays/minimap/TerritoryOverlay.java
```diff
@@ -22,18 +25,34 @@ public TerritoryOverlay() {
...
+        return showOwner.get().getTemplate();
     }
 
     @Override
     public String getPreviewTemplate() {
-        return TEMPLATE;
+        return getTemplate();
+    }
+
+    public enum TerritoryOwnerDisplay {
```
ShadowCat117:
```suggestion
    private enum TerritoryOwnerDisplay {
```

kristofbolyai (approved):
Looks good, although I wouldn't add too many of these, I think we should provide sensible defaults, then everyone can customize their UI to their needs. Either way, extending this overlay to have the guild owner is a nice touch.

ShadowCat117:
> Looks good, although I wouldn't add too many of these, I think we should provide sensible defaults, then everyone can customize their UI to their needs.

I think these 3 options are nice, if someone wants both name and tag or for it to be coloured when the owner matches their guild then that should be in a custom overlay

PR #2521, opened 2024-05-23
feat: Allow hiding the infobar [skip ci]

cmkohnen:
Allow to hide the info bar.
Default value is set to disabled, as I imagine not every user wants this enabled.

### DIFF
--- a/common/src/main/java/com/wynntils/core/consumers/features/FeatureManager.java
+++ b/common/src/main/java/com/wynntils/core/consumers/features/FeatureManager.java
@@ -137,6 +137,7 @@ import com.wynntils.features.ui.CustomLoadingScreenFeature;
 import com.wynntils.features.ui.CustomSeaskipperScreenFeature;
 import com.wynntils.features.ui.CustomTerritoryManagementScreenFeature;
 import com.wynntils.features.ui.CustomTradeMarketResultScreenFeature;
+import com.wynntils.features.ui.HideInfoBarFeature;
 import com.wynntils.features.ui.LobbyUptimeFeature;
 import com.wynntils.features.ui.ProfessionHighlightFeature;
 import com.wynntils.features.ui.SoulPointTimerFeature;
@@ -350,6 +351,7 @@ public final class FeatureManager extends Manager {
         registerFeature(new CustomSeaskipperScreenFeature());
         registerFeature(new CustomTerritoryManagementScreenFeature());
         registerFeature(new CustomTradeMarketResultScreenFeature());
+        registerFeature(new HideInfoBarFeature());
         registerFeature(new LobbyUptimeFeature());
         registerFeature(new ProfessionHighlightFeature());
         registerFeature(new SoulPointTimerFeature());
--- a/common/src/main/java/com/wynntils/features/ui/HideInfoBarFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/HideInfoBarFeature.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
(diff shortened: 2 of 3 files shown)


magicus:
I wonder if this is the right way to do it. It feels like we tack on something downstream that should be dealt with closer to the actual action bar handling.

Also, I think it is not advisable to avoid staying away from too many negatives. Or to put it in other words: A "Show infobar" with default `true` is better...

cmkohnen:
> I wonder if this is the right way to do it. It feels like we tack on something downstream that should be dealt with closer to the actual action bar handling.

As far as I understand the codebase, this event is specifically used to track new bars being added and to track whether or not they should be rendered.

> Also, I think it is not advisable to avoid staying away from too many negatives. Or to put it in other words: A "Show infobar" with default true is better...

Will change that

ShadowCat117:
I did mention in https://github.com/Wynntils/Artemis/pull/2502 that eventually the overlay selection/management screen should be able to handle vanilla overlays and I think being able to move and disable boss bars should be done there.

It will require some effort though so this is fine for now until that has been setup.

cmkohnen:
> I did mention in #2502 that eventually the overlay selection/management screen should be able to handle vanilla overlays and I think being able to move and disable boss bars should be done there.

How about a custom bossbar overlay, that allows users to change position, width and texture of the bars? That would be something I'd be interested in exploring how to implement later.
That could also then come with custom bars such as just Guild Level as well as an added bar if bombs are active in the world (instead of them being in the same bar). This would also obviously supersede this feature then.

cmkohnen:
> Also, I think it is not advisable to avoid staying away from too many negatives. Or to put it in other words: A "Show infobar" with default true is better...

Is there a way to still listen to events if a feature is disabled? Because how it works now is that if the feature is enabled all events regarding adding the infobar get automatically cancelled. If the feature is disabled however the event will not be propagated to the listener in the feature.
So I have no idea how to reverse the handling visually without some hacky workarounds

ShadowCat117:
> How about a custom bossbar overlay, that allows users to change position, width and texture of the bars? That would be something I'd be interested in exploring how to implement later. That could also then come with custom bars such as just Guild Level as well as an added bar if bombs are active in the world (instead of them being in the same bar). This would also obviously supersede this feature then.

I think being able to move and disable the vanilla bars is a better option as there will always be people who prefer the original styled bars, but may want to put them in a different place.

Being able to create your own bars or just regular info boxes with the information the hidden bars currently display would still be a worthwhile addition imo. For example, with your recent additions I think we have functions for all of the information from the info bar, except bombs but the bomb bell overlay does have that, so the option to disable that bar and functions to still view that information however the user wants is great.

Unfortunately that isn't the case for the other bars we hide currently, I know someone was recently requesting mana bank functions as currently they are forced to use our overlay or the vanilla bar. The options to hide the vanilla and our custom one and use functions to make their own is the ideal situation

kristofbolyai:
> I think being able to move and disable the vanilla bars is a better option as there will always be people who prefer the original styled bars, but may want to put them in a different place.

Agreed.

Overall, I think there is room for hiding vanilla bars, or any other GUI elements, but I think, it should be integrated into the overlay system.

kristofbolyai:
So, maybe the best option would be to not do this right now, and we will try to add the overlay infrastructure for this soon.

What do you think @ShadowCat117 @cmkohnen?

ShadowCat117:
I think that would be better, as otherwise this is just going to be a temporary feature until it can be supported as an overlay later

cmkohnen:
As of now, I'm just a drive-by contributor, so I'll leave this decision entirely up to you. I'm fine with either way.

kristofbolyai:
Okay, let's not do this right now. I'll see if I can work on adding this the proper way, soon.

PR #2523, opened 2024-05-24
feat: Improvements to profession highlight feature for crafting stations

ShadowCat117:
Just makes crafting stations only have the option for the relevant profession type, not much point highlighting armouring ingredients in the scribing station for example


common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
kristofbolyai, line 1: Perhaps we should not add the button for these containers.

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
kristofbolyai, line 1: Most definitely do not create these by hand. Supply the profession types as the constructor, we don't need the separate classes. That way, you can create the instances with a for loop.

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
ShadowCat117: I think it's worthwhile to see the items in your inventory highlighted as they could be from different prof types

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
ShadowCat117: Ah like how the stat providers does it, yeah that's better

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
```diff
@@ -182,6 +183,19 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
 
             ProfessionType profession = feature.getSelectedProfession();
 
+            // For crafting stations, the only option should be the profession type for the station
+            if (Models.Container.getCurrentContainer() instanceof CraftingStationContainer craftingStation) {
+                if (profession == null) {
+                    profession = craftingStation.getProfessionType();
+                } else {
```
cmkohnen, line 198:
```suggestion
                } else if (profession != craftingStation.getProfessionType()) {
```
Otherwise this logic would nullify the currently selected profession if it matches the crafting station, no?

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
kristofbolyai: I meant that we either highlight a single type, or don't. I think we might as well highlight it always, and get rid of the button in these cases.

common/src/main/java/com/wynntils/models/containers/ContainerModel.java
ShadowCat117: Hmm, only issue with this is that the per container stores them via the class name so it'll be storing all types under "CraftingStation"

kristofbolyai: Can't you override the getName method?

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
ShadowCat117: Ah right, yeah that works

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
ShadowCat117: This will cause you to never be able to disable the highlight which wasn't the original intent but I'll be changing it to work like this now anyway

common/src/main/java/com/wynntils/features/ui/ProfessionHighlightFeature.java
ShadowCat117: Hmm actually I think having the toggle is nice, if you know you only have the relevant type on you but you want to see the tier, you'd have to hover all the ingredients or disable the feature

kristofbolyai: Okay, let's keep this then :).

kristofbolyai:
One problem is, that you need to handle global mode. Example: You set the filter to armouring in your inventory, it does not make sense in the crafting containers. Perhaps make it so if it's "any", then use the profession type, if it's "none", then keep it none.

ShadowCat117:
I think this works nicely.

If none is selected, opening a crafting station remains on none. If anything else was selected, when you open a crafting station it will now set the global to that stations type

kristofbolyai (approved):

PR #2524, opened 2024-05-24
feat: Add InStream function [skip ci]

ShadowCat117:
Useful for info boxes that show world name, since that no longer gets overwritten by the "-"


kristofbolyai (approved):
Do we have a default overlay that can "leak" the world? If so, please change it to use this function. If we don't, feel free to merge this.

ShadowCat117:
I already changed the server uptime and player list overlays to not display it, can't find any others that use it.

I think the only thing that can leak it is the bomb bell overlay as that tracks the info bar even without champion, so someone with champion could find you that way, but we don't want to display "Profession XP on -" for that

kristofbolyai:
> I think the only thing that can leak it is the bomb bell overlay as that tracks the info bar even without champion, so someone with champion could find you that way, but we don't want to display "Profession XP on -" for that

Okay, we can deal with that later, when we add current world bomb functions.

PR #2525, opened 2024-05-24
feat: Add option to limit number of leaderboard badges shown [skip ci]

ShadowCat117:
Rendering so many at once is it a bit excessive and especially since we'll soon be adding more badges this felt like it should be implemented. I set a default of 7 as I think that's a reasonable amount but it can support anything from 1 to however many badges there are.

It scrolls through the badges every 40 ticks, so it's not like any badges outside of the range aren't viewable.

Default 7
image
Original 13 for comparison
image

https://github.com/Wynntils/Artemis/assets/8337467/b1b2036a-03a8-4106-a64e-8557d5755415

### DIFF
--- a/common/src/main/java/com/wynntils/features/players/CustomNametagRendererFeature.java
+++ b/common/src/main/java/com/wynntils/features/players/CustomNametagRendererFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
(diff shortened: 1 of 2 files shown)


kristofbolyai (approved):

kristofbolyai:
I'm sure we will need to improve our leaderboard/badge model too.

ShadowCat117:
> I'm sure we will need to improve our leaderboard/badge model too.

Yes, I was going to do a bit of that too but I'll save it for when athena actually has the information

PR #2526, opened 2024-05-24
feat: Add guild objective functions [skip ci]

cmkohnen:
Allows the following:
Bildschirmfoto vom 2024-05-24 18-57-02

Notice that I haven't tested a player completing guild objectives and new objectives being assigned in game yet.


kristofbolyai (commented):
Essentially this should be good, but someone should also take a closer look.

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -119,6 +141,19 @@ public void onChatMessage(ChatMessageReceivedEvent e) {
             guildRank = GuildRank.valueOf(rankChangedMatcher.group(2).toUpperCase(Locale.ROOT));
             WynntilsMod.info("User's guild rank changed to " + guildRank);
```
ShadowCat117, line 142: Add a `return` to this and the objective completed if too

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -119,6 +141,19 @@ public void onChatMessage(ChatMessageReceivedEvent e) {
...
         }
+
+        Matcher objectiveCompletedMatcher = message.getMatcher(MSG_OBJECTIVE_COMPLETED);
+        if (objectiveCompletedMatcher.matches()) {
+            updateObjectivesCompletedProgress(objectivesCompletedProgress.current() + 1);
+            if (objectiveCompletedMatcher.group("player").equals(McUtils.playerName())) {
```
ShadowCat117, line 162: I'm fairly sure it does display the real name, but have you checked it doesn't display nicknames for players with one?

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -161,9 +197,29 @@ public void parseGuildContainer(ContainerContent container) {
...
+            }
+        }
+
         WynntilsMod.info("Successfully parsed guild info for guild " + guildName);
     }
 
+    private void updateObjectivesCompletedProgress(int completed) {
+        int currentGoal = 0;
+        for (int goal : OBJECTIVE_GOALS) {
+            if (completed > currentGoal) currentGoal = goal;
```
ShadowCat117: I might be wrong but I don't think we do variable assignments on the same line like this

common/src/main/java/com/wynntils/models/players/GuildModel.java
cmkohnen: I'm 99% sure that it will display the player name, as I logged a message testing the OphanimBar overlay of a nicked GuildMember completing their objective (which just displayed their username).

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -161,9 +197,29 @@ public void parseGuildContainer(ContainerContent container) {
         long required = Long.parseLong(levelProgressMatcher.group("required").replace(",", ""));
         guildLevelProgress = new CappedValue((int) (((double) current / required) * 100d), 100);
 
+        for (StyledText line : LoreUtils.getLore(objectivesItem)) {
+            Matcher objectivesCompletedMatcher = line.getMatcher(OBJECTIVES_COMPLETED_PATTERN);
+            if (objectivesCompletedMatcher.matches()) {
+                updateObjectivesCompletedProgress(Integer.parseInt(objectivesCompletedMatcher.group("completed")));
```
ShadowCat117: You can add a `continue` after this

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -161,9 +197,29 @@ public void parseGuildContainer(ContainerContent container) {
...
+            }
+
+            Matcher objectiveStreakMatcher = line.getMatcher(OBJECTIVE_STREAK_PATTERN);
+            if (objectiveStreakMatcher.matches()) {
+                objectiveStreak = Integer.parseInt(objectiveStreakMatcher.group("streak"));
```
ShadowCat117, line 228: I don't have a streak at the moment so can't check but if the streak line will always come after the objectives completed line you can `return` here. If the other comes first though then swap them round

cmkohnen: Do you mean `break`?

common/src/main/java/com/wynntils/models/players/GuildModel.java
ShadowCat117: I think you can also `break` when the goal has been set

common/src/main/java/com/wynntils/models/players/GuildModel.java
ShadowCat117: Yes, there's the info message afterwards, missed that

ShadowCat117:
Just a few minor things but good otherwise :)

common/src/main/java/com/wynntils/models/players/GuildModel.java
cmkohnen: break wouldn't work here.
Say a certain guild has 23 people who have completed their objective, you would first check against the first goal (5) and set `currentGoal` to 5. Then you would check agains 15 and set  `currentGoal` to 15. Finally you would check against 30 and set it to 30.

ShadowCat117: Can you parse the goal from the objectives completed pattern then? And then when you update it from the chat message just keep the same max.

We really should be able to update CappedValues without creating new instances of them, I think the soul point CappedValue has a comment about it

kristofbolyai: CappedValues are meant to be immutable, though.

ShadowCat117: Oh ok

cmkohnen: > Can you parse the goal from the objectives completed pattern then? And then when you update it from the chat message just keep the same max.

The problem I'm trying to solve with this function is that if by completing an objective the next milestone is reached, it should automatically update. During initial parse that can be skipped however, as it is right there

ShadowCat117: Ah of course, hadn't thought about that.

I mainly wanted it in case Wynn ever changed the goal amounts it would just automatically work but if that does happen we'll just have to update the list

ShadowCat117 (approved):

common/src/main/java/com/wynntils/models/players/GuildModel.java
cmkohnen: Although I can try optimising this a bit (redoing the logic in itself as well as parse the goal from the item initially)

kristofbolyai (approved):
Looks good! Thank you.

PR #2528, opened 2024-05-25
feat: Add guild diplomacy tracking

cmkohnen:
Track guild diplomacy.
- Enables tribute tracking from the diplomacy screen as well as chat.
- Adds an is_ally function for users to be able to replicate the info bar coloring via functions
- Adds `/wynntils rescan` to manually invoke character and guild scanning
- Parse silverbull only once on startup and then only when the welcome message is sent in chat


common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -213,13 +215,20 @@ private void scanCharacterInfo() {
...
+                .conditionalThen(
+                        guildMember,
+                        QueryStep.clickOnSlot(GUILD_MENU_SLOT)
+                                .expectContainerTitle(ContainerModel.GUILD_MENU_NAME)
+                                .processIncomingContainer(Models.Guild::parseGuildContainer))
+                .conditionalThen(
+                        guildMember,
+                        QueryStep.clickOnSlot(GuildModel.DIPLOMACY_MENU_SLOT)
+                                .expectContainerTitle(ContainerModel.GUILD_DIPLOMACY_MENU_NAME)
+                                .processIncomingContainer(Models.Guild::parseDiplomacyContainer));
```
cmkohnen: We need to scan this container for two reasons:
- Wynn doesnt show sent tributes in just the icon in contrast to recieved tributes
- We need to figure out which Guild is sending which tributes so we can accurately update information based on text messages

common/src/main/java/com/wynntils/screens/territorymanagement/widgets/GuildOverallProductionWidget.java
```diff
@@ -35,23 +36,33 @@ protected void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, flo
...
+        int emeraldProduction = holder.getOverallProductionForResource(GuildResource.EMERALDS)
+                + Models.Guild.getRecievedTributesForResource(GuildResource.EMERALDS);
+        int oreProduction = holder.getOverallProductionForResource(GuildResource.ORE)
+                + Models.Guild.getRecievedTributesForResource(GuildResource.ORE);
+        int woodProduction = holder.getOverallProductionForResource(GuildResource.WOOD)
+                + Models.Guild.getRecievedTributesForResource(GuildResource.WOOD);
+        int fishProduction = holder.getOverallProductionForResource(GuildResource.FISH)
+                + Models.Guild.getRecievedTributesForResource(GuildResource.FISH);
+        int cropsProduction = holder.getOverallProductionForResource(GuildResource.CROPS)
+                + Models.Guild.getRecievedTributesForResource(GuildResource.CROPS);
```
cmkohnen: Also I would like feedback if this should be done here or in the holder.
Or if tributes should be displayed separately alltogether.

cmkohnen (commented):
~~This is currently missing all tracking based on chat messages~~
done

common/src/main/java/com/wynntils/functions/GuildFunctions.java
```diff
@@ -48,6 +49,19 @@ public String getValue(FunctionArguments arguments) {
         }
     }
 
+    public static class IsAlliedFunction extends Function<Boolean> {
```
ShadowCat117: Maybe call this IsAlliedGuild and add `is_allied` as an alias, maybe even some others like `is_ally`

ShadowCat117:
All of the parsing looks good to me, I'll let bolyai decide on your displaying tributes question

fabric/src/test/java/TestRegex.java
```diff
@@ -363,6 +366,42 @@ public void GuildModel_MSG_NEW_OBJECTIVES() {
...
     }
 
+    @Test
+    public void GuildModel_MSG_TRIBUTE_SCEDULED() {
+        PatternTester p = new PatternTester(GuildModel.class, "MSG_TRIBUTE_SCEDULED");
+        p.shouldMatch("§3[INFO]§b ExamplePlayer1 scheduled 2 Emeralds per hour to Example Guild");
+        p.shouldMatch("§3[INFO]§b ExamplePlayer1 scheduled Ⓙ 1 Crop per hour to Example Guild");
+        p.shouldMatch("§3[INFO]§b ExamplePlayer1 scheduled Ⓚ 1 Fish per hour to Example Guild");
+        p.shouldMatch("§3[INFO]§b ExamplePlayer1 scheduled Ⓒ 1 Wood per hour to Example Guild");
+        p.shouldMatch("§3[INFO]§b ExamplePlayer1 scheduled Ⓑ 1 Ore per hour to Example Guild");
```
kristofbolyai: Does scheduling 2 crop"s" change the message? According to `GuildModel_MSG_TRIBUTE_STOPPED`, it might. If so, please add tests for it.

kristofbolyai: Also, check this with fish/wood too, crops is a special case for plurality.

kristofbolyai: And finally, make sure it can actually be parsed by the GuildProduction (or however) it is called enum, so we don't just match the line, but actually interpret it.

common/src/main/java/com/wynntils/screens/territorymanagement/widgets/GuildOverallProductionWidget.java
kristofbolyai: We can keep this here, I think.

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -87,12 +108,21 @@ public class GuildModel extends Model {
...
     private static final Pattern OBJECTIVE_STREAK_PATTERN = Pattern.compile("^§a- §7Streak: §f(?<streak>\\d+)$");
 
+    // Test in GuildModel_TRIBUTE_PATTERN
+    private static final Pattern TRIBUTE_PATTERN = Pattern.compile(
+            "^§[abef6](?<symbol>[ⒷⒸⓀⒿ]?) ?(?<amount>[+-]\\d+) (Ore|Wood|Fish|Crops?|Emeralds?) per Hour$");
+
     private static final int MEMBERS_SLOT = 0;
     private static final int OBJECTIVES_SLOT = 13;
+    public static final int DIPLOMACY_MENU_SLOT = 26;
+    private static final List<Integer> DIPLOMACY_SLOTS = List.of(2, 3, 4, 5, 6, 7, 8);
```
kristofbolyai, line 125: Are these fixed slots, do they change with guild level? Can you screenshot the container for me? (I can't login atm, sorry).

fabric/src/test/java/TestRegex.java
cmkohnen: > Does scheduling 2 crop"s" change the message?

Yes, both "Emeralds" and "Crops" have a special case for just a singular resource

> And finally, make sure it can actually be parsed by the GuildProduction (or however) it is called enum, so we don't just match the line, but actually interpret it.

To account for the plural s being missing, the GuildResource is aquired by matching the symbol the actual name of the resource is ignored in this case.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
kristofbolyai: As I've mentioned before, these queries are starting to get a bit too much. In fact, playing with the last patch already felt slower than before. However, I do understand that for the guild functionality, we need this parsing to happen. Give me a bit to think about how we can make this optional, or do this less times, etc.

fabric/src/test/java/TestRegex.java
kristofbolyai: Great, thanks! Can you add a final test case to show-case this?

common/src/main/java/com/wynntils/models/players/GuildModel.java
cmkohnen: Bildschirmfoto vom 2024-05-26 11-27-03

According to the [wiki](https://wynncraft.wiki.gg/wiki/Guilds#Leveling_Guild) there is a total of 7 ally slots to unlock, so this is an exhaustive list.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
cmkohnen: There is definitely a way to skip this container now that I think of it.
If a guild has no allies (which can be parsed form the previous container) this scan shouldnt be necessary. However due to Wynncraft itself not showing sent tributes in the previous menu (in contrast to recieved ones) we cant rule out tributes affecting global usage.

cmkohnen: If it werent for https://forums.wynncraft.com/threads/guilds.281813/page-3#post-3658780
we could also skip the parsing of that container if the tooltip says "No tributes sent or received".

kristofbolyai: Nothing you can do about that. But I can offer you something to work on, that can "compensate" for this additional tracking. Everything that is parsed from `parseCratesBombsCosmeticsContainer` can't really change during gameplay, except the silverbull rank. So that parsing only needs to happen once per launch. The problem is detecting when a player purchased silverbull. Neither you or me has the regexes for the chat messages... So I can't request you to do that change here, however, if in some way, you can do it, that would be the best.

kristofbolyai: Otherwise, I'll think about creating configs, or something else to compensate for the speed..

cmkohnen: > Nothing you can do about that. But I can offer you something to work on, that can "compensate" for this additional tracking. Everything that is parsed from `parseCratesBombsCosmeticsContainer` can't really change during gameplay, except the silverbull rank. So that parsing only needs to happen once per launch. The problem is detecting when a player purchased silverbull. Neither you or me has the regexes for the chat messages... So I can't request you to do that change here, however, if in some way, you can do it, that would be the best.

The problem would still be that while you are disconnected or in class selection/hub you wouldn't be able to parse chat or other changes.

How about having a threshold of time that skips guild and silverbull parsing if the time delta between logging out and back in is short enough (i. e. short enough for skipping it on entering wars and /switch as well as party world switching).

kristofbolyai: > The problem would still be that while you are disconnected or in class selection/hub you wouldn't be able to parse chat or other changes.

You don't get these messages when disconnected, they happen on join. Store messages work like this, never seen silverbull though.

cmkohnen: Ooooh, I in fact might have a regex for that then

cmkohnen: Okay: good and bad news: There is a chat message, but I do not know if it behaves similarily to the purchase message, as it is in different nature to regular shop messages:

```
[21:43:21] [Render thread/INFO]: [CHAT] À
[21:43:21] [Render thread/INFO]: [CHAT]          &3Welcome to the &b✮ Silverbull Trading Company&3!
[21:43:21] [Render thread/INFO]: [CHAT]            &7Your subscription has been activated.
[21:43:21] [Render thread/INFO]: [CHAT] À
[21:43:21] [Render thread/INFO]: [CHAT]                         &a✔ &fLower market taxes
[21:43:21] [Render thread/INFO]: [CHAT]                         &a✔ &fGuild boost ability
[21:43:21] [Render thread/INFO]: [CHAT]                         &a✔ &fDouble daily rewards
[21:43:21] [Render thread/INFO]: [CHAT]                         &a✔ &fFree monthly shares
[21:43:21] [Render thread/INFO]: [CHAT]                       &f&oAnd more unlocked!
[21:43:21] [Render thread/INFO]: [CHAT] À
```

kristofbolyai: I think it's fine to assume, it is.

cmkohnen: I have added the requested changes, however I'm not able to test them myself

kristofbolyai: Can you, @ShadowCat117?

ShadowCat117: I don't think I can trigger the welcome message again , it'll just be the renewal one which I'm pretty sure doesn't look like that

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -81,7 +83,7 @@ public final class CharacterModel extends Model {
     public final Storage<Long> silverbullExpiresAt = new Storage<>(0L);
 
     @Persisted
-    public final Storage<Boolean> silverbullSubscriber = new Storage<>(false);
+    public final Storage<Boolean> silverbullSubscriber = new Storage<>(null);
```
kristofbolyai: Could you use `ConfirmedBoolean` here?

common/src/main/java/com/wynntils/models/character/CharacterModel.java
```diff
@@ -155,21 +157,30 @@ public void onWorldStateChanged(WorldStateEvent e) {
...
+            lastDeathLocation = Location.containing(lastPositionBeforeTeleport);
+            CharacterDeathEvent deathEvent = new CharacterDeathEvent(lastDeathLocation);
+            WynntilsMod.postEvent(deathEvent);
+
+            if (deathEvent.isCanceled()) {
+                e.setCanceled(true);
+            }
+            return;
+        }
+        ;
```
kristofbolyai: typo

common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -87,12 +112,25 @@ public class GuildModel extends Model {
...
     private static final int MEMBERS_SLOT = 0;
     private static final int OBJECTIVES_SLOT = 13;
+    private static final int DIPLOMACY_MENU_SLOT = 26;
+    private static final List<Integer> DIPLOMACY_SLOTS = List.of(2, 3, 4, 5, 6, 7, 8);
 
     private static final List<Integer> OBJECTIVE_GOALS = List.of(5, 15, 30);
 
     private Map<String, GuildProfile> guildProfileMap = new HashMap<>();
+    // sent, received
+    private final Map<String, Pair<Map<GuildResource, Integer>, Map<GuildResource, Integer>>> guildDiplomacyMap =
```
kristofbolyai: It may be better to create a holder class/record for this, and "name" the objects in the pairs. That way you can do `guildDiplomacyMap.getGuild("ANO").getSent()`, which is much more verbose/readable than it is currently.

kristofbolyai (commented):
Those should be the last changes needed. We can't test the silverbull regex. Worst case, it doesn't work, and it becomes unsynced, for a single instance. We can fix that, if it becomes a problem.

common/src/main/java/com/wynntils/models/character/CharacterModel.java
magicus: I have never had a Silverbull subscription. Let me know in Discord if you need my help.

kristofbolyai: > I have never had a Silverbull subscription. Let me know in Discord if you need my help.

Me neither. If we really want to test this, we can get Scyu to gift either of us a month of silverbull subscription, from the store.

magicus: Probably better you get it, I guess you could actually put it to some good use. :)

kristofbolyai (approved):
Looks good! Thank you for waiting. I'll be merging this soon, after I got the chance to the test Silverbull tracking.

cmkohnen:
Thanks for actually testing!

kristofbolyai:
> Thanks for actually testing!

No problem, sorry it took so long, it took a while to coordinate. I'll do a final test in a bit. 
image

PR #2529, opened 2024-05-25
fix: Fix tomes being counted as defective, if they had no identifications [skip ci]

ShadowCat117:


kristofbolyai (approved):
Can you check the charm implementation..? I did this for gears already, but I think that one is still missing.

ShadowCat117:
Charms all have ids so they shouldn't need it

kristofbolyai:
> Charms all have ids so they shouldn't need it

The code should get fixed anyways.

ShadowCat117:
https://github.com/Wynntils/Artemis/pull/2530

PR #2530, opened 2024-05-25
fix: Fix charms being counted as defective, if they had no identifications [skip ci]

ShadowCat117:
All charms currently have identifications but good to be future proof


kristofbolyai:
Thank you!

PR #2532, opened 2024-05-27
refactor: Make label levels Optional<Integer> in preparation for MapData

magicus:
Contributed by @kristofbolyai.

This is extracted from #2494. This is the part that affects the currently used label system. It turned out that this was, in the end, mostly irrelevant to the changes needed in #2494.


kristofbolyai (approved):

kristofbolyai:
Looks good :) (Yes, it's 90% of my code, but let's merge if we both agree :))

PR #2537, opened 2024-05-29
port: Migrate Custom Pois and Loot Chests to map data

kristofbolyai:
**This PR targets `mapdata-ports`. All mapdata related changes should be merged to that branch, to prevent regressions.** 

Migration was tested by a personal collection of POIs, as well as data from Poi Providers, around 600 Pois.


kristofbolyai:
@magicus This PR brings up a few questions. 

Should we keep custom pois and loot chests in storage/config (where exactly is unrelated to this question, although I think loot chests should go to storage now, while pois can stay in hidden config, or can be moved as well) or, should we put them in separate files? That would mean extending JsonProvider to save to files. It would also mean we can't "upfix" them.

If we keep them in persisted (storage/config), should we serialize them as JsonLocation, or should we keep a minimal, non mapdata format, like they are now?

magicus:
My thought was that they should be stored as separate, individual files. And yes, I think definitely they should be stored in the new json format.

I have been toying with the idea of extending the persistance system to allow for arbitrary files, not only for fields in a class. Like you set up a `Map<File, DataStructure>` that saves each data structure in the corresponding file.

magicus:
But to be honest, I have not really thought this part of the story through.

kristofbolyai:
We'll sleep on this for a few night then, and see what we come up with.

kristofbolyai:
Coming back to this today. This is an important step for porting mapdata. Do you think the compromise of storing custom/loot chest map locations as the JsonLocation mapdata format **in storage** is acceptable? That way, we can semi-easily do this now, we only need a hacky way to move it from storage to config. Aaand, sometime later, we will separate them into other files. I most certainly do not want to sacrifice us not being able to upfix the custom files, so we need proper infrastructure to do it, which is a bit of work.

magicus:
Yes, that sounds like a reasonable intermediate step.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
```diff
@@ -94,7 +88,7 @@ public class MainMapFeature extends Feature {
     public final Config<Boolean> autoWaypointChests = new Config<>(true);
 
     @Persisted
-    public final Config<LootChestType> minTierForAutoWaypoint = new Config<>(LootChestType.TIER_3);
+    public final Config<LootChestTier> minTierForAutoWaypoint = new Config<>(LootChestTier.TIER_3);
```
magicus: I think this should go away. Once we started separating collected chest locations from user added waypoints, we should collect all of them. Then we can chose which to *display* in the map, but that is a different story.

magicus: That way you can e.g. show tier-1 chests in the minimap, but only 3-4 on the main map. And if we start collecting tier1 right away, we have a bunch of chests to display if/when the user changes her mind and wants to display chests of a lower tier. So the collection should not be tied to what we display. (Which it is right now)

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
```diff
@@ -162,32 +156,25 @@ public void onScreenOpened(ScreenOpenedEvent.Post event) {
...
         }
 
-        PoiLocation location = new PoiLocation(lastChestPos.getX(), lastChestPos.getY(), lastChestPos.getZ());
-        CustomPoi newPoi = new CustomPoi(
-                location,
-                chestType.getWaypointName(),
-                CommonColors.WHITE,
-                chestType.getWaypointTexture(),
-                CustomPoi.Visibility.DEFAULT);
+        Location location = new Location(lastChestPos.getX(), lastChestPos.getY(), lastChestPos.getZ());
```
magicus: I'm not sure what type lastChestPos is, but can't it implement Location? Or have a getLocation()?

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
```diff
@@ -162,32 +156,25 @@ public void onScreenOpened(ScreenOpenedEvent.Post event) {
...
 
-        if (customPois.get().stream().noneMatch(customPoi -> customPoi.equals(newPoi))) {
-            customPois.get().add(newPoi);
+        if (foundChestLocations.get().stream()
```
magicus: It's kind of weird that we have this in MainMapFeature. It should be in a Model.

common/src/main/java/com/wynntils/models/containers/type/LootChestTier.java
```diff
@@ -8,36 +8,40 @@
...
         return waypointTexture;
     }
 
+    public int getWaypointTier() {
+        return waypointTier;
+    }
+
     public String getWaypointName() {
-        return waypointName;
+        return "Loot Chest " + waypointTier;
```
magicus, line 36: i18n? Or at least a i18n FIXME?

magicus:
I reviewed the existing code. It looks good, I think, but I think you can continue a bit deeper with the cleaning up that becomes possible with the new design.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
kristofbolyai: Agreed, I can remove this.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
kristofbolyai: It's a Minecraft BlockPos. Can't add any implementations (without mixins, but we don't want to do that).

common/src/main/java/com/wynntils/models/containers/type/LootChestTier.java
kristofbolyai: Not really something to i18n IMO, it's more like internal knowledge of the wynn world, than something we represent to the user (this method is not even used atm).

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
kristofbolyai: Should we move both the found chests and custom markers to a (new?) model?

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
magicus: Maybe make a new constructor to Location that takes in a BlockPos then? (Maybe there already is one..?)

kristofbolyai: Okay, I can do that.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
magicus: Yes and no. :-) They should not live inside a Feature.

However, actual loot chests align with the concept of a Model -- it is something that exists in the Wynncraft world. The user-added waypoints however, are a complete add-on provided by Wynntils, so it is definitely a Service, like favorite items etc.

kristofbolyai:
This PR is now waiting on https://github.com/Wynntils/Artemis/pull/2546, as it's too hard to keep track of both of them.

common/src/main/java/com/wynntils/models/containers/type/LootChestTier.java
magicus: Ok.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
kristofbolyai: Okay. When migrating from the old customPois config, I'll make sure to migrate to two different components' (LootChestModel and UserWaypointsService?) storage instead.

magicus: Yep!

Did we decide how to handle the old chests? The problem is that we cannot be 100% sure that the user did not manually enter a custom poi with the same name and icon as an automatically collected chest location. (Perhaps they marked the entrance of a cave that way.)

I think we said something about migrating them to a "legacy" storage, and that when we find a new chest, we add it to a "proper" storage, and also check if it is in the "legacy" storage, and if so, remove it.

magicus: Then maybe we need to offer a way to handle remaining legacy storage chests after users have been playing for awhile, by either removing them, or moving them to user-added waypoints.

kristofbolyai: I was thinking of moving all loot chests that are of the format we used to the found chests automatically. I think that might be the simplest, but the most comprehensive solution. We can check for the visibility and color too, to rule out custom added ones.

kristofbolyai:
This PR is now ready to be reviewed/merged.

common/src/main/java/com/wynntils/models/containers/LootChestModel.java
```diff
@@ -192,4 +259,68 @@ private void resetNormalDryStatistics() {
...
+        customPois.get().removeAll(migratedPois);
+        customPois.touched();
+    }
+
+    // This feature ports old custom poi data to the new mapdata system
+    // This is a one-time migration, but can't be removed in the foreseeable future,
+    // so we can keep upfixing old configs
+    private boolean migrateToMapdata(CustomPoi customPoi) {
+        boolean isLootChest = isCustomPoiLootChest(customPoi);
+        if (!isLootChest) return false;
```
magicus, line 304: I'm a bit tired so I might be missing something obvious, but I'll ask anyway. :) You are only migrating the found loot chests? Should not the rest of the custom pois migrate to waypoints?

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
```diff
@@ -568,7 +568,7 @@ public void deletePoi(CustomPoi poiToDelete) {
 
         customPois.get().remove(poiToDelete);
         customPois.touched();
-        Managers.Feature.getFeatureInstance(MainMapFeature.class).updateWaypoints();
+        // Managers.Feature.getFeatureInstance(MainMapFeature.class).updateWaypoints();
```
magicus, line 571: Why is this commented out? If it is not needed anymore, remove it?

common/src/main/java/com/wynntils/services/mapdata/UserWaypointService.java
```diff
@@ -0,0 +1,118 @@
...
+import com.wynntils.services.mapdata.providers.builtin.WaypointsProvider;
+import com.wynntils.services.mapdata.providers.json.JsonMapAttributes;
+import com.wynntils.services.mapdata.providers.json.JsonMapAttributesBuilder;
+import com.wynntils.services.mapdata.providers.json.JsonMapVisibility;
+import com.wynntils.utils.mc.type.Location;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class UserWaypointService extends Service {
```
magicus: I would like to argue that we should call this just `Waypoints`, and use that term consistently in the UI for the locations added by the user. The system with the compass marker etc that we called waypoint in the old code I would argue should be called "markers". And that we try to keep "User" out of the name of mapdata types.

common/src/main/java/com/wynntils/services/mapdata/UserWaypointService.java
```diff
@@ -0,0 +1,118 @@
...
+        customPois.get().removeAll(migratedPois);
+        customPois.touched();
+    }
+
+    // This feature ports old custom poi data to the new mapdata system
+    // This is a one-time migration, but can't be removed in the foreseeable future,
+    // so we can keep upfixing old configs
+    private boolean migrateToMapdata(CustomPoi customPoi) {
+        boolean isLootChest = LootChestModel.isCustomPoiLootChest(customPoi);
+        if (isLootChest) return false;
```
magicus: Okay, I see, the waypoint migration happens here. Ignore my comment above.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/LootChestsProvider.java
```diff
@@ -0,0 +1,47 @@
...
+        PROVIDED_FEATURES.forEach(this::notifyCallbacks);
+        PROVIDED_FEATURES.clear();
+        foundChests.forEach(LootChestsProvider::registerFeature);
+    }
+
+    public static void registerFeature(FoundChestLocation foundChest) {
+        PROVIDED_FEATURES.add(foundChest);
+    }
+
+    public static final class FoundChestLocation extends JsonMapLocation {
```
magicus, line 38: It seems a bit odd to have this extend "Json" MapLocation. I had to check the code to understand why. I see now that there are several "Json" implementations that are basic implementations of interfaces which just maps getters to fields, and sets those using a constructor. That kind of implementation is good, but maybe we should not call it "Json". 

I'm not sure what to rename it to, though. Something like "BasicMapLocation" or "SimpleMapLocation", maybe? "ConcreteMapLocation"?

common/src/main/java/com/wynntils/services/mapdata/UserWaypointService.java
kristofbolyai: Okay, I can live with that.

common/src/main/java/com/wynntils/screens/maps/PoiManagementScreen.java
kristofbolyai: The PoiManagmentScreen needs full porting, I've not done anything in that regard in this PR. I did not want to remove these old update methods, for anyone porting in the future, it might be useful. This'll be removed before the actual mapdata PR is merged.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/LootChestsProvider.java
kristofbolyai: How would the new class look like? I thought it's obvious that we have to use a json map location here, so these can be serialized/deserialized by GSON.

magicus: Ok, I see how you mean. I didn't really think about it like that, but sure, I can agree with that.

magicus (approved):

PR #2538, opened 2024-05-29
chore: Revert "feat: Start using MapData for some providers [skip ci]" [skip ci]

kristofbolyai:
Reverts Wynntils/Artemis#2536

I've noticed that mapdata rendering performance is a significant problem. The recursive nature of resolving attributes creates a big impact in CPU cycles and memory allocations. There are also some minor visual issues.

I'll go ahead and revert this PR to give us time to fix these issues, without holding back main, or otherwise we wouldn't be able to do any releases, until this problem is fixed.

I'll share profiler data in Discord about the issue, if you have time to take a look, magicus.

### DIFF
--- a/common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
+++ b/common/src/main/java/com/wynntils/features/map/WorldWaypointDistanceFeature.java
@@ -220,8 +220,7 @@ public class WorldWaypointDistanceFeature extends Feature {
                                 false,
                                 scale.get(),
                                 1,
-                                50,
-                                true);
+                                50);
                 bufferSource.endBatch();
                 poseStack.popPose();
             }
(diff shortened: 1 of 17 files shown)


magicus:
That is fine. We'll have to check the performance. I did not look into that at all. 

The visual issues is a bit more disturbing. I was running with the old and the new markers alongside and tried to check that they looked consistent.

kristofbolyai:
> That is fine. We'll have to check the performance. I did not look into that at all.
> 
> The visual issues is a bit more disturbing. I was running with the old and the new markers alongside and tried to check that they looked consistent.

Sadly, after spending time with checking this, I can say the opposite. Performance bothers me much more than the visual issue, I think the visual issue is a matter of a check missing, while the performance is an inherited design issue.

magicus:
I did have the thought, though, that resolving the ResolvedMapAttributes a single time and store it as actual values seemed like a better idea than resolving it dynamically. This probably means that we should break the MapAttributes implementation as well, so we can have it return actual values instead of optionals, and maybe also make it a record? We could cache a resolve MA in each feature, and just invalidate them if we load a new provider.

magicus:
> Performance bothers me much more than the visual issue,

I see what you mean, but I feel pretty confident we can get the performance back on par with quite easy methods, like the ones I sketched out above. It needs to be done, sure, but I don't think it is that hard.

kristofbolyai:
> > Performance bothers me much more than the visual issue,
> 
> I see what you mean, but I feel pretty confident we can get the performance back on par with quite easy methods, like the ones I sketched out above. It needs to be done, sure, but I don't think it is that hard.

Let me know if I can help out. I've shared everything related on Discord.

PR #2539, opened 2024-05-29
feat: Start using mapdata for some providers - remerge

kristofbolyai:
This PR is on hold until performance issues with mapdata are resolved.


magicus:
I have now remade ResolvedMapAttributes so it is a record, with values computed once, at resolve time. This resolved attributes record is furthermore stored in a cache, keyed by feature.

I have not implemented cache invalidation. This should basically occur if we add a new provider, since they can add new attributes to categories.

magicus:
We are still creating new MapFeaturePoiWrappers in getFeaturesAsPois(). This is done once per render, so it is probably causing a bit of object allocation churn. Otoh, the objects are now cheap to create. To get rid of those, we need to change how rendering is called from the map screen/overlay, so that it could work directly on a MapFeature. 

(Basically, I think the render method of MapFeaturePoiWrappers should be broken out, and perhaps added to MapDataService or a separate MapFeatureRenderer.)

magicus:
@kristofbolyai Can you please check if the performance is still bad, or if this solved the problem?

magicus:
And as a bonus, I think the code got even cleaner and better...

kristofbolyai:
Yes. I'll go ahead and check performance tomorrow.

kristofbolyai:
> We are still creating new MapFeaturePoiWrappers in getFeaturesAsPois(). This is done once per render, so it is probably causing a bit of object allocation churn. Otoh, the objects are now cheap to create. To get rid of those, we need to change how rendering is called from the map screen/overlay, so that it could work directly on a MapFeature. 
> 
> 
> 
> (Basically, I think the render method of MapFeaturePoiWrappers should be broken out, and perhaps added to MapDataService or a separate MapFeatureRenderer.)

If it's not a major point in the profiling, then it'll be done later (when we ported all Pois, and we start to phase them out fully). If it becomes a problem earlier, we will see how we can create a bridge to make it work.

kristofbolyai:
> I have not implemented cache invalidation. This should basically occur if we add a new provider, since they can add new attributes to categories.

Do you want me to do this? It'll be necessary when we add user providers / configs.

magicus:
I fixed it now. It was not that it was hard, it was that I was too tired yesterday to do it. :)

There is one caveat that we need to look out for: if we add the functionality in the UI for users to override attributes for certain categories (basically giving them a personal category provider), which I think we should, then if that is *modified* we need to handle that. Either we need to remove/re-add it, or add functionality to tell the system that a provider has changed. 

Currently the system assumes that a provider stays constant once added.

But I think we'll figure out the best way to handle that once we get there. Also, it will be obviously clear that modifying a provider will not result in any changes so there is no risk of inadvertent bugs. :-)

kristofbolyai:
I'll keep this in mind. Thank you.

magicus:
Let me know when you have verified the performance. If there are no hiccups, I think this is (finally!) good to merge properly.

kristofbolyai:
I've done a new round of profiling. The changes seem good, but more is required. Specifically, `com.wynntils.services.mapdata.MapDataService#getIcon` needs to be optimized or cached.

I've also created a patch to fix a bootstrap issue with the invalidation patch, otherwise the mod can't launch. You'll see it when you have time to run it. I'll send my patch in Discord, but I am sure you'll also do essentially the same change as I did.

kristofbolyai (commented):
This looks good now. I can't approve my own PR, but consider this as approved. Performance looks on par, or better than the old methods.

magicus (approved):

PR #2543, opened 2024-05-30
feat!: Introduce Mapdata service to users

### LINKED ISSUE #2582: [MapData] - Different attributes for different MapFeatures
I am running into issues where MapAttribute is too close to MapLocation, and other implementations, like MapArea need a custom set of different attributes. I think we should consider different attributes for each MapFeature type. I also know it will be useful when implementing MapPath.

kristofbolyai:
This branch is tracking all the porting work that needs to be done for mapdata. A single commit for moving to mapdata was already merged to main, I don't want to revert that. Everything else should be targetting this branch, to ensure we do not have significant regressions.

edited (by @magicus): 

Here is a checklist of all remaining work:

- [x] resolve #2582 -- i.e. design a more proper way to inherit attributes for non-location map features as well
- [x] "Block" areas and path from having marker option attributes
- [x] update GUIs to use mapdata
- [x] add support for area types
- [x] The MarkerProvider framework must be folded into MapData
  - [x] Fix Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation() so it does not require a Texture, but an iconId, and remove AbstractTexture
- [x] Fix the map screens
- [x] Figure out if we should change attribute resolving to first go through providers in order, and go up the hierarchy per provider instead of going through the category hierarchy, and checking each hierarchy level in provider order (see https://github.com/Wynntils/Wynntils/pull/2570#issuecomment-2156226956)
   - Note: Override providers can be used to change built in provider defaults in a way that they will always take priority. This should resolve this issue.
- [x] Consider renaming "Json" classes that are not json but built-in
- [ ] ~~Draw Lootrun paths as mapfeature paths~~ (EDIT: As of right now, this is not planned for the initial release of MapData)
  - [ ] ~~add support for path types~~
- [x] Add back the pois that are still not converted to MapData:
  - [x] Provided custom pois
  - [x] Marker waypoints
  - [x] Remote players
  - [x] Territory pois
- [ ] Provider Manager GUI
- [ ] Category Manager + assume custom waypoints are primarily category driven, only using explicit attributes as override


magicus:
I apologize if I merged something you did not intend to get in. Miscommunication apparently. :(

kristofbolyai:
> I apologize if I merged something you did not intend to get in. Miscommunication apparently. :(

I did intend to get it in :). Until I realized that MainMapScreen is too dependent on poi class instanceof checks. We have some minor regressions, like not being able to remove or edit custom pois. We can live with that, I guess. (I would not have known this if it was not pointed out by testing, so there is nothing to be done.)

kristofbolyai:
Overall, I would much rather do the rolling-releases like you do, and just merge every PR into main. But that comes at a great cost, testing wise. We either spend too much time testing and doing temp fixes, or we don't test changes thoroughly and they eventually become problems.

magicus:
> Until I realized that MainMapScreen is too dependent on poi class instanceof checks

Ah, I see. Some additional proof as to why instanceof checks are bad. :-/

magicus:
It's been so long so I forgot. What remains to be fixed to be able to push this?

magicus:
I've brought this up to date with main. It was definitely a non-trivial undertaking. :-/ But now it is done..

kristofbolyai:
I accidentally pushed a PR to here directly! I've forced push to revert it.

magicus:
I apologize for being absent from this work for so long. I just tested the mapdata-ports at its current state, and it looks really good! However, there are a few things I noticed that are troubling me. I don't know if any of these are blockers for an integration into main, but I think they might/should be. At least 1 seems just like it should be fixed, but I think that unless we do 2 and 3, much of the benefit of MapData will be lost on users, and unless we do 2 we risk having users that use a lot of overridden special personal waypoints, that will then later on be hard to retrofit onto a proper category system.

1) The waypoint management screen:
    * What seems like a bug or missing implementation: the waypoint management screen says I have no waypoints, but I did create a couple of waypoints. (I have no old waypoints in my dev environment, nor did I try to import anything). 
    * The waypoint management screen has no button for actually adding a waypoint. That seems like it should be a simple and useful addition.
    * The texts still say "POIs" instead of waypoints.

2) Improved category support for waypoints. 
I see **categories** as the lifeblood of mapdata. While it is technically possible to edit a personal waypoints icon, the label visibility, etc, I think this should be done only as a special case or a "last resort". The typical workflow should be that the user sets up a bunch of categories (maybe we can even provide some default ones?), like `good-grinding-spots` or `strange-things-to-examine` or `best-entrance-to-cave` or whatever. They can assign icons, colors and visibility (and perhaps even a default label) to these categories. When creating a new waypoint, they should just select which category it should belong to. And then perhaps there can be an "Advanced" or "Override" button, that when pressed, allows you to customize this particular waypoint. But the GUI should guide (or steer) the user towards utlizing categories as the first-hand choice. 

Also, the category system is hierarchical in nature. That means it should be possible, indeed easy, for the user to create and select sub-categories, as well as top-level categories, such as `good-grinding-spots:level-30`. Sub-categories inherit properties of their parents, so for each attribute available, it should be clear that the default mode is "inherit", and to change it you need to be explicit to say that the mode should be "override" and then you can provide a value. It should also be easy to go back to "inherit". Maybe this can be done GUI-wise not by an "inherit/override" switch, but by having like dimmed showing of the values and a greyed-out "Reset" button for inherited values, and opaque values and an active "Reset" button for overridden values. The point is that this inheritance needs to be clear for the user.

Finally, to really appreciate the power that categories gives us, we would need a way to select which categories are displayed on the map. We had something similar in the legacy Wynntils, but Artemis has never had anything like that. I'm imagining some kind of GUI where the categories are displayed like a collapsible tree, where you can select either a parent or leave category, and enable if it should be shown on the map or not. Or perhaps you would like to highlight a category, so you select "Discovered Tier 3 chests" and "Discovered Tier 4 chests" and click on "Always show", which will give the category "wynntils:chests:tier4" (or whatever we called it) the override property of "label visibility: always".

Once we have this method in place of easily hiding or showing mapdata entries on the map, we can actually start adding much more detailed mapdata suppliers, like NPCs or quest starts. And it makes it more feasible to import someone else's shared mapdata json files, since if that file is properly designed, you can select to only show certain aspects of it.

3) Handling of mapdata providers
I think we are still missing a mapdata provider management screen. A place where we can add new local or online providers, where the latter should be either downloaded just once, or "subscribed" to, so we check regularly for updates, like the event-based daily easter egg hunts. (We might need to add some more basic support in the mapdata service provider itself for the latter.) 
The provider management screen should also allow you to re-order providers, and enable/disable individual providers.

ShadowCat117:
> 1. The waypoint management screen:

It's basically fully working on https://github.com/Wynntils/Wynntils/pull/3123 it just needs to display found chest locations too, I'll hopefully finish that on Friday

magicus:
Fortunately, I think doing 2 is actually not that hard; most of the pieces are already in place, they just needs to be connected differently. 

I one big thing we need to add is a CategoryManagementScreen, which would allow us to add/remove or customize a category. But the customization is basically already the same screen as in the current WaypointCreationScreen, only that it applies to the category and not an individual waypoint. (But with the inherit/override distinction I talked about above.)
I think it would be helpful if you could also select a category, and be shown how the resolved attributes will turn out for that category. But this might be bells and whistles that we can add later on, and we can certainly improve this iteratively, as long as we mark the general path for the users on how to work with categories. (I think what I'm saying is: category handling need not be perfect in the initial release of MapData, but we need to clearly steer the user towards using it, so they understand that this is how it is supposed to work.)

Also, we should not forget that categories can (and is encouraged to) have a name, not just the formal path, so `good-grinding-spots:level-30` could have its name set to e.g. "(lvl 30) Grinding spots" by the user. The GUI should highlight this functionality, as in it should assume that new categories should have a name, and display that when possible, and only fall back to showing the path if a name is unavailable.

magicus:
I was also thinking, did we ever figure out the correct (or at least best) way of dealing with different override priority based on category inheritance and provider ordering?

Basically, if we have this scenario:
* Provider A provides category "foo" with attribute "labelColor: red"
* Provider B provides category "foo:subfoo" with attribute "labelColor: green"

If I want to render a label for category "foo:subfoo", should it be always green, regardless of the ordering of providers, or should provider A be able to have a higher priority so it will match already on "foo" and never fall back on provider B?

When I state it like this, it seems obvious that we should select the most specific category match regardless of provider ordering, but I vaguely remember having a real-world problem where the opposite seemed to be the best solution.

In any case, ordering of providers will be important for scenarios like this:
* Provider A provides category "foo" with attribute "labelColor: red"
* Provider B provides category "foo" with attribute "labelColor: green"

Basically, it will allow the user to add a new provider that overrides built-in provider categories. In effect, a re-implementation of the legacy feature of custom map icon packs.

edit: Maybe the solution is that we should resolve each level of the category path completely, where providers are tested in priority order, and higher priority providers override the values of lower order providers. Then we move on to the child category, and do the same. That way we could have like:
* The built-in provider says "wynntils:services" has "labelColor: blue", and "wynntils:services:identifier" has "icon: wynntils:icons:identifier"
* The user adds a local provider with higher priority that says "wynntils:services" has "labelColor: purple". 

This way, all services will get a different label color, but the identifier will still have the default icon.

kristofbolyai:
> I was also thinking, did we ever figure out the correct (or at least best) way of dealing with different override priority based on category inheritance and provider ordering?
> 
> Basically, if we have this scenario: Provider A provides category "foo" with attribute "labelColor: red" Provider B provides category "foo:subfoo" with attribute "labelColor: green"
> 
> If I want to render a label for category "foo:subfoo", should it be always green, regardless of the ordering of providers, or should provider A be able to have a higher priority so it will match already on "foo" and never fall back on provider B?
> 
> When I state it like this, it seems obvious that we should select the most specific category match regardless of provider ordering, but I vaguely remember having a real-world problem where the opposite seemed to be the best solution.
> 
> In any case, ordering of providers will be important for scenarios like this: Provider A provides category "foo" with attribute "labelColor: red" Provider B provides category "foo" with attribute "labelColor: green"
> 
> Basically, it will allow the user to add a new provider that overrides built-in provider categories. In effect, a re-implementation of the legacy feature of custom map icon packs.

We go down until the most specific category. I've been using the "override provider" concept for a while now, which can provide attribute overrides that apply on top of a features and it's categories' attributes. All of the "overrides" that are applicable to a feature is merged together, and when overrides conflict, the first registered provider takes priority.

See the PR's description:
> Figure out if we should change attribute resolving to first go through providers in order, and go up the hierarchy per provider instead of going through the category hierarchy, and checking each hierarchy level in provider order (see https://github.com/Wynntils/Wynntils/pull/2570#issuecomment-2156226956)
Note: Override providers can be used to change built in provider defaults in a way that they will always take priority. This should resolve this issue.

kristofbolyai:
> 3\. Handling of mapdata providers
>     I think we are still missing a mapdata provider management screen. A place where we can add new local or online providers, where the latter should be either downloaded just once, or "subscribed" to, so we check regularly for updates, like the event-based daily easter egg hunts. (We might need to add some more basic support in the mapdata service provider itself for the latter.)
>     The provider management screen should also allow you to re-order providers, and enable/disable individual providers.

Incoming. https://github.com/Wynntils/Wynntils/pull/3130 or later.

magicus:
It seems you got most of it covered. Great work! I'm eternally grateful for your work @ShadowCat117 and @kristofbolyai in actually managing to get this to a deliverable state. 

Sorry if I've been beating on open doors. I just have not kept up at all with MapData and now decided to go through the backlog.

magicus:
> We go down until the most specific category. I've been using the "override provider" concept for a while now, which can provide attribute overrides that apply on top of a features and it's categories' attributes. All of the "overrides" that are applicable to a feature is merged together, and when overrides conflict, the first registered provider takes priority.
> 
> See the PR's description:
> 
> > Figure out if we should change attribute resolving to first go through providers in order, and go up the hierarchy per provider instead of going through the category hierarchy, and checking each hierarchy level in provider order (see [#2570 (comment)](https://github.com/Wynntils/Wynntils/pull/2570#issuecomment-2156226956))
> > Note: Override providers can be used to change built in provider defaults in a way that they will always take priority. This should resolve this issue.

I think I need to think a bit more on this. I don't really understand how "override providers" fit in; I think if we did this right there would need to be no such thing. My gut feeling is that the correct way of resolving the attributes for a map feature with attributes "attr1: value1" and category "foo:bar:baz" is to first go to category "foo" and resolve all it's attributes, from all providers in the correct provider ordering, so that if several providers provide "attr2" with e.g. "value2a" and "value2b", then the one with the highest priority will be selected. And then we descend into "foo:bar", and do the same. If an attribute is not specified for "foo:bar", it will be inherited from "foo" as we resolved it, otherwise it will be overridden, in the same way that provider priority decides. And finally we go to "foo:bar:baz", doing the same all over again.

And then finally finally, we take the individual map feature and applies the "attr1: value1" it provides, overriding all possible values set by the categories.

I'm imagining the providers as a list where a provider further down the list is able to override providers higher up, basically meaning tha the last provider has the highest priority. But it can also be envisioned the other way around, that the highest provider is checked first, and when it can give an answer for a specific attribute, it wins and we check no further. There are some conceptual and implementational differences between those to ways of thinking, but the end result is the same. We just need to be clear with the user what is happening. 

The reason I think the "later overrides" model is better is that we can fix our built-in values as the very first element, and then users can override by adding stuff below. And if we want to override things, we can basically have two built-in providers, the "Always" provider, and an overriding (hence lower in the list) "Currently overridden" provider, that serves the role of an "overriding provider". 

If we do it the other way around, we would have to fix our built-in providers at the very bottom of the list, which just feels more odd from a UI perspective.

magicus:
I'm not sure if is 100% necessary that we get this right before releasing. We can probably change what in effect is some obscure ordering principles after release and it will not really affect the majority of map features. But I do think it would be great if we really could nail this before release 1.0 of MapData, since it is still kind of essential for the conceptual model of what we're doing here.

magicus:
> > 3. Handling of mapdata providers
> > I think we are still missing a mapdata provider management screen. [... ]
> > (We might need to add some more basic support in the mapdata service provider itself for the latter.)
> 
> Incoming. #3130 or later.

Am I interpreting this correctly that #3130 will lay the foundational groundwork for proper provider management, and that we will then need to add a GUI screen on top of this?

magicus:
Actually, this comment from the PR description sounds like we are actually doing right now what I propose we do:

> Figure out if we should change attribute resolving to first go through providers in order, and go up the hierarchy per provider instead of going through the category hierarchy, and checking each hierarchy level in provider order

🤣 

I read the linked comment, which basically suggested we should have the functionality so "That way you could add a provider last that would override even very general attributes."

But that also seems like it could be very much broken, if I have added a provided that says "labelColor: green" for category "wynntils", then every label for all we do will be green, and that does not make sense. If we do this, we lose the ability to have more specified attributes per sub-category. So this is clearly a bad idea.

Maybe this is where the override providers comes in? Allowing a select handful of providers that actually are allowed to do this kind of sweeping changes? Do I understand them correctly?

kristofbolyai:
@magicus can you please check the in code usages of override providers? I don't have the energy to explain it currently, but I can agree working towards removing it. 

Edit: I can now see that you are not explicitly against these "overrides".

magicus:
You don't have to explain, it's on me to dig in and understand the code.

common/src/main/java/com/wynntils/core/components/Models.java
```diff
@@ -69,6 +68,7 @@ public final class Models {
     public static final AbilityTreeModel AbilityTree = new AbilityTreeModel();
     public static final BankModel Bank = new BankModel();
     public static final BeaconModel Beacon = new BeaconModel();
+    public static final ActivityModel Activity = new ActivityModel(Beacon);
```
magicus: If ActivityModel constructor depends on Beacon, it should be moved under the (currently removed) heading "// Models with constructor dependencies, ordered alphabetically as far as possible" below.

common/src/main/java/com/wynntils/core/components/Services.java
```diff
@@ -47,5 +49,8 @@ public final class Services {
     public static final StopwatchService Stopwatch = new StopwatchService();
     public static final TranslationService Translation = new TranslationService();
     public static final UpdateService Update = new UpdateService();
+    public static final WaypointsService Waypoints = new WaypointsService();
     public static final WynntilsAccountService WynntilsAccount = new WynntilsAccountService();
+
+    public static final UserMarkerService UserMarker = new UserMarkerService(MapData);
```
magicus: Kind of sad that we need to introduce a Service with dependencies on another Service. :(

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -253,69 +256,69 @@ private void renderPois(
...
-            float dX = (poi.getLocation().getX() - (float) playerX) / zoomRenderScale;
-            float dZ = (poi.getLocation().getZ() - (float) playerZ) / zoomRenderScale;
-
-            if (followPlayerRotation.get()) {
-                float tempdX = dX * cosRotationRadians - dZ * sinRotationRadians;
-
-                dZ = dX * sinRotationRadians + dZ * cosRotationRadians;
-                dX = tempdX;
-            }
+        // FIXME: Add back the pois that are still not converted to MapData
```
magicus: This FIXME is addressed now, isn't it?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -296,52 +261,31 @@ public void doRender(GuiGraphics guiGraphics, int mouseX, int mouseY, float part
...
-    }
+        mapFeatures = mapFeatures.filter(feature -> !(feature instanceof SeaskipperDestinationArea));
 
-    private Stream<PlayerMainMapPoi> getPlayerPois(
-            boolean renderRemotePartyPlayers, boolean renderRemoteFriendPlayers) {
-        return Services.Hades.getHadesUsers()
-                .filter(hadesUser -> (hadesUser.getRelation() == PlayerRelation.PARTY && renderRemotePartyPlayers)
-                        || (hadesUser.getRelation() == PlayerRelation.FRIEND && renderRemoteFriendPlayers))
-                .map(PlayerMainMapPoi::new);
+        // FIXME: Add back the pois that are still not converted to MapData
```
magicus, line 311: Same here...

common/src/main/java/com/wynntils/services/map/PoiService.java
```diff
@@ -77,31 +47,11 @@ public PoiService() {
...
-
-    public Stream<ServicePoi> getServicePois() {
-        return servicePois.stream();
-    }
-
-    public Stream<CombatPoi> getCombatPois() {
-        return Stream.concat(combatPois.stream(), cavePois.stream());
-    }
-
     public List<CustomPoi> getProvidedCustomPois() {
```
magicus: Is this still used?

common/src/main/java/com/wynntils/services/mapdata/features/type/MapArea.java
```diff
@@ -0,0 +1,26 @@
...
+import com.wynntils.services.mapdata.attributes.type.MapAreaAttributes;
+import com.wynntils.utils.mc.type.Location;
+import com.wynntils.utils.type.BoundingPolygon;
+import com.wynntils.utils.type.BoundingShape;
+import java.util.List;
+
+public interface MapArea extends MapFeature<MapAreaAttributes> {
+    // The area is described by a polygon. This list is the sequence of
+    // vertices of that convex polygon, ordered in a counterclockwise orientation.
+    // The last segment of the polygon connects from the last vertice to the first.
```
magicus: typo: "vertex"

common/src/main/java/com/wynntils/services/mapdata/providers/type/MapDataOverrideProvider.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Get the attributes that are overriden by this provider.
+     *
+     * @param mapFeature The map feature to get the attributes for.
+     * @return The attributes that are overriden by this provider.
+     */
+    MapAttributes getOverrideAttributes(MapFeature mapFeature);
+
+    /**
+     * Get the feature ids that are overriden by this provider.
```
magicus: The wording is similar to getOverrideAttributes, but I think the meaning is rather different: This is the list of feature id:s that this provider does serve overridden attributes for.

I'm not sure how we should phrase this, but the getOverrideAttributes should not share description with getOverridden[Feature|Category]Ids.

magicus:
FWIW, I skimmed through all the changes in this entire PR. I think this mostly looks good and makes sense. (I'm glad we already had a baseline MapData implementation in main, so a lot of that code reviewing need not be done now.)

I added a handful of comments where I found something out of the ordinary.

magicus:
And also, I think I understand the current override functionality, and I'm fine with it.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
kristofbolyai: They'll be removed in their specific PRs. I think this list is maintained correctly, and there is a reason I've left it there. We'll come back to all FIXMEs in this PR before the final merge anyways.

common/src/main/java/com/wynntils/services/map/PoiService.java
kristofbolyai: It's removed in #3130.

common/src/main/java/com/wynntils/services/mapdata/providers/type/MapDataOverrideProvider.java
kristofbolyai: Okay, I've changed this.

common/src/main/java/com/wynntils/core/components/Services.java
kristofbolyai: It's not longer needed anymore.

common/src/main/java/com/wynntils/core/components/Models.java
kristofbolyai: Done.

magicus:
This is starting to look really, really good! I added two more point to the bullet list in the description. But I believe that is really the last part, then we can actually push this.

magicus:
Does `chore: Rough merge of main` mean that the previous merge was incomplete, and that this commit fixes it, or that this commit is just a partial fix and more work is needed?

(I don't know if I should interpret "rough" as in "it was rough weather", or as in "this is a rough sketch")

kristofbolyai:
It's documented on DC; tldr is it runs but the merge added fixmes.

magicus:
What is really remaining here? Is it the GUI part? (I've forgotten)

kristofbolyai:
> What is really remaining here? Is it the GUI part? (I've forgotten)

Mostly. We are planning on working on it after 3.1.0 (as the next milestone).

ShadowCat117:
Something else I think is a blocker for this is the issue with storage files failing to save correctly (and therefore resetting on next launch)

With current data it's annoying when it gets reset but manageable, with waypoints moving to storage with this it's more of a pain to lose those and with waypoints storing a lot of data, the issue is more likely to occur (as it seems to happen more often with larger storage files)

ShadowCat117:
I'm not sure if I broke something with updating the branch or if some logic we've changed in main is just causing the Poi migration to break but running this branch is now getting `Caused by: java.lang.IllegalArgumentException: Feature class com.wynntils.features.map.MainMapFeature is not registered`

ShadowCat117:
https://github.com/Wynntils/Wynntils/pull/4231

PR #2545, opened 2024-05-31
chore: Once again, use old POI system instead of MapData

magicus:
Until #2543 is properly done, keep using the old POI system in release.


kristofbolyai (approved):
Works great in-game. Can you submit the reverse patch for this to the mapdata-ports branch?

PR #2546, opened 2024-05-31
feat: Add a system for providers to signal when they change

magicus:
Unfortunately this goes a bit against the design philosophy that a provider would just provide a stream of features, and not really care about what it exports in that stream.

Due to performance reasons, we need to cache this, and that means we need to track any changes.

### DIFF
--- a/common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
+++ b/common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
@@ -32,7 +32,9 @@ public enum ServiceKind {
     TRADE_MARKET("Trade Market", Texture.TRADE_MARKET, "trade-market"),
     WEAPON_MERCHANT("Weapon Merchant", Texture.WEAPON_MERCHANT, "merchant:weapon"),
     WEAPONSMITHING_STATION("Weaponsmithing Station", Texture.WEAPONSMITHING_STATION, "profession:weaponsmithing"),
-    WOODWORKING_STATION("Woodworking Station", Texture.WOODWORKING_STATION, "profession:woodworking");
+    WOODWORKING_STATION("Woodworking Station", Texture.WOODWORKING_STATION, "profession:woodworking"),
+    MERCHANT("Merchant", null, "merchant"),
+    STATION("Station", null, "profession");
 
     private final String name;
     private final Texture texture;
--- a/common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
+++ b/common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
@@ -4,7 +4,6 @@
  */
 package com.wynntils.services.mapdata.providers.builtin;
 
-import com.wynntils.core.components.Services;
 import com.wynntils.services.map.pois.CustomPoi;
 import com.wynntils.services.mapdata.attributes.AbstractMapAttributes;
 import com.wynntils.services.mapdata.attributes.FixedMapVisibility;
@@ -33,12 +32,15 @@ public class WaypointsProvider extends BuiltInProvider {
         return PROVIDED_FEATURES.stream();
     }
 
-    public static void resetFeatures() {
+    public void updateWaypoints(List<CustomPoi> customPois) {
+        PROVIDED_FEATURES.forEach(feature -> notifyCallbacks(feature));
         PROVIDED_FEATURES.clear();
-        Services.MapData.invalidateAllCaches();
+        customPois.forEach(customPoi -> {
+            registerFeature(customPoi);
+        });
     }
 
-    public static void registerFeature(CustomPoi customPoi) {
+    private void registerFeature(CustomPoi customPoi) {
         int tier =
                 switch (customPoi.getIcon()) {
                     case CHEST_T1 -> 1;
--- a/common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
+++ b/common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
@@ -32,11 +32,11 @@ public class WaypointsProvider extends BuiltInProvider {
         return PROVIDED_FEATURES.stream();
     }
 
-    public void updateWaypoints(List<CustomPoi> customPois) {
+    public void updateWaypoints(List<CustomPoi> waypoints) {
         PROVIDED_FEATURES.forEach(feature -> notifyCallbacks(feature));
         PROVIDED_FEATURES.clear();
-        customPois.forEach(customPoi -> {
-            registerFeature(customPoi);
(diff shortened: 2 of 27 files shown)


common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
```diff
@@ -32,11 +32,15 @@ public Stream<MapFeature> getFeatures() {
...
     }
 
-    public static void resetFeatures() {
+    public void updateWaypoints(List<CustomPoi> customPois) {
+        PROVIDED_FEATURES.forEach(feature -> notifyCallbacks(feature));
         PROVIDED_FEATURES.clear();
+        customPois.forEach(customPoi -> {
+            registerFeature(customPoi);
+        });
     }
```
kristofbolyai: The main problem is, I don't want to update the list of custom pois (or the equivalent JsonMapLocation) by hand. I can't, storage does not allow me. What I want to do is to tell the provider, hey, this is the list of map locations. Do what you want to do with it. This goes the same way for loot chests.

What do you think, how should we do this? In reality, this PR is partially a preparation for for https://github.com/Wynntils/Artemis/pull/2537, where we need to handle these "edge-cases".

magicus: I'm not sure I understand what the problem should be. You can determine that *some* change has happened, like adding a method call whenever you do storage.touch(). If you then call this updateWaypoints function, it will tell the mapdata service to invalidate all waypoints you have provided before, and then register them all anew.

I think that will solve your problem exactly as you want?

magicus: Also, I have tried to use the term "waypoint" instead of "custom poi".

kristofbolyai: The problem is that I don't have a "storage is loaded, here is the first/original list of pois". Yes, I can determine and call this when a change has happened. The problem is the initial set.

magicus:
I've pushed some changes here that does not really belong to a PR for changing the signalling about change. They were triggered by the fact that I exposed the individual built-in providers from MapDataService.

I think this is good changes that we want. Can you accept to bring them in with this PR (essentially making it solve two completely unrelated issues), or should I split it up?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
magicus: Hm. Maybe we need to actually add some kind of callback for that. It would make sense. The config can be updated from the GUI without the control of the owning class, but that is not the case for storage. However, it can and will be loaded by the manager, without the control of the owning class, and so it seems reasonable that we could get a notification for that.

common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
```diff
@@ -32,7 +32,9 @@ public enum ServiceKind {
     TRADE_MARKET("Trade Market", Texture.TRADE_MARKET, "trade-market"),
     WEAPON_MERCHANT("Weapon Merchant", Texture.WEAPON_MERCHANT, "merchant:weapon"),
     WEAPONSMITHING_STATION("Weaponsmithing Station", Texture.WEAPONSMITHING_STATION, "profession:weaponsmithing"),
-    WOODWORKING_STATION("Woodworking Station", Texture.WOODWORKING_STATION, "profession:woodworking");
+    WOODWORKING_STATION("Woodworking Station", Texture.WOODWORKING_STATION, "profession:woodworking"),
+    MERCHANT("Merchant", null, "merchant"),
+    STATION("Station", null, "profession");
```
kristofbolyai, line 37: What's this change?

common/src/main/java/com/wynntils/utils/render/type/AbstractTexture.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.render.type;
+
+import net.minecraft.resources.ResourceLocation;
+
+public interface AbstractTexture {
```
kristofbolyai, line 9: I've been thinking about some map/poi only texture interface/class/enum for a long time. However, I think this should be done backwards. Imo, there should be a `AbstractMapFeatureTexture` class, that is either a `WrappedTexture` (an icon that is defined in the `Texture` enum) or a `MapIconTexture` (a wrapper for `MapIcon`s). What do you think?

While we are here, we need to handle user provided map icon textures too. I don't think we impose any limits on them, which is bad. We should limit them to a 1:1 aspect ratio, with 16x16, 32x32 and maybe 64x64 pixel textures.

kristofbolyai (commented):
The callbacks look good, although we still might have problems getting them to work with custom markers. I have some worries about the texture interface. If we are not going to do many rounds of reviews, then this PR can stay as a whole. If we need to discuss a part of it for much longer, it'll be beneficial to split and merge parts that are accepted.

common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
magicus: Basically it was for testing that you could do `/location service merchant` and get a list of all merchants. I'm not 100% convinced it is useful, so we can revert this.

kristofbolyai: If the "kind" doesn't exist in the services file we provide, then it should not be kept (for now, at least).

common/src/main/java/com/wynntils/utils/render/type/AbstractTexture.java
magicus: Ok, I should have commented a bit what I was trying to do here. The AbstractTexture and icon wrapper was only intended as a temporary measure, which should be removed before integrating `mapdata-ports` into main.

magicus: The better solution would have been to fix `Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation()` so it did not require a Texture, but an `iconId` or something like that. But that likely requires aligning the current waypoint poi system better with MapData, which is a undertaking on it's own. (It is like a partial, but different, implementation of MapData)

common/src/main/java/com/wynntils/services/map/type/ServiceKind.java
magicus: I'll revert them next time I get around to starting IDEA.

kristofbolyai (approved):
I'll merge this now and remove the ServiceKind manually (I can't push to your branch).

PR #2550, opened 2024-06-01
fix: Fix wrong minimum hover/render alphas

### LINKED ISSUE #2548: [MapData] - Minimum visibility to render is slightly off, allowing hovering unrendered features
image

1 more zoom step
image

kristofbolyai:
Fixes #2548

It seems like OpenGL ignores the first and last 10% render alpha, so `0.1f` is considered as `0f` alpha, and `0.9f` is considered as `1f`.
Taking all these into account, I've found two different, weird behavior:
- Due to `hoverAlphaFactor`, fully visible features were still rendered with `0.9f` alpha. Due to the weird OpenGL alphas, this was considered as `1f`.
- `0.1f` alphas were considered as rendered, but did not render, causing features to become rendered on hover, although there were invisible (#2548)

It seems like `hoverAlphaFactor` was introduced in `LabelPoi`, but I can't find any reason for their existence. It seems magicus ported this design detail, to preserve the original behavior, which is wrong, or confusing, at the very least.

# Debug Screenshots
## Old behavior
image
Observe that features with `0.075f` alpha is fully invisible. It can also be observed that features with "max" visibility are rendered with `0.9f` alpha (but are shown with max alpha).

## New Behavior
image
image
image
Observe that all rendered alphas are actually represented in the rendering of the icons. Also observe that fully visible features are rendered with the correct, `1f` alphas.

### DIFF
--- a/common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
+++ b/common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
@@ -11,6 +11,7 @@ import com.wynntils.services.mapdata.attributes.type.MapDecoration;
 import com.wynntils.services.mapdata.attributes.type.MapIcon;
 import com.wynntils.services.mapdata.attributes.type.ResolvedMapAttributes;
 import com.wynntils.services.mapdata.type.MapFeature;
+import com.wynntils.utils.colors.CommonColors;
 import com.wynntils.utils.colors.CustomColor;
 import com.wynntils.utils.render.FontRenderer;
 import com.wynntils.utils.render.buffered.BufferedFontRenderer;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
```diff
@@ -71,16 +74,25 @@ public static void renderMapFeature(
                     0,
                     iconWidth,
                     iconHeight);
+            BufferedFontRenderer.getInstance()
```
magicus: What happened here?

magicus:
And yeah, I don't know about hoverAlphaFactor, just copied it from the old code as you say.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
kristofbolyai: Left behind debug code to render the zoom level. I've thought I force pushed to remove this.. Anyways, does this look acceptable otherwise?

magicus (approved):
Yes, looks good. Thanks!

PR #2553, opened 2024-06-04
feat: Port PoiCreationScreen to MapData

ShadowCat117:


kristofbolyai:
So, @magicus, how do you like the UI? I'll review the screen code, but as the original author, what are your thoughts on the screen's options?

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -53,8 +53,8 @@
 public abstract class AbstractMapScreen extends WynntilsScreen {
     protected static final float SCREEN_SIDE_OFFSET = 10;
     private static final float BORDER_OFFSET = 6;
-    private static final int MAP_CENTER_X = -400;
-    private static final int MAP_CENTER_Z = -1700;
+    protected static final int MAP_CENTER_X = -360;
+    protected static final int MAP_CENTER_Z = -3000;
```
kristofbolyai: Perhaps put these above the border offset, if they are protected.

common/src/main/java/com/wynntils/screens/maps/CustomWaypointIconScreen.java
```diff
@@ -0,0 +1,338 @@
...
+import net.minecraft.client.gui.components.Tooltip;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.client.resources.language.I18n;
+import net.minecraft.network.chat.Component;
+
+public class CustomWaypointIconScreen extends WynntilsGridLayoutScreen {
+    private static final int ICONS_PER_PAGE = 16;
+    private static final int SCROLLBAR_HEIGHT = 40;
+    private static final float SCROLL_FACTOR = 10f;
+    private static final Pattern ICON_NAME_PATTERN = Pattern.compile("^[a-z0-9/._-]+$");
```
kristofbolyai:
```suggestion
    // Accepted resource location name pattern `^[a-z0-9/._-]+`
    // This is made stricter to avoid weird names
    private static final Pattern ICON_NAME_PATTERN = Pattern.compile("^[a-z0-9-]+$");
```

common/src/main/java/com/wynntils/screens/maps/WaypointCategoryScreen.java
```diff
@@ -0,0 +1,374 @@
...
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.client.resources.language.I18n;
+import net.minecraft.network.chat.Component;
+
+public final class WaypointCategoryScreen extends WynntilsGridLayoutScreen {
+    private static final int CATEGORIES_PER_PAGE = 16;
+    private static final int SCROLLBAR_HEIGHT = 40;
+    private static final float SCROLL_FACTOR = 10f;
+    private static final String DEFAULT_CATEGORY = "wynntils:personal:waypoint";
+    private static final Pattern CATEGORY_PATTERN = Pattern.compile("^[a-zA-Z0-9:-]+$");
```
kristofbolyai: I think I told you A-Z is fine, but maybe we shouldn't do that. a-z0-9:- should be enough.

kristofbolyai:
Starting to look really good. As far as I know, we should not expose priority to be an user set field.

kristofbolyai:
When you change some visibility settings, go to change category, when you come back, your changes won't keep.

ShadowCat117:
> As far as I know, we should not expose priority to be an user set field.

Ah so it that something sort of hard coded in to categories then?

> When you change some visibility settings, go to change category, when you come back, your changes won't keep.

Think I know what's causing that

common/src/main/java/com/wynntils/screens/maps/WaypointCreationScreen.java
```diff
@@ -0,0 +1,1303 @@
...
+                .build();
+        this.addRenderableWidget(nextIconButton);
+
+        addCustomIconButton = new Button.Builder(
+                        Component.translatable("screens.wynntils.poiCreation.addCustomIcon"),
+                        (button) -> McUtils.mc().setScreen(CustomWaypointIconScreen.create(this)))
+                .pos((int) (dividedWidth * 8) + 20, (int) (dividedHeight * 25) + 20)
+                .size((int) (dividedWidth * 10), 20)
+                .build();
+        this.addRenderableWidget(addCustomIconButton);
```
kristofbolyai: This button could be the same width as the displayed icons itself, or at least, center aligned with those buttons.

kristofbolyai:
Overall, all UIs should have titles of some sort, it can be pretty overwhelming at first look, if you don't know what you see.

kristofbolyai:
For the visibility UI, perhaps hide the sliders if always/never is selected. That should make it less overwhelming.

kristofbolyai:
Also, the edit visibility/edit waypoint buttons stay hovered forever for some reason, after clicking.

kristofbolyai:
Aaand, the category selection screen should have a "back" button that changes nothing.

common/src/main/java/com/wynntils/services/mapdata/WaypointsService.java
```diff
@@ -18,18 +18,25 @@
...
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class WaypointsService extends Service {
     @Persisted
     private final Storage<List<WaypointLocation>> waypoints = new Storage<>(new ArrayList<>());
 
+    @Persisted
+    private final Storage<List<JsonIcon>> customIcons = new Storage<>(new ArrayList<>());
```
kristofbolyai, line 38: Obviously, this is not an ideal location to keep these, but we don't really have working JsonProviders at the moment, and we've agreed with magicus that this is an acceptable porting compromise, that we can fix later. I personally don't want to give up upfixability of these waypoints/icons, which is only possible if we keep them a persisted value (storage/config).

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
```diff
@@ -12,6 +13,7 @@
 
 public class WaypointsProvider extends BuiltInProvider {
     private static final List<MapFeature> PROVIDED_FEATURES = new ArrayList<>();
+    private static final List<MapIcon> PROVIDED_ICONS = new ArrayList<>();
```
kristofbolyai, line 16: What we can do is however, provide these icons in a separate provider. Perhaps, the "personal settings provider" we don't have yet. We'll come back to it.

common/src/main/java/com/wynntils/screens/maps/WaypointCreationScreen.java
ShadowCat117: Oh yeah it was meant to be the same size but doesn't work on other gui scales

ShadowCat117:
> Overall, all UIs should have titles of some sort, it can be pretty overwhelming at first look, if you don't know what you see.

As in a "Waypoint Creation" title or titles for each region, like "Label Options", "Icon Options". Or maybe even both

> For the visibility UI, perhaps hide the sliders if always/never is selected. That should make it less overwhelming.

Maybe, I did leave them so people can actually see what "always" or "never" values look like, but I guess if they change from "always" to "custom" the values will be visible there.

> Also, the edit visibility/edit waypoint buttons stay hovered forever for some reason, after clicking.

Lack of a mouseReleased call I presume

ShadowCat117:
> Also, the edit visibility/edit waypoint buttons stay hovered forever for some reason, after clicking.

I think it's actually intended behaviour by Minecraft's default buttons, if you look at some of the buttons in vanilla options they work the same

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/WaypointsProvider.java
magicus: In general, I think this is a good idea. That way, people can like download "icon packs" that provide additional icons, and if they want to add new ones of their own, they will essentially be starting to produce their own icon pack.

In fact, when thinking of it, maybe we should *only* support this method of setting icons for waypoints in the GUI.

magicus:
@ShadowCat117 I updated your branch with a merge of the latest version of mapdata-ports. I have not verified that it works, but despite some merge conflicts I think I resolved them correctly. You might want to check it out, though,

ShadowCat117:
Just to note I'll probably start work on this again on the weekend, there's a few things I'd like to improve so want to spend some time on that

ShadowCat117:
https://github.com/Wynntils/Wynntils/pull/2969

PR #2554, opened 2024-06-04
fix: Rotate minimap based on camera rotation instead of player rotation [skip ci]

jadedbay:
Added the option to rotate the minimap based on camera rotation.
This is useful for when using other mods that lets you rotate the camera without rotating the player eg. BetterThirdPerson.

### DIFF
--- a/common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
+++ b/common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
@@ -154,7 +154,10 @@ public class MinimapOverlay extends Overlay {
         if (followPlayerRotation.get()) {
             poseStack.pushPose();
             RenderUtils.rotatePose(
-                    poseStack, centerX, centerZ, 180 - McUtils.mc().gameRenderer.getMainCamera().getYRot());
+                    poseStack,
+                    centerX,
+                    centerZ,
+                    180 - McUtils.mc().gameRenderer.getMainCamera().getYRot());
         }
 
         // avoid rotational overpass - This is a rather loose oversizing, if possible later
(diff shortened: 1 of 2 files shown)


jadedbay:
I'm using an enum instead of a boolean now:
```java
private enum RotationType {
        STATIC,
        PLAYER,
        CAMERA
}
```
but I could just update `McUtils.player().getYRot()` to `McUtils.mc().gameRenderer.getMainCamera().getYRot()` instead and leave it as a boolean, because I don't really see any point in having the `PLAYER` (old rotation behavior) option since it has the same behavior as `CAMERA` in first and third person (only different in second person and when using mods that adjust how the camera works), and this would simplify the setting for users. Let me know if you want me to update it.

kristofbolyai:
Is second person view also achieved by mods?

jadedbay:
The third person view but you face yourself, sorry wasn't sure what to call it.

kristofbolyai:
So all in all, this is just mod compatibility fix (that happens to fix a weird 3rd person view issue). Okay. Let's revert the boolean then, and keep the fix :)

jadedbay:
Sure I'll make the changes now do you want me to just keep the variable named followPlayerRotation or change it to followCameraRotation?

kristofbolyai:
Don't rename the config, otherwise it will be reset for users (unless you add upfixers for it). You can correct the en_us file though.

magicus:
So what happens with this fix if I press F5 and view myself from the front? From the back?

jadedbay:
There's no difference if you're facing the back of the player (since the camera and player are facing the same direction) but if you are looking at the player's face (F5 twice from first person) the minimap will be rotated so that its pointing the same way the camera is and not where the player is looking.

magicus:
Hm, okay. Are you OK with this @kristofbolyai? If so, then so am I.

kristofbolyai (approved):
I think this is acceptable, the minimap should follow the camera rotation, not the player body (as it's an UI element).

kristofbolyai:
Thank you for fixing this!

PR #2555, opened 2024-06-04
fix: Include authToken in Athena requests [skip ci]

kristofbolyai:
This is done as part 1 of some Athena patches. Without auth-tokens, we can rate-limit users later on. This is not a new change, in fact, legacy had been sending auth tokens, only Artemis was missing this.

I've had to change WynntilsAccount to a manager, otherwise NetManager would force class loading of Services. This is a compromise we have to live with, or we can do other hacks..

I would like an explicit review from @magicus about the manager/service changes, and from @DevScyu to confirm the auth tokens are sent correctly.

### DIFF
--- a/common/src/main/java/com/wynntils/core/net/NetManager.java
+++ b/common/src/main/java/com/wynntils/core/net/NetManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.core.net;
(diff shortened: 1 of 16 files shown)


DevScyu (approved):
Web Headers LGTM

common/src/main/java/com/wynntils/core/net/NetManager.java
```diff
@@ -106,28 +106,40 @@ public void openLink(UrlId urlId, Map<String, String> arguments) {
...
                 .uri(uri)
                 .timeout(Duration.ofMillis(REQUEST_TIMEOUT_MILLIS))
                 .header("User-Agent", USER_AGENT)
                 .header("Content-Type", "application/json")
-                .POST(HttpRequest.BodyPublishers.ofString(jsonArgs.toString()))
-                .build();
+                .POST(HttpRequest.BodyPublishers.ofString(jsonArgs.toString()));
+
+        // If we are just bootstrapping, we don't have a token (or WynntilsAccountManager) yet
+        if (Managers.WynntilsAccount != null && athenaUrl) {
```
magicus: I'm not entirely happy how knowledge about athena auth is hard-coded in the NetManager. I wonder if we should add a `Map<String, String> headers` argument to the NetManager methods instead. At some point of course we need to determine if this is an athena call and if we have a token, but this feels like the wrong place.

magicus: Even just adding a `createAthenaPostRequest` here that forwards to `createPostRequest` and adds an Athena header would be an improvement. But I think there could be a better place. Let me look around a bit.

magicus:
If I read the code correctly, all paths that come in with the isAthenaUrl argument as true come via NetManager.callApi(). I'd propose that you instead propagate the headers argument as I suggested above to this function, and add a similar `callApi` method to `WynntilsAccountService`, that sets the athena header.

The you'll just have to change a call like
```
Managers.Net.callApi(UrlId.API_ATHENA_UPDATE_CHECK, Map.of("stream", stream));
```
into 
```
Services.WynntilsAccount.callApi(UrlId.API_ATHENA_UPDATE_CHECK, Map.of("stream", stream));
```

I think this will be both clearer and less hacky.

magicus (approved):
Much better, thank you!

PR #2556, opened 2024-06-04
fix: Make Minimap pointer follow player rotation [skip ci]

jadedbay:
Rotates the player pointer on minimap to align with player rotation


kristofbolyai (approved):
Thank you for fixing it!

PR #2558, opened 2024-06-04
feat: Port labels, services and combat locations to mapdata

kristofbolyai:
It's really coming together :)

I've ported all Static-Storage sources to mapdata in this PR. I've also been doing some general cleanup, slowly chipping away from the old classes. I've also needed to do some structural changes, let me know what you think. 

I also had the idea to have all 3 types in different models, but keeping them together may be better overall.

## Change Showcase
- Caves, dungeons, boss alters display their levels
image
image
image

- Lootrun camps
image

- Boss altars show up (these had been coded previously, but I don't remember them showing up)
image


common/src/main/java/com/wynntils/core/net/UrlId.java
```diff
@@ -32,7 +32,7 @@ public enum UrlId {
     DATA_STATIC_ABILITIES("dataStaticAbilities"),
     DATA_STATIC_CAVE_INFO("dataStaticCaveInfo"),
     DATA_STATIC_CHARMS("dataStaticCharms"),
-    DATA_STATIC_COMBAT_LOCATIONS("dataStaticCombatLocations"),
+    DATA_STATIC_MAPDATA_COMBAT_LOCATIONS("dataStaticMapdataCombatLocations"),
```
magicus: I'd like to avoid "mapdata" as a term other than when we are doing this transition, so we do not encode it in user-fronting code, or essentially in anything except the `mapdata` package. It kind of makes sense now, but when it is the only kind of map data we have it will sound wonky.

In contrast, calling these entries for "map features" is something I encourage.

What about `dataStaticCombatMapFeatures` or `dataStaticCombatMapLocations`?

kristofbolyai: I specifically include mapdata in the URLs because otherwise Static-Storage becomes quite hard to differentiate. I guess "MapFeatures" can work.

common/src/main/java/com/wynntils/core/net/UrlId.java
```diff
@@ -42,9 +42,9 @@ public enum UrlId {
     DATA_STATIC_LOOTRUN_TASKS_NAMED("dataStaticLootrunTasksNamed"),
     DATA_STATIC_MAPS("dataStaticMaps"),
     DATA_STATIC_MATERIAL_CONVERSION("dataStaticMaterialConversion"),
-    DATA_STATIC_PLACES("dataStaticPlaces"),
+    DATA_STATIC_MAPDATA_PLACES("dataStaticMapdataPlaces"),
     DATA_STATIC_SEASKIPPER_DESTINATIONS("dataStaticSeaskipperDestinations"),
-    DATA_STATIC_SERVICES("dataStaticServices"),
+    DATA_STATIC_MAPDATA_SERVICES("dataStaticMapdataServices"),
```
magicus: Also, on these two, maybe we should try to keep a consistent naming scheme, so like `dataStaticServiceMapFeatures` (`dataStaticServiceMapLocations`, if that is what you prefer).

common/src/main/java/com/wynntils/services/mapdata/MapFeatureService.java
```diff
@@ -0,0 +1,53 @@
...
+        loadServices();
+        loadCombatLocations();
+    }
+
+    private void loadPlaces() {
+        Download dl = Managers.Net.download(UrlId.DATA_STATIC_MAPDATA_PLACES);
+        dl.handleReader(reader -> {
+            TypeToken<List<PlaceLocation>> type = new TypeToken<>() {};
+            List<PlaceLocation> places = Managers.Json.GSON.fromJson(reader, type.getType());
+            Services.MapData.PLACE_LIST_PROVIDER.updatePlaces(places);
```
magicus, line 32: I had some kind of idea that the Providers themselves should be responsible for loading the json file. The `update` hooks were just a way to get them to work with the existing framework.

Maybe it does not turn out good, but that was at least my idea.

magicus: So, like, either in the constructor of e.g. CombatListProvider, or in MapDataService, when we do `registerBuiltInProvider(COMBAT_LIST_PROVIDER)`.

kristofbolyai: I've been thinking about something like this, but I've failed to turn out anything solid. Also, I just remember that I did not implement the "reload" hook in this service.

common/src/main/java/com/wynntils/core/net/UrlId.java
magicus: MapFeatures is better than map data, so if that is okay, please go with it.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureService.java
magicus: But you intend to look at this? Or did you mean that you thought that my suggestion was a bad idea that was too hard to implement?

kristofbolyai: > But you intend to look at this? Or did you mean that you thought that my suggestion was a bad idea that was too hard to implement?

I had the idea to use something like `createOnlineProvider` to download these. But I really have no idea how they are meant to be used, and I have yet to figure it out, because it's only strictly needed for porting "poi providers". All in all, I think with `MapFeatureService` does not look too bad, so if we keep it as is, I won't have sleepless nights about it.

magicus: Ok, let's take this in as it is, and then I can play around a bit with it and see if I find some different solution I'm more happy with.

magicus (approved):

PR #2562, opened 2024-06-05
feat: Port Hades player waypoints to mapdata

kristofbolyai:
These are going to be one of the most hacky changes we need to do for mapdata. Overall, I think they work well with mapdata. On the main map, the player label is the real location of the player. On the minimap, it's the center of the head (like it used to be). This is a compromise I had to do to make labels obey label attributes.

image
image


common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
```diff
@@ -71,10 +118,168 @@ public Optional<MapVisibility> getIconVisibility() {
...
+                poseStack.popPose();
+            }
+        }
+    }
+
     private static final class CharacterLocation implements MapLocation {
         @Override
         public String getFeatureId() {
-            return "built-in:character:location";
+            return "player-you";
```
kristofbolyai: Every other map feature used id's in a format like this, not the old "category-like" one.

common/src/main/java/com/wynntils/services/mapdata/MapFeaturePoiWrapper.java
```diff
@@ -114,6 +114,7 @@ public void renderAt(
             float zoomRenderScale,
             float zoomLevel,
             boolean showLabels) {
+        // showLabels = fullScreenMap, to be removed when POIs are fully removed, and replaced by MapFeatureRenderer
```
kristofbolyai, line 117: This only stays while we are porting. When we get rid of Pois and MapFeaturePoiWrapper, we can change showLabels to fullscreenMap officially.

magicus:
I'll look closer into this and others later. Right now I'm at a four-day rock music festival with my wife and daughter. :)

kristofbolyai:
> I'll look closer into this and others later. Right now I'm at a four-day rock music festival with my wife and daughter. :)

I'll make sure to port everything else over while you are busy then. After you are back, I still want to discuss that issue with the minimap. Also, we forgot about adding the "in-world marker" rendering that we talked about some time about Discord. We'll discuss it when you are free. 

Most importantly, have fun!

magicus:
> Also, we forgot about adding the "in-world marker" rendering that we talked about some time about Discord.

I think it is as easy to add a MapVisibility `distance`, but where we will interpret min and max as percentage of a predefined distance (perhaps 1000 meters)?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
magicus: Hm? I thought I tried to use a hierarchical, colon-separated naming scheme for feature id:s?

magicus:
> On the main map, the player label is the real location of the player. [...] This is a compromise I had to do to make labels obey label attributes.

I'm not sure I understand why, but it does not sound ideal. I'd assumed that the player was at the center of the icon, even if there is a label. That is what I tried to do with other map features that had both an icon and a label. (As like when hovering.)

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: I don't think so. At least, you did not complain to me before that. I've not used hierarchical naming for feature ids, only categories. Check https://github.com/Wynntils/Static-Storage/blob/main/Reference/combat_mapfeatures.json as an example of the format I use. As featureId is going to be something used in commands for example, I think my format makes sense. But you had this idea originally, so let me know.

magicus:
Btw, I started looking a bit into this some days ago, but never got anywhere. My approach was to try to create a new icon from the player skin (the head with the "hat" rendered on top), into e.g. a NativeImage that could be returned as a ResourceLocation. Never really got a working solution though. It's kind of annoying that we really have what is essentially an icon, and still cannot use it as such. Not sure how much fuss it is worth to try and fix it though.

kristofbolyai:
> Btw, I started looking a bit into this some days ago, but never got anywhere. My approach was to try to create a new icon from the player skin (the head with the "hat" rendered on top), into e.g. a NativeImage that could be returned as a ResourceLocation. Never really got a working solution though. It's kind of annoying that we really have what is essentially an icon, and still cannot use it as such. Not sure how much fuss it is worth to try and fix it though.

Hm, I forgot to include this in the description. I've originally tried to render the head as a MapIcon, creating them dynamically. The problem is that Minecraft populates a resource location for a whole skin, and I don't know a way to edit the native image to only include the "head" + "hat", and not the body, etc. Right now, the resource location is just like a Minecraft skin you would upload when you change your skin.

kristofbolyai:
> > Also, we forgot about adding the "in-world marker" rendering that we talked about some time about Discord.
> 
> I think it is as easy to add a MapVisibility `distance`, but where we will interpret min and max as percentage of a predefined distance (perhaps 1000 meters)?

I think it has to be customizable. Let's say, you have "quest hint" waypoints. You don't want them to show any time, except if you are really zoomed in on the map of that quest location and when you are only a few blocks away from the puzzle door / puzzle location.

kristofbolyai:
> > > Also, we forgot about adding the "in-world marker" rendering that we talked about some time about Discord.
> > 
> > 
> > I think it is as easy to add a MapVisibility `distance`, but where we will interpret min and max as percentage of a predefined distance (perhaps 1000 meters)?
> 
> I think it has to be customizable. Let's say, you have "quest hint" waypoints. You don't want them to show any time, except if you are really zoomed in on the map of that quest location and when you are only a few blocks away from the puzzle door / puzzle location.

In fact, the whole general idea is quest hints. Whether they are achieved by labels, by icons, or by MapPaths (which are not supported at all at the moment, but I'll be working on them shortly. They are basically lootruns, they can be rendered as lootruns. In fact, maybe they should deprecate lootruns at some point.)

magicus:
> I've originally tried to render the head as a MapIcon, creating them dynamically. The problem is that Minecraft populates a resource location for a whole skin, and I don't know a way to edit the native image to only include the "head" + "hat", and not the body, etc. Right now, the resource location is just like a Minecraft skin you would upload when you change your skin.

Yep, that was what got me stuck as well. I tried to find a way to just "render" a part of the texture into a new nativeimage, but there seemed to be no support for that in vanilla minecraft.

I'll look a bit more. If we could find a way to do it, then it would allow folks to override the icons of their friends, just by adding a json with the icon-id for their friend and a personalized icon. It's not a super biggie, but it would be nice to have that option.

magicus:
I made some progress. There is `Screenshot.takeScreenshot` which turns a `RenderTarget` into a `NativeImage`. We could call this, or do something similar. I think we can setup a "virtual" rendertarget and just render the head part of the player texture. But it seems like to much effort, at least right now.

magicus:
> . In fact, maybe they should deprecate lootruns at some point.

Yes, I think lootruns should be a kind of MapPaths.

kristofbolyai:
> I made some progress. There is `Screenshot.takeScreenshot` which turns a `RenderTarget` into a `NativeImage`. We could call this, or do something similar. I think we can setup a "virtual" rendertarget and just render the head part of the player texture. But it seems like to much effort, at least right now.

I am sure we could do some hacks to get it working. I propose we keep this version for the sake of progressing and we can get back to it, if we really want to.

kristofbolyai:
> > . In fact, maybe they should deprecate lootruns at some point.
> 
> Yes, I think lootruns should be a kind of MapPaths.

Once the porting from Poi to MapLocation finished, we are going to have to shift full focus to MapPaths and MapAreas. Perhaps we can do the lootrun port there. Or we keep lootruns separate for a bit, and share it's code with MapPaths.

magicus:
> > On the main map, the player label is the real location of the player. [...] This is a compromise I had to do to make labels obey label attributes.

> I'm not sure I understand why, but it does not sound ideal. I'd assumed that the player was at the center of the icon, even if there is a label. That is what I tried to do with other map features that had both an icon and a label. (As like when hovering.)

Ok, I think I get this now. Since we don't have any icon, the rendering will treat this as a pure-label situation (like for places). 

I'd say the fix is to provide a 100% transparent placeholder icon for the remote player hierarchy, and draw the map decoration in the same place.

I really think we need to player marker to have the icon center at where the player is.

kristofbolyai:
> I'd say the fix is to provide a 100% transparent placeholder icon for the remote player hierarchy, and draw the map decoration in the same place.

Should we provide it as a static sized "texture", or should we make it dynamic, somehow? or have multiple ones like "wynntils:...:transparent-16", "..-20", etc?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
```diff
@@ -4,39 +4,86 @@
...
+    private void reloadHadesUsers() {
+        PLAYERS.forEach(this::notifyCallbacks);
+        PLAYERS.clear();
+        Services.Hades.getHadesUsers().map(HadesPlayerLocation::new).forEach(PLAYERS::add);
+    }
+
     @Override
     public String getProviderId() {
-        return "character";
+        return "characters";
```
magicus: Maybe we should rename this `PlayerProvider` and have `players` as provider id? We've tried to reserve "character" for the user playing the game, and call everyone else "player", so we should try and stick with that.

(I realize I named this wrong)

kristofbolyai: Do you want to separate Hades into PlayerCategory, and keep the "character/us" here?

kristofbolyai: Also, can you explain what the `CharacterLocation` provides? It has no texture, no label, and we still render the "arrow" on the map separately? Do you want to replace that special rendering with this provider, but did not finish writing it?

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
```diff
@@ -71,10 +118,168 @@ public Optional<MapVisibility> getIconVisibility() {
...
+        }
+
+        @Override
+        public String getFeatureId() {
+            return "player-" + hadesUser.getName().toLowerCase(Locale.ROOT);
+        }
+
+        @Override
+        public String getCategoryId() {
+            return "wynntils:players:hades" + (hadesUser.getRelation().isEmpty() ? "" : ":" + hadesUser.getRelation());
```
magicus: I don't think we should expose "hades" here. I know we have done so in some other places, and I don't like it. It's an internal-only name for us developers. I think just `wynntils:players` is good enough. Perhaps we should also change so we *always* have a third part, so if we have no known hades relation we should use `wynntils:players:other`. I'm ot sure how it could show up then (perhaps by implementing a "entity radar" function?) but I think it is better to be prepared.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
magicus: After thinking a bit about this, I think you are correct. My thinking was that the "fully qualified" name of a feature is it's category + feature id, so the feature id should be seen in this context.

That also means that the feature ids for hades players should probably be just the player name, not "player-", so we would get like `wynntils:player:friend:magicus`.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
magicus: Also, I think I tried to always use singular in category ids. If I didn't, we should fix it. It might not feel like the best fit in all situations, but that is basically the only way to be consistent which I think is more important.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
magicus: > Do you want to replace that special rendering with this provider, but did not finish writing it?

That was my thinking, but in retrospect I'm not sure it was a good idea. I'd say at this point it is better to remove it, and focus on making this provider do the "other players" things. If we figure out a need for the user marker to be a map feature, we can restore it.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Okay. I can fix that up.

magicus:
> Should we provide it as a static sized "texture", or should we make it dynamic, somehow? or have multiple ones like "wynntils:...:transparent-16", "..-20", etc?

Well, it needs to be the same size as `playerHeadRenderSize`... Do we want to allow players to change that? If not, we can create a fixed icon, I guess. Otherwise I assume we need to provide it dynamically.

I must say I have not really thought this through completely. But it really breaks things that we cannot provide the player head as an icon. :(

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Okay. I'll fix that up here, and the other PRs as well (in Static-Storage).

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Fixed.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Fixed.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Fixed (here).

kristofbolyai:
I've pushed a fix for the texture positioning.

common/src/main/resources/assets/wynntils/textures/icons/map/steve_head.png
magicus, line 1: That's not Steve! :-o

kristofbolyai: Wynn-Steve 😉

kristofbolyai: (it's from the Wynncraft resource pack)

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -76,13 +76,7 @@ public class MinimapOverlay extends Overlay {
...
 
     @Persisted
-    public final Config<Boolean> renderRemoteFriendPlayers = new Config<>(true);
-
-    @Persisted
-    public final Config<Boolean> renderRemotePartyPlayers = new Config<>(true);
-
-    @Persisted
-    public final Config<Float> remotePlayersHeadScale = new Config<>(0.4f);
+    public final Config<Float> remotePlayersHeadScale = new Config<>(0.8f);
```
magicus: This got me thinking. It feels wrong to treat player map features specially like this. Is this really a useful thing to change? If so, maybe all map features should have a iconScale attribute?

common/src/main/java/com/wynntils/overlays/PartyMembersOverlay.java
```diff
@@ -78,7 +78,12 @@ protected List<PartyMemberOverlay> getPreviewChildren() {
     }
 
     @SubscribeEvent
-    public void onHadesUserAdded(HadesUserAddedEvent event) {
+    public void onHadesUserAdded(HadesUserEvent.Added event) {
```
magicus, line 81: Since you do the same for both "subevents", you can do like:
```
 public void onHadesUserChange(HadesUserEvent event) {
 ```

File: common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java (hunk 5 above)
Comment (kristofbolyai) [reply common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java]: That's something we should decide on right now. I've just started working on natively using map features (https://github.com/Wynntils/Artemis/pull/2578), and I've had to remove "poiScale" configs, because we can't treat them. Icons can be scaled for map locations, but what about map paths, and map areas? Map areas can have icons, in the center, but map paths, thats a weirder story. Overall, I have many ideas for map paths and map areas now, so you'll probably see an endless stream of PRs from me, depending on each other.

I think if we want to keep "icon scale", we should definitely make it an attribute.

File: common/src/main/java/com/wynntils/overlays/PartyMembersOverlay.java (hunk 6 above)
Comment (kristofbolyai) [reply common/src/main/java/com/wynntils/overlays/PartyMembersOverlay.java:81]: You can't :). I've thought this is the case! But turns out, event bus does not work like that, at least not now. I have no idea if this is an effect of the LexForge/NeoForge split, Lex have been updating the event bus... Anyways, this only works this way.

File: common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java [hunk 7]
```diff
@@ -0,0 +1,266 @@
...
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class PlayerProvider extends BuiltInProvider {
+    private static final List<RemotePlayerLocation> PLAYERS = new ArrayList<>();
+
+    // Provided features and categories
+    private static final List<MapCategory> PROVIDED_CATEGORIES = List.of(new PlayersCategory());
+
+    @SubscribeEvent
+    public void onHadesAuthenticatedEvent(HadesEvent.Authenticated event) {
```
magicus, line 38: Same here, you can combine these into a single `HadesEvent` listener. (And of course with `HadesUserEvent` below, too.)

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
+import java.util.stream.Stream;
+import net.minecraft.client.renderer.MultiBufferSource;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+public class PlayerProvider extends BuiltInProvider {
+    private static final List<RemotePlayerLocation> PLAYERS = new ArrayList<>();
```
magicus: For consistency with other providers, maybe rename it `PROVIDED_FEATURES` and declare it as `List<MapFeature>`? Unless you need to treat it like RemovePlayerLocation somewhere, if so, you can keep the name as well.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+        @Override
+        public Optional<MapAttributes> getAttributes() {
+            return Optional.of(new AbstractMapAttributes() {
+                @Override
+                public Optional<String> getLabel() {
+                    return Optional.of(hadesUser.getName());
+                }
+
+                @Override
+                public Optional<CustomColor> getLabelColor() {
```
magicus: This can move to the category. In general, a feature should override/define *only* the attributes that are essential to have different for this particular feature. In this case, overriding label is correct. (And in theory icon, if we could have fixed so that we could actually provide an icon.)

Basically, we need to add a category `wynntils:player:friend` which sets label color to the relation color of friend, etc. 

I think this means that the knowledge of coding hades relation to color can reside purely in the category (as it should), and that hadesUser.getRelationColor() can be removed.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+                    return Optional.of(hadesUser.getRelationColor());
+                }
+
+                @Override
+                public Optional<String> getIconId() {
```
magicus: Same here, this can be moved to the top `wynntils:player`.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+                    return Optional.of("wynntils:icon:player:head");
+                }
+
+                @Override
+                public Optional<MapDecoration> getIconDecoration() {
+                    return Optional.of(new MainMapIconDecoration());
```
magicus: I think I'd prefer to make `MainMapIconDecoration` static and pass the hadesUser to the constructor instead. Using nested classes with an implicit parent this is tricky, I think, and should only be used if it is the only possible solution.

Also perhaps rename the class to `PlayerMapDecoration`?

magicus: ... in fact, seeing how this is used, maybe we should even rename the class from `MapDecoration` to `MapDecorator`? If you agree, we can do that in a follow-up change, so we don't mix more changes into this PR.

kristofbolyai: > ... in fact, seeing how this is used, maybe we should even rename the class from `MapDecoration` to `MapDecorator`? If you agree, we can do that in a follow-up change, so we don't mix more changes into this PR.

Sure, I don't mind either way.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+
+                poseStack.pushPose();
+                // center the player icon
+                poseStack.translate(-playerHeadRenderSize / 2f, -playerHeadRenderSize / 2f, 0);
+
+                HadesUser user = RemotePlayerLocation.this.hadesUser;
+                ResourceLocation skin = SkinUtils.getSkin(user.getUuid());
+
+                if (!fullscreenMap) {
+                    // outline
```
magicus, line 151: I've been thinking about this a bit, like could we give the user the option to control whether outline and health bar should be drawn or not? I toyed with the idea of passing custom attributes but I don't think that's a good idea.

If I said that we can split this up into three different providers, `wynntils:player:head:<relation>`,  `wynntils:player:outline:<relation>` and  `wynntils:player:health:<relation>`, what would you say? Is that a good idea? My thought was that you can then select if you want the outline, head and health bar individually, if we allow different category selections for the minimap and the main map.

I'm not sure this is a good idea, but it feels kind of okay to me right now at least.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,266 @@
...
+    public Stream<MapFeature> getFeatures() {
+        return PLAYERS.stream().map(MapFeature.class::cast);
+    }
+
+    @Override
+    public Stream<MapCategory> getCategories() {
+        return PROVIDED_CATEGORIES.stream();
+    }
+
+    private static final class PlayersCategory implements MapCategory {
```
magicus: I have put all other "built-in" categories in `CategoriesProvider`. It is of course arguable if this is a good design. An alternative is to get rid of CategoriesProvider, and split it up and put the categories in the closest matching feature providers. But if we should keep that design, then this provider should move there too, including the changes I suggested below.

common/src/main/resources/assets/wynntils/textures/icons/map/steve_head.png
magicus: Okaaaay. The icon itself is fine, but I think the name is a bit misleading. Maybe name it "default_head" or something like that? (And propagate that name up into the enum etc)

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
kristofbolyai: It was you that put this here, I wondered why, but I see, it was just a mistake. I'll move it.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
kristofbolyai: I think this concept will break very easily, when we are about to add changes like "iconScale". It'll also be confusing when hovering/clicking on the map, as it won't propagate down. I see what you want to achieve here, but I am against splitting this into pieces.

common/src/main/resources/assets/wynntils/textures/icons/map/steve_head.png
kristofbolyai: player_head should work.

kristofbolyai:
I've done the review changes. Only thing left is to decide on iconScale.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
magicus: Ok, your objections sound reasonable. I withdraw that idea. :)

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
magicus: We could keep a global icon scale, as a config of the main map and minimap, that will scale all icons. That would be good to have, I think.

I am not completely sure it makes sense to scale other icons to different sizes. Was it ever a good idea to make player heads differently sized from other icons? If we think it was, and want to keep that, then we will need an iconScale attribute.

common/src/main/java/com/wynntils/overlays/PartyMembersOverlay.java
magicus: Oh, really? 🤔 That sounds like a new thing, I'm pretty sure it used to work.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CategoriesProvider.java
```diff
@@ -359,4 +366,64 @@ public Optional<MapVisibility> getLabelVisibility() {
...
+    private static final class RemotePlayerCategory implements MapCategory {
+        private final PlayerRelation relation;
+
+        private RemotePlayerCategory(PlayerRelation relation) {
+            this.relation = relation;
+        }
+
+        @Override
+        public String getCategoryId() {
+            return "wynntils:players:" + relation.name().toLowerCase();
```
magicus:
```suggestion
            return "wynntils:player:" + relation.name().toLowerCase();
```

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/PlayerProvider.java
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        @Override
+        public String getFeatureId() {
+            return hadesUser.getName().toLowerCase(Locale.ROOT);
+        }
+
+        @Override
+        public String getCategoryId() {
+            return "wynntils:players:" + (hadesUser.getRelation().name().toLowerCase(Locale.ROOT));
```
magicus:
```suggestion
            return "wynntils:player:" + (hadesUser.getRelation().name().toLowerCase(Locale.ROOT));
```

magicus (approved):
I think this looks good now, apart from any iconScale changes. Maybe push this as it is, and then we can discuss changes to iconScale separately?

kristofbolyai:
Okay, I'll merge this, and make an issue for iconSize.

PR #2564, opened 2024-06-06
fix: Fix bomb tracking issues [skip ci]

cmkohnen:
This fixes two issues:
- When a bomb is thrown in the current world, it may take a few seconds to actually register due to the infobar being slow sometimes, so if you were to relay the bomb to e. g. guild chat, the bomb still hasnt registered yet and you'll be sending either nothing or the previous bomb.
- Due to delay in the bombell the timer might be a little longer than the actual bomb duration. This isn't the problem. The problem is if a new bomb is thrown directly afterwards, and the bell reaches the user before the local timer has run out, the bomb will simply get ignored.

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
+++ b/common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
@@ -4,10 +4,14 @@
  */
 package com.wynntils.models.items.items.gui;
 
+import com.google.gson.JsonElement;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
 import com.wynntils.models.territories.type.GuildResource;
 import com.wynntils.models.territories.type.GuildResourceValues;
 import com.wynntils.models.territories.type.TerritoryUpgrade;
 import com.wynntils.utils.type.CappedValue;
+import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
 
@@ -126,5 +130,12 @@ public class TerritoryItem extends GuiItem {
                 + treasuryBonus + ", alerts="
                 + alerts + ", upgrades="
                 + upgrades + '}';
+    }
+
+    public static class TerritorySerializer implements JsonSerializer<TerritoryItem> {
+        @Override
+        public JsonElement serialize(TerritoryItem territoryItem, Type type, JsonSerializationContext context) {
+            return null;
+        }
     }
 }
--- a/common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
+++ b/common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
@@ -4,14 +4,10 @@
  */
 package com.wynntils.models.items.items.gui;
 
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
```diff
@@ -127,4 +131,11 @@ public String toString() {
                 + alerts + ", upgrades="
                 + upgrades + '}';
     }
+
+    public static class TerritorySerializer implements JsonSerializer<TerritoryItem> {
+        @Override
+        public JsonElement serialize(TerritoryItem territoryItem, Type type, JsonSerializationContext context) {
+            return null;
+        }
+    }
```
kristofbolyai: Left behind?

common/src/main/java/com/wynntils/models/worlds/BombModel.java
```diff
@@ -82,15 +86,25 @@ public void onChat(ChatMessageReceivedEvent event) {
         }
     }
 
+    private void addBombFromChatClue(String user, String bomb, String server) {
```
kristofbolyai:
```suggestion
    private void addBombFromChat(String user, String bomb, String server) {
```

common/src/main/java/com/wynntils/models/worlds/BombModel.java
```diff
@@ -114,11 +128,17 @@ public BombInfo getLastBomb() {
...
 
             // Ensure no duplicate bombs are added
-            if (bombs.containsKey(key) && bombs.get(key).isActive()) return;
+            if (bombs.containsKey(key)) {
+                if (replaceIfExists) {
+                    bombs.remove(key);
+                } else if (bombs.get(key).isActive()) {
+                    return;
+                }
+            }
```
kristofbolyai:
```suggestion
            removeOldTimers();
            
            if (bombs.containsKey(key) && !replaceIfExists) {
                return;
            }
```

And `bombs.put` will override the old bomb.

common/src/main/java/com/wynntils/models/worlds/BombModel.java
```diff
@@ -114,11 +128,17 @@ public BombInfo getLastBomb() {
     private static final class ActiveBombContainer {
         private final Map<BombKey, BombInfo> bombs = new ConcurrentHashMap<>();
 
-        public void add(BombInfo bombInfo) {
+        public void add(BombInfo bombInfo, boolean replaceIfExists) {
```
kristofbolyai: Can you make this method private, and add a public `add(BombInfo)` and `forceAdd(BombInfo) methods? Reading code that has booleans is hard, because you don't know what the boolean does.

common/src/main/java/com/wynntils/models/items/items/gui/TerritoryItem.java
cmkohnen: Yeah, IntelliJ and Git don't seem to like each other

kristofbolyai (approved):
Much better, thank you!

PR #2569, opened 2024-06-06
feat: Add crafting level progress bar [skip ci]

cmkohnen:
Bildschirmfoto vom 2024-06-06 21-56-39

Adds a bar to crafting stations indicating current level progress


kristofbolyai:
Hm. Honestly, these bars always rendered so weirdly. Looks like it's narrow now. I should probably add proper rendering for them, so people can stretch it properly in custom bars too.

kristofbolyai:
I'll see if I can do that tomorrow. If not, this'll be merged as is.

kristofbolyai:
Doesn't this bar look much better? It feels like it's naturally there, because it's the same size as the normal XP bar.

image

kristofbolyai (approved):

magicus:
Nice! Thanks for your contribution!

PR #2570, opened 2024-06-07
feat: Introduce MarkerOptions to Mapdata, merge MarkerModel consumer features

kristofbolyai:
This is basically what I've been talking about for the few weeks. How I introduced the change may need changes, I'll be looking forward to your review, @magicus.

The plan is to make MapAttributes able to define how a marker is rendered in world, if it is. That is by providing beacon colors, whether distance, label, icon is rendered, and with proper fading/visibility. 

With these changes done, I am also working on merging `WaypointDistanceFeature` and `BeaconBeamFeature`. They do essentially the same thing, extending each other's features.

Finally, `MarkerModel` can be removed, and `MarkerProvider`s can be replaced with mapdata providers. That means we got rid off an old system, added many great functionality to mapdata, and cleaned up two very similar features, in one go.

This PR will most likely only contain the systematic changes, as I expect this to be controversial. The actual porting of the `MarkerProvider`s will come as a separate PR. 

# Demo
[Kooha-2024-06-07-12-16-16.webm](https://github.com/Wynntils/Artemis/assets/49001742/c85731c3-a6d1-41e9-8ca8-fbaff7bbbc07)


common/src/main/java/com/wynntils/services/mapdata/attributes/type/MarkerOptions.java
```diff
@@ -0,0 +1,79 @@
...
+    Optional<Float> getFade();
+
+    // The color of the beacon beam
+    // If empty, or alpha is 0, no beacon beam will be rendered
+    Optional<CustomColor> getBeaconColor();
+
+    // Whether to render the label
+    Optional<Boolean> renderLabel();
+
+    // Whether to render the distance (below the label)
```
magicus, line 70: Is this something that we want to change on a per-feature or per-category basis? I would have guessed that it should be a system-wide config, to the in-world drawing feature itself.

magicus:
The `renderDistance` attribute got me thinking. I was thinking like "oh, but if you don't want distance rendering on any markers, you can just add a json provider that gives a top-level category like "wynntils" and defines "renderDistance" to false".

And then I realized that this will not work, since we will try to find the most specific category value first, before looking at a "globally" overriding provider.

I started to think that this is kind of a problem. Maybe we should always try to resolve this in provider order, and go up the hierarchy per provider instead of going throught the category hierarchy, and checking each hiearchy level in provider order.

That way you could add a provider last that would override even very general attributes.

I think that ability would be good to have, but I am not convinced it is correct solution to achieve it. I am also a bit too sleepy to fully understand what consequences such a change would have...

magicus:
But if we do it strictly provider-order based, I think we could supply a "built-in" provider that is always applied first, that can contain some simple overrides that we can help the user to make, like hiding entire categories, or changing the global iconSize (if it becomes a thing), or turning on labels on services etc.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MarkerOptions.java
kristofbolyai: I think there is value in making it per feature. Also, it really does not cost anything at this point, and just adds more customization. For example, the quest guide waypoints, most would not want a distance text, but some really specifically might, for example in mazes.

magicus:
Do we want to do this at the same time as MapData, or can it be done as a follow-up?

magicus:
I tried to merge the current mapdata-ports into this, but failed. This PR interacts heavily with https://github.com/Wynntils/Wynntils/pull/2753, and I could not really figure out how the data flow is supposed to go in this new `WorldMarkersFeature` world compared to the old one in `WorldWaypointDistanceFeature` that was modified by #2753. I'll have to leave that to you, @kristofbolyai.

magicus:
I did merge everything up until that, though, so even though there are three more merge conflicts, they are pretty trivial to resolve. The real problem is only about transplanting the `showAdditionalTextInWorld` functionality.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/MarkerOptions.java
```diff
@@ -0,0 +1,79 @@
...
+
+    // The fade distance for the marker visibility circles
+    Optional<Float> getFade();
+
+    // The color of the beacon beam
+    // If empty, or alpha is 0, no beacon beam will be rendered
+    Optional<CustomColor> getBeaconColor();
+
+    // Whether to render the label
+    Optional<Boolean> renderLabel();
```
magicus: I don't like these names. They sound like operations, verb, rather than a boolean. What about `labelVisible`, `distanceVisible` and `iconVisible` instead?

magicus:
I remember being rather skeptical of this PR when you created it, but now I can't really understand what fuelled that skepticism. I understand what you are trying to achieve, and I think it is overall good. 

Edit: I have some opinions on naming. I'd like for us to get the names right. It's not only the configs, but also the new marker attributes.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,535 @@
...
+
+    private boolean isInBound(Position position, Window window) {
+        return position.x() > 0
+                && position.x() < window.getGuiScaledWidth()
+                && position.y() > 0
+                && position.y() < window.getGuiScaledHeight()
+                && position.z() < 1;
+    }
+
+    private double getMarkerVisibility(Location location, ResolvedMarkerOptions resolvedMarkerOptions) {
```
magicus: This feels like it should be part of the mapdata services, somewhere at least, rather than a Feature.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,535 @@
...
+import org.joml.Vector4f;
+
+@ConfigCategory(Category.MAP)
+public class WorldMarkersFeature extends Feature {
+    private static final float MINIMUM_RENDER_VISIBILITY = 0.1f;
+    private static final int RAINBOW_CHANGE_RATE = 10;
+
+    private static final WaypointPoi DUMMY_WAYPOINT = new WaypointPoi(() -> null, "");
+
+    @Persisted
```
magicus, line 63: I'm not sure all these config names make sense in this new combined beacon + marker world. Also, there seem to be a bit of old-style names like waypoint. I'm too tired now but I'll return with some more naming suggestions.

We should be clear with what we mean by terms like "waypoint" and "marker"; right now it feels like it is consistent to 90%, but not 100%.

kristofbolyai:
The merge is completed, see https://github.com/Wynntils/Wynntils/issues/2932 for additional details.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: I remember discussions about this method's implementation, but I have no idea if it's location was ever discussed. As of right now, I can't think of anything else rendering map features in-world, so this method might very well not be needed anywhere else.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: I think I resolved your issues with the consistency. I've also fixed i18n and added an upfixer.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,539 @@
...
+                    0.166f,
+                    0.33f);
+
+            poseStack.popPose();
+        }
+
+        BUFFER_SOURCE.endBatch();
+    }
+
+    private void updateRenderedMapLocations(Matrix4f projectionMatrix, Camera camera) {
```
magicus, line 188: This is done one for every rendered frame? My gut feeling is that this is way too heavy for that.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
magicus: Do we not do something very similar when calculating visibility for icons on the map? Where is that done? In the MapScreen code? If so, I think it too should move.

This is not so much about code re-usability as encoding the "knowledge" on how to convert the attributes into a visibility percentage. It's an integral part on how the `fade` attribute is supposed to work.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
```diff
@@ -66,33 +69,44 @@ private <T> T getAttribute(Function<MapAttributes, Optional<T>> attributeGetter)
...
 
+    private ResolvedMarkerOptions getResolvedMarkerOptions(
+            Function<MapAttributes, Optional<MarkerOptions>> attributeGetter) {
+        return new ResolvedMarkerOptions(
+                getInheritedValue(MarkerOptions::getInnerRadius, attributeGetter),
+                getInheritedValue(MarkerOptions::getOuterRadius, attributeGetter),
+                getInheritedValue(MarkerOptions::getFade, attributeGetter),
+                getInheritedValue(MarkerOptions::getBeaconColor, attributeGetter),
+                getInheritedValue(MarkerOptions::labelVisible, attributeGetter),
+                getInheritedValue(MarkerOptions::distanceVisible, attributeGetter),
```
magicus, line 80: I'm still not really happy about the names. Let start with the two first. Sure, it technically describes a radius, but that could also just be described as the player's distance from the target, which I think seems more intuitive. So that then points towards something like `getSmallestDistance` and `getLargestDistance`. But that is unwieldy. So maybe just try to and align it with the visibility attribute and call it `getMin` and `getMax`? 

Or... maybe `getMin/MaxDistance`, and rename the visibility attributes `getMinZoom` and `getMaxZoom`, to clarify the difference a bit more?

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
```diff
@@ -66,33 +69,44 @@ private <T> T getAttribute(Function<MapAttributes, Optional<T>> attributeGetter)
...
+                getInheritedValue(MarkerOptions::iconVisible, attributeGetter));
```
magicus: I'm still not satisfied (I know I suggested this myself, yeah yeah :-D).

What about `hasIcon` and `hasLabel`? I think that would be the best.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
```diff
@@ -66,33 +69,44 @@ private <T> T getAttribute(Function<MapAttributes, Optional<T>> attributeGetter)
         return attributeGetter.apply(DefaultMapAttributes.INSTANCE).get();
     }
 
+    private ResolvedMarkerOptions getResolvedMarkerOptions(
+            Function<MapAttributes, Optional<MarkerOptions>> attributeGetter) {
+        return new ResolvedMarkerOptions(
+                getInheritedValue(MarkerOptions::getInnerRadius, attributeGetter),
+                getInheritedValue(MarkerOptions::getOuterRadius, attributeGetter),
```
magicus: I'm still not really happy about the names. Let start with the two first. Sure, it technically describes a radius, but that could also just be described as the player's distance from the target, which I think seems more intuitive. So that then points towards something like `getSmallestDistance` and `getLargestDistance`. But that is unwieldy. So maybe just try to and align it with the visibility attribute and call it `getMin` and `getMax`? 

Or... maybe `getMin/MaxDistance`, and rename the visibility attributes `getMinZoom` and `getMaxZoom`, to clarify the difference a bit more?

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
magicus: This still sounds like it's saying at what distance it becomes visible. Aligning with `hasIcon` and `hasLabel`, maybe `hasDistanceLabel`, or `hasDistanceDisplay`?

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,539 @@
...
+    @Persisted
+    public final Config<Float> textRenderScale = new Config<>(1.0f);
+
+    @Persisted
+    public final Config<Float> iconRenderScale = new Config<>(1.0f);
+
+    @Persisted
+    public final Config<TextShadow> textShadow = new Config<>(TextShadow.NONE);
+
+    @Persisted
```
magicus, line 75: These kind of conflicts with the attributes, yes? If you'd want to override those, should you just not have it as an overriding category with map attributes that ... hm... you can't really set a clipping limit there. Right. Ok, fine.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,539 @@
...
+    public final Config<Float> topBoundingDistance = new Config<>(40f);
+
+    @Persisted
+    public final Config<Float> horizontalBoundingDistance = new Config<>(30f);
+
+    @Persisted
+    public final Config<Integer> maxMarkerDistance = new Config<>(5000);
+
+    @Persisted
+    public final Config<CustomColor> markerBeaconColor = new Config<>(CommonColors.RED);
```
magicus: But this surely conflicts with the map attributes, right?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1423,28 +1419,28 @@
...
+  "feature.wynntils.worldMarkers.description": "Adds the ability to see the distance to your waypoint in the world.",
+  "feature.wynntils.worldMarkers.horizontalBoundingDistance.description": "How far should off-screen indicator be from two sides of the screen?",
+  "feature.wynntils.worldMarkers.horizontalBoundingDistance.name": "Side Indicator Distance",
+  "feature.wynntils.worldMarkers.iconRenderScale.description": "How large should the marker's icon be?",
+  "feature.wynntils.worldMarkers.iconRenderScale.name": "Icon Rendering Scale",
+  "feature.wynntils.worldMarkers.markerBeaconColor.description": "What default color should the beacon beam have?",
+  "feature.wynntils.worldMarkers.markerBeaconColor.name": "Marker Beacon Beam Color",
+  "feature.wynntils.worldMarkers.maxMarkerDistance.description": "At what distance should the waypoint text stop rendering?",
+  "feature.wynntils.worldMarkers.maxMarkerDistance.name": "Maximum Distance for Waypoint Text",
+  "feature.wynntils.worldMarkers.name": "Waypoint Distance",
```
magicus: This does not seem like an adequate name and description of the feature. "World Markers and Beacons"? Or just "World Markers" and have the description as "Adds the ability to show in-world markers and beacons to certain map locations"

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1423,28 +1419,28 @@
...
-  "feature.wynntils.worldWaypointDistance.textColor.description": "What color should distance text be?",
-  "feature.wynntils.worldWaypointDistance.textColor.name": "Distance Text Color",
-  "feature.wynntils.worldWaypointDistance.textShadow.description": "What should the text shadow look like?",
-  "feature.wynntils.worldWaypointDistance.textShadow.name": "Text Shadow",
-  "feature.wynntils.worldWaypointDistance.topBoundingDistance.description": "How far should off-screen indicator be from top of the screen?",
-  "feature.wynntils.worldWaypointDistance.topBoundingDistance.name": "Top Indicator Distance",
+  "feature.wynntils.worldMarkers.bottomBoundingDistance.description": "How far should off-screen indicator be from bottom of the screen?",
+  "feature.wynntils.worldMarkers.bottomBoundingDistance.name": "Bottom Indicator Distance",
+  "feature.wynntils.worldMarkers.description": "Adds the ability to see the distance to your waypoint in the world.",
+  "feature.wynntils.worldMarkers.horizontalBoundingDistance.description": "How far should off-screen indicator be from two sides of the screen?",
```
magicus: Ah, now I understand what this does. I think these need some renaming too. Let me try. We have three variables relating to the top, bottom and sides. They say how far in on the screen the marker should be drawn. 
`screenMarginTop`, `screenMarginBottom` and `screenMarginSides`, perhaps?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1423,28 +1419,28 @@
...
+  "feature.wynntils.worldMarkers.horizontalBoundingDistance.name": "Side Indicator Distance",
+  "feature.wynntils.worldMarkers.iconRenderScale.description": "How large should the marker's icon be?",
+  "feature.wynntils.worldMarkers.iconRenderScale.name": "Icon Rendering Scale",
+  "feature.wynntils.worldMarkers.markerBeaconColor.description": "What default color should the beacon beam have?",
+  "feature.wynntils.worldMarkers.markerBeaconColor.name": "Marker Beacon Beam Color",
+  "feature.wynntils.worldMarkers.maxMarkerDistance.description": "At what distance should the waypoint text stop rendering?",
```
magicus: I guess it affects not just the text but also the icon? Also, does it affect beacon rendering too? This is not clear from neither the name nor the description. If not, is there some other variable to control beacon max rendering distance?

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: It's basically has been like this since the original feature existed, it's not too much calculations, if anything, it's collecting the map locations that might introduce some overhead, but since basically all of the attributes get cached, it's just a glorified getter. Before merging the overall mapdata branch, I do plan to profile map and marker performance, and if anything comes up then, we can fix it.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
kristofbolyai: I've renamed the two, and left `distanceVisible` explicitly as is, as `hasDistance` doesn't make much sense.

kristofbolyai: And I now see the next comment. Don't mind me.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: It's a default value. If for some reason, beacon rendering options are defined, but the color is NONE, this is used as a fallback/default color.

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: I've tried to fix up the i18n.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
magicus: Can we have that as beacon color in our default base attribute instead?

kristofbolyai: Yes, sure, I haven't thought about that.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
magicus: So what do you say about using min/max of some sort instead of inner/outer radius?

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
magicus: The new `has` names are much better. It needs to be updated everywhere. I assume the getter will need to be `getHasLabel` etc, even though it sounds a bit ungrammatical. The alternative is `isHasLabel` (javabean-style getter) or just `hasLabel`, but then that will look like a record-style getter with no prefix, which will be inconsistent.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: In fact, what if I just remove the default color, and leave it up to the user for it to be "required" for rendering? I think it makes sense that the user has to define the color, the min max ranges, etc.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
kristofbolyai: I only worry that the min/max naming is too similar to MapVisibility's mix/max, and by that, the end result wouldn't be so distinct.

kristofbolyai: Do you perhaps have an alternative name idea?

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: The map calculation code is in `MapDataService#calculateVisibility`. I'll move this code now too, and perhaps rename both of them to better reflect in which "space" they are used in.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
magicus: Since they are part of the marker sub-attribute object, I think that makes it clear enough. In pseudo-json, it'll look like:
```
attributes: {
  name: foo,
  marker: {
    minDistance: 12,
    maxDistance: 33,
    fade: 5
  },
  visibility: {
   minZoom: 1,
   maxZoom: 50,
   fade: 12
 }
}
```

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,513 @@
...
+
+@ConfigCategory(Category.MAP)
+public class WorldMarkersFeature extends Feature {
+    private static final float MINIMUM_RENDER_VISIBILITY = 0.1f;
+    private static final int RAINBOW_CHANGE_RATE = 10;
+
+    private static final MultiBufferSource.BufferSource BUFFER_SOURCE =
+            MultiBufferSource.immediate(new ByteBufferBuilder(256));
+
+    private static final WaypointPoi DUMMY_WAYPOINT = new WaypointPoi(() -> null, "");
```
magicus, line 61: What is this used for, btw? `WaypointPoi` sounds like its from the old design.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
```diff
@@ -0,0 +1,513 @@
...
+            CustomColor color = resolvedMarkerOptions.beaconColor();
+
+            if (color == CustomColor.NONE) {
+                // If the color is set to NONE, we skip rendering the beacon beam
+                // This is not logged as render time logging is heavy
+                continue;
+            }
+
+            int colorInt;
+            if (color == CommonColors.RAINBOW) {
```
magicus, line 163: Can we express this in json btw?

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: We have some code that depends on it, we can't remove it (yet!).

common/src/main/java/com/wynntils/services/mapdata/MapDataService.java
```diff
@@ -259,6 +262,34 @@ public float calculateVisibility(ResolvedMapVisibility mapVisibility, float zoom
         return 0;
     }
 
+    public double getMarkerVisibility(Location location, ResolvedMarkerOptions resolvedMarkerOptions) {
```
magicus: This was the placement I was looking for, thanks! :) To tie in better with the function above, maybe rename it `calculateMarkerVisibility`? It's not really a trivial getter which the name currently kind of implies.

common/src/main/java/com/wynntils/services/mapdata/attributes/MarkerOptionsBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+import com.wynntils.services.mapdata.attributes.type.MarkerOptions;
+import com.wynntils.utils.colors.CustomColor;
+import java.util.Optional;
+
+public class MarkerOptionsBuilder implements MarkerOptions {
+    private Float innerRadius = null;
+    private Float outerRadius = null;
+    private Float fade = null;
+    private CustomColor beaconColor = null;
+    private Boolean renderLabel = null;
```
magicus: hasLabel, etc, here too.

common/src/main/java/com/wynntils/services/mapdata/attributes/type/ResolvedMarkerOptions.java
```diff
@@ -0,0 +1,16 @@
...
+package com.wynntils.services.mapdata.attributes.type;
+
+import com.wynntils.utils.colors.CustomColor;
+
+public record ResolvedMarkerOptions(
+        float innerRadius,
+        float outerRadius,
+        float fade,
+        CustomColor beaconColor,
+        boolean renderLabel,
```
magicus: `hasLabel` etc here too.

common/src/main/java/com/wynntils/services/mapdata/providers/json/JsonMarkerOptions.java
```diff
@@ -0,0 +1,71 @@
...
+import com.wynntils.services.mapdata.attributes.type.MarkerOptions;
+import com.wynntils.utils.colors.CustomColor;
+import java.util.Optional;
+
+public class JsonMarkerOptions implements MarkerOptions {
+    private final Float innerRadius;
+    private final Float outerRadius;
+    private final Float fade;
+    private final CustomColor beaconColor;
+    private final Boolean renderLabel;
```
magicus: Here too

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: We can (and do) in built in providers. That's enough for me.

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
kristofbolyai: What do you think about this, @magicus?

common/src/main/java/com/wynntils/features/map/WorldMarkersFeature.java
magicus: Okay.

magicus (approved):
Ship it!

Issue #2560, opened 2024-06-05
MalformedJsonException + error 'Signature is missing from Property textures' spamming in the console


Aeldit:
## The errors
There are multiple errors like this that appear :
```log
[10:22:10] [ForkJoinPool.commonPool-worker-5/WARN]: Failure in net manager [handleJsonObject], processing API:API_ATHENA_USER_INFO
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 8 path $
	at com.google.gson.JsonParser.parseReader(JsonParser.java:76) ~[gson-2.10.1.jar:?]
	at com.wynntils.core.net.NetResult.lambda$handleJsonObject$2(NetResult.java:57) ~[wynntils-1.2.19-fabric+MC-1.20.2.jar:?]
	at com.wynntils.core.net.NetResult.lambda$handleReader$1(NetResult.java:46) ~[wynntils-1.2.19-fabric+MC-1.20.2.jar:?]
	at com.wynntils.core.net.NetResult.lambda$wrappingHandler$5(NetResult.java:111) ~[wynntils-1.2.19-fabric+MC-1.20.2.jar:?]
	at java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:718) ~[?:?]
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[?:?]
	at java.util.concurrent.CompletableFuture.postFire(CompletableFuture.java:614) ~[?:?]
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:844) ~[?:?]
	at java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:483) ~[?:?]
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373) ~[?:?]
	at java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182) ~[?:?]
	at java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655) ~[?:?]
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622) ~[?:?]
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165) ~[?:?]
[...]

PR #2571, opened 2024-06-07
chore: Don't spam the log with exception stack traces [skip ci]

magicus:
"Fixes" #2560.

### DIFF
--- a/common/src/main/java/com/wynntils/core/net/NetResult.java
+++ b/common/src/main/java/com/wynntils/core/net/NetResult.java
@@ -56,7 +56,7 @@ public abstract class NetResult {
                     try {
                         handler.accept(JsonParser.parseReader(reader).getAsJsonObject());
                     } catch (Throwable t) {
-                        WynntilsMod.warn("Failure in net manager [handleJsonObject], processing " + desc, t);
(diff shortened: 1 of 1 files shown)


kristofbolyai (approved):
Yeah, sure, this is marginally better. Thanks.

magicus:
"marginally better"?! :-D I thought this was actually good -- you get to see that something happens but the log is not dominated by irrelevant stack traces.

But we really need to fix Athena. Or disable the Athena functionality completely.

kristofbolyai:
> "marginally better"?! :-D I thought this was actually good -- you get to see that something happens but the log is not dominated by irrelevant stack traces.

Some people that doesn't understand it will go and panic, whether they see a 1 line error, or a 10 line one :).

> But we really need to fix Athena. Or disable the Athena functionality completely.

I've been advocating that in Discord for a long time. At this point, I really don't know, what else I can do.

ShadowCat117:
As a side note, the error for failing to connect could be made a little better. The Hades one could mention that friends/party members won't be visible on the map to hopefully get less people asking about that.

And then the account one mentions cloud config syncing won't work, which I assume was just copied from legacy as Artemis doesn't have cloud configs

kristofbolyai:
Agreed, all of those messages are bad, outdated, or both.

PR #2572, opened 2024-06-07
feat: Add profession XP gain message feature [skip ci]

cmkohnen:
Bildschirmfoto vom 2024-06-07 15-53-21

Similarly to the compat xp message feature, add profession xp gain as well.
Also enables to filter out chat messages on crafting xp gain.


kristofbolyai:
Hm. Maybe it should be colored differently? But I guess this also works.

kristofbolyai:
I'll check out the code in depth later. Thank you!

cmkohnen:
> Hm. Maybe it should be colored differently? But I guess this also works.

Yeah, I wasn't sure about the coloring (given that i struggle to differentiate them sometimes). Feel free to change them

ShadowCat117:
To be honest we've released a lot of features recently that have had colours chosen by us, most of them could've had config 
choices

In fact, maybe that's another case for a special config type. We could render a preview of the text for something and allow users to select the colour of each pre defined "section"

ShadowCat117:
Code looks good though :)

kristofbolyai (commented):
To be honest, I would rather you merge this with `CombatXpGainMessageFeature` to `ExperienceGainMessageFeature` or something along those lines. Add configs for combat XP / profession XP, to be able to only have one. The delay could be shared, IMO 10 seconds should work. Try using Upfixers to make the configs not reset, but if you can't figure that out, I can help out with that.

cmkohnen:
> Try using Upfixers to make the configs not reset, but if you can't figure that out, I can help out with that.

Is there any documentation on that? I'm a little lost on starting an implementation

ShadowCat117:
You'll want to do it like MapToMainMapRenamedConfigsUpfixer, but instead of mapFeature (old name) it'll be combatXpGainMessageFeature and then mainMapFeature (new name) will be experienceGainMessageFeature

common/src/main/java/com/wynntils/core/persisted/upfixers/config/CombatXpGainToExperienceGainUpfixer.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+package com.wynntils.core.persisted.upfixers.config;
+
+import com.wynntils.core.persisted.upfixers.RenamedPrefixesUpfixer;
+import com.wynntils.utils.type.Pair;
+import java.util.List;
+
+public class CombatXpGainToExperienceGainUpfixer extends RenamedPrefixesUpfixer {
+    private static final List<Pair<String, String>> RENAMED_PREFIXES =
+            List.of(Pair.of("combatXpGainMessage.", "experienceGainMessage."));
```
ShadowCat117:
```suggestion
            List.of(Pair.of("combatXpGainMessageFeature.", "experienceGainMessageFeature."));
```

common/src/main/java/com/wynntils/features/ExperienceGainMessageFeature.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright © Wynntils 2022-2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.features;
```
kristofbolyai: If the config category is Utilities, you should move the class to the corresponding package.

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -180,18 +187,21 @@ public void resetValueFromItem(ItemStack professionInfoItem) {
...
         if (newPercentage == 100) {
             newPercentage = 0;
         }
 
-        professionProgressMap.put(type, new ProfessionProgress(oldValue.level(), newPercentage));
+        professionProgressMap.put(profession, new ProfessionProgress(oldValue.level(), newPercentage));
 
-        rawXpGainInLastMinute.get(type).put(xpGain);
+        rawXpGainInLastMinute.get(profession).put(event.getGainedXpRaw());
     }
```
kristofbolyai: Nit, but can you make sure all event methods are below each other? We usually put them as the first methods of a class. (Here, it means switch up resetValueFromItem and onXpGain's place).

common/src/main/java/com/wynntils/features/ExperienceGainMessageFeature.java
```diff
@@ -0,0 +1,119 @@
...
+            lastCombatMessage = Managers.Notification.queueMessage(
+                    getCombatXpGainMessage(event.getGainedXpRaw(), event.getGainedXpPercentage()));
+        }
+
+        lastCombatXpDisplayTime = System.currentTimeMillis();
+    }
+
+    @SubscribeEvent
+    public void onProfessionXpGain(ProfessionXpGainEvent event) {
+        if (!(professions.get() && Models.WorldState.onWorld())) return;
```
kristofbolyai:
```suggestion
        if (!professions.get() || !Models.WorldState.onWorld()) return;
```

common/src/main/java/com/wynntils/features/ExperienceGainMessageFeature.java
```diff
@@ -0,0 +1,119 @@
...
+    public final Config<Boolean> filterChat = new Config<>(true);
+
+    private long lastCombatXpDisplayTime = 0;
+    private final Map<ProfessionType, Long> lastProfessionXpDisplayTimes = new EnumMap<>(ProfessionType.class);
+
+    private MessageContainer lastCombatMessage = null;
+    private final Map<ProfessionType, MessageContainer> lastProfessionMessages = new EnumMap<>(ProfessionType.class);
+    private float lastRawCombatXpGain = 0;
+    private final Map<ProfessionType, Float> lastRawProfessionXpGains = new EnumMap<>(ProfessionType.class);
+    private float lastPercentageXpGain = 0;
```
kristofbolyai:
```suggestion
    private MessageContainer lastCombatMessage = null;
    private float lastRawCombatXpGain = 0;
    
    private final Map<ProfessionType, MessageContainer> lastProfessionMessages = new EnumMap<>(ProfessionType.class);
    private final Map<ProfessionType, Float> lastRawProfessionXpGains = new EnumMap<>(ProfessionType.class);
    private float lastPercentageXpGain = 0;
```

kristofbolyai (commented):
Should be working correctly after the upfixer fix.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -485,6 +481,16 @@
...
+  "feature.wynntils.experienceGainMessage.combat.description": "Shows combat Xp gain messages as a game update notification.",
+  "feature.wynntils.experienceGainMessage.combat.name": "Show Combat Xp Messages",
+  "feature.wynntils.experienceGainMessage.description": "Shows experience gain messages as a game update notification.",
+  "feature.wynntils.experienceGainMessage.filterChat.description": "Should xp gain messages be hidden from chat?",
+  "feature.wynntils.experienceGainMessage.filterChat.name": "Filter Chat Messages",
+  "feature.wynntils.experienceGainMessage.name": "Experience Gain Messages",
+  "feature.wynntils.experienceGainMessage.professions.description": "Shows profession Xp gain messages as a game update notification.",
+  "feature.wynntils.experienceGainMessage.professions.name": "Show Profession Xp Messages",
+  "feature.wynntils.experienceGainMessage.secondDelay.description": "How long (in seconds) should a XP gain message collect XP before it is a separate message?",
+  "feature.wynntils.experienceGainMessage.secondDelay.name": "XP Gain Message Consolation Time",
```
ShadowCat117: Can you make these all use "XP" for consistency, after that everything should be good to go

common/src/main/java/com/wynntils/features/utilities/ExperienceGainMessageFeature.java
```diff
@@ -0,0 +1,120 @@
...
+import com.wynntils.models.characterstats.event.CombatXpGainEvent;
+import com.wynntils.models.profession.event.ProfessionXpGainEvent;
+import com.wynntils.models.profession.type.ProfessionType;
+import com.wynntils.models.worlds.event.WorldStateEvent;
+import java.util.EnumMap;
+import java.util.Map;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UTILITIES)
+public class ExperienceGainMessageFeature extends Feature {
```
magicus: Sorry for coming in late in the review, but we (or at least I) have strived for using the term "XP" consistently, instead of using "experience". (In Java classes, this means using the camel-case form `Xp`.)

I think it would be great if we stayed that way, otherwise there will be an endless confusion if stuff is named "experience" or "xp".

This comments applies to all changed code in this PR that uses "experience" instead of "xp".

kristofbolyai (approved):
Looks good to me! Thanks.

PR #2573, opened 2024-06-07
refactor: Move json loading to providers

magicus:
Also I realized it was wrong to expose individual providers. What we really want is to get stuff from a certain category.

We still expose two providers, but that is due to incomplete integration with the providers. We should fix that and then all providers can be private once again.


magicus:
The three `loadFoo` functions are basically copied code, but I could not really figure out a way to unify them, due to the json loading needing the specialized type knowledge.

kristofbolyai (commented):
We need to add back the "reloadData" hook we used to have (that is called by /wynntils reload). After that, this should work just like it used to.

PR #2574, opened 2024-06-07
chore: Update the failed to connect messages [skip ci]

ShadowCat117:
image


kristofbolyai (commented):
I think we might as well do "Click __here__(underline) to try to reconnect".

ShadowCat117:
image

kristofbolyai (approved):
Thanks!

magicus:
I made a review comment but it was apparently not posted due to spotty Internet connection. :( If we're updating this message, I suggest we remove the "Welps!". Is that even a real word? It makes the error message sound not very serious, and do not fit with the rest of the style of our error messages, which are plain and to the point.

ShadowCat117:
I suppose it can be removed, didn't really think that much about it. I won't make a pr just for that change, I'm sure something will be done soon we can bundle it with

PR #2577, opened 2024-06-08
fix: Add back provider reload hooks

kristofbolyai:
Extends https://github.com/Wynntils/Artemis/pull/2573 by adding back reload hooks.


magicus:
Heh. I started working on this yesterday but never got very far.

But I added a `reloadData` to `MapDataProvider`, and not just built-in providers. I think that if we reload data, we will likely reload user-provided jsons as well, no?

kristofbolyai:
Hm, yes, that should work like that. If you want to, take over this PR, if not, I'll fix it up after the Hades one.

magicus:
Nah, you can continue, I did not get much further than thinking that reloadData needs to be at the provider top level. :)

kristofbolyai:
Okay, I've moved the method to the interface. I don't have much else to do, at least, now.

common/src/main/java/com/wynntils/services/mapdata/providers/MapDataProvider.java
```diff
@@ -19,4 +19,8 @@ public interface MapDataProvider {
     Stream<MapIcon> getIcons();
 
     void onChange(Consumer<MapDataProvidedType> callback);
+
+    default void reloadData() {
```
magicus: I think we should force implementors to think about reloading, i.e. it should not have a default implementation.

kristofbolyai: Okay.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
```diff
@@ -44,6 +44,9 @@ public Stream<MapCategory> getCategories() {
         return PROVIDED_CATEGORIES.stream();
     }
 
+    @Override
+    public void reloadData() {}
```
magicus: Now there is a bit of conflict between the inflight PRs, but maybe we should re-query Hades for this?

common/src/main/java/com/wynntils/services/mapdata/providers/json/JsonProvider.java
```diff
@@ -161,6 +161,11 @@ public void onChange(Consumer<MapDataProvidedType> callback) {
         // reload of the file, so we do not need to register callbacks.
     }
 
+    @Override
+    public void reloadData() {
+        // To be implemented if needed
```
magicus: I think we should implement it, if /reload is supposed to do anything. I guess it means we need to let the json provider somehow have a hook back to how it was loaded/created in the first place.

I am also fine with that you do not fix it right now, but then please turn this into a `FIXME`.

kristofbolyai: I did not do this because I was not sure how to. I'll definitely fix it when it comes to that.

common/src/main/java/com/wynntils/services/mapdata/providers/builtin/CharacterProvider.java
kristofbolyai: Yes, I did it.

magicus (approved):

PR #2578, opened 2024-06-08
feat: Render MapLocations natively, prepare for other MapFeatures

kristofbolyai:
This PR needs https://github.com/Wynntils/Artemis/pull/2570 (and it's follow ups) to be merged to be able to integrate mouse clicks correctly. For this reason, this PR was repurposed to only contain ports to mapfeature rendering, making it mergeable, as is. This PR needs to be merged so additional PRs, like MapArea and MapPath implementations don't depend on it in a chain.


common/src/main/java/com/wynntils/services/mapdata/type/MapArea.java
```diff
@@ -5,11 +5,17 @@
...
 import java.util.List;
 
 public interface MapArea extends MapFeature {
     // The area is described by a polygon. This list is the sequence of
     // vertices of that polygon, ordered in a counterclockwise orientation.
     // The last segment of the polygon connects from the last vertice to the first.
     List<Location> getPolygonArea();
+
+    @Override
+    default boolean isVisible(BoundingBox boundingBox) {
```
magicus: Just a note, this is technically not correct. An area could have all its points outside the bounding box, but it is still visibile inside the bounding box. In fact, this problem might be tricky to solve correctly. Or at least, I'm not a good enough game developer to know the answer straight ahead. Might be better to leave this unimplemented than pushing an incorrect implementation.

common/src/main/java/com/wynntils/services/mapdata/type/MapPath.java
```diff
@@ -5,9 +5,15 @@
...
+import com.wynntils.utils.type.BoundingBox;
 import java.util.List;
 
 public interface MapPath extends MapFeature {
     // The path is described by a sequence of locations
     List<Location> getPath();
+
+    @Override
+    default boolean isVisible(BoundingBox boundingBox) {
+        return getPath().stream().anyMatch(location -> boundingBox.contains(location.x(), location.z()));
```
magicus: Same here, the path might stretch across the bounding box (from e.g. down of it to up of it).

kristofbolyai:
Okay, you got me. I think you are right, in fact, I should not add any implementations for MapArea/MapPath, until we have at least one feature that uses it. I'll focus on MapLocations here.

magicus:
I think that sounds like a good plan. I must admit that I have not spend half as much time thinking about areas and paths as I have done for locations, so we might run into issues when we try to add those.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -178,24 +182,24 @@ public void render(PoseStack poseStack, MultiBufferSource bufferSource, float pa
                     textureZ,
                     width * extraFactor,
                     height * extraFactor,
-                    zoomRenderScale);
+                    1f / zoomRenderScale);
```
kristofbolyai, line 185: Turns out this was ported wrongly when we added the new map zooming. For some reason, noone noticed.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -242,54 +246,47 @@ private void renderPois(
             cosRotationRadians = (float) -StrictMath.cos(rotationRadians);
         } else {
             sinRotationRadians = 0f;
-            cosRotationRadians = 0f;
+            cosRotationRadians = 1f;
```
kristofbolyai, line 249: This is the "sane" default value, these rotations applied to the render coordinates produce the same output as the original input.

common/src/main/java/com/wynntils/screens/maps/CustomSeaskipperScreen.java
kristofbolyai, line 1: This class had a bunch of divisions that should have been floats instead..

kristofbolyai:
Phew, the merge conflicts. This is why I want this merged, rather than having to roll it for days.

@magicus This should be the next PR to be merged (after this PR, https://github.com/Wynntils/Artemis/pull/2570, just so you know). Please review it the next time you have time to spend on Wynntils. This is much less mapdata changes, and is just general porting + housekeeping, so I would assume not many review rounds will be needed.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
```diff
@@ -46,7 +46,7 @@ public class MainMapFeature extends Feature {
     public final Config<Float> playerPointerScale = new Config<>(1.5f);
 
     @Persisted
-    public final Config<Float> poiScale = new Config<>(1f);
```
magicus, line 49: Do we need an upfixer?

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
magicus: What was the effect of this bug?

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
kristofbolyai: I'd like to think that an explicit "reset" won't hurt with the new changes. But probably I can add it too..

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
```diff
@@ -242,54 +246,47 @@ private void renderPois(
...
         }
 
         float currentZoom = 1f / zoomRenderScale;
 
         // Get all MapData features as Pois
-        Stream<? extends Poi> poisToRender = Services.MapData.getFeaturesAsPois();
+        Stream<Pair<MapFeature, ResolvedMapAttributes>> mapFeatures = Services.MapData.getFeatures()
```
magicus, line 255: I've been kind of toying with the idea if we should have a ResolvedMapFeature, which basically would correspond to this pair. I think I got stuck when trying to figure out how this would result in a new hierarchy of ResolvedMapLocation etc, that is distinct from the MapFeature hierarchy...

But maybe we should pursue that idea a bit further.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
kristofbolyai: The zoom levels were off, in a subtle way. It had given me trouble, as I've unified feature location calculations, including rotation, scaling, and zoom levels. I've had code that worked perfectly on the maps, but not on the minimap. I suspected something was off in this port, but I found the different values (1/scale vs scale) for the main and mini maps. From testing and reading the old code, I'm confident that this is correct.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -76,8 +83,11 @@ public abstract class AbstractMapScreen extends WynntilsScreen {
     // Zooming updates zoomLevel, but we also cache zoomRenderScale for rendering
     protected float zoomLevel = MapRenderer.DEFAULT_ZOOM_LEVEL;
     protected float zoomRenderScale = MapRenderer.getZoomRenderScaleFromLevel(zoomLevel);
+    protected BoundingBox mapBoundingBox = BoundingBox.EMPTY;
 
+    // TODO: This is not used anymore. It's only here to make the code compile.
```
magicus, line 88: ... and you can't remove the code that still references it?

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
kristofbolyai: Maybe, although I feel like it might get problematic with generics, and not result in much cleaner code.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
kristofbolyai: All of the "mouseClicked" methods use it. You know, that big mess of "keybinds" :). I'll be working on unifying that, after the marker options are added, and we figured out feature/category overrides.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
```diff
@@ -130,4 +184,75 @@ public static void renderMapFeature(
...
+            float renderZ = screenCenter.y() + dZ;
+
+            int labelHeight = (int) (FontRenderer.getInstance().getFont().lineHeight * featureRenderScale * TEXT_SCALE);
+
+            int yOffset = 0;
+
+            float iconAlpha = Services.MapData.calculateVisibility(attributes.iconVisibility(), zoomLevel);
+            Optional<MapIcon> icon = Services.MapData.getIcon(attributes.iconId());
+            boolean drawIcon = iconAlpha > MINIMUM_RENDER_ALPHA;
+            if (icon.isPresent() && drawIcon) {
```
magicus, line 217: Feels like we are doing a lot over overlap here with the rendering. I wonder if we can express this logic only once... hmm....

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -857,13 +857,13 @@
   "feature.wynntils.mainMap.description": "Adds a map to view the world.",
   "feature.wynntils.mainMap.holdGuildMapOpen.description": "Should Control need to be held to open guild territories on the map. If this is disabled, control will act as a toggle.",
   "feature.wynntils.mainMap.holdGuildMapOpen.name": "Hold Control to Show Guild Territories",
+  "feature.wynntils.mainMap.mapFeatureScale.description": "How big should map map features be?",
+  "feature.wynntils.mainMap.mapFeatureScale.name": "Map Point of Interest Scale",
```
magicus: I think we should drop the "Point of Interest" name and just go with "map feature" (or just "feature", if the context is clear).

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -857,13 +857,13 @@
   "feature.wynntils.mainMap.description": "Adds a map to view the world.",
   "feature.wynntils.mainMap.holdGuildMapOpen.description": "Should Control need to be held to open guild territories on the map. If this is disabled, control will act as a toggle.",
   "feature.wynntils.mainMap.holdGuildMapOpen.name": "Hold Control to Show Guild Territories",
+  "feature.wynntils.mainMap.mapFeatureScale.description": "How big should map map features be?",
```
```suggestion
  "feature.wynntils.mainMap.mapFeatureScale.description": "How big should main map features be?",
```

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: Yes, my mistake, I wanted to do this originally.

magicus (approved):
Seems okay, just rename "point of interest" to "feature" in the i18n text, and it's good to merge.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
kristofbolyai: It's really hard to create such abstractions sadly. The other map features will be simpler to hover check.

common/src/main/java/com/wynntils/overlays/minimap/MinimapOverlay.java
magicus: Let's keep with adhoc pairing like this for now.

common/src/main/java/com/wynntils/features/map/MainMapFeature.java
magicus: Ok, fair enough, old values might not really apply anymore.

PR #2579, opened 2024-06-08
feat: Introduce TerritoryArea and MapArea rendering

kristofbolyai:
_This PR targets https://github.com/Wynntils/Artemis/pull/2578 as it implements MapArea rendering, etc._

# Working Demo
image

# Comparison
*Left is old POI rendering, right is mapdata.*
Note that the guild map is not complete yet. This PR is marked as ready as it's purpose is to tackle map area rendering.
image
image


common/src/main/java/com/wynntils/services/mapdata/type/MapArea.java
```diff
@@ -5,17 +5,21 @@
...
 
 import com.wynntils.utils.mc.type.Location;
+import com.wynntils.utils.type.BoundingPolygon;
 import com.wynntils.utils.type.BoundingShape;
 import java.util.List;
 
 public interface MapArea extends MapFeature {
     // The area is described by a polygon. This list is the sequence of
-    // vertices of that polygon, ordered in a counterclockwise orientation.
+    // vertices of that convex polygon, ordered in a counterclockwise orientation.
```
magicus, line 15: Is it necessary to limit this to convex polygons? I can think of many reasons for having non-convex polygons, e.g. if trying to create areas that cover geographical features on the Wynncraft map.

kristofbolyai: This is very elaborate, the reasoning is that some operations are much more reasonable if they are only done on convex polygons, with these operations being rendering, and different kinds of clipping checks, both of which are vital performance wise.

magicus: Okay. I'm honestly not very good at that kind of algorithms (I remember that I did have exams on stuff like convex hulls etc in uni, but I've forgotten all details...).

Maybe we can rephrase the text (which basically is our specification) to something like:
```
    // The area is described by a polygon. This list is the sequence of
    // vertices of that polygon, ordered in a counterclockwise orientation.
    // Currently, only convex polygons are supported.
```

to at least indicate that we could be open for something else, if the need arises.

kristofbolyai:
So from what I understand, this is the next "step", right, @magicus?

magicus:
It's one of the next steps, yes. :) If you can spend time on it, that'd be great!

kristofbolyai:
Ah, yes! I've now booted up the game, and now I know what this PR is missing! We need a new, MapArea-only attribute, for borderWidth and borderColor. I assume we don't need a hasBorder if we can set a 0 width border..

kristofbolyai:
Any ideas for map area attributes? Next time I get to work on this, I'll add some MapArea attributes, and remove all that are too specific to MapLocation. I suppose that needs to be a separate PR, as many MapAttribute attributes should actually only be MapLocationAttributes.

magicus:
Yes, we can probably restrict some of the current attributes to be just location attributes.

I started thinking that maybe, now that we have a story for handling sub-attributes (marker and visibility), maybe we should group some of the existing attributes as well, e.g. label (text and color), icon (id and color), and now also border (color and width). It increases the complexity of the code a bit, but it also makes the resulting structure more intuitive.

kristofbolyai:
This should be ready now, please review it fully and then we can discuss anything that comes up.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
```diff
@@ -131,6 +132,9 @@ protected void doInit() {
         mapHeight = renderHeight - renderedBorderYOffset * 2f;
         centerX = renderX + renderedBorderXOffset + mapWidth / 2f;
         centerZ = renderY + renderedBorderYOffset + mapHeight / 2f;
+
```
magicus, line 135: Is this related to the territory or maparea changes? Does not look like it.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
```diff
@@ -35,17 +35,20 @@ public static ResolvedMapAttributes resolve(MapFeature feature) {
...
+                resolver.getAttribute(MapAttributes::getLabel),
                 resolver.getResolvedMapVisibility(MapAttributes::getLabelVisibility),
                 resolver.getAttribute(MapAttributes::getLabelColor),
                 resolver.getAttribute(MapAttributes::getLabelShadow),
+                resolver.getAttribute(MapAttributes::getIconId),
                 resolver.getResolvedMapVisibility(MapAttributes::getIconVisibility),
                 resolver.getAttribute(MapAttributes::getIconColor),
                 resolver.getAttribute(MapAttributes::getIconDecoration),
-                resolver.getResolvedMarkerOptions(MapAttributes::getMarkerOptions));
+                resolver.getResolvedMarkerOptions(MapAttributes::getMarkerOptions),
```
magicus, line 48: Did we miss this in the other PR where you added the area attributes?

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
```diff
@@ -185,6 +199,59 @@ public static void renderMapLocation(
...
+                attributes.fillColor(),
+                attributes.borderColor(),
+                attributes.borderWidth(),
+                screenVertices,
+                0);
+
+        BoundingPolygon boundingPolygon = BoundingPolygon.fromVertices(screenVertices);
+        Vector2f centroid = boundingPolygon.centroid();
+
+        int labelWidth = (int)
```
magicus, line 234: I get the feeling this code should be possible to share with drawing labels for locations. Or is it slightly different enough that it is not worth the effort to try and unify them?

common/src/main/java/com/wynntils/services/mapdata/type/MapArea.java
magicus: I still think this applies.

common/src/main/java/com/wynntils/utils/type/BoundingPolygon.java
magicus, line 1: I'm just gonna assume that you've done this correct... I guess we'll notice if this fails. Or, if you are feeling very enthusiastic, you could consider adding a few unit tests for the geometric bounding functionality.

common/src/main/java/com/wynntils/screens/maps/AbstractMapScreen.java
kristofbolyai: In fact it it, we use this bounding shape to check collision with other map features, and this was improperly inited among maps.

common/src/main/java/com/wynntils/utils/type/BoundingShape.java
```diff
@@ -1,44 +1,40 @@
 /*
- * Copyright © Wynntils 2023.
+ * Copyright © Wynntils 2023-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.utils.type;
 
 public interface BoundingShape {
```
magicus: This would btw be a perfect candidate for the new `sealed` keyword:

```
public sealed interface BoundingShape permits BoundingBox, BoundingCircle, BoundingPolygon {
```

 It's saying that the only possible BoundingShapes are Boxes, Circles and Polygons. This can help the compiler generate more efficient code in some cases, and it clarifies the logic. Not saying you must, just pointing out that this is one of the few examples I've ever seen where it would make sense.

magicus (approved):
Overall this looks good, I think. I had a few comments here and there, only.

Also, when do we start removing the old framework? Should we do that partially as we fix things, or clear out all in the end? (I'm not seeing much code removal here so I'm guessing there will be some dead code in the old territory handing after this PR)

kristofbolyai:
I usually try to check, but I can't remove much before I adapt the guild screen logic first.

common/src/main/java/com/wynntils/services/mapdata/MapAttributesResolver.java
kristofbolyai: Yes.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
kristofbolyai: I only like to "dry out code" if it is repeated 3 times. This is only the second case :).

common/src/main/java/com/wynntils/utils/type/BoundingPolygon.java
kristofbolyai: That would require adding it for all the other shapes. Let's trust this code for now, and we can make some tests when it fails for the first time.

common/src/main/java/com/wynntils/services/mapdata/type/MapArea.java
kristofbolyai: I do not think that we should be open for something else. Concave polygons can be described as multiple convex polygons, and the complexity drastically changes when you introduce concave polygons into the mix.

common/src/main/java/com/wynntils/services/mapdata/MapFeatureRenderer.java
magicus: Kind of depends on how long the repeated piece of code is. But okay...

PR #2584, opened 2024-06-10
fix: Correct war horn configs [skip ci]

ShadowCat117:
TowerEffectOverlayFeature is already implemented correctly

image

### DIFF
--- a/common/src/main/java/com/wynntils/features/embellishments/WarHornFeature.java
+++ b/common/src/main/java/com/wynntils/features/embellishments/WarHornFeature.java
@@ -28,6 +28,6 @@ public class WarHornFeature extends Feature {
 
     @SubscribeEvent
     public void onWarQueued(GuildWarQueuedEvent event) {
-        McUtils.playSoundAmbient(WAR_HORN_SOUND, soundVolume.get(), soundPitch.get());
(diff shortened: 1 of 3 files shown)


kristofbolyai (commented):
Uh oh. Let me think if there is a better solution, this seems like something that'll take equally as long to notice, once it's fixed by MC.

magicus:
Is this an error in the mappings, naming the vanilla arguments incorrectly?

ShadowCat117:
> Is this an error in the mappings, naming the vanilla arguments incorrectly?

Yeah, so it's putting the "volume" argument as the pitch and the "pitch" argument as the volume.

If they just rename the arguments but don't actually swap the order they need to be passed in, then it shouldn't be a problem but obviously we won't know how they'll fix it

magicus:
Are mojang using it "backwards" in their own code as well? 

I think we should keep our order internally, but be prepared to swap it when we call into vanilla code if they ever change that.

ShadowCat117:
`SimpleSoundInstance.forLocalAmbience(SoundEvents.PORTAL_TRIGGER, this.random.nextFloat() * 0.4F + 0.8F, 0.25F)`

Is used for the sound played when in a nether portal, so "volume" is the random one and "pitch" is the constant 0.25F. But in actuality the volume is the constant and the pitch is the random one

kristofbolyai (approved):

PR #2585, opened 2024-06-11
feat: Add option to disable Discord rich presence in streamer mode

ShadowCat117:
One place I forgot we can still leak the world. I was going to just hide the world but I think hiding the whole presence is more useful as most people use streamer mode to hide that they are online vs just hiding their current world.

### DIFF
--- a/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
+++ b/common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
@@ -96,9 +96,9 @@ public class DiscordRichPresenceFeature extends Feature {
 
     @SubscribeEvent
     public void onStreamToggle(StreamModeEvent e) {
-        if (e.isEnabled()) {
+        if (disableInStream.get() && e.isEnabled()) {
             onDisable();
-        } else {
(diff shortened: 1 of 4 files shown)


magicus (approved):

common/src/main/java/com/wynntils/features/DiscordRichPresenceFeature.java
```diff
@@ -90,6 +94,15 @@ public void onWorldChange(WorldStateEvent event) {
         }
     }
 
+    @SubscribeEvent
+    public void onStreamToggle(StreamModeEvent e) {
+        if (disableInStream.get() && e.isEnabled()) {
+            onDisable();
+        } else if (!e.isEnabled()) {
+            onEnable();
+        }
```
kristofbolyai: Can you extract what onEnable/onDIsable does into enable/disable methods, and call that here and in the onX methods? It feels very wrong to call such a feature "event" method

kristofbolyai: Also, you may need to add hooks for "config change" here.

kristofbolyai (approved):
Thanks!

PR #2588, opened 2024-06-13
fix: Fix discoveries not being queried when only showing undiscovered [skip ci]

ShadowCat117:


kristofbolyai:
Does this fix anything? Afaik undiscovered stuff are resolved from the API (hence my design)?

ShadowCat117:
If you have the filter set to only show undiscovered, it will never show anything as `reloadDiscoveries` will just get all false so it'll never query anything

ShadowCat117:
Current release, pressing reload leaves it empty


With this fix, pressing reload shows the undiscovered

kristofbolyai:
Oh, hm, I guess this doesn't work as I thought it does..

kristofbolyai:
Probably it was changed when the content book was updated..

ShadowCat117:
Yeah content book contains everything for discoveries, you just can't track secret ones so we get the location info from the wiki

PR #2590, opened 2024-06-14
feat: Item Filter screen smooth scrolling [skip ci]

ShadowCat117:
Also fixes a crashing issue and just improves the sort widget slightly.

Same smooth scrolling implementation as on other recent screens, only a few screens left to get it :)


kristofbolyai (approved):
Code looks good, feels good in-game.

PR #2594, opened 2024-06-14
fix: Properly detect class joining

### LINKED ISSUE #2593: Entering the class edit screen causes /guild xp to be run multiple times upon join
**Version:** 1.2.22

### Expected behaviour

Editing a class would have no impact and the /guild xp command would only be run once upon joining.

### Actual behaviour

The /guild xp command will be run for each time you open the class edit menu once you join.

### Steps

1. Open the class selection screen
2. Press the edit button
3. Exit the class edit menu
4. Repeat steps 2-3 *n* times
5. Join with a class
6. /guild xp will be run *n* times

### Visual steps

https://github.com/Wynntils/Artemis/assets/71361356/e33dd560-dc34-4fc7-b65b-b92b674487ff

kristofbolyai:
Previously, we detected "any" clicks that was not air, so the edit buttons could "signal" a class update..

Fixes #2593

### DIFF
--- a/common/src/main/java/com/wynntils/models/character/CharacterModel.java
+++ b/common/src/main/java/com/wynntils/models/character/CharacterModel.java
@@ -23,6 +23,7 @@ import com.wynntils.models.character.event.CharacterDeathEvent;
 import com.wynntils.models.character.event.CharacterUpdateEvent;
 import com.wynntils.models.character.type.ClassType;
 import com.wynntils.models.containers.ContainerModel;
+import com.wynntils.models.items.items.gui.CharacterItem;
 import com.wynntils.models.worlds.event.WorldStateEvent;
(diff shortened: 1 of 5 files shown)


ShadowCat117:
I was going to suggest maybe we should start parsing gamemodes from this but it seems like we need to 
image

magicus (approved):
I think this looks good but did not scrutinize it.

kristofbolyai:
Okay @ShadowCat117, I made the regex support game modes, but I did not add any functionality to parse/display it.

ShadowCat117:
That's fine, I have this screen on my todo list for a redesign anyway so unless someone gets to it first I'll display them then :)

ShadowCat117 (approved):

PR #2595, opened 2024-06-14
feat: Add the ability to have different attributes for each map feature type

### LINKED ISSUE #2582: [MapData] - Different attributes for different MapFeatures

kristofbolyai:
Proposed fix for #2582 (fixes #2582)

I've explicitly wanted to avoid using generics in the implementation. Using generics in this case would "litter" the code everywhere, whereas, now, we only have a tiny amount of casting done in consumer classes, and most of the heavily lifting is done in MapAttributesResolver.

It's also on purpose that the new attribute types do not have any custom values, rather, they only have defaults defined. I think it's sane to implement these custom attributes the first time we use them (so, in the territory area PR for area, and the first map path PR).


magicus:
I rebased this on the updated mapdata. A merge would have just made this one-commit PR harder to understand.

PR #2596, opened 2024-06-15
fix: Fix emitting ProfessionXpGainEvent twice for woodcutting [skip ci]

### LINKED ISSUE #2586: ProfessionXpGainedEvent can get emitted twice for woodcutting
https://github.com/Wynntils/Artemis/assets/59118042/b242d028-0e53-4369-83b7-26fb51b29f7f

Trees have multiple locations where the label can spawn, moving depending on where the player stands. This can trigger the `ProfessionXpGainedEvent` again.

cmkohnen:
fix #2586

### DIFF
--- a/common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
+++ b/common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
@@ -101,7 +101,7 @@ public class ProfessionModel extends Model {
     }
 
     @SubscribeEvent
-    public void onLabelSpawn(EntityLabelChangedEvent event) throws TimedValue.ValueExpiredException {
+    public void onLabelSpawn(EntityLabelChangedEvent event) {
         StyledText label = event.getName();
 
         Matcher professionNodeExperienceMatcher = label.getMatcher(PROFESSION_NODE_EXPERIENCE_PATTERN);
@@ -118,9 +118,9 @@ public class ProfessionModel extends Model {
             ProfessionType profession = ProfessionType.fromString(professionNodeExperienceMatcher.group("name"));
 
             // Woodcutting labels can move during "display", so position based checks don't always work
-            if (profession == ProfessionType.WOODCUTTING
-                    && !lastProfessionLabel.isExpired()
-                    && label.equals(lastProfessionLabel.get())) return;
+            if (profession == ProfessionType.WOODCUTTING && lastProfessionLabel.matches(label)) {
+                return;
+            }
 
             lastProfessionLabel.set(label);
             gatheredNodes.put(entityPosition);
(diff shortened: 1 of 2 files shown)


kristofbolyai (commented):
Do you think it's viable to get the same label, from different nodes? If you, perhaps some timed mechanism should be used.

ShadowCat117:
> Do you think it's viable to get the same label, from different nodes? If you, perhaps some timed mechanism should be used.

At lower levels it's probably more likely to get the same xp amount. Once you're earning thousands per node the range is probably too big for it to be a common occurrence

cmkohnen:
> Do you think it's viable to get the same label, from different nodes? If you, perhaps some timed mechanism should be used.

It is definitely possible if yet very improbable.
However just to be on the safe side, I've made it time based.

For this time based implementation I've introduced `TimedValue` a bit similar to `TimedSet`.

common/src/main/java/com/wynntils/utils/type/TimedValue.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.utils.type;
+
+import java.util.concurrent.TimeUnit;
+
+public class TimedValue<T> {
```
kristofbolyai, line 10: This class should have a bit of a different API, to make it easier/clearer to use.

This class should behave like a provider class, with the duration/unit settings being kept as a final value. The value itself should be changeable, and should have it's expiry reset when the value changes. The get method itself should behave like an optional, aka, it should throw, if the value is expired. That way, we the developer has to explicitly check for `isExpired`.

kristofbolyai: Perhaps it would also be beneficial to have a `reset` method, which clears the current value, and throws in get/return false in `isExpired`.

kristofbolyai:
Overall, I do like the idea here, I would have done the same thing :). The class itself needs a bit of refinement though, it's very weird to have the timer part reconstructed in-code.

common/src/main/java/com/wynntils/models/profession/ProfessionModel.java
```diff
@@ -110,21 +115,29 @@ public void onLabelSpawn(EntityLabelChangedEvent event) {
                 return;
             }
 
+            ProfessionType profession = ProfessionType.fromString(professionNodeExperienceMatcher.group("name"));
+
+            // Woodcutting labels can move during "display", so position based checks don't always work
+            if (profession == ProfessionType.WOODCUTTING
+                    && !lastProfessionLabel.isExpired()
+                    && label.equals(lastProfessionLabel.get())) return;
```
kristofbolyai: Add braces, the if is too long.

kristofbolyai (approved):

PR #2597, opened 2024-06-15
port: Port to 1.21, using NeoForge EventBus

kristofbolyai:
### port: Port to 1.21 NeoForge (buildscript, mixins)
This commit ports to 1.21 NeoForge. Notable changes are the fixed NeoForge/Fabric only mixins (I've separated old, require = 0 mixins to Fabric, so it's easier to see what works where). I've also added improvements to our buildscript for both Fabric/NeoForge, to aid version changes.

This commit does not compile, as NeoForge's new EventBus needs some changes to the code base.

### port: Port to NeoForge EventBus
This PR is mostly just import changes. A few rules were added to NeoForge's event bus, to make it easier and less error-prone to use.
Affected parts:
- `@Cancelable` annotation is now an interface.
- With the annotations removed, the event bus does not need pre-processing, so we no longer depend on EventBusTransformer.
- Generic events were removed. Only PackedEvent used it, it's a simple port.
- Abstract classes cannot register events with @SubscribeEvent. For overlays, I've introduced manager level ticking. For special overlays, I've added WynntilsMod#registerListener to specially register these events.
- Abstract events cannot be listened to. This resulted in a small amount of code duplication. Not enough to consider doing any refactors.


ShadowCat117:
> It's also likely that we are going to end up switching to NeoForge. Before making that decision, I'll wait for some mod loader statistics for 1.21.

I think we just go for NeoForge, we already tell people whenever they have Forge issues to swap to Fabric so if we are going to support 1 of the Forge versions we should just go for NeoForge, unless whenever people have issues with NeoForge we tell them to swap to Fabric in which case why even bother supporting a Forge version

Rafii2198:
Also, Forge Wynntils users are such minority, pretty much everybody is on Fabric, and most of that people on Forge are using it because "this is what I've been always using" and that's the only reason, the reason is not available mods or ease of use because that is what is better about Fabric, it's not Forge's domain, they just don't want to try something new.
And as for mods that are relevant for Wynncraft, i.e client sided mods then NeoForge is already doing better in that regard as many Fabric mods got ported to NeoForge, even Sodium and Iris have interest to supporting it natively.

And if people really want to use Forge then there is always Sinytra Connector.

common/src/main/resources/wynntils.accessWidener
```diff
@@ -50,7 +50,6 @@ accessible method net/minecraft/client/multiplayer/MultiPlayerGameMode startPred
 accessible method net/minecraft/client/renderer/RenderType create (Ljava/lang/String;Lcom/mojang/blaze3d/vertex/VertexFormat;Lcom/mojang/blaze3d/vertex/VertexFormat$Mode;ILnet/minecraft/client/renderer/RenderType$CompositeState;)Lnet/minecraft/client/renderer/RenderType$CompositeRenderType;
 accessible method net/minecraft/client/renderer/RenderType create (Ljava/lang/String;Lcom/mojang/blaze3d/vertex/VertexFormat;Lcom/mojang/blaze3d/vertex/VertexFormat$Mode;IZZLnet/minecraft/client/renderer/RenderType$CompositeState;)Lnet/minecraft/client/renderer/RenderType$CompositeRenderType;
 accessible method net/minecraft/client/renderer/blockentity/BeaconRenderer renderBeaconBeam (Lcom/mojang/blaze3d/vertex/PoseStack;Lnet/minecraft/client/renderer/MultiBufferSource;Lnet/minecraft/resources/ResourceLocation;FFJIIIFF)V
-accessible method net/minecraft/client/resources/server/DownloadedPackSource tryParseSha1Hash (Ljava/lang/String;)Lcom/google/common/hash/HashCode;
```
ShadowCat117: Is this not required? I had to add it back to be able to run

kristofbolyai: It is.. No idea when it got removed, probably rebase nuked it..

neoforge/src/main/resources/META-INF/neoforge.mods.toml
```diff
@@ -0,0 +1,29 @@
...
+[[dependencies.wynntils]]
+modId = "neoforge"
+mandatory = true
+versionRange = "${neoforge_version_range}"
+
+[[mixins]]
+config = "wynntils.mixins.json"
+
+[[mixins]]
+config = "wynntils.mixins.neoforge.json"
```
magicus, line 29: Sorry, late review. This file is missing a trailing newline.
(Why don't we check for that? Maybe we ignore toml files in Spotless?)

PR #2599, opened 2024-06-16
feat: Lootrun Mission Tracker Overlay [skip ci]

wiji1:
The first couple of commits aims to add more lootrun statistics to the LootrunFinishedEvent. These will be used in future PRs that I have planned.


common/src/main/java/com/wynntils/functions/LootrunFunctions.java
```diff
@@ -121,6 +122,24 @@ public FunctionArguments.Builder getArgumentsBuilder() {
...
+            MissionType mission = Models.Lootrun.getMission(missionIndex);
+            if (mission == null) return "§cNone";
+
+            return mission.getColoredName();
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.RequiredArgumentBuilder(
+                    List.of(new FunctionArguments.Argument<>("index", Integer.class, -1)));
```
kristofbolyai:
```suggestion
                    List.of(new FunctionArguments.Argument<>("index", Integer.class, 
null)));
```

Required arguments have null default values.

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
```diff
@@ -121,6 +122,24 @@ public FunctionArguments.Builder getArgumentsBuilder() {
         }
     }
 
+    public static class LootrunMissionFunction extends Function<String> {
+        @Override
+        public String getValue(FunctionArguments arguments) {
+            int missionIndex = arguments.getArgument("index").getIntegerValue();
+
+            MissionType mission = Models.Lootrun.getMission(missionIndex);
+            if (mission == null) return "§cNone";
```
kristofbolyai: We don't do coloring for function outputs 99% of the time, it just makes them more annoying to use, and less flexible.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -107,6 +111,11 @@ public class LootrunModel extends Model {
     //                        ÀÀÀ§7Challenges Completed: §f7
 
     private static final Pattern LOOTRUN_FAILED_PATTERN = Pattern.compile("[À\\s]*§c§lLootrun Failed!");
+    private static final Pattern CHALLENGE_FAILED_PATTERN = Pattern.compile("À.*?§c§lChallenge Failed!");
+
+    private static final Pattern MISSION_COMPLETED_PATTERN = Pattern.compile("[À\\s]*§b§lMission Completed");
+    private static final Pattern ADVANCED_MISSION_PATTERN = Pattern.compile("À.*?§b§l");
+    private static final Pattern MISSION_PATTERN = Pattern.compile(".*?§.");
```
kristofbolyai: I would have really liked tests for this in TestRegex, for two reasons. This is really not a strict regex, I fear it will match too much. If I can see the missions texts, I can either help with a stricter regex, or I can see and accept it, as is.

common/src/main/java/com/wynntils/models/lootrun/type/MissionType.java
```diff
@@ -0,0 +1,51 @@
...
+    CLEANSING_RITUAL("Cleansing Ritual", '5'),
+    EQUILIBRIUM("Equilibrium", '5'),
+    INNER_PEACE("Inner Peace", '5'),
+    BACKUP_BEAT("Backup Beat", 'a'),
+    STASIS("Stasis", 'a'),
+    GAMBLING_BEAST("Gambling Beast", 'c'),
+    REDEMPTION("Redemption", 'c'),
+    ULTIMATE_SACRIFICE("Ultimate Sacrifice", 'c'),
+    FAILED("Failed", '4'),
+    ;
```
kristofbolyai:
```suggestion
    FAILED("Failed", '4');
```

common/src/main/java/com/wynntils/models/lootrun/type/MissionType.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final String name;
+    private final char color;
```
kristofbolyai: This could be a ChatFormatting, not a char.

common/src/main/java/com/wynntils/models/lootrun/type/MissionType.java
kristofbolyai: I also don't like that "Failed" is a mission type. It's not. I know it can happen, but I think it should be handled specially. This goes for "unknown" as well, since this class hard-codes all missions, it can easily happen that Wynn adds new ones, then, the mod crashes, errors, etc.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
kristofbolyai: Okay, I see.. These are not real patterns, but a "builder" for the mission types. It might be better not to call them "patterns" then, and keep them as a string, or at the very least, comment that they are not used as declared here.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -158,6 +167,12 @@ public class LootrunModel extends Model {
 
     private Map<BeaconColor, Integer> selectedBeacons = new TreeMap<>();
 
+    @Persisted
+    private final Storage<Map<String, List<MissionType>>> missionStorage = new Storage<>(new TreeMap<>());
+
+    private final Pattern[] missionPatterns = new Pattern[2];
+    private final StyledText[] recentChatMessages = new StyledText[5];
```
kristofbolyai: I am not happy about either of these, but especially about `recentChatMessages`. Let's keep this comment here, and we can get back to it, when most of the other requests are done, and the code is finalising.

kristofbolyai (commented):
The first two commits look good, but the mission tracking needs some work. Don't get afraid of the comments, it's there to help. If you have any questions, let me know.

common/src/main/java/com/wynntils/models/lootrun/type/MissionType.java
kristofbolyai: To be clear, I am not a 100% sure what would be the best here. Maybe, having some state class for mission state with the states, "not started", "failed", "started", "finished", with the mission either being a mission type, or an "unknown mission".

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
wiji1: How would you suggest I color the Mission names on the Overlay?

common/src/main/java/com/wynntils/models/lootrun/type/MissionType.java
wiji1: In my recent commit I have aimed to address this. Please let me know if you want me to do this differently.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
wiji1: If it wasn't clear from the code, the recent chat messages are required to identify that a mission has been completed instead of starting. Please let me know if you can think of a better way to achieve this. I do agree that the missionPatterns array is a little sketchy and could probably be done in a better way.

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
kristofbolyai: You can add a secondary argument for colored/not.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
kristofbolyai: You can do exactly what I did above your code, to get rid of `recentChatMessages`.  If `LOOTRUN_COMPLETED_PATTERN/LOOTRUN_FAILED_PATTERN` matches, then a flag is set, that expect the next message to contain fail/complete data. You can do the same thing here, set an "expectMission"/"expectMissionComplete" flag, and the next time you parse a mission from chat, you know what it is.

Secondly, the added code around `onChatMessage` needs changes. At first, I had no idea why it even worked. I see now. `StyledText#toString` does not return a string representation of the text. If you check the method, it's clear that it's a debug-style text. You have multiple `getString` methods available for that. Then, you can replace all the `matcher.find()` calls with `"matcher.matches()`.

wiji1: I have taken your advice and replaced the recentChatMessages system with one with a flag that expects mission completion. Would you still like me to change the missionPatterns array? It's currently in an array to avoid adding too many variables at the top of the file, but since I don't anticipate any more entries being added in the future, I could change it to a simple two-variable solution.

common/src/main/java/com/wynntils/functions/LootrunFunctions.java
```diff
@@ -121,6 +121,23 @@ public FunctionArguments.Builder getArgumentsBuilder() {
...
+            boolean colored = arguments.getArgument("colored").getBooleanValue();
+
+            return Models.Lootrun.getMissionStatus(missionIndex, colored);
+        }
+
+        @Override
+        public FunctionArguments.Builder getArgumentsBuilder() {
+            return new FunctionArguments.RequiredArgumentBuilder(List.of(
+                    new FunctionArguments.Argument<>("index", Integer.class, null),
+                    new FunctionArguments.Argument<>("colored", Boolean.class, false)));
```
kristofbolyai:
```suggestion
                    new FunctionArguments.Argument<>("colored", Boolean.class, null)));
```

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -158,6 +170,13 @@ public class LootrunModel extends Model {
 
     private Map<BeaconColor, Integer> selectedBeacons = new TreeMap<>();
 
+    @Persisted
+    private final Storage<Map<String, List<String>>> missionStorage = new Storage<>(new TreeMap<>());
```
kristofbolyai: Move this closer to the storages, so switch it's place with selectedBeacons

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -246,6 +266,9 @@ public void onCharacterChange(CharacterUpdateEvent event) {
         selectedBeacons = selectedBeaconsStorage.get().get(id);
 
         selectedBeaconsStorage.touched();
+
+        missionStorage.get().putIfAbsent(id, new LinkedList<>());
```
kristofbolyai:
```suggestion
        missionStorage.get().putIfAbsent(id, new ArrayList<>());
```

Storages use `ArrayList` as the implementation for lists. Do you need LinkedLists specifically?

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -509,11 +574,37 @@ public void addToRedBeaconTaskCount(int changeAmount) {
...
         redBeaconTaskCountStorage.touched();
     }
 
+    private void resetMissions() {
+        List<String> missions = missionStorage.get().get(Models.Character.getId());
+        if (missions == null) return;
+
+        missions.clear();
+        missionStorage.get().put(Models.Character.getId(), missions);
+        missionStorage.touched();
```
kristofbolyai:
```suggestion
        List<String> missions = missionStorage.get().getOrDefault(Models.Character.getId(), new ArrayList<>());
        missions.clear();
        missionStorage.touched();
```

The put is not needed, it's already a reference from the storage itself. The other change is just a nit.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -107,6 +111,14 @@ public class LootrunModel extends Model {
...
 
     private static final Pattern LOOTRUN_FAILED_PATTERN = Pattern.compile("[À\\s]*§c§lLootrun Failed!");
+    private static final Pattern CHALLENGE_FAILED_PATTERN = Pattern.compile("[À\\s]*§c§lChallenge Failed!");
+
+    private static final Pattern MISSION_COMPLETED_PATTERN = Pattern.compile("[À\\s]*§b§lMission Completed");
+
+    // These two prefixes are used to identify when a mission is completed,
+    // as this information is presented in multiple ways
+    private static final String ADVANCED_MISSION_PREFIX = ".*[À\\\\s]*§b§l";
+    private static final String MISSION_PREFIX = ".*?§.";
```
kristofbolyai: So can you explain a bit more about these patterns? I need lines for them, because these regexes are too broad. Why the ".*" in both cases? Is one of the patterns for completion, one for mission start? 

Also, you could create the regex in-line with something like `Pattern.compile("...(" + Arrays.stream(MissionType.values()).map(MissionType::getName).collect(Collectors.joining("|")) + ")");`

kristofbolyai:
I still have concerns about keeping missions as a string, but I can only suggest a change after I know what the regexes are for, why they are different, and if we are able to tell the difference between mission start and mission complete.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
wiji1: To be honest, I'm not too familiar with regex, so most of these patterns are taken from other parts of the project with some assistance from AI development tools. 

When a mission is completed, it can be presented in one of two ways:

-   It is completed during a challenge, presenting the "Mission Complete" message before displaying the mission name.
-   It is completed as a part of a challenge's finishing phase, where it is presented in a format without the "Mission Complete" text and where the mission name is always the Bolded Aqua color. 

Each of these regex patterns covers its corresponding case.

Currently, this feature does not track when missions are started, as I did not feel it was necessary due to the current mission already being tracked on the scoreboard along with its completion requirement.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
wiji1: The order of the missions needs to be retained in the storage. Since the values are already being stored in strings, I could  append prefixes to them to order them once fetched from storage, though it would be a rather messy solution.

kristofbolyai: ArrayLists already retain the order? We are not talking about sets here

wiji1: My apologies, I mixed up List index order with Map key index order.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
kristofbolyai: Okay. Can you send the chat messages from some logs then? Only then, I can help with the regex.

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -509,11 +574,34 @@ public void addToRedBeaconTaskCount(int changeAmount) {
...
+
+    private void addMission(String identifier) {
+        List<String> missions = missionStorage.get().getOrDefault(Models.Character.getId(), new ArrayList<>());
+
+        for (String mission : missions) {
+            if (mission.equals(identifier)) {
+                WynntilsMod.info("Mission already exists.");
+                return;
+            }
+        }
```
kristofbolyai:
```suggestion
    if (missions.contains(identifier)) {
            WynntilsMod.warn("Mission already completed: " + identifier + ", mission list: " + String.join(", ", missions));
            return;
    }
```

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
wiji1: First case:
`[CHAT]                          ÀÀ&b&lMission Completed
18:18:40.301
wynntils
Render thread
[CHAT]                     ÀÀÀ&7You have finished 1 mission!
18:18:40.302
wynntils
Render thread
[CHAT]                                   ÀÀÀ&eStasis
18:18:40.302
wynntils
Render thread
[CHAT]                    À&7While choosing a beacon, your
18:18:40.302
wynntils
Render thread
[CHAT]                 À&7timer does not decrease (Max 5m)
`

Second case:
`[CHAT]            À&7Parsimonious&r                      ÀÀ&b&lMaterialism
17:43:09.708
wynntils
Render thread
[CHAT]                    À                   ÀÀÀ&7All Challenges now additionally
17:43:09.708
wynntils
Render thread
[CHAT]                    À                  ÀÀ&7have the Yellow Beacon Effect.
17:43:09.708
wynntils
Render thread
[CHAT]                    À                   &7Yellow Beacons cannot appear
17:43:09.708
wynntils
Render thread
[CHAT]                    À                                 À&7anymore`

wiji1: [patterns.txt](https://github.com/user-attachments/files/15864484/patterns.txt)

wiji1:
Can we use Dark Red for the Failed type?

kristofbolyai:
> Can we use Dark Red for the Failed type?

Sure.

kristofbolyai:
Okay, please test this version @wiji1, if you think it works fine, let me know.

wiji1:
> Okay, please test this version @wiji1, if you think it works fine, let me know.

The version's mission tracking works perfectly. The overlay had some issues though. I'm not sure if you intended this with the overlay script you pushed, but it displayed missions before they were completed, causing them to show up as Unknown. New line characters were also not implemented.

I went ahead and fixed those two issues in my recent commit.

If the changes are to your liking, I can confirm that this version is tested and works well.

kristofbolyai:
The unknown thing was kind of what I wanted, but I realize it does not look so great. I wasn't sure if newlines were needed, but this way we know it was. Thanks for testing.

kristofbolyai (approved):

kristofbolyai:
Thanks for contributing!

PR #2601, opened 2024-06-16
fix: Fix hades users appearing in full size on minimap

kristofbolyai:


magicus:
Oops! Sorry. Thanks for fixing.

PR #2605, opened 2024-06-18
feat: Rewrite ResourcePackService to work better with Minecraft (1.21 backport)

kristofbolyai:
# Tests:
- [x] Disabling the feature
- [x] Deleting options.txt
- [x] Deleting the resource pack
- [x] Changing the resource pack/disabling it
- [x] Joining a Wynn with a resource pack
- [x] Joining a Wynn without a resource pack
- [x] Joining a Wynn with a different resource pack
- [x] Deleting the preloaded pack hash
# Additional, non-critical tests:
- [x] Join a non-Wynn server without a resource pack with a preloaded pack, gets removed on join, and preloaded after disconnect
- [x] Join a non-Wynn server with a resource pack with a preloaded pack, gets removed on join, and preloaded after disconnect


common/src/main/java/com/wynntils/mc/event/ServerResourcePackLoadEvent.java
```diff
@@ -8,15 +8,17 @@
...
 
-/** Fires on receiving {@link net.minecraft.network.protocol.game.ClientboundResourcePackPacket} */
-@EventThread(EventThread.Type.IO)
+/**
+ * Fired on receiving {@link net.minecraft.network.protocol.common.ClientboundResourcePackPacket}
+ */
+@EventThread(EventThread.Type.ANY)
 @Cancelable
-public class ResourcePackEvent extends Event {
+public class ServerResourcePackLoadEvent extends Event {
```
magicus: Maybe you could combine the two under a ServerResourcePackEvent with a Load and Clear nested subclass? Just to keep down on the number of event files.

common/src/main/java/com/wynntils/mc/mixin/ClientCommonPacketListenerImplMixin.java
```diff
@@ -24,8 +26,15 @@ public abstract class ClientCommonPacketListenerImplMixin {
             at = @At("HEAD"),
             cancellable = true)
     private void handleResourcePackPre(ClientboundResourcePackPacket packet, CallbackInfo ci) {
-        ResourcePackEvent event = new ResourcePackEvent(packet.getUrl(), packet.getHash(), packet.isRequired());
-        MixinHelper.post(event);
+        // Hash is SHA-1 of the URL in 1.20.2
+        // Hash is packet#hash() in 1.21
```
magicus, line 30: Oh, vanilla is joining the 21st century and abandoning sha-1! :-D

kristofbolyai: Iirc it's still Sha-1, just calculated a bit differently :) I don't remember, there are too many IDs and hashes around here :).

magicus (approved):
I like the clear code and the comments. Resource pack loading suddenly became very much less black magic...

kristofbolyai:
I very much agree, I hope this should also make it less of a hassle to port in the future, as this feature took the most effort every version change. Considering this is a 1.21 backport, I can say, only a few lines needs changing, e.g. settings are now in records for packs and the resource hashes are provided by MC.

common/src/main/java/com/wynntils/mc/event/ServerResourcePackEvent.java
```diff
@@ -0,0 +1,43 @@
...
+            return hash;
+        }
+
+        public boolean isRequired() {
+            return required;
+        }
+    }
+
+    @EventThread(EventThread.Type.ANY)
+    public static class ServerResourcePackClearEvent extends ServerResourcePackEvent {}
```
magicus: We used to have a convention where an inner class like this would be named only `Clear`, so the total name would be `ServerResourcePackEvent.Clear`. This convention is used in e.g. `RenderEvent`. 

Now it seems like most new classes has followed this new convention where the names just stack, and the full name becomes almost duplicated.

I'm not sure I like that. I kind of liked the old, shorter format as well, even if it meant that the actual event class name did not end in "Event". But I am open to other views as well. :-) At the very least, we should be consistent. 

Fixing this consistently is a separate issue, but I'll just want to note that what I envisioned when I suggested this was the shorter format. Not saying you should change anything, just trying to explain that I did not really think ` ServerResourcePackEvent.ServerResourcePackLoadEvent` was better than what was before. :-)

magicus (approved):

common/src/main/java/com/wynntils/mc/event/ServerResourcePackEvent.java
kristofbolyai: Oh, I just forgot to rename these, don't worry about it 😅

ShadowCat117 (approved):

common/src/main/java/com/wynntils/mc/event/ServerResourcePackEvent.java
magicus: Ok, thanks, but my observation still stands -- we have a mix of both kinds of naming, and we should really do something about it...

PR #2606, opened 2024-06-18
port: Start using WynntilsCheckbox exclusively (1.21 backport) [skip ci]

kristofbolyai:
This change is easily done, and allows us to not depend on Minecraft's checkbox, which received breaking changes in 1.21 that makes it impractical to use for us.

### DIFF
--- a/common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
+++ b/common/src/main/java/com/wynntils/screens/base/WynntilsGridLayoutScreen.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © Wynntils 2023.
+ * Copyright © Wynntils 2023-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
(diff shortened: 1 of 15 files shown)


magicus (approved):

ShadowCat117:
The consuming checkbox on chat tabs screen doesn't seem to be toggleable, though that also happens in the current release so not caused by this. Might also be worth extending its max text width as it's scrolling on gui scale 3+ even though there's plenty of room for the text

kristofbolyai:
@ShadowCat117 I fixed the width issue, but I seem to be able to toggle the checkbox.

ShadowCat117:
Hmm for me it makes the clicked sound but doesn't actually update. Does it actually require a valid pattern to update, I don't use chat tabs so don't really know how it works honestly

kristofbolyai:
Hm. I will leave it up to you to fix this then, I have no idea, it works for me. Also, it doesn't really block this PR, so it's fine.

PR #2610, opened 2024-06-25
feat: Action bar and held item overlays [skip ci]

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/c57c65f2-1728-4ef9-b0b8-3bdb8d616ff6


common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
```diff
@@ -167,4 +170,33 @@ private void onRenderHeartsPre(
...
+            at = @At("HEAD"),
+            ordinal = 0)
+    private Component setActionBarMessage(Component value) {
+        if (!MixinHelper.onWynncraft()) return value;
+
+        ActionBarMessageSetEvent event = new ActionBarMessageSetEvent(value);
+        MixinHelper.post(event);
+
+        return event.getMessage();
+    }
```
ShadowCat117, line 201: In 1.21 there is a `renderOverlayMessage` so this can be replaced with that

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
```diff
@@ -167,4 +170,33 @@ private void onRenderHeartsPre(
             ci.cancel();
         }
     }
+
+    @Inject(
+            method = "Lnet/minecraft/client/gui/Gui;renderSelectedItemName(Lnet/minecraft/client/gui/GuiGraphics;)V",
```
kristofbolyai, line 175:
```suggestion
            method = "renderSelectedItemName(Lnet/minecraft/client/gui/GuiGraphics;)V",
```

Generally, we do this weird, semi-qualified method naming for mixins. I can see the benefits. If you only used the method names, things can break and go unnoticed when porting. The fully qualified name's first part, the holder class itself, however, has no meaningful use here.

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
```diff
@@ -167,4 +170,33 @@ private void onRenderHeartsPre(
...
+            at = @At("HEAD"),
+            cancellable = true)
+    private void onRenderSelectedItemNamePre(GuiGraphics guiGraphics, CallbackInfo ci) {
+        if (!MixinHelper.onWynncraft()) return;
+
+        RenderEvent.Pre event =
+                new RenderEvent.Pre(guiGraphics, 0, this.minecraft.getWindow(), RenderEvent.ElementType.HELD_ITEM_NAME);
```
kristofbolyai, line 182: Help me out. For the other ElementTypes, do we always fire a Pre/Post event, or just the ones we use? If we do, please fix it here. If not, well, I hope it won't cost someone time in the future.

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
kristofbolyai: I think you should try Forge. I am at least 90% sure this mixin won't work. Forge has a separate `ForgeGui` class on this version (and patches the regular `Gui` class on 1.21`. This will be a bit of a pain to port, but oh well, this is the reason I've separated Fabric/Forge only mixins in 1.21.

common/src/main/java/com/wynntils/overlays/minecraft/ActionBarOverlay.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public void render(PoseStack poseStack, MultiBufferSource bufferSource, float partialTicks, Window window) {
+        if (hideActionBar.get()) return;
+        if (messageTimer <= 0) return;
+
+        // Render it the same way vanilla renders item changes
+        int alpha = (int) Math.min((float) messageTimer * 256.0F / 10.0F, 255.0F);
+        if (alpha <= 0) return;
```
kristofbolyai, line 79: If you check MapData or the GameUpdateOverlay (the notification one), there is a transparency cutoff at 0.1f or higher, as <0.1f alphas tend to be rendered with full transparency, so it kind of "flashes" when rendering. Not sure if you've ever encountered that here, or with that other overlay, it's been not so long I've fixed it.

common/src/main/java/com/wynntils/overlays/minecraft/HeldItemOverlay.java
```diff
@@ -0,0 +1,139 @@
...
+    public void onTick(TickEvent event) {
+        ItemStack currentItem = InventoryUtils.getItemInHand();
+
+        if (!ItemStack.matches(currentItem, heldItem)) {
+            heldItem = currentItem;
+
+            if (currentItem.getItem() == Items.AIR) {
+                message = StyledText.fromString("");
+            } else {
+                // FIXME: Italics don't render
```
kristofbolyai, line 77: So, how come :)? What are we talking about? Those custom named items, that are made italic by Minecraft?

kristofbolyai (commented):
This was not an in-depth review, more of just reading code and commenting my thoughts. Here is what I've been thinking about since I read the title. "Hm, it's getting to be the time to refactor overlays". I think we are close to needing it now :). The fact that you keep needing to repeat yourself on the management screen also tells me that (custom, minecraft, etc).

Secondly. Previously, when people mentioned "Minecraft Overlays", I've thought about them as a special overlay, like TextOverlay. A wrapper class, that can position the vanilla overlay, basically wrapping the appropriate RenderEvent. I see how you lose the benefits of font styling there. Maybe it can be wrapped specifically, separately for text overlays and other ones, like the HUD. Copying what Minecraft does never ends well, see the nametag rendering fix in 1.21 (from Discord), or beacon rendering that I've removed from 1.21 and merged with MInecraft code. So, we'll have to think about a way to make this more of a wrapper, than an invention, in that sense.

Lastly, it's weird to have these overlays be "disabled" with a config, that disables rendering the overlay, but keeps it enabled? When you disable a "disabled" (hidden) overlay, it becomes enabled, when you disable an "enabled" overlay, it resets its position. If this is how it works, and I understand it correctly, we need to address that. Whether that means a patch to the management screens, or starting an overlay refactor, is up to discussion.

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
ShadowCat117: Ah, I mainly wanted to start doing the "vanilla overlays" so I didn't have to port removehud again so if it's easier it can be done after 1.21, it's not necessarily needed for 1.20

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
ShadowCat117: Oops, Minecraft dev plugin copies the whole thing, didn't think to check how we usually do it :)

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
ShadowCat117: Health bar, food bar, tab overlay and crosshair (Do we even do anything with that?) all only have pre. The `render` method is the only one to have pre and post

common/src/main/java/com/wynntils/overlays/minecraft/ActionBarOverlay.java
ShadowCat117: I copied this from the spell cast overlay, I don't think I've ever seen the flash occur on that but will double check

common/src/main/java/com/wynntils/overlays/minecraft/HeldItemOverlay.java
ShadowCat117: I suppose "don't render" isn't a good description of the issue. `heldItem.getHoverName()` just doesn't return the `§o`, it was most noticeable on soul points, though we won't have to worry about that for long, but it was also missing the colour for the soul point count. the 15 is meant to be yellow and italics but it's just white.

I also had some other item that was just italics and one colour but the italics still didn't work

ShadowCat117: Top is overlay, bottom vanilla
image
image

common/src/main/java/com/wynntils/overlays/minecraft/ActionBarOverlay.java
ShadowCat117: No, it's not an issue

kristofbolyai: Even with outline text shadow?

ShadowCat117: Even with it

ShadowCat117:
I did start out doing this separately from the overlay system, just modifying the x/y position variables like how removehud does it, see here https://github.com/ShadowCat117/Artemis/commit/8dc02049a4aad80b6af27a036b87a72d4851a686 but that was just a bunch of offset configs which isn't great. I don't think these two are that bad for copying how Minecraft does it, but the xp bar and hotbar which would be the next ones to do, may run into some issues. That's why I've done these first just to get some feedback on the implementation.

And yeah the disabling is weird. If "our" one is being rendered, it always cancels the vanilla one, but if you turn off the overlay then the vanilla one will render, there's not currently a nice way to disable both without having to turn off the overlay itself, and a config to disable the vanilla one.

common/src/main/java/com/wynntils/mc/mixin/GuiMixin.java
ShadowCat117: Action bar works for forge but not held item, for some reason they have a parameter to shift the y level in their `Gui` class `renderSelectedItemName(GuiGraphics arg, int yShift)` 

It's not in `ForgeGui` so I don't think it can be injected in `ForgeGuiMixin` unless I'm just doing the mixin wrong. So definitely probably better saving this for 1.21 and if we do want to refactor overlays then maybe that first too

ShadowCat117:
To be revisited later after overlay refactors

PR #2612, opened 2024-06-29
fix: fix bigger config/storage files not getting saved properly

kristofbolyai:
This is a long shot, there is no solid documentation on this from Java's part. Flushable/Closable are different interfaces, flushable does not extend closable. That means that a flushing before closing is an implementation specific detail, which may cause these issues.


ShadowCat117 (approved):
Hopefully it works :)

PR #2617, opened 2024-06-29
feat: Lootrun Task Name Force Coordinates option [skip ci]

Rafii2198:
Adds an option that makes it so the Lootrun Task Name overlay will always show the coordinates instead of names.
I did it by adding a new function `{lootrun_task_location}` that returns a location of a given task, people can use for their own functions, for example showing the distance from the player to each one or showing the closest one, etc.

image

Somebody asked if it is possible, so I decided to implement such option


common/src/main/java/com/wynntils/overlays/lootrun/LootrunTaskNameOverlay.java
```diff
@@ -1,18 +1,26 @@
...
..._GRAY\"))); \"]\";\"\\n\"));IF_STRING(STRING_EQUALS(lootrun_task_name(\"GREEN\"); \"\");\"\";CONCAT(\"§aGreen: \"; lootrun_task_type(\"GREEN\");\" at \";\"[\"; str(x(lootrun_task_location(\"GREEN\"))); \", \"; str(y(lootrun_task_location(\"GREEN\"))); \", \"; str(z(lootrun_task_location(\"GREEN\"))); \"]\";\"\\n\"));IF_STRING(STRING_EQUALS(lootrun_task_name(\"RED\"); \"\");\"\";CONCAT(\"§cRed: \"; lootrun_task_type(\"RED\");\" at \";\"[\"; str(x(lootrun_task_location(\"RED\"))); \", \"; str(y(lootrun_task_location(\"RED\"))); \", \"; str(z(lootrun_task_location(\"RED\"))); \"]\";\"\\n\"));IF_STRING(STRING_EQUALS(lootrun_task_name(\"WHITE\"); \"\");\"\";CONCAT(\"§fWhite: \"; lootrun_task_type(\"WHITE\");\" at \";\"[\"; str(x(lootrun_task_location(\"WHITE\"))); \", \"; str(y(lootrun_task_location(\"WHITE\"))); \", \"; str(z(lootrun_task_location(\"WHITE\"))); \"]\";\"\\n\"));IF_STRING(STRING_EQUALS(lootrun_task_name(\"RAINBOW\"); \"\");\"\";CONCAT(\"§4R§ca§6i§en§ab§2o§bw§9: §f\"; lootrun_task_type(\"RAINBOW\");\" at \";\"[\"; str(x(lootrun_task_location(\"RAINBOW\"))); \", \"; str(y(lootrun_task_location(\"RAINBOW\"))); \", \"; str(z(lootrun_task_location(\"RAINBOW\"))); \"]\"))))}";
```
magicus, line 22: Our "scripting" language might be powerful, but this is horribly unreadable. I wonder if we can change the syntax to e.g. allow CR as a whitespace, and merge multiple whitespace into a single space. That would at least allow us to put in line breaks and indentations into this mess...

(Not a comment on the PR itself, just a remark about future potential development)

Rafii2198: This is of one the issues new users have, it is really hard to read and thus write, especially that functions are usually repetitive which means making one typo will break whole thing and will be hard to spot.
At least you can use new lines just fine, but I don't think it is possible for us to do in code easily unless I am missing something.

kristofbolyai: You can use whitespaces and newlines (they will get ignored). In fact, some overlays have formatted code, with `"""` blocks. This is just not one of them.

kristofbolyai (approved):
I guess this works, although, while we are here, I would like to check what happens when we don't have name data.

kristofbolyai:
It shows "Unknown at [coord]". That should work.

PR #2618, opened 2024-07-01
refactor: rewrite gammabright

nerd:
this commit rewrites the Gammabright feature.

### motivation
the current Gammabright feature writes directly to Minecraft's settings, which can potentially lead to state desync between Artemis and Minecraft, and subsequently writing out of bounds values to the Minecraft settings file. the code that manages this state is overly complex for what should be a very simple feature.

### the change
this rewrite instead simply writes our desired value (maximum brightness) directly to the lightmap. the lightmap is recalculated every render frame and never written to disk, so there is nothing to clean up. this stateless approach avoids interfering with Minecraft settings and significantly reduces the code complexity of the feature.

### DIFF
--- a/common/src/main/java/com/wynntils/mc/mixin/LightTextureMixin.java
+++ b/common/src/main/java/com/wynntils/mc/mixin/LightTextureMixin.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.mc.mixin;
+
+import com.mojang.blaze3d.platform.NativeImage;
+import com.wynntils.core.events.MixinHelper;
+import com.wynntils.mc.event.LightmapEvent;
+import net.minecraft.client.renderer.LightTexture;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@Mixin(LightTexture.class)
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/mc/mixin/LightTextureMixin.java
```diff
@@ -0,0 +1,25 @@
...
+public class LightTextureMixin {
+    @Redirect(
+            method = "updateLightTexture",
+            at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/platform/NativeImage;setPixelRGBA(III)V"))
+    private void updateLightmapRGB(NativeImage image, int x, int y, int rgb) {
+        final LightmapEvent lightmapEvent = new LightmapEvent(rgb);
+        MixinHelper.post(lightmapEvent);
+        image.setPixelRGBA(x, y, lightmapEvent.getRgb());
+    }
+}
```
kristofbolyai: I am not happy with this mixin. Can you use `WrapOperation` to make this more compatible with other mods?

kristofbolyai (commented):
Overall, I like the idea.

common/src/main/java/com/wynntils/mc/mixin/LightTextureMixin.java
nerd: sure thing. i've pushed an updated version. not super familiar with MixinExtras so let me know if the updated one is what you had in mind

kristofbolyai (approved):
Hm, looks good.

kristofbolyai:
Thank you!

PR #2620, opened 2024-07-02
chore: Update mod incompatibilities [skip ci]

ryanzhoudev:
Weirdchat removed because the mod is dead (700 downloads, no update for a year). Also no version exists for 1.21.
Memory Leak Fix removed because 1.1.2 is only for Minecraft 1.16 or lower.

### DIFF
--- a/fabric/src/main/resources/fabric.mod.json
+++ b/fabric/src/main/resources/fabric.mod.json
@@ -4,12 +4,11 @@
   ],
   "breaks": {
     "exordium": "*",
-    "memoryleakfix": "<1.1.2"
+    "ias": "*"
   },
   "conflicts": {
     "avomod": "*",
-    "chatpatches": "*",
-    "weirdchat": "*"
+    "chatpatches": "*"
   },
   "contact": {
     "homepage": "https://wynntils.com",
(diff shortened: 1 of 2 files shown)


Essentuan:
IAS should probably be conflicts not breaks no?

ryanzhoudev:
I see where you're coming from, some user can just not switch and it'll be fine, but imho users just shouldn't have any switcher installed in the same instance as wynntils to avoid the risk of configs being wiped

Essentuan:
I know of significantly more people who would be hurt by labeling any account switcher (such as essentials, unless that does things differently than IAS?) as 'breaks' instead of 'conflicts'. Frankly if you *are* using an account switcher I'd view that as you're willing to accept the risk of your config being wiped. I've also used IAS since the early days of artemis and have never had my config deleted. As well wouldn't most people who use an account switcher prefer the config from their main anyways?

Essentuan:
A much better solution imo would be to warn the user the first time they log onto wynntils with IAS (or any other account switcher mod)

ryanzhoudev:
yeah i guess i can look into that

kristofbolyai:
I mean, just add it into conflicts then. Really no point in having a separate feature.

kristofbolyai (approved):

PR #2621, opened 2024-07-02
fix: 2.1 attack timer fixes and don't send defence levels on failed attacks [skip ci]

### LINKED ISSUE #2542: WAR_MESSAGE_PATTERN does not handle x minute and y second case
Current regex: `"§3\\[WAR\\]§c The war for (?<territory>.+) will start in (?<remaining>.+) (?<type>minutes|seconds)\\."`

`[CHAT] &3[WAR]&c The war for Aldorei's Arch will start in 14 minutes and 59 seconds.`

ryanzhoudev:
I kind of turned this into a weird queue styled thing. Should be much better though.
Fixes #2542


common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1205,6 +1205,7 @@
   "feature.wynntils.territoryAttackTimerOverlay.overlay.territoryAttackTimer.textShadow.description": "What should the text shadow look like?",
   "feature.wynntils.territoryAttackTimerOverlay.overlay.territoryAttackTimer.textShadow.name": "Text Shadow",
   "feature.wynntils.territoryDefenseMessage.description": "Sends the territory defense message when you attack a territory.",
+  "feature.wynntils.territoryDefenseMessage.message": "%s defense is %s",
```
ShadowCat117: Might be wrong but don't we get the defense from this chat message in the attack timer model? Probably not a good idea to make it translatable if so

ShadowCat117:
I think it also sends the message when you don't have enough emeralds to queue an attack if you wanted to fix that too, not sure what happens if the territory is on cooldown, that might work too

common/src/main/resources/assets/wynntils/lang/en_us.json
kristofbolyai: That's right, that's almost the whole point of the message.

ShadowCat117: Can we not get it from the advancement? Or does that not update frequently enough

kristofbolyai: The one in the attack menu is basically resource tick accurate, the advancement one updates like every 10 mins or so

ryanzhoudev:
> I think it also sends the message when you don't have enough emeralds to queue an attack if you wanted to fix that too, not sure what happens if the territory is on cooldown, that might work too

Unfortunately I can't test either of these cases, I'm in my own empty holding guild at the moment. I'll see if someone can provide screenshots

ShadowCat117:
Cooldown should be easy enough to check, just try to attack a territory that's been held for less than 10 minutes. Emeralds one I can check next time my guild goes on the attack and fix that later if it is indeed an issue

ShadowCat117:
Ok can confirm not having enough emeralds will send the message, it will have `✖` by the price instead so maybe check for that or the tick
image

Cooldown isn't an issue though, you can't open the attack menu whilst it's on cooldown

ryanzhoudev:
I'm gonna rework this entirely, I can just wait for a chat message after the click and send it if it appears, no need for all of these special cases

Essentuan:
GuildAttackTimerModel relies on the defenses for a territory being sent before the attack message.

ryanzhoudev:
> GuildAttackTimerModel relies on the defenses for a territory being sent before the attack message.

..why? Also, what part of it? Seems like it's fine? image

Essentuan:
Ok I'm ass at github so I dont know how to show this easily but

```
        matcher = event.getOriginalStyledText().getMatcher(GUILD_DEFENSE_CHAT_PATTERN);
        if (matcher.matches()) {
            String territory = matcher.group(1);
            territoryDefenses.put(territory, GuildResourceValues.fromString(matcher.group(2)));
            return;
        }
```

this part from [GuildAttackTimerModel](https://github.com/Wynntils/Artemis/blob/main/common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java)

ryanzhoudev:
The only places I see the defense value being used is in the beacon coloring and the overlay. Both of which still work

Essentuan:
When adding a new timer GuildAttackTimerModel will first check if the defense message has been sent. If no defense message was sent it will then get the defense from the advancement menu. So it requires that defense message be sent before the timer gets added otherwise the shown defense will be significantly less accurate.

Essentuan:
Ok that was inaccurate its used in [TerritoryAttackTimer](https://github.com/Wynntils/Artemis/blob/main/common/src/main/java/com/wynntils/models/territories/TerritoryAttackTimer.java#L48) and doesn't break.  Sorry.

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
```diff
@@ -10,19 +10,28 @@
...
 import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.EventPriority;
 import net.neoforged.bus.api.SubscribeEvent;
 
 @ConfigCategory(Category.UNCATEGORIZED)
 public class TerritoryDefenseMessageFeature extends Feature {
     private static final Pattern ATTACK_SCREEN_TITLE = Pattern.compile("Attacking: (.+)");
     private static final Pattern TERRITORY_DEFENSE_PATTERN = Pattern.compile("Territory Defences: (.+)");
+    private static final Pattern TERRITORY_ATTACK_MESSAGE_PATTERN =
+            Pattern.compile("§3\\[WAR]§c ?The war for (.+) will start in \\d+ minutes?\\.");
```
ShadowCat117: https://github.com/Wynntils/Artemis/issues/2542 is also relevant here

kristofbolyai: Maybe it would be the best not to duplicate this, rather make it public in the model. Same with the ones above, if they are in a model.

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
```diff
@@ -10,19 +10,28 @@
...
+    private static final long MESSAGE_TIMEOUT = 3000; // 3 seconds for the server to respond to an attack command
```
kristofbolyai:
```suggestion

    // 3 seconds for the server to respond to an attack command
    private static final long MESSAGE_TIMEOUT = 3000; 
```

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
```diff
@@ -37,10 +46,32 @@ public void onInventoryClick(InventoryMouseClickedEvent event) {
...
+                        Pair.of(
+                                System.currentTimeMillis(),
+                                "g %s defense is %s".formatted(titleMatcher.group(1), matcher.group(1))));
                 return;
             }
         }
     }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onChatMessage(ChatMessageReceivedEvent e) {
```
kristofbolyai: Sounds like this would also send the defence in this case: 
- A clicked attack, it failed, defense was saved
- B clicked attack, it happened, both send the defence.

So, perhaps, this needs to be a timed map. I would also advise against using a `Pair<Long, String>`. At least, make it a `Pair<Long, TerritoryDefense(idr the name)>`, or optionally, add a private record to this model that holds a `long timestamp, TerritoryDefense defense`. Lastly, this was my bad originally, but please extract the format string to a constant, while you are at it. Thanks.

ryanzhoudev: How would a timed map fix this? In both situations, if a second person attacks the same territory within MESSAGE_TIMEOUT, both will send it. We can adjust this to be lower, but it might not send when the server is laggy then. But if someone else attacks after MESSAGE_TIMEOUT, the first person won't send it.

kristofbolyai: Hm. Is the war attack sound played for everyone, or just the "attacker"? I am not sure if we have any "clue" that we can use to fix this, or we are left with this issue.

kristofbolyai:
So did this change with 2.1 in any way? Maybe we can do something more accurate now.

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
ryanzhoudev: The sound solution does work. I've just finished it now

kristofbolyai:
Does this mean the attack sound is only played for the player who actually attacked? Or, are we using sounds now, but the double messages are still a problem?

ryanzhoudev:
Sounds are only played for the attack initiator. Now that I think about it, isn't this what the war horn was supposed to solve?

common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java
```diff
@@ -50,8 +50,9 @@
 public final class GuildAttackTimerModel extends Model {
     private static final Pattern GUILD_ATTACK_PATTERN = Pattern.compile("§b- (.+):(.+) §3(.+)");
     private static final Pattern GUILD_DEFENSE_CHAT_PATTERN = Pattern.compile("§3.+§b (.+) defense is (.+)");
+    // Test in GuildAttackTimerModel_WAR_MESSAGE_PATTERN
     private static final Pattern WAR_MESSAGE_PATTERN = Pattern.compile(
-            "§3\\[WAR\\]§c The war for (?<territory>.+) will start in (?<remaining>.+) (?<type>minutes|seconds)\\.");
+            "§c(?:\uDAFF\uDFFC\uE006\uDAFF\uDFFF\uE002\uDAFF\uDFFE|\uDAFF\uDFFC\uE001\uDB00\uDC06) The war for (?<territory>.+)(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? will(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? start(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? in(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? ((?<minutes>\\d+)(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? minute(?:s)?)?(?:(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? and )?((?<seconds>\\d+)(?: ?\\n\uDAFF\uDFFC\uE001\uDB00\uDC06)? second(?:s)?)?\\.");
```
kristofbolyai: What's up with all these alignment characters..? Maybe StyledText should have a method for removing them? Maybe like `trimAlignment`?

kristofbolyai:
> Sounds are only played for the attack initiator. Now that I think about it, isn't this what the war horn was supposed to solve?

Probably yes, but I am glad to have found a solution.

common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java
kristofbolyai: Okay, since this is the only thing holding this PR back, I'll add this shortly.

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
```diff
@@ -37,10 +49,32 @@ public void onInventoryClick(InventoryMouseClickedEvent event) {
...
                 return;
             }
         }
     }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onAttackSound(SoundPlayedEvent event) {
+        if (queuedTerritories.isEmpty()
+                || event == null
+                || !event.getSoundInstance().getLocation().equals(ATTACK_SOUND)) return;
```
kristofbolyai:
```suggestion
        if (queuedTerritories.isEmpty()
                || !event.getSoundInstance().getLocation().equals(ATTACK_SOUND)) {
                return;
        }
```

event can't be null

kristofbolyai: Did you mean event.getSoundInstance() == null?

common/src/main/java/com/wynntils/models/territories/GuildAttackTimerModel.java
```diff
@@ -74,11 +75,11 @@ public GuildAttackTimerModel(MarkerModel marker) {
     public void onMessage(ChatMessageReceivedEvent event) {
         if (event.getRecipientType() != RecipientType.GUILD) return;
 
-        Matcher matcher = event.getOriginalStyledText().getMatcher(WAR_MESSAGE_PATTERN);
+        Matcher matcher = event.getStyledText().getMatcher(WAR_MESSAGE_PATTERN);
```
kristofbolyai: You don't seem to use the "stripAlignment" method here? Also, watch out for double whitespaces, that are after those unicode texts, you may need to do a `replaceAll("\s\s", "")`

common/src/main/java/com/wynntils/features/TerritoryDefenseMessageFeature.java
ryanzhoudev: No, it was just left over when I was testing the sound mixin.

common/src/main/java/com/wynntils/core/text/StyledText.java
```diff
@@ -272,6 +273,36 @@ public StyledText stripAlignment() {
         });
     }
 
+    /**
+     * Cleans a StyledText into a single-line string, preserving only ascii, formatting codes, and single whitespaces.
+     * @return a new cleaned {@link StyledText}
+     */
+    public StyledText clean() {
```
kristofbolyai: I am sorry that I still have to nag you about this PR, but I don't think this method in it's current form is suitable for `StyledText`. I think it's implementation does not fit into every other `StyledText` native method, that uses the text's parts, and does string manipulation to the parts' string itself. However, this is not my biggest concern. I think that this method does too many things, so it wouldn't be useful to other models apart from this single use case.

Here is my proposal to fix it:
1. Create a method in `StyledTextUtils` to join newlines into spaces. Basically `StyledTextUtils#joinAllLines` that is a simple call to `styledText.replaceAll("\n", " ")`. Sidenote: Do you really need to join with a space? Wouldn't that create double spaces in the place of newlines? Maybe it'd be wiser to just remove all newlines, and replace them with nothing.
2. Now, you can produce a `StyledText` without alignments, with lines joined. What does that single Unicode character look like, that you want to get rid of? If it does not change, maybe you could just call `replace(All)` to get rid of it? I understand the idea of wanting to create a method that removes all custom Unicode characters from a string, but since you don't know which characters are retextured, which are not, etc, this is not really viable at the moment.

ryanzhoudev: For 2. I made a new method that essentially remakes the StyledText in a single part. Turns out the regex doesn't work for that unicode because it's on separate parts, but I didn't want to modify the replaceAll since maybe the by-part modification was intentional? Seems like it because it mentions PartStyle

PR #2623, opened 2024-07-06
fix: Switching accounts in game breaks configs

ryanzhoudev:
This appears to work for now. I only tested once though. If someone else experiences problems let me know.

On hold for new mixin version that allows us to "watch" a variable.


ShadowCat117:
Should this really be a feature? I feel like it should just be part of the manager.

Also don't forget to reload the storage file too

ryanzhoudev:
> Should this really be a feature? I feel like it should just be part of the manager.

Yes, since some people mentioned wanting to be able to use one account's config for another.

> Also don't forget to reload the storage file too

Oh yeah thanks.

ShadowCat117:
> Yes, since some people mentioned wanting to be able to use one account's config for another.

I don't think this solution will fully support that. If you log in as account A, then swap to account B with this feature disabled you will still be saving changes to account A's config. So if you then loaded up as account B first you wouldn't have the changes saved to account A's config.

I think it would just be better to have the global config that was mentioned on discord the other day, but I don't know if/when we plan on doing that

kristofbolyai:
Fully agree with ShadowCat. Let's just fix config wipes, nothing else should be done here and now (without an explicit config rework, which won't be a priority during beta).

common/src/main/java/com/wynntils/core/persisted/config/ConfigManager.java
```diff
@@ -333,4 +333,10 @@ private <T> void setConfigValue(Config<T> config, Object value) {
             config.setValue(typedValue);
         }
     }
+
+    public void userSwitched() {
```
magicus: I think a better name would be like `reloadAllConfigs` or something like that. The method should describe what it does, not when it is called.

kristofbolyai:
I have an idea, not sure how feasible. What if you mixin into the Minecraft code, where the user uuid is stored, and basically "inject on change". Iirc there is a way to do that with mixin. If that's possible, you can, well, not signal an event, because it might be too early :), but check for `WynntilsMod#isInitalized`, then call Config/Storage manager directly, to reload the file. 

This is probably a long shot, but I feel like this is the most "complete" way. This also means, you don't have to add a new, kind of weird, feature or service.

ryanzhoudev:
> I have an idea, not sure how feasible. What if you mixin into the Minecraft code, where the user uuid is stored, and basically "inject on change". Iirc there is a way to do that with mixin. If that's possible, you can, well, not signal an event, because it might be too early :), but check for `WynntilsMod#isInitalized`, then call Config/Storage manager directly, to reload the file.
> 
> This is probably a long shot, but I feel like this is the most "complete" way. This also means, you don't have to add a new, kind of weird, feature or service.

I'm not sure this will work. The `User user` variable is `private final`, and IAS uses an accessor to mess with it. But it's also possible that some other mods or clients use AW, which if we inject into the getUser method, we won't detect. I didn't find anything that allows us to "watch" a variable. Thoughts?

ryanzhoudev:
Not interested in trying to update and resolve this mess, when the new mixins stuff comes out we can look at it again. I will make an issue.

PR #2637, opened 2024-07-07
feat: Rework custom bank page names and quick jumps for 2.1

ShadowCat117:
https://github.com/Wynntils/Artemis/assets/8337467/aaa1162a-8bc8-4e4c-8c5d-ac680474a3ee


common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
```diff
@@ -83,6 +85,12 @@ public void onContainerRender(ContainerRenderEvent event) {
                 ? containerScreen.leftPos + 2
                 : screen.width - containerScreen.leftPos - 2;
 
+        if (Models.Container.getCurrentContainer() instanceof PersonalStorageContainer
+                && Managers.Feature.getFeatureInstance(PersonalStorageUtilitiesFeature.class)
+                        .isEnabled()) {
+            textureX -= 10 + Texture.BANK_PANEL.width();
+        }
```
ShadowCat117, line 92: Don't like this but not sure what the best alternative is, maybe changing the texture locations for banks to be above on either side of the bank type name?

ShadowCat117:
Some changes may be made to the textures but those can be changed later

common/src/main/java/com/wynntils/features/inventory/InventoryEmeraldCountFeature.java
kristofbolyai: Hm, yes. This is more of a texture issue than a code issue, so there is no point in holding this PR for it,

kristofbolyai (approved):

PR #2627, opened 2024-07-06
fix: Update containers for 2.1

ShadowCat117:
image

PR #2638, opened 2024-07-07
fix: Fix skill point annotator [skip ci]

ShadowCat117:
image
image


ryanzhoudev (approved):
Can you revert the CustomBankPageNamesFeature.java change tho?

ShadowCat117:
I think it's just because spotless wasn't ran on #2627.

I've removed the feature in #2637 anyway and merged it with the new feature

kristofbolyai (commented):
Actually, it would be wise to check all uses of À, because it was removed from many places.

ShadowCat117:
> Actually, it would be wise to check all uses of À, because it was removed from many places.

Sure, I can take a look at some

PR #2640, opened 2024-07-08
fix: Fix emerald pouch annotator [skip ci]

ShadowCat117:
image
image


kristofbolyai (commented):
Maybe leave the space in the regex. They have done these kind of changes back before and reverted it.

PR #2641, opened 2024-07-09
fix: Horse keybind no longer relies on soul point item [skip ci]

ryanzhoudev:


common/src/main/java/com/wynntils/features/combat/HorseMountFeature.java
```diff
@@ -119,11 +118,43 @@ private void mountHorse(Entity horse) {
...
+            Optional<HorseItem> horseItemOpt = Models.Item.asWynnItem(itemStack, HorseItem.class);
+            if (horseItemOpt.isPresent()) {
+                return i;
+            }
+
+            // weapons (temp. disabled due to wynn bug)
+            //            Optional<GearItem> gearItemOpt = Models.Item.asWynnItem(itemStack, GearItem.class);
+            //            if (gearItemOpt.isPresent() && gearItemOpt.get().getGearType().isWeapon()) {
+            //                return i;
+            //            }
```
kristofbolyai: I think, just enable this, and write it off as a wynn bug. They'll fix it, and we won't have to do a new patch.

kristofbolyai (approved):
Looks good otherwise.

PR #2643, opened 2024-07-09
chore: remove (almost) everything related to soul points [skip ci]

kristofbolyai:
I've got side tracked while working on the action bar, so I might as well do this now.


kristofbolyai:
@DonkeyBlaster I've left a FIXME for you in skill point model, when you have time, could you look at it?

PR #2646, opened 2024-07-09
fix: Fix crashing in character selection screen

ShadowCat117:


kristofbolyai (approved):
Thanks. I checked the logs, knew it was the pose stacks somewhere..

ShadowCat117:
Yeah honestly that whole render method is a mess with all those posestack translations and scales, I need to get around to redoing it at some point

PR #2649, opened 2024-07-10
fix: SkillPointModel port to beta [skip ci]

ryanzhoudev:
Skill Point Loadouts still don't work because TomeAnnotator isn't fixed yet. But that is something to do later.


kristofbolyai (approved):
Thank you!

PR #2650, opened 2024-07-10
fix: Fix BufferSource memory leaks

kristofbolyai:
This was actually quite stupid, this should have caused issues earlier. These buffer sources use MALLOC to allocate memory, so detecting it with regular Java Profilers was not possible before (as they generally profile heap allocations).

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/ItemScreenshotFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/ItemScreenshotFeature.java
@@ -120,8 +120,9 @@ public class ItemScreenshotFeature extends Feature {
         // draw tooltip to framebuffer, create image
         McUtils.mc().getMainRenderTarget().unbindWrite();
 
-        GuiGraphics guiGraphics =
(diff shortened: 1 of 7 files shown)


common/src/main/java/com/wynntils/features/inventory/ItemScreenshotFeature.java
```diff
@@ -141,6 +142,10 @@ private void takeScreenshot(Screen screen, Slot hoveredSlot, List<Component> ite
 
         BufferedImage bi = SystemUtils.createScreenshot(fb);
 
+        // Free the buffer source to prevent memory leaks
+        byteBuffer.close();
```
magicus, line 146: If the buffer implements Closeable you can use try-with-resources.

kristofbolyai: Yeah, I'd have to check, it might not, I'd have expected the BufferSource itself to be closable, but it is not.

PR #2652, opened 2024-07-11
fix: Improve reliability of diplomacy tracking [skip ci]

Essentuan:
These changes should improve the reliability of diplomacy tracking. It also will always update whenever the diplomacy menu is opened (mainly as a failsafe so the player can actually fix it whenever it does fail). 

I also am probably using the wrong event for `onContainerSetContent` but it seems to work fine.


kristofbolyai (approved):
Looks good, thanks.

PR #2656, opened 2024-07-13
feat: Port ProfessionModel to TextDisplay [skip ci]

kristofbolyai:


common/src/main/java/com/wynntils/models/profession/label/GatheringNodeLabelParser.java
```diff
@@ -16,12 +16,14 @@
 import net.minecraft.world.entity.Entity;
 
 public class GatheringNodeLabelParser implements LabelParser<ProfessionGatheringNodeLabelInfo> {
-    // Note: At the moment, only Dernic appends to the end of the label, but not consistently..
-    private static final Pattern GATHERING_NODE_LABEL = Pattern.compile("^§(.)(.+?)(:?\\s(Fish|Seed|Ore|Wood))?$");
+    // Note: At the moment, only "Dernic" tier does not have the types consistently as a suffix
```
magicus, line 19: Is this still the case or has Wynncraft fixed it?

kristofbolyai: Yes.

magicus: Too bad. :(

I wonder if we should compile a wish-list of minor annoyances like that, and sent them to zeer0, so maybe they can start fixing them as a background task.

PR #2657, opened 2024-07-13
feat: Port ShamanTotemModel to TextDisplay (tracking still broken) [skip ci]

kristofbolyai:
This fixes the pattern, but tracking itself is still broken, likely due to item/texture changes.


magicus:
Is the tracking fixed? Do we have a list somewhere with whatever remains to be fixed for 2.1?

PR #2662, opened 2024-07-14
fix: Fix raid tracking for 2.1

ShadowCat117:
The Canyon Colossus is not yet enabled in the beta and will need the scoreboard pattern for entering that boss fight but all the logic is there so that can be reviewed


kristofbolyai (approved):
Code looks good! Thanks.

common/src/main/java/com/wynntils/models/raid/scoreboard/RaidScoreboardPart.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public void onSegmentRemove(ScoreboardSegment segment) {
+        // Do nothing here, we will know when the raid is over from patterns
+    }
+
+    @Override
+    public void reset() {
+        // Do nothing here, we will know when the raid is over fro patterns
```
magicus, line 103: Typo...

ShadowCat117: Oopsie

PR #2664, opened 2024-07-16
feat: Improve chat tab switching hotkeys [skip ci]

girlbossdev:
Adds Ctrl+[1-9] hotkeys for switching chat tabs and changes the Shift+Tab hotkey to Ctrl+Tab instead, which is the same as most other software and also allows for Ctrl+Shift+Tab to cycle tabs backwards.

I've also added a setting to change the latter back since this seems to be a personal preference, though I've made it off by default because in my opinion it's a bit unintuitive. (I had no idea this was already implemented until I looked at the source code)

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
@@ -50,6 +50,9 @@ public class ChatTabsFeature extends Feature {
             new ChatTab("Private", false, "/r  ", Sets.newHashSet(RecipientType.PRIVATE), null),
             new ChatTab("Shout", false, null, Sets.newHashSet(RecipientType.SHOUT), null))));
 
+    @Persisted(i18nKey = "feature.wynntils.chatTabs.oldTabHotkey")
+    public final Config<Boolean> oldTabHotkey = new Config<>(false);
+
     // We do this here, and not in Services.ChatTab to not introduce a feature-model dependency.
     @SubscribeEvent(priority = EventPriority.LOWEST)
     public void onChatReceived(ChatMessageReceivedEvent event) {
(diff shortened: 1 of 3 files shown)


kristofbolyai:
So with shift tab, you get to cycle forward? I feel like that can be weird to get used to, as it cycled backwards in legacy.

Can you target the development branch? We've temporarily frozen development on main, so we can get the 2.1 beta and 1.21 versions ready.

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
```diff
@@ -125,13 +128,38 @@ public void onScreenRender(ScreenRenderEvent event) {
...
+        }
+
+        if (keyCode >= GLFW.GLFW_KEY_1 && keyCode <= GLFW.GLFW_KEY_9) {
+            ChatTab newTab = Services.ChatTab.getTab(keyCode - GLFW.GLFW_KEY_1);
+            if (newTab != null) {
+                Services.ChatTab.setFocusedTab(newTab);
+            }
+            event.setCanceled(true);
+            return;
+        }
```
kristofbolyai: I can't see where this checks for CTRL being held down. In it's current form, it works by pressing the numbers only. Do I miss something?

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
```diff
@@ -125,13 +128,38 @@ public void onScreenRender(ScreenRenderEvent event) {
...
+                    reverse = true;
+                }
+            }
+
+            if (KeyboardUtils.isControlDown() && !oldTabHotkey.get()) {
+                int newTab = reverse ? Services.ChatTab.getLastFocusedTab() : Services.ChatTab.getNextFocusedTab();
+                Services.ChatTab.setFocusedTab(newTab);
+                event.setCanceled(true);
+                return;
+            }
```
kristofbolyai: I think this could be simplified a bit, so the logic is more easily readable.

```
let newTab = -1;
if (oldTabHotkey) {
   if (shiftDown) {
      newTab = nextTab;
   }
} else {
   if (controlDown) {
      newTab = shiftDown ? lastTab : nextTab;   
   }
}

if (newTab == -1) return;

setFocusedTab + event.cancel
```

What do you think?

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
girlbossdev: Well that's embarrassing, somehow I didn't notice I managed to miss this until now

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
girlbossdev: That's better, I thought it was pretty ugly myself but I was out of ideas

girlbossdev:
> So with shift tab, you get to cycle forward? I feel like that can be weird to get used to, as it cycled backwards in legacy.

Well, this is just the current behavior. Unless I'm missing something here, I didn't mess with it.

Also, I must've gotten confused because I just noticed that the PR I mentioned introduces the same behavior I did and that it got reverted later on rather than changing it to what it currently is, so I've edited my comment now to make more sense.

common/src/main/java/com/wynntils/features/chat/ChatTabsFeature.java
```diff
@@ -50,6 +50,9 @@ public class ChatTabsFeature extends Feature {
             new ChatTab("Private", false, "/r  ", Sets.newHashSet(RecipientType.PRIVATE), null),
             new ChatTab("Shout", false, null, Sets.newHashSet(RecipientType.SHOUT), null))));
 
+    @Persisted(i18nKey = "feature.wynntils.chatTabs.oldTabHotkey")
```
kristofbolyai:
```suggestion
    @Persisted
```

No need to specify the key, unless you want something different from the default key.

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
```diff
@@ -81,6 +81,12 @@ public int getNextFocusedTab() {
         return (getTabIndex(getFocusedTab()) + 1) % getTabCount();
     }
 
+    public int getLastFocusedTab() {
```
kristofbolyai: Last is not the "opposite of" _next_. Maybe `getPreviousFocusedTab`?

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
```diff
@@ -81,6 +81,12 @@ public int getNextFocusedTab() {
...
+        int tabCount = getTabCount();
+        int lastTab = getTabIndex(getFocusedTab()) - 1;
+        return lastTab == -1 ? tabCount - 1 : lastTab;
```
kristofbolyai:
```suggestion
    int tabIndex = getTabIndex(getFocusedTab());
    return (tabIndex - 1 + getTabCount()) % getTabCount();
```

I think this expresses the edge-case where only 1 tab is in the list better (and since you are already changing the name, this is not a "hassle").

kristofbolyai (approved):
Really close now, I have 2 minor change requests just for code consistency :tada:

common/src/main/java/com/wynntils/services/chat/ChatTabService.java
girlbossdev: Sounds better to me too, yeah

kristofbolyai (approved):
Thanks :)

PR #2665, opened 2024-07-19
fix: Fix everything lootrun related for 2.1

kristofbolyai:
Brings back lootrun and beacon model to 2.0 behavior. Beacon model is slightly too integrated with lootruns to my liking, but it was basically like this previously.


ryanzhoudev (approved):

common/src/main/java/com/wynntils/models/lootrun/LootrunModel.java
```diff
@@ -398,7 +398,7 @@ public void onLootrunCompleted(LootrunFinishedEvent.Completed event) {
     @SubscribeEvent
     public void onWorldStateChanged(WorldStateEvent event) {
         // The world state event is sometimes late compared to lootrun events (beacons, scoreboard)
-        // Reseting once when leaving the class is enough
+        // Resting once when leaving the class is enough
```
magicus, line 401: Should maybe have been "Resetting"?

kristofbolyai: Yep, I didn't see that😅

PR #2676, opened 2024-07-27
fix: Fix SpellCastVignetteFeature bugging out when a spell's mana cost is 0 [skip ci]

kristofbolyai:


magicus:
Just curious, when can the mana cost be 0? Does that mean you can just spam the spell without end?

kristofbolyai:
> Just curious, when can the mana cost be 0? Does that mean you can just spam the spell without end?

The ability tree has options for 0 mana cost spells, they usually either require spell cycling (basically, cast 3 spells, get the 4th for free 😅, or special cases, like, spend health instead of mana, if you are out).

PR #2681, opened 2024-07-27
fix: Fix activity parsing (by changing CustomColor#fromInt alpha behavior)

kristofbolyai:


fabric/src/test/java/TestCustomColor.java
```diff
@@ -90,4 +90,26 @@ public void customColor_fromRGBWorks() {
...
+        CustomColor result = CustomColor.fromInt(color);
+
+        Assertions.assertEquals(expected, result, "CustomColor#fromInt() did not return the expected value");
+    }
+
+    @Test
+    public void customColor_fromIntWithAlphaWorks() {
+        final int color = 0xcc00aabb;
+
+        final CustomColor expected = new CustomColor(0, 170, 187, 204);
```
magicus, line 109: Why is the expected alpha value 0 and not 0xcc?

kristofbolyai: The alpha is the 4th int in the constructor, while the int is in 0xaarrggbb format. So 0xCC = 204

magicus: Ah, I see. Thanks!

I just assumed it was the first... Having the hard-coded values in hex instead of decimal might have made it easier to read; but I'm not going to request a separate change for that.

kristofbolyai: The tests are for testing integer to color conversion, if anything, writing the expected result in hex could be better, although hex is #rgba, so it's still not the 0xargb format.

PR #2687, opened 2024-07-28
feat: Add better support for confirmationless dialogues

kristofbolyai:
This PR should make non-standard NPC messages be parsed in confirmationless dialogues, which should provide a better overall dialogue support, with less flickering, and better translations.

This PR was tested with two quests, namely _Hollow Serenity_ and _Hero of Gavel_. Both of these quests have a wide variety of dialogues, and from my testing, there are no regressions. Hollow Serenity has some specific issues with selection dialogues with some NPCs, however they are not regressions of this PR, so no change was made to better support them.

Reviewing this PR is hard, as the dialogue system is complex with many edge-cases, so finding faults in logic is basically impossible. For this reason, I ask that reviewers try the new parsing, and let me know about their findings. Obviously, we can't test everything, so if we only look at a few quests, but we don't find any _new_ issues, then we may very well release this to the public.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
@@ -94,7 +94,8 @@ public final class ChatHandler extends Handler {
     private String oneBeforeLastRealChat = null;
 
     private long lastSlowdownApplied = 0;
-    private List<StyledText> lastScreenNpcDialog = List.of();
+    private List<StyledText> lastScreenNpcDialogue = List.of();
+    private StyledText lastConfirmationlessDialogue = null;
     private List<StyledText> delayedDialogue;
     private NpcDialogueType delayedType;
     private long chatScreenTicks = 0;
@@ -108,7 +109,8 @@ public final class ChatHandler extends Handler {
         lastRealChat = null;
         oneBeforeLastRealChat = null;
         lastSlowdownApplied = 0;
-        lastScreenNpcDialog = List.of();
+        lastScreenNpcDialogue = List.of();
+        lastConfirmationlessDialogue = null;
         delayedDialogue = null;
         delayedType = NpcDialogueType.NONE;
     }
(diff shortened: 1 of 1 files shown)


ShadowCat117 (approved):

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -282,6 +331,16 @@ private void processNewLines(LinkedList<StyledText> newLines) {
             // This is an NPC dialogue screen.
             // First remove the "Press SHIFT/Select an option to continue" trailer.
             newLines.removeFirst();
+
+            // If this happens, the "Press SHIFT/Select an option to continue" got appended to the last dialogue
+            // NOTE: Currently, we do nothing in this case, as it seems to work without any issues
```
magicus, line 336: Did you encounter any situation where a confirmationless dialogue was turned into a confirmation-requiring one?

kristofbolyai: Yes, multiple ones.

magicus: Interesting. Do you remember any specific situation (a quest or some other instance)? I never noticed this, but maybe it is a new thing in Wynncraft as well. Or I just did not look around in enough places...

kristofbolyai: [Kooha-2024-07-28-10-43-39.webm](https://github.com/user-attachments/assets/186c2dbe-dfd9-4c37-9aba-8dc78d0401ef)

I only have a video of these special confirmationless dialogues, that are basically time or distance based. `Hero of Gavel` has quite a few dialogues that are regular NPC dialogues, they start out as confirmationless, and they turn out to be SHIFT to continue (with a time based auto-continue in most cases).

magicus: I need to go and check that in play myself. I wonder if I have a character with enough levels that have not done Hero of Gavel...

kristofbolyai: Iirc Hollow Serinity has them tok and it's way less levels (but Hero of Gavel is much better for testing). Hollow Serentiy however has NPCs with special selection dialogues that are still broken :)

magicus: So basically what you are saying is that if/when I get the time to work with the dialogue system again, I should replay Hollow Serenity and fix so it works, and then I also get the chance to check out these transmogrifying dialogues. :) I'll bear that in mind.

PR #2688, opened 2024-08-01
fix: Fix character ID parsing for 2.1

ShadowCat117:
image
image


kristofbolyai (approved):
Thank you! I assume this is now live on beta, so this is not theoretical only. If so, please merge this :)

ShadowCat117:
Yup, tried it on all the available servers on the beta so seems to be fully live

PR #2692, opened 2024-08-03
fix: Don't spam tracker update sound when tracking world events [skip ci]

ShadowCat117:


kristofbolyai (approved):

magicus:
I have not played on the beta, but do we need some kind of countdown timer to these world events? Maybe it would be nice to have an audio cue when there is like 5 min, 1 min and 15 seconds left?

ShadowCat117:
> I have not played on the beta, but do we need some kind of countdown timer to these world events? Maybe it would be nice to have an audio cue when there is like 5 min, 1 min and 15 seconds left?

When tracked, the name of the event is like "Prelude to Annihilation (5m left)" so that displays a countdown. I think(?) it also sends chat messages as the timer gets lower

PR #2693, opened 2024-08-03
fix: Fix player nametags rendering incorrectly [skip ci]

ShadowCat117:
Current 
image
Fixed
 image-1


kristofbolyai:
Just to verify, should(n't) the first one also be negative? You can confirm it by looking at the original nametag rendering source code. Don't waste too much time on it though; if you can't find it, I think I remember where I can.

ShadowCat117:
It was removed in the 1.21 port https://github.com/Wynntils/Artemis/commit/b12ee1563b382fc73e4d71da1e3997de4c8c4584

kristofbolyai (approved):
Should work fine then :)

PR #2701, opened 2024-08-05
feat: Add MappingProgressFeature (for development purposes) [skip ci]

kristofbolyai:


magicus:
Ooooh! ✨

ShadowCat117:
This could probably be reworked into the hiding unexplored areas feature that's been highly requested

kristofbolyai:
> Ooooh! ✨ 

I've not added a picture for the PR, but here is one from discord (after I mapped everything for the next patch). image

kristofbolyai:
> This could probably be reworked into the hiding unexplored areas feature that's been highly requested

I believe mapdata bounding shapes, aka. mapareas should be used for those. After 1.21/2.1 is done, we should continue work on Mapdata.

magicus:
> I believe mapdata bounding shapes, aka. mapareas should be used for those. 

Well, yes and no. I agree that we should expose the "discovered area" information using the mapdata system, but theMapAreas are utterly untested and are currently mostly a sketch, rather than a working implementation. As I imagined them, I though of something like generalized territories, basically polygons with a limited number of edges (4 in the common case of rectangles). Maybe a bit more in e.g. the seaskipper destination maps, to have target shapes that somewhat better match the island shapes than pure rectangles.

The data you collect here is more "pixel-like" than "vector-like". Converting that to an enclosing polygon seems computationally intensively and possible pointless. Maybe we need a fourth option, like "MapMesh" or whatever, that is more like a bitmask of included/excluded cells.

PR #2703, opened 2024-08-05
feat: Add ModMenu support for Fabric and configuration screen for NeoForge [skip ci]

ShadowCat117:


fabric/build.gradle
```diff
@@ -25,11 +25,19 @@ repositories {
...
+        url "https://maven.terraformersmc.com/"
+    }
 }
 
 dependencies {
     modImplementation "net.fabricmc:fabric-loader:${fabric_loader_version}"
     include(modImplementation(fabricApi.module("fabric-resource-loader-v0", fabric_version)))
+    include(modImplementation(fabricApi.module("fabric-screen-api-v1", fabric_version)))
+    include(modImplementation(fabricApi.module("fabric-key-binding-api-v1", fabric_version)))
+    include(modImplementation(fabricApi.module("fabric-lifecycle-events-v1", fabric_version)))
```
kristofbolyai: Are these actually needed? From what I understand, modmenu is optional. _If modmenu is present_, then these become dependencies of modmenu, not us.

magicus: I think you are correct. All that is needed for ModMenu is to provide a proper implementation of its `ModMenuApi` class.

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
```diff
@@ -255,6 +259,7 @@ protected void doInit() {
 
     @Override
     public void doRender(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick) {
+        this.renderBackground(guiGraphics, mouseX, mouseY, partialTick);
```
magicus: Why is this change necessary?

ShadowCat117: Otherwise, the modmenu background is still visible and creates some visual issues. Can't call the super render method either as that also causes issues so just need the background

kristofbolyai: What if we override renderBackground so it only renders a background if "McUtils.mc().level" is null?

ShadowCat117: https://github.com/Wynntils/Artemis/pull/2703/commits/5223418af691789fb205d36553bf12fe177f1d57 Like this?

kristofbolyai (approved):

common/src/main/java/com/wynntils/screens/settings/WynntilsBookSettingsScreen.java
kristofbolyai: Yes

PR #2705, opened 2024-08-06
feat(major): Release Wynntils 2.0 (Wynncraft 2.1 + MC 1.21 - Fabric/NeoForge)

kristofbolyai:
This should not be merged before Wynncraft goes down for maintenance on August 10th.


magicus:
Should not the PR title be `feat!: ....`? From the user PoV, this is where all the goodies are coming in; chore is supposed to be behind-the-scenes work...

kristofbolyai:
It's a chore! so it's a major bump (because we bump MC version), while feat! is only a minor bump.

magicus:
Wait... do `chore!` imply major bump? How/when did we decide that was a good idea? 🤔

kristofbolyai:
> Wait... do `chore!` imply major bump? How/when did we decide that was a good idea? 🤔 

It's either chore! or chore(bump-mc). It can really be anything in the commit, but when I was reworking CI versioning, I opted to use this.

magicus:
Can we change that to e.g. `feat!!` before we release 2.0? I'd hate it for our changelog for 2.0 to go out as a "chore" to end users...

magicus:
Or will we concatenate all individual commits that make up this PR when creating the changelog, so the "version bump" itself just becomes a minor item on the list? If so, a `chore` is okay.

kristofbolyai:
I can change the system, I think.

PR #2709, opened 2024-08-08
feat: Localize Wynntils to 13 languages using GPT4-o (hu, nl, pl, es, fr, de, it, pt, ru, ja, ko, zh_CN, zh_TW)

kristofbolyai:
This PR also does changes font rendering, replacing scissor testing with stencil masks, as stencils work way better, and are PoseStack compatible.

The language files had slight modifications done. Mainly, `Á` was replaced with an alternative Unicode character (as currently, in the Wynn texture pack, this font is a big black box).


common/src/main/resources/assets/wynntils/lang/de_de.json
```diff
@@ -1,47 +1,2655 @@
...
+  "core.wynntils.category.commands": "Befehle",
+  "core.wynntils.category.debug": "Debuggen",
+  "core.wynntils.category.embellishments": "Verzierungen",
   "core.wynntils.category.inventory": "Inventar",
   "core.wynntils.category.map": "Karte",
   "core.wynntils.category.overlays": "Overlays",
   "core.wynntils.category.players": "Spieler",
-  "core.wynntils.category.redirects": "Umleitungen von Spielbenachrichtigungen",
-  "core.wynntils.category.tooltips": "Gegenstand Tooltips",
+  "core.wynntils.category.redirects": "Spielbenachrichtigungs-Weiterleitungen",
```
magicus, line 65: I'm not sure GPT does this better than a human translator. Look at this example, for instance.

magicus:
I think you pushed this very fast, without discussion. :-( I have been objecting to machine translation before. 

I agree that it is reasonable to assume that GPT4 might do a better job than previous machine translation services, but I still would like a better assessment on the quality of the translations. Especially when it replaces a human-made translation, but also when adding new translations -- I don't know about you, but I've encountered way to many bad translations to Swedish where the quality is so inferior that it would have been better not even to pretend it was translated.

kristofbolyai:
We did discuss this on Discord for a day or two, we did some refinements to the GPT prompt. We've also looked into our own languages, hungarian for me, dutch for Chromium and we found it reasonable.

I'd also like to think that anyone, with a decent enough English knowledge (and will to learn) opts for the English locale in-game. This is mostly for those specific users, who are having a really hard time with english (like russians, japanese, chinese and polish people, we generally get the most amount of requests from them).

magicus:
Can you add a Swedish translation as well, so I can check it?

kristofbolyai:
> Can you add a Swedish translation as well, so I can check it?

I don't have enough credit left at the moment :sweat:. The cost is not really high, the 13 languages here was around 7-8 dollars (not counting the bad attempts before we got a stable system working). I'll likely consult Scyu to open up an "official" Wynntils Org in OpenAI, and we can use that from now on (even for NPC dialogue translations, at some point). I'll get back to you!

PR #2714, opened 2024-08-09
refactor: Java 21 related refactors [skip ci]

kristofbolyai:
This should fix an actual bug in ChatHandler where the comparison between a holder and a normal object was wrong. Other than that, these are mostly Java 21 changes.

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
@@ -150,7 +150,7 @@ public final class ChatHandler extends Handler {
     public void onStatusEffectUpdate(MobEffectEvent.Update event) {
         if (event.getEntity() != McUtils.player()) return;
 
-        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN
+        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN.value()
                 && event.getEffectAmplifier() == 3
                 && event.getEffectDurationTicks() == 32767) {
             if (delayedDialogue != null) {
@@ -168,7 +168,7 @@ public final class ChatHandler extends Handler {
     public void onStatusEffectRemove(MobEffectEvent.Remove event) {
         if (event.getEntity() != McUtils.player()) return;
 
-        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN) {
+        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN.value()) {
             lastSlowdownApplied = 0;
         }
     }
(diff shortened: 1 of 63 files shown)


kristofbolyai:
I believe you might also be interested in this, @magicus, although, if you won't have time, I'll merge this before the 10th.

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
```diff
@@ -150,7 +150,7 @@ public void onSystemChatReceived(ChatPacketReceivedEvent.System event) {
     public void onStatusEffectUpdate(MobEffectEvent.Update event) {
         if (event.getEntity() != McUtils.player()) return;
 
-        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN
+        if (event.getEffect() == MobEffects.MOVEMENT_SLOWDOWN.value()
```
magicus: Is this a bug fix? I can't see how both the code before and after could be valid, so it's not a refactoring.

kristofbolyai: It's the fix mentioned in the PR/commit description.

common/src/main/java/com/wynntils/screens/itemsharing/SavedItemsScreen.java
```diff
@@ -270,10 +270,8 @@ public boolean mouseClicked(double mouseX, double mouseY, int button) {
                         (int) (categoryRenderX + 97),
                         (int) categoryRenderY,
                         (int) (categoryRenderY + 8))) {
-            if (!editingCategory) {
-                editingCategory = true;
-                addCategoryInput();
-            }
+            editingCategory = true;
+            addCategoryInput();
```
magicus, line 274: I don't have the context, but I'm noting tha tthis will make `addCategoryInput` being unconditionally called. Is this okay?

kristofbolyai: If you check the if that this code is wrapped in, you can see it was a duplicate condition.

common/src/main/java/com/wynntils/handlers/chat/ChatHandler.java
magicus: I juuuust noticed and was about to reply to myself. :)

magicus:
Okay. 

So, the changes themselves look good. However, the PR as a whole is perhaps not ideal.

First of all, the title is confusing. In what way is this "suggestions from the Artemis Mini profile"? My best guess is that you are digging out some changes from a branch that tried to implement profiles. But the origin of the code is not really relevant (at least not in this case), so a better title would be "Java 21 refactorings" or something like that.

And that would also highlight the dual nature of this PR. It contains a ton of systematic refactorings, and a small bug fix hidden away in the middle. I don't like that kind of PRs. Please separate out the bug fix into a separate PR. In general, if you have like 50% refactoring and 50% bug fixing, and the refactoring was needed to fix the bug, then it is okay to mix them. But when you have huge refactoring-only commits, sneaking in anything else in them is not a good idea, since it is hopeless to track changes properly.

That would further allow you to use the `refactor` and `fix` PR title prefixes correctly, instead of lumping it under the misc bin `chore`. (Which I tbh think has been a bit over-used in Wynntils; it is supposed to be stuff like bumping versions of dependencies, or fixing build scripts.)

kristofbolyai:
We've used to use this commit time for a while now. Artemis Mini is an inspection profile we've committed a while back. I'll separate the fix from the refactors though.

PR #2718, opened 2024-08-09
fix: Change ReplaceRecipeBookFeature to DisableRecipeBookFeature [skip ci]

kristofbolyai:


magicus:
Do we want to add a new GUI element to be able to open the item guide from the inventory?

kristofbolyai:
Maybe eventually yes, I just wanted to get rid of this “bug” for now.

PR #2723, opened 2024-08-10
fix: Fix widgets with scrolling text not being masked [skip ci]

ShadowCat117:


kristofbolyai (commented):
Why use scissor instead of stencil masks? I've recently made a change for FontRenderer to use stencil masks instead of scissor, because it didn't work correctly in the settings screen (with longer, localized text, the reset button was empty).

ShadowCat117:
Yeah that change isn't playing nice with the masks for smooth scrolling, causes this issue that's why I changed it to scissor
image

kristofbolyai (approved):
Sure, let's go with this. I believe scissors are better used here, than in text rendering.

PR #2730, opened 2024-08-11
fix: Fix tutorial content book

ShadowCat117:
Value is 16 on normal content books, checked on 2 separate characters.


kristofbolyai (commented):
Doesn't this soft-crash if the item is null, or if the item doesn't have a custom model data?

ShadowCat117:
> Doesn't this soft-crash if the item is null, or if the item doesn't have a custom model data?

Possibly, I assume the book should always be in that slot but I can add a check

kristofbolyai (approved):

PR #2735, opened 2024-08-12
feat: Bulk buy rework

ryanzhoudev:
https://github.com/user-attachments/assets/c6eb70a3-82a7-42d8-ab90-76df122186ac

https://github.com/user-attachments/assets/51554e05-08cc-4127-b442-8cea0b74daef


ShadowCat117:
Resizing window causes the widget to dissapear, not sure you can use the SetSlotEvent for this

ShadowCat117:
Once you fix that though I think it looks good!

common/src/main/java/com/wynntils/features/ui/BulkBuyFeature.java
```diff
@@ -13,46 +13,161 @@
...
+
+        initBulkBuyWidget(McUtils.mc().screen);
+    }
+
+    @SubscribeEvent
+    public void onContainerOpened(ScreenInitEvent.Post e) {
+        // This event handles any reinit required, eg window resize
+        // It will not make a widget on first init
+        initBulkBuyWidget(e.getScreen());
+    }
```
ShadowCat117, line 84: If you only need this for any init after the first, can you not use `isFirstInit()` from the existing ScreenInitEvent and not create the widget if it is?

ryanzhoudev: Can't, the .Pre has the old screen coordinates, so when I do screen.topPos and .leftPos and the screen has been resized, it'll be in a weird position

ShadowCat117: Ah I see

ShadowCat117 (approved):
Looks good to me!

kristofbolyai:
Changing the ScreenEventInit is a huge thing, so I'll have to review this when I'm not this sleepy. In the meantime, let's also ask @Essentuan whether fuygg will break because of this event change or not, as it's pretty significant.

ryanzhoudev:
The summary for the ScreenInit stuff is the previous usages of it are all .Pre, for which the mixins are identical and not changed. Same functionality, different name. The only new part is .Post, which sends the event after draw.

Essentuan:
> Changing the ScreenEventInit is a huge thing, so I'll have to review this when I'm not this sleepy. In the meantime, let's also ask @Essentuan whether fuygg will break because of this event change or not, as it's pretty significant.

Nope you guys are good, fuy.gg only uses MenuOpenedEvent

ryanzhoudev:
Side effect of the overlap fix is that the bank inventory emerald count gap is also fixed

common/src/main/java/com/wynntils/mc/event/ScreenInitEvent.java
```diff
@@ -26,4 +26,17 @@ public Screen getScreen() {
...
         return firstInit;
     }
+
+    public static class Pre extends ScreenInitEvent {
+        public Pre(Screen screen, boolean firstInit) {
+            super(screen, firstInit);
+        }
+    }
+
+    // FIXME: This event does not trigger on first init. See the section near ScreenMixin.java L61
```
kristofbolyai: I don't understand why this is really an issue. We already have a "Pre" mixin, can't you just copy it, and shift it "after" the target?

common/src/main/java/com/wynntils/features/ui/BulkBuyFeature.java
```diff
@@ -13,46 +13,161 @@
...
+
+    private void initBulkBuyWidget(Screen screen) {
+        // Neither event guarantees that they do not re-fire on the same screen
+        if (bulkBoughtSlotNumber != -1) return;
+
+        if (!(screen instanceof ContainerScreen containerScreen)) return;
+        if (!(containerScreen.getMenu() instanceof AbstractContainerMenu acm)
+                || acm.getItems().size() != 90) return;
```
kristofbolyai: You know the drill, multi-line ifs have brackets. I assume this was spotless.

common/src/main/java/com/wynntils/features/ui/BulkBuyFeature.java
```diff
@@ -13,46 +13,161 @@
...
-            }
+        if (bulkBoughtSlotNumber == -1) {
+            // we're starting a new bulk buy
+            bulkBoughtSlotNumber = e.getSlotNum();
+            bulkBoughtContainer = container;
+            bulkBoughtAmount = bulkBuyAmount.get();
+            bulkBoughtPrice = itemPrice;
+
+            bulkBuyWidget.setBulkBoughtPrice(bulkBoughtPrice);
+            bulkBuyWidget.setBulkBoughtItemStack(itemStack);
```
kristofbolyai: Maybe introduce a `BulkBuyInfo` record to send all of the data at once?

common/src/main/java/com/wynntils/mc/event/ScreenInitEvent.java
ryanzhoudev: I did try to use the @ return of the same method, but it has the same effect as the Pre mixin. The screen's coordinates are still the ones from before the re-init.

common/src/main/java/com/wynntils/features/ui/BulkBuyFeature.java
ryanzhoudev: Ah yea

common/src/main/java/com/wynntils/features/ui/BulkBuyFeature.java
ryanzhoudev: I did originally have something like this, but removed it since there are only two instances where I send the price and itemstack together. Do you still think it is worth it?

kristofbolyai: I do think it's worth it, it makes sure you can't missmatch and forget to update part of the data.

kristofbolyai:
What if the bulk buy UI part only appeared when an active buy is happening? I think it's taking up a lot of space unnecessarily. I understand that the original idea had use cases for this UI part, but in this current form (limited by rules), it's not needed.

ryanzhoudev:
> What if the bulk buy UI part only appeared when an active buy is happening? I think it's taking up a lot of space unnecessarily. I understand that the original idea had use cases for this UI part, but in this current form (limited by rules), it's not needed.

I gave this a shot and I'm not a huge fan of such a large UI element popping and flashing in and out like that. Especially if the user buys a block of 4, then a different item block of 4, and such. I'll see if I can do something else with this

PR #2737, opened 2024-08-12
feat: Add world event container as a reward container [skip ci]

ShadowCat117:


ShadowCat117:
Also will note here, I believe the Annihilation world event gives a special reward chest, if someone manages to add that that'd be great as I will likely miss the first one

kristofbolyai (approved):

PR #2738, opened 2024-08-13
fix: Fix TradeMarketPriceConversionFeature [skip ci]

ShadowCat117:


common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceConversionFeature.java
```diff
@@ -45,13 +44,4 @@ public void onClientChat(ChatSentEvent event) {
...
     }
-
-    @SubscribeEvent
-    public void onGuiOpen(ScreenOpenedEvent.Post event) {
-        if (TRADE_MARKET_PATTERN
-                .matcher(event.getScreen().getTitle().getString())
-                .matches()) {
-            shouldConvert = false;
-        }
-    }
```
ShadowCat117, line 56: You are no longer able to open the tm until inputting something so I presume this is no longer needed

common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceConversionFeature.java
```diff
@@ -10,26 +10,25 @@
...
 @ConfigCategory(Category.TRADEMARKET)
 public class TradeMarketPriceConversionFeature extends Feature {
-    private static final Pattern PRICE_PATTERN =
-            Pattern.compile("^§6Type the price in emeralds or type 'cancel' to cancel:$");
-    private static final Pattern TRADE_MARKET_PATTERN = Pattern.compile("^What would you like to sell\\?$");
-    private static final Pattern CANCELLED_PATTERN = Pattern.compile("^You moved and your chat input was canceled\\.$");
+    private static final Pattern PRICE_PATTERN = Pattern.compile(
+            "^§5(\uE00A\uE002|\uE001) \n\uE001 Type the price in emeralds or type \n\uE001 'cancel' to cancel:\n\uE001 ");
+    private static final Pattern CANCELLED_PATTERN =
+            Pattern.compile("^§4(\uE008\uE002|\uE001) You moved and your chat input was canceled.$");
```
kristofbolyai, line 22: We really should be able to strip these prefixes with RecipientType/StyledTextUtils soon, without adding them to every regex.

kristofbolyai (approved):

common/src/main/java/com/wynntils/features/trademarket/TradeMarketPriceConversionFeature.java
kristofbolyai: In fact, if you want to take this up, maybe it'd be the best. Although @DonkeyBlaster also mentioned wanting to work on RecepientTypes. We can not only easily parse the RecipientTypes from the headers, you can also easily get rid of them, as `StyledTextUtils#unwrap` already exists, you'd only need to add a method for removing the initial header.

ShadowCat117: Sure I can probably take a look this weekend if it hasn't already been done/started

PR #2750, opened 2024-08-16
fix: Totem tracking doesn't bind timers [skip ci]

ryanzhoudev:
So I think the problem was that wynn spawns two TextDisplays now, one of them is empty and the other is the timer/hpr one. Since they're both unnamed at spawn and receive a name after one update, we can just check them during the rename. Doesn't fix the time rift problem yet, but one at a time right?


kristofbolyai (approved):
Looks good. Perhaps release this with the party list fix? Friend messages seem to get parsed.

PR #2753, opened 2024-08-16
feat: Show waypoint names in world [skip ci]

Rafii2198:
This is probably my biggest PR I have ever made, so I suspect there will be something wrong with it.
This began as an accessibility setting for lootrun beacons but became more broad.

This is how it looks in game, everything is optional and off by default
image
image
image


kristofbolyai:
I'll review this tomorrow or the day after. Thanks.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1362,6 +1366,8 @@
   "feature.wynntils.worldWaypointDistance.name": "Waypoint Distance",
   "feature.wynntils.worldWaypointDistance.scale.description": "How large should the marker be?",
   "feature.wynntils.worldWaypointDistance.scale.name": "Scale",
+  "feature.wynntils.worldWaypointDistance.showAdditionalTextInWorld.description": "Wether the name of compass traget should be displayed",
```
ShadowCat117:
```suggestion
  "feature.wynntils.worldWaypointDistance.showAdditionalTextInWorld.description": "Whether the name of compass target should be displayed",
```

Essentuan:
Would it not be better to put the text under the icon? I feel like its just gonna be significantly slower to 'read' the waypoint.

Rafii2198:
I think top is better, these are only accessibility options that are disabled by default, they are more like a title of the icon which I think it makes more sense to put on top, tho I of course can change it if devs want it

Essentuan:
Two things 

A) This definitely should be enabled by default, helping locate things faster is always a good thing (and its not like this massively increases visual clutter)
B) Humans will always read text if presented with it. I honestly do not see why you would wanna make the player spend more time moving their eyes to read everything. (Personally I think most people would say moving lower would look better anyway)

Essentuan:
For context this is how it looks with fuygg

image

ShadowCat117:
I don't agree that they should be enabled by default.

For the lootrun markers, most people aren't colourblind so they won't need the text colour, they can get the colour from the existing text.
For the activity, the name doesn't really have any use. You can see the name on the content tracker overlay or scoreboard and as it uses the quest scroll icon you know it's for tracked activity.
And for the other markers, like fast travel, that's probably only something newer players who are unfamiliar with the icons would want

Essentuan:
Why does it matter if people don't 'need' it? Sure people can (and do) function without it, it just a little thing that goes a long way with helping people? Plus like I said there is no downside to keeping it enabled by default.

ShadowCat117:
As for top vs bottom, I also don't agree with it being on the bottom. Them being separated helps you to distinguish between the distance text and the name. If you're just looking for the purple lootrun beacon you look above, if you care about the distance you look below

Essentuan:
> Why does it matter if people don't 'need' it? Sure people can (and do) function without it, it just a little thing that goes a long way with helping people? Plus like I said there is no downside to keeping it enabled by default.

For example its not like its easy to tell at a glance which beacon is what color in lootruning. I suspect most people will enable it anyway.

Essentuan:
> As for top vs bottom, I also don't agree with it being on the bottom. Them being separated helps you to distinguish between the distance text and the name. If you're just looking for the purple lootrun beacon you look above, if you care about the distance you look below

This is just... wrong??? I've and many other people have done so many wars with the fuy.gg territory markers and never have I once been unable to quickly distinguish the two. They're on separate lines its nearly impossible to distinguish them.

ShadowCat117:
> Why does it matter if people don't 'need' it? Sure people can (and do) function without it, it just a little thing that goes a long way with helping people? Plus like I said there is no downside to keeping it enabled by default.

Wynntils already gets a lot of criticism for having a "cluttered" UI which I do somewhat agree with, we do have a lot showing on the screen and I don't think this is a large enough help to the vast majority of players to be on by default. I can guarantee if it was on by default, we'd get multiple people in discord asking how to hide it

Essentuan:
> > Why does it matter if people don't 'need' it? Sure people can (and do) function without it, it just a little thing that goes a long way with helping people? Plus like I said there is no downside to keeping it enabled by default.
> 
> Wynntils already gets a lot of criticism for having a "cluttered" UI which I do somewhat agree with, we do have a lot showing on the screen and I don't think this is a large enough help to the vast majority of players to be on by default. I can guarantee if it was on by default, we'd get multiple people in discord asking how to hide it

The cluttered UI issues are mostly from wynntils replacing things that don't need to be replaced (pretty much all of the gameplay related bars are just much worse than just leaving them off). And at least personally that also applies to the objective bars, taking up more space for the same functionality just seems weird to me. And also this doesn't really create visual clutter? Its one extra line.

magicus:
Nice addition @Rafii2198, this looks really great. I will have a look at the patch as soon as I can.

kilieli:
> > As for top vs bottom, I also don't agree with it being on the bottom. Them being separated helps you to distinguish between the distance text and the name. If you're just looking for the purple lootrun beacon you look above, if you care about the distance you look below
> 
> This is just... wrong??? I've and many other people have done so many wars with the fuy.gg territory markers and never have I once been unable to quickly distinguish the two. They're on separate lines its nearly impossible to distinguish them.

I do have to agree with Essentuan here, the fuy.gg way of doing it has worked very well, is readable at a single glance, and looks more elegant than having the text split over multiple lines with a graphic in between. With split text, the different outlines of the icons can make the top text seem like it has different spacing from the bottom text and icon itself, which then impacts readability and the overall look negatively.

magicus:
@Essentuan @kilieli The ultimate decision on how this feature will be implemented lies with the Wynntils developers. Everyone is free to add suggestions on Github, but 8 posts just arguing for design change is not helping that cause. 

Give us space to implement the feature as we find best. If it still does not work out for you, feel free to add a suggestion for improvement in our Discord.

Essentuan:
> @Essentuan @kilieli The ultimate decision on how this feature will be implemented lies with the Wynntils developers. Everyone is free to add suggestions on Github, but 8 posts just arguing for design change is not helping that cause.
> 
> Give us space to implement the feature as we find best. If it still does not work out for you, feel free to add a suggestion for improvement in our Discord.

Im trying to get across how the community feels about Wynntils (and how it designs ui elements). Nobody really talks about it in wynntils discord because they can't be bothered. But I've spoken to a lot of people and have specifically asked. So far a decent chunk of people want to either stop or have stopped using Wynntils.

Essentuan:
Ok I've been asking other people so far everyone has agreed that text on the bottom looks better. And ~50% of people have agreed it should be enabled by default. Imo this is a good reason *to* enable it by default. Since people who need it are unlikely to know it exists without being told and its not like everyone dislikes it.

kristofbolyai:
What if placement is a config? It wouldn't be too hard.

Secondly, this is kind of a mix between accessibility and not. The lootrun tasks should be "Task Name\n(Color)" to match the other icons' descriptions, which are less accessibility focused.

Rafii2198:
The commit names should be pretty self-explanatory, this is now how it looks (the scale is set to 10)
image

But there is one issue that I don't get so I need a bit of help with, the text for tracked content specifically displays the distance 2 times for some reason, only tracked content, everything else is fine. I tried to find the cause but I just can't
image

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
```diff
@@ -35,7 +36,8 @@ public void setSpawnLocation(Location spawnLocation) {
                                 Texture.QUESTS_SCROLL_ICON,
                                 CommonColors.WHITE,
                                 CommonColors.WHITE,
-                                CommonColors.WHITE),
+                                CommonColors.WHITE,
+                                null),
```
kristofbolyai: Maybe put `ACTIVITY_LOCATION_NAME` here, if the config is enabled?

common/src/main/java/com/wynntils/models/marker/UserWaypointMarkerProvider.java
```diff
@@ -24,7 +24,13 @@ public class UserWaypointMarkerProvider implements MarkerProvider<WaypointPoi> {
...
     public void addLocation(Location location, Texture texture, CustomColor beaconColor, CustomColor textColor) {
         addLocation(new MarkerInfo(
-                "Waypoint", new StaticLocationSupplier(location), texture, beaconColor, textColor, CommonColors.WHITE));
+                "Waypoint",
+                new StaticLocationSupplier(location),
+                texture,
+                beaconColor,
+                textColor,
+                CommonColors.WHITE,
+                ""));
```
kristofbolyai: Why "" when the other instances have null?

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
```diff
@@ -411,7 +412,13 @@ public boolean doMouseClicked(double mouseX, double mouseY, int button) {
...
                         } else {
                             Models.Marker.USER_WAYPOINTS_PROVIDER.addLocation(
-                                    new Location(hovered.getLocation()), iconPoi.getIcon());
+                                    new Location(hovered.getLocation()),
+                                    iconPoi.getIcon(),
+                                    Managers.Feature.getFeatureInstance(WorldWaypointDistanceFeature.class)
+                                                    .showAdditionalTextInWorld
+                                                    .get()
+                                            ? hovered.getName()
+                                            : null);
```
kristofbolyai: Shouldn't this logic be applied in UserWaypointsProvider so it applies to all custom pois?

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
Rafii2198: Well, then it will show "Activity Location" text overlapped with name of the activity. Tho I think I get it now, it is like placing 2 different markers in the same place, one is coming from `setSpawnLocation` and second `setTrackedActivityLocation`. Is there a necessity for it to do that?
image

kristofbolyai: We set duplicated markers? That's a separate issue.. I guess ignore this change, and we'll fix it eventually.

common/src/main/java/com/wynntils/screens/maps/MainMapScreen.java
Rafii2198: So if I get this right, the right way would be to change all `null` values I put in UserWaypointsProvider to this check?

Rafii2198:
So I did the changes, and it seems everything works well, including custom pois, territory names and so on (first image) as well as the compass command (second image)
image
image
I would assume the guild wars ones too show it, but I can't easily test that.

Tho with this PR I could at most "mask" the second Activity marker, it looks normal but when you look deeper into it, you will notice that there are actually 2 of them, with the second one having empty text and overlaid distance which makes the background darker. You can notice that darker background on live version too when you put manual marker next to activity one.
image

And there is one more bug related to Activity markers, tho I assume it is again something out of the scope of this PR, when you set activities from the vanilla book for the ones we don't have in our menus, the name either won't be there or will copy the previous one if that one was the one we have support for
image

Rafii2198:
Actually wait, the second issue is due to `setSpawnLocation` method which for some reason returns the name of previous content but with the new content location (?) ~~I don't think I can easily mask that actually~~ Actually I had to just think lol

common/src/main/java/com/wynntils/models/activities/markers/ActivityMarkerProvider.java
```diff
@@ -35,7 +36,12 @@ public void setSpawnLocation(Location spawnLocation) {
                                 Texture.QUESTS_SCROLL_ICON,
                                 CommonColors.WHITE,
                                 CommonColors.WHITE,
-                                CommonColors.WHITE),
+                                CommonColors.WHITE,
+                                Managers.Feature.getFeatureInstance(ContentTrackerFeature.class)
+                                                .showAdditionalTextInWorld
+                                                .get()
+                                        ? ""
```
kristofbolyai, line 45: Can you comment in the code what this empty string is supposed to fix? So whenever we get back to this, we know what happened here.

Rafii2198: Alright, do I tag it with the FIXME?

kristofbolyai (approved):
Looks good. I think enabled by default is acceptable.

PR #2754, opened 2024-08-16
feat: Update various "paper" textures with new Wynn style [skip ci]

ShadowCat117:
Scroll button looks a little off on poi manager but for some reason it's always been small, will fix that when the screen is ported to MapData

Custom Character Selection screen textures have also not been changed as I plan to redo that screen at some point soon.
And Custom Trade Market textures not touched either as the feature has not been updated for 2.1 yet and likely won't use the existing textures anyway due to the new vanilla trade market textures


magicus:
Frankly, it looks a bit boring. :( I have not played enough to really recall how the new style looks, so maybe this is the way we have to go, but I must say I am going to miss the old "weathered" look.

ShadowCat117:
They are a little more uninspired than the original paper ones to be honest, but this fits better with the new art style they're going with

ryanzhoudev:
Is it viable at all to make this an enum toggle? I honestly really liked the old look and I still use the paper texture for my minimap

ShadowCat117:
A resource pack can be used to get the older textures, some of these screens don't belong to a feature so no way to add  config

kristofbolyai:
I like this. Wynntils is made to be an addition of Wynncraft, so in theory, we do need to match their style. People can customise the game to their liking, with resource packs.

kristofbolyai (approved):

kristofbolyai:
@magicus are you fine with this?

kristofbolyai:
Okay. I'll merge this now, and we'll get back to any discussion, if needed.

magicus:
You did the right call to merge this without waiting for me. But I feel a bit like Donkey here. We should perhaps think a bit about having a story to change the theme for Wynntils. I'm not saying it is a high-priority issue, though.

kristofbolyai:
There are some Wynncraft packs that already change our textures as well. But we can change our defaults as well.

PR #2760, opened 2024-08-20
feat: Fully parse items without Wynn Item API (UnknownGear parsing)

kristofbolyai:
This is useful for adding UnknownGear sharing in chat. It also fixes these items not being parsed as unknown gear and not being able to be used with spell cast keybinds.

### DIFF
--- a/common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
+++ b/common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
@@ -86,6 +86,9 @@ public final class WynnItemParser {
     private static final Pattern SKILL_REQ_PATTERN =
             Pattern.compile("^§(c✖|a✔)§7 (?<skill>[a-zA-Z]+) Min: (?<value>-?\\d+)$");
 
+    // Test in WynnItemParser_QUEST_REQ_PATTERN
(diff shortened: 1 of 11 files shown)


common/src/main/java/com/wynntils/models/wynnitem/parsing/WynnItemParser.java
```diff
@@ -217,17 +226,44 @@ public static WynnItemParseResult parseItemStack(
...
+                RangedValue range = RangedValue.fromString(damageMatcher.group("range"));
+                damages.add(Pair.of(DamageType.fromSymbol(symbol), range));
+            }
+
+            Matcher defenceMatcher = coded.getMatcher(ITEM_DEFENCE_PATTERN);
+            if (defenceMatcher.matches()) {
+                String symbol = defenceMatcher.group("symbol");
+                int value = Integer.parseInt(defenceMatcher.group("value"));
+                defences.add(Pair.of(Element.fromSymbol(symbol), value));
+            }
```
ShadowCat117: You can add a `continue` if any of these match, otherwise it'll continue checking for matches that *shouldn't* be matching. Looks like a few other checks in this loop don't have a continue either so probably worth adding to those as well

kristofbolyai: Indeed.

ShadowCat117 (approved):

PR #2770, opened 2024-08-22
fix: Fix auto progress NPC dialogue not releasing sneak [skip ci]

ShadowCat117:


kristofbolyai (approved):
We usually don't like using the Packet Event, but I understand it.

magicus:
> We usually don't like using the Packet Event, but I understand it.

Hm, can you explain? Because I think we should not use it for anything other than the packet dump debugging, period.

I'm looking at this code and I see two things:
1) Canceling a shift press
2) Delaying the shift release send to the next tick

Although I do now know the exact problem this PR is trying to resolve, I am just wondering if doing just 2) would have sufficed? 

Also, if 1 really is needed, I wonder if it is possible to do with a proper mixin instead.

PR #2775, opened 2024-08-23
fix: Fix hide player armor hiding new cosmetics [skip ci]

ShadowCat117:
Another iron horse armour change :)


kristofbolyai (approved):
Yes... Also, we'll need to use the API for the iron_horse_armor GearType ranges, because they vary too frequently now, and can't be hard-coded like we did before.

PR #2777, opened 2024-08-23
feat: Added the ability to manually override stat ranges from api

YoavNahumus:
Made it possible to edit a file in %appdata%/Wynntils/gear_info.json to change ranges for items
Ideally you would be able to edit it in game later


ShadowCat117:
Isn't this already possible with https://github.com/Wynntils/Wynntils/pull/2351 just setup your own fork of Static-Storage using the beta API values

magicus:
I'm not really fond of this. Perhaps ShadowCat's solution is working out for you instead of this hack?

YoavNahumus:
If I'm understanding correctly the solution is to host my own api and use it instead of wapi? if so then what i want is to be able to just override specific ranges of specific items without needing all the items

kristofbolyai:
What if the solution is waiting that week or two for the Wynn API...?

kristofbolyai (commented):
I guess I'll have to be the "annoying developer" here and close this. I think no other developers like the idea of "hacking" in changes to items (as it was already expressed above).

I know how annoying not having the latest item changes are. I'd love to provide a way to add this changes now, but there is no real way, when the Wynncraft API Beta releases 2-3 months late, compared to the actual game's beta. When I tried implementing the _beta_ API, I've found around 6-7 bugs, which I've all reported, but are yet to be fixed. Item stats themselves are not up to date on the beta API, they are not in sync with the game. Same goes for item models, meaning we'd have to show random models in the guides. Ingredients have 50% of their actual stat info missing. Tomes, charms, and the item ID collection also each have their small issues.

Despite all these issues, a draft PR is already present to add the new items, the backend infrastructure is ready for the beta API to get fixed, or the production API release. Trust me, we'd like this change as much as you, players.

Thank you for understanding.

PR #2780, opened 2024-08-24
fix: Fix durability arcs not rendering on all items in inventory [skip ci]

ShadowCat117:
Current:
image

Fixed:
image

### DIFF
--- a/common/src/main/java/com/wynntils/features/inventory/EmeraldPouchFillArcFeature.java
+++ b/common/src/main/java/com/wynntils/features/inventory/EmeraldPouchFillArcFeature.java
@@ -37,7 +37,7 @@ public class EmeraldPouchFillArcFeature extends Feature {
     }
 
     @SubscribeEvent
-    public void onRenderSlot(SlotRenderEvent.Pre e) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/inventory/EmeraldPouchFillArcFeature.java
```diff
@@ -37,7 +37,7 @@ public void onRenderHotbarSlot(HotbarSlotRenderEvent.Pre e) {
     }
 
     @SubscribeEvent
-    public void onRenderSlot(SlotRenderEvent.Pre e) {
+    public void onRenderSlot(SlotRenderEvent.Post e) {
```
ShadowCat117, line 40: Doesn't seem to be required here but better to keep it the same in case it changes

kristofbolyai (commented):
Do we know why this change fixed the issue?

ShadowCat117:
I assume something with the new textures. It also needs to be changed for searching as that has the same issue
image

kristofbolyai:
Ah, that explains it.

PR #2781, opened 2024-08-24
fix: Fix guild chat not ignoring username in chat mentions [skip ci]

ShadowCat117:
Hopefully last change to this, I didn't test it on guild chat originally which doesn't have the whitespace after the `:` so it wasn't matching the pattern and would detect your username as a mention

### DIFF
--- a/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatCoordinatesFeature.java
@@ -26,7 +26,7 @@ import net.neoforged.bus.api.SubscribeEvent;
 
 @ConfigCategory(Category.CHAT)
 public class ChatCoordinatesFeature extends Feature {
-    private static final Pattern END_OF_HEADER_PATTERN = Pattern.compile(".*:\\s");
+    private static final Pattern END_OF_HEADER_PATTERN = Pattern.compile(".*:\\s?");
 
     @SubscribeEvent(priority = EventPriority.HIGH)
     public void onChatReceived(ChatMessageReceivedEvent e) {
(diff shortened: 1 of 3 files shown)


kristofbolyai:
Perhaps it's time to add unit testing for this?

ShadowCat117:
Hmm, doesn't really work here that well as it's matching a split up StyledText so the case would just be ":" and ": "

magicus:
Why is this pattern duplicated across three features? It seems like it needs a single, proper place to live in.

Also, we can make other unit tests apart from the pure regexp ones. If we have string operations that have been tricky to get right, it can certainly be a point in writing unit tests to make sure we handle/split/modify/whatever correctly.

ShadowCat117:
> Why is this pattern duplicated across three features? It seems like it needs a single, proper place to live in.

I'm not really sure of a good place to keep it. An idea I did have whilst doing this was a way to strip the header from the messages so the current and any other future features won't need to handle it themselves.

kristofbolyai:
Ideally, this would have a separate Unit Test, for the whole "replacement". This includes those other features. However, there is no point in holding this back fix.

kristofbolyai (approved):

magicus:
> An idea I did have whilst doing this was a way to strip the header from the messages so the current and any other future features won't need to handle it themselves.

Like a `ChatHandler.stripHeader(chatMsg)` perhaps? Sounds reasonable. 

An alternative, after looking at the code, is that all three use places are doing a similar text replacement using `iterateBackwards`. Maybe the common functionality there can be extracted to a method that just takes a matcher and a lambda of what to do when a match is found? Or something like that.

Anyway, if you'd care to try to generalize this in any way, I'd appreciate it.

PR #2782, opened 2024-08-24
fix: Fix Nexus of Light tracking issues [skip ci]

ShadowCat117:
Fortunately since it's the last room it doesn't fully break tracking and the printed timer at the end shows the boss timer correctly, just the overlay won't be updated.

### DIFF
--- a/common/src/main/java/com/wynntils/models/raid/RaidModel.java
+++ b/common/src/main/java/com/wynntils/models/raid/RaidModel.java
@@ -81,7 +81,7 @@ public class RaidModel extends Model {
         }
     }
 
-    // One challenge is Nexus of Light does not display the scoreboard upon challenge completion
+    // One challenge in Nexus of Light does not display the scoreboard upon challenge completion
     // so we have to check for the chat message
     @SubscribeEvent
     public void onChatMessage(ChatMessageReceivedEvent event) {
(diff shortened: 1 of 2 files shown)


ShadowCat117:
Found the issue causing NoL to not count challenge 2 as completed so will fix that here too

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -79,6 +81,17 @@ public void onTitle(TitleSetTextEvent event) {
         }
     }
 
+    // One challenge is Nexus of Light does not display the scoreboard upon challenge completion
```
magicus: Is the first "challenge" here just a remark about difficulties implementing the mod? In any other situation I would not have cared, but since this is about "challenges" in the game, it sounds a bit confusing. Perhaps replace with "difficulty"?

ShadowCat117: It's referring to one of the challenge rooms in the raid

ShadowCat117: Oh I see the confusion now, it should be "One challenge in" not "One challenge is"

kristofbolyai (approved):

common/src/main/java/com/wynntils/models/raid/RaidModel.java
magicus: Ah, that makes sense. Somehow my brain failed to find that alternative interpretation.

PR #2788, opened 2024-08-25
fix: fix transparent layer compatibility with 3D Skin Layer mod [skip ci]

SteveKunG:
This PR fixes incompatibility with 3D Skin Layer mod with their 3D Layer with Ghost Transparent config.

Before
image

After
image

I've planned to fixes cape layer but it quite tricky for me, maybe for the next PR.
This is my first PR on the project if I'm doing something wrong please let me know. Thanks!


kristofbolyai:
Do we really need a separate mixin library? What if you just set that specific mixin to not error on fail? I'd love to accept this fix, but embedding a new library, previously unknown to me/us is not something I'd want to do, if I can avoid it.

SteveKunG:
I've remove conditional mixin as your request. I just realized vanilla mixin can handle missing classes in runtime when its not present.

kristofbolyai:
I think I like the `compat` subdirectory in mixins. I'll give @magicus a few days to express their opinion.

Thank you for contributing!

magicus:
Ooohhh... This is a mixin, not for Vanilla code, but for another mod? Hmmm. I am not sure I like that, at all.

What is the problem we are trying to solve here? @SteveKunG Can you please write an issue describing what you want to achieve? Maybe there is some other way; like we try to understand what the incompatibility is, or try to get in touch with the developer of that other mod.

common/src/main/java/com/wynntils/mc/mixin/compat/CustomLayerFeatureRendererMixin.java
```diff
@@ -0,0 +1,37 @@
...
+            method = "renderLayers",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "dev/tr7zw/skinlayers/api/Mesh.render(Lnet/minecraft/client/model/geom/ModelPart;Lcom/mojang/blaze3d/vertex/PoseStack;Lcom/mojang/blaze3d/vertex/VertexConsumer;III)V"),
+            index = 5)
+    private int setTranslucentFor3DSkinLayer(
+            int original, @Local(argsOnly = true) AbstractClientPlayer abstractClientPlayer) {
+        boolean isGhostPlayer = Models.Player.isPlayerGhost(abstractClientPlayer);
```
magicus, line 31: And even if we chose to go ahead on this path with modding other mods, this code is not acceptable. Calling a Model from within a mixin is a no-no in Wynntils. 

The abstraction level of the fix needs to be raised, in such a case. But, before you start changing anything, I think we need to reconsider our options here.

SteveKunG:
> Ooohhh... This is a mixin, not for Vanilla code, but for another mod? Hmmm. I am not sure I like that, at all.
> 
> What is the problem we are trying to solve here? @SteveKunG Can you please write an issue describing what you want to achieve? Maybe there is some other way; like we try to understand what the incompatibility is, or try to get in touch with the developer of that other mod.

3DSkinLayer mod added a 3D Layer to your skin, the current issue is when you see other players in the Wynncraft that is not in the same world. Their 3D Layer appear to be full opaque and does not rendered as transparent.

So this PR will fix the issue as I described above but this is not a good way to solve the current issue. I will create an issue to request a proper API on their repo.
https://github.com/tr7zw/3d-Skin-Layers/issues/173

kristofbolyai:
So, what's the state of this PR? Can we close this, and assume other mods will fix compatibility?

SteveKunG:
I will close it for now until Skin Layer dev has time to make the API. In the meantime I will create a small mod to fix this but the code might looks cursed and janky.

PR #2790, opened 2024-08-25
fix: fix transparent cape for ghost players [skip ci]

SteveKunG:
image


magicus:
Thank you for wanting to contribute to Wynntils. However, your code is breaking a number of fundamental rules of separation of concern we have in Wynntils, so even if it technically might fix the problem, it needs to be redesigned to be accepted into Wynntils. 

Unfortunately, there is no good documentation of these rules, so I understand it's hard to know that you've broken them. :-)

Basically, the idea is that from mixins, all you are allowed to do is to send events. The actual decisions about modifications is done by a Model (if it is more general and high-level), or a Feature (which my initial guess is would be the correct place in this particular case).

Also, please have a look at other mixins on how to call instance methods on the mixin target class, instead of declaring the mixin class as extending stuff.

SteveKunG:
Thanks for guiding me on how to contributing to Wynntils, I'm a little bit confusing with Forge-like event style because I haven't familiar with it for years haha.

The last changes to this PR I use `RenderType.itemEntityTranslucentCull` instead of `RenderType.entityTranslucent`. Please don't merge it until I get home (For about 8 hours from now) and see it my for myself, even if it's good for the team.

kristofbolyai:
I don't think we should be using an "item entity" render type for cape rendering. Perhaps look into creating a CustomRenderType (see the class). Even if all the settings are the same at the moment, it may change in the future, breaking us.

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +53,32 @@ private void render(
...
+                                    "net/minecraft/client/renderer/MultiBufferSource.getBuffer(Lnet/minecraft/client/renderer/RenderType;)Lcom/mojang/blaze3d/vertex/VertexConsumer;"))
+    private RenderType setTranslucenceCapeRenderType(
+            RenderType original,
+            @Local(argsOnly = true) AbstractClientPlayer livingEntity,
+            @Local PlayerSkin playerSkin) {
+        Minecraft minecraft = Minecraft.getInstance();
+        boolean isBodyVisible = !livingEntity.isInvisible();
+        boolean translucent = !isBodyVisible && !livingEntity.isInvisibleTo(minecraft.player);
+
+        LivingEntityRenderTranslucentCheckEvent event =
```
kristofbolyai: Does this mean this event is now called from 2 different sites? If so, is it called once for the entity itself, once for the cape layer? If so, it'd be wise to add a way for the event to specify which layer it's referring to, or alternatively, create a new, unique event.

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +53,32 @@ private void render(
             ci.cancel();
         }
     }
+
+    @ModifyArg(
+            method = "render",
```
kristofbolyai: One other quirk we have in Wynntils is that we use "semi-qualified" mixin method naming. This is not common in the scene, but basically it gives us all the benefits of using fully-qualified method names, but the code stays readable. You can check the format  
from other mixins, but it's basically a fully qualified method target, without the "Lnet/.../MixinClass" prefix.

Don't forget to change both of the mixins in the PR. Thanks!

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +53,32 @@ private void render(
...
+        Minecraft minecraft = Minecraft.getInstance();
+        boolean isBodyVisible = !livingEntity.isInvisible();
+        boolean translucent = !isBodyVisible && !livingEntity.isInvisibleTo(minecraft.player);
+
+        LivingEntityRenderTranslucentCheckEvent event =
+                new LivingEntityRenderTranslucentCheckEvent(translucent, livingEntity, translucent ? 0.15f : 1.0f);
+        MixinHelper.post(event);
+
+        // Translucence value needs to pass into PlayerModel class, because renderCloak method has no 'color' argument
+        PlayerModelExtension playerModelExtension = (PlayerModelExtension) this.getParentModel();
```
kristofbolyai: Referring to my earlier comment;

Would you think it's feasible change to make the event itself hold the PlayerModelExtension, and the event users can modify the translucence (and perhaps, in the future, other specific settings) directly, without the event acting as kind of a middle man.

kristofbolyai (commented):
Thank you! It looks much better now. Now that we've separated the mod's more higher level code from the mixins, we can focus on clearing up some details about the events and the mixins.

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
SteveKunG: Done!

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
SteveKunG: Done! I created a new event `PlayerFeatureRenderTranslucentCheckEvent` specific just for player. If somehow Wynncraft added an Elytra as a new armor or cosmetic, we can expand an additional event here.

SteveKunG: Also I reverted `RenderType` back to `entityTranslucent`, hope that should be enough for translucent cape.

SteveKunG:
Um well, I think I should request an API for WaveyCapes as well.

image

Just compiled and testing in production (Fabric), is there anything else to improve and fix?

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -1,22 +1,35 @@
...
+import net.minecraft.client.resources.PlayerSkin;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 @Mixin(CapeLayer.class)
-public abstract class CapeLayerMixin {
+public abstract class CapeLayerMixin extends RenderLayer<AbstractClientPlayer, PlayerModel<AbstractClientPlayer>> {
```
magicus: Please revert this change.

common/src/main/java/com/wynntils/mc/mixin/PlayerModelMixin.java
```diff
@@ -0,0 +1,51 @@
...
+import com.wynntils.mc.extension.PlayerModelExtension;
+import com.wynntils.utils.colors.CommonColors;
+import net.minecraft.client.model.PlayerModel;
+import net.minecraft.client.model.geom.ModelPart;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+
+@Mixin(PlayerModel.class)
+public class PlayerModelMixin implements PlayerModelExtension {
```
magicus:
```suggestion
public abstract class PlayerModelMixin implements PlayerModelExtension {
```

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
kristofbolyai: This is actually not so evil as you think and is allowed in mixins. What's your reasoning? Or are the superclasses not used anywhere?

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +47,28 @@ private void render(
...
+    private RenderType setTranslucenceCapeRenderType(
+            RenderType original,
+            @Local(argsOnly = true) AbstractClientPlayer livingEntity,
+            @Local PlayerSkin playerSkin) {
+        Minecraft minecraft = Minecraft.getInstance();
+        boolean isBodyVisible = !livingEntity.isInvisible();
+        boolean translucent = !isBodyVisible && !livingEntity.isInvisibleTo(minecraft.player);
+
+        PlayerFeatureRenderTranslucentCheckEvent.Cape event = new PlayerFeatureRenderTranslucentCheckEvent.Cape(
+                translucent, livingEntity, translucent ? 0.15f : 1.0f, ((PlayerModelExtension)
```
magicus: Why the hard-coded 0.15f? That is something the Feature can figure out itself as long as the `translucent` argument is passed in the event.

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +47,28 @@ private void render(
             ci.cancel();
         }
     }
+
+    @ModifyArg(
```
kristofbolyai, line 55: Does ModifyArg work on Neoforge? IIrc either `ModifyArg` or `ModifyArgs` did not use to work with Forge. I think the latter, as it caused some module generation - dependency resolution issues. But please do confirm this for me. Thanks.

common/src/main/java/com/wynntils/mc/mixin/PlayerModelMixin.java
```diff
@@ -0,0 +1,51 @@
...
+    private float wynntilsTranslucence;
+
+    @WrapOperation(
+            method = "renderCloak(Lcom/mojang/blaze3d/vertex/PoseStack;Lcom/mojang/blaze3d/vertex/VertexConsumer;II)V",
+            at =
+                    @At(
+                            value = "INVOKE",
+                            target =
+                                    "net/minecraft/client/model/geom/ModelPart.render(Lcom/mojang/blaze3d/vertex/PoseStack;Lcom/mojang/blaze3d/vertex/VertexConsumer;II)V"))
+    private void setTranslucenceCape(
```
magicus: The name is not ideal, since it makes it an overload of the setter below. Something like `renderCloakWithTransparency` would be better.

common/src/main/java/com/wynntils/mc/event/PlayerFeatureRenderTranslucentCheckEvent.java
```diff
@@ -0,0 +1,48 @@
...
+
+        /**
+         * Translucence value needs to pass into {@link net.minecraft.client.model.PlayerModel} class,
+         * because {@link net.minecraft.client.model.PlayerModel#renderCloak(PoseStack, VertexConsumer, int, int)} method does not accept {@code color}
+         * argument
+         * @param translucence
+         */
+        @Override
+        public void setTranslucence(float translucence) {
+            super.setTranslucence(translucence);
```
kristofbolyai: So when we set the cape translucence, we also set the player/entity translucence? Is that expected/wanted behavior?

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
```diff
@@ -40,4 +47,28 @@ private void render(
...
+        Minecraft minecraft = Minecraft.getInstance();
+        boolean isBodyVisible = !livingEntity.isInvisible();
+        boolean translucent = !isBodyVisible && !livingEntity.isInvisibleTo(minecraft.player);
+
+        PlayerFeatureRenderTranslucentCheckEvent.Cape event = new PlayerFeatureRenderTranslucentCheckEvent.Cape(
+                translucent, livingEntity, translucent ? 0.15f : 1.0f, ((PlayerModelExtension)
+                        ((CapeLayer) (Object) this).getParentModel()));
+        MixinHelper.post(event);
+
+        return event.getTranslucence() < 1.0f ? RenderType.entityTranslucent(playerSkin.capeTexture()) : original;
```
magicus: Overall, I think the logic of the entire fix seems very convoluted. I'm trying to wrap my head around what is happening, and is kind of failing. 

To get the correct transparancy, you need to do things:
1) Update the RenderType argument to getBuffer (done here)
2) Make sure PlayerModel.renderCloak renders the cape with an alpha value

Is my understanding of this correct?

If so, why can't you do the getParentModel.setTranslucenceCape() here, if the event has an updated translucence? Right now, the logic is spread out across feature - event - two different mixins.

I'd like to see a single, trivial event at the point of interception. The feature can listen to that event and flag it as translucent, and then when the event is checked in the point of interception, all changes that are needed to be done are done there.

magicus: If you do that, you can get rid of the strange hierarchy you are introducing below LivingEntityRenderTranslucentCheckEvent. 

Instead, you should go to the actual class file of LivingEntityRenderTranslucentCheckEvent, make it abstract, and create two nested subclasses in that file, one for Body and one for Cape. They will need to have no additional fields or methods, but will just be markers. Then replace the current usage of LivingEntityRenderTranslucentCheckEvent with LivingEntityRenderTranslucentCheckEvent.Body, and replicate it for your new LivingEntityRenderTranslucentCheckEvent.Cape as well. (Unfortunately with the recent update to the neoforge bus we cannot listen on abstract events :-().

magicus: Also, please rename that class to just `RenderTranslucentCheckEvent`. The current name is just cumbersome, not helpful and (after my suggested change) straight out wrong.

common/src/main/java/com/wynntils/mc/mixin/CapeLayerMixin.java
magicus: The only reason to do this is to have access to more methods on `this`. We used to do this before, and it ended up getting messy, forcing us to add dummy constructors, etc, and in the end it did not work in all cases anyway. So I had to check how to resolve that. And then I found that a much better way is to make a double cast:

`(VanillaTargetClass) (Object) this`. It is a bit more verbose at the call point, but we get rid of all the other cruft. Steve has fixed it this way already in the PR now.

magicus: So I went a pass through all our mixins to make sure we did not do that anywhere, but used double-casting instead. I'd like to keep it that way. So the rule is: all mixins should be `public abstract class <VanillaClassName>Mixin {` and that's it. Ideally, I'd like to have some check to enforce this.

common/src/main/java/com/wynntils/mc/event/PlayerFeatureRenderTranslucentCheckEvent.java
magicus: Actually, I don't think that is what is happening. It is just propagating the value up the strange event inheritance hierarchy.

But the code is way too convoluted for what it is supposed to be doing, so I understand that you interpret it as this. (And then again, it might be me who is wrong about this as well...)

magicus (approved):
The code looks fine now. Thank you @SteveKunG for working with us to get it to this point!

The only question remaining before we push this is if there is a problem on NeoForge as bolyai says. This needs to be checked.

SteveKunG:
I'm not sure why my reviews are in Pending state, not sure someone already seeing this but I'm tested on NeoForge yesterday.
image

magicus:
That happens when you press "Start a review" rather than "Post single comment". To get them posted, you need to go to Review changes and press "Post review". Idiotic UI from Gihub's side.

kristofbolyai (approved):
LGTM, thanks.

magicus:
(I guess the idea is that you can make tentative comments when starting to review a PR and then when you find the answer to your question further down you can remove the comment and noone will be the wiser, instead of you looking like a fool. But it's annoying that they try to push it as default with the green colored buttons.)

PR #2791, opened 2024-08-25
fix: fix chat redirects for 2.1 [skip ci]

Markus-Rost:
Fix chat redirects by adding the new chat icons to the patterns

### DIFF
--- a/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
@@ -130,7 +130,7 @@ public class ChatRedirectFeature extends Feature {
     public void onChatMessage(ChatMessageReceivedEvent e) {
         if (e.getRecipientType() != RecipientType.INFO) return;
 
-        StyledText message = e.getOriginalStyledText();
+        StyledText message = e.getOriginalStyledText().stripAlignment();
         MessageType messageType = e.getMessageType();
 
         for (Redirector redirector : redirectors) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
```diff
@@ -961,8 +964,8 @@ protected StyledText getNotification(Matcher matcher) {
     }
 
     private final class MerchantRedirector extends SimpleRedirector {
-        private static final Pattern FOREGROUND_PATTERN =
-                Pattern.compile("^§5(?<merchant>.*): §dThank you for your business\\. Come again!$");
+        private static final Pattern FOREGROUND_PATTERN = Pattern.compile(
+                "^§5(?:\uE00A\uE002|\uE001) (?<merchant>.*):§d Thank you for your \\n§5\uE001§d business\\. Come again!$");
```
kristofbolyai: As we've discussed in Discord, I'd like you to use unwrap. If you check the implementation, it will only remove the "wrapping" new lines, the ones followed by |. Regular newlines should be unaffected.

You may ask why I consider this so important. I think the current chat wrapping in Wynncraft is dialed way too high. And if they change the "wrapping width" on the server side, every pattern relying on these seemingly fixed locations will break. (stripAlignment and unwrap also exists for this reason, to ensure easier compatibility now, and in the present, when changes happen on Wynn).

Markus-Rost: I've added the unwrap and tested that both redirect are still working.

kristofbolyai (approved):
Looks good, thanks!

magicus:
Thanks for your contribution!

magicus (approved):

PR #2793, opened 2024-08-25
refactor: Rename BossBarModel to AbilityModel and move AnnihilationSunBar to WorldEventModel [skip ci]

ShadowCat117:
Completely forgot we can register bars from models.. (:

### DIFF
--- a/common/src/main/java/com/wynntils/core/components/Models.java
+++ b/common/src/main/java/com/wynntils/core/components/Models.java
@@ -4,8 +4,8 @@
  */
 package com.wynntils.core.components;
 
+import com.wynntils.models.abilities.AbilityModel;
 import com.wynntils.models.abilities.ArrowShieldModel;
-import com.wynntils.models.abilities.BossBarModel;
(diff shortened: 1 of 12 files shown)


magicus:
Why do we even have a BossBarModel? I was confused at first since it should be a handler. Which it is; we just also have a BossBarModel. Did we not find any better place to put these boss bars? Maybe we should try a bit harder.

ShadowCat117:
Probably CharacterModel is a good place to put the current ones as that already has some class information and the bars in there are class related

kristofbolyai:
> Probably CharacterModel is a good place to put the current ones as that already has some class information and the bars in there are class related

So, are you planning to do it here, or should we merge this as is?

ShadowCat117:
Yeah I can do it here

ShadowCat117:
Actually, the BossBarModel is already in the abilities package so will just rename it to AbilityModel

kristofbolyai (approved):

PR #2797, opened 2024-08-27
feat: Introduce graphed, dependency based downloads

kristofbolyai:
`DownloadManager` has the documentation I'd put in the PR description.

TLDR; The goal is to eliminate download dependency race once and for all, all while taking benefit of having a manager behind all downloads, which include proper failure handling / retries, maximum parallel download settings, a clear and easily calculated download progress percentage and a list of download failures which can be shown to the user.

### DIFF
--- a/common/src/main/java/com/wynntils/core/net/DownloadDependencyGraph.java
+++ b/common/src/main/java/com/wynntils/core/net/DownloadDependencyGraph.java
@@ -1,0 +1,281 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.net;
+
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.components.CoreComponent;
+import com.wynntils.utils.type.Pair;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+
+public final class DownloadDependencyGraph {
+    private final Map<Node, NodeState> nodeMap = new ConcurrentHashMap<>();
+
+    private DownloadDependencyGraph(List<Node> nodes) {
+        nodes.forEach(node ->
+                nodeMap.put(node, node.dependencies.isEmpty() ? NodeState.QUEUED : NodeState.WAITING_ON_DEPENDENCY));
+    }
+
+    public static DownloadDependencyGraph build(List<QueuedDownload> downloads) {
+        // Short circut on no downloads, to handle the edge-case one and for all in all sanity checks
+        if (downloads.isEmpty()) return new DownloadDependencyGraph(List.of());
+
+        // Firsly, run sanity checks on the queued downloads to determine if a valid graph can be built
+        // 1. Check for duplicate downloads
+        checkDuplicateDownloads(downloads);
+
+        // 2. Check for missign dependencies
+        checkMissingDependencies(downloads);
+
+        // Now, let's build a directed graph of the download dependencies
+        List<Node> nodes = downloads.stream().map(Node::new).toList();
+        nodes.forEach(node -> node.calculateDependencies(nodes));
+
+        // The latter part of the sanity checks can be ran after building the graph:
+        // 1. Check for circular dependencies
+        checkCircularDependencies(nodes);
+
+        return new DownloadDependencyGraph(nodes);
+    }
+
+    // region Processing
+
+    public QueuedDownload nextDownload() {
+        Node nextNode = nodeMap.entrySet().stream()
+                .filter(entry -> entry.getValue() == NodeState.QUEUED)
+                .map(Map.Entry::getKey)
+                .findAny()
+                .orElse(null);
+
+        if (nextNode == null) return null;
+
+        nodeMap.put(nextNode, NodeState.IN_PROGRESS);
+
+        return nextNode.download;
+    }
+
+    public void markDownloadCompleted(QueuedDownload download) {
+        Node node = nodeMap.keySet().stream()
+                .filter(n -> n.download == download)
+                .findAny()
+                .orElseThrow(() -> new IllegalStateException("Download not found in graph: " + download.urlId()));
+
+        nodeMap.put(node, NodeState.COMPLETED);
+
+        // Mark all dependents as ready to be downloaded, if all their dependencies are completed
+        node.dependents.forEach(dependent -> {
+            if (dependent.dependencies.stream()
+                    .allMatch(dependency -> nodeMap.get(dependency) == NodeState.COMPLETED)) {
+                nodeMap.put(dependent, NodeState.QUEUED);
+            }
+        });
+    }
+
+    public void markDownloadError(QueuedDownload download) {
+        Node node = nodeMap.keySet().stream()
+                .filter(n -> n.download == download)
+                .findAny()
+                .orElseThrow(() -> new IllegalStateException("Download not found in graph: " + download.urlId()));
+
+        nodeMap.put(node, NodeState.ERROR);
+
+        // Mark all dependents as error, as they cannot be downloaded
+        node.dependents.forEach(dependent -> nodeMap.put(dependent, NodeState.ERROR));
+    }
+
+    // endregion
+
+    // region State and Progress
+
+    public boolean isFinished() {
+        return nodeMap.values().stream().allMatch(state -> state == NodeState.COMPLETED || state == NodeState.ERROR);
+    }
+
+    public boolean hasError() {
+        return nodeMap.values().stream().anyMatch(state -> state == NodeState.ERROR);
+    }
+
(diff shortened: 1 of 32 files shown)


common/src/main/java/com/wynntils/core/net/DownloadManager.java
```diff
@@ -0,0 +1,70 @@
...
+import com.wynntils.core.components.Managers;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Manages downloading files from the internet. This manager acts as a middle layer between the caller components
+ * and {@link NetManager}. It's main purpose is to provide the components a well regulated and managed way
+ * of downloading dependency files.
+ * <p><b>This manager should not be used for spontainous downloads. Use one {@link NetManager}'s download methods
```
magicus: typo

magicus:
Oh, very nice!!!

I think you have the completely right idea here. The design sketched here looks sound; consider it approved. :)

But sometimes I wonder why we have to do so much ground-work ourselves. Some of the hardest part of the mod to get right have had nothing to do with Wynntils or Minecraft modding at all, like config/storage, downloading and event buses. *sigh* There is an old saying: "Any sufficiently large software project will end up re-implementing emacs", or something like that. I feel that is kind of what we are doing.

kristofbolyai:
Now you can check how the code API would look, without an actual backing implementation. I wanted to break away from how `NetManger#download` is always callable, so I introduced a proper time to register these downloads. I think they turned out great, and removed most of the previous mess we've had dealing with dependencies.

common/src/main/java/com/wynntils/models/ingredients/IngredientModel.java
```diff
@@ -17,40 +16,23 @@
...
 public class IngredientModel extends Model {
     private static final Map<ChatFormatting, Integer> TIER_COLOR_CODES = Map.of(
             ChatFormatting.DARK_GRAY, 0,
             ChatFormatting.YELLOW, 1,
             ChatFormatting.LIGHT_PURPLE, 2,
             ChatFormatting.AQUA, 3);
+
     private final IngredientInfoRegistry ingredientInfoRegistry = new IngredientInfoRegistry();
 
     public IngredientModel(WynnItemModel wynnItem) {
```
magicus: I'm guessing the dependency in wynnItem can be removed now? My reading of this code was that it was added either when we loaded the ingredient DB in the constructor, or as a safe guard against a race condition when it was made to be asynchronously loaded.

kristofbolyai: That's likely correct. Or maybe when our "dependency" requirement was more strict. I'll remove it and check other cases as well.

common/src/main/java/com/wynntils/services/map/PoiService.java
```diff
@@ -75,87 +74,21 @@ public class PoiService extends Service {
...
     }
 
     @Override
-    public void reloadData() {
-        loadData();
+    public void registerDownloads(DownloadRegistry registry) {
+        registry.registerDownload(UrlId.DATA_STATIC_PLACES).handleReader(this::handlePlaces);
+        registry.registerDownload(UrlId.DATA_STATIC_SERVICES).handleReader(this::loadServices);
+        registry.registerDownload(UrlId.DATA_STATIC_COMBAT_LOCATIONS).handleReader(this::loadCombat);
+        registry.registerDownload(UrlId.DATA_STATIC_CAVE_INFO).handleReader(this::handleCaves);
```
magicus, line 84: The naming is a bit inconsistent. Maybe rename all to "handle*"?

kristofbolyai: Indeed. I think these were the only ones I missed though, I wanted all named handleX.

magicus:
This looks really, really great.

magicus:
We have some way to trigger a full re-download from a command. I recall not being entirely happy on how that was implemented. Maybe that is something that can be better integrated into the DownloadManager and made nicer as well?

kristofbolyai:
> We have some way to trigger a full re-download from a command. I recall not being entirely happy on how that was implemented. Maybe that is something that can be better integrated into the DownloadManager and made nicer as well?

Yes. It's wynntils reload / wynntils clearcaches that calls Model#reload or smth like that. It's definitely planned for renovation either in this PR, or as a follow up. I want all these download changes released as a minor version bump (likely 2.2.0, as 2.1.0 will be the Wynn V3.3 API update this weekend), so I'll retarget this PR to `feat/data-loading`.

kristofbolyai:
@magicus This should be reviewable now. Now, the mod uses the new manager, but new features, automatic retry on error is not implemented. In its current form, it's a replacement for the old messy dependencies, except dependency management actually works. Reloading also works as intended. We can already calculate some statistics, like error rate, however it's only in logs at the moment. I plan on adding each new feature as a separate PR, while also fixing some older issues related to data loading, hence the new branch this PR targets.

After a lot of iterations on my end, I've decided to keep it simple. I've went through a lot of iterations with CompletableFuture, implementing custom NetResult types. Currently, I am happy with the code's current form, although that may change in the upcoming days, which is why I am keeping these batch of changes out of main for now.

magicus:
I'll try to look at this as soon as I can.

common/src/main/java/com/wynntils/core/net/DownloadDependencyGraph.java
```diff
@@ -0,0 +1,286 @@
...
+
+    public static DownloadDependencyGraph build(List<QueuedDownload> downloads) {
+        // Short circut on no downloads, to handle the edge-case one and for all in all sanity checks
+        if (downloads.isEmpty()) return new DownloadDependencyGraph(List.of());
+
+        // Firsly, run sanity checks on the queued downloads to determine if a valid graph can be built
+        // 1. Check for duplicate downloads
+        checkDuplicateDownloads(downloads);
+
+        // 2. Check for missign dependencies
```
magicus:
```suggestion
        // 2. Check for missing dependencies
```

common/src/main/java/com/wynntils/core/net/DownloadDependencyGraph.java
```diff
@@ -0,0 +1,286 @@
...
+                .filter(download -> downloads.stream().anyMatch(other -> download != other && download.equals(other)))
+                .findAny()
+                .orElse(null);
+
+        if (duplicatedInSameComponent != null) {
+            throw new IllegalStateException(duplicatedInSameComponent.callerComponent()
+                    + " downloaded the same data twice: " + duplicatedInSameComponent.urlId());
+        }
+
+        // Downloading the same resouce in different core components is not a direct confirmation of a schematic issue,
```
magicus:
```suggestion
        // Downloading the same resource in different core components is not a direct confirmation of a schematic issue,
```

common/src/main/java/com/wynntils/core/net/DownloadManager.java
```diff
@@ -0,0 +1,220 @@
...
+        if (jsonArrayHandler != null) {
+            download.handleJsonArray(
+                    wrapDownloadHandler(jsonArrayHandler, queuedDownload), wrapDownloadFailure(queuedDownload));
+            return download;
+        }
+
+        throw new IllegalStateException("Queued download has no handler set: " + queuedDownload);
+    }
+
+    private void replaceDownload(QueuedDownload toBeReplaced) {
```
magicus: The name got me thinking that you first downloaded to a temporary file name, and then moved the file in place one the download was successful. (If we don't do that elsewhere in the NetManager, it is probably something we should do.)

But after scratching my head a bit, I realize this is the code responsible for repopulating the queue. I'd suggest renaming the method. Maybe `queueNextDownload()` instead?

common/src/main/java/com/wynntils/services/map/PoiService.java
```diff
@@ -75,87 +74,21 @@ public class PoiService extends Service {
...
-
-                providedCustomPois.put(poiProvider, ImmutableList.copyOf(pois));
-            });
-        }
-    }
-
-    @SubscribeEvent
-    public void onDataLoaded(NetResultProcessedEvent.ForUrlId event) {
-        if (!WynntilsMod.isDevelopmentEnvironment()) return;
-        // Serialize loading of POIs when on dev env
```
magicus, line 139: So this was just about triggering a race, which will be resolved by itself in the new system? Asking this as I see that you have added no dependency constraints for the DownloadManager. 

Or, to phrase it differently: are you sure that there are no dependencies between these downloads that you have missed to express?

magicus (approved):
Overall it looks very good! I left a few comments/questions.

common/src/main/java/com/wynntils/services/map/PoiService.java
kristofbolyai: We really have two options here. The serialized loading for POIs originally happened because of a multi-thread class loading race between threads in the fork join pool. As far as I know it's been fixed in Arch-Transformer / EventBus (or Bus, as they call it now..). I can either add the dependency back, or what I would rather do, remove it, as it's really likely to be fixed, and even if it is not, it's a race that only happens in development environment, meaning users won't have issues with it.

magicus: Right, it was some class loading stuff. Now I recall. Yeah, remove it, and hope it's been fixed. Otherwise we'll deal with it.

PR #2798, opened 2024-08-27
fix: Parse health when over 100k [skip ci]

ShadowCat117:
https://github.com/user-attachments/assets/bd0b6919-557a-499e-951b-db0055028a63


ryanzhoudev (approved):
I wonder if we'll ever need anything above k...

ShadowCat117:
> I wonder if we'll ever need anything above k...

They are in the resource pack so who knows...

kristofbolyai (approved):
So I did add support for this! Guess I'm not going insane. I also guess that I completely forgot that regex existed and I did not have a way to test this :)

PR #2800, opened 2024-08-27
feat: Make SpellCastMessageOverlay support Blood Pact

Markus-Rost:
I made SpellCastMessageOverlay support spells cast using the Blood Pact ability.

### DIFF
--- a/common/src/main/java/com/wynntils/overlays/SpellCastMessageOverlay.java
+++ b/common/src/main/java/com/wynntils/overlays/SpellCastMessageOverlay.java
@@ -55,8 +55,8 @@ public class SpellCastMessageOverlay extends Overlay {
         int manaCost = event.getManaCost();
         int healthCost = event.getHealthCost();
         if (healthCost > 0) {
-            spellMessage = StyledText.fromString(
-                    "§7" + event.getSpellType().getName() + " spell cast! §3[§b-" + manaCost + " ✺§3] §4[§c-" + healthCost + " ❤§4]");
+            spellMessage = StyledText.fromString("§7" + event.getSpellType().getName() + " spell cast! §3[§b-"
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/wynntils/overlays/SpellCastMessageOverlay.java
```diff
@@ -53,8 +53,14 @@ public void onItemRename(ItemRenamedEvent event) {
...
-        spellMessage = StyledText.fromString(
-                "§7" + event.getSpellType().getName() + " spell cast! §3[§b-" + manaCost + " ✺§3]");
+        int healthCost = event.getHealthCost();
+        if (healthCost > 0) {
+            spellMessage = StyledText.fromString("§7" + event.getSpellType().getName() + " spell cast! §3[§b-"
+                    + manaCost + " ✺§3] §4[§c-" + healthCost + " ❤§4]");
+        } else {
+            spellMessage = StyledText.fromString(
+                    "§7" + event.getSpellType().getName() + " spell cast! §3[§b-" + manaCost + " ✺§3]");
+        }
```
kristofbolyai, line 63: We have the opportunity to add support for cases where both mana and health is used, whenever Wynn decides to add an ability like that, we won't have to update this code.

Markus-Rost: Blood Pact is already using both Mana and Health, where mana used is always displayed even when 0 mana is used. So I'm not sure how a blood only option could be added.

kristofbolyai: I completely forgot to read that line actually. Sorry, it looks good.

kristofbolyai (approved):

PR #2801, opened 2024-08-28
fix: fix GuildBankRedirector pattern [skip ci]

cmkohnen:
Also since, these messages appear under `RecipientType.GUILD` remove the check for only INFO messages.

### DIFF
--- a/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
@@ -13,7 +13,6 @@ import com.wynntils.core.persisted.config.ConfigCategory;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.handlers.chat.event.ChatMessageReceivedEvent;
 import com.wynntils.handlers.chat.type.MessageType;
-import com.wynntils.handlers.chat.type.RecipientType;
 import com.wynntils.models.players.type.PlayerRank;
 import com.wynntils.utils.mc.StyledTextUtils;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


kristofbolyai (approved):
It doesn't seem like removing the recipient type check has any performance impact. However, I'll do revert this if anything significant pops up (that I couldn't see during my testing).

common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
```diff
@@ -133,8 +132,6 @@ private void register(Redirector redirector) {
 
     @SubscribeEvent(priority = EventPriority.HIGH)
     public void onChatMessage(ChatMessageReceivedEvent e) {
-        if (e.getRecipientType() != RecipientType.INFO) return;
```
magicus, line 136: So why did you not change it to GUILD instead?

kristofbolyai: Because this event applies all chat redirects. We are still in debt regarding this one.

magicus: Ah, now I see. Previously we only redirected info messages, now we also have a guild one. 

But let me rephrase it:

So why did you not change it to INFO or GUILD instead?

That this is dropped means we will check for redirects in all global chat messages etc.

kristofbolyai: Yes, we can do that. I've not had that idea. If I remember next time I'm developing, I'll PR it.

PR #2804, opened 2024-08-29
feat: Add CorruptedCacheAnnotator [skip ci]

ShadowCat117:
Not something that's very easy to get a hold of and test so I'm just basing it on this screenshot.

Won't count towards any dry streak but once the WorldEventModel is setup to detect event completion we should add one for this. For now this will just prevent users from closing the reward chest if the mythic blocker feature is enabled.

Cache


magicus (approved):

kristofbolyai (approved):
Hopefully it works.

PR #2803, opened 2024-08-29
feat: Update the visuals of the custom player list

ShadowCat117:
Category titles now come from the vanilla list so when "Party" changes to "Island info" or "Raid info" it no longer still says party at the top.
Text shadow settings, default to normal.
Grey text no longer becomes black.
Wider texture to support longer names.

PR #2806, opened 2024-08-30
fix: Make (Buffered)FontRenderer support custom color rendering [skip ci]

kristofbolyai:
This should fix rendering components like the ones in player tab, where Wynn is already using custom colors.


kristofbolyai:
@ShadowCat117 This should make raid components render correctly for #2803. Please do some more tests before approving this, as I could only take a short look, but it appears it works in most places.

ShadowCat117 (approved):
Did a raid with the changes and they rendered correctly!

PR #2812, opened 2024-08-31
fix: Various party management screen fixes [skip ci]

ILikePlayingGames:
- Fix texture not found in SuggestionPlayerWidget#renderWidget when trying to load the Steve skin for players with unknown skins
- Add pattern for new party leave message in Models.Party
- Fix "and" not being trimmed from the username of the last party member
- Hide party member column when the player is not in a party

### DIFF
--- a/common/src/main/java/com/wynntils/models/players/PartyModel.java
+++ b/common/src/main/java/com/wynntils/models/players/PartyModel.java
@@ -63,13 +63,14 @@ public final class PartyModel extends Model {
     private static final Pattern PARTY_LIST_SELF_FAILED =
             Pattern.compile(PARTY_PREFIX_REGEX + "You must be in a party to use this\\.");
 
+    // This message has no period unlike the others. Add a period here when Wynn adds one.
+    private static final Pattern PARTY_LEAVE_SELF =
+            Pattern.compile(PARTY_PREFIX_REGEX + "You have left your current party");
     private static final Pattern PARTY_LEAVE_OTHER =
             Pattern.compile(PARTY_PREFIX_REGEX + "(\\w{1,16}) has left the party\\.");
     private static final Pattern PARTY_LEAVE_SELF_ALREADYLEFT =
             Pattern.compile(PARTY_PREFIX_REGEX + "You must be in a party to leave\\.");
-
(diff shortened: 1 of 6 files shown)


kristofbolyai (approved):
Looks good to me. Thank you for your contribution.

@DonkeyBlaster as the original creator, would you like to review this as well?

ryanzhoudev:
> @DonkeyBlaster as the original creator, would you like to review this as well?

Yes please, thanks for notifying. My flight is arriving in around 11 hours so I'll look at it and merge then.

ryanzhoudev (commented):
What you have here is fine, but a couple things I want to bring up:
- When you don't render the party list on the left when not in a party, the entire screen feels very off centered. Perhaps you could instead write a "Currently not in party" text in the center of the left half of the screen?
- When you create a party using the screen button and disband it with the button next to your name, under the "kick" column, the screen does not refresh and makes it seem like you are still in your solo party. Are you able to look into this as well? It's fine if you don't want to, just let me know.

Thanks for contributing!

ILikePlayingGames:
Thanks for your feedback!

- I thought about that but the text was small and a lot of area was still blank, maybe I can make the text bigger and see.
- I guess Wynn fixed the missing period I reported, I’ll test later

ILikePlayingGames:
Which one is better? (I committed top)
| Top | Middle |
| ----|--------|
| image | image |

Also Wynn didn't add a period, they removed one from the disband message.

ryanzhoudev:
I kind of like the middle one more, it feels more in line with trying to take up screen space and is less off centered imo, but up to you. If you like what's here now, I am okay to merge.

ILikePlayingGames:
I’ll switch it to the middle. I also want to add a party member count out of 10 somewhere.

ILikePlayingGames:
New member count:

image

ILikePlayingGames:
@DonkeyBlaster I’m done with the changes now
, it is ready to merge.

ryanzhoudev:
Changes look good. Final thing I will ask of you is to align this title, so that it's in the middle like the Promote and Kick buttons. Then I will merge.
image

ILikePlayingGames:
I've aligned it.

image

ryanzhoudev (approved):

PR #2818, opened 2024-09-01
feat: Add colour picker for colour inputs [skip ci]

ShadowCat117:
https://github.com/user-attachments/assets/85ee1c6a-b202-40c2-ba6e-2538dbfdf72c


Rafii2198:
I think there should be a text field with the color that would dynamically change with the color as well as could be inputted to change the color and also in legacy there were some predetermined colors that could be selected, I don't remember what these colors were exactly tho.
There is not that much space for that, but I think the bars could be halved in size as they are pretty tall, that should give enough area for a text box and some squares with predetermined colors. Also, if there is enough space perhaps there could be 2 rows of these colors, first one would be the predetermined ones and second would remember last inputted colors like graphic apps do, but just a suggestion, but definitely would make it easier to set same color across many settings without having to copy it.

Rafii2198:
But other than that, great work, I should emphasize it from the beginning, it's just that when I look at these features I have the Legacy Missing Feature list in my head

ShadowCat117:
Wasn't able to do the recently input colours, as I'd like to keep that in storage but we can't do that from screens currently

magicus:
Screens don't support Storage? I don't think that would be very hard to fix.

ShadowCat117:
> Screens don't support Storage? I don't think that would be very hard to fix.

Don't think so, the screens that currently have a "setting" to save all have a respective feature to use.

magicus:
Settings are different, since they need to be shown in the interface. I thought all Components supported Storage. But hm, maybe Screens are not Components. Anyway, I don't think it would be hard to add.

kristofbolyai:
As a follow up, or even a dependency for this PR, we should remove all "x opacity" configs and use the custom colors alpha.

ShadowCat117:
> As a follow up, or even a dependency for this PR, we should remove all "x opacity" configs and use the custom colors alpha.

Only place I can find opacity configs that use a CustomColor too is in item highlights, but that has 2 separate opacity configs 1 for inventory and 1 for hotbar so can't really use the one from the CustomColor there

kristofbolyai:
Sorry, I'll have to take my time reviewing this, likely I can do it tomrorrow.

kristofbolyai (approved):
This looks good code wise, I can't test it unfortunately. If you are confident this should work, feel free to merge it whenever.

PR #2824, opened 2024-09-05
fix: Fix ability tree annotators [skip ci]

ShadowCat117:
Removed the color field from ArchetypeAbilitiesItem as it's no longer a ChatFormatting character and it wasn't used anywhere anyway

image


kristofbolyai (approved):
Great tests!

PR #2825, opened 2024-09-05
fix: Fix some more chat redirects [skip ci]

ShadowCat117:


common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
```diff
@@ -773,28 +772,6 @@ protected StyledText getNotification(Matcher matcher) {
...
-        }
-
-        @Override
-        protected StyledText getNotification(Matcher matcher) {
-            return StyledText.fromComponent(
-                    Component.translatable("feature.wynntils.chatRedirect.scrollTeleport.notificationMobs")
-                            .withStyle(ChatFormatting.DARK_RED));
-        }
-    }
-
```
ShadowCat117, line 797: This message doesn't seem to exist anymore as with the new scroll system the teleport is cancelled if you take damage

kristofbolyai (approved):
I believe we still could strip these symbols kind of easily, but let's get back to it later.

PR #2826, opened 2024-09-05
fix: Fix getting stuck in interim state when changing world via housing join [skip ci]

ShadowCat117:
Fixes the main issue of https://github.com/Wynntils/Wynntils/issues/2388 however world name is still unknown but I don't think anything can be done about that unless we run /find or check world via API for players currently on the housing, but they could be in /stream so that info is hidden so there's no consistent method

### DIFF
--- a/common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
+++ b/common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
@@ -31,7 +31,7 @@ import net.neoforged.bus.api.SubscribeEvent;
 public final class WorldStateModel extends Model {
     private static final UUID WORLD_NAME_UUID = UUID.fromString("16ff7452-714f-2752-b3cd-c3cb2068f6af");
     private static final Pattern WORLD_NAME = Pattern.compile("^§f {2}§lGlobal \\[(.*)\\]$");
-    private static final Pattern HOUSING_NAME = Pattern.compile("^§f  §l([^§\"\\\\]{1,18})$");
+    private static final Pattern HOUSING_NAME = Pattern.compile("^§f  §l([^§\"\\\\]{1,35})$");
     private static final Pattern HUB_NAME = Pattern.compile("^\n§6§l play.wynncraft.com \n$");
     private static final Pattern STREAMER_MESSAGE = Pattern.compile("§2Streamer mode (disabled|was enabled)\\.");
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
```diff
@@ -31,7 +31,7 @@
 public final class WorldStateModel extends Model {
     private static final UUID WORLD_NAME_UUID = UUID.fromString("16ff7452-714f-2752-b3cd-c3cb2068f6af");
     private static final Pattern WORLD_NAME = Pattern.compile("^§f {2}§lGlobal \\[(.*)\\]$");
-    private static final Pattern HOUSING_NAME = Pattern.compile("^§f  §l([^§\"\\\\]{1,18})$");
+    private static final Pattern HOUSING_NAME = Pattern.compile("^§f  §l([^§\"\\\\]{1,35})$");
```
ShadowCat117, line 34: 18 is the max limit that Wynn allows, but by default a guild housing name is `<guildname>'s HQ` and guilds have a name limit of 30 characters according to [the wiki](https://wynncraft.wiki.gg/wiki/Guilds#Creating)

common/src/main/java/com/wynntils/models/worlds/WorldStateModel.java
```diff
@@ -182,6 +182,9 @@ public void update(PlayerDisplayNameChangeEvent e) {
     private boolean setWorldIfMatched(Matcher m, boolean housing) {
         if (m.find()) {
             String worldName = housing ? currentWorldName : m.group(1);
+            if (worldName.isEmpty() && housing) {
+                WynntilsMod.warn("Changed world via housing join, current world name is unknown");
+            }
```
magicus, line 189: Maybe set worldName to "Unknown Housing" instead of the empty string?

ShadowCat117: World name should be "WC" followed by the number rather than the housing name, that is set below. So maybe "WC??" is better?

magicus: Sure.

magicus (approved):

kristofbolyai (approved):

PR #2827, opened 2024-09-05
feat: Add Range Visualizer for Raid Buff Major Ids [skip ci]

ben9964:
This MR adds range circles to players and yourself when taking major ids that give ranged abilities such as Guardian, Heart of the pack etc. It's pretty useful to know if your teammate takes guardian and see where you can stand to keep that benefit!

The way that it does this is by tracking the buffs that players take when in a raid by extending the already present chat listener in the RaidModel class. Each buff is kept track of and then the range visualizer will check if any of these are major ids that require a range to be displayed and display them.

The list of raid buffs and major ids that they give is hard coded in RaidKind.java for each of the raids.


common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -39,13 +44,17 @@ public class RaidModel extends Model {
     private static final Pattern RAID_COMPLETED_PATTERN = Pattern.compile("§f§lR§#4d4d4dff§laid Completed!");
     private static final Pattern RAID_FAILED_PATTERN = Pattern.compile("§4§kRa§c§lid Failed!");
 
+    private static final Pattern RAID_CHOOSE_BUFF_PATTERN = Pattern.compile("(\\w+) has chosen the (\\w+ \\w+) buff!");
```
ShadowCat117: This pattern is a bit too weak, although unlikely if someone sent a regular message of "player has chosen the X buff!" this will catch it. You should make it expect the raid chat banner and also keep the colours, as far as I'm aware all raids use the same colours for the buff messages so no need to ignore those.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -85,6 +94,19 @@ public void onTitle(TitleSetTextEvent event) {
     // so we have to check for the chat message
     @SubscribeEvent
     public void onChatMessage(ChatMessageReceivedEvent event) {
+        if (inBuffRoom()) {
+            Matcher matcher = RAID_CHOOSE_BUFF_PATTERN.matcher(
+                    ColorUtils.stripColors(event.getStyledText().getString()));
+            if (matcher.find()) {
```
ShadowCat117: As mentioned above, you shouldn't need to remove the colours and should be able to use `matches` not `find`

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -85,6 +94,19 @@ public void onTitle(TitleSetTextEvent event) {
...
+                String playerName = matcher.group(1);
+                String buff = matcher.group(2);
+
+                partyRaidBuffs
+                        .computeIfAbsent(playerName, k -> new ArrayList<>())
+                        .add(buff);
```
ShadowCat117: What if here, instead of just storing the name and buff name you calculated the major ID here. So pass the buff group, eg "Bedrock III" into `RaidBuffs` and return "Heart of the Pack"

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -221,6 +246,22 @@ public RaidKind getCurrentRaid() {
...
+
+        for (String buffName : buffNames) {
+            String majorId = RaidBuffs.majorIdFromName(buffName);
+            if (majorId == null) continue;
+
+            majorIds.add(majorId);
+        }
+
+        return majorIds;
+    }
```
ShadowCat117: if you did my above comment you should be able to just return `buffNames`

common/src/main/java/com/wynntils/models/raid/type/RaidBuffs.java
ShadowCat117, line 1: I think this information can be somewhat linked to `RaidKind` so that when you are in NOG, you only need to check for the relevant buffs

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Noted, I can see how someone could trigger it by sending such a message. The reason I went the route of stripping the colours as you've mentionned in other comments is that it felt like features that rely on colours could become a chore to update if wynn just happens to decide they like pink more than orange for example. Happy to change it back though as originally I did have it this way.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: I would be happy to change the structure such that we only track major Ids rather than every raid buff. The original logic was that they may be a useful feature in extending this tracking to support seeing the buffs your teammates have taken somewhere on screen as well as the buffs you've taken (similar to how lootrunning missions that you've taken are displayed)

common/src/main/java/com/wynntils/models/raid/type/RaidBuffs.java
ben9964: Would the static buff -> major id map being a part of the Enum for each RaidKind be sufficient? Would it be clean enough to have something like

```java
NEST_OF_THE_GROOTSLANGS(
            "Nest of the Grootslangs",
            StyledText.fromString("§2Nest of The Grootslangs"),
            Pattern.compile("^Slay the Restless$"),
            Map.of(
                "Beserk", Map.of(1, "Explosive Impact"),
                "Lightbearer", Map.of(3, "Transcendence"),
                ...
            )
)
```

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ShadowCat117: It's not too much of a bother. Wynn changes various things that we have to adapt for, so having to change the colour in a pattern is at least an easy fix

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ShadowCat117: Ah if that is something you want to do in the future then by all means keep it this way

common/src/main/java/com/wynntils/models/raid/type/RaidBuffs.java
ShadowCat117: I think that would be acceptable,

ben9964: Resolved

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Should be resolved now. I've removed the color Util file that I added and made the pattern a complete match with what chat says when gaining a buff (either hex code for others or yellow for yourself)

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Resolved

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -39,13 +42,18 @@ public class RaidModel extends Model {
     private static final Pattern RAID_COMPLETED_PATTERN = Pattern.compile("§f§lR§#4d4d4dff§laid Completed!");
     private static final Pattern RAID_FAILED_PATTERN = Pattern.compile("§4§kRa§c§lid Failed!");
 
+    private static final Pattern RAID_CHOOSE_BUFF_PATTERN = Pattern.compile(
+            "§#d6401effó\u008F¿¼î€\u0081ó\u0090€† §(#fa7f63ff|e)(\\w+)§#d6401eff has chosen the §#fa7f63ff(\\w+ \\w+)§#d6401eff buff!");
```
ShadowCat117: If you change the pattern to `§#d6401eff(\uE009\uE002|\uE001) §#fa7f63ff(\w+)§#d6401eff has chosen the §#fa7f63ff(\w+ \w+)§#d6401eff buff!` then you can use `event.getOriginalStyledText().stripAlignment().getMatcher(RAID_CHOOSE_BUFF_PATTERN)` in the event to make it a little cleaner.

ShadowCat117: Also I will note here this pattern is not perfect, as nicknamed members will have the `§o` before their name and you will have to get their actual username from the hover event, see `RevealNicknamesFeature` on how you can do that

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -85,6 +93,19 @@ public void onTitle(TitleSetTextEvent event) {
     // so we have to check for the chat message
     @SubscribeEvent
     public void onChatMessage(ChatMessageReceivedEvent event) {
+        if (inBuffRoom()) {
+            WynntilsMod.info(event.getStyledText().getString());
```
ShadowCat117: Leftover debug

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ShadowCat117: The yellow is from the chat mentions feature, by default it it the same for everyone

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Removed

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: I'm now using `getOriginalStyledText()`, thanks for the heads up! its great to have the message as it was before wynntils touched it.

as for nicknames I've implemented what I think will work but I am not champion so I cannot nickname myself to test that unfortunately

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Using original message removes the yellow (or other colour possibilities as i found out) as a problem

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -94,10 +95,14 @@ public void onTitle(TitleSetTextEvent event) {
...
     public void onChatMessage(ChatMessageReceivedEvent event) {
         if (inBuffRoom()) {
-            WynntilsMod.info(event.getStyledText().getString());
-            Matcher matcher = event.getStyledText().getMatcher(RAID_CHOOSE_BUFF_PATTERN);
+            Matcher matcher = event.getOriginalStyledText().stripAlignment().getMatcher(RAID_CHOOSE_BUFF_PATTERN);
             if (matcher.matches()) {
                 String playerName = matcher.group(2);
+                // if the player is nicknamed
+                if (playerName.startsWith("§o")) {
+                    playerName = RevealNicknamesFeature.getNameAndNick(event).a();
```
ShadowCat117: This is not the right way to do this, we don't like features referencing each other like this.

I would suggest either add a new method to StyledTextUtils that will extract just the username or simply moving the `getNameAndNick` method to StyledTextUtils

ben9964: I will refactor this as soon as I'm home later today, I had a look at the architecture wiki page and couldn't see anything about not exposing parts of features publicly. Maybe I missed it though, a util function is perfect either way.

ShadowCat117: I don't know if the wiki is up to date, a lot has changed since it was originally made but the idea is that a feature should only rely on itself, any relevant Models/Services/Managers/Handlers and other util classes. It should never rely on something contained in another feature.

If 2 features are doing the same, or similar things like in this case, then it's best to move that logic somewhere else

Rafii2198: I don't think Wiki ever changed well except for addition of functions at some point, but the wiki is the same as it was when I joined which was during 1.18 days (there wasn't even a config screen at the time)

common/src/main/java/com/wynntils/features/chat/RevealNicknamesFeature.java
```diff
@@ -123,6 +125,40 @@ public void onPlayerChat(ChatMessageReceivedEvent event) {
...
+            if (nickname.get() == null && username.get() == null) {
+                return IterationDecision.CONTINUE;
+            }
+
+            return IterationDecision.BREAK;
+        });
+
+        return new Pair<>(username.get(), nickname.get());
+    }
+
```
kristofbolyai: I'll spare sending the same message as @ShadowCat117 did in https://github.com/Wynntils/Wynntils/pull/2827/files#r1748046610. TLDR, move this method, but also don't forget to use it in this feature itself. Currently you are just "duplicating" the method, and using it in another feature which breaks feature abstraction.

The code itself also needs to be refined a bit. `StylexText#iterate` is meant to be used for changing parts of a `StyledText`, this case, you are ignoring it. You can use `StyledText` with for-each styled loops as it has `iterator` implemented.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
kristofbolyai: Other than that. `playerName.startsWith("§o")` can be solved by an optional group in the regex, and a separate group for an unformatted name string.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -221,6 +250,22 @@ public RaidKind getCurrentRaid() {
         return currentRaid;
     }
 
+    public List<String> getRaidBuffMajorIds(String playerName) {
```
kristofbolyai:
```suggestion
    public List<String> getMajorIdRaidBuffs(String playerName) {
```

I think this wording makes more sense.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -221,6 +250,22 @@ public RaidKind getCurrentRaid() {
...
+
+        for (String buffName : buffNames) {
+            String majorId = this.currentRaid.majorIdFromBuff(buffName);
+            if (majorId == null) continue;
+
+            majorIds.add(majorId);
+        }
+
+        return majorIds;
+    }
```
kristofbolyai:
```suggestion
        if (!partyRaidBuffs.containsKey(playerName)) return List.of();

        List<String> buffNames = partyRaidBuffs.get(playerName);
        List<String> majorIds = new ArrayList<>();

        for (String buffName : buffNames) {
            String majorId = this.currentRaid.majorIdFromBuff(buffName);
            if (majorId == null) continue;

            majorIds.add(majorId);
        }

        return majorIds;
    }
```

I am not sure if this was an attempt of (premature) optimization, or you just coded it this way, but simple code is always better when performance is not an issue. 

You can optionally try to rewrite this using streams, so something like `playerRaidBuffs.stream().map(...)...`

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
```diff
@@ -5,34 +5,66 @@
...
             StyledText.fromString("§#5f968bffThe Canyon Colossus"),
-            Pattern.compile("^Calm the canyon's$")),
+            Pattern.compile("^Calm the canyon's$"),
+            Map.of(
+                    "Intrepid",
+                            Map.of(
+                                    1, "Heart of the Pack",
+                                    2, "Greed",
+                                    3, "Guardian"),
+                    "StoneWalker", Map.of(3, "Explosive Impact"))),
```
kristofbolyai: Isn't it "Stonewalker"?

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
```diff
@@ -55,6 +87,17 @@ public static RaidKind fromName(String name) {
         return null;
     }
 
+    public String majorIdFromBuff(String name) {
+        String[] split = name.split(" ");
+        if (split.length == 1) return null;
+
+        if (majorIdBuffs.containsKey(split[0])) {
+            return majorIdBuffs.get(split[0]).get(MathUtils.integerFromRoman(split[1]));
+        }
```
kristofbolyai: This will error if a new major id is added to the game, but not the list (so basically it make older versions of the mod not work). I suggest you move the tier parsing and error handling to the model itself, and just do a simple lookup here.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: This method does return the list of Major Ids though not the list of Raid buffs

common/src/main/java/com/wynntils/features/chat/RevealNicknamesFeature.java
ben9964: It's not possible to use the extracted util method as it is doing a different job from the one in `RevealNicknamesFeature`.

The one present there is performing an action for each hover part that *isnt* the nickname, while the util is just extracting the nick and username. I could obviously add a consumer to the method for a possible action to be done when dealing with parts of the styled text that aren't nickname parts, but this seems overcomplicated and messy.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: I'm happy with the refactor you suggested and have refactored that. I did just write it that way.

If streams is the convention that you'd like to follow (as I saw in the range visualizer itself) then I can do so but I think that there's no benefit to using them. With such little data streams are slower and a simple for loop is usually more readable too.

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
ben9964: Resolved

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
ben9964: I don't see how it would error when a new major id is added to a buff or a new buff is added to the game? If the buff isn't present in that raid's list (e.g. Test II was added) then it will return null and the loop in the calling method will just continue to the next iteration as if it didn't find a major id from that buff at all. 

If a major id is added to an existing buff (e.g. Hollowed I begins giving the player Explosive Impact), then it will also return null and continue in the calling method's for loop as if the major id doesn't exist. I don't see where the error would be raised.

common/src/main/java/com/wynntils/features/chat/RevealNicknamesFeature.java
kristofbolyai: I still think it's possible to create some method that can find, parse and return "nickname" StyledTextParts. Although, you'd have to move it to a model now, as `StyledTextUtils` is reserved for special, but not feature specific utility methods. (It's hard to imagine the boundaries, but since you need a nickname regex, we should keep the code in a model).

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -120,21 +122,36 @@ private void checkMajorIdCircles(Player player) {
...
-                    case "Guardian" -> Pair.of(CommonColors.RED.withAlpha(TRANSPARENCY), 7.9f);
-                    default -> null;
-                }))
+                .flatMap(gearInfo ->
+                        gearInfo.fixedStats().majorIds().stream().map(majorId -> circleFromMajorId(majorId.name())))
                 .filter(Objects::nonNull)
-                .toList();
+                .collect(Collectors.toList());
+
+        // add circles gained from raid major id buffs
```
kristofbolyai, line 129: Perhaps keep this a stream, make the code below also be a stream, concat them and then put them into a list. That'd keep the code style the same, and also shorten the code a tiny bit.

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
```diff
@@ -120,21 +122,36 @@ private void checkMajorIdCircles(Player player) {
...
+                }
+            }
+        }
 
         if (!circles.isEmpty()) {
             circlesToRender.put(player, circles);
         }
     }
 
+    private Pair<CustomColor, Float> circleFromMajorId(String majorIdName) {
```
kristofbolyai:
```suggestion
    private Pair<CustomColor, Float> getCircleFromMajorId(String majorIdName) {
```

More conventional, although it's more of a nit.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
kristofbolyai: `getRaidMajorIds`?

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
kristofbolyai: What if "Hollowed 2's Guardian" buff is moved to "Hollowed 3"? Basically, you check the first get, but not the second one.

common/src/main/java/com/wynntils/utils/mc/StyledTextUtils.java
```diff
@@ -198,4 +207,26 @@ public static StyledText unwrap(StyledText styledText) {
 
         return StyledText.fromParts(newParts);
     }
+
+    public static Pair<String, String> extractNameAndNick(StyledText styledText) {
```
kristofbolyai, line 211: I'll mark this for myself tomorrow. I may or may not want this in a model, rather than this util, as it's too specific to StyledText. It's likely that a model of some sort should hold this code.

ben9964: Fair enough I'll leave that to you then. I believe I've done what was necessary for this util function to work as expected.

With regard to the error if a major id was moved, if the major id was moved then the first level of the map in the code would still have Hollowed as a key but it would have 2 as the secondary map's key rather than 3 so the second get would return null and everything would continue as expected. calling get() on a map where the key isn't present simply returns null rather than raising an error. I believe a containsKey() check is implicit to the get call

common/src/main/java/com/wynntils/features/combat/RangeVisualizerFeature.java
ben9964: Resolved

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ben9964: Resolved

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
magicus: Also, we want to keep the `type` classes as free from logic as possible; ideally they should just be simple Records, Enums, Interfaces etc; just keeping data but not any "code". That should be handled by the corresponding Model.

common/src/main/java/com/wynntils/utils/mc/StyledTextUtils.java
magicus: Yeah, this could just live in RaidModel, right?

ShadowCat117: Although I can't think of any currently, there's probably other places where getting username is useful so RaidModel could work for now but not a permenant place for it imo

ben9964: It's definitely a util method rather than something that is exclusive to this Model or part of the game. Granted those use cases haven't presented themselves yet so I'm happy to move it to the RaidModel until that point if that's the consensus

kristofbolyai: Okay, let's keep this here for now.

common/src/main/java/com/wynntils/utils/mc/StyledTextUtils.java
```diff
@@ -198,4 +207,26 @@ public static StyledText unwrap(StyledText styledText) {
...
+            for (StyledText partText : partTexts) {
+                Matcher nicknameMatcher = partText.getMatcher(NICKNAME_PATTERN);
+
+                if (nicknameMatcher.matches()) {
+                    return new Pair<>(nicknameMatcher.group("username"), nicknameMatcher.group("nick"));
+                }
+            }
+        }
+
+        return new Pair<>(null, null);
```
kristofbolyai, line 230: I think in this case it's generally smarter to return null, as it makes error handling code nicer.

common/src/main/java/com/wynntils/models/raid/type/RaidKind.java
ben9964: Logic has been moved out of the Enum

kristofbolyai (approved):
Thanks for going through with all the reviews! This looks great now!

PR #2828, opened 2024-09-06
feat: Add Custom Chat Banner

ILikePlayingGames:
This chat banner is to be used for all messages sent by the mod to match the new Wynncraft chat style. A new utility class and a major refactor will be required to migrate all mod messages to use this banner.

# Notes
- The banner design is just a test texture to get it working. If anyone has a better texture (must be 8px wide by 9px tall), please suggest it below or on Discord.
- The spacing and formatting characters aside from the logo depend on Wynnpack's chat font. This will break if Wynncraft changes their font.
- The character `\uE100` was chosen to not interfere with Unicode (it's in the private use section) and Wynncraft's chat font.
- The spacing is the same as what Wynncraft uses for their banners, the only change is the use of our own banner character.
- To use the banner, all message components sent with the banner must use the tag `"font": "wynntils:chat"
  - Remember to switch back to the `default` font to send other characters (shown in test command below)
- Test the output in a command block using the following command: `tellraw @p [{"text":"\uDAFF\uDFFC\uE100\uDAFF\uDFFF\uE002\uDAFF\uDFFE", "color":"dark_green", "font":"wynntils:chat","extra": [{"text":" Wooooooo! Custom Wynntils banner!\n", "font":"default"},{"text":"\uDAFF\uDFFC\uE001\uDB00\uDC06","font":"wynntils:chat"},{"text":" Look, multiple lines!\n","font":"default"},{"text":"\uDAFF\uDFFC\uE001\uDB00\uDC06","font":"wynntils:chat"},{"text":" Another one!","font":"default"}]}]`

# Screenshot
image


build.gradle
```diff
@@ -109,6 +109,10 @@ allprojects {
         }
         json {
             target "src/**/*.json"
+            // schemaVersion must be at the top or else Fabric will complain
+            targetExclude("src/main/resources/fabric.mod.json")
+            // spotless replaces the Unicode escapes with the actual characters and rearranges the order, worsening readability.
```
magicus: The order should be alphabetical, even if it worsen readability. That's the only way to get consistency. I accept that you need to exclude it from spotless if it removes the escape sequence, but please let spotless order the file once first.

ILikePlayingGames:
image

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
```diff
@@ -168,4 +176,42 @@ public static MutableComponent makeObfuscated(
...
     }
+
+    /**
+     * Adds a Wynntils chat banner to the left side of the provided text in the style of Wynncraft 2.1 chat banners.
+     * The formatting of the provided text is preserved.
+     *
+     * @param formattedText the formatted text to add the Wynntils chat banner to
+     * @return a {@code Component} holding the formatted text with the Wynntils chat banner added
+     */
+    public static Component addWynntilsBanner(FormattedText formattedText) {
```
magicus: I don't get it..? This takes a FormattedText, not a Component? But it's called with a Component? How does this even compile?

magicus: Ah, right, FormattedText, not StyledText. Nevertheless, maybe it should be Component anyway, for consistency?

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
```diff
@@ -168,4 +176,42 @@ public static MutableComponent makeObfuscated(
...
+        // - 2 to have room for the banner character and a space on the left
```
magicus: What is this -2 referring to?

magicus:
As for the actual chat banner image, maybe we should check if we have any artists on the Content Team that can make something that looks a bit like our "shield" logo? It might be impossible in that small space, but if we could have something that resembles our logo I think that would be better. Otherwise this'll have to do.

kristofbolyai:
> As for the actual chat banner image, maybe we should check if we have any artists on the Content Team that can make something that looks a bit like our "shield" logo? It might be impossible in that small space, but if we could have something that resembles our logo I think that would be better. Otherwise this'll have to do.

I'll ask for that.

@ILikePlayingGames would you be interested in migrating all of our messages to this new "system", if we agree on an API?

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
ILikePlayingGames: It was originally component but the splitcomponent method didn’t work right. It split each character instead of splitting by my width, so I used another way of splitting. I’ll see if I can make it work with component.

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
ILikePlayingGames: Sorry that was probably left over from an earlier revision.

ILikePlayingGames:
@kristofbolyai I probably won't be able to migrate all of them at once but I can do some here and there.

kristofbolyai:
> @kristofbolyai I probably won't be able to migrate all of them at once but I can do some here and there.

Then my main problem with this PR becomes is, when and if we would ever use this.. Sadly everyone has limited time, so we can't expect anyone to port the messages. I do think the idea is good though, so maybe we can find a compromise of some sort.

magicus:
Huh? If this is done in `sendMessageToClient`, what else do we need to do to "migrate" messages?

ILikePlayingGames:
There’s also a number of messages that do not use that method for some reason. I’ve noticed a few during testing but I haven’t looked at where or why.

magicus:
So we need to find and fix those places? I say that can be done as a follow-up to this.

ILikePlayingGames:
Sure, I think we can merge it after we hear back about the banner art and fix the leftovers after.

kristofbolyai:
Oh, so that single method uses it, any only some random ones don't use this? In that case, it's more than fine. I've already talked to Arctwig a few days ago, we'll likely have a texture ready this weekend.

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
```diff
@@ -168,4 +177,38 @@ public static MutableComponent makeObfuscated(
...
+                    .append(Component.literal(" ")
+                            .withStyle(Style.EMPTY.withFont(ResourceLocation.withDefaultNamespace("default")))
+                            .append(formattedTextToComponent(lines.get(i))));
+            if (i != lines.size() - 1) {
+                output.append("\n");
+            }
+        }
+
+        return output;
+    }
```
kristofbolyai: Why not use `ComponentUtils#splitComponent` and add the formatting for every element in the list?

fabric/src/main/resources/fabric.mod.json
kristofbolyai, line 1: whatever happened here, can you revert it?

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
ILikePlayingGames: It doesn’t work with wrapping the last line due to #2854.

fabric/src/main/resources/fabric.mod.json
ILikePlayingGames: I’m worried spotless will redo it again if I try but I’ll try

ILikePlayingGames:
> Oh, so that single method uses it, any only some random ones don't use this? In that case, it's more than fine. I've already talked to Arctwig a few days ago, we'll likely have a texture ready this weekend.

Messages sent by the mod don’t use a consistent api to send them (`McUtils#sendMessageToClient`). We would have to fix the inconsistent ones to send the message that way. Also, thanks for asking about the texture!

Essentuan:
Would it not be better to create a new method (something like `WynntilsMod#sendMessage`)? I find it likely that at some point wynntils would like to send messages without the banner (such as TranslationFeature).

magicus:
> Would it not be better to create a new method (something like WynntilsMod#sendMessage)?

That would mean we would have to update all places to the current method. In fact, I'd say that we can assume that the current helper method is designed for sending Wynntils-marked messages.

If there is a place where this is incorrect (e.g. translated messages), we need to change those places instead, maybe have a sendRawMessage or sendMessageWithoutBanner.

magicus:
@ILikePlayingGames It seems something has gone wrong when you merged from main. There is a lot of unrelated changes now in this PR. Most often, this can be fixed by doing yet another `git merge main`, but if you have gotten yourself too deep into a git mess (happens to all of us!), it might be easier to just cherrypick the correct changes and start a new PR.

ILikePlayingGames:
@magicus I don’t see any extra charges. Are you referring to me fixing some strings and the build.gradle issue bolyai mentioned?

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -974,6 +975,12 @@
   "feature.wynntils.mythicBoxScaler.scale.description": "How much more should mythic boxes should be scaled by?",
   "feature.wynntils.mythicBoxScaler.scale.name": "Mythic Box Scale",
   "feature.wynntils.mythicFound.description": "Do various things when a mythic item is found.",
+  "feature.wynntils.mythicFound.lootChestDryBoxes": "%s dry boxes",
```
magicus: What about this then? Is it supposed to go into this PR?

ILikePlayingGames: Hmm that probably could have been separated yes. I was trying to add a banner to this string but found it wasn’t localized so I localized it.

magicus: Ok, I see, so the PR is intentionally a mix between different fixes.

Can I ask you to separate these two concerns? So there is one PR that just adds the chat banner, and one PR for localizing the messages, and other similar changes. I think the chat banner stuff is just 👍 but the other changes looked like it could do with some extra scrutinity. There were some regressions from StyledText back to Component, for instance, that I did not like. (In general, we're trying to move away from Component to StyledText).

ILikePlayingGames: Okay. I’ll try to separate them later.

common/src/main/java/com/wynntils/utils/mc/ComponentUtils.java
kristofbolyai: Okay, so we either fix that issue, or use StyledText. This solution does way too much work for essentially not much change for my liking, making me paranoid to accept this as a frequently used Util method.

fabric/build.gradle
```diff
@@ -105,3 +105,8 @@ sourcesJar {
     dependsOn commonSources
     from commonSources.archiveFile.map { zipTree(it) }
 }
+
+spotless.json {
+    // schemaVersion must be at the top or else Fabric will complain
+    targetExclude("src/main/resources/fabric.mod.json")
+}
```
kristofbolyai: I don't actually like this. Not sure if we requested this, or you thought it's a good idea, but I value all spotless rules being in a single place. Please revert it.

kristofbolyai:
I honestly can't still can't feel good about this PR. I like the idea, but I won't accept doing this with component modifications. I mentioned using StyledText before, however it was not clear in my head whether it is a requirement to this PR or not. Now, I think I am confident it is. I am also not too keen on combining splitting into lines and adding banners into the same method. 

I'll now try to write down precisely how I want this PR to turn out:
Create a new method in `StyledTextUtils` to split a single line component into multiple ones. The util class already has the multi-line to single-line processing step `StyledTextUtils#unwrap`. I want you to create `wrap`, a method that does the reverse operation, adding the soft-wrapping, then adding the banner. However, soft-wrapping should happen in it's own method, `StyledTextUtils#softWrap(StyledText text, int maxWidth)`, so the `wrap` method is basically just iterating (see `StyledText#iterate` for example) to add the banners.

build.gradle
```diff
@@ -109,8 +109,8 @@ allprojects {
         }
         json {
             target "src/**/*.json"
-            // schemaVersion must be at the top or else Fabric will complain
```
magicus, line 112: Why are you removing this exclude?

fabric/build.gradle
ILikePlayingGames: Sorry, I added it here since I couldn't get it to work in the main build.gradle file but I figured it out and I moved it.

build.gradle
ILikePlayingGames: Sorry, I could not get multiple excludes to work initially so this one was moved to the fabric build.gradle but I fixed the issue and moved it back.

ShadowCat117:
@ILikePlayingGames I'm planning to fix the translation feature working with chat banners soon and I will likely have to implement the wrap methods as mentioned by bolyai above so I may as well finish this pr at the same time if you have no issues with that?

ILikePlayingGames:
That's okay. Thank you so much!

magicus:
@kristofbolyai Did you get an alternative texture?

kristofbolyai:
I think we have one in the Discord ready.

ShadowCat117:
Superseded by https://github.com/Wynntils/Wynntils/pull/3969

Thank you for the initial work on this :)

PR #2829, opened 2024-09-06
fix: Fix lootrun notes not rendering and saving incorrectly [skip ci]

ShadowCat117:
Another place that needed this fix! Can't find it anywhere else so should be the last one


ryanzhoudev (approved):
Yep notes work now, nice

kristofbolyai (approved):

PR #2830, opened 2024-09-06
fix: Fix outline text shadow not rendering around vanilla hud elements [skip ci]

ShadowCat117:
The display mode was changed to normal to fix an issue with ImmediatelyFast causing the text to render as completely black, however I have tried running this fix with just ImmediatelyFast and with other mods too and have been unable to reproduce that issue so it appears to have been fixed on their side.

Normal:

See through:

See the top around boss bar and the right by scoreboard


ryanzhoudev (approved):
Looks like it works fine

kristofbolyai (approved):

kristofbolyai:
If an issue with IF comes back, we'll investigate further.

PR #2833, opened 2024-09-07
feat: Add auto attacking

xaryphon:
Allows you to just hold down attack to attack continuously instead of spamming.


ryanzhoudev (dismissed):
There needs to be some sort of system limiting the click speed of this feature, essentially it cannot give an advantage over the typical player, and most people click around 9 or 10 CPS max

xaryphon:
The attack speed is limited on the server side anyways so shouldn't really make that much of a difference, but it's now limited to 10 CPS.

ryanzhoudev:
> The attack speed is limited on the server side anyways so shouldn't really make that much of a difference, but it's now limited to 10 CPS.

Yeah, but spamming it this quickly essentially allows you to instantly attack when the weapon cooldown is up, and if you're spamming normally you can't time it that perfect

Also, for your solution, I would rather you do a % 2 check on the gametime (preferred), or move your variable check to the top. Not necessary that we only countdown when holding a weapon, since it's possible for you to click when not holding a weapon :). And just for defensiveness, you can do a > 0 check instead of != 0 if you really want to keep your countdown variable

xaryphon:
Modulo is definitely cleaner.
There are some edge cases where auto attacks are blocked, such as:
- if you start a spell cast but then wait for it to expire
- start a spell cast, open inventory and do stuff

but not really sure if these can be handled without a lot of infrastructure that (as far as I can tell) doesn't currently exist.

Essentuan:
Why not just add a separate keybind? Also epoch tstack is the main tstack build right now, so you definitely should add support for it.

ShadowShift:
10 cps is kinda an eyebrow raise btw

ryanzhoudev:
oh make it % 3 then

ShadowShift:
4.7 is fine since thats the attack speed. its an issue of going beyond it cause it can cause the queueing of main attacks iirc

ryanzhoudev:
> 4.7 is fine since thats the attack speed. its an issue of going beyond it cause it can cause the queueing of main attacks iirc

What about 5.0? Then we can do % 4 easily, otherwise we need to do % 5 for 4.0 cps

xaryphon:
> Why not just add a separate keybind?

To me it feels better to have it on the same button as attacking normally, but we could have an _option_ to have it on a separate key.

> Also epoch tstack is the main tstack build right now, so you definitely should add support for it.

Archer already seems to have auto attacking on the server-side. Does it not reach the weapon attack rate (I can't really test as I don't have archer levelled)?

Also regarding the game rules talk about allowing this kind of macros only for melee... so I'm not sure if this should be restricted to only warrior and assassin, but I asked the moderators for clarification regarding this and also what they feel about the attack rates.

Essentuan:
> 4.7 is fine since thats the attack speed. its an issue of going beyond it cause it can cause the queueing of main attacks iirc

10 CPS is fine mechanically since the in game cps cap is like 12-15. 

> Archer already seems to have auto attacking on the server-side. Does it not reach the weapon attack rate (I can't really test as I don't have archer levelled)?

No its terrible.

kristofbolyai:
Can someone here request a review when you figured out the specifics, TPS and implementation wise?

kristofbolyai:
So, any progress on this? Otherwise, I'll have to play the "evil maintainer" role, and close this PR.

ryanzhoudev:
> So, any progress on this? Otherwise, I'll have to play the "evil maintainer" role, and close this PR.

Waiting on max CPS allowed confirmation still

xaryphon:
I tried using the "Moderator Questions Form" found  [here](https://forums.wynncraft.com/threads/moderation-information-and-announcements-thread.302289/), but haven't gotten a response. So now I tried to ask through the #contact_staff channel on discord. If I don't get a response there I'll try DMing a mod directly.

ryanzhoudev:
I have confirmation that you can do whatever hits the 4.7. Or, you can ditch the onTick method completely, and trigger an attack as soon as the weapon cooldown is over. There's that ender pearl cooldown animation, which you should be able to mixin and detect.

xaryphon:
Okay so doing a % 4 is should be fine, but the reason I don't want to base it off on the cooldown animation is network delay. I typically have 120-170 ping (no European servers unfortunately), which means we would have to do network prediction for it to be anywhere near useful.

xaryphon:
Would you prefer Archer support be in this PR or a separate one?

kristofbolyai:
This PR.

Essentuan:
> I have confirmation that you can do whatever hits the 4.7. Or, you can ditch the onTick method completely, and trigger an attack as soon as the weapon cooldown is over. There's that ender pearl cooldown animation, which you should be able to mixin and detect.

If this pr tries to hit exactly 4.7 (or 5) cps or uses the enderpearl cd animation it will be next to useless since you're likely going to end up missing a melee (because you can't attack yet) or some other issue meaning you won't get your max DPS.

xaryphon:
Noticed I had the % == check the wrong way around also I decided to set it to mod 2, as anything higher would drop a little bit of dps and anything less causes the server to not register a right-click if pressed while auto-attacking.
I also don't seem to have a request for review button anymore, but it's now ready (imo).

magicus:
It is honestly a bit disturbing of having a PR updated after several weeks with a comparison changed from `==` to `!=`. This raises a lot of questions, about the quality of the code, and the testing done. Like, why was this not discovered earlier? Could the code have been written in another way so it would have been obvious from the start that this was wrong?

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.COMBAT)
+public class AutoAttackFeature extends Feature {
+    static final int TICKS_PER_ATTACK = 2;
+
+    int lastSelectedSlot;
```
magicus: These should be private

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,90 @@
...
+        if (event.getActionContext() != ArmSwingContext.ATTACK_OR_START_BREAKING_BLOCK) return;
+        if (event.getHand() != InteractionHand.MAIN_HAND) return;
+        if (Models.Character.getClassType() != ClassType.ARCHER) return;
+        lastSelectedSlot = McUtils.inventory().selected;
+        preventWrongCast = true;
+    }
+
+    @SubscribeEvent
+    public void onUseItem(UseItemEvent event) {
+        if (Models.Character.getClassType() == ClassType.ARCHER) return;
```
magicus, line 61: We just added a feature to swap archer left and right button. I'm not sure what happens here if that is enabled. I guess this works just fine. But if not, we might need to create a component that keeps track on what is left click and right click for each class, given the status of the invert feature.

In any case you need to enable that feature and test if the code here works.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,90 @@
...
+            if (!McUtils.options().keyAttack.isDown()) return;
+        }
+
+        LocalPlayer player = McUtils.player();
+        ItemStack heldItem = player.getItemInHand(InteractionHand.MAIN_HAND);
+
+        if (!ItemUtils.isWeapon(heldItem)) return;
+        if (ItemUtils.getItemName(heldItem).contains("Unidentified")) return;
+        for (StyledText lore : LoreUtils.getLore(heldItem)) {
+            if (lore.contains("✖")) return;
```
magicus: This is quite a lot of work to do for each tick. You need to validate the weapon once. Hopefully we have a event for when the user switches the current slot on the hotbar, or when a slot is updated with new content.

In fact, this sounds like something that belongs to a Model. Maybe PlayerModel, I'm not sure without looking at the code. We should have a `Optional<WynnItem> getCurrentWeapon`, which returns the weapon being held, or an empty optional if not. 

I'm not sure if you are up to adding such functionality..?

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        if (player.tickCount % TICKS_PER_ATTACK != 0) return;
+
+        if (isArcher) {
+            SpellDirection.RIGHT.getSendPacketRunnable().run();
+        } else {
+            player.swing(InteractionHand.MAIN_HAND);
```
magicus, line 107: This asymmetry just looks weird. Why does not `player.attack()` or `SpellDirection.LEFT...` work in the opposite case?

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,90 @@
...
+
+        LocalPlayer player = McUtils.player();
+        ItemStack heldItem = player.getItemInHand(InteractionHand.MAIN_HAND);
+
+        if (!ItemUtils.isWeapon(heldItem)) return;
+        if (ItemUtils.getItemName(heldItem).contains("Unidentified")) return;
+        for (StyledText lore : LoreUtils.getLore(heldItem)) {
+            if (lore.contains("✖")) return;
+        }
+        if (player.tickCount % TICKS_PER_ATTACK != 0) return;
```
magicus: This is a quick short-circuit, that should be tested as early as possible.

Also, is this really correct? I mean, now you will only do the attacking on even-numbered ticks. But if an attack takes two ticks, and the user started their attack on an odd-numbered tick, then the next attack should come 2 ticks afterwards, but now you will wait 3. Maybe it's not a big deal, you will lose a tick with 50% probability at the start of holding down the attack button, but if you think this is an acceptable trade-off for code simplicity, this should be acknowledged and documented here. Otherwise, you should record the tick the attack started, and check for that instead of player.tickCount. I think it makes not much extra work and will make the feature work correctly as advertised.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: Just tested it and it works as expected.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: Changed it so that it only checks when the items have changed (or slot changed). As for adding it to a Model, unfortunately I don't really have too much spare time anymore so I'd rather not.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: `player.attack` seems to require a target entity (which we don't have). `SpellDirection.LEFT` doesn't do the swing animation, so it looks weird imo.

As for replacing `SpellDirection.RIGHT` I didn't see a `player.use` so this was simpler. While we could send a the packet manually, I don't really see a point in duplicating the code that SpellDirection has.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: > Also, is this really correct?

I don't think it's wrong per se, but without more in-depth knowledge on how the server handles the attacks I think anything more complex would be unnecessary speculation.

> will make the feature work correctly as advertised.

The language file doesn't advertise how it continues attacking, so I don't think this is a problem.

xaryphon:
Rebased it as I tested the button swap (not sure if there was a better way than force pushing).

> This raises a lot of questions, about the quality of the code, and the testing done. Like, why was this not discovered earlier?

Depends how dramatic you want to be, but mistakes happen. As for why I didn't notice, when I first added it, it was modulo 2 and worked as expected (might have also had the check be reversed, as in do the attack instead of return). Then when I increased the modulo I just took a quick look and didn't test attacks per second.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.COMBAT)
+public class AutoAttackFeature extends Feature {
+    private static final int TICKS_PER_ATTACK = 2;
+
+    private Boolean canUseWeapon = null;
```
magicus: Don't use `Boolean` with `null` as a tri-state. Either make an nested enum, or two different booleans, one which says if this is checked, and one what the value is.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,114 @@
...
+        lastSelectedSlot = McUtils.inventory().selected;
+        preventWrongCast = true;
+    }
+
+    @SubscribeEvent
+    public void onSpellCastCompleted(SpellEvent.Completed event) {
+        preventWrongCast = false;
+    }
+
+    private boolean checkWeapon() {
```
magicus:
```suggestion
    private boolean isHoldingWeapon() {
```

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
magicus: Can you write this down in the code? Just something like `// SpellDirection.LEFT doesn't do the swing animation` would go a long way, to document the reason of these different methods.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: Changed the name to isHoldingUsableWeapon as it doesn't only check if we're holding a weapon but also if it's can use it.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
xaryphon: Not sure on the enum member names, but probably good enough (also went with NOT_USABLE instead of UNUSABLE for a little bit better readability at a quick glace).

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,121 @@
...
+        lastSelectedSlot = McUtils.inventory().selected;
+        preventWrongCast = true;
+    }
+
+    @SubscribeEvent
+    public void onUseItem(UseItemEvent event) {
+        if (Models.Character.getClassType() == ClassType.ARCHER) return;
+        lastSelectedSlot = McUtils.inventory().selected;
+        preventWrongCast = true;
+    }
```
ShadowCat117: I think you should be able to combine these using the `SpellEvent.Partial` event, then you won't need to do any archer checks.

Also, if you start casting a spell but do not finish then the feature will not work until you complete a spell so if a partial spell is in progress you should set `preventWrongCast` back to false after the spell inputs have expired, I believe it lasts for 3 seconds

ShadowCat117: And the same goes for a failed spell actually, so use `SpellEvent.Failed` for that

xaryphon: `SpellEvent` is based on a packet being received from the server and therefore not really usable for this use case.
`SpellEvent.Completed` is used as, without good network prediction, it's the only way to check for spell completion (even though it can drop a small amount of DPS due to network delay). `SpellEvent.Completed` is also called for failed casts as well.

Added a timeout for the spell cast, though it can cast a spell in some rare cases (i.e. user is holding down left click while the spell cast is expiring).

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private boolean isHoldingUsableWeapon() {
+        ItemStack heldItem = McUtils.player().getItemInHand(InteractionHand.MAIN_HAND);
+        if (!ItemUtils.isWeapon(heldItem)) return false;
+        if (ItemUtils.getItemName(heldItem).contains("Unidentified")) return false;
+        for (StyledText lore : LoreUtils.getLore(heldItem)) {
+            if (lore.contains("✖")) return false;
+        }
+        return true;
+    }
```
ryanzhoudev: Still not a fan of this entire function - it's a duplicate of information that already exists.
Since we now save the usability of the weapon, can you look into using parsing the weapon as a WynnItem and using .meetsActualRequirements()?

Example at InventoryModel.java Line 77-81, if you need help.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,122 @@
...
+        } else {
+            lastSelectedSlot = currentSelectedSlot;
+            preventWrongCast = Integer.MIN_VALUE;
+        }
+
+        if (player.tickCount % TICKS_PER_ATTACK != 0) return;
+
+        if (weaponStatus == WeaponStatus.UNKNOWN) {
+            weaponStatus = isHoldingUsableWeapon() ? WeaponStatus.USABLE : WeaponStatus.NOT_USABLE;
+        }
```
ryanzhoudev, line 97: Do you think we need to even run this `== UNKNOWN` every tick? It's up to you since the check is cheap, but I think the chances of someone collecting a buff while holding an unusable weapon, then being able to use it, is incredibly low. Perhaps just do it once on hotbar switch?

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,122 @@
...
+            if (!McUtils.options().keyUse.isDown()) return;
+        } else {
+            if (!McUtils.options().keyAttack.isDown()) return;
+        }
+
+        if (isArcher) {
+            SpellDirection.RIGHT.getSendPacketRunnable().run();
+        } else {
+            // SpellDirection.LEFT doesn't do the swing animation
+            player.swing(InteractionHand.MAIN_HAND);
```
ryanzhoudev: Why not combine these if's, then you don't need to save the isArcher boolean?

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final int TICKS_PER_ATTACK = 2;
+    private static final int SPELL_TIMEOUT_TICKS = 60;
+
+    private enum WeaponStatus {
+        UNKNOWN,
+        NOT_USABLE,
+        USABLE
+    }
+
+    private WeaponStatus weaponStatus = WeaponStatus.UNKNOWN;
```
ryanzhoudev: I'm not sure of leaving the enum definition up here (maybe move to bottom of file?), but if @kristofbolyai or @magicus says it's fine, no objection

magicus: I agree. Nested classes (which this technically is) should be at the bottom of the file.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
magicus: The costly check is just done once, if it is in an unknown state, then it will be either usable or nonusable. The comparison is cheap and can remain. I think this is a reasonable way to solve the problem of not doing too many checks, but still doing enough. If some pre-condition changes, we flag the status as being unknown, and then we re-check it once next tick.

common/src/main/java/com/wynntils/features/combat/AutoAttackFeature.java
magicus: Good point. :-)

ryanzhoudev (approved):
I think this is fine now.

magicus (approved):
I agree. Let's ship this.

magicus:
@Xaryphon Thank you for your contribution! And for your patience in tweaking and polishing this patch.

PR #2837, opened 2024-09-08
fix: Fix custom territory management screen

ShadowCat117:
Handling multiple pages of territories may need to/could be changed but I'm unable to test that


kristofbolyai:
I think SetContent vs SetSlot matters for multi-page handling sadly. Let's try to test this.

Essentuan:
I just tested and it does work (though it still takes like .5s for two pages)

kristofbolyai (approved):
Looks good then.

PR #2839, opened 2024-09-08
fix: Fix issues with slot rendering [skip ci]

ShadowCat117:
First commit is a better fix than https://github.com/Wynntils/Wynntils/pull/2780

2nd is a hacky fix for item text overlays not rendering correctly in containers that use a background from a font. (bank, tm etc)


kristofbolyai (approved):
As discussed in Discord, this is the only fix we can do at the moment. Ideally, we'd have to clean up how we use FontRenderer to make this much nicer.

PR #2843, opened 2024-09-09
fix: Handle null acquired and no treasury in guild map [skip ci]

ShadowCat117:


Essentuan:
Territories are always at least 'Very Low' (or at least should be) so I think it would be better to just replace none with vlow here. (unless this changed very recently none territories have gotten medium treasury before)

ShadowCat117:
GuildResourceValues handles both treasury and defences, the problematic territory currently has a none treasury
image

ShadowCat117:
API appears to be fixed now so it's no longer causing issues but still worth including this in case it happens again

Essentuan:
weird

common/src/main/java/com/wynntils/models/territories/type/GuildResourceValues.java
```diff
@@ -1,12 +1,13 @@
...
- * Copyright © Wynntils 2022-2023.
+ * Copyright © Wynntils 2022-2024.
  * This file is released under LGPLv3. See LICENSE for full license details.
  */
 package com.wynntils.models.territories.type;
 
 import net.minecraft.ChatFormatting;
 
 public enum GuildResourceValues {
+    NONE("None", ChatFormatting.WHITE, 0),
```
kristofbolyai, line 10: Does the API actually return "NONE", or where does this get set?

ShadowCat117: This is from the advancements menu I believe, which displays "None" as seen in the image I posted above

kristofbolyai: Ah!

kristofbolyai (approved):
Wynn moment.

PR #2844, opened 2024-09-09
feat: Rewrite ChatTimestampFeature

Essentuan:
This PR rewrites ChatTimestampFeature to render the timestamp separately from the message. This improves timestamps so they don't ruin Wynncrafts formatting. Tested on Fabric & Forge. This could be improved a little bit by calling ChatComponent.refreshTrimmedMessages whenever the timestamp format is updated so that they're always the proper format.

I'm not entirely sure whether this is inside the scope of Wynntils, but figured I may as well make it. 

image
image


common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
```diff
@@ -22,9 +41,151 @@ public abstract class ChatComponentMixin implements ChatComponentExtension {
...
+
+    @WrapOperation(
+            method = "addMessageToDisplayQueue",
+            at = @At(value = "INVOKE", target = "Ljava/util/List;add(ILjava/lang/Object;)V"))
+    private void addMessageToDisplayQueue(
+            List<GuiMessage.Line> instance,
+            int i,
+            Object line,
+            Operation<Void> original,
+            @Local(name = "j") int index,
```
kristofbolyai: How many integers are used? Perhaps it'd be wiser to use the field's ordinal.

While writing this comment, I've checked Mixin docs,

> Names of the local variable. This will not work on obfuscated code like Minecraft. For targeting unobfuscated code this is normally the least brittle option.

Can you convert both of the use cases to ordinals?

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
```diff
@@ -22,9 +41,151 @@ public abstract class ChatComponentMixin implements ChatComponentExtension {
...
     }
+
+    @Unique
+    private ChatTimestampFeature getTimestampFeature() {
+        if (timestampFeature == null) {
+            timestampFeature = Managers.Feature.getFeatureInstance(ChatTimestampFeature.class);
+        }
+
+        return timestampFeature;
+    }
```
kristofbolyai: The mixin looks good, but unfortunately accessing a manager instance like this shouldn't be done, especially from a mixin. What I'd propose is creating a new event, `RenderChatTimestampsEvent`, with it being disabled by default, and `ChatTimestampFeature` can listen to it and enable it. That easily solves the "reverse" dependency situation you have here.

kristofbolyai (commented):
In my opinion this looks really exciting! As expected, most of my review comments are in relation to the mixins...

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
Essentuan: Yup my bad, I forgot about this. (Though funnily enough I think using name here technically would work since those are the obfuscated names; obviously a bad practice though.)

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
Essentuan: ok I've swapped it to use ChatTimestampFeature.RenderEvent. Not resolving for now in case you have a better idea on which package the event should be in.

kristofbolyai: Just create a new event, and put it in the appropriate package in near the mixins package, inner class events are weird. As suggested above, `RenderChatTimestampsEvent` works, or any other describing name. One might ask why create an unique event for just this single use case, but this how the mod's architecture is.

kristofbolyai:
Currently I don't have any review comments but I'd like to check this out tomorrow to see if anything comes up in a different state of mind.

magicus:
It looks like the patch is doing a bit too much work in ChatComponentMixin.java... Rendering and stuff should likely reside with the Feature instead.

Essentuan:
Using mixins here is simply the best way to ensure most mods that modify chat component rendering are compatible with Wynntils. The only other option here would be to post an event for each line in chat which would be terrible for performance.

magicus:
Events are high-performance. Once per chat line is *nothing*; we sends hundereds of events per second when you e.g. open your inventory.

Also, this is the rules for optimization. They apply to all software projects in general, but definitely to Wynntils specifically:

1) Don't optimize.

2) (Only for experts) Don't optimize yet.

That is, you should not write bad code out of a guess that it will perform badly. Write nice code, and then measure. If there is a performance issue, then -- and only then -- are you allowed to uglify the code.

magicus:
And our architectural design says we should to just that: mixins intercept minecraft code and sends event. Listeners (features or models) can act and do rendering or whatever when they receive an event, and if they need to communicate values back, they can modify the event.

You can get the exactly same functionality as you have now, but you just need to refrain from doing anything else than preparing a suitable event in the mixin. There is no greater risk for mod incompability by doing it this way, it is just a way to make sure we have proper separation of concern between injecting code in vanilla, and having a single point of responsibility for related modifications (the feature or model).

Essentuan:
> That is, you should not write bad code out of a guess that it will perform badly. Write nice code, and then measure. If there is a performance issue, then -- and only then -- are you allowed to uglify the code.

You should uglify BufferedFontRenderer, FontRenderer, and Templates (I don't even have that many TextOverlays enabled yet simply checking if they should be rendered is 5% of each frame, minimap is only 4%, and overlays as a whole are 10%.). I'm not saying premature optimization is good, its not. But simply being aware of performance (especially in templates) probably could've shrunk overlay rendering to less than 5% of the frame.

> You can get the exactly same functionality as you have now, but you just need to refrain from doing anything else than preparing a suitable event in the mixin. There is no greater risk for mod incompability by doing it this way, it is just a way to make sure we have proper separation of concern between injecting code in vanilla, and having a single point of responsibility for related modifications (the feature or model).

However there are mods that allow you to extend chat, and make the unfocused/focused height higher. The best way to do what this PR does with events is to post an event when the chat renders the background for each line & when it renders the text for that line (and one to disable/enable rendering of course). If a player extends their chat box there could be hundreds of lines being rendered. And having 2 events per line per frame is a lot. And if you want to make every mixin an event you'd need 5 events, where 2 of those are 2 lines right now (and probably could be 1). In addition these events would need to be hyper specialized since we only include the locals we need meaning they wouldn't be useful outside of rendering timestamps. 

For the record I think the slot render event (whatever its called) is fine, sure its called often but it used by multiple features. These events could only be used by chat timestamps.

common/src/main/java/com/wynntils/mc/event/ChatComponentRenderEvent.java
```diff
@@ -0,0 +1,144 @@
...
+
+        public GuiGraphics getGuiGraphics() {
+            return guiGraphics;
+        }
+
+        public int getX() {
+            return x;
+        }
+
+        public int getO() {
```
magicus: What are these? The names are highly non-descriptive.

magicus:
I don't have much more time to look at this tonight, but I get the feeling that there is something non-optimal about the approach. Are we really doing this the right way? Can't you inject code elsewhere in Vanilla to shift the chat messages to the right on the screen, for instance?

Essentuan:
> I don't have much more time to look at this tonight, but I get the feeling that there is something non-optimal about the approach. Are we really doing this the right way? Can't you inject code elsewhere in Vanilla to shift the chat messages to the right on the screen, for instance?

Anywhere else would need an additional mixin since you'd have to push and pop the posestack, no?

common/src/main/java/com/wynntils/mc/mixin/GuiMessageLineMixin.java
```diff
@@ -0,0 +1,56 @@
...
+    @Unique
+    private Optional<Pair<Component, Integer>> timestamp;
+
+    @Inject(method = "<init>", at = @At("TAIL"))
+    private void init(
+            int addedTime, FormattedCharSequence content, GuiMessageTag tag, boolean endOfEntry, CallbackInfo ci) {
+        timestamp = Optional.empty();
+    }
+
+    @Override
```
kristofbolyai, line 37: These methods could be annotated `@Unique` as well (not that it should matter on any modern mod loader).

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
```diff
@@ -22,9 +35,104 @@ public abstract class ChatComponentMixin implements ChatComponentExtension {
...
+    private void addMessageToDisplayQueue(
+            List<GuiMessage.Line> instance,
+            int i,
+            Object line,
+            Operation<Void> original,
+            @Local(ordinal = 1) int index,
+            @Local(argsOnly = true) GuiMessage message) {
+        MixinHelper.post(new AddGuiMessageLineEvent(message, (GuiMessage.Line) line, index));
+
+        original.call(trimmedMessages, i, line);
```
kristofbolyai, line 59: Does the event have some side effects that change anything in this method? If so, we likely have to express that better. If not, this could be turned into a simple Inject.

common/src/main/java/com/wynntils/features/chat/ChatTimestampFeature.java
```diff
@@ -43,20 +47,84 @@ protected void onConfigUpdate(Config<?> config) {
...
+                        .withStyle(ChatFormatting.GRAY))
+                .append(Component.literal("] ").withStyle(ChatFormatting.DARK_GRAY)));
+    }
+
+    @SubscribeEvent
+    public void onChatComponentRenderPre(ChatComponentRenderEvent.Pre event) {
+        timestampWidth = 0;
+
+        for (GuiMessage.Line line : ((ChatComponentAccessor) event.getChatComponent()).getTrimmedMessages()) {
+            GuiMessageLineExtension extension = (GuiMessageLineExtension) (Object) line;
```
kristofbolyai, line 72: Do you need to cast to object here? Can't you just go to the extension directly? (Iiirc, we only needed the object cast in mixins, when working with `this`.

common/src/main/java/com/wynntils/features/chat/ChatTimestampFeature.java
```diff
@@ -43,20 +47,84 @@ protected void onConfigUpdate(Config<?> config) {
...
+
+    @SubscribeEvent
+    public void onChatComponentTranslate(ChatComponentRenderEvent.Translate event) {
+        if (timestampWidth != 0) event.setX(event.getX() + 4 + timestampWidth);
+    }
+
+    @SubscribeEvent
+    public void onChatComponentMapMouseX(ChatComponentRenderEvent.MapMouseX event) {
+        if (timestampWidth != 0) event.setX(event.getX() - (4 + timestampWidth));
+    }
```
kristofbolyai: Can we get some 1-liner comments on what these do (and why they differ slightly in calculation)? Also, why you are at it, can you put in braces? We don't like not putting them for anything that's not code flow-control, like return/break/continue.

kristofbolyai (commented):
Arguably, the "proper" way to do this loses a bit of code readability thanks to the events, but not much, considering you have to use 3-4 mixin methods anyways.

common/src/main/java/com/wynntils/features/chat/ChatTimestampFeature.java
Essentuan: You have to cast to object because GuiMessage & GuiMessage.Line are final. (Or at least I think so, without casting to object it just doesn't compile.)

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
Essentuan:
```
this.trimmedMessages.add(0, new GuiMessage.Line(message.addedTime(), formattedCharSequence, message.tag(), bl2));
```

This is what the original method call is, I don't currently know of a way to get the GuiMessage.Line with a normal inject (I am also not an expert on mixins so)

magicus:
> Anywhere else would need an additional mixin since you'd have to push and pop the posestack, no?

If we could drop all other changes you are proposing here, and just add a new mixin with a push and pop of the posestack, then I think that would be much better, yes. That would be a clean injection, with less risk of bad interaction with other mods.

Then your timestamp functionality basically comes down to:
1) shift the vanilla chat to the right
2) draw your own timestamps

I'd say that such a solution is vastly superior, if it is possible.

Essentuan:
> If we could drop all other changes you are proposing here, and just add a new mixin with a push and pop of the posestack, then I think that would be much better, yes. That would be a clean injection, with less risk of bad interaction with other mods.

Nothing would change about this PR except an additional mixin at the end of the method to pop the posestack.

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
kristofbolyai: Okay, then this looks reasonable.

magicus:
I don't have more time to spend on this PR. I think it looks convoluted, with logic dispersed, and a lot more complicated than the old solution. My gut feeling is still that it does not need to be this way; but I don't have the time to dig into vanilla code myself to understand how to do this properly.

I don't like this PR, but I'll accept it if @kristofbolyai does. I'll leave it to him to decide. (To be clear: I am not objecting to the functionality itself, but that I have not been convinced that this complex solution is the only way possible to achieve this.)

common/src/main/java/com/wynntils/mc/mixin/ChatComponentMixin.java
kristofbolyai, line 1: This mixin is still missing the `@Unqiue` for extension methods + the semi-qualified mixin names I've talked about somewhere in this PR before.

kristofbolyai (commented):
I understand @magicus' view on how this feature is quite complicated, however, the end result is actually something very nice. 

Essentially I feel like this change transforms the kind of short and nice feature code into a bit of "screen" like code where you have to render and calculate a lot of screen coordinates precisely. What if you transformed this PR just into that? Even though `ChatComponent` is not a `Screen` by definition, you can just treat it like that. What that means is just create a `WynntilsChatComponent` (or an alternative name would be `TimestampedChatComponent`) and override these methods regularly there. I think that should be simple enough, and the only mixin we'd need is to replace the `ChatComponent` instances with our version. 

However, I do understand how these changes might have to be whole method overrides, as I'm sure Mojang's original `ChatComponent` code was not meant to be extensible. What that costs us, apart from readability, is mod compatibility with some other mods. However, it is likely that the mod itself wouldn't break, but just straight up wouldn't work, in some very specific cases. I think we can live with that. 

What do you think @Essentuan?

Essentuan:
> What do you think @Essentuan?

I think the community backlash from breaking existing chat mods (such as [ChatShot](https://modrinth.com/mod/chatshot); see [here](https://github.com/DeDiamondPro/ChatShot/blob/master/src/main/java/dev/dediamondpro/chatshot/mixins/ChatHudMixin.java), [MoreChatHistory](modrinth.com/mod/morechathistory); see [here](https://github.com/JackFred2/MoreChatHistory/blob/main/src/client/java/red/jackf/morechathistory/mixins/ChatComponentMixin.java)) negates any benefit of improving code readability. Unfortunately I don't think mc modding has the tools to do this properly. MoreChatHistory (and mods like it) breaking with Wynntils would be especially bad.

If we were to use a custom ChatComponent (I would just override methods in ChatTab personally) we'd have to rewrite `ChatComponent#addMessageToDisplayQueue` and `ChatComponent#render`.

magicus:
An alternative is to just say that to get chat timestamps another mod is needed. It is a borderline case for Wynntils in the first place.

kristofbolyai:
The examples brought up by Essentuan are mods that are likely to break, confirming my fears and making my suggestion not maintainable and practical in the real world.

Generally, chat tabs, chat timestamps, and chat message stacking all are really problematic chat-extension features which all require their share of not-so-great code.

I'm leaning towards accepting this code and keeping chat timestamps a Wynntils feature (along with chat tabs). While it does have 3-4 different mixins and events for basically a "single change", the use of events makes it not break any hard rules set by Wynntils' code structure, rather, it just uses them in a different way to achieve mod compatibility. While I'd be against most other PRs doing this, I'm ready to accept this one specifically, as I think this is just before that "line" we shouldn't cross. **However, I can say, with most certainty, that if this PR causes porting issues for _every single Minecraft version change_, just like the old resource pack pre-loading did, I am ready to either remove this feature, or rewrite it in an alternative, likely not so mod-cross-compatible way**.

I can't say the same for, for example, "message stacking", a legacy feature that could stack and add "[x2]" after repeating messages in chat. That feature's implementation does cross that theoretical line, and I consider it's implementation "great it works", not, "it works greatly".

I think that the best we could do is spend just 1 more evening with this PR, meaning I'll make my final decision by tonight, or by tomorrow.

kristofbolyai (approved):
Let's roll with this for now. Sorry for making this PR so hard to finish, chat-related features always get into this kind of gray area...

Thank you for getting this PR through.

magicus:
I think your reasoning is sound and reasonable.

PR #2845, opened 2024-09-09
feat!: Introduce the new data dependency loading systems

kristofbolyai:
This branch collects all PRs related to the refactors in data loading, which are aimed to improve dtra loading both systematically and for the user with new UIs.


kristofbolyai:
@ShadowCat117 would you be interested in creating some user-facing system for this PR? Either a download progress notification at the top of the screen, like resource packs, some kind of UI, or even some commands, alternatively. I do think that "click to reload data" is possible with this PR, so maybe we could add an UI for that. However, it might never be used, so maybe just a notification and a "not all of your data downloaded correctly" popup is enough.

ShadowCat117:
Yeah sure I can look into something

magicus:
Or maybe just a semi-transparent spinner while we are still downloading data? A progress bar might be tricky since we'd basically need to know how many bytes to download. Downloading the map tiles will take long time, and downloading e.g. label json will be quick.

PR #2847, opened 2024-09-09
refactor: dont return a pair of nulls [skip ci]

ben9964:
Small refactor that adds an early return if the player is null as well as returning null rather than a boxed pair.


kristofbolyai:
Thanks, I really just forgot to check that comment..

PR #2850, opened 2024-09-10
feat: Poi Creation Screen redesign [skip ci]

ShadowCat117:
This is essentially the improvements from https://github.com/Wynntils/Wynntils/pull/2553/ that don't rely on MapData. (Better icon selection, set coords to location/map center)

A few changes are also leftover from the MapData version like using Location instead of PoiLocation and setting 0 as a default Y.


magicus:
It's sad that the MapData effort got stalled. :( In the meantime, I agree that it is better to not waste the improvements done to the UI. I did not look specifically into the code, but from what I saw this looks good.

ShadowCat117:
1.21/Wynncraft 2.1 is pretty stable atm, only a few minor issues for 2.1 remain so if the main MapData branch is updated from main I can start working on the UI again, I know bolyai is quite busy atm so not sure he'll have time to finish up what still needs doing but that just gives me more time to polish the UI.

magicus:
Yeah, I started looking at the MapData branch. It has diverged a bit from the main branch, but that is not the biggest issue. We need to redesign a core aspect about how attributes are inherited and shared across locations / paths / areas and "subclasses" of these, that didn't work out the way I have thought. And there was no easy solution either; we will have to sacrifice flexibility, or readability, or something like that no matter how we do. :(

kristofbolyai:
> Yeah, I started looking at the MapData branch. It has diverged a bit from the main branch, but that is not the biggest issue. We need to redesign a core aspect about how attributes are inherited and shared across locations / paths / areas and "subclasses" of these, that didn't work out the way I have thought. And there was no easy solution either; we will have to sacrifice flexibility, or readability, or something like that no matter how we do. :(

MapData basically stalled because the two of us couldn't make progress on this attribute inheritance part. And now, I have a bit less time for it, so I can only work on it every so often, and only after I'm done with the downloading rewrites.

kristofbolyai (approved):
Looks good!

magicus:
> MapData basically stalled because the two of us couldn't make progress on this attribute inheritance part. 

Yeah, I remembered that after going back and reading up on the comments. I'll see if I can try to get some time to think on that a bit, so we can get it done.

PR #2852, opened 2024-09-10
feat: Handle hash conflicts in url files

kristofbolyai:
This PR aims to handle hash conflicts in the 3 versions of our URL list in a way that we always aim to re-download the data, if we are not "sure" our hash is correct. During the PR, I also discovered that loading data sources in constructors are VERY race-prone, as it's very likely the constructors run before the remote urls finished downloading, essentially limiting the mod to only use the local cache/bundled urls.json file.

### DIFF
--- a/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
+++ b/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
@@ -163,7 +163,7 @@ public class WynntilsCommand extends Command {
                         false);
 
         Services.Hades.tryDisconnect();
-        Services.WynntilsAccount.reauth();
+        Services.WynntilsAccount.reloadData();
         Models.Player.reset();
         // No need to try to re-connect to Hades, we will do that automatically when we get the new token
 
@@ -213,6 +213,8 @@ public class WynntilsCommand extends Command {
                         () -> Component.translatable("command.wynntils.reloadCaches.reloading")
                                 .withStyle(ChatFormatting.YELLOW),
                         false);
+
+        // FIXME: Special URL reloading
 
         // Reload all downloaded data
         WynntilsMod.reloadAllComponentData();
--- a/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
+++ b/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
@@ -214,12 +214,11 @@ public class WynntilsCommand extends Command {
                                 .withStyle(ChatFormatting.YELLOW),
                         false);
 
+        // This reloads all URLs, and will then trigger a re-download by the DownloadManager
         Managers.Url.loadUrls();
 
         // Reload all downloaded data
         WynntilsMod.reloadAllComponentData();
-
-        // Explicitly, do NOT call DownloadManager here, as it will get called when UrlManager finishes reloading
 
         return 1;
     }
--- a/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
+++ b/common/src/main/java/com/wynntils/commands/WynntilsCommand.java
@@ -214,11 +214,12 @@ public class WynntilsCommand extends Command {
                                 .withStyle(ChatFormatting.YELLOW),
                         false);
 
-        // FIXME: Special URL reloading
+        Managers.Url.loadUrls();
 
         // Reload all downloaded data
         WynntilsMod.reloadAllComponentData();
-        Managers.Download.download();
+
+        // Explicitly, do NOT call DownloadManager here, as it will get called when UrlManager finishes reloading
 
         return 1;
     }
--- a/common/src/main/java/com/wynntils/core/net/UrlManager.java
+++ b/common/src/main/java/com/wynntils/core/net/UrlManager.java
@@ -190,7 +190,7 @@ public final class UrlManager extends Manager {
         // unlike other core components.
     }
 
-    private void loadUrls() {
+    public void loadUrls() {
         // Reset the URL lists
         urlListMap.clear();
         urlList = UrlList.EMPTY;
--- a/common/src/main/java/com/wynntils/core/net/UrlManager.java
+++ b/common/src/main/java/com/wynntils/core/net/UrlManager.java
@@ -112,12 +112,10 @@ public final class UrlManager extends Manager {
     // -Dwynntils.url.list.force.source=local
     private static final String URLS_OVERRIDE_FORCE_SOURCE_ENV = "wynntils.url.list.force.source";
     private static final String FORCE_SOURCE = System.getProperty(URLS_OVERRIDE_FORCE_SOURCE_ENV);
-    private final UrlListType forceSource;
+    private final UrlMapperType forceSource;
 
-    private Map<UrlListType, UrlList> urlListMap = new ConcurrentHashMap<>();
-
-    private Object urlMapLock = new Object();
-    private UrlList urlList = UrlList.EMPTY;
+    private Map<UrlMapperType, UrlMapper> urlMappersByType = new ConcurrentHashMap<>();
+    private UrlMapper urlMapper = UrlMapper.EMPTY;
 
     public UrlManager(NetManager netManager) {
         super(List.of(netManager));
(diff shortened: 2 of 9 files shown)


kristofbolyai:
This has a few bugs + TODOs regarding custom JVM args, but essentially, the patch shouldn't change much in other regards.

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -64,11 +97,18 @@ public final class UrlManager extends Manager {
     // -Dwynntils.url.list.ignore.cache=true
     private static final String URLS_IGNORE_CACHE_ENV = "wynntils.url.list.ignore.cache";
 
+    // -Dwynntils.url.list.debug.log=true
+    private static final String URLS_DEBUG_LOG_ENV = "wynntils.url.list.debug.log";
```
magicus, line 109: At some point we might want to collect all these system properties in a single place, if you expect to add more like these. Perhaps a SystemPropertiesManager or such. If for nothing else, just to have them all collected in one place and not scattered all over. Otherwiese any name standard is sure to get lost over time.

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -203,33 +254,141 @@ private void downloadAndReadRemoteUrls(URI uri) {
...
+        dl.handleInputStream(
+                inputStream -> {
+                    try {
+                        UrlList urlList = readUrls(inputStream);
+                        urlListMap.put(UrlListType.REMOTE, urlList);
+
+                        // Merge the lists with the remote list as the primary source
+                        mergeUrlLists();
+                    } catch (IllegalStateException e) {
+                        // IllegalStateExceptions are thrown on cricical errors detected by the manager
```
magicus:
```suggestion
                        // IllegalStateExceptions are thrown on critical errors detected by the manager
```

common/src/main/java/com/wynntils/core/net/UrlManager.java
kristofbolyai: Good idea, I'll make sure to do it sometime.

kristofbolyai:
I think should be reviewable now, @magicus.

common/src/main/java/com/wynntils/commands/WynntilsCommand.java
```diff
@@ -214,9 +214,12 @@ private int reloadCaches(CommandContext<CommandSourceStack> context) {
                                 .withStyle(ChatFormatting.YELLOW),
                         false);
 
+        Managers.Url.loadUrls();
```
magicus, line 218: I think it is better to say something "positive" about this instead of something "negative" about the removed code, e.g.
`// This reloads all URLs, and will then trigger a re-download by the DownloadManager`

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -342,4 +528,18 @@ private static final class UrlProfile {
         String md5;
         String encoding;
     }
+
+    private record UrlList(int version, Map<UrlId, UrlInfo> urls) {
```
magicus: This name confused the heck out of me. It's not a list? Perhaps use `VersionedUrlMap`? Or just `UrlMap`?

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -342,4 +528,18 @@ private static final class UrlProfile {
...
+        public static final UrlList EMPTY = new UrlList(-1, Map.of());
+
+        public UrlInfo get(UrlId urlId) {
+            return urls.get(urlId);
+        }
+    }
+
+    private enum UrlListType {
```
magicus: Maybe `UrlSourceType`?

common/src/main/java/com/wynntils/core/net/UrlManager.java
magicus: Ok, I think I understand now. We would need a clear name on what the file "urls.json" represents. It is not a "url", it is a set of mapping urlid->url, a version, and a provenance. 

Maybe "URLs holder" (`UrlsHolder`)? Getting the plural -s might help saying that we are dealing with a collection of urls.

magicus: Urgh. I dunno, really. UrlsHolder does not sound good, either. UrlCollection? UrlMapper? Hm, yeah, that might work. Then `readUrls` would be `readUrlMapper`, and `urlListMap` would be `urlMappers`. I think that would be good.

common/src/main/java/com/wynntils/core/net/UrlManager.java
magicus: ... `UrlMapperType`. Sorry for my stream-of-concisousness reviewing here.

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -203,33 +263,156 @@ private void downloadAndReadRemoteUrls(URI uri) {
...
 
-    private void tryUpdateUrlMap(Pair<Integer, Map<UrlId, UrlInfo>> tryMap, boolean forceUpdateOnSameVersion) {
-        // Even if the version is the same, we might want to update the URLs if the urls are downloaded from the
-        // internet since the URLs might have changed. (local cache and bundled URLs are not always updated)
-        if (tryMap.a() > version || (forceUpdateOnSameVersion && tryMap.a() == version)) {
-            urlMap = tryMap.b();
-            version = tryMap.a();
+    private void mergeUrlLists() {
+        // In theory, this method shouldn't be called concurrently ever, but just in case
+        synchronized (urlMapLock) {
```
magicus: You don't need a separate object for that! You can just lock on the urlListMap object itself.

magicus: (That's actually a design decision from Java 1.0 that costs us all a lot these days -- all objects need to reserve bits in the object header since it could potentially be used as a lock...)

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -61,50 +98,62 @@
...
             WynntilsMod.info(
-                    "Loading urls.json in the normal mode. Url cache is " + (ignoreCache ? "ignored" : "used") + ".");
+                    "Loading urls.json in the normal mode. Url cache is " + (IGNORE_CACHE ? "ignored" : "used") + ".");
         } else {
             WynntilsMod.info("Loading urls.json from " + urlListOverrideUri + ". Url cache is "
-                    + (ignoreCache ? "ignored" : "used") + ".");
+                    + (IGNORE_CACHE ? "ignored" : "used") + ".");
         }
 
+        forceSource = UrlListType.valueOf(FORCE_SOURCE.toUpperCase(Locale.ROOT));
```
magicus: This is really putting the spotlight on the need of a SystemPropertiesManager, that can read and initialize values retrieved from system properties, and then just expose those values in their preferred form.

I'll accept it for now, but I'd really appreciate if you could try to make it a priority to refactor out these things afterwards.

magicus:
I think the overall logic is correct. My comments has mostly been about the readability of the code. 

However, it really makes me wonder, once again: is it worth all this trouble to have an updatable URL list?

common/src/main/java/com/wynntils/core/net/UrlManager.java
kristofbolyai: I'll try my best in the upcoming days.

kristofbolyai:
> However, it really makes me wonder, once again: is it worth all this trouble to have an updatable URL list?

It's more like, is it worth the trouble to have 3 different URL lists? Maybe not, although, at this point, it should be handled once and for all..

kristofbolyai:
I think I addressed everything, @magicus.

common/src/main/java/com/wynntils/core/net/UrlManager.java
```diff
@@ -203,33 +263,156 @@ private void downloadAndReadRemoteUrls(URI uri) {
...
         }
     }
 
-    private void tryUpdateUrlMap(Pair<Integer, Map<UrlId, UrlInfo>> tryMap, boolean forceUpdateOnSameVersion) {
-        // Even if the version is the same, we might want to update the URLs if the urls are downloaded from the
-        // internet since the URLs might have changed. (local cache and bundled URLs are not always updated)
-        if (tryMap.a() > version || (forceUpdateOnSameVersion && tryMap.a() == version)) {
-            urlMap = tryMap.b();
-            version = tryMap.a();
+    private void mergeUrlLists() {
```
magicus: One more refactoring: Name this `mergeUrlMappers`, and then I'm happy.

magicus (approved):

magicus:
And thank you for renaming to UrlMapper. The logic became much clearer.

kristofbolyai:
Thanks for reviewing it!

PR #2856, opened 2024-09-11
fix: Update TM bulk sell for 2.1 [skip ci]

ryanzhoudev:
The sell buttons moving a bit is just to account for the new texture.

### DIFF
--- a/common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
+++ b/common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
@@ -5,6 +5,7 @@
 package com.wynntils.features.trademarket;
 
 import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.components.Models;
 import com.wynntils.core.consumers.features.Feature;
 import com.wynntils.core.persisted.Persisted;
 import com.wynntils.core.persisted.config.Category;
@@ -14,6 +15,7 @@ import com.wynntils.core.text.PartStyle;
 import com.wynntils.core.text.StyledText;
 import com.wynntils.handlers.chat.event.ChatMessageReceivedEvent;
 import com.wynntils.mc.event.ContainerSetSlotEvent;
+import com.wynntils.models.containers.containers.TradeMarketSellContainer;
 import com.wynntils.screens.base.widgets.WynntilsButton;
 import com.wynntils.utils.mc.McUtils;
 import com.wynntils.utils.wynn.ContainerUtils;
@@ -32,7 +34,6 @@ public class TradeMarketBulkSellFeature extends Feature {
     private static final Pattern ITEM_NAME_PATTERN = Pattern.compile("(.+)À");
     private static final String SOLD_ITEM_SLOT = "Empty Item Slot";
     private static final String CLICK_TO_SET_AMOUNT = "Set Amount";
-    private static final String SELL_DIALOGUE_TITLE = "§f\uDAFF\uDFE8\uE014";
     private static final String TYPE_SELL_AMOUNT =
             "\uDAFF\uDFFC\uE001\uDB00\uDC06 Type the amount you wish to sell or ";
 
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
```diff
@@ -56,25 +56,28 @@ public void onSellDialogueUpdated(ContainerSetSlotEvent.Pre e) {
...
-        if (!title.equalsString(SELL_DIALOGUE_TITLE, PartStyle.StyleType.NONE)) return;
+        // The sell title changes based on the state of the screen but always contains this string
+        if (!title.contains(SELL_DIALOGUE_TITLE)) return;
 
         StyledText amountItemName = StyledText.fromComponent(
                 containerScreen.getMenu().getSlot(AMOUNT_ITEM_SLOT).getItem().getHoverName());
         if (!amountItemName.equalsString(CLICK_TO_SET_AMOUNT, PartStyle.StyleType.NONE)) return;
 
         String soldItemName = getSoldItemName(containerScreen);
+        System.out.println(soldItemName);
```
kristofbolyai: debug

common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
```diff
@@ -56,25 +56,28 @@ public void onSellDialogueUpdated(ContainerSetSlotEvent.Pre e) {
...
 
         removeSellButtons(containerScreen);
 
         if (soldItemName == null) return;
 
         addSellButtons(containerScreen, soldItemName);
+        System.out.println("Added sell buttons");
```
kristofbolyai: debug

common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
```diff
@@ -56,25 +56,28 @@ public void onSellDialogueUpdated(ContainerSetSlotEvent.Pre e) {
...
     }
 
     @SubscribeEvent
     public void onChatMessage(ChatMessageReceivedEvent e) {
         if (!sendAmountMessage) return;
-        if (!e.getOriginalStyledText().equalsString(TYPE_SELL_AMOUNT, PartStyle.StyleType.NONE)) return;
+        if (!e.getOriginalStyledText().contains(TYPE_SELL_AMOUNT)) return;
```
kristofbolyai, line 77: StartsWIth perhaps? although I wouldn't be sure the symbols are always the long one, maybe it's a "second" message in a line, so perhaps adapt the regex?

common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
```diff
@@ -56,25 +56,28 @@ public void onSellDialogueUpdated(ContainerSetSlotEvent.Pre e) {
         if (!(McUtils.mc().screen instanceof ContainerScreen containerScreen)) return;
 
         StyledText title = StyledText.fromComponent(containerScreen.getTitle());
-        if (!title.equalsString(SELL_DIALOGUE_TITLE, PartStyle.StyleType.NONE)) return;
+        // The sell title changes based on the state of the screen but always contains this string
+        if (!title.contains(SELL_DIALOGUE_TITLE)) return;
```
ShadowCat117: Maybe register a container in ContainerModel that will match both, that way when the price match feature is fixed it won't need to duplicate this title check

common/src/main/java/com/wynntils/features/trademarket/TradeMarketBulkSellFeature.java
ryanzhoudev: For this, they send a block of 4 lines, and the part I'm using .contains on is guaranteed to be the second line
image

So what's there now is actually the shorter symbols, and it's guaranteed to be the shorter symbols. That's also why I didn't use startswith

kristofbolyai: Okay, sure.

kristofbolyai (approved):

PR #2857, opened 2024-09-12
fix: Increase delay between territory updates when not in a guild

ShadowCat117:
Not a requested change by Wynncraft devs, just a "does it need to call it this often" which for a large amount of players, even those in a guild, it doesn't. When we eventually have model configs there could even be a setting for the delay

### DIFF
--- a/common/src/main/java/com/wynntils/models/players/GuildModel.java
+++ b/common/src/main/java/com/wynntils/models/players/GuildModel.java
@@ -12,6 +12,7 @@ import com.wynntils.core.WynntilsMod;
 import com.wynntils.core.components.Handlers;
 import com.wynntils.core.components.Managers;
 import com.wynntils.core.components.Model;
+import com.wynntils.core.components.Models;
 import com.wynntils.core.net.ApiResponse;
 import com.wynntils.core.net.DownloadRegistry;
 import com.wynntils.core.net.UrlId;
@@ -23,6 +24,7 @@ import com.wynntils.handlers.container.type.ContainerContent;
 import com.wynntils.mc.event.ContainerSetContentEvent;
 import com.wynntils.models.character.CharacterModel;
 import com.wynntils.models.containers.ContainerModel;
+import com.wynntils.models.players.event.GuildEvent;
 import com.wynntils.models.players.label.GuildSeasonLeaderboardHeaderLabelParser;
 import com.wynntils.models.players.label.GuildSeasonLeaderboardLabelParser;
 import com.wynntils.models.players.profile.GuildProfile;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/models/players/GuildModel.java
```diff
@@ -274,6 +278,7 @@ public void parseGuildInfoFromGuildMenu(ItemStack guildInfoItem) {
             Matcher guildNameMatcher = line.getMatcher(GUILD_NAME_MATCHER);
             if (guildNameMatcher.matches()) {
                 guildName = guildNameMatcher.group("name");
+                Models.Territory.scheduleTerritoryUpdate(true);
```
ShadowCat117: Did originally have this as a `GuildEvent.Set` but not sure how often that would be used so changed it to this. Can change it back if that's a better approach

kristofbolyai: What if you started scheduling with the small timer, and on every reschedule, you check the guild model, so you know whether you are in a guild or not. And then, you only need `GuildEvent.Joined` to "cancel" the current timer (and optionally `GuildEvent.Left`, but that's only a micro-optimization).

common/src/main/java/com/wynntils/models/territories/TerritoryModel.java
```diff
@@ -251,5 +277,7 @@ private void updateTerritoryProfileMap() {
                             .collect(Collectors.toSet());
                 },
                 onError -> WynntilsMod.warn("Failed to update territory data."));
+
+        scheduleTerritoryUpdate(Models.Guild.isInGuild());
```
kristofbolyai: Shouldn't you do this in a "finally" way? So basically after processing, and after error handling as the Download object doesn't have support for "finally".

common/src/main/java/com/wynntils/models/players/GuildModel.java
kristofbolyai: Do you still need the whole scheduling method still? I think you really could just do one of `schedule`-ing with the `schedule` method, and have the download itself handle schedule itself after finishing.

_That's a lot of scheduling :)_

kristofbolyai (approved):
Thanks.

common/src/main/java/com/wynntils/models/territories/TerritoryModel.java
```diff
@@ -172,6 +176,16 @@ public void onAdvancementUpdate(AdvancementUpdateEvent event) {
         }
     }
 
+    @SubscribeEvent
+    public void onGuildJoined(GuildEvent.Joined e) {
+        scheduleTerritoryUpdate();
```
magicus, line 181: I see this is already merged, and maybe this comment does not matter, but if you join a guild, it will take 15 s to update the territory. I was thinking that this should trigger an instant update, and then the next after 15 s.

ShadowCat117: It actually does work like that, since we put 0 as the initialDelay in scheduleWithFixedDelay so it's immediately called

magicus: Ah, I see. That's great. Thanks!

PR #2858, opened 2024-09-14
fix: Don't crash when stat filters exceed the integer limit [skip ci]

ShadowCat117:


kristofbolyai:
Can you check other filters? I suspect there may be some other integer parsing ones.

ShadowCat117:
All the integer ones are done in RangedStatFilters after the refactor for the filter UI. PercentageStatFilter uses floats so doesn't have this issue

PR #2860, opened 2024-09-15
fix: custom loading screen fixes [skip ci]

ILikePlayingGames:
- Enable custom loading screen for world switching
- Fix incorrect width calculations when a non-default font is used in BufferedFontRenderer
- Fix Wynncraft logo on custom loading screen
- Rename "title" and "subtitle" fields in LoadingScreen to un-hide superclass field

### DIFF
--- a/common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
+++ b/common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
@@ -5,7 +5,6 @@
 package com.wynntils.features.ui;
 
 import com.wynntils.core.consumers.features.Feature;
-import com.wynntils.core.events.MixinHelper;
 import com.wynntils.core.persisted.config.Category;
 import com.wynntils.core.persisted.config.ConfigCategory;
 import com.wynntils.mc.event.LoadingProgressEvent;
@@ -40,12 +39,6 @@ public class CustomLoadingScreenFeature extends Feature {
             // Ensures baseConnectScreen is cleared after the initial handshake occurs
             // Only our LoadingScreen and ConnectScreen should be able to work with baseConnectScreen
             baseConnectScreen = null;
-        }
-
-        // Enable the custom loading screen for world transfers
-        if (MixinHelper.onWynncraft() && event.getScreen() instanceof ReceivingLevelScreen && loadingScreen == null) {
-            loadingScreen = LoadingScreen.create();
-            McUtils.mc().setScreen(loadingScreen);
         }
 
         if (loadingScreen == null) return;
@@ -125,7 +118,10 @@ public class CustomLoadingScreenFeature extends Feature {
                 McUtils.mc().setScreen(loadingScreen);
             }
             case INTERIM -> {
-                if (loadingScreen == null) return;
+                if (loadingScreen == null) {
+                    loadingScreen = LoadingScreen.create();
+                    McUtils.mc().setScreen(loadingScreen);
+                }
 
                 loadingScreen.setMessage(I18n.get("feature.wynntils.customLoadingScreen.joiningWorld"));
             }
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
```diff
@@ -41,6 +42,12 @@ public void onScreenOpenPre(ScreenOpenedEvent.Pre event) {
             baseConnectScreen = null;
         }
 
+        // Enable the custom loading screen for world transfers
+        if (MixinHelper.onWynncraft() && event.getScreen() instanceof ReceivingLevelScreen && loadingScreen == null) {
```
ShadowCat117: As the name implies, MixinHelper should only be used in Mixins.

ShadowCat117: There also already is a ReceivingLevelScreen case in this method, so perhaps instead you should create the screen in the `INTERIM` case for `onWorldStateChange` if there is no current loading screen

kristofbolyai (commented):
Hold your horses! This PR is currently a flashing in red and buzzing in my head. It's a mix of at least 3 different changes with 0 explanation at all. I can break it up, as 

- Fixed `LOGO_STRING` and logo rendering: Okay, I can understand this. This change should be the _only_ change in the PR.

- Renamed (sub)title -> (sub)heading: Why? This is standard Minecraft terminology, I don't see why it warrants a rename, and mixing this rename with other changes makes the PR confusing. Please either provide a reason, or revert this.

- `BufferedFontRenderer` changes: I see what you are getting at, likely `(Buffered)FontRenderer` runs checks on the formatted text itself, but this doesn't work when you introduce fonts and other special component coloring in the mix. However, this is not the way to fix this. This change alone warrants a separate PR, as it's not only likely that this same mistake appears in more places in the renderer, it's also very possible that some other code "depends" on this mistake, so it incorrectly assumes something, and fixing the rendering makes it break, which we have to fix. What we (and most other OSS projects) would expect you to do, is either create an issue, or PR the fix (if you are confident in the fix, and don't think it's worth discussing with the other developers beforehand) for the issue separately, before or in parallel to this PR. (Note that if you choose to PR the fix in parallel, make sure both PRs' description clearly states their relation).

common/src/main/java/com/wynntils/features/ui/CustomLoadingScreenFeature.java
ILikePlayingGames: Thanks! I implemented your suggestions.

kristofbolyai:
Okay, now this PR is getting cleaner. I hope you didn't take my previous comment wrong, I am not here to harass you or your changes in any way, however it's part of my job to try to keep the repository maintainable.

I do now see that you had mentioned the FontRenderer fixes in the PR description, I missed that. Sorry about it. However, I'd still like to see that separated and fixed, I assume this PR still needs that fix to be working correctly. Would you like to work on it, or should we start an issue and track it separately?

ILikePlayingGames:
Thanks for your feedback. I get what you mean, and I’ll try to keep the changes more separate in the future. Regarding the title rename, the issue as mentioned in the PR is the title field overrides the superclass field Screen#title . Ideally they should be differently named, but if you think that’s not necessary we can leave it. The FontRenderer issue should probably be tracked as an issue. I probably won’t have enough time to go through everything that depends on it.

This PR does not need the FontRenderer fix, it has a workaround for the issue.

kristofbolyai:
I don't think you have to go through all the uses of FontRenderer though. Let me check. 

As for the title/heading change, I understand the concern, but I don't think it's worth dealing with/changing anything about for now.

kristofbolyai (approved):

PR #2861, opened 2024-09-15
feat: Add SystemPropertiesManager to handle custom manager JVM properties

kristofbolyai:
This should give us a nice, generalized way of using JVM arguments as properties. Currently, I have these properties limited to managers, however I think that makes a lot of sense from a design standpoint, as the user shouldn't be able to modify other CoreComponents' settings in this way.

### DIFF
--- a/common/src/main/java/com/wynntils/core/properties/SystemPropertiesManager.java
+++ b/common/src/main/java/com/wynntils/core/properties/SystemPropertiesManager.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.core.properties;
+
+import com.google.common.collect.ImmutableMap;
+import com.wynntils.core.WynntilsMod;
+import com.wynntils.core.components.Manager;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+public class SystemPropertiesManager extends Manager {
+    private static final Map<Class<?>, Function<String, Object>> MAPPERS = ImmutableMap.of(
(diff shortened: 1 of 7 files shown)


kristofbolyai:
@magicus The parsing and testing is not fully done, but I think the API should look something like this. What do you think?

common/src/main/java/com/wynntils/core/properties/SystemPropertiesManager.java
```diff
@@ -0,0 +1,65 @@
...
+                T enumValue = (T) EnumUtils.searchEnum((Class<? extends Enum>) property.getClassType(), propertyValue);
+                return enumValue;
+            } catch (IllegalArgumentException e) {
+                WynntilsMod.warn("Could not parse enum from JVM property " + property.getFullJvmArgumentPath()
+                        + ". Using null. Value: " + propertyValue);
+                return null;
+            }
+        }
+
+        // Then fall back to using the mappers†
```
magicus:
```suggestion
        // Then fall back to using the mappers
```

magicus (approved):
This looks fantastic! Much better than I was thinking about; my vague idea was just to collect the properties in a single class, but this is waaaay better.

kristofbolyai:
> This looks fantastic! Much better than I was thinking about; my vague idea was just to collect the properties in a single class, but this is waaaay better.

I did actually think about doing that first, however then it made me wonder how to access values, etc, it came with many problems.

I'm not confident in limiting this to managers only, in fact, I'll likely extend properties to all CoreComponents when the need arises. I do plan to add a JVM switch to disable Hades once the download changes are merged (which shouldn't be far away, as I am only waiting on Jamie RN).

magicus:
I guess as long as our names does not clash, even with the <CoreComponentType> part removed, that should be just fine.

PR #2862, opened 2024-09-16
feat: held item cooldown overlay + focused mob health overlay

phantamanta44:
## Features

This PR adds two new overlays, both of which are disabled by default:

* Held item cooldown overlay: displays an indicator for the cooldown of the held item (e.g. from using your main attack).
  * Also includes a feature that override vanilla crosshair rendering with our own; otherwise, the crosshair is off-center. This is toggleable as a separate feature.
  * Also includes a normal function `held_item_cooldown` for accessing the held item cooldown.

* Focused mob health bar overlay: displays a custom replacement for the focused mob health bar, which is usually displayed as a boss bar
  * Useful for when players want to move that information elsewhere on the screen.
  * Also includes normal functions for accessing the focused mob's name, health, and elemental properties (as an unparsed string for now).

[Here's a brief video demonstration of these overlays.](https://www.youtube.com/watch?v=rIhDmUHa0VE)

## Implementation Notes

* The properties of the focused mob are captured by `DamageModel`, which is responsible for parsing the focused mob health bar as a boss bar.
  * Some high-health mobs display their health as an abbreviated number, so there's a new utility function for parsing such numbers.
    * Since we support suffixes up to the trillions, mob health is now represented as a `long`. **This is technically a breaking change!** However, nothing in Wynntils currently uses these private values, so it's unlikely that anything was actually broken.
* The focused mob's elemental properties are currently retrieved as the raw string extracted from the boss bar. In this form, it's usable for displaying as a text overlay, but it'll need to be parsed into something more structured if players want to use it for template logic. This is left as future work.


Rafii2198:
>  In this form, it's usable for displaying as a text overlay, but it'll need to be parsed into something more structured if players want to use it for template logic

If you mean that the new `focused_mob_health()` function would return something like `12345678900` and want to have it abbrivated then the `format()` function can do that. In this case `format(focused_mob_health)` would return `12.3M` given the number from the example.
If I understand your function correctly ofc.

Rafii2198:
Also both types of functions in FunctionMenager are not sorted alphabetically. Same for features.

phantamanta44:
> > In this form, it's usable for displaying as a text overlay, but it'll need to be parsed into something more structured if players want to use it for template logic
> 
> If you mean that the new `focused_mob_health()` function would return something like `12345678900` and want to have it abbrivated then the `format()` function can do that. In this case `format(focused_mob_health)` would return `12.3M` given the number from the example. If I understand your function correctly ofc.

that last bullet point is referring to the mob's elemental properties, i.e. its weaknesses, resistances, and damage types



> Also both types of functions in FunctionMenager are not sorted alphabetically. Same for features.

aren't the features in alphabetic order? and i noticed that not all the functions in `FunctionManager` are alphabetically ordered, so i assumed that it was supposed to be grouped by feature

Rafii2198:
My bad on the sorting of features, I probably missread it, as for functions I believe they are grouped by type and sorted alphabetically kind of like features.

I also read again that last bullet point and my conclusion is that I should leave such reviews to devs especially at 3am... My bad, sorry.

common/src/main/java/com/wynntils/features/overlays/FixCrosshairPositionFeature.java
```diff
@@ -0,0 +1,54 @@
...
+import com.wynntils.core.persisted.config.Category;
+import com.wynntils.core.persisted.config.ConfigCategory;
+import com.wynntils.mc.event.RenderEvent;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.Gui;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.OVERLAYS)
```
kristofbolyai: Isn't this more of an UTILITY than an OVERLAY?

kristofbolyai: However, this begs the question, whether we should fix this, or just make users use separate mods, if they care to

common/src/main/java/com/wynntils/functions/InventoryFunctions.java
```diff
@@ -310,4 +313,25 @@ protected List<String> getAliases() {
...
+            Player player = McUtils.player();
+            ItemStack itemStack = player.getItemInHand(InteractionHand.MAIN_HAND);
+            ItemCooldowns cooldowns = player.getCooldowns();
+            ItemCooldowns.CooldownInstance cooldown = cooldowns.cooldowns.get(itemStack.getItem());
+            if (cooldown == null || cooldown.startTime >= cooldown.endTime) return CappedValue.EMPTY; // Sanity check
+
+            int remaining = cooldown.endTime - cooldowns.tickCount;
+            if (remaining <= 0) return CappedValue.EMPTY;
+
+            return new CappedValue(remaining, cooldown.endTime - cooldown.startTime);
```
kristofbolyai: Extract this to the appropriate model, functions should essentially be "public getters" of core components data.

common/src/main/java/com/wynntils/models/damage/DamageModel.java
```diff
@@ -14,26 +14,30 @@
...
 import java.util.regex.Pattern;
+import net.minecraft.client.gui.components.LerpingBossEvent;
 import net.neoforged.bus.api.SubscribeEvent;
 
 public final class DamageModel extends Model {
     // Test in DamageModel_DAMAGE_BAR_PATTERN
-    private static final Pattern DAMAGE_BAR_PATTERN =
-            Pattern.compile("^§[ac](.*) - §c(\\d+)§4❤(?:§r -( (§.(.+))(Dam|Weak|Def))+)?$");
+    private static final Pattern DAMAGE_BAR_PATTERN = Pattern.compile(
+            "^\\s*§[ace](.*) - §c(\\d+(?:\\.\\d+)?[kKmM]?)§4❤(?:§r - ( ?(§.(.+))(Dam|Weak|Def))+)?\\s*$");
```
kristofbolyai: Does Wynn do the "kKmM" now? We used to have to add that client side.

common/src/main/java/com/wynntils/utils/StringUtils.java
```diff
@@ -19,6 +19,7 @@
 
 public final class StringUtils {
     private static final String[] SUFFIXES = {"", "k", "m", "b", "t"}; // kilo, million, billion, trillion (short scale)
+    private static final long[] SUFFIX_MULTIPLIERS = {-1L, 1_000L, 1_000_000L, 1_000_000_000L, 1_000_000_000_000L};
```
kristofbolyai: Why not 1L? I see that it is not used, but multiplying with 1L wouldn't mess up any math, even if you miss-handled something.

common/src/main/java/com/wynntils/models/damage/DamageModel.java
```diff
@@ -76,22 +96,31 @@ public double getAverageAreaDamagePerSecond(int seconds) {
...
             super(DAMAGE_BAR_PATTERN);
         }
 
+        @Override
+        public void setEvent(LerpingBossEvent event) {
+            super.setEvent(event);
+            // Wynncraft updates the focused mob health bar by repeatedly destroying and recreating the boss bar. We
+            // grab the progress here, when the bar's recreation event is received, to avoid the brief delay before
+            // the first progress update event is received, during which the state would otherwise be inconsistent
+            onUpdateProgress(event.getProgress());
```
kristofbolyai: This sounds like something that should be handled on a `BarHandler` (if that's how it's called) level. I'll leave this note here, and take a look, or alternatively, you can too, if you want to get this PR in faster.

kristofbolyai (commented):
Just as an overall comment, didn't the actual "attack cooldown" feature used to work with Vanilla? Or was the timing wrong? Maybe it could be patched and fixed, and that would mean much less work for us.

phantamanta44:
> Just as an overall comment, didn't the actual "attack cooldown" feature used to work with Vanilla? Or was the timing wrong? Maybe it could be patched and fixed, and that would mean much less work for us.

there is the vanilla swing speed indicator for weapon swings, but that's a separate subsystem from item cooldowns, and as far as i know, it's never been used by Wynncraft. in principle, it might be possible to reuse it, but it would require either an ugly mixin hack or for us to just re-render it entirely, which kinda defeats the point of reusing it in the first place

(also i just like the way mine looks more lol)

common/src/main/java/com/wynntils/features/overlays/FixCrosshairPositionFeature.java
phantamanta44: wasn't sure about the category; it's technically a "vanilla overlay" of sorts. at any rate, i included this change because the off-center vanilla reticle made the centered item cooldown indicator look terrible

common/src/main/java/com/wynntils/models/damage/DamageModel.java
phantamanta44: i found it on a few mobs in SE with the abbreviation feature disabled

common/src/main/java/com/wynntils/models/damage/DamageModel.java
phantamanta44: i initially considered just dealing with it in `BossBarHandler`, but i was hesitant to make what is technically a breaking behavioural change, especially considering the comment here
https://github.com/Wynntils/Wynntils/blob/1a3e676a8ad51c3e4bd1ec4e160d966d2ba24d4c/common/src/main/java/com/wynntils/models/abilities/bossbars/BloodPoolBar.java#L30-L32
which seems to imply that not updating progress immediately is the expected behaviour

common/src/main/java/com/wynntils/features/overlays/FixCrosshairPositionFeature.java
kristofbolyai: I think that's why it should move to the utilities category, because the OVERLAY category is strictly reserved for our custom overlays.

kristofbolyai:
Sounds reasonable, I originally didn't check how it looks, but now I agree it looks much better than the vanilla one.

common/src/main/java/com/wynntils/overlays/gamebars/FocusedMobHealthBarOverlay.java
```diff
@@ -0,0 +1,78 @@
...
+    public void onFocusChange(FocusedDamageEvent.MobFocused event) {
+        this.currentProgress = progress().progress();
+    }
+
+    @Override
+    protected String text() {
+        String healthString = Managers.Feature.getFeatureInstance(AbbreviateMobHealthFeature.class)
+                        .isEnabled()
+                ? StringUtils.integerToShortString(Models.Damage.getFocusedMobHealth())
+                : Long.toString(Models.Damage.getFocusedMobHealth());
```
kristofbolyai, line 59: Should you use `Long.toString`? IIRC it adds thousand separators and in some locales it looks really weird.

phantamanta44: it shouldn't; it's the normal toString map for `long`s

common/src/main/java/com/wynntils/features/utilities/FixCrosshairPositionFeature.java
```diff
@@ -0,0 +1,54 @@
...
+import com.wynntils.mc.event.RenderEvent;
+import com.wynntils.utils.mc.McUtils;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.Gui;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.UTILITIES)
+public class FixCrosshairPositionFeature extends Feature {
+    private static boolean shouldOverrideCrosshair(Minecraft mc) {
```
kristofbolyai: It's a bit weird to pass in MC, no? You really could just access it statically.

common/src/main/java/com/wynntils/models/damage/DamageModel.java
kristofbolyai: Okay, sounds sane enough.

phantamanta44:
ran a raid yesterday and realized that raid bosses have dark red and bold names, so i've just allowed any colour in the health bar regexp

common/src/main/java/com/wynntils/models/damage/DamageModel.java
magicus: The comment in BloodPoolBar is just odd? I think bolyai has a point in that this would be better handled in the BossBarHandler. Can you try doing it there anyway and see if it breaks BloodPool bars? Otherwise I think this is basically a "bug fix" of BossBarHandler.

common/src/main/java/com/wynntils/overlays/HeldItemCooldownOverlay.java
```diff
@@ -0,0 +1,76 @@
...
+        renderOverlay(poseStack, bufferSource, cdPercent);
+    }
+
+    @Override
+    public void renderPreview(
+            PoseStack poseStack, MultiBufferSource bufferSource, DeltaTracker deltaTracker, Window window) {
+        renderOverlay(poseStack, bufferSource, 1F);
+    }
+
+    private void renderOverlay(PoseStack poseStack, MultiBufferSource bufferSource, float cdPercent) {
```
magicus: Can I ask you to spell out `coolDownPercent`?

common/src/main/java/com/wynntils/overlays/HeldItemCooldownOverlay.java
```diff
@@ -0,0 +1,76 @@
...
+
+        // Draw bars
+        float barWidth = 3f;
+        float offset = (1f - cdPercent) * (width - barWidth) / 2f;
+        BufferedRenderUtils.drawRect(poseStack, bufferSource, color, x1 + offset, y, 0, barWidth, height);
+        BufferedRenderUtils.drawRect(poseStack, bufferSource, color, x2 - barWidth - offset, y, 0, barWidth, height);
+    }
+
+    @Override
+    protected void onConfigUpdate(Config<?> config) {}
```
magicus, line 75: Do we still need to implement `onConfigUpdate` for overlays? I thought we fixed that.

magicus:
Overall I think this looks great! Both in terms of the functionality provided, and the quality of the source code. 

Thanks for contributing to Wynntils.

common/src/main/java/com/wynntils/overlays/HeldItemCooldownOverlay.java
phantamanta44: it does still appear to be defined `abstract` in `Overlay`, yes

common/src/main/java/com/wynntils/models/damage/DamageModel.java
phantamanta44: i'll give it a shot, but admittedly i'm not much of a shaman player lol

common/src/main/java/com/wynntils/overlays/HeldItemCooldownOverlay.java
magicus: *sigh* Ok.

kristofbolyai: I can fix that.

common/src/main/java/com/wynntils/models/damage/DamageModel.java
magicus: You can always ask for help testing on Discord if you need.

common/src/main/java/com/wynntils/models/damage/DamageModel.java
```diff
@@ -60,6 +71,37 @@ public void onLabelIdentified(LabelIdentifiedEvent event) {
...
+    }
+
+    public String getFocusedMobName() {
+        checkFocusedMobValidity();
+        return focusedMobName;
+    }
+
+    public String getFocusedMobElementals() {
+        checkFocusedMobValidity();
+        return focusedMobElementals; // TODO: Parse this into specific elements and expose as functions
```
kristofbolyai: We usually don't do same-line TODOS, but really just a nit.

```suggestion
        // TODO: Parse this into specific elements and expose as functions
        return focusedMobElementals;
```

kristofbolyai (approved):

kristofbolyai:
This also looks really good, thanks! I'll go ahead and test the blood pool bar, then merge this.

kristofbolyai:
Blood pool does work! I'll merge this now (without doing a release just yet.)

PR #2863, opened 2024-09-16
fix: Time Rift, Pandemonium, and overall totem performance fixes

ryanzhoudev:
Trift and Pandemonium now bypass the cast check, since they can cast willy nilly. Apparently pandemonium was also patched so that other classes can't cast totems, so now all events check for shaman class.

~~For the performance fix, essentially we slow down the rate of trying to match a timer to a totem after the first few attempts. This should help with whatever performance issues there were before, but still gives the totem/timer an opportunity to rebind if the totem and timer are able to move (with the ability) and realign themselves.~~

Also fixes an inaccurate warning log where the first timer rename would be considered a failed match. Usually the first timer instance is spawned very far away (around 5 blocks diagonal), then teleported to the totem once the totem makes that "boom" sound.


kristofbolyai:
> For the performance fix, essentially we slow down the rate of trying to match a timer to a totem after the first few attempts. This should help with whatever performance issues there were before, but still gives the totem/timer an opportunity to rebind if the totem and timer are able to move (with the ability) and realign themselves.

The performance issue was basically from logging a warning that we couldn't match the timer, as the logging was heavy IO-wise.

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -102,9 +115,28 @@ public void onTotemSpawn(AddEntityEvent e) {
...
+            Optional<GearItem> gearItemOpt = Models.Item.asWynnItem(item, GearItem.class);
+            if (gearItemOpt.isEmpty()) continue;
+
+            GearItem gearItem = gearItemOpt.get();
+            if (AUTO_CASTER_ITEMS.contains(gearItem.getItemInfo().name())) {
+                return true;
+            }
+        }
+
+        return false;
```
kristofbolyai: This could be a one liner with streams, no?

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -47,6 +53,9 @@ public class ShamanTotemModel extends Model {
 
     private final ShamanTotem[] totems = new ShamanTotem[MAX_TOTEM_COUNT]; // 0-indexed list of totems 1-4
     private final Integer[] pendingTotemVisibleIds = new Integer[MAX_TOTEM_COUNT]; // ID of totems that need a timer
+    private final Map<Integer, Integer> orphanedTimers =
+            new ConcurrentHashMap<>(); // IDs of timers that can't really find a totem
```
kristofbolyai: Why a concurrent map? From what I see, this field is only accessed from a single event/method (with it being `TextDisplayChangedEvent.Text`). Maybe this was left over from some alternative change?

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -116,6 +148,11 @@ public void onTimerRename(TextDisplayChangedEvent.Text event) {
         int parsedTime = Integer.parseInt(m.group("time"));
         int timerId = textDisplay.getId();
 
+        if (orphanedTimers.getOrDefault(timerId, 0) >= 3) {
+            // check every other second to find the totem to avoid performance problems
+            if (parsedTime % 2 == 0) return;
+        }
+
```
kristofbolyai: Given my comment above, do we need this? Since only error/warn logging took time, not the actual matching code, this is incorrect optimization, as rather, the logging should be reduced.

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -38,6 +41,7 @@
 public class ShamanTotemModel extends Model {
     // Test in ShamanTotemModel_SHAMAN_TOTEM_TIMER_PATTERN
     private static final Pattern SHAMAN_TOTEM_TIMER = Pattern.compile("§c(?<time>\\d+)s(\n\\+(?<regen>\\d+)❤§7/s)?");
+    private static final List<String> AUTO_CASTER_ITEMS = List.of("Time Rift", "Pandemonium");
```
ShadowCat117: I think checking for major id vs item is probably a better solution as this can get outdated very easily, the same can be said for major id but I'd say it's less likely to be changed often. Especially now that we can track major id's from raids and I know Sorcery is one of the possible id's

common/src/main/java/com/wynntils/models/abilities/ShamanTotemModel.java
```diff
@@ -71,7 +83,8 @@ public void onTotemSpawn(AddEntityEvent e) {
                 () -> {
                     // didn't come from a cast within the delay, probably not casted by the player
                     // this check needs to be ran with a delay, the cast/spawn order is not guaranteed
-                    if (System.currentTimeMillis() - totemCastTimestamp > CAST_MAX_DELAY_MS) return;
+                    if (!hasAutoCasterItem() && System.currentTimeMillis() - totemCastTimestamp > CAST_MAX_DELAY_MS)
```
magicus: If you want to make this a "performance fix", maybe you should change order here and allow the time stamp check to short-circuit the test before jumping into the more complex hasAutoCasterItem.

kristofbolyai (approved):
I think this looks good, but someone else should also take a look.

ShadowCat117 (approved):
Looks good!

PR #2866, opened 2024-09-16
fix: Fix (Buffered)FontRenderer not calculating width/height correctly in every case [skip ci]

kristofbolyai:
I think this should be everything needed for #2860, but I couldn't test this just yet. Can you perhaps try this, @ILikePlayingGames?

### DIFF
--- a/common/src/main/java/com/wynntils/utils/render/FontRenderer.java
+++ b/common/src/main/java/com/wynntils/utils/render/FontRenderer.java
@@ -215,7 +215,7 @@ public final class FontRenderer {
                 renderX,
                 cursorRenderY,
                 0,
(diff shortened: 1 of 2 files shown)


kristofbolyai:
From my testing, this seems to work. Anyone found anything, or can we merge this?

ShadowCat117:
I'll give it a quick test

ShadowCat117 (approved):
Seems fine

kristofbolyai:
Thanks!

ILikePlayingGames:
> I think this should be everything needed for #2860, but I couldn't test this just yet. Can you perhaps try this, @ILikePlayingGames?

Thanks I’ll try to test it tonight

PR #2867, opened 2024-09-16
fix: Fix auto join party [skip ci]

ShadowCat117:


kristofbolyai (commented):
Why, Wynn? Maybe do an or, whenever Wynn fixes this :)

ShadowCat117:
I'm not sure if they will add the banner to this message but sure I can add it

kristofbolyai (approved):
Ah, I didn't know this message is centered. Whatever, let's not revert the regex now..

PR #2871, opened 2024-09-18
feat: inventory interaction improvements and desync fixes

phantamanta44:
This PR makes various improvements related to inventory interactions (i.e. interactions with "real" items rather than with placeholders or UI elements). In particular, it does the following:

* Forces synchronization after interactions that are known to cause desyncs
  * In particular, after:
    1. interactions that don't modify nonempty slots and which leave the held stack empty
    2. item "spread" interactions (i.e. dragging an item stack to split it across multiple slots)
  * This is achieved by sending a "hotbar swap" interaction to some random slot; since Wynncraft doesn't support swapping operations on "real" items, this allows us to get back a sync packet without actually affecting the inventory
* Adds an `InventoryInteractionEvent` that describes actual inventory interactions with "real" items
  * This is done by computing the expected interaction on the client after a slot click, then verifying that the expected interactions actually went through by inspecting sync packets received in a small time window after the interaction
  * Currently not used for anything, but it'll be the basis for a future PR
* Adds a "fix stack sizes" feature that correctly displays received stack sizes for items that are unstackable in vanilla
  * This was an issue for e.g. broken forgery keys and crafter bags from raids
  * Does this by hooking into `SimpleContainer#setItem`, which truncates stacks after they're received from the server
* Adds a "synchronize inventories" feature that synchronizes all slot clicks with the server before applying them on the client
  * This means the client inventory state will *always* correctly reflect the server inventory state, but at the cost of noticeably reduced responsiveness in all inventories (since all clicks now take an entire RTT!)
  * This is probably not a worthwhile tradeoff for most players, so it's disabled by default


common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -717,6 +719,8 @@
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightColor.name": "0 Star Ingredient Highlight Color",
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightEnabled.description": "Enable highlighting of 0 star ingredient items",
   "feature.wynntils.itemHighlight.zeroStarIngredientHighlightEnabled.name": "0 Star Ingredient Highlight",
+  "feature.wynntils.itemInteractionSound.description": "Plays sounds when items are moved around in inventories.",
```
magicus: Is this a planned further addition to this PR, or something that never happened/got removed?

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                // SWAP and CLONE are not supported as inventory operations on Wynncraft
+        }
+    }
+
+    @SubscribeEvent
+    public void onSetContents(ContainerSetContentEvent.Pre event) {
+        long now = System.currentTimeMillis();
+        int containerId = event.getContainerId();
+        List<ItemStack> newInv = event.getItems();
+        pending.removeIf(entry -> now >= entry.expiryTime || !entry.interaction.onContentUpdate(containerId, newInv));
```
magicus: We have a TimedSet (or something like that) that removes items after certain amount of time. 

But I am not sure what you are using this pending list for. I can't see any reads from it?

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    @SubscribeEvent
+    public void onWorldStateChange(WorldStateEvent event) {
+        running = null;
+        pending.clear();
+    }
+
+    public void forceSync(AbstractContainerMenu menu) {
+        McUtils.sendPacket(new ServerboundContainerClickPacket(
```
magicus: Please add some comments why this forces a sync.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+        }
+    }
+
+    private static final class PendingInteractionEntry {
+        private final PendingInteraction interaction;
+        private final long expiryTime;
+
+        private PendingInteractionEntry(PendingInteraction interaction) {
+            this.interaction = interaction;
+            this.expiryTime = System.currentTimeMillis() + 500L;
```
magicus: Please move this "magic" numbers to a constants i the top-level handler class instead.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                    if (heldStack.isEmpty()) {
+                        if (!slotStack.isEmpty()) {
+                            expect(new PendingInteraction.PickUp(menu, slotNum, slotStack.copy()));
+                        }
+                    } else if (slotStack.isEmpty()) {
+                        if (event.getMouseButton() == 0 || heldStack.getCount() == 1) {
+                            forceSyncLater(menu);
+                        }
+                        expect(new PendingInteraction.Place(menu, slotNum, heldStack.copy()));
+                    } else if (areItemsSimilar(heldStack, slotStack)) {
```
magicus: I am not sure I understand this `areItemsSimilar` logic. Are you assuming that it is the same item stack if the name and count matches? Or do we need to take some special action if we are dealing with two different stacks of the same kind of item?

magicus:
My brain fell asleep while looking at the `PendingInteraction` hierarchy. I'll have to come back and review it later.

common/src/main/resources/assets/wynntils/lang/en_us.json
phantamanta44: oh, whoops. this was the original goal for the PR, but I decided that there was already so much stuff going on that it made more sense to save it for a later one

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: I decided that `TimedSet` had too much overhead given how small the pending interaction set should be (probably two elements at most!). the reads are the `removeIf` calls, which invoke the hooks in `PendingInteraction`. i could probably make this more clear with comments

kristofbolyai:
HI! Thanks for the contribution! Before I start reviewing the code itself, I have a few suggestions and/or requests for now and in the future.

As this PR extends the code on the CoreComponenent (Handler) level, we are stricter and more picky in reviewing it. In fact, I usually prefer when people talk to us before submitting such changes, just so the reviewing process is easier for both of us, and we can filter changes that are out of scope. You can easily do this through Discord, or alternatively GH issues, although we do prefer Discord for more direct conversations.

This PR does not fall into the category of "out of scope" additions, however I would like some semi-major changes first. I'd like to request, that any "optimizations"/'fixes" the `InventoryHandler` does is optional, so it is toggled by a feature, or a feature's configs. Otherwise I'd fear that the upkeeping/maintainability of the code would be too hard, and would create game-breaking bugs in the mod itself, if Wynncraft itself changes something on their end. 

Lastly, I'd prefer if you mentioned some examples where these inventory "bug/desync-fixes" are useful, just for further reference, or if we need to come back to this PR in the future.

Thank you in advance!

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: yes, i'm making that assumption on the basis that pretty much every Wynncraft item has a distinct name. the item type is an overapproximation because of certain interactions where picking up a stack changes its item type without changing its identity, e.g. picking up weapons or helmets with gear skins equipped

magicus: So what happens if you pick up one stack with e.g. materials and put it down on top of another stack with the same material?

kristofbolyai:
> Adds an `InventoryInteractionEvent` that describes actual inventory interactions with "real" items
> * This is done by computing the expected interaction on the client after a slot click, then verifying that the expected interactions actually went through by inspecting sync packets received in a small time window after the interaction
> * Currently not used for anything, but it'll be the basis for a future PR

If the event itself is not too deep in the Handler/Code structure, can I ask you to not add it until it's actually being used? Once again, this is mostly just a concern of maintainability and keeping the repository clean of dead (and in this case, "not yet born") code.

> Adds a "synchronize inventories" feature that synchronizes all slot clicks with the server before applying them on the client
> * This means the client inventory state will _always_ correctly reflect the server inventory state, but at the cost of noticeably reduced responsiveness in all inventories (since all clicks now take an entire RTT!)
> * **This is probably not a worthwhile tradeoff for most players, so it's disabled by default**

Do you think some players would actually benefit from this? How much code does this feature need? If it's not an "useful" feature, but it adds too much code to maintain, it might be wiser to not include it afterall.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: 1. `onSlotClick` determines that it should expect a `Place` interaction and registers an entry in `pending`
2. the game sends the click packet and Wynncraft responds one or more sync packets that update the slot with the materials and the held stack
3. either `onSetContents` or `onSetSlot` relays the sync data to the pending interaction
4. the pending interaction verifies that the held materials actually went into the slot and posts an `InventoryInteractionEvent` with the interaction data

phantamanta44:
> I'd like to request, that any "optimizations"/'fixes" the InventoryHandler does is optional, so it is toggled by a feature, or a feature's configs.

the issue with this would be that the "fixes" are merely a side effect of the event infrastructure. in particular, we actually don't receive enough information from Wynncraft to reliably recover the full inventory state, so we force synchronizations when necessary to get that information. this information comes in the form of inventory sync packets, which are also what fix the desync bugs.

in principle, it could be possible to make this toggleable anyways with one of the following resolutions:

* just throw in the config option and let players trade off interaction event precision at their own discretion; any event consumers will have to be aware of the possible loss of precision
* implement a feature dependency system so that any features which rely on interaction events can require the option to be enabled
* instead of toggling the fixes, let the handler always run correctly and make the config option toggle an "unfixer" that reverts the inventory to a previous untouched state

i don't particularly like any of these options, and given that more synchronization should in theory never actually break anything, i would say the best option is to just leave the "fixes" enabled

phantamanta44:
> If the event itself is not too deep in the Handler/Code structure, can I ask you to not add it until it's actually being used? Once again, this is mostly just a concern of maintainability and keeping the repository clean of dead (and in this case, "not yet born") code.

really, the event is the principal contribution here; the rest of the fixes are either side effects or afterthoughts. at any rate, my PR with the feature relying on the event should come in the very near future

> Do you think some players would actually benefit from this? How much code does this feature need? If it's not an "useful" feature, but it adds too much code to maintain, it might be wiser to not include it afterall.

with the desync fixes in place, the feature needs negligible code; all it actually does is prevent clientside processing of slot clicks, and the rest follows from correctness of synchronization. i'm not sure if anyone will actually use it, but i can at least imagine the use case. it would especially be useful to players with very low latency, which is not me

magicus:
> the issue with this would be that the "fixes" are merely a side effect of the event infrastructure. 

I am not sure this is true. Afaict, the only visible "fix" here is the call to `forceSync`; the rest seem to just parse the current state and convert it into events for further consumption.

What would happen if you made the `forceSync` call optional? Would the entire infrastructure come apart, or would it just risk being a bit more unreliable? From my understanding it would be the latter, not the former, but I must admit that I don't fully grasp what you are trying to accomplish in `InventoryHandler`.

With that said, I think I am more positive towards this PR than what bolyai seem to be. The code is clear, so my lack of understanding seems more to be founded in the fact that you are trying to solve a complex problem. If anything, `InventoryHandler` could gain from a bit longer high-level presentation at the top of the file how the handler works.

Could you share a draft PR with your planned upcoming work that will utilize the inventory events? If we agree that this is a good goal, then I see no problem with pushing preparatory code that introduces an event that is not used. The problem here is that we do not know what you plan to use it for, nor that we will agree that it is a good feature that should be accepted once you present it.

Basically, I think you have published a set of very high-quality patches, that I am more than happy to accept -- but they all appeared quite out of the blue for us. So we'd like to have more communication with you about what you are planning.

magicus:
Also, please merge rather than rebase on an open PR. Otherwise it can make existing comments dangling, and overall makes it harder to understand how a PR was reviewed. We will squash the PR upon integration anyway.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                default -> {
+                    running = null;
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+
+    private interface PendingInteraction {
```
magicus: This file is rather long, and the majority of it is nested classes. I don't advocate moving every single nested class into a top-level class, but perhaps try to move the PendingInteraction hierarchy to a separate file, and see if it makes the files more balanced?

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+            @Override
+            public boolean onSlotUpdate(int containerId, int slotNum, ItemStack newStack) {
+                if (containerId == -1) {
+                    tryDispatch(newStack);
+                    return false;
+                }
+                return true;
+            }
+
+            protected abstract void tryDispatch(ItemStack newHeldStack);
```
magicus: The name `tryDispatch` throw me off a bit when trying to understand this code. I recommend `sendEvent` instead. The "try" part is not needed; you can always assume that any method will only perform its core functionality if the correct conditions hold. And we use the terminology "send", not "dispatch", of events.

magicus:
To clarify what I meant above:

If you were to change your calls to `forceSyncLater` with sending a `PossibleInventoryDesyncEvent` (or something like that), and this was listened to by a feature, that did the scheduling of the `forceSync` operation, I think that would be a much better solution.

kristofbolyai:
Replying to https://github.com/Wynntils/Wynntils/pull/2871#issuecomment-2358116943,

I am actually quite happy about the PR, despite how my comments sound. What I could describe in my head, but couldn't successfully communicate here is that we need to see how this PR is going to be used, and we need some details on how it could be disabled. Most parts of the mod is optional/toggle-able, and I would like to keep that here, if possible (we did do and still have exceptions to this, like ChatHandler). If that's not possible, there is nothing left but to review the code throughly of any edge-cases to make sure it's ready to be "mission-critical" code.

I also wonder, if these desyncs get fixed on Wynn's side, do we just handle that and not do any excess work? Another question regarding this, do we send/request a considerable amount of requests? From what I've seen in the code, we don't, but admittedly, I haven't went through it yet.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
magicus: So how does this differ from if you were to pick up one stack with materials, and put it down on another stack with another material? Basically, what I'm asking is what would break if you were to remove this `if (areItemsSimilar(heldStack, slotStack))` check.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
magicus: What kind of overhead is it you are afraid of? Performance-wise? I'd like to see us utilize our existing functionality as much as possible, instead of creating new similar functionality. And if the existing functionality is missing something, it is better to improve it.

This is of course always a judgement call, and I understand why this particular instance is really a borderline case. Nevertheless, using `TimedSet` would make it show directly even at the declaration of `pending` that it is a, well, timed set. And it would make `PendingInteractionEntry` unnecessary.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,449 @@
...
+        if (running != null && running.handle(event)) return;
+
+        AbstractContainerMenu menu = event.getContainerMenu();
+        ItemStack heldStack = menu.getCarried();
+        int slotNum = event.getSlotNum();
+        switch (event.getClickType()) {
+            case PICKUP -> {
+                if (slotNum < 0) { // Slot -999 -> tossed an item outside the GUI
+                    if (!heldStack.isEmpty()) {
+                        if (event.getMouseButton() == 0 || heldStack.getCount() == 1) {
```
kristofbolyai: What do you think of using `GLFW.GLFW_MOUSE_BUTTON_LEFT` constants instead of 0? It's a bit longer code but lessens the mental load on the reader.

kristofbolyai (commented):
I don't have any other comments other than what magicus pointed out. Maybe once the InventoryHandler subclasses get extracted, I can take a look at them once again.

Looks really great overall!

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: the idea is that the collection is small, but updated relatively frequently, so ideally i want:

1. linear scan to find the entry matching a given slot number
2. cheap insertion and removal

thus, i think the best options are either a linked list or a nullable array. then, my issue with `TimedSet` are:

* i can't back it with a straightforward linked list or nullable array because it requires set semantics (unless i implement my own non-compliant `Set`, which would not be great)
* i don't have control over when pruning operations happen because i can't call `TimedSet#releaseEntries` even if auto-clear is disabled

the second issue is easy enough to solve, but the first would require a breaking change (i.e. weakening the collection semantics)

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: oh, i guess it wouldn't actually make that much of a difference because `PendingInteraction.Swap` handles placements as an edge case. good catch

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: i think i would still like the name to indicate that the function is responsible for validating the interaction; maybe `checkAndSendEvent`?

phantamanta44:
> What would happen if you made the forceSync call optional? Would the entire infrastructure come apart, or would it just risk being a bit more unreliable? From my understanding it would be the latter, not the former, but I must admit that I don't fully grasp what you are trying to accomplish in InventoryHandler.

you are right, but i see it as an essential bit of reliability. without the `forceSync` calls, we would not be able to track any operations that don't modify an occupied slot and which leave the cursor empty (e.g. placing an item into an empty slot), as well as most spread operations. probably the biggest issue with this is that it'll difficult to articulate to users exactly what they're trading off by choosing to enable or disable these "fixes"

i suppose that if the "fixes" are disabled, we *could* try to overapproximate these operations by just immediately sending the events without verification; i'm unsure of whether this is better or worse than not sending the events at all

> Could you share a draft PR with your planned upcoming work that will utilize the inventory events?

will work on it

> I also wonder, if these desyncs get fixed on Wynn's side, do we just handle that and not do any excess work?

if Wynncraft fixes the sync issues on their side, then we could simply remove the `forceSync` calls and the event infrastructure will work transparently

> Another question regarding this, do we send/request a considerable amount of requests?

we send one additional packet per operation introducing a potential desync, which would be the non-slot-modifying cursor-emptying operations i mentioned above. the bigger overhead would be from the additional packets received: Wynncraft responds to almost every inventory operation with several sync packets, but it currently does not respond to these desyncable operations at all (which is why they're desyncable in the first place!); then, this change will cause these operations to elicit sync packets in parity with all the other inventory operations

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: hmm, not quite, actually. `Swap` checks that the placed stack has exactly the same size as the previously held stack (e.g. placing an item into an accessory slot) whereas `Place` only checks that the items are similar. however, it *does* seem like a good idea to merge them considering that there are some items that Wynncraft considers distinct but which the client can't differentiate (e.g. keys from world events and keys from key guardians)

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
magicus: Sure, that's okay.

magicus:
> you are right, but i see it as an essential bit of reliability. without the forceSync calls, we would not be able to track any operations that don't modify an occupied slot and which leave the cursor empty (e.g. placing an item into an empty slot), as well as most spread operations. probably the biggest issue with this is that it'll difficult to articulate to users exactly what they're trading off by choosing to enable or disable these "fixes"

> i suppose that if the "fixes" are disabled, we could try to overapproximate these operations by just immediately sending the events without verification; i'm unsure of whether this is better or worse than not sending the events at all

First of all, I think what makes me and bolyai worried is that the suggested patch unconditionally modifies the Minecraft client behavior. As long as we only react locally to stuff sent by the server, and as long as any responses we send back are governed by features that can be turned off by the user, in case they turn out to be bad for some reason, then it's all fine. But hard-coding sending data to the server inside a Handler like this is not really something we would like.

This is mostly a question of robustness of the mod. If, for some reason, these "re-sync" packets cause some problem for some people, there is no way to turn them off.

Second, I am still not sure I understand how making these calls optional would change the robustness. You schedule the resync packet to be sent in the next tick, and then you send an event immediately anyway, without awaiting for this next tick to happen. It might be that I don't fully understand neither the problem nor how inventory updates are communicated between the server and the client, but to me it would seem like we would need to have a flow like:

1) user does something that can cause a desync
2) you schedule a re-sync operation
3) you send an event
4) at the next tick, the re-sync packet is sent to the server
5) the server replies to this packet in some way, that ends up inside the intentory handler
6) presumably, you catch this new situation and sends a new event confirming that the server has seen the action.

What am I missing here?

phantamanta44:
the way it works is generally:

1. user clicks a slot and the click listener in `InventoryHandler` catches it
2. the click listener figures out what kind of interaction it's supposed to be and inserts an appropriate entry into `pending`, optionally scheduling a force-sync packet if it decides it's necessary
3. the game sends the click packet to the server, as well as the force-sync packet if it was scheduled in (2)
4. the server responds with one or more sync packets, which are caught by the sync listeners in `InventoryHandler`
5. the sync listeners pass the data to the pending interaction
6. the pending interaction examines the data to validate the interaction; if successful, it posts the event

no event is posted until the validation at the very end, and so if we take the force-sync out of the equation, we'd have to either:

* post events without validation in (2) when we don't know whether we'll get a sync packet or not
* keep the process the same and just don't post events when we don't get sync packets

common/src/main/java/com/wynntils/handlers/inventory/PendingInteraction.java
```diff
@@ -0,0 +1,261 @@
...
+import net.minecraft.world.item.ItemStack;
+
+interface PendingInteraction {
+    boolean onSlotUpdate(int containerId, int slotNum, ItemStack newStack);
+
+    default boolean onContentUpdate(int containerId, List<ItemStack> newInv) {
+        return true;
+    }
+
+    static boolean areItemsSimilar(ItemStack a, ItemStack b) {
```
magicus: I was about to suggest that you put this in `ItemStackUtils`, but then I double checked and to my surprise we did not have such a class. We really ought to have one, in `com.wynntils.utils.utils.mc`. I suggest you create one and move this function there.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,186 @@
...
+
+        private boolean handle(ContainerClickEvent event) {
+            if (event.getClickType() != ClickType.QUICK_CRAFT
+                    || event.getContainerMenu().containerId != menu.containerId) {
+                running = null;
+                return false;
+            }
+
+            int mask = event.getMouseButton();
+            if (((mask & 0x4) == 0) == single) {
```
magicus: Please use constants instead of magic numbers in code. (Also applies to line 161 below.)

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
magicus: LinkedList is perhaps not as performant as you may think, but whatever -- I doubt it matter any much way.

I've been looking at your code a couple of times now and I finally get what you are doing. But the removeIf call has a very high "logic density". I think it would be better if you wrapped it in a method call, so you could do something like:
`checkAndRemovePending(entry -> entry.interaction.onSlotUpdate(containerId, slotNum, newStack));`

and have the expiry time check and the `!` inversion inside `checkAndRemovePending`.

magicus:
I think I'm getting a better understanding how this is supposed to work now, thanks for your explanations. 

If we were to remove the force sync packets, then not all events will be affected, right? It's only those that would require such a force sync packets, such as throwing items. 

That would mean that much of the code would make sense even if we were to disable force sync packets. Whatever solution we ends up with, I can't really see a way that we can accept this without any possibility to disable sending such packets. So the question is: if we disable such packets, then what do we do? As you say, we can either post events optimistically without confirmation, or we can not post such events at all. Of course, we could support both solutions -- the flag in question could be a tri-state: SEND_SYNC_PACKETS, ASSUME_SUCCESS, IGNORE_REPORTING.

Once again the question of how you are planning to use these events become relevant. What happens if the events are sent optimistically but not verified, or not sent at all? Exactly how broken would your new feature be in that case?

magicus:
But I agree that the sending of sync packets cannot and should not really be moved outside the handler to a feature -- it is not so much a "repair broken vanilla behavior" (as I first perceived this fix), but a "this is needed to keep our model intact".

Which itself also leads to a second problem: we don't really have a story for setting configuration variables in Handlers. In the worst case, we'll have to create a dummy Feature which is only used to store a config variable, and then look up that feature directly from inside the Handler. That is not at all nice; but we have made similar solutions before.

common/src/main/java/com/wynntils/handlers/inventory/PendingInteraction.java
phantamanta44: i think that's probably what `ItemUtils` is for, right? it has functions like `boolean isWeapon(ItemStack)`

phantamanta44:
i thought of another option: the interaction events could have a "confidence" attached to them which describes whether the interaction is verified or not, and the consumers can then decide for themselves whether they want to accept unverified interactions or not. obviously consumers that need verified events will still lose data, but it's better than forcing users to pick one or the other

also i suspect most players will just leave improved sync enabled anyways, so this is likely a very niche concern

phantamanta44:
as a bit of a funny anecdote, i cleared reincarnation on one of my low-level classes yesterday to get a bob weapon, but once i got to the temple of legends, i discovered that the stone of myths was no longer in my inventory! i can only assume that when i was moving it out of my hotbar in the tomb, i suffered a desync (since placing it in an empty slot doesn't sync!) and it ended up dropping on the ground without my noticing. at any rate, i now play with inventory synchronization enabled lol

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -0,0 +1,265 @@
...
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.handlers.inventory;
+
+import static net.minecraft.world.inventory.AbstractContainerMenu.QUICKCRAFT_HEADER_CONTINUE;
+import static net.minecraft.world.inventory.AbstractContainerMenu.QUICKCRAFT_HEADER_END;
+import static net.minecraft.world.inventory.AbstractContainerMenu.QUICKCRAFT_HEADER_START;
+import static net.minecraft.world.inventory.AbstractContainerMenu.QUICKCRAFT_TYPE_GREEDY;
+import static net.minecraft.world.inventory.AbstractContainerMenu.getQuickcraftHeader;
+import static net.minecraft.world.inventory.AbstractContainerMenu.getQuickcraftType;
```
kristofbolyai: Please don't use static imports. Afaik spotless should complain, but maybe we couldn't add this as a rule efficiently.

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
```diff
@@ -0,0 +1,66 @@
...
+import it.unimi.dsi.fastutil.ints.Int2ObjectOpenHashMap;
+import java.util.Arrays;
+import net.minecraft.network.protocol.game.ServerboundContainerClickPacket;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.INVENTORY)
+public class ImprovedInventorySyncFeature extends Feature {
```
kristofbolyai, line 24: Didn't this used to be disabled as it needed 1 tick to update, or something around the lines of that? I can't find the original message I've read this..

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
```diff
@@ -60,6 +61,14 @@ private TooltipFlag onGetTooltipLines(TooltipFlag flags) {
         return event.getFlags();
     }
 
+    @ModifyReturnValue(method = "copy", at = @At("RETURN"))
```
kristofbolyai: Can you use that semi-qualified mixin format we use? I think I've seen you use it in your other mixins, but it's basically the fully qualified method target without the L<classpath>; prefix.

kristofbolyai (commented):
Essentially this looks good to me, but due to the size of the PR, it's likely that both me and @magicus need a round or two to review this. Sorry about it taking so long!

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: it didn't complain, no

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
phantamanta44: the feature is just a marker for enabling/disabling all the sync changes; the force-sync thing is controlled by the `forceSync` sub-option now

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
phantamanta44: i note that there are a handful of other non-qualified mixin targets: image

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
kristofbolyai: Sounds good! So that means we can disable the changes done by the handler, right?

kristofbolyai (approved):
Essentially this looks good to me! Let's wait for magicus' review, and in the meantime, I can also test this in-game.

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
phantamanta44: the extra packets, yes. the event infrastructure will always be there, but when the feature is disabled, the events that require extra sync are just sent without verification

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
```diff
@@ -0,0 +1,66 @@
...
+    @Persisted
+    public Config<Boolean> forceSync = new Config<>(false);
```
magicus, line 26: You can have this as `true` as default. I don't doubt that it works and is helpful, if you claim it is. All I wanted was a way to disable it. If you leave this as false as default, not many players will turn it on.

kristofbolyai: But it delays inventory actions? Parts of the fix are enabled by default, and this more "hardcore" one is not. Am I correct in this?

phantamanta44: that's correct

common/src/main/java/com/wynntils/mc/mixin/ItemStackMixin.java
magicus: Those are because we have been sloppy in reviewing, combined with that we have no automatic checks. The format bolyai described is the one we should be using.

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
magicus: We're talking about two things. The forceSync config is for controlling the re-sync packets from the InventoryHandler. That is what I propose to be on by default.

The Feature itself is, I believe, the extra safety-belts sychronization thing that probably is not needed iirc. That should be off by default. I can't see right now if it is or not (github cuts just before that) but if it isn't, it should be.

magicus:
This is starting to look ready for merging! Thanks for your efforts, and for working with us in getting the patch in shape for inclusion. 

There's just a few remarks left, then I think we're done.

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
phantamanta44: no, i've flipped them because the force-sync feature depends on the sync packets so it makes more sense this way around

common/src/main/java/com/wynntils/features/embellishments/ItemInteractionSoundsFeature.java
```diff
@@ -0,0 +1,228 @@
...
+            case InventoryInteraction.PickUp ixn -> playSoundEvent(PICK_UP_SOUND);
+            case InventoryInteraction.Place ixn -> playItemSound(ixn.placed());
+            case InventoryInteraction.Spread ixn -> playItemSound(ixn.stack());
+            case InventoryInteraction.Swap ixn -> playItemSound(ixn.placed());
+            case InventoryInteraction.Transfer ixn -> playItemSound(ixn.transferred());
+            default -> {}
+        }
+    }
+
+    public void playItemSound(ItemStack stack) {
```
magicus: I find the overloaded name here a confusing and unsuitable. Combined with the switch above, where it is not easy to know what item type is returned from the events, that makes it impossible to see just by reading the code if e.g. InventoryInteraction.Spread calls this method or the other one. 

In general, I think name overloading is a mis-feature, and the only good usage is if the essentially same method but with a different set of arguments (or a different, but similar type, for one or more of the arguments.)

common/src/main/java/com/wynntils/features/inventory/ImprovedInventorySyncFeature.java
magicus: Ah, I see now that `isImprovedSyncEnabled` checks if this feature is enabled; I thought it checked if forceSync was turned on. Then I agree, it's all fine!

magicus (approved):
I think this looks fine now. Thanks once again for your contribution!

PR #2872, opened 2024-09-18
feat: attack/spell keybind inversion [skip ci]

phantamanta44:
Some guildmates asked for a feature that allows the main attack and spell keybinds to be inverted for archers to maintain consistency with the other classes; this PR is a generalization of that, allowing selective inversion of any class' attack and spell keybinds.


magicus:
Jeez, you're on a roll! :-D

common/src/main/java/com/wynntils/features/combat/InvertKeybindsFeature.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import net.minecraft.client.KeyMapping;
+import net.minecraft.client.Options;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@StartDisabled
+@ConfigCategory(Category.COMBAT)
+public class InvertKeybindsFeature extends Feature {
```
magicus: Maybe `InvertAttackKeybindFeature` instead? Otherwise the name suggests a very sweeping functionality.

magicus (approved):
Overall, this looks good.

common/src/main/java/com/wynntils/features/combat/InvertAttackKeybindsFeature.java
```diff
@@ -0,0 +1,108 @@
...
+            default -> false;
+        }) {
+            return null;
+        }
+
+        // Ensure the held item is a weapon
+        Optional<WynnItem> wynnItemOpt = Models.Item.getWynnItem(InventoryUtils.getItemInHand());
+        if (wynnItemOpt.isEmpty()
+                || !(wynnItemOpt.get() instanceof GearTypeItemProperty gearTypeItem)
+                || !gearTypeItem.getGearType().isWeapon()) return null;
```
ShadowCat117: `ItemUtils` has an `isWeapon` that will do this for you

ShadowCat117:
My only concern with a feature like this is the way the spell cast indicators won't accurately reflect the correct mouse input. An inverted left click input will still display the right click icon which could get confusing for those who like to keep track of any misinputs

common/src/main/java/com/wynntils/features/combat/InvertAttackKeybindsFeature.java
```diff
@@ -0,0 +1,108 @@
...
+        if (!switch (Models.Character.getClassType()) {
+            case MAGE -> invertMage.get();
+            case ARCHER -> invertArcher.get();
+            case WARRIOR -> invertWarrior.get();
+            case ASSASSIN -> invertAssassin.get();
+            case SHAMAN -> invertShaman.get();
+            default -> false;
+        }) {
+            return null;
+        }
```
kristofbolyai: nit; but since ShadowCat has a valid comment just below, can you extract this to a separate method?

kristofbolyai (approved):
Essentially looks great, please address the short comments. Thank you!

Essentuan:
I am not entirely convinced this should be added. While this isn't enabled by default players would likely have to still think about the fact they're playing archer since Wynncraft DOES distinguish between left and right click (sometimes), meaning some actions you'll have to think about which button to use. In my opinion the fact this inconsistency exists (and isn't possible to solve because of the input queue) should leave this kind of functionality to a JSMacros script over a Wynntils feature.

ShadowCat117:
I do agree that this doesn't really need to be a feature, the fact that archer is the only one with inverted enabled by default shows that this is just more of a "invert archer inputs", as it's the only 1 out of the 5 classes to not use left click as main attack. I do think a mod like JSMacros is probably better suited for something niche like this but if bolyai and magicus think it's fine then I'm not going to stop it

kristofbolyai:
So, what do you think @phantamanta44? Do you have any ideas on making this feature more connected with Wynncraft (and in turn, Wynntils)?

To be clear, I am not opposed to this PR, however I would like to hear a valid idea, just like Essentuan brought one up.

phantamanta44:
i really have no better argument for this feature than "some wynntils users asked me to implement it". i'll ask them to weigh in

magicus:
Hm, does this mean you have to use the "wrong" click to e.g. open a chest, if you have invert keymapping turned on? Maybe that will limit the usefulness.

Otoh, I can't see how adding this would hurt, either. In general, as long as functionality is contained within features and with additions according to our architecture, I have little to say if something should be added or not. Especially if it is off by default. "Someone wanted this" is basically a good enough reason for me.

kristofbolyai:
Just to write down what I think, I support both sides. We try to add functionality in the mod that is relevant to Wynncraft, and mostly Wynncraft, however some more generic, Minecraft features do essentially make their way into the mod. Whether that's good or not is up for debate. Why we didn't reject this PR instantly in the first place is that this functionality could be implemented neatly in the mod, in a disable-able, and non-intrusive way. The same thing couldn't be said about, for example the chat-related PRs, which is why they caused so much headache. Given however, that even that PR got accepted, I think this PR is opt to be accepted as well.

kristofbolyai (approved):
No comments left. Looks good, thank you!

ShadowCat117:
> Hm, does this mean you have to use the "wrong" click to e.g. open a chest, if you have invert keymapping turned on? Maybe that will limit the usefulness.

Chests do allow both left and right click inputs but other cases such as speaking to npcs only support right click so those will be inverted if you are holding a weapon whilst trying to speak to them. And I believe there are other areas in the game where only 1 click type works.

My original concern was only for the spell indicators being swapped but I hadn't considered other areas of the game that will only support the 1 click type, for the people who only want archer to match the other classes for attack inputs I don't think any of them have realised how this will affect other features and how those inputs will now also be reversed

magicus:
> for the people who only want archer to match the other classes for attack inputs I don't think any of them have realised how this will affect other features and how those inputs will now also be reversed

I think you are right. I hate the archer button flip too, I always gets confused when I switch between classes. But I think this is something that really needs to be solved at the Wynn level. But now we have done what we can; it might not get used much but it's not really problematic to have it as well.

PR #2874, opened 2024-09-18
refactor: Don't require onConfigUpdate to be overridden in overlays [skip ci]

kristofbolyai:
This was already the case with features. It's fascinating this took so long.


magicus:
Thanks!

PR #2876, opened 2024-09-19
feat: Reworked ItemCompareFeature

uTa4u:
Changes inside of ItemCompareFeature class:
- Automatically compares with equipped items (including accessories and held weapon)
- Only compare with items of the same type
- Item selection indicator is correctly rendered, unlike in old version
- More than one item can be selected. Selected items can be of different types - feature will only pick items of the same type
- Tooltips don't clip outside of screen and among themselves even on gui scale 3
- Removes lore and set info from items during comparison to safe vertical space. Both can be toggled off. Set stats are not affected
- Displays tags above items for ease of distinction between equipped/selected/hovered items. Can be toggled off. Russian lang file included
- Config option to enable centered item names. Since I'm sure players will install LegendaryTooltips and then report lack of centered item names during comparison as a bug

Other changes:
- New method in ItemUtils to check if item is `empty accessory slot`
- Refactored InventoryModel to use aforementioned method
- New helper methods in TooltipUtils
- A few patterns made public in WynnItemParser
-  ItemStatInfoFeature and TooltipFittingFeature now use methods from TooltipUtils


magicus:
Do you have a screenshot or screencast that showcases the functionality in this PR?

uTa4u:
> Do you have a screenshot or screencast that showcases the functionality in this PR?

Sure, here

https://github.com/user-attachments/assets/9063500d-5e1e-41a5-894f-b95025480621

magicus (approved):
I think this looks overall good, but I'd like for @kristofbolyai to take a look as well.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,590 @@
...
+
+        poseStack.pushPose();
         poseStack.translate(0, 0, 300);
 
-        int toBeRenderedWidth = abstractContainerScreen.getTooltipFromItem(McUtils.mc(), itemToCompare).stream()
-                .map(component -> McUtils.mc().font.width(component))
-                .max(Integer::compareTo)
-                .orElse(0);
+        List<Component> hoveredLines = new ArrayList<>(event.getTooltips());
+        if (removeFlavourText.get()) removeFlavourText(hoveredLines);
```
kristofbolyai: We add brackets for every if that is multi-line, or does not have a flow-control statement (return/continue/break) as it's expression. Please go through this method and fix them.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,590 @@
...
+
+    private final Map<WynnItem, ItemStack> selectedItems = new HashMap<>();
+    private static final int COMPARE_ITEM_PAD = 6;
+    private static final String EQUIPPED_KEY = "feature.wynntils.itemCompare.tag.equipped";
+    private static final String HOVERED_KEY = "feature.wynntils.itemCompare.tag.hovered";
+    private static final String SELECTED_KEY = "feature.wynntils.itemCompare.tag.selected";
+    private static final String HOVERED_SELECTED_KEY = "feature.wynntils.itemCompare.tag.hovered_selected";
+
+    // First equippedCount items in itemsToCompare will have "Equipped" tag, others will have "Selected" tag
+    private int equippedCount = 0;
```
kristofbolyai, line 99: Essentially this could be changed so these items are kept track of in a separate list from Selected items. I think that would make it less error-prone, as it'd be much harder to "miss-count" items accidentally. The current implementation is fair, and is likely to not miss-count, but in the future, I think this code would be just a tiny bit more maintainable if we didn't have to keep track of the count in this way.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,590 @@
...
     }
 
     private void onSelectKeyPress(Slot hoveredSlot) {
         if (hoveredSlot == null) return;
+        // Selecting items behaves weirdly on tm updates, so no selecting in tm
+        if (Models.Container.getCurrentContainer() instanceof TradeMarketContainer) return;
+        int slot = hoveredSlot.getContainerSlot();
+
+        // Selecting armor and accessory slots is not allowed
+        if (slot == 9 || slot == 10 || slot == 11 || slot == 12 || slot == 36 || slot == 37 || slot == 38 || slot == 39)
```
kristofbolyai: And long ifs should also have brackets.

kristofbolyai: Also, I wonder, shouldn't InventoryModel keep track of these slots? I don't think it's great that they are hard-coded here, in such an "easy-to-miss" way.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,590 @@
...
+        // This is not the exact center, but good enough
+        int count = Math.round((tooltipWidth - font.width(string)) / 2.0f / font.width(" "));
+        return Component.literal(" ".repeat(count) + string);
+    }
+
+    private void removeFlavourText(List<Component> lines) {
+        while (lines.getLast().getString().startsWith("§8")) {
+            lines.removeLast();
+        }
+    }
```
kristofbolyai: I'd like it so you refactor this method to use StyledText and check the first components actual ChatFormatting. I fear this could easily break if we changed how the lore component looks like upstream, `Component#getString` is not known to be the most stable result producing method..

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,590 @@
...
+
+    private void removeSetInfoText(List<Component> lines) {
+        for (int i = lines.size() - 1; i >= 0; --i) {
+            StyledText line = StyledText.fromComponent(lines.get(i)).getNormalized();
+            if (line.getMatcher(WynnItemParser.SET_ITEM_PATTERN).matches()) {
+                lines.remove(i);
+            } else if (line.getMatcher(WynnItemParser.SET_PATTERN).matches()) {
+                lines.remove(i);
+                lines.remove(i - 1);
+                return;
```
kristofbolyai, line 480: Why remove just two lines? There are sets with more items in it. Or am I missing something here?

kristofbolyai (commented):
This code is actually really great as a first contribution! Thank you for working on it. I have some questions and overall code-style related reviews, however the overall picture is looking really good.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
uTa4u: > Why remove just two lines? There are sets with more items in it. Or am I missing something here?

First if branch removes item lines eg. "- Morph-Emerald" or "- Morph-Ruby". Second branch is when we hit set name eg. "Morph Set (1/8)", then we remove the set name and an empty line above it.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
uTa4u: I'm having a bit of a problem doing this. In particular, I don't understand why lore lines have style= {color=dark_purple,!italic} if in `GearTooltipComponent#buildFooterTooltip` all lore `Components` have `ChatFormatting.DARK_GRAY` set as color

kristofbolyai: Weird, I can take a look in a bit if you haven't figured it by then.

uTa4u: I figured out that `GearTooltipComponent#buildFooterTooltip` is only used to build tooltips for fake items like ones that are displayed in Item Guide. So I guess lore having `dark_purple` as color in Style is some error(?) on Wynncraft's side

uTa4u:
Also found out that TM is not the only place where selected items behave weirdly. Bank container is also affected (maybe more). Better solution would be to clear selectedItems list when TM updates and when Bank page changes. In general when container's contents change. Maybe will look into it more later

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
uTa4u: If you don't mind, I will leave it as it is for now. I'm planning to implement comparing tomes and charms in the future and do some refactoring at the same time.

kristofbolyai:
> Also found out that TM is not the only place where selected items behave weirdly. Bank container is also affected (maybe more). Better solution would be to clear selectedItems list when TM updates and when Bank page changes. In general when container's contents change. Maybe will look into it more later

What if you just cleared the selected items on `ScreenCloseEvent`? Iirc TM and bank works by "opening" a new container on page switches, and I think using this other event may solve it. If it doesn't, what if we just don't clear the selected item on screen close? I've kind of deliberately left that out from the original implementation, so you can compare items cross containers. I know it's not the best solution though..

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,634 @@
...
+
+    private void removeFlavourText(List<Component> lines) {
+        TextColor loreColor = TextColor.fromLegacyFormat(ChatFormatting.DARK_GRAY);
+        for (int i = lines.size() - 1; i >= 0; --i) {
+            Component line = lines.get(i);
+
+            // Since line has dark_purple as color we get Component with actual color value with this
+            Component comp = StyledText.fromComponent(line).getComponent();
+            // This Component is empty, the Component we need is stored as it's sibling
+            List<Component> siblings = comp.getSiblings();
```
kristofbolyai: Sorry, but I'd still like you to change this up a little bit. Basically, use StyledText to get the first "StyledTextPart", and then turn it into a component, and check that color. StyledText does not gurantee that the resulting components always looks like (empty -> allchildren), it just guarantees that the output's style matches visually/in meaning what was passed in.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
kristofbolyai: Okay, I can let this slip, it's not a huge issue IMO.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,634 @@
...
-            itemToCompare = comparedItem;
+                    Optional<ItemStack> matchingArmorOpt = armors.stream()
+                            .filter(itemStack -> isMatchingType(itemStack, hoveredGearItemProperty))
+                            .findFirst();
+                    if (matchingArmorOpt.isPresent()) {
+                        ItemStack armor = matchingArmorOpt.get();
+                        Models.Item.getWynnItem(armor).ifPresent(wynnItem -> {
+                            if (armor != hoveredItemStack) {
+                                itemsToCompare.add(new Pair<>(wynnItem, armor));
+                                ++equippedCount;
```
magicus: A style issue: Please use postfix increment, unless required by the context to read the value (which, btw, is discouraged). This applies to all these  ++equippedCount lines.

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,648 @@
...
         Slot slot = event.getSlot();
-        if (slot.getItem() == comparedItem) {
-            RenderUtils.drawArc(event.getPoseStack(), CommonColors.LIGHT_BLUE, slot.x, slot.y, 200, 1, 6, 8);
-        }
+        drawSelectionArc(event.getPoseStack(), slot.getItem(), slot.x, slot.y, false);
     }
 
     @SubscribeEvent
-    public void onItemTooltipRenderEvent(ItemTooltipRenderEvent.Post event) {
+    public void onHotbarSlotRenderEvent(HotbarSlotRenderEvent.Pre event) {
```
ShadowCat117, line 131: Is there a reason the selection arc needs to be drawn on the hotbar? Afaik this is for the hotbar in your hud not in the inventory and you clear selected items on closing containers

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
```diff
@@ -4,162 +4,648 @@
...
+
+    private void onSelectKeyPress(Slot hoveredSlot) {
+        if (hoveredSlot == null) return;
+
+        int slot = hoveredSlot.getContainerSlot();
+
+        // Selecting armor and accessory slots is not allowed
+        for (int i : InventoryAccessory.getSlots()) {
+            if (slot == i) return;
+        }
```
ShadowCat117, line 388: What's the reason for this? If I wanted to compare my necklace with one in a container this forces me to select the container one which for some could be a minor inconvenience

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
uTa4u: For consistency only tbh. Since hotbar is still visible while inventory is open

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
uTa4u: If by 'my necklace' you mean the one you have equipped, it is being compared against already. To compare with a necklace in a container just hold compare button over it

uTa4u: Since equipped items are added to `itemsToCompare` first and only then selected items are added to it (if not present already) equipped items being selected will never change anything, so I just disabled selecting them. And the reason why I check if selected item is present in `itemsToCompare` is because held item (weapon) can be selected and not allowing to select held item is kinda weird (you can select it and then take it in your hand anyways)

common/src/main/java/com/wynntils/features/tooltips/ItemCompareFeature.java
ShadowCat117: I suppose it can be kept, I just don't see any reason to display that it's been selected twice when the focus will be on the inventory

kristofbolyai (approved):
Sorry for the delay!

PR #2877, opened 2024-09-19
fix: Do not send nor display crafted item names from item encodings

magicus:
This can be misused by users sending invalid characters or using unsuitable language that will not be blocked by Wynncraft chat filters.

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/encoding/impl/item/CraftedConsumableItemTransformer.java
+++ b/common/src/main/java/com/wynntils/models/items/encoding/impl/item/CraftedConsumableItemTransformer.java
@@ -60,13 +60,9 @@ public class CraftedConsumableItemTransformer extends ItemTransformer<CraftedCon
         level = requirementsData.requirements().level();
 
         // Optional blocks
-        NameData nameData = itemDataMap.get(NameData.class);
-        if (nameData != null) {
-            name = nameData.name();
-        } else {
-            name = "Crafted "
-                    + StringUtils.capitalizeFirst(consumableType.name().toLowerCase(Locale.ROOT));
-        }
+        // Unfortunately, we cannot use the NameData from crafted items, since it can be
+        // set to unsuitable values by the users.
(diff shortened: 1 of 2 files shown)


ryanzhoudev (approved):

Essentuan:
Would it not be better to check if the name is invalid, and if so then sending the generic name? If we don't know what characters aren't allowed we could just allow only ascii.

magicus:
We will not try to keep a list of bad words in Wynntils. EOD.

Essentuan:
> We will not try to keep a list of bad words in Wynntils. EOD.

Why not just make it a toggle then? I really don't see the value in removing this outright even if sometimes it can break. And is naming items blocked words even legal? What words are even being blocked here?

ryanzhoudev:
> > We will not try to keep a list of bad words in Wynntils. EOD.
> 
> Why not just make it a toggle then? I really don't see the value in removing this outright even if sometimes it can break. And is naming items blocked words even legal? What words are even being blocked here?

The problem is not naming items, Wynn has it's own filter yes. But we can't trust any encoding at all right now, so in theory, someone could encode a slur and we would not have a way to filter it out.

Essentuan:
> The problem is not naming items, Wynn has it's own filter yes. But we can't trust any encoding at all right now, so in theory, someone could encode a slur and we would not have a way to filter it out.

Why does this matter??? Thats still bannable, you're making a feature significantly worse in my opinion just because there are a few bad actors.

magicus:
To clarify: People have built their own encoders using our encoding protocol (see e.g. https://github.com/Zatzou/idmangler-lib). While we did sort of did encourage this by documenting the protocol properly, it also means that people are free to send any kind of strings. This is definitely an annoyance, and could potentially be a security problem. Hence, we cannot accept any free-form strings from this protocol.

EnderNon:
man when i thought you were going to fix it I didn't think it would involve completely hiding the name

magicus:
We could possibly add some additional functionality to share the original name of the crafted item alongside the encoding in the copied text, to restore the relevant part of the lost functionality.

magicus:
See https://github.com/Wynntils/Wynntils/issues/2878

PR #2879, opened 2024-09-19
feat: Add the name of crafted items to the string shared when sharing encoded items [skip ci]

magicus:
Solves https://github.com/Wynntils/Wynntils/issues/2878.

PR #2880, opened 2024-09-20
feat: Add back sanitized custom name sharing to chat encoding, as an opt-in config [skip ci]

kristofbolyai:
There are two main concerns for allowing custom name sharing:
- Anyone could encode strings that inject something into chat
- Wynncraft or us can't filter any profanity in custom names

With the names being sanitized on the client, even if bad actors can encode anything, the decoded output will be always valid from now on.

Custom names are now opt-in. This basically means the user accepts the "terms" of this config, and they accept that anything they see won't be filtered by Wynn or by us. As I brought this up on Discord, I essentially see this equivalent to how you can just send an URL of a pastebin to someone, and if they decide to open it, and they might see something rude there, or they may just see what they expected to see. I essentially agree with @Essentuan in #2877 that we shouldn't dumb down the feature just because of those few bad actors.

If anyone has something to add to this PR, please do. I'd be happy to leave this PR open for discussion for 1-2 days, so everyone can bring up their concerns.

### DIFF
--- a/common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
+++ b/common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
@@ -6,9 +6,23 @@ package com.wynntils.models.items.encoding.data;
 
 import com.wynntils.models.items.encoding.type.ItemData;
 import com.wynntils.models.items.properties.IdentifiableItemProperty;
+import java.util.regex.Pattern;
 
 public record NameData(String name) implements ItemData {
+    private static final int MAX_NAME_LENGTH = 48;
+    private static final Pattern NAME_SANITIZER_PATTERN = Pattern.compile("[^a-zA-Z0-9 ]");
+
+    public static NameData from(String itemName) {
+        // Sanitize the name
+        itemName = itemName.substring(0, Math.min(itemName.length(), MAX_NAME_LENGTH));
+        itemName = NAME_SANITIZER_PATTERN.matcher(itemName).replaceAll("");
+        itemName = itemName.trim().replaceAll("\\s{2,}", " ");
+
+        return new NameData(itemName);
+    }
+
     public static NameData from(IdentifiableItemProperty property) {
-        return new NameData(property.getName());
+        String itemName = property.getName();
+        return from(itemName);
     }
 }
--- a/common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
+++ b/common/src/main/java/com/wynntils/features/chat/ChatItemFeature.java
@@ -78,6 +78,9 @@ public class ChatItemFeature extends Feature {
     @Persisted
     public final Config<Boolean> showPerfectOrDefective = new Config<>(true);
 
+    @Persisted
+    public final Config<Boolean> showCraftedItemCustomNames = new Config<>(false);
+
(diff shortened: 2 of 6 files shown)


common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
```diff
@@ -6,9 +6,23 @@
 
 import com.wynntils.models.items.encoding.type.ItemData;
 import com.wynntils.models.items.properties.IdentifiableItemProperty;
+import java.util.regex.Pattern;
 
 public record NameData(String name) implements ItemData {
+    private static final int MAX_NAME_LENGTH = 48;
+    private static final Pattern NAME_SANITIZER_PATTERN = Pattern.compile("[^a-zA-Z0-9 ]");
```
ryanzhoudev, line 13: I believe we also need to allow `'` and `-`, some default custom crafted names have this

ryanzhoudev: And I don't think we need 0-9, but not sure

ryanzhoudev (commented):
I'm not super keen on having custom names, but I won't outright disagree for it either. Other than allowing people to have some customization or identification of crafted items, what benefit does it have? I would argue that it even makes it less convenient at a glance to see what item type someone is sharing.

I'm not sure Wynncraft would even like this kind of stuff to be possible in a Wynn-related setting. What would someone's first impression be if they saw some Wynncraft gameplay and there were unfiltered items containing rude words in chat? Is this not just a way to bypass chat filter?

I'm ok with adding it back on the conditions that 
1) We check in #consultation that this is okay and allowed
2) We add a disclaimer and mention your kind of "terms" acceptance in the description of the feature, or as a whole other screen that pops up when you enable this.

Of course, you can proceed without my approval vote, but I think at the very least it would be rather inconsiderate of us to put it back without at least asking the Wynn mod team, now that they're aware of the possibilities.

ShadowCat117:
Although this solution would make chat items look strange for people on older versions and also vanilla players. What if we sent named items in a format like this "(encoding::customname) then the feature will be updated to recognise that a custom name has been given to to the item and apply it from there.

Since the name is being sent in chat, it will go through all of Wynn's filters for profanity and disallowed characters

Essentuan:
> Of course, you can proceed without my approval vote, but I think at the very least it would be rather inconsiderate of us to put it back without at least asking the Wynn mod team, now that they're aware of the possibilities.

I was talking to cal_and_ben about this, and he said that if someone abused this to send profanity he would still ban them. Obviously not the same as asking the mod team, but I figured I should still mention it

magicus:
I honestly think https://github.com/Wynntils/Wynntils/pull/2879 is a better solution.

1) The names will always go through normal chat, and will thus be filtered by Wynncraft if needed.

2) It is extremely clear what kind of item is being shared. There was a worry that you could trick or fool other players of buying e.g. a proper mythic by renaming your crafted item ; that will be impossible with my solution.

The name will still be there, but it will be passed as a part of the normal chat message. I am not sure why this is considered "dumbed down".

kristofbolyai:
Okay, but then, I want to extend #2879 so the mod itself can parse the name into the chat item from chat.

magicus:
Okay, that sounds fair enough.

PR #2882, opened 2024-09-21
feat: Add functions for damage in raids, use in raid overlay

ShadowCat117:
Tracks the damage dealt in each room of a raid. The total damage does not equal the total damage stat given by Wynn at the end of raids, ours is usually lower. Unsure why this is, I checked our damage tracking and it doesn't seem to be missing anything so either Wynn is hiding some damage dealt or their stat includes something we don't know about.

Works with all the existing config options.

No milliseconds

Default

Damage hidden still works

Total across multiple rooms

(Almost) End of raid

Chat message


common/src/main/java/com/wynntils/models/raid/RaidModel.java
```diff
@@ -348,6 +388,17 @@ private List<Long> getAllRoomTimes() {
...
+    private List<Long> getAllRoomDamages() {
+        List<Long> allRoomDamages = new ArrayList<>();
+
+        allRoomDamages.add(getRoomDamage(RaidRoomType.CHALLENGE_1));
+        allRoomDamages.add(getRoomDamage(RaidRoomType.CHALLENGE_2));
+        allRoomDamages.add(getRoomDamage(RaidRoomType.CHALLENGE_3));
+        allRoomDamages.add(getRoomDamage(RaidRoomType.BOSS_FIGHT));
+
+        return allRoomDamages;
+    }
```
kristofbolyai, line 400: This is called by functions, right? If so, it's likely wise to cache this list, and return an immutable view `Collections#unmodifiableList` of the cached version.

kristofbolyai (commented):
Really clean code, good to go once damages are cached.

common/src/main/java/com/wynntils/models/raid/RaidModel.java
ShadowCat117: No this is just to get all the damages for the raid completed event, the timers do the same thing

Essentuan:
> The total damage does not equal the total damage stat given by Wynn at the end of raids, ours is usually lower. Unsure why this is

Its because of the mob scaling and it accounts for that in a weird way, idk been a known 'problem' for a while.

kristofbolyai (approved):
Okay, looks good then!

ShadowCat117:
> Its because of the mob scaling and it accounts for that in a weird way, idk been a known 'problem' for a while.

Ah interesting, good to know it's not an issue on our end then.

PR #2884, opened 2024-09-22
feat: unidentified item improvements

phantamanta44:
I have often been annoyed by how hard it is to distinguish between identified items and unboxed unidentified items in the inventory---especially because when you go to identify a bunch of unboxed items, it's easy to accidentally click on your own items. This PR addresses this in two ways:

1. by extending the unidentified items icon so that it can also display icons over unboxed items.
2. by adding a new "prevent mixed rerolls" feature that disallows the insertion of identified and unidentified items simultaneously in an item identifier UI.


common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.INVENTORY)
+public class PreventMixedRerollsFeature extends Feature {
+    private static final String ITEM_IDENTIFIER_NAME = "\udaff\udff8\ue018";
```
kristofbolyai, line 39: Can you capitalize these Unicode chars? That's usually the format they are used in, and we also use that style.

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+    public void onScreenClose(ScreenClosedEvent event) {
+        state = State.NOT_OPEN;
+    }
+
+    @SubscribeEvent
+    public void onUpdateContents(ContainerSetContentEvent.Post event) {
+        if (state == State.NOT_OPEN) return;
+
+        AbstractContainerMenu menu = McUtils.containerMenu();
+        if (menu.containerId != event.getContainerId()) return;
```
kristofbolyai, line 64: Would it be sane to do a null-check for the menu itself? Wynn sending wrong packets, or Minecraft behaving weirdly could spam exceptions, theoretically. 

(If you do add it, add it for all events)

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+        updateState(menu);
+    }
+
+    @SubscribeEvent
+    public void onUpdateSlot(ContainerSetSlotEvent.Post event) {
+        if (state == State.NOT_OPEN) return;
+
+        AbstractContainerMenu menu = McUtils.containerMenu();
+        if (menu.containerId != event.getContainerId() || Ints.indexOf(ITEM_IDENTIFIER_SLOTS, event.getSlot()) == -1)
+            return;
```
kristofbolyai, line 75: Please fix all multi-line ifs to have brackets. We only allow same-line if-s if they fix the line.

kristofbolyai (commented):
Really clean code once again! I just have a single suggestion apart from code-style. What if you could add "shift-click" to override the identification blocking? I think that would make it fully usable, but wouldn't block users in the rare chance they do want to just dump everything in the identifier.

kristofbolyai:
For reference, here is the new added "question-mark" for the "revealed" items.

image

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
magicus: "fix" = "fit" (it took me a while to parse :))

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
```diff
@@ -11,39 +11,62 @@
...
-            Map.entry(GearType.CHESTPLATE, Pair.of(16 * 1, 16 * 0)),
-            Map.entry(GearType.LEGGINGS, Pair.of(16 * 2, 16 * 0)),
-            Map.entry(GearType.BOOTS, Pair.of(16 * 3, 16 * 0)),
-            Map.entry(GearType.MASTERY_TOME, Pair.of(16 * 0, 16 * 3)),
-            Map.entry(GearType.CHARM, Pair.of(16 * 1, 16 * 3)));
+    private static final StyledText QUESTION_MARK_TEXT =
+            StyledText.fromComponent(Component.literal("?").withStyle(ChatFormatting.BOLD));
+    private static final Map<GearType, Pair<Integer, Integer>> TEXTURE_COORDS = new EnumMap<>(GearType.class);
+
+    static {
```
magicus, line 44: Did you do this change out of stylistic reasons, or is there some other rationale? 

While technically very similar in the resulting byte code, I strongly dislike `static` blocks. They make for surprising code behavior imho. In contrast, constructing the map at the point of the `final` declaration makes it very clear what it contains.

kristofbolyai: I see it's not an EnumMap, perhaps that was the reason? If there wasn't any problem with using a regular HashMap, then probably should be reverted.

magicus:
First of all, I really like this idea! This problem has been nagging me for a long time, but never passing the threshold of getting me to do something about it.

There are some parts of the code that I don't understand, though. Either there are issues with the code, or I'm just a bit stupid. :) I'll add comments/questions inline.

magicus:
I don't understand where the new question-mark icon is coming from? I see no added png file. Did we already have such a texture, and if so, where do we use it right now?

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
```diff
@@ -55,41 +78,76 @@ public void onHotbarSlotRender(HotbarSlotRenderEvent.CountPre e) {
...
+                        textureCoords.a(),
+                        textureCoords.b() + yOffset,
+                        16,
+                        16,
+                        texture.width(),
+                        texture.height());
+            };
+        }
+
+        static IconRenderer forText(Function<GearType, StyledText> textMap) {
```
kristofbolyai, line 117: @magicus the question mark is rendered here, as a font.

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.bus.api.SubscribeEvent;
+
+@ConfigCategory(Category.INVENTORY)
+public class PreventMixedRerollsFeature extends Feature {
+    private static final String ITEM_IDENTIFIER_NAME = "\udaff\udff8\ue018";
+    private static final int[] ITEM_IDENTIFIER_SLOTS = {11, 12, 13, 14, 15, 20, 21, 22, 23, 24};
```
magicus, line 40: This looks like the kind of "knowledge" that should be owned by a Model. I'm not sure which one, though. We don't seem to have an IdentifierModel; perhaps it is time to create one. Or do you have any better suggestions, @kristofbolyai?

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onSlotClick(ContainerClickEvent event) {
+        if (state == State.NOT_OPEN || state == State.NO_FILTER) return;
+
+        AbstractContainerMenu menu = event.getContainerMenu();
+        int slotNum = event.getSlotNum();
+        if (slotNum < 0
+                || slotNum >= menu.slots.size()
+                || !isItemFiltered(menu.getSlot(slotNum).getItem())) return;
+        event.setCanceled(true);
```
magicus, line 89: I think the logic is more clear if you do like this:

```suggestion
        if (slotNum < 0 || slotNum >= menu.slots.size())  return;
        if (isItemFiltered(menu.getSlot(slotNum).getItem())) {
            event.setCanceled(true);
        }
```

magicus: We try to write methods like:

* check prereq # 1, if not, return 
* check prereq # 2, if not, return 
* ...
* Now we do the actual code

I see you have grasped that style. I think the difference here is to me, the "actual code" is not to cancel the event, but to "cancel the event if the item is filtered".

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
kristofbolyai: It could live in the container specific class for the Identifier, although we don't have one yet. I forgot what the superclass is called, but it's somewhere around ContainerModel.

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
magicus: Otoh... I noticed now that you have this as a pattern for all these methods. An alternative is to split the if into two: one to validate the range, and one (that is more "tightly bound" to the event cancelling) for the filtering check.

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+                500,
+                18f,
+                18f);
+    }
+
+    @SubscribeEvent(priority = EventPriority.LOWEST)
+    public void onRenderItemTooltip(ItemTooltipRenderEvent.Pre event) {
+        if (state == State.NOT_OPEN || state == State.NO_FILTER || !isItemFiltered(event.getItemStack())) return;
+
+        List<Component> newTooltips = new ArrayList<>(event.getTooltips());
```
magicus, line 114: This will add the tooltip to all items in the UI..?

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+
+        AbstractContainerMenu menu = McUtils.containerMenu();
+        if (menu.containerId != event.getContainerId() || Ints.indexOf(ITEM_IDENTIFIER_SLOTS, event.getSlot()) == -1)
+            return;
+
+        updateState(menu);
+    }
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public void onSlotClick(ContainerClickEvent event) {
```
magicus, line 81: Also, what is your intention with this method? Is it to prevent the user from confirming the identification request? If so, you should only block a single slot, right?

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
```diff
@@ -0,0 +1,157 @@
...
+
+        updateState(menu);
+    }
+
+    @SubscribeEvent
+    public void onUpdateSlot(ContainerSetSlotEvent.Post event) {
+        if (state == State.NOT_OPEN) return;
+
+        AbstractContainerMenu menu = McUtils.containerMenu();
+        if (menu.containerId != event.getContainerId() || Ints.indexOf(ITEM_IDENTIFIER_SLOTS, event.getSlot()) == -1)
```
magicus, line 74: If you make ITEM_IDENTIFIER_SLOTS a List you can use `contains` instead; that will express the intent much clearer.

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
magicus: Yeah... Enum maps are technically somewhat faster, but sacrificing code legibility for performance should only be done if it has a measured impact.

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
```diff
@@ -11,39 +11,62 @@
...
+    }
 
     @Persisted
     public final Config<UnidentifiedItemTextures> texture = new Config<>(UnidentifiedItemTextures.WYNN);
 
+    @Persisted
+    public final Config<Boolean> showOnUnboxed = new Config<>(true);
+
+    @Persisted
+    public final Config<UnidentifiedItemTextures> unboxedTexture = new Config<>(UnidentifiedItemTextures.QUESTION_MARK);
```
magicus, line 68: I don't like how you mix the old UnidentifiedItemTextures with this new function. They serve quite different purposes. For boxed items, the texture draws a "mini object" on top of the box, and the reason we have UnidentifiedItemTextures is to allow users to select between two different themes, one which mimicks Wynn style better, and one with clearer legibility in the small size.

The marker for "unboxed but unidentified" items is very different -- in this case it would make no sense to overlay an icon of what type it is, since the real icon is clearly visible. And in contrast, the question mark would make no sense to put on all boxed items.

I think the question mark is fine as it is -- you don't need to treat it as a texture, but only present the option to draw a question mark or not. If someone at a later stage gets mad at the question mark and wants to provide a different texture, then that is a future refactoring to allow different textures for "unboxed but un-id:d" items.

magicus: Doing this will also allow you to revert several changes to this file, and just add a new event listener that draws a question mark on top of unboxed-but-unidentified objects.

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1370,8 +1373,12 @@
   "feature.wynntils.translation.translationService.name": "Translation Service",
   "feature.wynntils.unidentifiedItemIcon.description": "Adds an icon to unidentified items.",
   "feature.wynntils.unidentifiedItemIcon.name": "Unidentified Item Icon",
+  "feature.wynntils.unidentifiedItemIcon.showOnUnboxed.description": "Should the icon be shown over unboxed unidentified items?",
```
magicus, line 1376:
```suggestion
  "feature.wynntils.unidentifiedItemIcon.showOnUnboxed.description": "Should an icon be shown over unboxed unidentified items?",
```

common/src/main/java/com/wynntils/models/items/properties/UnidentifiedItemProperty.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright © Wynntils 2024.
+ * This file is released under LGPLv3. See LICENSE for full license details.
+ */
+package com.wynntils.models.items.properties;
+
+/**
+ * Represents an item that may or may not be identified at an item identifier.
+ */
+public interface UnidentifiedItemProperty {
```
magicus, line 10: I'm not sure I like these WynnItem hierarchy changes. There's something ... itchy ... about them, that I can't really put my finger to, without thinking a bit more. So let me start by turning the tables: Why do you think you need these changes, instead of just checking for GearBoxItem vs GearItem.isUnidentified()?

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
magicus: I saw that later. Thanks.

common/src/main/java/com/wynntils/models/items/properties/UnidentifiedItemProperty.java
kristofbolyai: I think you mean that having "UnidentifedItemProperty" return false for isUnidentified is really weird.

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
magicus: Yes, that is indeed the correct place. 

Look at e.g. `com.wynntils.models.containers.containers.SeaskipperContainer`; you need to create a similar `IdentifierContainer` (or if you like being verbose, `ItemIdentifierContainer`).

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
phantamanta44: i *think* it should be impossible for this field to be null; the game should be setting this back to the player's inventory container whenever it's cleared from a container screen

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
phantamanta44: it was just a performance improvement. i usually say static init blocks are fine so long as they're placed near the thing they're initializing. otherwise, rather than reverting, i could wrap the old map literal with the `EnumMap(Map<E, V>)` constructor to get the same thing but with better readability and a slightly higher init cost

phantamanta44: or maybe it would even be useful to add a helper somewhere that constructs `EnumMap`s with the same syntax as `Map::ofEntries`

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
phantamanta44: prevents the user from clicking any filtered items to add them to the identification slots

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
phantamanta44: to all filtered items to better indicate that they can't be added. the overlay might be sufficient, though

common/src/main/java/com/wynntils/features/inventory/PreventMixedRerollsFeature.java
phantamanta44: it's probably quite clear at this point that i think too much about premature optimizations lol

common/src/main/java/com/wynntils/features/inventory/UnidentifiedItemIconFeature.java
kristofbolyai: Then, just wrap it in an Enum map. GC during init is fine. Although the positive gain from this change is likely minimal. Don't get us wrong, we like performant code, but not at the expense of readability and maintainability.

common/src/main/java/com/wynntils/models/items/properties/UnidentifiedItemProperty.java
phantamanta44: this way, it also catches tomes and charms. i feel that it would be a bit clunky to manually check every identifiable item type

phantamanta44: i would have liked to call it `IdentifiableItemProperty`, but that's already used for items with known stat ranges, which is not the case for gear boxes

magicus:
@phantamanta44 Did you give up on this? If so, would you mind if someone else took your code to the finish line?

phantamanta44:
oh no, I've just been exceedingly busy for the past few months and I haven't really had the energy for side projects... At any rate, I'd be happy to let someone else work out whatever I was cooking here

kristofbolyai:
Supersed by #3078

PR #2887, opened 2024-09-22
feat: Add the name of crafted items to the string shared when sharing encoded items [skip ci]

magicus:
Solves https://github.com/Wynntils/Wynntils/issues/2878.


kristofbolyai:
@magicus I can't request a review, but I'd like your opinion on the implementation. As far as I'm aware, this leaves no ground for unsanitized input to be used in the mod, even if sanitization in Wynn fails.

common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
```diff
@@ -6,9 +6,22 @@
...
+
     public static NameData from(IdentifiableItemProperty property) {
-        return new NameData(property.getName());
+        return NameData.from(property.getName());
+    }
+
+    public static NameData from(String name) {
+        name = SANITIZE_PATTERN.matcher(name).replaceAll("");
+        name = name.trim();
+        name = name.replaceAll("\\s\\s", " ");
```
magicus: Shouldn't this be
```suggestion
        name = name.replaceAll("\\s+", " ");
```

Otherwise e.g. "\t\t\t" would be " \t", right?

common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
```diff
@@ -6,9 +6,22 @@
 
 import com.wynntils.models.items.encoding.type.ItemData;
 import com.wynntils.models.items.properties.IdentifiableItemProperty;
+import java.util.regex.Pattern;
 
 public record NameData(String name) implements ItemData {
+    private static final int MAX_NAME_LENGTH = 48;
+    private static final Pattern SANITIZE_PATTERN = Pattern.compile("[^a-zA-Z0-9'\\s]");
```
magicus, line 13: Are you sure `-` is not allowed? If not, names like `Uruk-Hai Blade` would not be allowed.

common/src/main/java/com/wynntils/models/items/encoding/impl/item/CraftedConsumableItemTransformer.java
```diff
@@ -60,9 +60,16 @@ public ErrorOr<CraftedConsumableItem> decodeItem(ItemDataMap itemDataMap) {
         level = requirementsData.requirements().level();
 
         // Optional blocks
-        // Unfortunately, we cannot use the NameData from crafted items, since it can be
-        // set to unsuitable values by the users.
-        name = "Crafted " + StringUtils.capitalizeFirst(consumableType.name().toLowerCase(Locale.ROOT));
+        // Warning!: The name data from Crafted items is deliberately removed from the item data map to prevent
```
magicus: `!:` looks a bit odd; remove one of them. Same goes below.

common/src/main/java/com/wynntils/models/items/ItemEncodingModel.java
```diff
@@ -50,4 +53,17 @@ public boolean canEncodeItem(WynnItem wynnItem) {
     public Pattern getEncodedDataPattern() {
         return ENCODED_DATA_PATTERN;
     }
+
+    public String makeItemString(WynnItem wynnItem, EncodedByteBuffer encodedItem) {
+        String itemName = "";
+
+        // Gear items are named, but their names are encoded in the data
+        if (shareItemName.get()
```
magicus, line 61: Why is this not mirroring the check below in line 98 in the registry? It seems like we will add the chat name to everything that has a name (isn't that like kind of anything, e.g. `Potion of Healing`?) as long as it is not a gear item; but we will only use the chat name for crafted items. I think we should only append the chat name if it really is needed and used.

magicus: In fact, I see now that this was how I originally implemented this function...

common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
ShadowCat117: image

Numbers are allowed even though it says otherwise in the message

EnderNon, line 13: Why use pattern match instead of just checking if the name is in the printable characters area?

kristofbolyai: Because Wynn overrides many of those non-latin characters.

magicus: Ok!

common/src/main/java/com/wynntils/models/items/encoding/impl/item/CraftedConsumableItemTransformer.java
magicus:
```suggestion
        // Warning: The name data from Crafted items is deliberately removed from the item data map to prevent
```

common/src/main/java/com/wynntils/models/items/encoding/impl/item/CraftedGearItemTransformer.java
```diff
@@ -73,10 +73,16 @@ public ErrorOr<CraftedGearItem> decodeItem(ItemDataMap itemDataMap) {
         requirements = requirementsData.requirements();
 
         // Optional blocks
-        // Unfortunately, we cannot use the NameData from crafted items, since it can be
-        // set to unsuitable values by the users.
-        name = "Crafted "
-                + StringUtils.capitalizeFirst(gearTypeData.gearType().name().toLowerCase(Locale.ROOT));
+        // Warning!: The name data from Crafted items is deliberately removed from the item data map to prevent
```
magicus:
```suggestion
        // Warning: The name data from Crafted items is deliberately removed from the item data map to prevent
```

common/src/main/java/com/wynntils/models/items/encoding/data/NameData.java
magicus: @kristofbolyai Do you agree with this?

common/src/main/java/com/wynntils/models/items/ItemEncodingModel.java
magicus: @kristofbolyai Is there a problem if I revert this back to how I originally did this?

kristofbolyai: I've done so myself.

kristofbolyai:
@magicus I think we are now in agreement here. This is originally your PR with my changes, so let's both review it.

kristofbolyai (approved):

magicus (commented):
Yeah, looks good. I can't formally approve since this is "my" PR.

PR #2889, opened 2024-09-22
fix: Fix guild map not showing territories

ShadowCat117:
https://github.com/Wynntils/Wynntils/pull/2885 caused TerritoryModel to not have the url loaded by the time it tried to make the call, so moving it to reloadData fixes that.

Also moved the not in guild check to before it actually makes the call


kristofbolyai (approved):
I wonder how many tries it took to get this right for us..

PR #2891, opened 2024-09-22
fix: Don't throw error when clicking outside container GUI [skip ci]

ShadowCat117:
Relevant log

`[Render thread/ERROR] (wynntils) Exception in event listener not belonging to a feature
 java.lang.IndexOutOfBoundsException: Index -999 out of bounds for length 46
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100) ~[?:?]
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106) ~[?:?]
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302) ~[?:?]
	at java.base/java.util.Objects.checkIndex(Objects.java:385) ~[?:?]
	at java.base/java.util.ArrayList.get(ArrayList.java:427) ~[?:?]
	at net.minecraft.core.NonNullList.get(NonNullList.java:47) ~[minecraft-merged-894a333e29-1.21.1-loom.mappings.1_21_1.layered+hash.652182843-v2.jar:?]
	at net.minecraft.world.inventory.AbstractContainerMenu.getSlot(AbstractContainerMenu.java:287) ~[minecraft-merged-894a333e29-1.21.1-loom.mappings.1_21_1.layered+hash.652182843-v2.jar:?]
	at com.wynntils.handlers.inventory.InventoryHandler.onSlotClick(InventoryHandler.java:96)`

PR #2892, opened 2024-09-22
fix: deal with inventory sync edge cases (fixes #2888, addresses #2891)

### LINKED ISSUE #2888: Placing items in misc bucket & bookshelf with inventory sync enables tries to pick up another item
Whilst in the misc bucket or bookshelf containers, placing an item either into the container or your inventory displays the "You cannot place this item in this bank!" message even when placing items that are valid for that container. The issue only occurs whilst the Improved Inventory Sync feature is enabled

https://github.com/user-attachments/assets/07fefea0-4ea8-45b2-a034-a1b6c0542c5a

phantamanta44:
This PR addresses #2888 by first trying to click a placeholder slot (the snow stacks in the bank navigation bar), which should prompt an inventory update without causing the "cannot place this item in this bank!" error. It also correctly handles the `THROW` case for out-of-inventory clicks as in #2891

### DIFF
--- a/common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
@@ -93,15 +93,35 @@ public class InventoryHandler extends Handler {
                 }
             }
             case THROW -> {
-                ItemStack slotStack = menu.getSlot(slotNum).getItem();
-                if (heldStack.isEmpty() && !slotStack.isEmpty()) {
-                    if (isImprovedSyncEnabled()) {
-                        forceSyncLater(menu);
-                        expect(new PendingInteraction.ThrowFromSlot(menu, slotNum, slotStack.copy()));
-                    } else {
-                        ItemStack thrown = event.getMouseButton() == 0 ? slotStack.copyWithCount(1) : slotStack.copy();
-                        WynntilsMod.postEvent(new InventoryInteractionEvent(
-                                menu, new InventoryInteraction.ThrowFromSlot(slotNum, thrown), Confidence.UNCERTAIN));
+                if (slotNum < 0) { // Slot -999 -> tossed an item outside the GUI
+                    if (!heldStack.isEmpty()) { // This probably shouldn't happen, but we'll check anyways
+                        boolean throwAll = event.getMouseButton() == GLFW.GLFW_MOUSE_BUTTON_LEFT;
+                        if (throwAll || heldStack.getCount() == 1) {
+                            if (isImprovedSyncEnabled()) {
+                                forceSyncLater(menu);
+                            } else {
+                                ItemStack thrown = throwAll ? heldStack.copy() : heldStack.copyWithCount(1);
+                                WynntilsMod.postEvent(new InventoryInteractionEvent(
+                                        menu, new InventoryInteraction.ThrowFromHeld(thrown), Confidence.UNCERTAIN));
+                                return;
+                            }
+                        }
+                        expect(new PendingInteraction.ThrowFromHeld(menu, heldStack.copy()));
+                    }
+                } else {
+                    ItemStack slotStack = menu.getSlot(slotNum).getItem();
+                    if (heldStack.isEmpty() && !slotStack.isEmpty()) {
+                        if (isImprovedSyncEnabled()) {
+                            forceSyncLater(menu);
+                            expect(new PendingInteraction.ThrowFromSlot(menu, slotNum, slotStack.copy()));
+                        } else {
+                            ItemStack thrown =
+                                    event.getMouseButton() == 0 ? slotStack.copyWithCount(1) : slotStack.copy();
+                            WynntilsMod.postEvent(new InventoryInteractionEvent(
+                                    menu,
+                                    new InventoryInteraction.ThrowFromSlot(slotNum, thrown),
+                                    Confidence.UNCERTAIN));
+                        }
                     }
                 }
             }
--- a/common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
+++ b/common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
@@ -17,6 +17,7 @@ import com.wynntils.models.items.items.gui.IngredientPouchItem;
 import com.wynntils.models.worlds.event.WorldStateEvent;
 import com.wynntils.utils.mc.McUtils;
 import com.wynntils.utils.type.Confidence;
+import com.wynntils.utils.wynn.InventoryUtils;
 import com.wynntils.utils.wynn.ItemUtils;
 import it.unimi.dsi.fastutil.ints.Int2ObjectMaps;
 import it.unimi.dsi.fastutil.ints.IntArrayList;
@@ -189,7 +190,7 @@ public class InventoryHandler extends Handler {
         }
 
         // Method 2: swap the content book into an empty slot
-        ItemStack contentBookStack = McUtils.player().getInventory().getItem(8);
+        ItemStack contentBookStack = McUtils.player().getInventory().getItem(InventoryUtils.CONTENT_BOOK_SLOT_NUM);
         if (!contentBookStack.isEmpty()) {
             for (int slotNum = 0; slotNum <= menu.slots.size(); slotNum++) {
                 Slot slot = menu.slots.get(slotNum);
@@ -199,7 +200,7 @@ public class InventoryHandler extends Handler {
                             menu.containerId,
                             menu.getStateId(),
                             slotNum,
-                            8,
+                            InventoryUtils.CONTENT_BOOK_SLOT_NUM,
                             ClickType.SWAP,
                             ItemStack.EMPTY,
                             Int2ObjectMaps.emptyMap()));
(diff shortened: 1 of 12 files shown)


common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -147,6 +168,25 @@ public void onWorldStateChange(WorldStateEvent event) {
     }
 
     public void forceSync(AbstractContainerMenu menu) {
+        // Find a non-slot placeholder item in a bank-like inventory
+        // We try this first to avoid the "cannot place in this bank!" error in certain filtered inventories
```
kristofbolyai: Are we sure only banks have these placeholder items?

phantamanta44: reasonably sure yes, but even if other inventories have placeholder items, that would just mean this sync method can be used with those inventories too

phantamanta44:
i would suggest that this PR be pushed through ASAP to avoid more issues like the one mentioned in discord. admittedly the reversed dependency on the item model is quite ugly, but it's getting quite hard to find suitable and generalizable sync methods given how inconsistent Wynncraft's inventory handling seems to be

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -147,41 +170,62 @@ public void onWorldStateChange(WorldStateEvent event) {
...
+        if (!contentBookStack.isEmpty()) {
+            for (int slotNum = 0; slotNum <= menu.slots.size(); slotNum++) {
+                Slot slot = menu.slots.get(slotNum);
+                if (slot.getItem().isEmpty() && slot.mayPlace(contentBookStack)) {
+                    // Swap the book into the slot, which should be no-op, but which will prompt an update packet
+                    McUtils.sendPacket(new ServerboundContainerClickPacket(
+                            menu.containerId,
+                            menu.getStateId(),
+                            slotNum,
+                            8,
```
kristofbolyai: Did you hardcode the content book slot here? Please use a constant from InventoryModel (iirc it has one for this). Also, wouldn't swapping items with the content book open the book itself?

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
```diff
@@ -147,41 +170,62 @@ public void onWorldStateChange(WorldStateEvent event) {
...
+                            Int2ObjectMaps.emptyMap()));
+                    return;
+                }
             }
         }
-        // At this point, our view of the inventory must be completely empty; there's not much we can do
+
+        // Method 3: swap the ingredient pouch into the content book
+        // This is the method of last resort because in certain inventories with filtered interactions (e.g. housing
+        // inventories or blacksmiths), swapping the pouch will prompt an annoying error message in the chat
```
kristofbolyai: are u sure this is a good way to do this? Afaik, opening the ingredient pouch has to be blocked in some inventories, as it can override reward UIs for example, essentially voiding the loot inside.

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: generally swapping the content book into a slot (i.e. cursor over the content book, press a numkey) will open it, but swapping another slot into the content book (i.e. cursor over some other slot, press 9) will not

common/src/main/java/com/wynntils/handlers/inventory/InventoryHandler.java
phantamanta44: from my testing it seems like the ingredient pouch never opens from a `SWAP` click, but i suppose i can't really *guarantee* that given how inconsistent everything's been so far...

kristofbolyai:
I think the wisest hotfix is to disable the feature for now (with StartDisabled) so we have enough time to go over everything related to it and fix it, if it pops up. Can you PR that?

kristofbolyai:
Sorry @phantamanta44, I've been not so clear! I think we should merge a PR and release a version that temporarily disables the feature, so we can go and review this PR in no such hurry. Also, we can likely find some edge-cases, if only us, developers play with it.

phantamanta44:
apparently none of these sync methods work if you're in a reward container GUI with all slots filled. really the modeling issue i'm seeing here is that different GUIs have different sets of working "safe" sync methods, so it's likely that either we'll have to select sync methods based on the specific open GUI or we'll have to include at least one method in the fallback chain that's "unsafe" for at least one GUI. the former is probably the better option, but the implementation will be a bit unpleasant

magicus:
We have systematic "next/previous" page button slots for many containers. I guess adding safe sync slots could be done on a per-container basis in a similar way.

phantamanta44:
the next/previous page buttons are described by `ScrollableContainerProperty` and are therefore part of the container model, but i would argue that the resync method chain associated with a container is not an intrinsic part of the game object, and so it shouldn't be part of the model. in that case, the correct implementation is probably to have a big mapping of `Container` subclass ↦ resync method chain

magicus:
Maybe I misunderstand things, but I thought what you are saying is that you need to know a "safe" slot for every kind of "GUI container" in the game. If so, then that knowledge is definitely part of the container model, and instead of having a huge map (we've tried that before, it was a bad idea), we have split this up into one class per container, that knows everything there is to know about that particular GUI.

phantamanta44:
not so much a "safe slot" as a safe strategy; consider the table in `ImprovedInventorySyncFeature` in this revision, for instance

magicus:
This is getting out of hand...

Now what was supposed to be a simple fix for desync has ballooned into a major thing involving multiple classes. The code is complex and hard to understand. I looked at it and I have no idea if it is correct or not. To me that means taking a single place that looked kind of suspicious but still okay (but which turned out to be broken), and multiplying it manyfold. Now we have like a dozen places that look suspicious and that I cannot say they are okay. The risk here is that we have multiple more places for bugs to hide.

magicus:
At this point, I think we need to step back and ask ourself if the problem this is supposed to solve is so important that it warrants a fix like this. I don't think so. Before your initial PR, I have never even heard of the inventory desync as an issue. Maybe I have a lag-free connection, or whatever, so I have just been blissfully unaware. But my feeling is that this really only became a problem when you wanted the exact events to be sent.

And the inventory sound effects are cute, by all means, but they are in no way enough to warrant a patch like this.

magicus:
I see two ways forward: Either you are able to find a way to properly fix the resync problem in a way that it is possible to review and say "there are obviously no problems here", not "there are no obvious problems here". Or we need to remove this functionality, and you will have to live with the unconfirmed events. 

Going forward with this patch as it looks now is not an option. There are too many places where things can go wrong. There are too many special cases, meaning that the testing burden for keeping this functionality if Wynn changes things are way too high to motivate its existence.

ShadowCat117:
I do have to agree, given the issue that has occured from this feature I don't think it should be kept unless we can be 100% sure that nothing like it can happen again as even with the benefits it brings it is not worthwhile if there's even a 0.1% chance of something going wrong again

magicus:
I am sorry if I sound too harsh @phantamanta44. :-( I have really appreciated your contributions, they have had a code quality and shown an understand of our architecture that few first time contributors have. Really. I have been very happy to see them. ❤️ 

However, as you have now witnessed first-hand, it is way too easy to break the client completely with incorrect implementations, and our Hippocratic attitude here is "First, do no harm". That is, we must not break things or crash the game, or lose peoples' items. And that is basically about risk management. After a while, you get a good nose for what is a risky change. Your initial PR smelled quite a lot on the risk scale, but we accepted it once it was possible to turn off. But this one just "stinks" on the risk scale. 

So I don't want to shoot you down or discourage you from contributing, I just want to minimize the risk for Wynntils users and the long-term development of Wynntils.

phantamanta44:
understood. in that case, would it be acceptable to just patch out the sync feature, but to leave the desync/resync infrastructure (i.e. the `InventoryResyncStrategy` api and `DesyncableContainerClickEvent`) so that they can be consumed by an addon mod?

magicus:
Ooohh... You are thinking of moving this into a "Wynntils add-on"? Hm, maybe that is not such a bad idea. Have you even tested if it works for a separate mod to receive our events? If it does, I think we should have an official way of using add-on mods; at the very least request that they register themselves with us, so we can include information about that in case of a crash or other failure.

If this idea of add-on mods work, then we can discuss what we should expose in terms of the code present here. I am not too happy to keep "dead code" around that is not for the benefit of us but for an add-on, but I can accept some reasonable exposed interface to add-ons.

phantamanta44:
i have not tried it yet, no, and i will probably not have time to work on that sort of thing for a bit. probably the best option for now is to close this PR, open a new PR splicing out the inventory sync feature, and then leave the possible add-on API for future work. what do you think?

kristofbolyai:
Wynntils "add-on" mods do work and they have been around for a while now, but they usually require frequent maintenance as Wynntils doesn't have a version stable core API, meaning little changes in events, mixins, CoreComponents, and Utils can cause major issues on mods depending on them.

magicus:
I agree, it sounds like it best to remove the broken feature for now, and then we can discuss if and how we can support an add-on mod, or if there is some better way to do this within Wynntils that is not so risky.

PR #2893, opened 2024-09-23
fix: Disable ImprovedInventorySyncFeature by default until refinements are made

kristofbolyai:
This should give us proper time to fix the handler, without having to rush different types of fix PRs. Relevant fix PR is #2892.


phantamanta44:
ah good you got to it before i noticed the message

PR #2894, opened 2024-09-24
feat: Add keybind options for opening overlay selection and management screens [skip ci]

ShadowCat117:


magicus (approved):
This is not really the correct feature, but I guess it does not matter that much.

The problem is that the key bind will not be present if this feature is disabled, which makes no sense. We should have non-feature-bound keybinds, but we do not have that currently, so I guess this is the best solution.

magicus:
Hm, I wonder if we should introduce a `CoreFeature` that can't be disabled, that could hold global keybinds like this, and where we could move all the `Config`s that should really be pushed into models.

ShadowCat117:
> The problem is that the key bind will not be present if this feature is disabled, which makes no sense. We should have non-feature-bound keybinds, but we do not have that currently, so I guess this is the best solution.

I did think that was the reason nobody had added these yet but we don't really have an "Overlay Feature" that these can sit in, so this seemed like the best place.

> Hm, I wonder if we should introduce a CoreFeature that can't be disabled, that could hold global keybinds like this, and where we could move all the Configs that should really be pushed into models.

It would be nicer from the user perspective to have the configs in models, so there isn't one option in the settings menu with a large amount of config options

magicus:
> It would be nicer from the user perspective to have the configs in models

Absolutely, but fixing that will require a fair bit of refactoring of our core code. It has been in the works for like 2 years, and I don't think I'm going to be able to finish it soon either. :(

magicus:
But yes, you have a point, I got carried away. We could have key binds in such a generic feature, and storage, but we should not move configs there.

PR #2895, opened 2024-09-25
fix: Fix beta warning feature sometimes failing to initialise

ShadowCat117:
Everywhere else either accesses a Url in a registered download or in some kind of event so this should be the only place where one wasn't

### DIFF
--- a/common/src/main/java/com/wynntils/features/wynntils/BetaWarningFeature.java
+++ b/common/src/main/java/com/wynntils/features/wynntils/BetaWarningFeature.java
@@ -5,14 +5,15 @@
 package com.wynntils.features.wynntils;
 
 import com.wynntils.core.WynntilsMod;
-import com.wynntils.core.components.Handlers;
 import com.wynntils.core.components.Managers;
 import com.wynntils.core.consumers.features.Feature;
 import com.wynntils.core.mod.event.WynncraftConnectionEvent;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/wynntils/features/wynntils/BetaWarningFeature.java
```diff
@@ -39,15 +38,19 @@ public void onDisplayNameChange(PlayerInfoEvent.PlayerDisplayNameChangeEvent e)
...
         warnType = WarnType.NONE;
     }
 
     private enum WarnType {
         NONE(Component.empty()),
         BETA(Component.translatable("feature.wynntils.betaWarning.usingBetaOnNormalServer")),
-        RELEASE(Component.translatable(
-                "feature.wynntils.betaWarning.usingReleaseOnBetaServer",
-                Managers.Url.getUrl(UrlId.LINK_WYNNTILS_DISCORD_INVITE)));
+        RELEASE(Component.translatable("feature.wynntils.betaWarning.usingReleaseOnBetaServer"));
```
kristofbolyai: What if you made this be "supplied"? So just make the enum contain a Supplier<Component>. I think that should work better and is much cleaner. Make sure to include a line about it in comments.

kristofbolyai (approved):

PR #2897, opened 2024-09-26
feat: Added favorite item hotkey [skip ci]

aprevrha:
A hotkey that makes favoriting items faster.

By default the hotkey only works on gear items and ingredients but there is a configuration option that allows favoriting of any item via the hotkey.
image


common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
```diff
@@ -0,0 +1,57 @@
...
+
+        gearItemOpt = Models.Item.asWynnItem(hoveredSlot.getItem(), GearItem.class);
+        if (gearItemOpt.isPresent()) itemName = gearItemOpt.get().getName();
+
+        ingredientItemOpt = Models.Item.asWynnItem(hoveredSlot.getItem(), IngredientItem.class);
+        if (ingredientItemOpt.isPresent()) itemName = ingredientItemOpt.get().getName();
+
+        if (allowFavoritingAllItems.get()) itemName = unformattedName;
+
+        if (itemName != null && !itemName.isBlank()) Services.Favorites.toggleFavorite(itemName);
```
ryanzhoudev: Fix styling (use { }) on all of these please - we only use one liners for simple flow control, eg. return/continue/break.

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
ShadowCat117, line 1: Can this not be added to the existing `ItemFavoriteFeature` why does it need to be separate?

magicus: I agree, that would be a much better fit.

ShadowCat117:
If I remember correctly we have a `NamedItemProperty` which I believe all of the current guide items are instances of so this could be simplified a lot by checking if the current item is one of those. Then if the allow all favourites setting is enabled then you can just use the unformatted name

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
antonprev: Yup agree, I'll do this

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
```diff
@@ -0,0 +1,57 @@
...
+    @Persisted
+    public final Config<Boolean> allowFavoritingAllItems = new Config<>(false);
+
+    private void tryChangeFavoriteStateOnHoveredSlot(Slot hoveredSlot) {
+        if (hoveredSlot == null || !hoveredSlot.hasItem()) return;
+
+        String unformattedName = PATTERN.matcher(
+                        StyledText.fromComponent((hoveredSlot.getItem().getHoverName()))
+                                .getStringWithoutFormatting())
+                .replaceAll("");
```
ShadowCat117: Will `WynnUtils#normalizeBadString` accomplish the same goal as this?

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
antonprev: done.

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
antonprev: This should be fixed now, let me know if there is anything else

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteHotkeyFeature.java
antonprev: Yeah it works and looks a lot nicer.

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
```diff
@@ -4,21 +4,27 @@
  */
 package com.wynntils.features.inventory;
 
+import static com.wynntils.utils.wynn.WynnUtils.normalizeBadString;
```
ShadowCat117: Don't use static imports

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
```diff
@@ -121,4 +137,22 @@ private static void renderFavoriteItem(SlotRenderEvent.Post event) {
...
+                Models.Item.asWynnItemProperty(hoveredSlot.getItem(), NamedItemProperty.class);
+        if (namedItemPropertyOpt.isPresent()) {
+            itemName = namedItemPropertyOpt.get().getName();
+        } else if (allowFavoritingAllItems.get()) {
+            itemName = normalizeBadString(
+                    StyledText.fromComponent((hoveredSlot.getItem().getHoverName()))
+                            .getStringWithoutFormatting());
+        }
+
+        if (itemName != null && !itemName.isBlank()) Services.Favorites.toggleFavorite(itemName);
```
ShadowCat117: Remember the brackets :)

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
antonprev: uh ok

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
antonprev: those sneaky one liners ay :)

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
antonprev: fixed :)

common/src/main/java/com/wynntils/features/inventory/ItemFavoriteFeature.java
antonprev: Hope there are no more formatting errors. I'm new to this modding :D

ShadowCat117 (approved):
Looks good to me. Thank you for the contribution! :)

magicus (approved):
This looks great now! Thanks for the contribition, and for working with us to get the patch to this shape.

PR #2898, opened 2024-09-29
chore: Remove ImprovedInventorySync and ItemInteractionSounds features [skip ci]

ShadowCat117:
Not a full revert of https://github.com/Wynntils/Wynntils/pull/2871/, FixStackSizeFeature did not rely on inventory sync and is a good addition, plus a few util features have been left


kristofbolyai (approved):
Thank you. Sad that we have to remove this, but it's hard to implement this reliably.

PR #2900, opened 2024-09-30
chore: Make original unwrapped styledtext available in chat messages [skip ci]

cmkohnen:
With chat wrapping being present in more chat messages, store the original unwrapped chat message in the chat event as well. At the moment this is only used by two methods, but there are a few regexes that are broken right now, which would rely on this as well.

### DIFF
--- a/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
+++ b/common/src/main/java/com/wynntils/features/redirects/ChatRedirectFeature.java
@@ -15,7 +15,6 @@ import com.wynntils.handlers.chat.event.ChatMessageReceivedEvent;
 import com.wynntils.handlers.chat.type.MessageType;
 import com.wynntils.models.players.type.PlayerRank;
 import com.wynntils.utils.StringUtils;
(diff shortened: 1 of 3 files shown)


ShadowCat117 (approved):

kristofbolyai (commented):
Give me a moment to think about this, the long method name and the various types of styled texts in the event puts a lot of mental overhead in the code, and makes it easy to "missread" code by not actually processing a word in the getter and thinking of an alternative styled text.

cmkohnen:
> Give me a moment to think about this, the long method name and the various types of styled texts in the event puts a lot of mental overhead in the code, and makes it easy to "missread" code by not actually processing a word in the getter and thinking of an alternative styled text.

Hmmm, would anything getting just the original styledtext break if that were also unwrapped? More specifically would unwrapping non-wrapped messages have side effects?
If not, it might be worth removing the `getOriginalStyledText()` method and rely on `getOriginalUnwrappedStyledText()`.

magicus:
Do we ever need the original but not unwrapped message?

kristofbolyai:
I don't think this is really necessary for now.

PR #2901, opened 2024-09-30
feat: Add texture options for item highlights [skip ci]

ShadowCat117:
Transparent circle (current), opaque circle, transparent box, opaque box and 2 gradient box options
Highlights

More can easily be added, render method will need slight changes if enough textures are added to move onto a 2nd row but this is fine for now


kristofbolyai (approved):
Looks good!

PR #2907, opened 2024-10-05
feat: Hide original activity beacon & marker [skip ci]

ShadowCat117:
Doesn't conflict with lootrun beacons by hiding those and vice versa if original lootrun beacons are hidden.

I think it's been mentioned before how BeaconModel was a bit too tied to LootrunModel so now it's more separate as it handles multiple beacon types


common/src/main/java/com/wynntils/features/combat/ContentTrackerFeature.java
```diff
@@ -31,6 +31,9 @@ public class ContentTrackerFeature extends Feature {
     @Persisted
     public final Config<Boolean> showAdditionalTextInWorld = new Config<>(true);
 
+    @Persisted
+    public final Config<Boolean> hideOriginalMakrer = new Config<>(true);
```
magicus: typo

ShadowCat117: Oops ty

common/src/main/java/com/wynntils/models/activities/beacons/ActivityBeaconMarkerKind.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final String MARKER_PREFIX = "\uDAFF\uDFF8[\uE010-\uE014]\uDAFF\uDFDE";
+    private static final String MARKER_SUFFIX = "(\n\\d+m (\uE000|\uE001)?)?";
+
+    private final Pattern iconPattern;
+
+    ActivityBeaconMarkerKind(String iconCharacter) {
+        this.iconPattern = Pattern.compile(MARKER_PREFIX + iconCharacter + MARKER_SUFFIX);
+    }
+
+    public boolean matches(StyledText styledText) {
```
kristofbolyai, line 33:
```suggestion
    @Override
    public boolean matches(StyledText styledText) {
```

common/src/main/java/com/wynntils/models/activities/beacons/ActivityBeaconKind.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private final int customModelData;
+    private final CustomColor customColor;
+
+    ActivityBeaconKind(int customModelData, CustomColor customColor) {
+        this.customModelData = customModelData;
+        this.customColor = customColor;
+    }
+
+    public boolean matches(ItemStack itemStack) {
```
kristofbolyai, line 38:
```suggestion
    @Override
    public boolean matches(ItemStack itemStack) {
```

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -6,26 +6,42 @@
...
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import net.minecraft.world.item.alchemy.PotionContents;
+import net.minecraft.world.item.component.CustomModelData;
 import net.neoforged.bus.api.EventPriority;
 import net.neoforged.bus.api.SubscribeEvent;
 
 public class BeaconModel extends Model {
+    public static final int COLOR_CUSTOM_MODEL_DATA = 79;
```
kristofbolyai, line 38: Where is this int exactly? Is it not in the MC source code as a constant?

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
```diff
@@ -72,5 +107,62 @@ public void onEntityRemoved(RemoveEntitiesEvent event) {
...
+        });
+    }
+
+    public void registerBeacon(String name, BeaconKind beaconKind) {
+        beaconRegistry.put(name, beaconKind);
+    }
+
+    public void registerBeaconMarker(String name, BeaconMarkerKind beaconMarkerKind) {
+        beaconMarkerRegistry.put(name, beaconMarkerKind);
+    }
```
kristofbolyai: I think the interfaces could have `name` methods on it, and when implementing on an enum, it'll work automatically.

kristofbolyai (commented):
Is it not reasonable to assume that a beacon and a marker comes "together"? It feels a bit weird to separate them.

Secondly, now that Activity and Lootrun models depend on BeaconModel, can you add them as explicit dependencies in their ctrors/Models.java?

common/src/main/java/com/wynntils/models/beacons/BeaconModel.java
ShadowCat117: It was in the original `LootrunBeaconKind` as that would do the check for unknown beacon kind but now it's done in the model

ShadowCat117:
> Is it not reasonable to assume that a beacon and a marker comes "together"? It feels a bit weird to separate them.

Yeah I guess they can be grouped together, I'll look into that tomorrow

magicus:
> Secondly, now that Activity and Lootrun models depend on BeaconModel, can you add them as explicit dependencies in their ctrors/Models.java?

Such dependencies need/should only be marked if the dependency is needed during construction. I did a pass some time ago and removed all other dependencies. It is fully okay for a model to use any other model at any time after construction, and that need not be declared.

common/src/main/java/com/wynntils/models/activities/beacons/ActivityBeaconKind.java
```diff
@@ -0,0 +1,51 @@
...
+import com.wynntils.utils.colors.CommonColors;
+import com.wynntils.utils.colors.CustomColor;
+import net.minecraft.core.component.DataComponents;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import net.minecraft.world.item.alchemy.PotionContents;
+import net.minecraft.world.item.component.CustomModelData;
+
+public enum ActivityBeaconKind implements BeaconKind {
+    QUEST(Models.Beacon.COLOR_CUSTOM_MODEL_DATA, CustomColor.fromInt(2739350)),
```
magicus: Can you please convert the color codes to hex?

ShadowCat117: It's easier to keep them as int as it's being compared against the color of the potion which we get as an int but can change to hex if you prefer

ShadowCat117:
I don't think we can assume the beacon and marker come together actually, `/toggle beacon` will hide the beacon but the marker will remain

common/src/main/java/com/wynntils/models/activities/beacons/ActivityBeaconKind.java
ShadowCat117: I will change it actually as LootrunBeaconKind does it that way

magicus: I just want to point out that both 123 and 0xa4 represent an `int` -- it is just different ways to express the literal value of the same underlying type. So chosing decimal or hexadecimal representation in the source code is just a matter of which is easiest to read. And color codes are traditionally written in hex, to clarify the RRGGBB components.

kristofbolyai (approved):
Looks good! Thanks.

PR #2908, opened 2024-10-06
feat: Quick Search TM feature

aprevrha:
Quality of life trade market feature that allows quickly searching for an item.
Please do tell me what you think, suggestions on how to fix my wacky code are welcomed :)

ToDo:
- [x] Figure out if and how to do the history - postponed
- [x] Option to add searches to chat history - postponed
- [x] Fix / Change everything that needs to be changed

https://github.com/user-attachments/assets/a928975b-7360-49f4-86c4-a217570a1894


magicus:
So you are adding functionality that adds a new hotkey, and if you press that while hovering an item in the trademarket, it will automatically search for that item's name?

It seems like you are also planning to add more functionality. That is all good, but I think you should do that as separate PRs. Let's focus on this functionality first.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @SubscribeEvent
+    public void onScreenOpen(ScreenOpenedEvent.Post event) {
+        if (!(Models.Container.getCurrentContainer() instanceof TradeMarketContainer)) {
+            inTradeMarket = false;
+            quickSearching = false;
+            return;
+        }
+        inTradeMarket = true;
+        WynntilsMod.info("TM Opened!");
```
magicus: This looks like debug support.

magicus:
The patch seems to meet all formal requirement wrt to code style etc. But there is something that bothers me with this PR, that I can't really put my finger to. Maybe it is just that the feature is doing stuff that really belongs in a TrademarketModel, see the discussion in https://github.com/Wynntils/Wynntils/pull/2910. Or maybe it is just nothing, and the code is fine.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: Yup Removed

aprevrha:
> Maybe it is just that the feature is doing stuff that really belongs in a TrademarketModel, see the discussion in #2910. Or maybe it is just nothing, and the code is fine.

Yes agree stuff should be moved to the Model can this be done in a separate PR?

magicus:
I think I'll leave it for bolyai and the others to evaluate this PR.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private static final Pattern TYPE_TO_CHAT_PATTERN = Pattern.compile(
+            "^§5(\uE00A\uE002|\uE001) \n\uE001 Type the .* or type (\n\uE001 'cancel' to|'cancel' to \n\uE001) cancel:\n\uE001 ");
+    // Maybe there is a better solution for this than regex, but the TM is very peculiar.
+    // \\[.*?\\] Crafting stuff, \\[\\uE000-\\uF8FF\\] Gathering Tools, \\u2B21 Shiny
+    private static final Pattern CUT_PATTERN =
+            Pattern.compile("(Emerald Pouch|\\[.*?\\]|[\\uE000-\\uF8FF]+|\\u2B21)\\s*");
+
+    private static final Pattern POTION_PATTERN = Pattern.compile("^Potion of (. \\w+)");
+    private static final Pattern EOL_PATTERN = Pattern.compile("À+$");
```
kristofbolyai: I think the use of these regexes is what bothers magicus, the fact that so many items need special handling done. 

My overall feeling is that this feature would be better if the old custom trade market searching worked, and overall, it would be great if we could wrap the trade market as a whole with a custom UI. However, I'm really afraid of doing that just because the impact of anything breaking is so significant and the code is really hard to review and manage.

antonprev: The regex seems a bit over done, but there are just some very weird edge cases with how the trade market search works. And I think I've done a pretty good job at making every item you could find on the tm searchable with quick search, tell me if I missed something or if there's a better way.
Yeah ofc this would be better with a whole custom UI, we wouldn't need use the chat and could have auto completion as someone mentioned. We’ll cross that bridge when we get there. 

I'd argue that this already provides value to the player, even if it's just QOL and not a complete overhaul.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,139 @@
...
+@ConfigCategory(Category.TRADEMARKET)
+public class TradeMarketQuickSearchFeature extends Feature {
+    @Persisted
+    public final Config<Boolean> instantSearch = new Config<>(true);
+
+    @Persisted
+    public final Config<Boolean> hidePrompt = new Config<>(true);
+
+    @Persisted
+    public final Config<Boolean> autoCancel = new Config<>(true);
```
ShadowCat117: Honestly I don't think any of these are necessary. Instant search should just be always on, I can't imagine a scenario where someone would want to quick search something but not have it be instant and require pressing enter to send the chat message.

If the message is sent instantly the prompt can always be hidden and you'll never need to cancel sending the search query

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,139 @@
...
+    private final KeyBind quickSearchKeyBind = new KeyBind(
+            "Quick Search TM",
+            GLFW.GLFW_MOUSE_BUTTON_MIDDLE,
+            InputConstants.Type.MOUSE,
+            true,
+            null,
+            this::tryQuickSearch);
+
+    private static final Pattern TYPE_TO_CHAT_PATTERN = Pattern.compile(
+            "^§5(\uE00A\uE002|\uE001) \n\uE001 Type the .* or type (\n\uE001 'cancel' to|'cancel' to \n\uE001) cancel:\n\uE001 ");
```
ShadowCat117: The pattern from `TradeMarketAutoOpenChatFeature` is a little nicer so I'd say use that here too. Then use `StyledTextUtils.unwrap(event.getOriginalStyledText()).stripAlignment()` to check for the match.

Annoying to duplicate the pattern but as mentioned before ideally this should all be handled in the model

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: Yeah it's the default, the shift to bring up the chat is what the other two options were originally for, but it was easy to add the option so I did for anyone who wants it, but ur right not a lot of people will use it


With instant search the prompt is always hidden except when opening the chat with shift

Sometimes it's good to have the option to edit the search, say if you want to delete "Unidentified" from the beginning

antonprev: Maybe use the hide prompt and auto cancel from the auto open chat feature? But that's also kinda janky

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,139 @@
...
+            McUtils.mc().setScreen(new ChatScreen(searchQuery));
+            if (hidePrompt.get()) {
+                event.setCanceled(true);
+            }
+            inSearchChat = true;
+        } else {
+            event.setCanceled(true);
+            McUtils.sendChat(searchQuery);
+        }
+    }
```
ShadowCat117: This doesn't need to high priority as the default will be above the low of the hide prompt option but it's fine. But do please put this method above `tryQuickSearch` as we usually put the @SubscribeEvent methods first

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
ShadowCat117: That's also partly why I don't like this, as it's 2 features doing the same thing. I think personally I would say remove these for now, then when the day comes where we finally get configs in models, the hide prompt and auto cancel could be moved there so it will apply to both

antonprev: Removing an option no one would use anyway (instant search off) I guess, tho why not leave it in as an option. But while this is not all in the model the other two are somewhat important. They were the reason I made the same option for "auto open chat"

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: Fixed [c52aaa4](https://github.com/Wynntils/Wynntils/pull/2908/commits/c52aaa467e843e76de65393e510514b171dda91b)

common/src/main/resources/assets/wynntils/lang/en_us.json
```diff
@@ -1373,6 +1373,14 @@
   "feature.wynntils.tradeMarketPriceMatch.undercutBy.description": "How much should the buttons undercut the existing prices by? Set to 0 to match the existing prices.",
   "feature.wynntils.tradeMarketPriceMatch.undercutBy.name": "Undercut By",
   "feature.wynntils.tradeMarketPriceMatch.youReceive": "You receive: ",
+  "feature.wynntils.tradeMarketQuickSearch.autoCancel.description": "Cancels the search if chat is exited.",
+  "feature.wynntils.tradeMarketQuickSearch.autoCancel.name": "Auto Cancel",
+  "feature.wynntils.tradeMarketQuickSearch.description": "Adds a hotkey to instantly seach for an item on the trade market.",
```
```suggestion
  "feature.wynntils.tradeMarketQuickSearch.description": "Adds a hotkey to instantly search for an item on the trade market.",
```

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,139 @@
...
+        if (!inSearchChat || !(event.getScreen() instanceof ChatScreen)) return;
+        if (autoCancel.get() && KeyboardUtils.isKeyDown(GLFW.GLFW_KEY_ESCAPE)) {
+            McUtils.sendChat("cancel");
+        }
+        inSearchChat = false;
+    }
+
+    // EventPriority.HIGH so that InventoryEmeraldCountFeature does not render.
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void MenuClosedEvent(MenuClosedEvent event) {
```
kristofbolyai:
```suggestion
    public void onMenuClosed(MenuClosedEvent event) {
```

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
kristofbolyai: Okay, we added auto cancel and hide prompt in the other feature. Why would we want to duplicate the same thing here? 

I've now checked the code and I realized that this and `TradeMarketAutoOpenChatFeature` inheritedly wants to be the same feature. Please merge the two into `TradeMarketQuickSearchFeature` (admittedly, I like your feature name better here). That'll also get rid of the duplicate configs. Me or @ShadowCat117 can help with a config upfixer for persisting the enabled state of the old feature. Finally, sure, let's, keep `instantSearch`.

kristofbolyai: Also, as Shadowcat mentioned in other places, the original feature is more refined, so please try to take everything over, and use that feature as a base, and extend this feature's changes into it.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
kristofbolyai: My problem is not with the regex existing, it's the fact that there is no explanation of the "problem" they try to solve. Does the TM not accept `[]` for pouches? Can I not search for `Potion of Health`? What do those unicode ranges stand for, what do they represent?

Don't get me wrong, I very much disagree with people writing 30 line comment-essays, most of the time, it means that the code was not properly designed, and they needed those comments to compromise on the lack of design code wise. However, in a case like this, we do need some kind of documentation on the issue itself. So, please, try to explain it, not here, but in code comments, in a reasonable way. Thank you :)

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: Okay if you put it that way, yeah I guess the whole opening chat automatically feature could be moved to Quick Search.
I just removed the options here as @ShadowCat117 was saying because I realized he was 100% correct because the options just worked out of the box even if they were disabled in QS.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: [6653343](https://github.com/Wynntils/Wynntils/pull/2908/commits/6653343f32347c662583cc701210cbf9074727e2)

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: It's all in Quick Search now :)

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,183 @@
...
+            Pattern.compile("(Emerald Pouch|\\[.*?\\]|[\\uE000-\\uF8FF]+|Unidentified \\u2B21|\\u2B21)\\s*");
+
+    // Similarly to the Pouches, the elemental/skill potions cannot be found by their full name (Other potions can).
+    // If we are searching such a potion, we simply search the second part. '.' matches the icon and '\w+' the skill.
+    // e.g. "Potion of ✤ Strength [2/2]" -> "Potion of ✤ Strength" -> "✤ Strength"
+    private static final Pattern POTION_PATTERN = Pattern.compile("^Potion of (. \\w+)");
+
+    // 'À' this char is behind all the items names and needs to be trimmed before we search.
+    // We cannot use WynnUtils.normalizeBadString() as this messes up dungeon keys.
+    // e.g. "CorruptedÀÀÀGalleon'sÀÀÀGraveyard KeyÀ" -> "CorruptedÀÀÀGalleon'sÀÀÀGraveyard Key"
```
magicus, line 83: What does this mean? That you need to search for "CorruptedÀÀÀGalleon'sÀÀÀGraveyard" instead of "Corrupted Galleon's Graveyard"? That surprises me, but if it really is the case then I guess that is what it is.

Or does it mean that normalizeBadString does not work properly? I would have assumed that it would generate "Corrupted Galleon's Graveyard" from "CorruptedÀÀÀGalleon'sÀÀÀGraveyard"  -- is that not so?

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,183 @@
...
+    private String searchQuery;
+    private boolean inTradeMarket = false;
+    private boolean inSearchChat = false;
+    private boolean openChatWhenContainerClosed = false;
+    private boolean quickSearching = false;
+    private boolean instantSearching = false;
+
+    @SubscribeEvent
+    public void onScreenOpen(ScreenOpenedEvent.Post event) {
+        WynntilsMod.info("Screen Opened");
```
magicus: Remove these logging outputs.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,183 @@
...
+            openChatWhenContainerClosed = false;
+            inSearchChat = true;
+        }
+    }
+
+    // EventPriority.HIGH so that InventoryEmeraldCountFeature does not render.
+    @SubscribeEvent(priority = EventPriority.HIGH)
+    public void onMenuClosed(MenuClosedEvent event) {
+        WynntilsMod.info("Menu Closed");
+        if (!inTradeMarket) return;
```
magicus:
```suggestion
        if (!inTradeMarket) return;

```

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: Yeah the market is weird like that, you need to search with the ÀÀÀ but not the À at the end, took me ages to figure out because I was using normalizeBadStrings at first.

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,183 @@
...
+    public void onScreenOpen(ScreenOpenedEvent.Post event) {
+        WynntilsMod.info("Screen Opened");
+        openChatWhenContainerClosed = false;
+        if ((Models.Container.getCurrentContainer() instanceof TradeMarketContainer)) {
+            inTradeMarket = true;
+            quickSearching = false;
+            inSearchChat = false;
+            return;
+        }
+        inTradeMarket = false;
```
magicus:
```suggestion
        inTradeMarket = (Models.Container.getCurrentContainer() instanceof TradeMarketContainer);
        if (!inTradeMarket) return;
        
        quickSearching = false;
        inSearchChat = false;
```

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    @SubscribeEvent
+    public void onChatMessageReceive(ChatMessageReceivedEvent event) {
+        if (!Models.WorldState.onWorld()) return;
+        StyledText styledText =
+                StyledTextUtils.unwrap(event.getOriginalStyledText()).stripAlignment();
+        if (!styledText.matches(TYPE_TO_CHAT_PATTERN)) return;
+        if (instantSearching) {
```
magicus, line 145: Please leave a blank line after the set of checks for early return, not just here but everywhere in the code.

magicus: The methods should be divided in two parts, first check if we really should be here, otherwise return, then if everything passes, do what we should do.

ShadowCat117 (commented):
You still need to add the upfixer so that configs from the auto open chat feature will be changed for the new feature. Look at other `RenamedPrefixesUpfixer` to see what to do

common/src/main/java/com/wynntils/features/trademarket/TradeMarketQuickSearchFeature.java
antonprev: [651e713](https://github.com/Wynntils/Wynntils/pull/2908/commits/651e713613b0c0675869651890dd5a0aab82988b)

aprevrha:
> You still need to add the upfixer so that configs from the auto open chat feature will be changed for the new feature. Look at other `RenamedPrefixesUpfixer` to see what to do

I hope I did it correctly... There is still the "tradeMarketAutoOpenChatFeature.userEnabled" field in the config, but the options seem to work.

Also there are some semantics that could be confusing now. It's called Quick Search, but also does other stuff, never even stated that it auto opens the chat, not only when searching mind you...
Just looking forward to when models can have configs.

ShadowCat117:
Upfixer looks fine, I'll double check it tomorrow. As for no mention of auto opening, you could change the feature description to mention that as this feature is no longer just the keybind

kristofbolyai:
> There is still the "tradeMarketAutoOpenChatFeature.userEnabled" field in the config, but the options seem to work.

That doesn't sound too good though, does it..?

common/src/main/java/com/wynntils/core/persisted/upfixers/UpfixerManager.java
```diff
@@ -58,6 +59,7 @@ public UpfixerManager() {
         registerConfigUpfixer(new NpcDialoguesOverlayConfigsMovedUpfixer());
         registerConfigUpfixer(new TowerAuraVignetteNameUpfixer());
         registerConfigUpfixer(new TowerAuraVignetteAndOverlayMovedToCommonFeature());
+        registerConfigUpfixer(new TradeMarketAutoOpenChatToTradeMarketQuickSearchUpfixer());
```
kristofbolyai: warning! This is not alphabetical ordering! These upfixers are run in order, so you should move this down to be the last one.

kristofbolyai (approved):
I think this looks much better now! Thanks.

common/src/main/java/com/wynntils/core/persisted/upfixers/UpfixerManager.java
antonprev: [ec634d5](https://github.com/Wynntils/Wynntils/pull/2908/commits/ec634d5e280ae9ca0ccffe7119e63de4ce586476)

ShadowCat117 (approved):
Upfixer works fine, so this looks good to me! Thanks for doing the requested changes, it's a lot nicer having this in one feature!

PR #2909, opened 2024-10-06
feat: Trade Market Auto Cancel option [skip ci]

aprevrha:
parity for #2908


ShadowCat117:
Could this be enabled by default? Seems to make sense to me as you aren't able to reopen the tm screen until entering something

kristofbolyai (approved):
Looks good, I think it can be enabled by default.

magicus:
I think we should have a GUI element where you can write the search string. I understand why Wynncraft has to resort to chat to support vanilla clients, but we can do so much better. With a specialized GUI with text input, we can e.g. have a search history, integrate with our item DB to tab expand names, etc.

PR #2910, opened 2024-10-06
feat: Trade Market Hide Prompt option [skip ci]

aprevrha:
parity for #2908


aprevrha:
Just noticed this clashes with quick search (#2908)

aprevrha:
fixed with [5a269d6](https://github.com/Wynntils/Wynntils/pull/2908/commits/5a269d6e0f7d1c75f3782f630b6fe48dde96f956)

common/src/main/java/com/wynntils/features/trademarket/TradeMarketAutoOpenChatFeature.java
```diff
@@ -34,6 +40,9 @@ public void onChatMessageReceive(ChatMessageReceivedEvent event) {
 
         if (styledText.matches(TYPE_TO_CHAT_PATTERN)) {
             openChatWhenContainerClosed = true;
+            if (hidePrompt.get()) {
+                event.setCanceled(true);
```
ShadowCat117, line 51: As other features rely on this chat message, you should set the event priority to lower so this doesn't cancel it before they use it

magicus:
With all these trademarket changes going on, I believe the time has come to create a `TrademarketModel`. There is too much knowledge about the trademarket spread over too many features, and now they start to step on each other's toes.

A TrademarketModel could have knowledge about what state the TM is in (as in which screen is currently shown), perform searches on request, etc.

ShadowCat117:
There already is a `TradeMarketModel` but I believe it's mainly used for the custom tm screen feature which has yet to be updated for 2.1

common/src/main/java/com/wynntils/features/trademarket/TradeMarketAutoOpenChatFeature.java
antonprev: Done, set it to LOW

aprevrha:
> A TrademarketModel could have knowledge about what state the TM is in (as in which screen is currently shown), perform searches on request, etc.

> There already is a `TradeMarketModel` but I believe it's mainly used for the custom tm screen feature which has yet to be updated for 2.1

Agreed, the states should all be in the model. Updating the custom tm screen would be the perfect opportunity to do so. 
I haven't worked with models yet, but could get started, will need some help tho :)

For now I'd like to merge this anyways as the issue should be temporarily fixed

ShadowCat117:
> I haven't worked with models yet, but could get started, will need some help tho :)

They're fairly straightforward, essentially it's just collecting information about a certain Wynncraft related system in one place that other components can access. Raid model for example does everything to do with tracking when you are in a raid, what room you are in, how long you've been in that room for etc. Then the functions just query the model for that information instead of them handling it themselves.

As always if you need any assistance the discord channel is there and someone will be around

ShadowCat117 (approved):
Thanks!

